PR #23, opened 2022-11-22
fix: Use os.TempDir() for creating temporary anchore/grype config

mkilchhofer:
I prepare a contribution to use read-only root filesystem in every kubescape component. Unfortunately, kubevuln writes the temporary grype config to the same directory where the bundled config is living and therefore there is no option to mount a temporary filesystem below this folder.

By moving the temporary config to /tmp (`os.TempDir()`) we can mount a kuberentes `emptyDir` volume to /tmp and we're done :-)

/cc: @dwertent


Bezbran (approved):

mkilchhofer:
can we merge this?

Bezbran:
> can we merge this?

Yep

PR #33, opened 2023-02-09
use hexagonal architecture

matthyx:
This is the blog post that sparked the idea: https://medium.com/@matiasvarela/hexagonal-architecture-in-go-cfd4e436faa3


matthyx:
@vladklokun current status:
- layout is pretty much final
- syft and grype integration are working
- I need to refactor HTTP API to match our latest designs
- same for scan workflow

matthyx:
cc @soft4rchitecture

adapters/v1/grype.go
```diff
@@ -0,0 +1,147 @@
...
+	}, nil
+}
+
+func (g *GrypeAdapter) UpdateDB(ctx context.Context) error {
+	g.mu.Lock()
+	defer g.mu.Unlock()
+	ctx, span := otel.Tracer("").Start(ctx, "UpdateDB")
+	defer span.End()
+	var err error
+	g.store, g.dbStatus, g.dbCloser, err = grype.LoadVulnerabilityDB(g.dbConfig, true)
```
vladklokun, line 140: issue: `LoadVulnerabilityDB` visits the `dbConfig.ListingURL`

This makes a live request, and if we exceed our quota or something like that, our tests will fail or will be flaky otherwise. Please see the comment regarding `dbConfig`.

adapters/v1/syft.go
```diff
@@ -0,0 +1,95 @@
...
+func NewSyftAdapter() *SyftAdapter {
+	return &SyftAdapter{}
+}
+
+func (s *SyftAdapter) CreateSBOM(ctx context.Context, imageID string, options domain.RegistryOptions) (domain.SBOM, error) {
+	ctx, span := otel.Tracer("").Start(ctx, "CreateSBOM")
+	defer span.End()
+	// ensure only one SBOM is created at a time
+	s.mu.Lock()
+	defer s.mu.Unlock()
```
vladklokun, line 34: question: why do we have to create only one SBOM at a time?

If I understand correctly, since we lock the mutex for the entire adapter, we would only be able to create an SBOM for one image at a time.

Why should we not have the ability to create SBOM for `image1` and `image2` concurrently?

adapters/v1/syft.go
```diff
@@ -0,0 +1,95 @@
...
+	// translate business models in syft models
+	userInput := "registry:" + imageID
+	sourceInput, err := source.ParseInput(userInput, "", true)
+	if err != nil {
+		panic(err)
+	}
```
vladklokun, line 40: question: is there a reason we panic here, instead of returning the `err`?

Is this a truly unrecoverable error that the whole application has to drop dead? If I am trying to wrap SBOM generation in an HTTP API and a user provides me a malformed input, I would expect to return them an error that tells them their input is invalid, not terminate on the spot.

adapters/v1/syft.go
```diff
@@ -0,0 +1,95 @@
...
+		InsecureSkipTLSVerify: options.InsecureSkipTLSVerify,
+		InsecureUseHTTP:       options.InsecureUseHTTP,
+		Credentials:           credentials,
+		Platform:              options.Platform,
+	}
+	// download image
+	src, cleanup, err := source.New(*sourceInput, registryOptions, []string{})
+	defer cleanup()
+	if err != nil {
+		panic(err)
```
vladklokun, line 60: question: do we have to panic here?

adapters/v1/syft.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+	// extract packages
+	catalogOptions := cataloger.Config{
+		Search: cataloger.DefaultSearchConfig(),
+	}
+	pkgCatalog, relationships, actualDistro, err := syft.CatalogPackages(src, catalogOptions)
+	if err != nil {
+		panic(err)
```
vladklokun, line 68: question: do we have to panic here?

adapters/v1/syft_test.go
```diff
@@ -0,0 +1,50 @@
...
+func Test_syftAdapter_CreateSBOM(t *testing.T) {
+	tests := []struct {
+		name    string
+		imageID string
+		format  string
+		options domain.RegistryOptions
+		wantErr bool
+	}{
+		{
+			name:    "hello-world",
```
vladklokun, line 27: suggestion: let’s use a more explicit name for the test case

The idea is the same as the one in a previous comment related to Grype.

adapters/v1/syft_test.go
```diff
@@ -0,0 +1,50 @@
...
+			imageID: "library/hello-world@sha256:aa0cc8055b82dc2509bed2e19b275c8f463506616377219d9642221ab53cf9fe",
+			format:  string(fileContent("testdata/hello-world-sbom.format.json")),
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			s := NewSyftAdapter()
```
vladklokun, line 34: question: is there a reason that the Grype adapter returned an error, while this one does not?

I believe they both solve a similar problem. With Grype, something can go wrong, but with Syft, it can’t?

adapters/v1/grype_test.go
```diff
@@ -0,0 +1,56 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			g, err := NewGrypeAdapter(context.TODO())
+			assert.Assert(t, err == nil)
+			got, err := g.ScanSBOM(context.TODO(), tt.sbom)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("ScanSBOM() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			ja := jsonassert.New(t)
+			ja.Assertf(string(got.Content), tt.format)
```
vladklokun, line 46: praise: that’s a useful thing I didn’t know existed!

cmd/cli/main.go
```diff
@@ -0,0 +1,7 @@
+package main
+
+import "github.com/kubescape/go-logger"
+
+func main() {
+	logger.L().Error("CLI not yet implemented")
```
vladklokun, line 6: issue: `Error()` will exit with return code 0.

But if the CLI is not implemented, I think it constitutes an non-successful execution of the application, which, in my opinion, should return a non-successful exit code.

How about:
```suggestion
	logger.L().Fatal("CLI not yet implemented")
```

cmd/http/main.go
```diff
@@ -0,0 +1,45 @@
...
+			url.URL{Host: otelHost})
+		defer logger.ShutdownOtel(ctx)
+	}
+
+	repository := repositories.NewMemoryStorage() // TODO add real storage
+	sbomAdapter := v1.NewSyftAdapter()
+	cveAdapter, _ := v1.NewGrypeAdapter(ctx)
+	platform := adapters.NewMockPlatform() // TODO add real platform
+	service := services.NewScanService(sbomAdapter, repository, cveAdapter, repository, platform)
+	controller := controllers.NewHTTPController(service)
```
vladklokun, line 35: praise: I have to tell you, I really like how injectable all this is!

docs/swagger.yaml
```diff
@@ -0,0 +1,216 @@
+basePath: /
```
vladklokun, line 1: question: is this an autogenerated Swagger spec?

cmd/http/main_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+func TestScan(t *testing.T) {
+	tests := []struct {
+		name         string
+		yamlFile     string
+		expectedCode int
+		expectedBody string
+	}{
+		{
+			"good",
```
vladklokun, line 25: chore: let’s name the test cases more explicitly

cmd/http/main_test.go
```diff
@@ -0,0 +1,67 @@
...
+			"../../api/v1/testdata/scan.yaml",
+			200,
+			"{\"data\":\"new CVEs created\",\"status\":\"success\"}",
+		},
+		{
+			"missing fields",
```
vladklokun, line 31: chore: let’s name the test cases more explicitly

cmd/http/main_test.go
```diff
@@ -0,0 +1,67 @@
...
+			service := services.NewScanService(sbomAdapter, repository, cveAdapter, repository, platform)
+			controller := controllers.NewHTTPController(service)
+
+			router := gin.Default()
+
+			router.POST("/v1/generateSBOM", controller.GenerateSBOM)
+			router.POST("/v1/scanImage", controller.ScanCVE)
+
+			file, err := os.Open(test.yamlFile)
+			assert.NoError(t, err)
```
vladklokun, line 52: issue: this function is supposed to assert that scans are handled properly, not that the test YAML file opens without errors

Let’s make assertions only on things we are supposed to verify.

cmd/http/main_test.go
```diff
@@ -0,0 +1,67 @@
...
+			req, _ := http.NewRequest("POST", "/v1/generateSBOM", file)
+			w := httptest.NewRecorder()
+			router.ServeHTTP(w, req)
+
+			file, err = os.Open(test.yamlFile)
+			assert.NoError(t, err)
```
vladklokun, line 58: issue: same non-tested assertion as above

controllers/http.go
```diff
@@ -0,0 +1,85 @@
...
+	err = h.scanService.GenerateSBOM(c.Request.Context(), newScan.ImageHash, domain.Workload(newScan))
+	if err != nil {
+		logger.L().Ctx(c.Request.Context()).Error("service error", helpers.Error(err))
+		c.JSON(http.StatusInternalServerError, nil)
+		return
+	}
+
+	c.JSON(http.StatusOK, gin.H{
+		"status": "success",
+		"data":   "new SBOMs created",
```
vladklokun, line 44: issue: the response does not give its clients a way to see what they have created

Usually, creation APIs return the ID of the created items, their bodies or some kind of a reference to them.

Is there a reason we don’t do something like that? Maybe we should at least give back the ID of the image for which we are creating the SBOM.

I believe this is a new API endpoint that did not exist previously, so we can do it properly.

As a consumer of this API I would like to see a reference to the SBOM that has been created. I know we are currently storing them in memory, so an ID of the SBOM would do. And, actually, it would help even when we store it somewhere else ;)

controllers/http.go
```diff
@@ -0,0 +1,85 @@
...
+	})
+}
+
+func (h HTTPController) Ready(c *gin.Context) {
+	c.Header("Content-Type", "application/json")
+
+	if !h.scanService.Ready() {
+		c.JSON(http.StatusServiceUnavailable, gin.H{
+			"status": "Scanner not ready",
```
vladklokun, line 53: question: is this message for backwards compatibility or can we change it to whatever API we want?

Designing APIs where statuses are messages like this is not really a best practice. If we are breaking things, we should do it properly, like using proper string enums for statuses and being descriptive with our errors.

controllers/http.go
```diff
@@ -0,0 +1,85 @@
...
+	err = h.scanService.ScanCVE(c.Request.Context(), newScan.Wlid, newScan.ImageHash, domain.Workload(newScan))
+	if err != nil {
+		logger.L().Ctx(c.Request.Context()).Error("service error", helpers.Error(err))
+		c.JSON(http.StatusInternalServerError, nil)
+		return
+	}
+
+	c.JSON(http.StatusOK, gin.H{
+		"status": "success",
+		"data":   "new CVEs created",
```
vladklokun, line 83: issue: same as the SBOM, we should return a reference

controllers/http.go
```diff
@@ -0,0 +1,85 @@
...
+
+	err = h.scanService.GenerateSBOM(c.Request.Context(), newScan.ImageHash, domain.Workload(newScan))
+	if err != nil {
+		logger.L().Ctx(c.Request.Context()).Error("service error", helpers.Error(err))
+		c.JSON(http.StatusInternalServerError, nil)
+		return
+	}
+
+	c.JSON(http.StatusOK, gin.H{
+		"status": "success",
```
vladklokun, line 43: question: is this the only status code we will support? Or it is just a placeholder for now?

I would expect the status codes to be defined in some kind of an Enum at least.

There are also alternative, arguably more ergonomic ways to report errors over a JSON API. For example, a lot of APIs feature a response that looks similar to the following object:
```
{
  "ok": false,
  "data": { ... },
  "info": {
    "status": "error",
    "code": "invalidCredentials",
    "message": "The provided credentials were not valid for the given image ID.",
    "suggestion": "Check the provided credentials and try again."
  }
```

You could check the [Stripe HTTP API](https://stripe.com/docs/api/errors) for inspiration. It’s really good.

core/domain/cve.go
```diff
@@ -0,0 +1,9 @@
+package domain
+
+type CVE struct {
+	ImageID            string
+	SBOMCreatorVersion string
+	CVEScannerVersion  string
+	CVEDBVersion       string
+	Content            []byte
+}
```
vladklokun, line 9: question: what is CVE?

Is it a list of vulnerabilities? If so, then, as Ben said, CVE is a pretty specific name. Moreover it usually refers to one vulnerability, not a collection of them, so if it’s a list of vulnerabilities, we should name it appropriately.

core/domain/sbom.go
```diff
@@ -0,0 +1,21 @@
+package domain
+
+type SBOM struct {
```
vladklokun, line 3: chore: we should also document our exported structures

core/domain/cve.go
```diff
@@ -0,0 +1,9 @@
+package domain
+
+type CVE struct {
+	ImageID            string
+	SBOMCreatorVersion string
+	CVEScannerVersion  string
+	CVEDBVersion       string
+	Content            []byte
```
vladklokun, line 8: question: what does this CVE object contain?

Is it a list of individual vulnerabilities?

core/domain/workload.go
```diff
@@ -0,0 +1,5 @@
+package domain
+
+import wssc "github.com/armosec/armoapi-go/apis"
+
+type Workload wssc.WebsocketScanCommand
```
vladklokun, line 5: question: why `Workload`?

I remember this object being exactly what it is called: a scan command. Personally, I would drop the websocket part, but otherwise it looked to me like a scan command or a scan request.

core/services/scan.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/kubescape/kubevuln/core/ports"
+	"go.opentelemetry.io/otel"
+)
+
+type ScanService struct {
+	sbomCreator    ports.SBOMCreator
+	sbomRepository ports.SBOMRepository
+	cveScanner     ports.CVEScanner
+	cveRepository  ports.CVERepository
+	armoPlatform   ports.Platform
```
vladklokun, line 19: issue: CNCF requires us to be vendor-neutral, so we should be agnostic of specific backends

We even had to remove all mentions of ARMO in Kubevuln, so adding them back is a no-no :)

core/services/scan.go
```diff
@@ -0,0 +1,105 @@
...
+		armoPlatform:   armoPlatform,
+	}
+}
+
+func (s *ScanService) GenerateSBOM(ctx context.Context, imageID string, workload domain.Workload) error {
+	ctx, span := otel.Tracer("").Start(ctx, "GenerateSBOM")
+	defer span.End()
+	if imageID == "" {
+		return errors.New("missing imageID")
+	}
```
vladklokun, line 39: nitpick: same thing with separating logical code blocks with newlines

core/services/scan.go
```diff
@@ -0,0 +1,105 @@
...
+
+func (s *ScanService) Ready() bool {
+	return s.cveScanner.Ready()
+}
+
+func (s *ScanService) ScanCVE(ctx context.Context, instanceID string, imageID string, workload domain.Workload) error {
+	ctx, span := otel.Tracer("").Start(ctx, "ScanCVE")
+	defer span.End()
+	if instanceID == "" {
+		return errors.New("missing instanceID")
```
vladklokun, line 64: question: how will users of a service be able to verify that an error they receive is about a missing instance ID and handle it appropriately?

If we create errors with `errors.New()`, clients of the function will have to compare the string contents of the error. This can get tedious very quick. If I’m correct, the current best practices are about defining custom errors and comparing against them with `errors.Is()`, as documented in the [Go blog](https://go.dev/blog/go1.13-errors) and [this](https://blog.logrocket.com/error-handling-golang-best-practices/) Logrocket blog.

vladklokun (changes requested):
Regarding the overall architecture, I really like it. It truly seems composable and supportable, at least at this stage. You did an amazing job!

I left comments here and there, but most of these are really small. Except, maybe, making the error handling consistent.

Again, great job and thank you!

adapters/v1/grype_test.go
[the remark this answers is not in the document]
matthyx: This is an interesting question... the fact is that we need a properly initialized `GrypeAdapter` for the test to run.
Do you mean writing a test for `NewGrypeAdapter` - yes I should do that.
Do you also mean removing `NewGrypeAdapter` from this test - I don't think I can, otherwise I have to the init by hand?
Or maybe you mean to move `NewGrypeAdapter` out of `t.Run()`?

adapters/v1/grype_test.go
```diff
@@ -0,0 +1,56 @@
...
+		})
+	}
+}
+
+func Test_grypeAdapter_Version(t *testing.T) {
+	g, err := NewGrypeAdapter(context.TODO())
+	assert.Assert(t, err == nil)
```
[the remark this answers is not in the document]
matthyx: Testing the DB version requires a properly initialized DB, which is done inside `NewGrypeAdapter`.

adapters/v1/grype.go
```diff
@@ -0,0 +1,147 @@
...
+	store    *store.Store
+}
+
+var _ ports.CVEScanner = (*GrypeAdapter)(nil)
+
+func NewGrypeAdapter(ctx context.Context) (*GrypeAdapter, error) {
+	dbConfig := db.Config{
+		DBRootDir:  "grypedb",
+		ListingURL: "https://toolbox-data.anchore.io/grype/databases/listing.json",
+	}
```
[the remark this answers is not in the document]
matthyx: +1 this will be required when we move to postgres for the clusterization, I have to check if that's doable

adapters/v1/grype.go
```diff
@@ -0,0 +1,147 @@
...
+			Python:     python.MatcherConfig{UseCPEs: true},
+			Dotnet:     dotnet.MatcherConfig{UseCPEs: true},
+			Javascript: javascript.MatcherConfig{UseCPEs: true},
+			Golang:     golang.MatcherConfig{UseCPEs: true},
+			Stock:      stock.MatcherConfig{UseCPEs: true},
+		},
+	)
+}
+
+func (g *GrypeAdapter) ScanSBOM(ctx context.Context, sbom domain.SBOM) (domain.CVE, error) {
```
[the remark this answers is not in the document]
matthyx: +1

adapters/v1/syft.go
matthyx: This is a requirement from @dwertent and @rcohencyberarmor because the image is downloaded during scan, they want to avoid disk usage issues.
The mutex is used (for the moment) because it's simpler regarding error reporting to the Operator than a pure event-driven solution (although they might want to go there).

adapters/v1/syft_test.go
matthyx: Yes, for the moment it only initializes the struct, which cannot fail... maybe for future proofing I should add error handling in the case this "constructor" needs it in the future?
Let's KISS for the moment.

docs/swagger.yaml
matthyx: Unfortunately no, and it should not be there... this is some experiment with LightHouse and unrelated to this PR.

cmd/http/main_test.go
matthyx: Do you suggest to not fail the test if this file cannot open? I know this is not the purpose of the test, but not handling the error of `os.Open` seems dangerous?

controllers/http.go
matthyx: This is part of the discussion I'm supposed to have with @Daniel-GrunbergerCA when defining APIs.
I would like to use this: https://github.com/mschneider82/problem

core/domain/workload.go
matthyx: At the beginning I wanted to scan a workload, and since for the moment we wanted to reuse the existing object from the Operator point of view, I have decided to abstract it to my initial thought.
I should probably rename it to `ScanCommand` and at some point use a generic (uncorrelated) representation.

core/services/scan.go
matthyx: `saasPlatform` then?

core/domain/cve.go
matthyx: this is the CVE manifest in JSON format straight from Grype

adapters/v1/grype_test.go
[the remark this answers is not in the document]
vladklokun: >  Do you also mean removing NewGrypeAdapter from this test - I don't think I can, otherwise I have to the init by hand?
No, you don’t have to. My point is that:

```go
g, err := NewGrypeAdapter(context.TODO())
assert.Assert(t, err == nil) // this line should be removed, since this does not concern the `SaveSBOM()` method
```

adapters/v1/grype.go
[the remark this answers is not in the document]
matthyx: I suggest we move that to a separate PR

controllers/http.go
matthyx: this was mostly a placeholder, but I am also worried about giving too much information with error messages, as it could help attackers

core/services/scan.go
matthyx: good point, let's do it later

adapters/v1/grype.go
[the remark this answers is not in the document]
vladklokun: Sure.

adapters/v1/syft.go
vladklokun: If that’s the path you’ve chosen, I don’t mind.

However, judging by the function docstring, if our goal here is to limit possible disk usage, a better solution might be using a semaphore instead of a mutex, since it will allow us to adjust the concurrency level as necessary.

vladklokun (approved):
LGTM!

adapters/v1/syft.go
matthyx: good idea :)

PR #35, opened 2023-02-14
use queue to control syft concurrency

matthyx:
Using https://github.com/gammazero/workerpool as a worker pool.
I have added a graceful shutdown mechanism to "theoretically" empty the worker queue before exit.

### DIFF
--- a/cmd/http/main.go
+++ b/cmd/http/main.go
@@ -2,8 +2,12 @@ package main
 
 import (
 	"context"
+	"net/http"
 	"net/url"
 	"os"
+	"os/signal"
+	"syscall"
+	"time"
 
 	"github.com/gin-gonic/gin"
 	"github.com/kubescape/go-logger"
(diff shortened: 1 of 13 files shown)


cmd/http/main.go
```diff
@@ -41,5 +47,36 @@ func main() {
 	router.POST("/v1/generateSBOM", controller.GenerateSBOM)
 	router.POST("/v1/scanImage", controller.ScanCVE)
 
-	logger.L().Ctx(ctx).Fatal("router error", helpers.Error(router.Run()))
+	srv := &http.Server{
+		Addr:    ":8080",
```
Daniel-GrunbergerCA, line 51: I think it's better to make the port configurable

matthyx: Indeed, this is part of the industrialization. Before it was defaulted to 8080, I had to make it explicit when declaring the server.
Please focus on the graceful shutdown sequence that is required by the queue you asked for :)

Daniel-GrunbergerCA (approved):

PR #37, opened 2023-02-21
add backend status reporting

matthyx:


Daniel-GrunbergerCA:
Small comment: Maybe we want to store "Inqueueing" and "Dequeueing" in consts, so you won't have to type every time and to avoid a typo

Daniel-GrunbergerCA (approved):

matthyx:
> Small comment: Maybe we want to store "Inqueueing" and "Dequeueing" in consts, so you won't have to type every time and to avoid a typo

good point, although I would like to check with Amir if we really use these terms

adapters/v1/armo.go
```diff
@@ -0,0 +1,74 @@
...
+const ActionName = "vuln scan"
+const ReporterName = "ca-vuln-scan"
+
+var details = []string{
+	sysreport.JobStarted,
+	sysreport.JobStarted,
+	sysreport.JobSuccess,
+	sysreport.JobDone,
+}
+var statuses = []string{
```
matthyx, line 38: @Daniel-GrunbergerCA I did an indirection to abstract from implementation

Daniel-GrunbergerCA: I see. I think you should check for that the index exists in the list in the `SendStatus` function just in case, so we don't panic. Also I think is better if you move the lists to the steps.go file, since it will be easier to understand how these lists are used and what is the purpose of the steps consts.

matthyx: no, the list is implementation specific, it should not spill out

core/domain/platform.go
```diff
@@ -0,0 +1,8 @@
+package domain
+
+const (
+	Accepted = iota
+	Started
+	Success
+	Done
+)
```
matthyx, line 8: @Daniel-GrunbergerCA is it better like that?

PR #44, opened 2023-03-03
add docker build

matthyx:


build/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM golang:1.19-alpine as builder
+
+ENV GO111MODULE=on
+ENV CGO_ENABLED=0
```
vladklokun, line 4: question (non-blocking): is there a reason we keep these ENV variables on separate lines and layers?

Maybe we keep them on one line and save ourselves a layer:
```suggestion
ENV GO111MODULE=on CGO_ENABLED=0
```

build/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+WORKDIR /work
+
+ADD go.mod go.sum /work/
+RUN go mod download
```
vladklokun, line 8: question (non-blocking): am I correct in understanding that you download modules separately to better leverage build caching?

So that the layers stay cached as long as dependencies stay the same, and cache is still valid. If so, then I’m stealing this trick :grin:

build/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+ADD . .
+RUN go build -o build/kubevuln cmd/http/main.go
+
+FROM alpine
+
+RUN addgroup -S ks && adduser -S ks -G ks
+USER ks
+WORKDIR /home/ks/
```
vladklokun, line 17: suggestion (non-blocking): since `ks` gets repeated multiple times, should we extract it into an `ARG`?

Could help re-use the value in the future.

vladklokun (approved):
LGTM! All comments are non-blocking, but feel free to address them. Good job!

build/Dockerfile
matthyx: it's not relevant for the final image, since we have a multistage build
but it might save 1s during the build process

build/Dockerfile
matthyx: this is copied from operator... if it was me I would run a randomUID and use `scratch` as the base image

build/Dockerfile
matthyx: yes, but build still takes ages given all deps we need to recompile each time
there's even better solutions: https://github.com/edwarnicke/imports-gen

build/Dockerfile
vladklokun: Ah, you’re right, since it’s multi-stage. Thanks for clarifying!

build/Dockerfile
vladklokun: Thanks!

PR #48, opened 2023-03-06
add clusterName to otel initialization

matthyx:


amirmalka:
Great.

One small change - as I did with the other components, let's remove the otel middleware/logging from health check methods such as `/ready` etc. to reduce noise...

amirmalka (approved):

amirmalka (approved):

PR #59, opened 2023-03-20
accept empty instanceID

matthyx:


repositories/apiserver.go
```diff
@@ -116,13 +125,20 @@ func (a *APIServerStore) StoreCVE(ctx context.Context, cve domain.CVEManifest, w
...
 }
 
 func (a *APIServerStore) GetSBOM(ctx context.Context, imageID, SBOMCreatorVersion string) (sbom domain.SBOM, err error) {
+	if imageID == "" {
+		return domain.SBOM{}, nil
+	}
 	manifest, err := a.StorageClient.SBOMSPDXv2p3s(a.Namespace).Get(ctx, hashFromImageID(imageID), metav1.GetOptions{})
 	if err != nil {
-		return domain.SBOM{}, err
+		logger.L().Ctx(ctx).Warning("failed to get SBOM from apiserver", helpers.Error(err), helpers.String("ID", imageID))
```
dwertent, line 140: What if the storage is not running? Will we still see this warning message?

repositories/apiserver.go
```diff
@@ -116,13 +125,20 @@ func (a *APIServerStore) StoreCVE(ctx context.Context, cve domain.CVEManifest, w
 		},
 	}
 	_, err := a.StorageClient.VulnerabilityManifests(a.Namespace).Create(ctx, &manifest, metav1.CreateOptions{})
-	return err
+	if err != nil {
+		logger.L().Ctx(ctx).Warning("failed to store CVE manifest into apiserver", helpers.Error(err), helpers.String("ID", cve.ImageID))
```
dwertent, line 129: What if the storage is not running? Will we still see this warning message?

repositories/apiserver.go
```diff
@@ -140,9 +156,13 @@ func (a *APIServerStore) GetSBOM(ctx context.Context, imageID, SBOMCreatorVersio
 }
 
 func (a *APIServerStore) GetSBOMp(ctx context.Context, instanceID, SBOMCreatorVersion string) (sbom domain.SBOM, err error) {
+	if instanceID == "" {
+		return domain.SBOM{}, nil
+	}
 	manifest, err := a.StorageClient.SBOMSPDXv2p3Filtereds(a.Namespace).Get(ctx, hashFromInstanceID(instanceID), metav1.GetOptions{})
 	if err != nil {
-		return domain.SBOM{}, err
+		logger.L().Ctx(ctx).Warning("failed to get relevant SBOM from apiserver", helpers.Error(err), helpers.String("ID", instanceID))
```
dwertent, line 164: What if the storage is not running? Will we still see this warning message?

dwertent: I think this will happen a lot because the relevant SBOM is ready only after X  minutes.
It should be a debug log or no log at all

repositories/apiserver.go
```diff
@@ -184,5 +207,8 @@ func (a *APIServerStore) StoreSBOM(ctx context.Context, sbom domain.SBOM) error
 		manifest.Spec.Metadata.Report.CreatedAt.Time = created
 	}
 	_, err = a.StorageClient.SBOMSPDXv2p3s(a.Namespace).Create(ctx, &manifest, metav1.CreateOptions{})
-	return err
+	if err != nil {
+		logger.L().Ctx(ctx).Warning("failed to store SBOM into apiserver", helpers.Error(err), helpers.String("ID", sbom.ID))
```
dwertent, line 211: Again, if it is not running, we should not print warnings...

dwertent (changes requested):

matthyx:
This PR covers cases when storage is enabled. I didn't fix the other issue where we try to access the storage when it's disabled...
This will be in second PR, where I remove generateSBOM.

dwertent (approved):

PR #60, opened 2023-03-21
implement single call scan for CVE

matthyx:


matthyx:
ongoing...

matthyx:
@dwertent ready

PR #63, opened 2023-03-22
move send chunks logs to debug

matthyx:


adapters/v1/armo_utils.go
```diff
@@ -94,7 +94,7 @@ func (a *ArmoAdapter) postResults(ctx context.Context, report *v1.ScanResultRepo
 		errorChan <- err
 		return
 	}
-	logger.L().Info(fmt.Sprintf("posting to event receiver image %s wlid %s finished successfully response body: %s", imagetag, wlid, body)) // systest dependent
+	logger.L().Debug(fmt.Sprintf("posting to event receiver image %s wlid %s finished successfully response body: %s", imagetag, wlid, body)) // systest dependent
```
matthyx, line 97: @Daniel-GrunbergerCA can you check if this is really needed for systests?

dwertent: Unfortunately it does, but the tests run in debug mode, so you should be fine

dwertent (approved):

PR #69, opened 2023-03-26
Update instanceID generating and handling 

dwertent:
### **This code was not tested!**

Generating instanceID as defined [here](https://github.com/kubescape/k8s-interface/blob/update-instanceid/instanceidhandler/interface.go)


go.mod
```diff
@@ -21,6 +22,7 @@ require (
 	github.com/hashicorp/go-multierror v1.1.1
 	github.com/kinbiko/jsonassert v1.1.1
 	github.com/kubescape/go-logger v0.0.11
+	github.com/kubescape/k8s-interface v0.0.114-0.20230326195717-d11a8f35efc2
```
matthyx, line 25: again, let's wait for release

PR #79, opened 2023-03-28
add syft version to SBOM metadata

matthyx:


adapters/v1/testdata/alpine-sbom.format.json
```diff
@@ -11,7 +11,7 @@
   "licenseListVersion": "<<PRESENCE>>",
   "creators": [
    "Organization: Anchore, Inc",
-   "Tool: syft-"
+   "Tool: syft-unknown"
```
matthyx, line 14: here it's unknown because package introspection only works on a compiled binary

PR #87, opened 2023-03-29
Fix vuln counter

Daniel-GrunbergerCA:
setting `relevantLabel` was being done on a copy of object, not object itself


matthyx (approved):
nice catch

PR #88, opened 2023-03-29
ensure proper labels and annotations on objects

matthyx:


internal/tools/tools.go
```diff
@@ -22,3 +26,25 @@ func PackageVersion(name string) string {
...
+func sanitize(s string) string {
+	return truncate.Truncate(offendingChars.ReplaceAllString(s, "-"), 63, "", truncate.PositionEnd)
+}
+
+// LabelsFromImageID returns a map of labels from an image ID.
+// Each label is sanitized and verified to be a valid DNS1123 label.
+func LabelsFromImageID(imageID string) map[string]string {
+	labels := map[string]string{}
+	match := reference.ReferenceRegexp.FindStringSubmatch(imageID)
+	labels["name"] = sanitize(match[1])
```
dwertent, line 41: Let's use a global variable that will look like:
```
kubescape.io/image-name
```

dwertent (changes requested):

PR #97, opened 2023-04-06
add support for big image scanning

matthyx:


dwertent (approved):

config/config.go
```diff
@@ -11,6 +11,7 @@ type Config struct {
 	BackendOpenAPI       string        `mapstructure:"backendOpenAPI"`
 	ClusterName          string        `mapstructure:"clusterName"`
 	EventReceiverRestURL string        `mapstructure:"eventReceiverRestURL"`
+	MaxImageSize         int64         `mapstructure:"maxImageSize"`
```
dwertent, line 14: @matthyx This config object is global for all of the in-cluster components.
The `maxImageSize` is specific to the kubevuln, so I don't this it should be part of this structure.
I suggest we should have a dedicated config file for such configurations.

matthyx: Agreed, can you create a Jira task for it? I'll check when I can split these configs, as it's not the only config that should remain specific to kubevuln.

PR #98, opened 2023-04-17
discard relevant SBOM based on validating func

matthyx:
also don't discard if sbom tool version is different

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -2,6 +2,7 @@ package repositories
 
 import (
 	"context"
+	"fmt"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


repositories/apiserver.go
```diff
@@ -190,6 +191,13 @@ func (a *APIServerStore) GetSBOM(ctx context.Context, imageID, SBOMCreatorVersio
 	return result, nil
 }
 
+func validateSBOMp(manifest *v1beta1.SBOMSPDXv2p3Filtered) error {
```
matthyx, line 194: @dwertent this is where you should make your changes

dwertent (approved):

PR #109, opened 2023-05-16
don't use imagetag + imagehash for scanid

matthyx:


Daniel-GrunbergerCA (approved):

core/services/scan.go
```diff
@@ -279,9 +279,6 @@ func generateScanID(workload domain.ScanCommand) string {
 	if workload.InstanceID != "" {
 		return workload.InstanceID
 	}
-	if workload.ImageTag != "" && workload.ImageHash != "" {
```
matthyx, line 282: either this, or we hash the string... @dwertent

dwertent: I think we should hash the string

dwertent: @matthyx

core/services/scan.go
```diff
@@ -280,7 +282,8 @@ func generateScanID(workload domain.ScanCommand) string {
 		return workload.InstanceID
 	}
 	if workload.ImageTag != "" && workload.ImageHash != "" {
-		return workload.ImageTag + workload.ImageHash
+		sum := sha256.Sum256([]byte(workload.ImageTag + workload.ImageHash))
```
matthyx, line 285: @dwertent fixed

Bezbran: Please use FNV. it's much cheaper and we have no need in cryptography grade hash function in this scenario

dwertent (approved):

PR #113, opened 2023-05-25
do not pull image for 10m after error 421

matthyx:


dwertent:
What will be the behavior in such a case? When will Kubevuln resume pulling the images?

matthyx:
Next time it is called by operator, but only if the timeout expired

dwertent (approved):

Issue #111, opened 2023-05-23
Error running kubevuln


craigbox:
Cluster has relevancy beta installed and I assume the scan was either triggered by me restarting a deployment to test to see if anything was generated, or by me hitting "Scan" in the ARMO UI when it seemed like nothing happened.

```
2023-05-23 21:12:50 {"level":"info","ts":"2023-05-23T20:12:50Z","msg":"starting server"}
2023-05-23 21:12:51 {"level":"info","ts":"2023-05-23T20:12:51Z","msg":"updating grype DB"}
2023-05-23 21:12:57 {"level":"info","ts":"2023-05-23T20:12:57Z","msg":"grype DB updated"}
2023-05-23 21:31:22 {"level":"info","ts":"2023-05-23T20:31:22Z","msg":"scan started","imageID":"docker://sha256:2edf9c994f199aecfea940f65e2582eea072a6c2e2a747db5af3933a77a8ce46","jobID":"7f7cd5fb-2527-46e0-b237-b42cc7fd7430"}
2023-05-23 21:31:22 panic: runtime error: index out of range [3] with length 0 [recovered]
2023-05-23 21:31:22     panic: runtime error: index out of range [3] with length 0 [recovered]
2023-05-23 21:31:22     panic: runtime error: index out of range [3] with length 0
2023-05-23 21:31:22 
2023-05-23 21:31:22 goroutine 1058 [running]:
2023-05-23 21:31:22 go.opentelemetry.io/otel/sdk/trace.(*recordingSpan).End.func1()
2023-05-23 21:31:22     /go/pkg/mod/go.opentelemetry.io/otel/sdk@v1.11.2/trace/span.go:383 +0x2a
2023-05-23 21:31:22 go.opentelemetry.io/otel/sdk/trace.(*recordingSpan).End(0xc0005ea000, {0x0, 0x0, 0x23100f0bb927f753?})
2023-05-23 21:31:22     /go/pkg/mod/go.opentelemetry.io/otel/sdk@v1.11.2/trace/span.go:421 +0x942
2023-05-23 21:31:22 panic({0x2424280, 0xc0005e2078})
2023-05-23 21:31:22     /usr/local/go/src/runtime/panic.go:884 +0x212
2023-05-23 21:31:22 go.opentelemetry.io/otel/sdk/trace.(*recordingSpan).End.func1()
[...]

PR #114, opened 2023-05-28
fix(panic): Enhanced Support for Diverse Image ID Formats in Hash Extraction

dwertent:
## Pull Request Description

Title: **Handling of Various Image ID Formats in Hash Extraction**
Resolves: #111 

### Background:
The current implementation of hash extraction from Image IDs assumes a certain pattern and does not account for all variations of Image IDs, specifically formats like:
- "docker://sha256:c1b135231b5b1a6799346cd701da4b59e5b7ef8e694ec7b04fb23b8dbe144137"
- "sha256:c1b135231b5b1a6799346cd701da4b59e5b7ef8e694ec7b04fb23b8dbe144137"

There are instances where the Image ID doesn't have enough submatches (less than 3) leading to panic as the existing code tries to access an index that does not exist.

### Changes:
This PR provides a robust mechanism to extract the SHA256 hash from Image IDs that can handle varying formats. The key changes include:

1. **Adding a check for plain SHA256 hashes:** If the Image ID is a plain SHA256 hash, the function will now directly return it as the hash.

2. **Improved Error Handling:** An error is now returned when there is an issue with the hash extraction, instead of the function panicking. 

3. **Fallback Mechanism:** In cases where the Image ID doesn't match the usual pattern, a fallback mechanism has been implemented to parse and extract the hash using a regular expression.

### Impact:
This change will make the hash extraction from Image IDs more robust and prevent unexpected panics. It also improves error handling and provides more detailed error messages to help with troubleshooting.

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -3,6 +3,7 @@ package repositories
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
@@ -35,6 +36,10 @@ var _ ports.CVERepository = (*APIServerStore)(nil)
 
 var _ ports.SBOMRepository = (*APIServerStore)(nil)
 
+var re = regexp.MustCompile(`([a-f0-9]{64})`)
+
+const ImageHashLength = 64
+
 // NewAPIServerStorage initializes the APIServerStore struct
 func NewAPIServerStorage(namespace string) (*APIServerStore, error) {
 	config, err := rest.InClusterConfig()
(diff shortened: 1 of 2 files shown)


matthyx:
Thanks for fixing it... It's weird that the official method doesn't account for these forms.
Anyway this should move to the common methods in order to use it for kubevuln, operator and backend.

dwertent:
@matthyx Where is the common method?

matthyx:
Doesn't exist yet... It would make sense to have one

dwertent:
@matthyx Should we merge this or should we write such a function? Do we know if the Operator needs such a capability?

matthyx:
let's merge it now, worst case we can extract it later...
thanks for the fix

matthyx (approved):

PR #120, opened 2023-06-02
rename crd

matthyx:


controllers/http.go
```diff
@@ -123,6 +137,9 @@ func websocketScanCommandToScanCommand(c wssc.WebsocketScanCommand) domain.ScanC
 		Args:            c.Args,
 		Session:         sessionChainToSession(c.Session),
 	}
+	if slug, err := names.ImageInfoToSlug(c.ImageTag, c.ImageHash); err == nil {
+		command.ImageID = slug
```
vladklokun, line 141: suggestion(non-blocking): let’s update the name of the field to be more descriptive

As far as I understand, `command.ImageID` contains not the image ID, but the image ID slug. Right?

If that’s correct, then let’s rename the field to something like `ImageSlug`. Seeing how this was used for both SBOMs and CVEs I personally found it confusing.

Another way to tackle this would be defining a special type, but not in your code and in `k8s-interface/names` instead:
```
type ImageSlug string
```
That way we can be explicit about the type of string we expect to be there.

Let me know what you thing and prefer.

controllers/http.go
```diff
@@ -176,6 +199,9 @@ func registryScanCommandToScanCommand(c wssc.RegistryScanCommand) domain.ScanCom
 		Args:            c.Args,
 		Session:         sessionChainToSession(c.Session),
 	}
+	if slug, err := names.ImageInfoToSlug(c.ImageTag, "nohash"); err == nil {
+		command.ImageID = slug
+	}
```
vladklokun, line 204: question: is it okay to pass this `nohash` value down to the SBOM creator?

Doesn’t this pollute the data? Will Syft be okay scanning images with the `nohash` tag?

vladklokun (commented):
See questions. If they concern nothing serious, LGTM :)

controllers/http.go
matthyx: it's fine

vladklokun (approved):
Thank you, LGTM!

PR #123, opened 2023-06-07
retrieve labels and annotations from storage

matthyx:
fixes system tests and probably other bugs


matthyx:
I made sure we cover that too

PR #127, opened 2023-06-15
Bump go version to 20

dwertent:
* Bumping go version to 20
* Adding e2e tests
* Adding github templates

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/kubescape/kubevuln
 
-go 1.19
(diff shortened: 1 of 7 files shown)


go.mod
rcohencyberarmor, line 1: are you sure that this change will run with the new workflow file, and not the old one?

rcohencyberarmor (changes requested):

go.mod
dwertent: yes

PR #130, opened 2023-06-22
fix summary chunk with IsLastReport=true

matthyx:
one of the few original pieces of code:
https://github.com/kubescape/kubevuln/blob/master/scanner/request.go#L202-L203
(code does the opposite of comment)


dwertent:
How did the tests pass until now?

matthyx:
I don't think it has any impact on the backend processing... But when I saw this I had to fix it.

PR #132, opened 2023-06-27
add imp for store summary CRD objects

rcohencyberarmor:


internal/tools/tools.go
```diff
@@ -55,3 +59,27 @@ func LabelsFromImageID(imageID string) map[string]string {
 	}
 	return labels
 }
+
+func FileContent(path string) []byte {
```
matthyx, line 63: good, in the future we can clean other tests with these

matthyx (approved):
thx!

PR #135, opened 2023-07-25
env var for config file added

### LINKED ISSUE #129: [question] can't load config file using `CONFIG` env
# Description

[As mentioned in the README](https://github.com/kubescape/kubevuln#configuration), we can load config file using the `CONFIG` env.
However, we are unable to load config file path using the `CONFIG` environment variable.
Kubevuln always defaults to searching the config file in the `/etc/config` path, even when we define a new path in the `CONFIG` env.

In order to run kubevuln:
- We either add the config file named `clusterData.json` in the path `/etc/config`, the default path where kubevuln searches for it.
- Or, we explicitly define a new path in the code [here](https://github.com/kubescape/kubevuln/blob/main/cmd/http/main.go#L29), where we want Kubevuln to search for the config file named `clusterData.json`, build kubevuln and then run it

If we follow any of the above 2 steps, then I am able to run it locally

# Environment
OS:  `Ubuntu 20.04.4 LTS` 
 
# Steps To Reproduce
Method 1:
1. Build kubevuln using `make`
2. Load config file using the CONFIG environment variable
`export CONFIG=path/to/clusterData.json`

Method 2:
1. Build the Docker image from the Dockerfile and run it using:
`docker run <kubevuln-built-image-name> -e CONFIG=path/to/clusterData.json`

# Additional Context
Is this an expected behavior or am I missing something?

siddhikhapare:
## Overview
Use the CONFIG environment variable to load the config file, but it currently doesn't allow custom file paths. Kubevuln always defaults to searching for the config file in the /etc/config path, regardless of the CONFIG env value.

## How to Test

Build kubevuln using ```make```
Load config file using the CONFIG environment variable
```export CONFIG=path/to/clusterData.json```

Set the PORT environment variable to 8081
```export PORT=8080```

## Related issues/PRs:
Resolved #129 

### DIFF
--- a/cmd/http/main.go
+++ b/cmd/http/main.go
@@ -26,7 +26,12 @@ import (
 func main() {
 	ctx := context.Background()
 
-	c, err := config.LoadConfig("/etc/config")
+	configPath := os.Getenv("CONFIG")
+	if configPath == "" {
+		configPath = "/etc/config"
+	}
+
+	c, err := config.LoadConfig(configPath)
 	if err != nil {
 		logger.L().Ctx(ctx).Fatal("load config error", helpers.Error(err))
 	}

### CODE REVIEW

File: cmd/http/main.go [hunk 1]
```diff
@@ -26,7 +26,12 @@ import (
 func main() {
 	ctx := context.Background()
 
-	c, err := config.LoadConfig("/etc/config")
+	configPath := os.Getenv("CONFIG")
```
dwertent, line 29: @matthyx what do you think is the proper solution here? Because we are thinking of having multiple configuration files...

matthyx: Exactly, the goal is to split the config between component and cluster like in [node-agent](https://github.com/kubescape/node-agent/blob/main/pkg/config/config.go).

siddhikhapare:
@dwertent  @matthyx I understand that config.LoadConfig function to load configurations from multiple JSON files so we can also load configurations from multiple JSON files in the "/etc/config" directory using the filepath.Glob approach  but viper.AddConfigPath(path) is look for config in the mentioned "/etc/config" directory  so What should I do then? because I think your approach is right then to handle loading configurations from multiple directories specified through environment variable is not correct which I have tried.   Thank you.

cmd/http/main.go
```diff
@@ -26,7 +26,12 @@ import (
...
+	if configPath == "" {
+		configPath = "/etc/config"
+	}
+
+	c, err := config.LoadConfig(configPath)
 	if err != nil {
 		logger.L().Ctx(ctx).Fatal("load config error", helpers.Error(err))
 	}
```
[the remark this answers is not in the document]
siddhikhapare: okay

PR #139, opened 2023-08-16
Store summary cr objects

rcohencyberarmor:


repositories/apiserver.go
```diff
@@ -222,24 +310,37 @@ func (a *APIServerStore) storeCVESummary(ctx context.Context, cve domain.CVEMani
 	if cve.Labels == nil {
```
matthyx, line 310: maybe we can skip this and make a nil check in the enriching func?

repositories/apiserver.go
```diff
@@ -222,24 +310,37 @@ func (a *APIServerStore) storeCVESummary(ctx context.Context, cve domain.CVEMani
...
 		cve.Labels = make(map[string]string)
 	}
+	if cve.Annotations == nil {
```
matthyx, line 313: same here

matthyx (approved):
LGTM, nit regarding creating maps for nothing

matthyx (approved):
much better, thx!

PR #141, opened 2023-08-16
Store summary cr objects

rcohencyberarmor:


repositories/apiserver.go
```diff
@@ -314,7 +314,11 @@ func getCVESummaryK8sResourceName(ctx context.Context) (string, error) {
 	if !ok {
 		return "", domain.ErrCastingWorkload
 	}
-	return workload.ImageSlug, nil
+	instanceID, err := instanceidhandler.GenerateInstanceIDFromString(workload.InstanceID)
+	if err != nil {
+		return "", nil
```
matthyx, line 319: Please document why no error

matthyx (changes requested):

repositories/apiserver.go
rcohencyberarmor: correct it

matthyx (approved):

PR #154, opened 2023-10-01
Support backend access key

rcohencyberarmor:
## PR Type:
Enhancement

___
## PR Description:
This PR introduces the support for access tokens in the backend adapter. The access token is passed as a parameter to the `NewBackendAdapter` function and is used for authorization in HTTP requests. The token is also loaded from a secret file. The PR includes changes in the function signatures, tests, and the addition of a new function to set headers for post requests. It also includes updates to dependencies in the go.mod file.

___
## PR Main Files Walkthrough:

`adapters/v1/backend.go`: The `NewBackendAdapter` function now accepts an additional parameter for the access token. The `getCVEExceptionsFunc` and `sendStatusFunc` functions have been updated to use the access token. The `accessToken` field has been added to the `BackendAdapter` struct.
`adapters/v1/backend_test.go`: The tests have been updated to reflect the changes in function signatures in the backend.go file.
`cmd/http/main.go`: The access token is now loaded from a secret file and passed to the `NewBackendAdapter` function.
`adapters/v1/backend_utils.go`: A new function `setPostResultHeaders` has been added to set headers for post requests, including the Authorization header using the access token.
`config/config.go`: Added functions to load and decode the access token from a secret file.
`go.mod`: The versions of several dependencies have been updated.
</details>

___


adapters/v1/backend_utils.go
```diff
@@ -62,6 +63,13 @@ func (a *BackendAdapter) postResultsAsGoroutine(ctx context.Context, report *v1.
 	}(report, eventReceiverURL, imagetag, wlid, errorChan, wg)
 }
 
+func (a *BackendAdapter) getRequestHeaders() map[string]string {
+	return map[string]string{
+		"Content-Type":              "application/json",
+		beServer.RequestTokenHeader: a.accessToken,
```
dwertent, line 69: As I wrote in the `backend` PR, we should add the header only if the string is not empty. (unless an empty header does not affect us)

dwertent (changes requested):

adapters/v1/backend_utils.go
amirmalka: Empty header does not affect

dwertent (approved):

adapters/v1/backend.go
```diff
@@ -122,11 +124,9 @@ func (a *BackendAdapter) SendStatus(ctx context.Context, step int) error {
 	report.ParentAction = workload.ParentJobID
 	report.Details = details[step]
 
-	ReportErrorsChan := make(chan error)
```
amirmalka, line 125: @rcohencyberarmor is this change part of the access token? is it a fix?

dwertent (approved):

adapters/v1/backend.go
rcohencyberarmor: yes, this is part of the change of the access token, the former version of this function is backend repo needed channel of errors, and in the newer versions it is not needed

Issue #155, opened 2023-10-10
Generation of VEX documents by the Kubescape relevancy engine


slashben:
## Overview
Kubescape calculates the relevancy of container image vulnerabilities by monitoring using eBPF the application behavior and produces a filtered list of vulnerabilities. Today the results are stored in the same format as the vulnerabilities, however the [VEX](https://www.ntia.gov/files/ntia/publications/vex_one-page_summary.pdf) seems to be a much better choice to store and publish this information. Kubescape needs to publish the filtered list of vulnerabilities in a VEX format. 

## Solution
In the current state, the *Kubevuln* is watching the filtered [SBOM objects](https://github.com/kubescape/storage/blob/main/pkg/apis/softwarecomposition/types.go), every time a new object is created or updated a filtered SBOM is created by the node-agent with only those modules that were loaded into the memory.

When a new filtered SBOM is available, the *Kubevuln* translates the SBOM to vulnerability list using Grype to create a filtered vulnerability list. 

In the same step when the filtered vulnerability is created,  *Kubevuln* should generate a VEX object. The object contains statements that all these vulnerabilities are loaded into the memory therefore they're relevant. This object should be stored as an API objects another vulnerability related

cc: @craigbox @puerco

puerco:
This is wonderful, please let us know if @openvex can help!

slashben:
@matthyx added support in Kubevuln PR #179 but it covers only creation and update. Where should we handle the cleanup of objects?

matthyx:
> @matthyx added support in Kubevuln PR #179 but it covers only creation and update. Where should we handle the cleanup of objects?

In the operator, check with @vladklokun

PR #179, opened 2023-10-18
Kubevuln support for VEX document creation

slashben:
## PR Type:
Enhancement

___
## PR Description:
This PR introduces the creation and update of VEX (Vulnerability EXchange) documents in Kubevuln. VEX is a document format that enables the sharing of vulnerability information about software vulnerabilities between different tools and systems. The implementation includes methods to create a new VEX document if it does not exist or update an existing one. The changes also include unit tests to verify the functionality.

___
## PR Main Files Walkthrough:

`repositories/apiserver.go`: Added methods to create and update VEX documents. These methods include logic to calculate a canonical hash for the VEX document, sort VEX statements, and create strings from vulnerabilities and components. Also, added the 'StoreVEX' method to the 'APIServerStore' struct.
`repositories/apiserver_test.go`: Added unit tests for the 'StoreVEX' method to verify the creation and update of VEX documents.
`core/ports/repositories.go`: Added the 'StoreVEX' method to the 'CVERepository' interface.
`core/services/scan.go`: Integrated the 'StoreVEX' method into the 'ScanCVE' method to store VEX documents during the scanning process.
`repositories/broken.go`: Implemented the 'StoreVEX' method in the 'BrokenStore' struct, which returns an expected error.
`repositories/memory.go`: Implemented the 'StoreVEX' method in the 'MemoryStore' struct, which currently does nothing and returns nil.
`go.mod`: Added the 'github.com/openvex/go-vex' package as a dependency for handling VEX documents.
</details>

___
## Overview

Implementation of VEX document creation and update flows in Kubevuln as per #155 
 
## Additional Information

VEX is a document format which enables the sharing of vulnerability information about software vulnerabilities in between different tools and systems.

Kubescape scans images for vulnerabilities and uses workload behavior information to determine whether these vulnerabilities are reachable for the attack (relevancy).

This enables this implementation to produce VEX document which describes which of the vulnerabilities need to be dealt with.

## How to Test

Added unit tests, otherwise VEX documents are produced together with filtered vulnerability reports.

## Related issues/PRs:

* Implements #155

put an [x] in the box to get it checked 

### DIFF
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -37,13 +37,14 @@ type ScanService struct {
 	cveRepository   ports.CVERepository
 	platform        ports.Platform
 	storage         bool
+	vexGeneration   bool
 	tooManyRequests *cache.Cache
 }
 
 var _ ports.ScanService = (*ScanService)(nil)
 
 // NewScanService initializes the ScanService with all injected dependencies
-func NewScanService(sbomCreator ports.SBOMCreator, sbomRepository ports.SBOMRepository, cveScanner ports.CVEScanner, cveRepository ports.CVERepository, platform ports.Platform, storage bool) *ScanService {
+func NewScanService(sbomCreator ports.SBOMCreator, sbomRepository ports.SBOMRepository, cveScanner ports.CVEScanner, cveRepository ports.CVERepository, platform ports.Platform, storage bool, vexGeneration bool) *ScanService {
 	return &ScanService{
 		sbomCreator:     sbomCreator,
 		sbomRepository:  sbomRepository,
@@ -51,6 +52,7 @@ func NewScanService(sbomCreator ports.SBOMCreator, sbomRepository ports.SBOMRepo
 		cveRepository:   cveRepository,
 		platform:        platform,
 		storage:         storage,
+		vexGeneration:   vexGeneration,
 		tooManyRequests: cache.New(cleaningInterval),
 	}
 }
@@ -234,10 +236,12 @@ func (s *ScanService) ScanCVE(ctx context.Context) error {
 				logger.L().Ctx(ctx).Warning("error storing CVE summary", helpers.Error(err),
 					helpers.String("imageSlug", workload.ImageSlug))
 			}
-			err = s.cveRepository.StoreVEX(ctx, cve, cvep, true)
-			if err != nil {
-				logger.L().Ctx(ctx).Warning("error storing VEX", helpers.Error(err),
-					helpers.String("imageSlug", workload.ImageSlug))
+			if s.vexGeneration {
+				err = s.cveRepository.StoreVEX(ctx, cve, cvep, true)
+				if err != nil {
+					logger.L().Ctx(ctx).Warning("error storing VEX", helpers.Error(err),
+						helpers.String("imageSlug", workload.ImageSlug))
+				}
 			}
 		}
 	}
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -234,6 +234,11 @@ func (s *ScanService) ScanCVE(ctx context.Context) error {
 				logger.L().Ctx(ctx).Warning("error storing CVE summary", helpers.Error(err),
 					helpers.String("imageSlug", workload.ImageSlug))
 			}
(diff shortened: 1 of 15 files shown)


matthyx:
At first glance, it looks OK, maybe I would add more unittest coverage for helper functions like `calculateVexCanonicalHash`.
We will probably need some system-tests coverage too.
cc @dwertent

repositories/memory.go
```diff
@@ -166,3 +166,11 @@ func (m *MemoryStore) StoreSBOM(ctx context.Context, sbom domain.SBOM) error {
 	m.sboms[id] = sbom
 	return nil
 }
+
+// StoreVEX stores a VEX to an in-memory map
+func (m *MemoryStore) StoreVEX(ctx context.Context, cve domain.CVEManifest, cvep domain.CVEManifest, withRelevancy bool) error {
+	_, span := otel.Tracer("").Start(ctx, "MemoryStore.StoreVEX")
```
dwertent, line 172: Remove span

repositories/apiserver.go
```diff
@@ -404,6 +408,325 @@ func (a *APIServerStore) StoreCVESummary(ctx context.Context, cve domain.CVEMani
...
+	// Calculate VEX
+	vexDoc := v1beta1.VEX{
+		Metadata: v1beta1.Metadata{
+			Context:     "https://openvex.dev/ns/v0.2.0",
+			Author:      "kubescape.io",
+			AuthorRole:  "smart vulnerability scanner :-)",
+			Timestamp:   timestamp,
+			LastUpdated: timestamp,
+			Version:     0,
+			Tooling:     "kubescape-vulnerability-analyzer",
```
dwertent, line 484: I think we should not have a new tool name here, we should either call this `kubescape`/`kubescape-operator`/`'kubevuln` so it is clear what is the tool.
WDYT?

repositories/apiserver.go
```diff
@@ -404,6 +408,325 @@ func (a *APIServerStore) StoreCVESummary(ctx context.Context, cve domain.CVEMani
...
+			Status:          v1beta1.Status(vex.StatusNotAffected),
+			Justification:   v1beta1.Justification(vex.VulnerableCodeNotPresent),
+			ImpactStatement: "Vulnerable component is not loaded into the memory",
+		})
+	}
+
+	// Now change the status of the filtered vulnerabilities to "Affected"
+	for _, v := range cvep.Content.Matches {
+		for i, s := range vexDoc.Statements {
+			if s.Vulnerability.ID == v.Vulnerability.ID && v.Artifact.PURL == s.Products[0].Subcomponents[0].ID {
```
dwertent, line 522: Are you sure the lists are not empty? `s.Products[0].Subcomponents[0]`

dwertent (changes requested):
It looks like the VEX is generated by default. We should add to the configuration file the "vex capability" so it will not always be generated.

repositories/apiserver.go
slashben: Kubescape and vulnerability is included already. This is ok to release since this is not a production grade feature anyway

repositories/memory.go
slashben: done

repositories/apiserver.go
slashben: Right, I have consolidated this with a proper loop and tested

slashben:
> It looks like the VEX is generated by default. We should add to the configuration file the "vex capability" so it will not always be generated.

Made this optional in commit https://github.com/kubescape/kubevuln/pull/179/commits/351ebe822ca5f60f24127bbad6fe0710dda9410b

core/services/scan.go
```diff
@@ -234,6 +236,13 @@ func (s *ScanService) ScanCVE(ctx context.Context) error {
 				logger.L().Ctx(ctx).Warning("error storing CVE summary", helpers.Error(err),
 					helpers.String("imageSlug", workload.ImageSlug))
 			}
+			if s.vexGeneration {
```
dwertent, line 239: @matthyx Where will you have this logic take place?

slashben: It is already done:
https://github.com/kubescape/helm-charts/pull/309

PR #185, opened 2023-11-08
send empty report if cve is known without relevant cve

matthyx:
## PR Type:
Enhancement

___
## PR Description:
This PR introduces a new feature where an empty report is sent if a CVE is known but without a relevant CVE. This is achieved by:
- Adding a new boolean field `wantEmptyReport` in the `MockPlatform` struct in `adapters/mockplatform.go`.
- Modifying the `NewMockPlatform` function to accept a new parameter `wantEmptyReport`.
- Modifying the `SubmitCVE` function to check if `wantEmptyReport` is true and the CVE content is nil, and if so, return an error.
- In `core/services/scan.go`, a new boolean variable `cveRetrievedFromStorage` is introduced to track if the CVE was retrieved from storage.
- If the CVE was retrieved from storage and the CVE' content is nil, an empty report is sent.
- The unit tests in `core/services/scan_test.go` are updated to reflect these changes.

___
## PR Main Files Walkthrough:

- `adapters/mockplatform.go`: Added a new boolean field `wantEmptyReport` in the `MockPlatform` struct. Modified the `NewMockPlatform` function to accept a new parameter `wantEmptyReport`. Modified the `SubmitCVE` function to check if `wantEmptyReport` is true and the CVE content is nil, and if so, return an error.
- `core/services/scan.go`: Introduced a new boolean variable `cveRetrievedFromStorage` to track if the CVE was retrieved from storage. If the CVE was retrieved from storage and the CVE' content is nil, an empty report is sent.
- `core/services/scan_test.go`: Updated the unit tests to reflect the changes in `adapters/mockplatform.go` and `core/services/scan.go`.
</details>

### DIFF
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -122,6 +122,7 @@ func (s *ScanService) ScanCVE(ctx context.Context) error {
 	defer span.End()
 
 	ctx = addTimestamp(ctx)
+	var cveRetrievedFromStorage bool // remember if we retrieved the CVE from storage
 
 	// retrieve workload from context
 	workload, ok := ctx.Value(domain.WorkloadKey{}).(domain.ScanCommand)
@@ -147,6 +148,7 @@ func (s *ScanService) ScanCVE(ctx context.Context) error {
 			logger.L().Ctx(ctx).Warning("error getting CVE", helpers.Error(err),
 				helpers.String("imageSlug", workload.ImageSlug))
 		}
+		cveRetrievedFromStorage = true
 	}
 
 	// if CVE manifest is not available, create it
@@ -189,6 +191,7 @@ func (s *ScanService) ScanCVE(ctx context.Context) error {
 		if err != nil {
 			return err
 		}
+		cveRetrievedFromStorage = false
 
 		// store CVE
 		if s.storage {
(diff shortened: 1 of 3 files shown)


core/services/scan.go
```diff
@@ -246,6 +249,11 @@ func (s *ScanService) ScanCVE(ctx context.Context) error {
 		}
 	}
 
+	// if CVE was retrieved from storage, and we don't have a CVE', we send an empty report
+	if cveRetrievedFromStorage && cvep.Content == nil {
```
Bezbran, line 253: I am not sure I follow.
As I understood the logic should be like
if the "new list of CVEs" ==="known list of CVEs" then don't send

matthyx: we don't store state... so with @dwertent we took this approach instead

Bezbran: So what we **do** check in this if statement?

Bezbran: what is the meaning of `cvep.Content == nil`?

matthyx: it means we don't have a relevant cve

Bezbran: Maybe I need to see it in action to understand the impact.
If we have relevant will we send the same full report over and over?

dwertent (approved):

PR #236, opened 2024-06-12
use application profile instead of sbomp

matthyx:


matthyx:
should not forget to update RBAC

PR #265, opened 2024-12-23
moved scanID to annotations, add timestamp annotation to summary

amirmalka:
# Overview

This PR changes the behavior so that:
1. Scan ID will be saved in the annotations instead of labels (label value is limited to 63 chars)
2. Add timestamp annotation to the summary, so that it will get updated on every scan request

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -325,6 +325,11 @@ func enrichSummaryManifestObjectAnnotations(ctx context.Context, annotations map
 	if !ok {
 		return nil, domain.ErrCastingWorkload
 	}
+	timestamp, ok := ctx.Value(domain.TimestampKey{}).(int64)
+	if !ok {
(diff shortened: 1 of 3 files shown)


repositories/apiserver.go
```diff
@@ -325,6 +325,11 @@ func enrichSummaryManifestObjectAnnotations(ctx context.Context, annotations map
 	if !ok {
 		return nil, domain.ErrCastingWorkload
 	}
+	timestamp, ok := ctx.Value(domain.TimestampKey{}).(int64)
+	if !ok {
+		return nil, domain.ErrMissingTimestamp
+	}
+	enrichedAnnotations["kubescape.io/timestamp"] = strconv.FormatInt(timestamp, 10) // TODO: use a constant
```
amirmalka, line 332: TODO: declare key in external package

matthyx (approved):

PR #266, opened 2024-12-25
Fix Application Profile Scans

amirmalka:
* Fix instance ID, Image ID in context - scan ID wasn't generated properly for each container scan
 * Scan only complete application profiles (skip partial)

### DIFF
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -25,6 +25,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/kubescape/go-logger"
 	"github.com/kubescape/go-logger/helpers"
+	v1 "github.com/kubescape/kubevuln/adapters/v1"
 	"github.com/kubescape/kubevuln/core/domain"
 	"github.com/kubescape/kubevuln/core/ports"
 	"go.opentelemetry.io/otel"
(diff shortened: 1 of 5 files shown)


core/services/scan.go
```diff
@@ -148,32 +149,36 @@ func (s *ScanService) ScanAP(ctx context.Context) error {
...
-		ctx = context.WithValue(ctx, domain.WorkloadKey{}, domain.ScanCommand{
+		ctx := enrichContext(mainCtx, domain.ScanCommand{
 			JobID:              uuid.NewString(),
 			ImageSlug:          slug,
 			ContainerName:      scan.ContainerName,
-			ImageHash:          scan.ImageID,
+			ImageHash:          v1.NormalizeImageID(scan.ImageID, scan.ImageTag),
 			ImageTagNormalized: scan.ImageTag,
-			InstanceID:         scan.InstanceIDString,
+			InstanceID:         instanceIDSlug,
```
amirmalka, line 175: Before using ScanAP API, the operator used to call ScanCVE API with `InstanceId` with the slug. (see https://github.com/kubescape/operator/blob/v0.2.49/mainhandler/vulnscan.go#L517)

In the new flow, the operator calls `ScanAP`, with empty `InstanceID` - which means, the logic has moved to kubevuln.
In addition, `InstanceIDString` (contains `/`) which is not aligned with the logic of `ValidateContainerScanID` - when using `InstanceIDString` it will cause a generation of new scan ID each time. this is why we are using instanceIDSlug.

core/services/scan.go
```diff
@@ -130,14 +131,14 @@ func (s *ScanService) Ready(ctx context.Context) bool {
 	return s.cveScanner.Ready(ctx)
 }
 
-func (s *ScanService) ScanAP(ctx context.Context) error {
-	ctx, span := otel.Tracer("").Start(ctx, "ScanService.ScanAP")
+func (s *ScanService) ScanAP(mainCtx context.Context) error {
+	mainCtx, span := otel.Tracer("").Start(mainCtx, "ScanService.ScanAP")
```
amirmalka, line 135: Separated the main context from the context that gets created in each iteration (container)

matthyx (approved):
Nice

PR #271, opened 2025-01-31
update k8s interface

amirmalka:
# Overview

Updating k8s-interface to include a fix for scanning statefulsets

https://github.com/kubescape/k8s-interface/pull/116


matthyx:
this fix should be done in operator, opening https://github.com/kubescape/operator/pull/288 instead

PR #274, opened 2025-02-11
add option to save filtered SBOMs

matthyx:


slashben:
Great @matthyx thanks!

slashben:
Do we need to add a configuration to the helm chart?

PR #278, opened 2025-03-03
normalize imageTag for scan AP

matthyx:
not sure it's actually useful...


amirmalka (approved):

amirmalka:
Can we check this with test?

matthyx:
> Can we check this with test?

good point, I think I've fixed it now

amirmalka (approved):

PR #284, opened 2025-04-04
fix: load config from path specified in env

### LINKED ISSUE #129: [question] can't load config file using `CONFIG` env

ian-fox:
Marking as draft and not requesting review until I figure out how to write a good test. Making PR against `main` because I don't actually see a `dev` branch (as mentioned in the PR template)

## Overview
Add a simple conditional to fix #129

## How to Test

`go test ./cmd/http` (once I figure that part out)

## Related issues/PRs:

Resolved #129


ian-fox:
Separate question: the PR template mentions making PRs against `dev`, which doesn't actually exist in this repo. It seems like it only exists in a few repos under the kubescape org, notably `kubescape/kubescape` and `kubescape/helm-charts`. Is that a branch that should exist in all of the other repos, or should I remove it from the template?

matthyx:
> Separate question: the PR template mentions making PRs against `dev`, which doesn't actually exist in this repo. It seems like it only exists in a few repos under the kubescape org, notably `kubescape/kubescape` and `kubescape/helm-charts`. Is that a branch that should exist in all of the other repos, or should I remove it from the template?

the template is stupid - you can fix it indeed :)

matthyx (approved):
thanks @ian-fox

ian-fox:
Thanks!

I'll make a PR to update the template and see what else I can do after work

PR #323, opened 2026-01-30
Feature: Vulnerabilitymanifest stores into workload namespace.

Mujib-Ahasan:
## Overview

Currenty Kubescape operator saves VulnerabilityManifest in the kubescape namespace. The VulnerabilityManifestSummary is saved in the workload namespace. Users need to have access rights to the kubescape namespace in order to view detailed findings. This is an issue when users have limited access rights due to cluster RBAC restrictions. In organizations where clusters are managed by platform teams, users only get access to their own namespaces to manage their own workloads. Currently we have to give users access rights to the Kubescape namespace and we would rather not do that.

This PR is raised to address this problem. Changes are made so that VulnerabilityManifest can now gets stored to workload namespace.

## Summary by CodeRabbit

* **New Features**
  * Store CVE manifests across multiple namespaces with workload-specific namespace resolution and fallback to a default namespace.
* **Bug Fixes / Reliability**
  * Improved error propagation and retry handling during store/update operations to avoid silent failures and ensure conflicts are retried correctly.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -205,12 +205,12 @@ func (a *APIServerStore) StoreCVE(ctx context.Context, cve domain.CVEManifest, w
 	if cve.Content != nil {
 		manifest.Spec.Payload = *cve.Content
 	}
-	// Regardless of relevency, VulnerabilityManifests gets stored to default namespace
+	// Regardless of relevancy, VulnerabilityManifests gets stored to default namespace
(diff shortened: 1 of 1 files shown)


matthyx:
thanks @Mujib-Ahasan I'll try it in the coming days

PR #370, opened 2026-06-02
fix(scan): classify MANIFEST_SCHEMA_UNSUPPORTED + write per-workload …

### LINKED ISSUE #369: kubevuln treats registry "manifest schema unsupported" as an unexpected error; the image is silently never scanned with no per-workload diagnostic
# Description

`classifySBOMError` matches `MANIFEST_UNKNOWN` (image-not-found) but not `MANIFEST_SCHEMA_UNSUPPORTED`. The Syft adapter wraps the underlying `go-containerregistry` error verbatim, so the latter falls into the catch-all `ReasonSBOMGenerationFailed` bucket:

```go
// kubevuln/core/services/scan_failure_reasons.go:18-53 (current upstream/main)
func classifySBOMError(err error) string {
    if err == nil {
        return scanfailure.ReasonUnexpected
    }
    // Sentinel/typed branches run first:
    if errors.Is(err, sbomscanner.ErrScannerCrashed) {
        return scanfailure.ReasonScannerOOMKilled
    }
    if errors.Is(err, context.DeadlineExceeded) {
    }
    var transportErr *transport.Error
    if errors.As(err, &transportErr) {
        switch {
        case transportErr.StatusCode == http.StatusUnauthorized ||
            transportErr.StatusCode == http.StatusForbidden:
            return scanfailure.ReasonImageAuthFailed
        }
        // NOTE: only 401/403 are matched here; a schema rejection arrives as
        // HTTP 400 + code MANIFEST_SCHEMA_UNSUPPORTED and falls through.
    }
    // String-based fallback — this is where the gap is:
    errStr := err.Error()
    switch {
    case strings.Contains(errStr, "401 Unauthorized") || strings.Contains(errStr, "403 Forbidden"):
        return scanfailure.ReasonImageAuthFailed
    case strings.Contains(errStr, "MANIFEST_UNKNOWN"):
        return scanfailure.ReasonImageNotFound
    }
    return scanfailure.ReasonSBOMGenerationFailed   // <- everything else, including schema rejection
}
```

Observed on a stock chart 1.40.2 install — the kubevuln scheduler enumerates every running image in the cluster including the Kind control-plane components, and one of them (`kindest/local-path-provisioner`) ships an OCI image whose registry rejects `go-containerregistry`'s pull:

```
{"level":"error","ts":"2026-05-28T19:26:14Z",
 "msg":"service error - ScanCVE",
 "error":"creating SBOM: an error occurred attempting to resolve 'index.docker.io/kindest/local-path-provisioner@sha256:3a195b…':
          oci-registry: failed to get image descriptor from registry:
          GET https://index.docker.io/v2/kindest/local-path-provisioner/manifests/sha256:3a195b…:
          MANIFEST_SCHEMA_UNSUPPORTED: manifest schema unsupported; unsupported schema manifest …",
 "wlid":"wlid://cluster-kind-kubescape-test/namespace-local-path-storage/deployment-local-path-provisioner",
 "imageSlug":"docker.io-kindest-local-path-provisioner-v20240813-c6f155d6-3a62ad", …}
```

Effect:

- The scan returns `fmt.Errorf("creating SBOM: %w", err)` and the loop `continue`s (`scan.go:230` / returns at line 431).
- No `VulnerabilityManifest` or `VulnerabilityManifestSummary` is ever written for this workload — the only signal that anything is wrong is the single `error` log line buried in the kubevuln pod.
- Because the reason is bucketed as `ReasonSBOMGenerationFailed`, the platform-reporting side (`s.platform.ReportScanFailure(...)`) sends a generic failure that is indistinguishable from a real SBOM-tooling crash. Operators triaging "this workload is missing vulnerability data" cannot tell whether to investigate Syft, registry creds, or an unsupported-schema source.
- There is no per-workload CR surfaced as `Annotations[kubescape.io/status]=unsupported-schema` or similar. UIs (Headlamp, ARMO dashboard) treat the workload as unscanned without explanation.

This is reproducible on every cluster running an image whose registry serves a schema `go-containerregistry`'s default resolver doesn't accept (OCI Image Index v1 with referrers, certain Harbor configurations, some ECR variants). It is *especially* visible on Kind because the default install scans Kind's own images.

# Environment

- Repo: `kubevuln` at `core/services/scan_failure_reasons.go`
- Affected callers: every `s.platform.ReportScanFailure(..., classifySBOMError(err), err)` site in `scan.go` (lines 119-120, 228-229, 429-430, 547-548). NOTE: `scan.go:252` is *not* affected — it calls `classifySBOMStatusWithAnnotation(sbom.Status, sbom.Annotations)` with a nil error (the Incomplete/TooLarge *status* path), which a schema rejection never reaches because that surfaces as an `err` from `CreateSBOM` and exits at 228/430 before the status check.
- Reproduced against `docker.io/kindest/local-path-provisioner:v20240813-c6f155d6` (chart 1.40.2 on Kind v1.31.0).
- Other known triggers: any image manifest with `mediaType: application/vnd.oci.image.index.v1+json` that the registry won't transcode; certain ECR cross-account pulls.

# Steps To Reproduce

```bash
kind create cluster
helm repo add kubescape https://kubescape.github.io/helm-charts/
helm upgrade --install kubescape kubescape/kubescape-operator -n kubescape \
   --create-namespace --set clusterName=$(kubectl config current-context)

# Trigger the cluster-wide scan (the chart's CronJob also fires daily)
kubectl -n kubescape create job --from=cronjob/kubevuln-scheduler oneshot

# Find the schema-rejected images
kubectl -n kubescape logs deploy/kubevuln | grep MANIFEST_SCHEMA_UNSUPPORTED

# The affected workload has no CR:
kubectl get vulnerabilitymanifest    -A | grep local-path-provisioner    # empty
kubectl get vulnerabilitymanifestsummary -A | grep local-path-provisioner # empty
```

# Expected behavior

- `classifySBOMError` should match `MANIFEST_SCHEMA_UNSUPPORTED` and return a dedicated reason (e.g. `ReasonImageSchemaUnsupported`) so the platform report and downstream UIs can distinguish "we can't read this registry's schema" from "Syft crashed". The reason already follows the same pattern as `MANIFEST_UNKNOWN`/`ReasonImageNotFound` — one-line analogue.
- Optionally: write a `VulnerabilityManifestSummary` stub with `status=unsupported-schema` and `severities` all zero, so consumers can see *something* for the workload instead of silence.
- Optionally: try one fallback resolver (e.g. pull as OCI image index → first manifest) before giving up; many "unsupported schema" rejections succeed when the resolver explicitly opts into OCI mediaTypes.

# Actual Behavior

- Schema rejection → `ReasonSBOMGenerationFailed`.
- No CR written; the workload is missing from `kubectl get vulnerabilitymanifest -A` results.
- Single `error` log line in the kubevuln pod is the only diagnostic.
- Platform report is indistinguishable from a real SBOM-tooling failure.

# Suggested fix

`kubevuln/core/services/scan_failure_reasons.go`:

```go
errStr := err.Error()
switch {
case strings.Contains(errStr, "401 Unauthorized") || strings.Contains(errStr, "403 Forbidden"):
    return scanfailure.ReasonImageAuthFailed
case strings.Contains(errStr, "MANIFEST_UNKNOWN"):
    return scanfailure.ReasonImageNotFound
case strings.Contains(errStr, "MANIFEST_SCHEMA_UNSUPPORTED") ||
     strings.Contains(errStr, "unsupported schema manifest"):
    return scanfailure.ReasonImageSchemaUnsupported   // new constant in armoapi-go/scanfailure
}
```

The primary token is the uppercase registry code `MANIFEST_SCHEMA_UNSUPPORTED`; the lowercase `unsupported schema manifest` phrase varies by registry, so keep the uppercase match as the primary clause. This string-based match is sufficient for **both** forms the error can take — verified that the rejection arriving as a typed `*transport.Error` (HTTP 400 + the code) still falls through the `errors.As` 401/403 switch to this string fallback, and `transport.Error.Error()` includes the literal `MANIFEST_SCHEMA_UNSUPPORTED` substring. No separate `transport.Diagnostic` code branch is required.

Add the constant to `armoapi-go/scanfailure` and tests:

```go
// scan_failure_reasons_test.go — new case
{"schema unsupported", errors.New("MANIFEST_SCHEMA_UNSUPPORTED: manifest schema unsupported"), scanfailure.ReasonImageSchemaUnsupported},
```

For the larger visibility fix, the `continue` at `scan.go:230` and the `return` at `scan.go:431` could be replaced by a "store an empty summary with annotated status" branch — gated on the new reason — so UIs see a per-workload record. That is a deeper change and can be a follow-up.

A separate audit: `go-containerregistry`'s `remote.Image` calls in the Syft adapter could be replaced with `remote.Get` followed by an explicit `OCI Image Index` branch. The schema-unsupported errors come specifically from the schema-version check inside the manifest fetcher; opting into OCI mediaTypes at the puller layer often makes the rejection go away.

# Source

- `kubevuln/core/services/scan_failure_reasons.go:18-53` — `classifySBOMError`, the missing match.
- `kubevuln/core/services/scan_failure_reasons_test.go` — table-driven `TestClassifySBOMError`; add the new case to the `tests` slice (~lines 22-89, e.g. right after the `MANIFEST_UNKNOWN` case). Line 94 is the assertion (`result := classifySBOMError(tt.err)`), not where cases are added.
- `kubevuln/core/services/scan.go:119-120` — caller (RegisterCVEExceptions path).
- `kubevuln/core/services/scan.go:226-230` — `creating SBOM, skipping scan` log + `continue` that swallows the schema rejection.
- `kubevuln/core/services/scan.go:425-431` — `ScanCVE` single-workload return path with `creating SBOM: %w`.
- `armoapi-go/scanfailure` — defines `Reason*` constants; add `ReasonImageSchemaUnsupported`.

harshitg927:
## Overview

When a registry serves a manifest schema go-containerregistry can't read, SBOM creation fails with `MANIFEST_SCHEMA_UNSUPPORTED` (OCI Image Index v1 with referrers, some Harbor/ECR configs, and Kind's own `kindest/local-path-provisioner`).

**Current behavior:**
- `classifySBOMError` doesn't recognize the code, so it falls into the catch-all `ReasonSBOMGenerationFailed` - indistinguishable from a real Syft crash.
- No CR is written for the workload, so it disappears from `kubectl get vulnerabilitymanifestsummary` and the UIs treat it as silently unscanned, with no explanation.

**Future behavior:**
- `classifySBOMError` maps the rejection to the dedicated `scanfailure.ReasonImageSchemaUnsupported`, so platform reports and downstream UIs can tell "unreadable registry schema" apart from "Syft crashed".
- On a schema rejection, the `ScanCP` and `ScanCVE` paths write a stub `VulnerabilityManifestSummary` (zeroed severities, `kubescape.io/status: unsupported-schema`), so the workload shows up with an explanation instead of vanishing. Gated strictly on the new reason - no other failure mode changes.

## Additional Information

- The string match catches both forms the error takes - a plain wrapped string and a typed `*transport.Error` (HTTP 400 + code), since `transport.Error.Error()` includes the code. The uppercase registry code is the stable primary token; the lowercase `unsupported schema manifest` phrase varies by registry.
- Added `CVERepository.StoreCVESummaryStub(ctx, status)`, implemented across all three stores (apiserver / memory / broken). The apiserver impl reuses the existing annotation/label enrichment and, on update, refreshes annotations/labels only - it never overwrites an existing summary's Spec with the zeroed stub.
- **Depends on armosec/armoapi-go#XXX**, which adds the `ReasonImageSchemaUnsupported` constant. This is held as a draft until that PR merges and is tagged; the armoapi-go version here will then be bumped and the local `replace` dropped.

## How to Test

```bash
go build ./...
go test ./core/services/... ./repositories/...
Relevant tests:

TestClassifySBOMError - new cases for the uppercase code, the lowercase phrase, and a typed 400 *transport.Error
TestScanService_ScanCVE_SchemaUnsupportedStub - asserts the stub summary is written on a schema error and not on a generic SBOM error
End-to-end (per the issue repro): on a Kind cluster running the operator, trigger the cluster scan and confirm kubectl get vulnerabilitymanifestsummary -A | grep local-path-provisioner now returns a record carrying kubescape.io/status: unsupported-schema, and the kubevuln log classifies the failure as image_schema_unsupported.

Related issues/PRs:
Resolved #369
Depends on [armosec/armoapi-go#653](https://github.com/armosec/armoapi-go/pull/653)

## Summary by CodeRabbit

* **New Features**
  * Scans now record per-image CVE summary stubs when SBOM generation fails due to unsupported image schemas, storing metadata-only summaries (no vulnerability details).
* **Behavior**
  * Storage-enabled runs write or merge stub summaries while preserving existing full summaries; classification now recognizes schema-unsupported SBOM errors.
* **Tests**
  * Added unit tests for SBOM error classification, stub storage behavior, and repository/store handling.

### DIFF
--- a/core/ports/repositories.go
+++ b/core/ports/repositories.go
@@ -18,6 +18,7 @@ type CVERepository interface {
 	GetCVESummary(ctx context.Context) (*v1beta1.VulnerabilityManifestSummary, error)
 	StoreCVE(ctx context.Context, cve domain.CVEManifest, withRelevancy bool) error
 	StoreCVESummary(ctx context.Context, cve domain.CVEManifest, cvep domain.CVEManifest, withRelevancy bool) error
+	StoreCVESummaryStub(ctx context.Context, status string) error
 	StoreVEX(ctx context.Context, cve domain.CVEManifest, cvep domain.CVEManifest, withRelevancy bool) error
 }
 
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/akyoto/cache"
 	"github.com/armosec/armoapi-go/armotypes"
+	"github.com/armosec/armoapi-go/scanfailure"
 	mapset "github.com/deckarep/golang-set/v2"
 	"github.com/docker/docker/api/types/registry"
 	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
@@ -26,7 +27,6 @@ import (
 	"github.com/kubescape/kubevuln/core/domain"
 	"github.com/kubescape/kubevuln/core/ports"
 	"github.com/kubescape/kubevuln/internal/tools"
-	"github.com/armosec/armoapi-go/scanfailure"
 	"github.com/kubescape/storage/pkg/apis/softwarecomposition/v1beta1"
 	"github.com/kubescape/storage/pkg/registry/file/dynamicpathdetector"
 	"go.opentelemetry.io/otel"
@@ -225,8 +225,17 @@ func (s *ScanService) ScanCP(mainCtx context.Context) error {
 					if err != nil {
 						logger.L().Ctx(ctx).Error("creating SBOM, skipping scan", helpers.Error(err),
 							helpers.String("imageSlug", slug))
+						reason := classifySBOMError(err)
 						_ = s.platform.ReportScanFailure(ctx, scanfailure.ScanFailureSBOMGeneration,
-							classifySBOMError(err), err)
+							reason, err)
+						// surface a per-workload record for schema-unsupported images so UIs
+						// do not show the workload as silently unscanned
+						if s.storage && reason == scanfailure.ReasonImageSchemaUnsupported {
+							if stubErr := s.cveRepository.StoreCVESummaryStub(ctx, statusUnsupportedSchema); stubErr != nil {
+								logger.L().Ctx(ctx).Warning("storing schema-unsupported summary stub", helpers.Error(stubErr),
+									helpers.String("imageSlug", slug))
+							}
+						}
 						continue // we need the SBOM
 					}
 					// store SBOM
@@ -426,8 +435,17 @@ func (s *ScanService) ScanCVE(ctx context.Context) error {
 				sbom, err = s.sbomCreator.CreateSBOM(ctx, workload.ImageSlug, workload.ImageHash, workload.ImageTagNormalized, optionsFromWorkload(workload))
 				s.checkCreateSBOM(err, workload.ImageHash)
 				if err != nil {
+					reason := classifySBOMError(err)
 					_ = s.platform.ReportScanFailure(ctx, scanfailure.ScanFailureSBOMGeneration,
-						classifySBOMError(err), err)
+						reason, err)
+					// surface a per-workload record for schema-unsupported images so UIs
+					// do not show the workload as silently unscanned
+					if s.storage && reason == scanfailure.ReasonImageSchemaUnsupported {
+						if stubErr := s.cveRepository.StoreCVESummaryStub(ctx, statusUnsupportedSchema); stubErr != nil {
+							logger.L().Ctx(ctx).Warning("storing schema-unsupported summary stub", helpers.Error(stubErr),
+								helpers.String("imageSlug", workload.ImageSlug))
+						}
+					}
 					return fmt.Errorf("creating SBOM: %w", err)
 				}
 				// store SBOM
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -231,7 +231,7 @@ func (s *ScanService) ScanCP(mainCtx context.Context) error {
 						// surface a per-workload record for schema-unsupported images so UIs
 						// do not show the workload as silently unscanned
 						if s.storage && reason == scanfailure.ReasonImageSchemaUnsupported {
-							if stubErr := s.cveRepository.StoreCVESummaryStub(ctx, statusUnsupportedSchema); stubErr != nil {
+							if stubErr := s.cveRepository.StoreCVESummaryStub(ctx, helpersv1.UnsupportedSchema); stubErr != nil {
 								logger.L().Ctx(ctx).Warning("storing schema-unsupported summary stub", helpers.Error(stubErr),
 									helpers.String("imageSlug", slug))
 							}
@@ -441,7 +441,7 @@ func (s *ScanService) ScanCVE(ctx context.Context) error {
 					// surface a per-workload record for schema-unsupported images so UIs
 					// do not show the workload as silently unscanned
 					if s.storage && reason == scanfailure.ReasonImageSchemaUnsupported {
-						if stubErr := s.cveRepository.StoreCVESummaryStub(ctx, statusUnsupportedSchema); stubErr != nil {
+						if stubErr := s.cveRepository.StoreCVESummaryStub(ctx, helpersv1.UnsupportedSchema); stubErr != nil {
 							logger.L().Ctx(ctx).Warning("storing schema-unsupported summary stub", helpers.Error(stubErr),
 								helpers.String("imageSlug", workload.ImageSlug))
 						}
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -228,9 +228,10 @@ func (s *ScanService) ScanCP(mainCtx context.Context) error {
 						reason := classifySBOMError(err)
 						_ = s.platform.ReportScanFailure(ctx, scanfailure.ScanFailureSBOMGeneration,
 							reason, err)
-						// surface a per-workload record for schema-unsupported images so UIs
-						// do not show the workload as silently unscanned
-						if s.storage && reason == scanfailure.ReasonImageSchemaUnsupported {
+						if s.storage &&
+							subWorkload.Wlid != "" &&
+							subWorkload.ContainerName != "" &&
+							reason == scanfailure.ReasonImageSchemaUnsupported {
 							if stubErr := s.cveRepository.StoreCVESummaryStub(ctx, helpersv1.UnsupportedSchema); stubErr != nil {
 								logger.L().Ctx(ctx).Warning("storing schema-unsupported summary stub", helpers.Error(stubErr),
 									helpers.String("imageSlug", slug))
@@ -439,8 +440,12 @@ func (s *ScanService) ScanCVE(ctx context.Context) error {
 					_ = s.platform.ReportScanFailure(ctx, scanfailure.ScanFailureSBOMGeneration,
 						reason, err)
 					// surface a per-workload record for schema-unsupported images so UIs
-					// do not show the workload as silently unscanned
-					if s.storage && reason == scanfailure.ReasonImageSchemaUnsupported {
+					// do not show the workload as silently unscanned; skip image-only
+					// scans where there is no workload to attach the stub to
+					if s.storage &&
+						workload.Wlid != "" &&
+						workload.ContainerName != "" &&
+						reason == scanfailure.ReasonImageSchemaUnsupported {
 						if stubErr := s.cveRepository.StoreCVESummaryStub(ctx, helpersv1.UnsupportedSchema); stubErr != nil {
 							logger.L().Ctx(ctx).Warning("storing schema-unsupported summary stub", helpers.Error(stubErr),
 								helpers.String("imageSlug", workload.ImageSlug))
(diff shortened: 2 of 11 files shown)


harshitg927:
@matthyx, Heads up: this depends on a new ReasonImageSchemaUnsupported constant I added in armoapi-go - PR here: [armosec/armoapi-go](https://github.com/armosec/armoapi-go/pull/653).

I've kept this one as a draft for now since it won't build until that lands and gets tagged. Once it's merged and released, I'll bump the armoapi-go version here, drop the local replace, and mark it ready for review.

harshitg927:
@matthyx , I have bumped the armoapi-go version to v0.0.718, PTAL and let me know what do you think about the changes. 
Thank!

matthyx (commented):
## Review — schema-unsupported classification + stub summary

I checked this out locally on `issue-369` (`75713104`): `go build ./...` and `go test ./core/services/... ./repositories/...` both pass, and the earlier blocker (local `../armoapi-go` replace) is gone — `armoapi-go` is bumped to the released `v0.0.718` and the new `scanfailure.ReasonImageSchemaUnsupported` resolves. The classification logic, the per-subworkload context plumbing in `ScanCP`, and the timestamp/workload context required by the annotation enrichment all look correct.

A few things I'd want addressed before merge:

### Blocker — the only production stub path (`APIServerStore.StoreCVESummaryStub`) has no test
`repositories/apiserver.go` is where the real logic lives: annotation/label enrichment, namespace fallback, the `AlreadyExists` → `RetryOnConflict` update branch, and the deliberate "refresh annotations/labels but keep existing Spec" behaviour. None of it is exercised. `TestScanService_ScanCVE_SchemaUnsupportedStub` only validates the in-memory recorder (`MemoryStore.summaryStubs`), which is a slice append — it can't catch a regression in the apiserver write. Please add a repository-level test (the existing `StoreCVESummary` tests are a template) covering at least: fresh create, and the update path asserting the existing `Spec` is preserved while the `kubescape.io/status` annotation is set.

### Question — does the consumer recognize `unsupported-schema`?
The whole point of the stub is that UIs / `vulnerabilitymanifestsummary` consumers stop treating the workload as silently unscanned. `unsupported-schema` is a new, kubevuln-coined `kubescape.io/status` value not in the shared `helpersv1` vocabulary (`incomplete`/`too-large`). If the operator or backend filters summaries by a known-status allowlist, a stub with an unrecognized status could still be dropped and the feature wouldn't achieve its goal. Can you confirm the downstream actually handles this value (link to the consuming code/PR)?

### Non-blocking — stale status on a previously-scanned workload
On the `AlreadyExists` branch the stub merges `status: unsupported-schema` onto an existing summary while keeping its old `Spec`. That yields a summary reporting real severities but an `unsupported-schema` status. It's a low-likelihood case (the same image digest won't flip schema support), but the combination is misleading — worth a short comment noting the intent, or skipping the annotation overwrite when a real Spec is already present.

### Non-blocking — PR description is now stale
The description still says "**Depends on armosec/armoapi-go#XXX**", "held as a draft until that PR merges", and references the local `replace`. The code has already moved past all of that (released `v0.0.718`, no replace, PR not a draft). Please update the description so reviewers aren't misled.

Nice work overall — the classification + context handling is solid; the gaps above are mostly about test confidence on the storage write and confirming the downstream contract.

harshitg927:
@matthyx Thanks for the review, I will take a look and update the pull request ASAP!

harshitg927:
@matthyx , I have made the required changes according to your review.

Also, I followed up on the downstream question - turns out two repos own the kubescape.io/status vocabulary: k8s-interface defines the valid values, and storage enforces them via ValidateStatusAnnotation (an allowlist that rejects anything unknown). unsupported-schema was in neither, so I've added it to both and opened PRs.
Since they chain on each other, merge order would be:
k8s-interface first - merge + release (it adds the UnsupportedSchema constant the others reference)
then storage and kubevuln - bump to that release, drop the temporary local replace, and merge
Links:
 [k8s-interface PR]: https://github.com/kubescape/k8s-interface/pull/155 
[storage PR]: https://github.com/kubescape/storage/pull/329. 
I'll swap the temp replaces for the tagged version once k8s-interface is out.

I am keeping this pr as draft till the other ones are merged and ready to go.

harshitg927:
@matthyx , PTAL and let know if anything else needs to be done. Thanks!

matthyx:
@harshitg927 https://github.com/kubescape/k8s-interface/releases/tag/v0.0.214 is out

harshitg927:
@matthyx , I have bumped the version of the k8s-interface in this pr and the storage pr as well, please review and share your thoughts! Thanks!

matthyx (approved):
thanks @harshitg927 !

PR #377, opened 2026-06-21
fix(sbomscanner): don't double-normalize an already-normalized image reference

kooomix:
## Problem

Registry image scans fail when the SBOM-scanner **sidecar** is enabled (`kubevuln.sbomScanner.enabled: true`). `normalizeImageID` is called **twice** on the sidecar path:

1. The adapter (`adapters/v1/sidecar.go`) normalizes the reference before sending the gRPC `CreateSBOMRequest`.
2. `scannerServer.CreateSBOM` (`pkg/sbomscanner/v1/server.go`) normalizes it **again**.

`normalizeImageID` assumes its `imageID` argument is empty or a bare digest. But on the second call it is already a complete reference. Registry scans always send an **empty image digest** (the operator does not resolve a digest for registry scans), so:

- 1st pass: `normalizeImageID("", tag)` → returns the tag.
- 2nd pass: `normalizeImageID(tag, tag)` → `tag` is non-empty and not a digest, so it gets grafted onto `imageTag` as a fake digest → `repo@sha256:<tag>`.

The result is an unparseable reference and the scan fails immediately:

```
"level":"error","msg":"service error - ScanRegistry",
 "error":"... could not parse reference: quay.io/systemtests/webgoat@sha256:quay.io/systemtests/webgoat:latest",
 "imageHash":""
```

No SBOM is produced and the scan never completes. (The in-process path normalizes once and is unaffected, so this only reproduces with the sidecar enabled.)

## Fix

Make `normalizeImageID` idempotent: if `imageID` is already a complete reference, return it unchanged —
- it equals `imageTag` (no digest known → scanned by tag), or
- it already carries a digest (`repo@sha256:...`).

This is the minimal, defensive fix and does not change the single-normalization (in-process) path.

### Alternative considered
`scannerServer.CreateSBOM` could simply not re-normalize and treat `req.ImageId` as the final reference (the adapter already normalized it). That's cleaner architecturally but changes the gRPC contract assumption; happy to take that direction instead if preferred.

## Tests
Added regression cases to `TestNormalizeImageID` for an already-normalized tag (the empty-digest case) and an already-normalized digest reference. `go test ./pkg/sbomscanner/v1/ -run TestNormalizeImageID` passes.

## Repro
kubevuln with `sbomScanner.enabled: true`; `POST /v1/scanRegistryImage` with `imageTag` set and `imageHash` empty.

## Release Notes

* **Bug Fixes**
  * Prevented incorrect handling of fully-formed image references by ensuring they’re passed through unchanged and only normalized at the correct point in the request flow.

* **Tests**
  * Added a regression test covering the case where a scan provides an empty digest while a tag is available, verifying the request is formed with the tag as expected.

### DIFF
--- a/adapters/v1/sidecar_test.go
+++ b/adapters/v1/sidecar_test.go
@@ -116,6 +116,34 @@ func TestSidecarSBOMAdapter_CreateSBOM_CrashRetry(t *testing.T) {
 	assert.Equal(t, 3, callCount)
 }
 
+// TestSidecarSBOMAdapter_CreateSBOM_EmptyDigestSendsTag is the regression guard for
+// the double-normalization bug. A registry scan arrives with an empty image digest
+// (imageID == "") and only a tag. The adapter is the single normalization point, so
+// it must send the tag as-is to the scanner — ImageID == ImageTag == the tag — and
+// must NOT graft the tag into a fake digest like repo@sha256:<tag>.
+func TestSidecarSBOMAdapter_CreateSBOM_EmptyDigestSendsTag(t *testing.T) {
+	const imageTag = "quay.io/systemtests/webgoat:latest"
+
+	var got sbomscanner.ScanRequest
+	mock := &mockScannerClient{
+		healthVersion: "v0.100.0",
+		healthReady:   true,
+		createSBOMFunc: func(_ context.Context, req sbomscanner.ScanRequest) (*sbomscanner.ScanResult, error) {
+			got = req
+			return &sbomscanner.ScanResult{Status: helpersv1.Learning}, nil
+		},
+	}
+
+	adapter := NewSidecarSBOMAdapter(mock, 5*time.Minute, 512*1024*1024, 20*1024*1024, false, "5Gi", nil)
+
+	_, err := adapter.CreateSBOM(context.Background(), "test-sbom", "", imageTag, domain.RegistryOptions{})
+	require.NoError(t, err)
+
+	assert.Equal(t, imageTag, got.ImageID, "empty digest must scan by tag, not be grafted into a digest")
+	assert.Equal(t, imageTag, got.ImageTag)
+	assert.Equal(t, got.ImageTag, got.ImageID, "ImageID and ImageTag must match for an empty-digest registry scan")
+}
+
 func TestSidecarSBOMAdapter_Version(t *testing.T) {
 	mock := &mockScannerClient{
 		healthVersion: "v0.100.0",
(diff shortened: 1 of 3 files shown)


matthyx:
@kooomix I'd prefer you implement the alternative, so we have one single normalization - I know it's more work but I think worth it

kooomix:
@matthyx done — switched to a single normalization in 55ba274.

The adapter (`SidecarSBOMAdapter.CreateSBOM` → `NormalizeImageID`) is now the only place the reference is normalized; `scannerServer.CreateSBOM` treats `req.ImageId` as the final pull reference and no longer re-normalizes. I removed the duplicate `normalizeImageID` helper in `server.go` (and its now-redundant test) — the canonical `NormalizeImageID` in `adapters/v1` stays, covered by `adapters/v1/syft_test.go`.

Net effect for the empty-digest (registry-scan) case: the adapter returns the tag, and the server scans by tag instead of grafting it into `repo@sha256:<tag>`. `go build`/`go vet` clean and the package + `adapters/v1` tests pass locally.

matthyx (commented):
This direction looks right to me, and I don't think it requires a proto/gRPC shape change: `image_id` is generic enough as the final pull reference, and the server comment now documents that the sidecar adapter is the single normalization point.

Could you add the regression coverage at that boundary, though? A focused `adapters/v1/sidecar_test.go` case should assert that a registry scan with an empty digest (`imageID == ""`, `imageTag == "quay.io/systemtests/webgoat:latest"`) sends `ScanRequest.ImageID == ScanRequest.ImageTag == "quay.io/systemtests/webgoat:latest"` to the scanner client. That covers the actual double-normalization bug now that the server helper and its tests are gone.

kooomix:
Added in 531e5f6 — `TestSidecarSBOMAdapter_CreateSBOM_EmptyDigestSendsTag` in `adapters/v1/sidecar_test.go`.

It drives `SidecarSBOMAdapter.CreateSBOM` with `imageID == ""`, `imageTag == "quay.io/systemtests/webgoat:latest"` and asserts the captured `ScanRequest.ImageID == ScanRequest.ImageTag == "quay.io/systemtests/webgoat:latest"` (reusing the existing `mockScannerClient` to capture the request). I also confirmed it's a real guard: re-introducing the double normalization makes it fail with exactly `quay.io/systemtests/webgoat@sha256:quay.io/systemtests/webgoat:latest`.

Thanks for the review!

matthyx (approved):
great, thanks @kooomix

PR #380, opened 2026-06-25
fix(sbomscanner): register sqlite driver for newer RPM databases

### LINKED ISSUE #378: sqlite driver is required for cataloging newer RPM databases, none registered
### Describe the bug

When scanning certain newer images (e.g. RPM-based or newer base images like redis), `kubevuln` fails to generate the SBOM with the following database driver error:

```
service error - ScanCVE, error: "creating SBOM: failed to generate SBOM: failed to run tasks: sqlite driver is required for cataloging newer RPM databases, none registered: sql: unknown driver \"sqlite\" (forgotten import?)"
```

### Steps to reproduce

1. Deploy kubevuln / kubescape-operator.
2. Scan a newer image containing an RPM-based database.
3. Observe the following error in `kubevuln` pod logs:
   `sqlite driver is required for cataloging newer RPM databases, none registered: sql: unknown driver "sqlite" (forgotten import?)`

### Expected behavior

Scan completes successfully and cataloging works.

### Potential Fix

In Go, newer versions of database tools (like those used inside Syft/cataloger) require importing the sqlite driver so it registers itself. Adding the import statement `_ "github.com/mattn/go-sqlite3"` (or the pure-Go alternative `_ "modernc.org/sqlite"`) to the imports section of the relevant scanner initialization code will fix the forgotten import issue.

matthyx:
## Problem

Scanning newer RPM-based images (e.g. recent `redis`/UBI-style bases) fails in the SBOM scanner sidecar with:

```
service error - ScanCVE, error: "creating SBOM: failed to generate SBOM: failed to run tasks:
sqlite driver is required for cataloging newer RPM databases, none registered:
sql: unknown driver \"sqlite\" (forgotten import?)"
```

Fixes #378.

## Root cause

Two changes combined to surface this:

1. **syft v1.32.0** (introduced in `cd2e51c`, "migrate to grype v0.99.1", 2025-09-17) added `ensureSqliteDriverAvailable()` to its RPM (redhat) cataloger. Newer RPM databases use the sqlite format (`rpmdb.sqlite`), and Syft now requires the consumer to register a `database/sql` driver under the name **`sqlite`** — it does not bundle one.

2. **The SBOM scanner sidecar** (`cmd/sbom-scanner` + `pkg/sbomscanner/v1`, introduced in `36cd416`, "add SBOM scanner sidecar", 2026-03-23) generates SBOMs via Syft **without importing grype**. The main `cmd/http` binary transitively pulls `glebarez/go-sqlite` (which registers `"sqlite"`) through grype's DB layer, so it works. The sidecar has no such transitive import, so the driver is never registered.

Verified per-binary (redhat cataloger present / sqlite driver present):

| binary | redhat cataloger | sqlite driver |
|---|---|---|
| `cmd/http` | ✅ | ✅ (glebarez via grype) |
| `cmd/cli` | ❌ | – |
| `cmd/sbom-scanner` | ✅ | ❌ → **bug** |

## Fix

Register the pure-Go (CGO-free) `modernc.org/sqlite` driver via a blank import in the sidecar entrypoint (`cmd/sbom-scanner/main.go`). `modernc.org/sqlite` registers under the name `"sqlite"`, which is exactly what the cataloger's `sql.Open("sqlite", ...)` needs.

It is placed in the **binary entrypoint**, not the shared `pkg/sbomscanner/v1` package, on purpose: `cmd/http` imports that package *and* already has glebarez's `"sqlite"` driver, so a blank import there would cause a `sql: Register called twice for driver sqlite` panic at startup. Putting it in `cmd/sbom-scanner` keeps each binary at exactly one `"sqlite"` driver.

> Note: the issue suggested `mattn/go-sqlite3` as one option — that registers as `"sqlite3"` (and requires CGO), so it would *not* satisfy `sql.Open("sqlite", ...)`. `modernc.org/sqlite` is the correct, pure-Go choice.

## Testing

- Added `cmd/sbom-scanner/sqlite_driver_test.go` — asserts exactly one `"sqlite"` driver is registered (catches both the missing-driver regression and an accidental double registration) and that an in-memory sqlite DB opens + pings.
- `go build ./cmd/...` passes; `go vet ./cmd/...` clean.
- Confirmed no binary ends up with two `"sqlite"` drivers (no double-registration panic).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Improved support for scanning images that rely on SQLite-backed metadata, helping scans complete more reliably in those cases.

* **Bug Fixes**
  * Fixed an issue where the SQLite driver could be missing or not loaded consistently during scans.
  * Added coverage to confirm the SQLite connection is available and working as expected.

### DIFF
--- a/cmd/sbom-scanner/main.go
+++ b/cmd/sbom-scanner/main.go
@@ -11,6 +11,15 @@ import (
 	sbomscanner "github.com/kubescape/kubevuln/pkg/sbomscanner/v1"
 	pb "github.com/kubescape/kubevuln/pkg/sbomscanner/v1/proto"
 	"google.golang.org/grpc"
+
+	// Register a pure-Go sqlite driver under the name "sqlite". Syft's RPM
+	// (redhat) cataloger requires it to read newer sqlite-backed RPM databases
+	// (rpmdb.sqlite) found in recent RPM-based images. Unlike the main HTTP
+	// binary, this sidecar does not import grype, so it does not transitively
+	// pull a sqlite driver; without this import the cataloger fails with
+	// "sqlite driver is required for cataloging newer RPM databases, none
+	// registered". See https://github.com/kubescape/kubevuln/issues/378.
+	_ "modernc.org/sqlite"
 )
 
 func main() {
--- a/cmd/sbom-scanner/sqlite_driver_test.go
+++ b/cmd/sbom-scanner/sqlite_driver_test.go
@@ -1,0 +1,36 @@
+package main
+
+import (
+	"database/sql"
+	"testing"
+)
+
+// TestSqliteDriverRegistered guards against a regression of
+// https://github.com/kubescape/kubevuln/issues/378: Syft's RPM (redhat)
+// cataloger requires a sqlite driver registered under the name "sqlite" to read
+// newer sqlite-backed RPM databases. The sidecar does not import grype (which
+// would otherwise pull one transitively), so it must register one itself via a
+// blank import in main.go.
+func TestSqliteDriverRegistered(t *testing.T) {
+	var count int
+	for _, d := range sql.Drivers() {
+		if d == "sqlite" {
+			count++
+		}
+	}
+	if count == 0 {
+		t.Fatalf(`no sqlite driver registered; Syft's RPM cataloger needs sql.Open("sqlite", ...) to work`)
+	}
+	if count > 1 {
+		t.Fatalf("sqlite driver registered %d times; risk of \"Register called twice\" panic", count)
+	}
+
+	db, err := sql.Open("sqlite", ":memory:")
+	if err != nil {
+		t.Fatalf("sql.Open(sqlite): %v", err)
+	}
+	defer db.Close()
+	if err := db.Ping(); err != nil {
+		t.Fatalf("ping in-memory sqlite: %v", err)
+	}
+}
--- a/go.mod
+++ b/go.mod
@@ -46,6 +46,7 @@ require (
 	k8s.io/apimachinery v0.35.0
 	k8s.io/client-go v0.35.0
 	k8s.io/utils v0.0.0-20260319190234-28399d86e0b5
+	modernc.org/sqlite v1.38.2
 	schneider.vip/problem v1.8.1
 )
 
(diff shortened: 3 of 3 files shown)


matthyx:
## Review — fix(sbomscanner): register sqlite driver for newer RPM databases

The fix itself is **correct and well-targeted**. Diagnosis (syft v1.32.0's redhat cataloger needs a `database/sql` driver named `sqlite` for `rpmdb.sqlite`; the sidecar lacks grype's transitive `glebarez` import) matches the code, and the placement choice is right:

- The blank import lives in `package main` of `cmd/sbom-scanner`, so it only affects that binary — `cmd/http` keeps its single `glebarez` `"sqlite"` driver and cannot hit a `Register called twice` panic. ✅
- `modernc.org/sqlite` registers under `"sqlite"` (not `"sqlite3"`), so it actually satisfies `sql.Open("sqlite", ...)`; pure-Go, no CGO. Moving it `indirect → direct` in `go.mod` (same `v1.38.2`, no `go.sum` churn) is correct. ✅
- The added `TestSqliteDriverRegistered` passed in CI (`ok cmd/sbom-scanner`) and guards both the missing-driver regression and accidental double registration. ✅

### Blockers

1. **DCO check failing (merge blocker, action required).** The single commit `347f0f8` has no `Signed-off-by:` trailer (`Co-Authored-By` does not satisfy DCO). Fix locally and force-push:
   ```
   git rebase HEAD~1 --signoff
   git push --force-with-lease origin fix/sqlite-driver-rpm-378
   ```

### Not a blocker from this PR (but the red check that's worrying)

2. **`pr-created / test / Create cross-platform build` FAILURE is pre-existing/unrelated.** The two failures are `Test_grypeAdapter_DBVersion` (`adapters/v1`) and `TestScanService_NginxTest` (`core/services`), both panicking at `GrypeAdapter.Ready` → `os.Exit(0)` (`grype.go:223`). That `os.Exit(0)` only fires when the grype DB update fails (the CI runner failed to download/load the vuln DB — a network/infra flake in these testcontainer-backed integration tests). This PR touches only `cmd/sbom-scanner/*` and `go.mod` — none of `grype.go`, `adapters/v1`, or `core/services` — so it cannot have caused these. A re-run (or running in an env where the grype DB downloads) should clear it. Worth a separate issue: `Ready()` calling `os.Exit(0)` on DB-update failure makes these tests crash the whole `go test` process rather than fail cleanly.

**Verdict:** code LGTM. Only the **DCO sign-off** must be fixed before merge; the cross-platform test red is unrelated infra flakiness, not a defect in this change.

PR #342, opened 2026-04-12
Integrate SecurityException CRDs into vulnerability scanning

slashben:
## Summary

- Adds `SecurityExceptionAdapter` using dynamic client to read `SecurityException` and `ClusterSecurityException` CRDs
- Converts CRD exceptions to `armotypes.VulnerabilityExceptionPolicy` for the existing matching flow
- Merges CRD exceptions with cloud exceptions in `GetCVEExceptions()`
- Handles expiry (`expiresAt`), `expiredOnFix`, and resource matching
- Wired at startup via in-cluster config

**Note:** `go.mod` currently has a `replace` directive pointing to the local storage checkout. This needs to be updated to the tagged storage release after kubescape/storage#309 is merged.

Depends on:
- kubescape/storage#309 (types)
- kubescape/helm-charts#817 (CRDs + RBAC)

Part of NAUT-1258: GitOps-native risk acceptance via SecurityException CRDs.

## Test plan

- [x] Conversion tests: basic CVE, expiry skip, expiredOnFix flag, resource matching (6 tests passing)
- [x] Merge test: cloud + CRD exceptions combined
- [ ] Remove `replace` directive after storage release is tagged
- [ ] `go build ./...` passes (currently blocked by replace directive — will pass after storage tag)
- [ ] E2E: apply a SecurityException CRD, trigger scan, verify CVE is excepted

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Release Notes

* **New Features**
  * Kubernetes SecurityException CRDs: Define vulnerability exception policies directly in your cluster
  * Unified exception management: Cloud and cluster-based exceptions are merged and applied together
  * Exception lifecycle: Configure expiration dates and automatic expiration when vulnerabilities are fixed

PR #383, opened 2026-07-14
[ LFX 2026 ] feat: scope vulnerability SecurityExceptions by match selectors

yugal07:
**Repo:** `kubescape/kubevuln`
**Branch:** `yugal07:lfx-securityexception-vuln-match` → `kubescape/kubevuln:main`
**Parent issue:** kubescape/kubescape#1982 (SecurityException CRD)
**Design doc:** `kubevuln/docs/security-exception-design.md`

## Summary

The vulnerability-scanning path only scoped `SecurityException` / `ClusterSecurityException`
CRDs by **namespace**. The `spec.match` selectors that the design doc defines —
`images`, `resources`, `objectSelector`, `namespaceSelector` — were never evaluated
when applying vulnerability exceptions, so an exception applied far more broadly than
declared (fail-open). This PR makes the vulnerability path honor all four selectors.

## Problem

In `adapters/v1/securityexception.go`, `ConvertToVulnerabilityExceptionPolicies` converted
every non-expired exception in the workload's namespace into an ignore policy keyed by CVE
ID, and `getCVEExceptionMatchCVENameFromList` (`adapters/v1/backend_utils.go`) matched
purely by CVE ID. Consequences:

- `match.images` was **never read** — a `ClusterSecurityException` scoped to
  `docker.io/library/nginx:*` suppressed its CVEs on **every** image cluster-wide.
- `match.objectSelector` / `match.namespaceSelector` were **never read** — a CSE scoped to
  `namespaceSelector: {env: staging}` suppressed its CVEs in production too.
- `match.resources` was converted into `PortalDesignator`s that the matcher then ignored,
  so kind/name scoping was not enforced either.
- Zero test coverage for any of these fields.

This is the same scoping the posture path already enforces
(kubescape/kubescape#2322 `namespaceSelector`, kubescape/kubescape#2480
`objectSelector.matchExpressions`); this PR brings the vulnerability path to parity.

## Solution

A new pure matcher decides, per scanned workload, whether an exception's `spec.match`
applies — then only matching exceptions are converted to ignore policies.

- **`adapters/v1/securityexception_match.go` (new)**
  - `ExceptionTarget` — the workload being scanned (namespace, kind, name, apiGroup,
    normalized image, and lazily-resolved workload/namespace labels).
  - `matchExceptionTarget(match, target, clusterScoped)` — AND across selector types,
    OR within `resources`/`images`; omitted selector matches everything.
    - `images`: `path.Match` glob against the fully-qualified `ImageTagNormalized`
      (`*` does not cross `/`, per the design doc).
    - `resources`: case-insensitive kind, optional name, optional apiGroup
      (enforced only when both sides are known).
    - `objectSelector`: standard label selector over the workload's labels.
    - `namespaceSelector`: standard label selector over the namespace's labels —
      **only** for `ClusterSecurityException`; ignored on a namespaced `SecurityException`
      (already scoped to its own namespace).
    - **Fail-closed**: an invalid selector, a label-lookup error, or a missing workload
      means the exception does **not** apply — a malformed exception never silently
      suppresses a finding.
  - `BuildExceptionTarget(...)` — assembles the target; workload/namespace labels are
    resolved **lazily**, only when an exception actually uses
    `objectSelector`/`namespaceSelector`, so the common path adds no extra API calls.
- **`adapters/v1/securityexception.go`** — `ConvertToVulnerabilityExceptionPolicies` now
  takes an `ExceptionTarget` and skips exceptions whose match does not apply.
- **`adapters/v1/backend.go`, `adapters/mockplatform.go`** — build the target and pass it in.
- **`core/ports/repositories.go`** — `SecurityExceptionRepository` gains
  `GetWorkloadLabels` and `GetNamespaceLabels`.
- **`repositories/apiserver.go`** — real label lookups via the dynamic client;
  the workload's `GroupVersionResource` is derived with
  `k8sinterface.GetGroupVersionResource` (same helper the store already uses).
  `repositories/noop_security_exception.go` returns nil.

Image matching relies on `ImageTagNormalized` being fully qualified, which
`tools.NormalizeReference` guarantees (e.g. `nginx` → `docker.io/library/nginx:latest`),
matching the design doc's "fully qualified, normalized image reference" contract.

## Testing

- **Unit** (`adapters/v1/securityexception_match_test.go`): glob matching (incl.
  `*`-does-not-cross-`/`, digest suffixes, malformed patterns), resource kind/name/apiGroup,
  label selectors (`matchLabels` + `matchExpressions`, invalid → fail-closed),
  AND/OR combination, cluster-vs-namespaced `namespaceSelector`, per-selector conversion
  scoping, and lazy label resolution in `BuildExceptionTarget`.
- **Integration** (`adapters/v1/backend_test.go`): a redis-image-scoped
  `ClusterSecurityException` is correctly **not** applied to an nginx workload
  (the exact fail-open regression).
- **Live** (kind cluster, `kubescape.io/v1beta1`): `GetWorkloadLabels` and
  `GetNamespaceLabels` resolve real labels; a missing workload returns empty/no-error;
  `GetGroupVersionResource` resolves `deployment → apps/v1/deployments` offline.
- `go build ./...`, `go vet` (changed non-test packages), and the SecurityException test
  suites pass. Existing exception behavior (expiry, `expiredOnFix`, CVE case-insensitive
  matching, cloud+CRD merge/precedence) is unchanged.

## Related

- **Parent:** kubescape/kubescape#1982 — SecurityException CRD
- **kubevuln (prior, this builds on):** #342 (integrate CRDs into vuln scanning),
  #349 (keepLocal mode), #358 (riskAcceptance flag)
- **kubescape posture parity:** #2322 (`namespaceSelector`), #2480 (`objectSelector`),
  #2023 (expiry filter), #2291 (exception match events)
- **helm-charts:** #817 (CRDs/RBAC), #859 (CEL validation), #875 (scanner RBAC)
- **headlamp-plugin:** #86 (SecurityException UI) · **operator:** #392 (watch + rescan)

## Summary by CodeRabbit

* **Bug Fixes**
  * Security exceptions are now applied only when their image, resource, workload labels, and namespace selectors match the scanned workload.
  * Cluster-scoped exceptions no longer affect unrelated workloads.
  * Invalid or missing selector information fails safely without applying unintended exceptions.

* **New Features**
  * Added support for matching exceptions using workload and namespace labels.
  * Workload and namespace labels are resolved automatically when required for exception evaluation.

### DIFF
--- a/adapters/v1/securityexception_match.go
+++ b/adapters/v1/securityexception_match.go
@@ -23,7 +23,11 @@ type ExceptionTarget struct {
 	Namespace string
 	Kind      string
 	Name      string
-	APIGroup  string
+	// APIGroup is the target's resolved API group. A nil pointer means the group
+	// is unknown (the kind could not be resolved); an empty string "" is the core
+	// group. These must stay distinct so a group-scoped exception never matches a
+	// core resource or an unverified one.
+	APIGroup *string
 	// Image is the fully-qualified, normalized image reference (as produced by
 	// tools.NormalizeReference), e.g. "docker.io/library/nginx:latest".
 	Image string
@@ -33,6 +37,14 @@ type ExceptionTarget struct {
 	// NamespaceLabels are the labels of the workload's namespace, resolved
 	// lazily and only when a ClusterSecurityException uses match.namespaceSelector.
 	NamespaceLabels map[string]string
+	// WorkloadLabelsResolved reports whether WorkloadLabels reflect a successful
+	// lookup. When an objectSelector is in play but resolution failed (missing
+	// workload, nil repo, lookup error), this stays false and matching fails
+	// closed — a negative selector (DoesNotExist/NotIn) would otherwise match an
+	// empty label set and wrongly suppress findings.
+	WorkloadLabelsResolved bool
+	// NamespaceLabelsResolved is the namespaceSelector equivalent.
+	NamespaceLabelsResolved bool
 }
 
 // matchExceptionTarget reports whether spec.match applies to the given target.
@@ -52,11 +64,17 @@ func matchExceptionTarget(match sev1beta1.ExceptionMatch, target ExceptionTarget
 	if !matchImages(match.Images, target.Image) {
 		return false
 	}
-	if !labelSelectorMatches(match.ObjectSelector, target.WorkloadLabels) {
-		return false
+	// objectSelector: when a selector is set but the workload's labels could not
+	// be resolved, fail closed rather than evaluate against an empty label set.
+	if match.ObjectSelector != nil {
+		if !target.WorkloadLabelsResolved || !labelSelectorMatches(match.ObjectSelector, target.WorkloadLabels) {
+			return false
+		}
 	}
-	if clusterScoped && !labelSelectorMatches(match.NamespaceSelector, target.NamespaceLabels) {
-		return false
+	if clusterScoped && match.NamespaceSelector != nil {
+		if !target.NamespaceLabelsResolved || !labelSelectorMatches(match.NamespaceSelector, target.NamespaceLabels) {
+			return false
+		}
 	}
 	return true
 }
@@ -74,9 +92,12 @@ func matchResources(resources []sev1beta1.ResourceMatch, target ExceptionTarget)
 		if r.Name != "" && r.Name != target.Name {
 			continue
 		}
-		// apiGroup is optional and defaults to all groups. Only enforce it when
-		// both the exception and the resolved target group are known.
-		if r.APIGroup != "" && target.APIGroup != "" && !strings.EqualFold(r.APIGroup, target.APIGroup) {
+		// apiGroup is optional. When the exception pins a group, the target's
+		// resolved group must equal it; if the target group is unknown (nil), the
+		// exception does not apply — fail closed rather than match a resource whose
+		// group was never verified. A non-nil "" is the core group and only matches
+		// an exception that pins "" (or none).
+		if r.APIGroup != "" && (target.APIGroup == nil || !strings.EqualFold(r.APIGroup, *target.APIGroup)) {
 			continue
 		}
 		return true
@@ -138,7 +159,8 @@ func BuildExceptionTarget(ctx context.Context, workload domain.ScanCommand, exce
 	// is simply skipped.
 	if kind != "" {
 		if gvr, err := k8sinterface.GetGroupVersionResource(kind); err == nil {
-			target.APIGroup = gvr.Group
+			group := gvr.Group
+			target.APIGroup = &group
 		}
 	}
 
@@ -146,21 +168,26 @@ func BuildExceptionTarget(ctx context.Context, workload domain.ScanCommand, exce
 		return target
 	}
 
+	// Labels are resolved only when a selector actually needs them. A failed
+	// resolution leaves the corresponding *Resolved flag false so the selector
+	// fails closed in matchExceptionTarget.
 	if usesObjectSelector(exceptions, clusterExceptions) && namespace != "" && kind != "" && name != "" {
 		if lbls, err := repo.GetWorkloadLabels(ctx, namespace, kind, name); err != nil {
-			logger.L().Ctx(ctx).Warning("failed to resolve workload labels for SecurityException objectSelector",
+			logger.L().Ctx(ctx).Warning("failed to resolve workload labels for SecurityException objectSelector; exception will not apply to this workload",
 				helpers.Error(err), helpers.String("namespace", namespace), helpers.String("kind", kind), helpers.String("name", name))
 		} else {
 			target.WorkloadLabels = lbls
+			target.WorkloadLabelsResolved = true
 		}
 	}
 
 	if usesNamespaceSelector(clusterExceptions) && namespace != "" {
 		if lbls, err := repo.GetNamespaceLabels(ctx, namespace); err != nil {
-			logger.L().Ctx(ctx).Warning("failed to resolve namespace labels for ClusterSecurityException namespaceSelector",
+			logger.L().Ctx(ctx).Warning("failed to resolve namespace labels for ClusterSecurityException namespaceSelector; exception will not apply to this namespace",
 				helpers.Error(err), helpers.String("namespace", namespace))
 		} else {
 			target.NamespaceLabels = lbls
+			target.NamespaceLabelsResolved = true
 		}
 	}
 
--- a/adapters/v1/securityexception_match.go
+++ b/adapters/v1/securityexception_match.go
@@ -11,6 +11,7 @@ import (
 	"github.com/kubescape/k8s-interface/k8sinterface"
 	"github.com/kubescape/kubevuln/core/domain"
 	"github.com/kubescape/kubevuln/core/ports"
+	"github.com/kubescape/kubevuln/internal/tools"
 	sev1beta1 "github.com/kubescape/kubevuln/pkg/securityexception/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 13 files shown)


yugal07:
@coderabbitai Addressed findings 2, 3 and 1 in f628fad.

**Finding 2 (label lookup fails open for negative selectors):** confirmed and fixed. `ExceptionTarget` now carries `WorkloadLabelsResolved` / `NamespaceLabelsResolved`, and `matchExceptionTarget` fails closed when a selector is set but its labels were not resolved. The apiserver repo now propagates `NotFound` as an error instead of returning nil labels, and the no-op repo returns an error too, so a missing workload, a nil repo, or a lookup error no longer lets a negative selector (`DoesNotExist` / `NotIn`) match an empty label set. Verified live against a kind cluster: a missing workload/namespace now returns a NotFound error, and an existing workload still resolves cleanly.

**Finding 3 (core group conflated with unknown group):** fixed. `ExceptionTarget.APIGroup` is now a `*string`, where nil means the group is unknown and "" means the core group. `matchResources` compares the pointed-to value, so a core-group resource no longer bypasses a group-scoped exception.

**Finding 1 (test expectation):** adopted the fail-closed behavior. When an exception pins an apiGroup and the target group is unknown, it now returns false, and I added cases for the core-group vs group-scoped and core-group vs groupless combinations.

**Finding 4 (objectSelector for cluster-scoped workloads):** declining for this PR. The vulnerability path only scans namespaced workloads, so an objectSelector on a cluster-scoped resource is never reached here; I kept the namespace guard. Happy to revisit if that assumption changes on the posture side.

New unit tests lock in the fail-closed behavior (unresolved vs resolved-but-empty labels, and lookup-error yields unresolved). Full adapters, repositories and core suites pass; build and vet are clean.

adapters/v1/securityexception_match.go
```diff
@@ -0,0 +1,218 @@
...
+// matches everything; a non-empty list never matches an empty image.
+func matchImages(patterns []string, image string) bool {
+	if len(patterns) == 0 {
+		return true
+	}
+	if image == "" {
+		return false
+	}
+	for _, p := range patterns {
+		if ok, err := path.Match(p, image); err == nil && ok {
```
matthyx: `path.Match` is a full-string anchored match, and `image` here is the fully-normalized `ImageTagNormalized`, which can include a digest suffix (`…@sha256:…`). So an **exact** pattern without a trailing `*` — e.g. `docker.io/library/nginx:1.25` — will silently fail to match `docker.io/library/nginx:1.25@sha256:abc`, and the exception won't apply. Safe (fail-closed) but a footgun for precise image pins.

Consider stripping the `@sha256:…` suffix before matching, or documenting that image patterns must account for an optional digest (end with `*`).

repositories/apiserver.go
```diff
@@ -149,6 +154,49 @@ func (a *APIServerStore) GetSecurityExceptions(ctx context.Context, namespace st
...
+		return nil, fmt.Errorf("failed to resolve GroupVersionResource for kind %q: %w", kind, err)
+	}
+	getCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), 30*time.Second)
+	defer cancel()
+	obj, err := a.DynamicClient.Resource(gvr).Namespace(namespace).Get(getCtx, name, metav1.GetOptions{})
+	if err != nil {
+		// Propagate NotFound as an error (rather than nil labels) so the caller
+		// fails closed: a negative objectSelector must not match a workload that
+		// could not be resolved.
+		return nil, err
```
matthyx, line 178: Propagating the error here (correctly) makes the exception fail closed — but note this also fires on RBAC `Forbidden` and transient API errors, not just `NotFound`. The scanner ServiceAccount must have `get` on every workload kind an `objectSelector` exception can target (and `get` on `namespaces` for `GetNamespaceLabels`), or those exceptions silently stop applying with only a `Warning` log. Please confirm the helm-charts RBAC (#875) covers all such kinds + namespaces, and consider a more visible signal than a log line when labels can't be resolved.

matthyx (commented):
## Review — scope vulnerability SecurityExceptions by match selectors

Nice work. This closes a genuine fail-open hole and the design is sound: matching is gated **before** conversion, selectors AND across types / OR within `resources`/`images`, `namespaceSelector` correctly restricted to `ClusterSecurityException`, and the whole thing fails **closed** on invalid selectors, lookup errors, and unresolved labels. Test coverage is thorough (glob edge cases, `*`-does-not-cross-`/`, negative-selector-vs-unresolved-labels, per-selector conversion scoping, lazy label resolution). `go build`/`vet` and the suites pass.

I did **not** find a hard merge-blocker. Two items below are worth resolving or explicitly confirming before merge, plus a couple of minor notes.

### 1. Operational (please confirm) — the label selectors depend on cluster RBAC, and fail *silently* closed
`objectSelector`/`namespaceSelector` now trigger live `GET`s on arbitrary workload kinds and on `namespaces` via the dynamic client. If the scanner ServiceAccount lacks `get` on a given workload kind (or on `namespaces`), `GetWorkloadLabels`/`GetNamespaceLabels` return an error → the exception fails closed → the CVE **reappears** — and the only signal is a `Warning` log line. That's the safe direction, but a mis-scoped RBAC will look like "exceptions randomly stopped working."

- Please confirm the helm-charts RBAC (referenced #875) grants `get` on **every** workload kind an `objectSelector` exception can target, plus `get` on `namespaces`.
- Consider surfacing "exception skipped because labels couldn't be resolved" somewhere more visible than a warning (metric/event), since transient API errors will silently un-apply exceptions too.

### 2. Correctness / UX — exact `match.images` patterns silently miss digest-bearing references
`matchImages` globs against the fully-normalized `ImageTagNormalized`, which can carry a digest suffix (`…:1.25@sha256:…` — your own test covers this for the wildcard case). Because `path.Match` is an anchored full-string match, a user who writes an **exact** pattern without a trailing `*`, e.g. `docker.io/library/nginx:1.25`, will **not** match `docker.io/library/nginx:1.25@sha256:abc` — the exception silently doesn't apply. Fail-closed, so not dangerous, but it's a real footgun for anyone writing a precise pin.

Suggest either normalizing away the `@sha256:…` suffix before matching, or documenting clearly (and ideally validating in the CRD/design doc) that image patterns must account for an optional digest (e.g. end with `*`).

### Minor
- `matchResources`: a fully-empty `ResourceMatch{}` entry (`Kind`/`Name`/`APIGroup` all empty) returns `true` and matches every workload. Probably fine if CRD validation forbids empty entries — worth a one-line guard or a validation note otherwise.
- `GetWorkloadLabels`/`GetNamespaceLabels` use `context.WithoutCancel(ctx)` + 30s timeout, so these lookups keep running after the scan context is cancelled. Intentional? A plain `ctx` with the timeout would still honor upstream cancellation.

None of these need to block if the RBAC in (1) is confirmed and the image-digest behavior in (2) is acceptable/documented.

yugal07:
On it. Thanks

yugal07:
Thanks @matthyx. Addressed both items in `90b6e4e`:

- **RBAC ( point 1):** confirmed it needed a chart change. Opened kubescape/helm-charts#879 to grant `kubevuln` read-only `get` on the standard workload kinds plus `namespaces`, gated under `capabilities.riskAcceptance`, and documented the required kinds in the design doc.
- **Image digests (point 2):** fixed. `matchImages` now tries each pattern against the equivalent forms of the reference (as-deployed, without the digest, bare repository), so an exact pin matches a digest-bearing ref while a digest-only ref still won't match a tag pattern.
- **Minor:** added a guard so an all-empty `ResourceMatch{}` no longer matches everything. Left `context.WithoutCancel` on purpose so label resolution completes if the scan's request context is cancelled. Happy to move the "labels couldn't be resolved" signal to a metric/event if you'd prefer that

matthyx:
Re-reviewed `90b6e4e`. Both concerns from my earlier review are resolved:

- **Image digest matching** — `matchImages` now tries each pattern against `tools.ReferenceMatchForms` (as-deployed / digest-stripped / bare-repo), so an exact tag pin like `docker.io/library/nginx:1.25` correctly matches a digest-bearing `…:1.25@sha256:…`, and the tag-vs-digest-only and registry-port cases are covered by tests. The design-doc table documenting the semantics is a nice touch.
- **RBAC** — the design doc now spells out the `get` verbs the scanner needs (namespaces + each targetable workload kind) and the fail-closed consequence.
- Empty `ResourceMatch{}` entries are now rejected with tests.

I didn't spot a new fail-open from the multi-form matching (bare-repo/tag-stripped widening is deliberate and documented). CI: unit tests + cross-platform build pass; the only red check is `build-and-push-image` failing at the Quay.io login step, which is the standard fork-PR secret limitation, not a code issue. LGTM from my side.

matthyx (approved):

PR #388, opened 2026-07-29
fix: revalidate SECURITY-INSIGHTS.yml and gate it in CI

1PoPTRoN:
## Overview

`SECURITY-INSIGHTS.yml` currently **fails validation against the very schema it declares** (`schema-version: 1.0.0`), and has drifted from reality for ~2 years. This corrects the manifest and adds a CI job so neither can happen silently again.

Validating the file on `main` against the official schema from the `ossf/security-insights` `v1.0.0` tag:

```
FAIL  SECURITY-INSIGHTS.yml: 2 schema violation(s)
  - project-lifecycle: Additional properties are not allowed ('stage' was unexpected)
  - project-lifecycle: 'status' is a required property
```

**Root cause.** The v1.0.0 property is `status`; `stage` was the pre-release name, renamed during v1.0.0 development ([ossf/security-insights#52](https://github.com/ossf/security-insights-spec/pull/52)). The file was authored against a draft and never machine-checked. Nothing in-tree reads it — no Go file, Makefile, Dockerfile or workflow — so nothing could ever fail because of it, and it rotted.

### What was wrong

| | Field | Was | Now |
|---|---|---|---|
| 1 | `project-lifecycle.stage` | invalid key → **schema failure** | `status: active` |
| 2 | `header.expiration-date` | `2024-10-12` (**656 days** past) | `2027-07-29` |
| 3 | `core-maintainers` | included `craigbox`, `dwertent` — **emeritus** per [project-governance/MAINTAINERS.md](https://github.com/kubescape/project-governance/blob/main/MAINTAINERS.md) | synced to the 5 core maintainers |
| 4 | `accepts-automated-pull-requests` | `false`, while Dependabot PRs #384–#386 were merged — and contradicting this file's own `security-testing` block | `true` + explicit `automated-tools-list` |
| 5 | `project-release` | `1.0.0` | `v0.3.159` (actual latest) |
| 6 | `documentation` | the **kubescape CLI's** docs | this repo's `docs/` + kubescape.io |
| 7 | `distribution-points` | GitHub repo only | + `quay.io/kubescape/kubevuln`, the actual artifact |
| 8 | `security-testing` | Dependabot only | + CodeQL (`sast`) and Snyk (`sca`), both confirmed in `kubescape/workflows/go-basic-tests.yaml` |
| 9 | CoC / security-policy | pointed at `kubescape/kubescape` | pointed at `project-governance`, matching this repo's own `CODE_OF_CONDUCT.md` / `SECURITY.md` |

Items 1 and 5–9 were found while fixing 2–4.

## Additional Information

**Deliberately not done: migration to spec v2.2.0.** v1.0.0 is three majors behind, but v2.0.0 was a breaking overhaul (`core-maintainers` → `core-team`). Migrating one repo while the rest of the org is on v1.0.0 would fragment the org's manifests, so this PR keeps `schema-version: 1.0.0` and makes the file *valid and true*. Happy to take the v2 direction instead if preferred.

**This is org-wide.** `kubescape/kubescape/SECURITY-INSIGHTS.yml` is byte-identical in every disputed field — same expiry, same 8-name list, same `accepts-automated-pull-requests: false`. This PR **intentionally diverges** from it. The org-wide sync and the v2 migration both feel like `project-governance` decisions rather than something to do unilaterally here — happy to open an issue there.

**Two calls for the reviewer:**

1. **`amitschendel` removed.** Governance lists him under *committers*, a separate table from core maintainers, and the schema field is `core-maintainers`. If the intent is "core team" (which is what v2.0.0 renamed the field to), say the word and I'll add him back — one line.
2. **`security-testing` omits GitGuardian, go-licenses and Scorecard** even though they run in CI. The `tool-type` enum is exactly `['sast','dast','iast','fuzzer','sca']` — there is no valid value for secret-scanning or license-scanning, so adding them would re-break validation.

**The `check-jsonschema` trap** (noted in the workflow): the obvious validator can't be used here. It metaschema-validates the schema file first, and upstream's `pgp-key` pattern is not a valid `format: regex`, so it errors before ever reading the manifest. The workflow uses `Draft7Validator` on the instance directly.

## How to Test

```bash
curl -sSL -o /tmp/si-schema.yaml \
  https://raw.githubusercontent.com/ossf/security-insights/v1.0.0/security-insights-schema.yaml
pip install pyyaml jsonschema

python3 - <<'EOF'
import yaml
from jsonschema import Draft7Validator
schema = yaml.safe_load(open("/tmp/si-schema.yaml"))
doc = yaml.safe_load(open("SECURITY-INSIGHTS.yml"))
errs = sorted(Draft7Validator(schema).iter_errors(doc), key=lambda e: list(e.path))
print("valid" if not errs else "\n".join(f"{'/'.join(map(str,e.path))}: {e.message}" for e in errs))
EOF
```

`git stash` the change and re-run to see the two failures on `main`.

Results:

```
main       → FAIL (2 violations: 'stage' unexpected / 'status' required)   exit=1
this PR    → SECURITY-INSIGHTS.yml is valid                                exit=0
```

Expiry gate:

```
2027-07-29 → 364 days remaining → pass
2024-10-12 → -656 days          → FAIL (expired)
```

Maintainer-drift check, parsing the governance table live:

```
governance: amirmalka Bezbran matthyx rotemamsa slashben
manifest:   amirmalka Bezbran matthyx rotemamsa slashben
→ IN SYNC
```

Blast radius: no in-tree consumer of the manifest; all 6 workflows parse; `go build ./...` and the hermetic test packages (`config`, `internal/tools`, `controllers`) unchanged and green.

## Related issues/PRs:

No tracking issue — filing directly per CONTRIBUTING's "small changes" path. Happy to open one if maintainers prefer.

## Summary by CodeRabbit

* **New Features**
  * Added automated validation for security metadata, including schema compliance and expiration-date checks.
  * Added advisory checks for discrepancies in listed core maintainers.
  * Added scheduled tracking for expiring security metadata.
* **Documentation**
  * Refreshed governance metadata, project links, lifecycle details, and security testing information, including CodeQL and Snyk.

### DIFF
--- a/.github/workflows/security-insights.yaml
+++ b/.github/workflows/security-insights.yaml
@@ -13,7 +13,11 @@ on:
       - 'SECURITY-INSIGHTS.yml'
       - '.github/workflows/security-insights.yaml'
   schedule:
-    - cron: '0 6 * * 1'
+    # Monthly, not weekly: the pull_request trigger above only fires when someone
+    # edits the manifest, and it went untouched from 2023 until this PR -- so a
+    # schedule is the only thing that ever runs the expiry check. Against the
+    # 60-day warning window below, monthly still fires twice before expiry.
+    - cron: '0 6 1 * *'
   workflow_dispatch:
 
 concurrency:
--- a/.github/workflows/security-insights.yaml
+++ b/.github/workflows/security-insights.yaml
@@ -1,0 +1,103 @@
+name: security-insights
+
+# SECURITY-INSIGHTS.yml is a hand-maintained manifest with a self-declared
+# expiry. Nothing consumes it in-tree, so it previously rotted unnoticed for
+# ~2 years (expired 2024-10-12, invalid `stage` key, emeritus maintainers).
+# This workflow makes both failure modes loud: schema drift fails on the PR
+# that introduces it, and impending expiry fails on a weekly schedule while
+# there is still time to review.
+
+on:
+  pull_request:
+    paths:
+      - 'SECURITY-INSIGHTS.yml'
+      - '.github/workflows/security-insights.yaml'
+  schedule:
+    - cron: '0 6 * * 1'
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  validate:
+    name: validate manifest
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+
+      - name: Validate against the OpenSSF Security Insights schema
+        run: |
+          # Schema tag is pinned to header.schema-version. Bump both together.
+          SCHEMA_VERSION=$(yq -r '.header.schema-version' SECURITY-INSIGHTS.yml)
+          echo "Validating against Security Insights v${SCHEMA_VERSION}"
+          curl -sSfL -o /tmp/si-schema.yaml \
+            "https://raw.githubusercontent.com/ossf/security-insights/v${SCHEMA_VERSION}/security-insights-schema.yaml"
+          # The [format] extra is required, not optional: the schema constrains 21
+          # fields with `format: iri` plus `date`, `date-time` and `idn-email`, and
+          # jsonschema silently skips every one of those checks unless the extra is
+          # installed. Without it FORMAT_CHECKER below is a no-op.
+          pip install --quiet pyyaml 'jsonschema[format]'
+          # NB: check-jsonschema cannot be used here. It metaschema-validates the
+          # schema file first, and the upstream v1.0.0 schema's `pgp-key` pattern
+          # is not a valid `format: regex`, so it errors out before ever looking
+          # at the manifest. Draft7Validator validates the instance directly.
+          python3 - <<'EOF'
+          import sys, yaml
+          from jsonschema import Draft7Validator
+
+          schema = yaml.safe_load(open("/tmp/si-schema.yaml"))
+          doc = yaml.safe_load(open("SECURITY-INSIGHTS.yml"))
+          validator = Draft7Validator(schema, format_checker=Draft7Validator.FORMAT_CHECKER)
+          errors = sorted(validator.iter_errors(doc), key=lambda e: list(e.path))
+          if not errors:
+              print("SECURITY-INSIGHTS.yml is valid")
+              sys.exit(0)
+          for e in errors:
+              loc = "/".join(str(p) for p in e.path) or "(root)"
+              print(f"::error file=SECURITY-INSIGHTS.yml::{loc}: {e.message}")
+          sys.exit(1)
+          EOF
+
+      - name: Fail if the manifest is expired or expires within 60 days
+        run: |
+          expiry=$(yq -r '.header.expiration-date' SECURITY-INSIGHTS.yml)
+          expiry_epoch=$(date -u -d "${expiry}" +%s)
+          now_epoch=$(date -u +%s)
+          days=$(( (expiry_epoch - now_epoch) / 86400 ))
+          echo "expiration-date: ${expiry} (${days} days remaining)"
+          if [ "${days}" -lt 0 ]; then
+            echo "::error file=SECURITY-INSIGHTS.yml::Manifest expired ${days#-} days ago. Re-review it against https://github.com/kubescape/project-governance/blob/main/MAINTAINERS.md and refresh header.last-reviewed / header.expiration-date."
+            exit 1
+          fi
+          if [ "${days}" -lt 60 ]; then
+            echo "::error file=SECURITY-INSIGHTS.yml::Manifest expires in ${days} days. Re-review it against https://github.com/kubescape/project-governance/blob/main/MAINTAINERS.md and refresh header.last-reviewed / header.expiration-date."
+            exit 1
+          fi
+
+      - name: Report core-maintainer drift from project-governance
+        # Advisory only: this parses a Markdown table in another repo, which is
+        # too brittle to gate a merge on. It annotates the run so the weekly
+        # schedule surfaces drift instead of waiting for the expiry to lapse.
+        continue-on-error: true
+        run: |
+          curl -sSfL -o /tmp/MAINTAINERS.md \
+            https://raw.githubusercontent.com/kubescape/project-governance/main/MAINTAINERS.md
+          # Core maintainers are the rows above the "commiters" heading.
+          sed -n '1,/commiters/p' /tmp/MAINTAINERS.md \
+            | grep -oE '\[@[A-Za-z0-9_-]+\]' | tr -d '[]@' | sort -fu > /tmp/governance.txt
+          yq -r '.project-lifecycle.core-maintainers[]' SECURITY-INSIGHTS.yml \
+            | sed 's/^github://' | sort -fu > /tmp/manifest.txt
+          if ! diff -iu /tmp/governance.txt /tmp/manifest.txt > /tmp/drift.txt; then
+            echo "::warning file=SECURITY-INSIGHTS.yml::core-maintainers differs from project-governance/MAINTAINERS.md"
+            cat /tmp/drift.txt
+          else
+            echo "core-maintainers matches project-governance/MAINTAINERS.md"
+          fi
--- a/.github/workflows/security-insights.yaml
+++ b/.github/workflows/security-insights.yaml
@@ -4,7 +4,7 @@ name: security-insights
 # expiry. Nothing consumes it in-tree, so it previously rotted unnoticed for
 # ~2 years (expired 2024-10-12, invalid `stage` key, emeritus maintainers).
 # This workflow makes both failure modes loud: schema drift fails on the PR
-# that introduces it, and impending expiry fails on a weekly schedule while
+# that introduces it, and impending expiry fails on a monthly schedule while
 # there is still time to review.
 
 on:
@@ -26,6 +26,10 @@ concurrency:
 
 permissions:
   contents: read
+  # Needed by the tracking-issue step below. Scheduled-run failure notifications
+  # go only to whoever last committed to this file, who may not be a maintainer,
+  # so a red run alone would leave the manifest rotting exactly as before.
+  issues: write
 
 jobs:
   validate:
@@ -42,7 +46,7 @@ jobs:
           # Schema tag is pinned to header.schema-version. Bump both together.
           SCHEMA_VERSION=$(yq -r '.header.schema-version' SECURITY-INSIGHTS.yml)
           echo "Validating against Security Insights v${SCHEMA_VERSION}"
-          curl -sSfL -o /tmp/si-schema.yaml \
+          curl -sSfL --connect-timeout 10 --max-time 60 -o /tmp/si-schema.yaml \
             "https://raw.githubusercontent.com/ossf/security-insights/v${SCHEMA_VERSION}/security-insights-schema.yaml"
           # The [format] extra is required, not optional: the schema constrains 21
           # fields with `format: iri` plus `date`, `date-time` and `idn-email`, and
@@ -71,32 +75,80 @@ jobs:
           EOF
 
       - name: Fail if the manifest is expired or expires within 60 days
+        id: expiry
         run: |
           expiry=$(yq -r '.header.expiration-date' SECURITY-INSIGHTS.yml)
           expiry_epoch=$(date -u -d "${expiry}" +%s)
           now_epoch=$(date -u +%s)
           days=$(( (expiry_epoch - now_epoch) / 86400 ))
           echo "expiration-date: ${expiry} (${days} days remaining)"
-          if [ "${days}" -lt 0 ]; then
-            echo "::error file=SECURITY-INSIGHTS.yml::Manifest expired ${days#-} days ago. Re-review it against https://github.com/kubescape/project-governance/blob/main/MAINTAINERS.md and refresh header.last-reviewed / header.expiration-date."
-            exit 1
+          # Classify on epoch seconds, not the truncated day count: a manifest that
+          # expired a few hours ago yields days=0, which would fall through to the
+          # "expires in 0 days" branch and report an already-invalid file as merely
+          # expiring soon.
+          if [ "${expiry_epoch}" -le "${now_epoch}" ]; then
+            message="SECURITY-INSIGHTS.yml EXPIRED on ${expiry}."
+          elif [ "${days}" -lt 60 ]; then
+            message="SECURITY-INSIGHTS.yml expires on ${expiry} (in ${days} days)."
+          else
+            exit 0
           fi
-          if [ "${days}" -lt 60 ]; then
-            echo "::error file=SECURITY-INSIGHTS.yml::Manifest expires in ${days} days. Re-review it against https://github.com/kubescape/project-governance/blob/main/MAINTAINERS.md and refresh header.last-reviewed / header.expiration-date."
-            exit 1
+          message="${message} Re-review it against https://github.com/kubescape/project-governance/blob/main/MAINTAINERS.md and refresh header.last-reviewed / header.expiration-date."
+          echo "message=${message}" >> "$GITHUB_OUTPUT"
+          echo "::error file=SECURITY-INSIGHTS.yml::${message}"
+          exit 1
+
+      - name: Open or refresh the expiry tracking issue
+        # Only on scheduled runs: on a pull_request the red check is already in
+        # front of someone who can act on it. On a schedule nothing else routes
+        # this to a maintainer, so raise an issue they will actually see.
+        if: ${{ always() && github.event_name == 'schedule' && steps.expiry.outcome == 'failure' }}
+        env:
+          GH_TOKEN: ${{ github.token }}
+          MESSAGE: ${{ steps.expiry.outputs.message }}
+          TITLE: 'SECURITY-INSIGHTS.yml needs review'
+        run: |
+          body="${MESSAGE}
+
+          Raised automatically by \`${GITHUB_WORKFLOW}\`: ${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}"
+          # Match on title rather than a label, so the step does not depend on a
+          # label existing in the repo.
+          existing=$(gh issue list --state open --search "in:title \"${TITLE}\"" \
+            --json number --jq '.[0].number // empty')
+          if [ -n "${existing}" ]; then
+            echo "refreshing existing issue #${existing}"
+            gh issue comment "${existing}" --body "${body}"
+          else
+            echo "opening a new tracking issue"
+            gh issue create --title "${TITLE}" --body "${body}"
           fi
 
       - name: Report core-maintainer drift from project-governance
         # Advisory only: this parses a Markdown table in another repo, which is
-        # too brittle to gate a merge on. It annotates the run so the weekly
+        # too brittle to gate a merge on. It annotates the run so the monthly
         # schedule surfaces drift instead of waiting for the expiry to lapse.
         continue-on-error: true
         run: |
-          curl -sSfL -o /tmp/MAINTAINERS.md \
+          curl -sSfL --connect-timeout 10 --max-time 60 -o /tmp/MAINTAINERS.md \
             https://raw.githubusercontent.com/kubescape/project-governance/main/MAINTAINERS.md
-          # Core maintainers are the rows above the "commiters" heading.
-          sed -n '1,/commiters/p' /tmp/MAINTAINERS.md \
-            | grep -oE '\[@[A-Za-z0-9_-]+\]' | tr -d '[]@' | sort -fu > /tmp/governance.txt
+          # Core maintainers are the rows above the committers heading. The range
+          # end is anchored to that heading and tolerates both spellings, so
+          # correcting the current upstream typo ("commiters") does not silently
+          # widen the range to the whole file and fold committers and emeritus
+          # maintainers into the comparison.
+          heading='^The following table lists the Kubescape project committ?ers'
+          if ! grep -qE "${heading}" /tmp/MAINTAINERS.md; then
+            echo "::warning file=SECURITY-INSIGHTS.yml::could not locate the committers heading in project-governance/MAINTAINERS.md; core-maintainer drift not checked"
+            exit 0
+          fi
+          # `|| true` keeps a no-match from tripping pipefail before the emptiness
+          # check below can distinguish "format changed" from "no drift".
+          sed -nE "1,/${heading}/p" /tmp/MAINTAINERS.md \
+            | grep -oE '\[@[A-Za-z0-9_-]+\]' | tr -d '[]@' | sort -fu > /tmp/governance.txt || true
+          if [ ! -s /tmp/governance.txt ]; then
+            echo "::warning file=SECURITY-INSIGHTS.yml::parsed no maintainers from project-governance/MAINTAINERS.md (link format may have changed); core-maintainer drift not checked"
+            exit 0
+          fi
           yq -r '.project-lifecycle.core-maintainers[]' SECURITY-INSIGHTS.yml \
             | sed 's/^github://' | sort -fu > /tmp/manifest.txt
           if ! diff -iu /tmp/governance.txt /tmp/manifest.txt > /tmp/drift.txt; then
--- a/SECURITY-INSIGHTS.yml
+++ b/SECURITY-INSIGHTS.yml
@@ -4,7 +4,6 @@ header:
   last-reviewed: '2026-07-29'
   expiration-date: '2027-07-29T00:00:00.000Z'
   project-url: https://github.com/kubescape/kubevuln/
-  project-release: v0.3.159
   changelog: https://github.com/kubescape/kubevuln/releases
   license: https://github.com/kubescape/kubevuln/blob/main/LICENSE
 project-lifecycle:
@@ -30,9 +29,7 @@ contribution-policy:
     path:
     - /
     comment: |
-      Dependabot opens dependency-bump PRs against main. Scanner dependencies
-      (github.com/anchore/syft, github.com/anchore/grype) are excluded from the
-      automated flow and are upgraded manually by maintainers.
+      Dependabot opens dependency-bump PRs against main.
   contributing-policy: https://github.com/kubescape/project-governance/blob/main/CONTRIBUTING.md
   code-of-conduct: https://github.com/kubescape/project-governance/blob/main/CODE_OF_CONDUCT.md
 documentation:
@@ -62,8 +59,9 @@ security-testing:
     ci: true
     before-release: false
   comment: |
-    CodeQL analysis runs on every pull request through the shared
-    kubescape/workflows go-basic-tests workflow.
+    CodeQL analysis runs on pull requests through the shared kubescape/workflows
+    go-basic-tests workflow (advisory; does not block merge), except for paths
+    excluded by that workflow's path filters.
 - tool-type: sca
   tool-name: Snyk
   tool-version: latest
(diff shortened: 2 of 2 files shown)


.github/workflows/security-insights.yaml
```diff
@@ -0,0 +1,98 @@
...
+          # schema file first, and the upstream v1.0.0 schema's `pgp-key` pattern
+          # is not a valid `format: regex`, so it errors out before ever looking
+          # at the manifest. Draft7Validator validates the instance directly.
+          python3 - <<'EOF'
+          import sys, yaml
+          from jsonschema import Draft7Validator
+
+          schema = yaml.safe_load(open("/tmp/si-schema.yaml"))
+          doc = yaml.safe_load(open("SECURITY-INSIGHTS.yml"))
+          errors = sorted(Draft7Validator(schema).iter_errors(doc), key=lambda e: list(e.path))
```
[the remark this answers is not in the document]
1PoPTRoN: Good catch, applied in 081eb8d - with one addition, because the suggestion as written would have been a no-op here.

`jsonschema` only runs a `format` checker if its optional dependency is installed; otherwise it silently skips it. This schema constrains **21** fields with `format: iri`, plus `date`, `date-time` and `idn-email` - and on a bare `pip install jsonschema` every one of those is inactive:

```
iri        active: False
date       active: False
date-time  active: False
idn-email  active: False
```

So `FORMAT_CHECKER` on its own changes nothing. The fix is `pip install 'jsonschema[format]'` **and** the format checker.

Verified by injecting three malformed values into the manifest:

| Injected | current code | FORMAT_CHECKER only | + `jsonschema[format]` |
|---|---|---|---|
| `changelog: "not a url"` | caught (via separate `pattern`) | caught (same reason) | caught as `iri` **and** `pattern` |
| `last-reviewed: "29-07-2026"` | **missed** | **missed** | **caught** |
| clean manifest | pass | pass | pass (no regression) |

The malformed date is the one that matters, since this PR is about date hygiene.

One caveat for the record: `not@an@email` still slips through `idn-email` - the `idna`-backed checker is lenient about the local part. Not worth chasing, but noting it so the check isn't assumed stricter than it is.

arp1traj:
Hi @matthyx , would appreciate a look when you have a moment.

It started as a small cleanup of `SECURITY-INSIGHTS.yml` (expired back in 2024, still listed craigbox and dwertent as core maintainers). While fixing it I found the file has never actually validated against the schema it declares, it uses stage: where `v1.0.0` wants status:, so anything doing schema validation rejects it outright. So the PR fixes that, the maintainer list, the accepts-automated-pull-requests: false contradiction with the Dependabot PRs we merge weekly, and a few stale fields. Then a small workflow so it can't quietly rot again.

Two calls I'd rather you make than guess at:

I dropped amitschendel from core-maintainers since governance lists him under committers but v2 of the spec renames that field to "core-team", so you may want him back in. Manifest fix and the CI workflow are in one PR. Happy to split them if you'd rather review separately. One thing I can't do from my side: it's from a fork, so the new workflow needs your approval before it'll run.

matthyx:
thanks @1PoPTRoN I wonder if we could have a global SECURITY-INSIGNTS.md file in https://github.com/kubescape/project-governance/tree/main that is common to all Kubescape projects?

arp1traj:
Yes @matthyx, and it turns out the spec has a first-class mechanism for it - though not quite as a single global file.

v2 splits the document into `project:` and `repository:`, and `header.project-si-source` lets a repo inherit the project half from a shared file. There are two official examples for exactly this layout: [multi-repository-project](https://github.com/ossf/security-insights/blob/v2.2.0/examples/example-multi-repository-project.yml) and [-reuse](https://github.com/ossf/security-insights/blob/v2.2.0/examples/example-multi-repository-project-reuse.yml).

So the shape would be one `security-insights.yml` in project-governance carrying `project:` (name, administrators, the repo list, vulnerability-reporting), and a slim file in each repo with `header.project-si-source` pointing at it plus a `repository:` block. The `repository:` half can't be centralized - it's per-repo by definition: url, status, core-team, license, security assessments.

Two catches:

1. This only works on v2.x. v1.0.0 (what all our repos are on) removed the parent mechanism in ossf/security-insights#57, so there's no inheritance available today. Centralizing means an org-wide v1 to v2 migration.
2. v2 ships a CUE schema rather than JSON Schema, so the validation step in this PR is v1-specific and would need reworking.

Given that, my suggestion would be to take this PR as-is for the immediate problem - the current file is expired *and* failing schema validation, and kubescape/kubescape has the byte-identical copy - then do the v2 centralization as a separate piece of work. Happy to prototype it: a `security-insights.yml` in project-governance plus the slim per-repo file for kubevuln as the first consumer, so we can see the shape before touching the other repos. Want me to open an issue there to scope it?

Also, whenever you get a chance - this is from a fork, so the new workflow needs your approval before it'll run.

.github/workflows/security-insights.yaml
```diff
@@ -0,0 +1,103 @@
...
+# that introduces it, and impending expiry fails on a weekly schedule while
+# there is still time to review.
+
+on:
+  pull_request:
+    paths:
+      - 'SECURITY-INSIGHTS.yml'
+      - '.github/workflows/security-insights.yaml'
+  schedule:
+    - cron: '0 6 * * 1'
```
matthyx: do we really need the weekly scan?

SECURITY-INSIGHTS.yml
```diff
@@ -1,30 +1,46 @@
...
+  accepts-automated-pull-requests: true
+  automated-tools-list:
+  - automated-tool: dependabot
+    action: allowed
+    path:
+    - /
+    comment: |
+      Dependabot opens dependency-bump PRs against main. Scanner dependencies
+      (github.com/anchore/syft, github.com/anchore/grype) are excluded from the
+      automated flow and are upgraded manually by maintainers.
```
matthyx: **Blocker — this describes a control that does not exist.**

There is no Dependabot configuration anywhere in this repo:

```
$ gh api repos/kubescape/kubevuln/git/trees/main?recursive=1 --jq '.tree[].path' | grep -i dependabot
(no match)
```

and there is none in `kubescape/.github` either (404). So there is no `ignore:` block, and nothing excludes `anchore/syft` or `anchore/grype` from anything.

With no `dependabot.yml`, what produced #384–#386 is Dependabot **security updates**, which are repo-setting-driven and cover *every* module in `go.mod` — syft and grype included, the moment an advisory lands on them. The manifest would be asserting the opposite of the actual behaviour.

That matters more here than it normally would: the premise of this PR is that the file drifted from reality and nobody noticed. Replacing an expired fact with an unimplemented one reintroduces the same failure class the workflow is meant to close.

Two ways out, either is fine by me:
- drop the second sentence and keep the comment to what is observable (`Dependabot opens dependency-bump PRs against main`), or
- add `.github/dependabot.yml` in this PR with the `ignore` entries for syft/grype, so the manifest and the config land together and the statement is true.

I'd lean towards the first — the exclusion policy is a maintainer decision we haven't actually taken.

SECURITY-INSIGHTS.yml
```diff
@@ -38,12 +54,32 @@ security-testing:
...
+- tool-type: sast
+  tool-name: CodeQL
+  tool-version: latest
+  integration:
+    ad-hoc: false
+    ci: true
+    before-release: false
+  comment: |
+    CodeQL analysis runs on every pull request through the shared
+    kubescape/workflows go-basic-tests workflow.
```
matthyx: **Blocker — "runs on every pull request" is not accurate**, on two counts.

1. **Path filters.** `pr-created.yaml` in this repo carries
   ```yaml
   paths-ignore:
     - '*.md'
     - '*.yaml'
     - '.github/workflows/*'
   ```
   and the shared `incluster-comp-pr-created.yaml` adds `'**/**.md'`. A PR that only touches workflows gets no CodeQL at all — which, incidentally, is true of the workflow half of *this* PR.

2. **CodeQL cannot fail a PR.** In `kubescape/workflows/.github/workflows/go-basic-tests.yaml` all three steps — `codeql-action/init`, `codeql-action/autobuild`, `codeql-action/analyze` — are `continue-on-error: true`. It is an advisory scan, not a gate.

`integration.ci: true` is the right value; it's the free-text `comment` that overstates. Something like "CodeQL analysis runs on pull requests via the shared kubescape/workflows go-basic-tests workflow (advisory; does not block merge), except for paths excluded by the workflow's path filters" would hold up.

SECURITY-INSIGHTS.yml
```diff
@@ -38,12 +54,32 @@ security-testing:
...
+- tool-type: sca
+  tool-name: Snyk
+  tool-version: latest
+  integration:
+    ad-hoc: false
+    ci: true
+    before-release: false
+  comment: |
+    Snyk runs on every pull request through the shared kubescape/workflows
+    go-basic-tests workflow when the SNYK_TOKEN secret is configured.
```
matthyx: **Blocker — same overstatement, and the fork case is not a corner case here.**

CodeRabbit flagged the fork exclusion; confirming it, and there is a second half it missed.

- Secrets are not exposed to `pull_request` runs from forks, so `Check-secret` resolves `run-snyk=false` and the Snyk step is **skipped entirely** on every fork PR. This PR is itself from a fork, so the very run that would demonstrate the claim is one where Snyk does not execute.
- The Snyk step is also `continue-on-error: true`, so even on same-repo PRs it never blocks a merge.

`when the SNYK_TOKEN secret is configured` reads as a config caveat; the real qualifier is "on same-repository pull requests, advisory only". Suggest saying that outright.

SECURITY-INSIGHTS.yml
```diff
@@ -1,30 +1,46 @@
...
   schema-version: 1.0.0
-  last-updated: '2023-10-12'
-  last-reviewed: '2023-10-12'
-  expiration-date: '2024-10-12T01:00:00.000Z'
+  last-updated: '2026-07-29'
+  last-reviewed: '2026-07-29'
+  expiration-date: '2027-07-29T00:00:00.000Z'
   project-url: https://github.com/kubescape/kubevuln/
-  project-release: 1.0.0
+  project-release: v0.3.159
```
matthyx: Not blocking, but worth a decision while we're here.

`project-release` is the fastest-rotting field in the file and it's the one thing the new workflow doesn't check. Release cadence over the last three months:

```
v0.3.136 2026-05-04  ...  v0.3.159 2026-07-21     (15 releases)
```

`v0.3.159` is correct today and will be wrong in a week or two. Schema-wise it's optional — `header.required` in v1.0.0 is `['expiration-date', 'project-url', 'schema-version']` — and `changelog:` (which you've added just below) already points at the releases page, which is self-updating.

Given the PR's stated goal, I'd drop `project-release` rather than pin a value that's guaranteed to be stale before the first scheduled run.

.github/workflows/security-insights.yaml
```diff
@@ -0,0 +1,107 @@
...
+        # Advisory only: this parses a Markdown table in another repo, which is
+        # too brittle to gate a merge on. It annotates the run so the weekly
+        # schedule surfaces drift instead of waiting for the expiry to lapse.
+        continue-on-error: true
+        run: |
+          curl -sSfL -o /tmp/MAINTAINERS.md \
+            https://raw.githubusercontent.com/kubescape/project-governance/main/MAINTAINERS.md
+          # Core maintainers are the rows above the "commiters" heading.
+          sed -n '1,/commiters/p' /tmp/MAINTAINERS.md \
+            | grep -oE '\[@[A-Za-z0-9_-]+\]' | tr -d '[]@' | sort -fu > /tmp/governance.txt
```
matthyx: **Blocker for what this step is for.** Both failure modes are reachable today, and I reproduced both.

**1. One upstream typo fix turns this into a permanent false positive.** The range key is the misspelling `commiters`, which is what `project-governance/MAINTAINERS.md` currently says. Correct it to `committers` and the range matches nothing, `sed` prints the whole file, and committers + emeritus get folded into the "governance" set:

```
$ sed 's/commiters/committers/' MAINTAINERS.md > fixed.md   # simulate the upstream fix
$ # ...same pipeline against fixed.md:
 amirmalka
-amitschendel
 Bezbran
-craigbox
-dwertent
 matthyx
```

So the step would warn about drift forever, naming exactly the three people this PR just correctly removed. `sed -n '1,/commiters/p'` → something anchored and spelling-tolerant, e.g. `sed -n '1,/^The following table lists the Kubescape project commit\?ers/p'`.

**2. A format change makes the step green *and silent* — the exact rot this workflow exists to prevent.** If the table stops using `[@handle](url)` link form, `grep -oE` matches nothing, `pipefail` (on by default in Actions' bash) kills the pipeline, and `continue-on-error: true` swallows it. No warning is emitted and the run is green:

```
$ # governance table rewritten without markdown links:
step exit=1 (continue-on-error swallows this; NO warning emitted)
```

Needs an explicit empty-parse branch that emits `::warning::` — otherwise the advisory check is indistinguishable from "no drift". This is the one I'd most want fixed before merge.

For what it's worth, against the current file the check does work — `governance == manifest`, the 5 names match, in sync. It's the failure modes, not the happy path.

---

Separately, and *not* a change I want: CodeRabbit's advice to rewrite the `yq` paths as `header["schema-version"]` / `project-lifecycle["core-maintainers"][]` is a false positive. Hyphenated keys parse fine as written:

```
$ yq --version
yq (https://github.com/mikefarah/yq/) version v4.53.3
$ yq -r '.header.schema-version' SECURITY-INSIGHTS.yml
1.0.0
$ yq -r '.project-lifecycle.core-maintainers[]' SECURITY-INSIGHTS.yml
github:amirmalka
github:Bezbran
...
```

Leave those alone.

.github/workflows/security-insights.yaml
```diff
@@ -0,0 +1,107 @@
...
+  pull_request:
+    paths:
+      - 'SECURITY-INSIGHTS.yml'
+      - '.github/workflows/security-insights.yaml'
+  schedule:
+    # Monthly, not weekly: the pull_request trigger above only fires when someone
+    # edits the manifest, and it went untouched from 2023 until this PR -- so a
+    # schedule is the only thing that ever runs the expiry check. Against the
+    # 60-day warning window below, monthly still fires twice before expiry.
+    - cron: '0 6 1 * *'
```
matthyx, line 20: **Blocker — nothing routes this failure to somebody who can act on it.**

The reasoning in the comment is right: `pull_request` can never catch expiry, because an expiring manifest is by definition one nobody is editing. So this cron is the only thing that ever runs the expiry gate, and a red scheduled run is the entire alerting mechanism.

The problem is who sees it. GitHub sends scheduled-workflow failure notifications to the user who last modified the workflow file — that's the fork contributor here, not the maintainers. A red run on `main` also doesn't surface in any PR check. Concretely: around 2027-05-30 this job starts failing every month, and the only person notified is someone without commit access.

That leaves the manifest rotting exactly as before, just with a failing job nobody reads. Options:

- have the expiry step open (or refresh) a tracking issue on failure — `gh issue create`, needs `issues: write` in `permissions`; this is the one I'd pick;
- or post to the maintainers' channel;
- or, at minimum, have a maintainer be the last committer to the cron so the notifications land somewhere useful — fragile, but better than nothing.

While you're in here: the header comment on line 7 says "impending expiry fails on a **weekly** schedule" and line 91 says "the **weekly** schedule", but the cron is monthly. Same on the two `curl` calls CodeRabbit noted — `--connect-timeout`/`--max-time` are worth adding since a stall in either external fetch just burns the job timeout. And the `days` truncation it flagged is real but cosmetic: a manifest that expired 6h ago yields `days=0`, skips the "expired" branch and lands in the `< 60` branch, so it still fails, just with the wrong message.

```
expired 6h ago -> days=0 ; days<0 ? no
```

matthyx (commented):
Went through this end-to-end and reproduced the claims locally rather than taking them on trust. The core thesis holds — I can confirm all of it:

```
main     -> project-lifecycle: Additional properties are not allowed ('stage' was unexpected)
            project-lifecycle: 'status' is a required property          exit=1
this PR  -> VALID                                                       exit=0
```

(validated against `ossf/security-insights@v1.0.0`, `Draft7Validator` + `FORMAT_CHECKER`, same as the workflow.) Also checked out: `v0.3.159` is the current release, `docs/` exists here, the four `project-governance` targets all resolve, the quay repo returns 200, and the 5-name `core-maintainers` list matches `MAINTAINERS.md` exactly. The `check-jsonschema` note is accurate. Good find on the `stage`/`status` break — that's been invalid since 2023 and nothing caught it.

**Blockers**, all inline:

1. `automated-tools-list` claims syft/grype are excluded from Dependabot. There is no `dependabot.yml` in this repo (or in `kubescape/.github`), so no such exclusion exists — and what's actually running is security updates, which cover those modules. Given this PR is about the file being untrue, that one needs to go or be made true.
2. `security-testing` overstates CodeQL and Snyk. Both are `continue-on-error: true` upstream, both are skipped by `pr-created.yaml`'s `paths-ignore`, and Snyk additionally never runs on fork PRs. They're advisory scans, not gates, and the comments should say so.
3. The drift step is silent on parse failure — `pipefail` + `continue-on-error` means a format change upstream gives you a green run with no warning, which is precisely the rot this workflow is supposed to catch. It's also keyed to the `commiters` misspelling upstream; fixing that typo turns it into a permanent false positive naming the three people you just removed.
4. Nothing routes the scheduled failure to a maintainer. The cron is the only path that ever evaluates expiry, and GitHub notifies the workflow file's last committer — a fork contributor. Worth having the step open a tracking issue instead.

None of these are structural; the shape of the PR is right. 1 and 2 are edits to free-text `comment` fields, 3 is a handful of lines in the drift step, 4 is a decision about where the alert should land.

On the two questions you raised:

- **`amitschendel`** — leave him out. Governance has him under committers, and the v1.0.0 field is `core-maintainers`; the list as it stands matches `MAINTAINERS.md` exactly, and having the manifest agree with governance is worth more than the broader reading. Revisit when we move to v2 and the field becomes `core-team`.
- **GitGuardian / go-licenses / Scorecard** — agreed, the `tool-type` enum has no value for secret- or license-scanning, so leaving them out is correct. Don't force them in.

On centralising into `project-governance`: your v2 `project-si-source` analysis matches what I read in the spec, and I agree with sequencing it after this. Please do open the issue there to scope it — the org-wide v1→v2 migration is the right frame for that conversation, and `kubescape/kubescape` carrying the byte-identical broken copy is a good argument for doing it properly rather than repo-by-repo.

I'll approve the workflow run so we can see it go green on CI.

.github/workflows/security-insights.yaml
1PoPTRoN: Fair - weekly was overkill. Switched to monthly in 878daea.

We do need *a* schedule though, or the expiry check never runs at all: `pull_request` only fires when someone edits the manifest or this workflow, and the manifest went untouched from 2023 until this PR. That is exactly how it drifted 656 days past expiry unnoticed.

Against the 60-day warning window monthly still fires twice before expiry (2027-06-01 and 2027-07-01 for the current date), and scheduled runs drop from 52/year to 12.

SECURITY-INSIGHTS.yml
1PoPTRoN: Went with your first option in e830ccc - the comment is now just "Dependabot opens dependency-bump PRs against main."

Confirmed there is no `dependabot.yml` in this repo or in `kubescape/.github`, so what runs is security updates over every module in `go.mod`, syft and grype included. The exclusion I described was not implemented anywhere, and you are right that stating it would have reintroduced the exact failure class this PR is meant to close.

Agreed the exclusion policy is a maintainer decision that has not actually been taken, so I have not added a `dependabot.yml` here.

SECURITY-INSIGHTS.yml
1PoPTRoN: Fixed in e830ccc, using your wording.

Confirmed both counts: all three `codeql-action` steps in `go-basic-tests.yaml` are `continue-on-error: true`, and the path filters mean a workflow-only PR gets no CodeQL run at all - including the workflow half of this PR.

SECURITY-INSIGHTS.yml
1PoPTRoN: Fixed in e830ccc.

It now reads "same-repository pull requests ... (advisory; does not block merge)" and states the fork case outright rather than framing it as a config caveat. Confirmed the step is gated on `needs.check-secret.outputs.run-snyk == 'true'` and is separately `continue-on-error: true`.

SECURITY-INSIGHTS.yml
1PoPTRoN: Dropped in e830ccc.

Agreed - pinning a value that is guaranteed to be stale before the first scheduled run is the failure mode this PR exists to close, and `changelog:` already points at the self-updating releases page.

.github/workflows/security-insights.yaml
1PoPTRoN: Both fixed in e830ccc, and thanks for reproducing them - I reproduced both before changing anything.

The range end is now anchored to the committers heading and tolerates either spelling, and there is an explicit empty-parse branch that emits `::warning::` instead of dying silently under pipefail. Behaviour across the shapes I could construct:

| upstream MAINTAINERS.md | before | after |
| --- | --- | --- |
| current (`commiters`) | in sync | in sync |
| typo fixed to `committers` | drift: craigbox, dwertent, amitschendel | in sync |
| markdown links removed | green and silent | explicit warning |
| heading renamed | drift (whole file matched) | explicit warning |

Two notes on the pattern itself, both of which cost me a wrong first attempt:

1. `commit\?ers` makes the *first* `t` optional, so it matches `commiters` but not `committers`. It would still have broken on the typo fix it was meant to survive. Used `committ?ers` instead.
2. `\?` is a GNU extension. On BSD sed it matches nothing, the range never closes, and sed prints the whole file - which is failure mode 1 again from a different cause. Switched to `sed -nE` with a plain `?`, which behaves the same on both.

The heading check is also a `grep -q` guard now, so a renamed heading warns rather than silently matching everything.

Left the `yq` paths as they were, agreed the hyphenated-key suggestion was a false positive.

.github/workflows/security-insights.yaml
1PoPTRoN: Implemented the tracking-issue option in e830ccc, since it was the one you would pick and it is the only one that does not depend on who last touched the file.

On scheduled runs only, a failing expiry check now opens an issue titled "SECURITY-INSIGHTS.yml needs review", or comments on the existing open one if there is one. It matches on title rather than a label so the step does not depend on a label existing in the repo. Added `issues: write` to `permissions` for this. It is skipped on `pull_request`, where the red check is already in front of someone who can act on it.

Also from your note:

- Corrected the stale "weekly" wording on lines 7 and 91.
- Added `--connect-timeout 10 --max-time 60` to both external fetches.
- Switched the expiry classification to epoch seconds. You were right that this is reachable: in bash a manifest expired 6h ago gives `days=0`, skips the expired branch and reports "expires in 0 days" for a file that is already invalid. Now:

```
expired 6h ago    days=0    was: "expires in 0 days"   now: EXPIRED
expired 40d ago   days=-40  was: EXPIRED               now: EXPIRED
expires in 30d    days=30   unchanged
```

matthyx (commented):
Re-checked the whole thing against `e830ccc`. All four blockers are cleared, and I verified each fix rather than reading the diff — including re-running the scenarios I used to raise them.

**1. Dependabot claim** — removed. Re-confirmed the underlying finding still holds: no `dependabot.yml` anywhere in the tree, and #384's body is a plain version-bump with no in-repo config behind it. Dropping the sentence was the right call.

**2. `security-testing` comments** — both now accurate. CodeQL reads as advisory with the path-filter caveat, Snyk names the same-repo restriction and the fork skip explicitly.

**3. Drift step** — this is the one I most wanted fixed, and it's properly fixed. All four paths behave:

```
current upstream ("commiters")      -> IN SYNC
typo fixed upstream ("committers")  -> IN SYNC          <- was a permanent false positive
link format changed                 -> warning, exit 0  <- was green and silent
heading removed entirely            -> warning, exit 0
```

`committ?ers` covers both spellings, and the `|| true` plus `-s` check is the right way to separate "parsed nothing" from "no drift" under `pipefail`. Nice.

**4. Tracking issue** — works. I checked the dedupe specifically, since a title full of dots and hyphens is where `in:title` phrase matching tends to fall over:

```
in:title "strings.HasPrefix"                    -> 1
in:title "conflict-safe retry"                  -> 1
--state open + --search                         -> combines correctly (open hit, closed control 0)
```

So the second month won't open a duplicate. The step's `body` also renders without the stray indentation you'd normally get building a multi-line string inside a block scalar, and the escaped backticks survive. `gh` picks up the repo from the checkout remote, which `persist-credentials: false` doesn't affect.

Also confirmed: `project-release` gone, both `curl` calls bounded, weekly/monthly wording consistent, manifest still validates clean against v1.0.0 with `FORMAT_CHECKER`, and the day-truncation fix does what it says —

```
expired 6h ago  -> EXPIRED on ... (days=0)     <- previously reported as "expires in 0 days"
expired 30d ago -> EXPIRED
expires in 30d  -> expires in 29 days
as written      -> pass (361 days)
```

Two residuals inline, both minor — one real (`if: always()` on the drift step, one line), one a judgement call you can decline. Neither blocks merge from my side.

One thing outstanding that isn't yours to fix: this workflow has still never actually executed. All five runs sit at `action_required` because the PR is from a fork, so everything above is my local reproduction rather than CI evidence. I said last time I'd get that approved and didn't — sorry for the round-trip. Chasing it now, and once it's green I'm happy to merge.

.github/workflows/security-insights.yaml
```diff
@@ -0,0 +1,159 @@
...
+          else
+            echo "opening a new tracking issue"
+            gh issue create --title "${TITLE}" --body "${body}"
+          fi
+
+      - name: Report core-maintainer drift from project-governance
+        # Advisory only: this parses a Markdown table in another repo, which is
+        # too brittle to gate a merge on. It annotates the run so the monthly
+        # schedule surfaces drift instead of waiting for the expiry to lapse.
+        continue-on-error: true
```
matthyx, line 130: Minor, and the only thing I'd still change: this step is skipped on exactly the runs that matter most.

It has `continue-on-error: true` but no `if:`. That flag only stops *this* step's failure from failing the job — it doesn't make the step run once the job is already in a failure state. The expiry step above exits 1, so every step after it without an `if:` is skipped:

```
- 'Fail if the manifest is expired or expires within 60 days'  if=None      continue-on-error=None
- 'Open or refresh the expiry tracking issue'                  if=always()  continue-on-error=None
- 'Report core-maintainer drift from project-governance'       if=None      continue-on-error=True
```

So on the monthly run where the manifest has lapsed — the run most likely to *also* have stale maintainers, since both rot for the same reason — the drift check never executes. Drift only ever gets reported when the manifest is otherwise healthy.

`if: always()` on this step, same as you did on the tracking-issue step, and it's covered.

.github/workflows/security-insights.yaml
```diff
@@ -0,0 +1,159 @@
...
+          message="${message} Re-review it against https://github.com/kubescape/project-governance/blob/main/MAINTAINERS.md and refresh header.last-reviewed / header.expiration-date."
+          echo "message=${message}" >> "$GITHUB_OUTPUT"
+          echo "::error file=SECURITY-INSIGHTS.yml::${message}"
+          exit 1
+
+      - name: Open or refresh the expiry tracking issue
+        # Only on scheduled runs: on a pull_request the red check is already in
+        # front of someone who can act on it. On a schedule nothing else routes
+        # this to a maintainer, so raise an issue they will actually see.
+        if: ${{ always() && github.event_name == 'schedule' && steps.expiry.outcome == 'failure' }}
```
matthyx, line 105: This is the right condition for the expiry case and the `always()` is correctly placed. One residual, noting it rather than asking for it:

gating on `steps.expiry.outcome` means a **schema** failure on a scheduled run raises no issue. If the validate step fails, the expiry step never runs, its outcome is `skipped` rather than `failure`, and this step is skipped too — leaving a red scheduled run whose only notification goes to whoever last touched this file. That's the same gap the tracking issue was added to close, just reached by a different door.

Low likelihood, since the schema URL is tag-pinned and the manifest can't change on `main` without passing the PR check. But if you want it airtight, give the validate step an `id` and OR the two outcomes, or drop to `failure() && github.event_name == 'schedule'` and let the message fall back when `steps.expiry.outputs.message` is empty.

Your call — I'm happy to merge without it.

PR #392, opened 2026-07-29
fix: report TooLarge status instead of Incomplete when image size lim…

### LINKED ISSUE #391: Bug: Inconsistent status returned when container image exceeds maximum size limit
# Description
When scanning a container image that exceeds the maximum size limit (`MaxImageSize` or `maxImageSize`), kubevuln incorrectly classifies the resulting error as `Incomplete` instead of using the dedicated `TooLarge` status. 

This causes other system components to treat the scan failure as a transient error and queue unnecessary retries.

# Environment
OS: `Kubernetes Cluster (Any OS)`
Version: `main branch / dev`
 
# Steps To Reproduce
1. Configure `kubevuln` with a strict `maxImageSize` limit.
2. Trigger a scan on an image that is larger than the configured limit.
3. Observe the generated SBOM/vulnerability report status in the logs or in-cluster CRDs.

# Expected behavior
When an image exceeds the size limit, the adapter should return a status of `TooLarge` (using the `helpersv1.TooLarge` constant), indicating that the scan failed due to size constraints.

# Actual Behavior
The adapter sets the scan status to `Incomplete` (using `helpersv1.Incomplete`), which is the same status used for general runtime scan failures.

# Additional context
The code path is located in:
* **Grype/Syft Adapter:** [adapters/v1/syft.go#L229](https://github.com/kubescape/kubevuln/blob/main/adapters/v1/syft.go#L229)
```go
	case err != nil && strings.Contains(err.Error(), image.ErrImageTooLarge.Error()):
		logger.L().Ctx(ctx).Warning("Image exceeds size limit",
			helpers.Int("maxImageSize", int(s.maxImageSize)),
		domainSBOM.Status = helpersv1.Incomplete // <-- Should be helpersv1.TooLarge
		return domainSBOM, nil

### PR DESCRIPTION
**Pull request by singhshresth26**
## Overview
When a container image size exceeds the maximum configured limit (`MaxImageSize` / `maxImageSize`), `kubevuln` was incorrectly classifying the failure as `Incomplete` instead of `TooLarge`. This causes other components to treat the failure as transient and repeatedly queue retries, leading to wasted resources.

This PR addresses the issue by:
1. Returning `helpersv1.TooLarge` instead of `helpersv1.Incomplete` when `image.ErrImageTooLarge` is encountered during SBOM generation in both the Syft adapter (`adapters/v1/syft.go`) and the gRPC server (`pkg/sbomscanner/v1/server.go`).
2. Down-grading a `logger.L().Ctx(ctx).Fatal` call to `Warning` inside the `src.Close()` defer block in `syft.go`. A failure to remove a temporary directory (commonly triggered by Windows file locking or antivirus in local environments) should not crash the entire `kubevuln` daemon.
3. Updating the test suite in `syft_test.go` to assert the `TooLarge` status when testing the image size limits.

## Additional Information
We also downgraded a `Fatal` logging level to a `Warning` log level in `syft.go`'s `src.Close()` cleanup defer, as file locking on Windows/development environments caused the unit test run (and would cause the production daemon) to crash completely rather than continuing gracefully.

## Related issues/PRs:
* Resolves #391

## Summary by CodeRabbit

* **Bug Fixes**
  * Oversized images and SBOMs now receive clear “too large” statuses with applicable size limits.
  * Scanner memory exhaustion is reported with a dedicated reason and configured memory limit.
  * Cached SBOMs are regenerated when limits or failure details are missing or outdated.
  * Improved handling of image download failures and source-close warnings.

* **Improvements**
  * Scan results now preserve detailed status reasons for better troubleshooting.
  * SBOM cleanup removes both standard and filtered results when requested.
  * Outdated SBOM data is handled more reliably during scans.

### CODE REVIEW

File: adapters/v1/syft.go [hunk 1]
```diff
@@ -226,7 +226,7 @@ func (s *SyftAdapter) CreateSBOM(ctx context.Context, name, imageID, imageTag st
 		logger.L().Ctx(ctx).Warning("Image exceeds size limit",
 			helpers.Int("maxImageSize", int(s.maxImageSize)),
 			helpers.String("imageID", imageID))
-		domainSBOM.Status = helpersv1.Incomplete
+		domainSBOM.Status = helpersv1.TooLarge
```
matthyx, line 234: 🔴 **Blocker — this pins the SBOM permanently.**

`kubescape/storage` v0.0.258 (`pkg/registry/file/storage.go:731-737`) makes `GuaranteedUpdate` a no-op for any object already annotated `TooLarge`:

```go
if annotations != nil && annotations[helpersv1.StatusMetadataKey] == helpersv1.TooLarge {
    logger.L().Debug("GuaranteedUpdate - already too large object, skipping update", ...)
    v.Set(reflect.ValueOf(origState.obj).Elem())
    return nil  // silently succeeds, writes nothing
}
```

`StoreSBOM` writes Create → on `AlreadyExists` → Update (`repositories/apiserver.go:1141-1156`), so the first store here freezes the `SBOMSyft` forever — and logs `"updated SBOM in storage"` on every later attempt because `retryErr` is `nil`.

Unlike `maxSBOMSize` (line 301, already `TooLarge`, and there the SBOM really was generated), `maxImageSize` is a knob operators turn to *remediate*. Raise it from 2Gi to 10Gi and the previously-failed image still never gets scanned: `GetSBOM` returns the frozen object with non-nil `Content` (it's always `&manifest.Spec.Syft`), so `scan.go:432` skips regeneration and `scan.go:472` returns `ErrIncompleteSBOM`. Even a syft version bump doesn't help — `GetSBOM` discards the stale manifest, kubevuln regenerates, and the write is silently dropped, so every subsequent scan redoes the full pull and fails to persist.

With `Incomplete` there is no short-circuit and the image recovers on the next version bump.

node-agent hit this and guards against it — see the comment at `pkg/sbommanager/v1/sbom_manager.go:378-386`, and the `ScannerMemoryLimitAnnotation` escape hatch at `:402-412` that re-opens the scan when the limit that caused the verdict changes.

Suggest recording the `maxImageSize` that produced this verdict as an annotation and re-scanning when the configured value differs — or confirming with maintainers that a config-change-proof permanent pin is intended.

matthyx, line 234: 🔴 **Blocker (knock-on) — the reason code reported to users becomes wrong.**

`classifySBOMStatus` (`core/services/scan_failure_reasons.go:65-72`) maps `TooLarge` → `scanfailure.ReasonSBOMTooLarge`, whose customer-facing text is *"Generated software inventory (SBOM) exceeds the maximum size limit"* (`armoapi-go/scanfailure/types.go:44`).

At this point the image was never even downloaded, so that message is affirmatively false. Before this change the reason was `ReasonSBOMIncomplete` — *"the scan may have timed out **or the image exceeded size limits**"* — which was at least accurate.

`scanfailure.ReasonImageTooLarge` (*"Image exceeds the maximum size limit for vulnerability scanning"*) exists for exactly this, and node-agent already emits it here (`pkg/sbommanager/v1/sbom_manager.go:570`).

Setting a marker annotation in this branch and extending `classifySBOMStatusWithAnnotation` to return `ReasonImageTooLarge` would follow the existing `"scanner OOM"` disambiguation pattern. Use a dedicated key rather than `helpersv1.StatusMetadataKey` — `StoreSBOM` overwrites that one with the status (`repositories/apiserver.go:1104`).

adapters/v1/syft.go
```diff
@@ -247,7 +247,7 @@ func (s *SyftAdapter) CreateSBOM(ctx context.Context, name, imageID, imageTag st
 		// make sure we clean the temp dir
 		defer func(src source.Source) {
 			if err := src.Close(); err != nil {
-				logger.L().Ctx(ctx).Fatal("failed to close source", helpers.Error(err),
+				logger.L().Ctx(ctx).Warning("failed to close source", helpers.Error(err),
```
matthyx, line 257: ✅ Good catch, and worth keeping — `logger.L().Ctx(ctx).Fatal` reaches `zap.Logger.Fatal`, which calls `os.Exit(1)`, so a temp-dir close failure really did take down the whole daemon. The temp dir is still cleaned by the outer `t.Cleanup()` defer at lines 180-187, so downgrading leaks nothing, and it makes this consistent with `pkg/sbomscanner/v1/server.go:145`, which already used `Warning`.

Non-blocking: this is unrelated to #391. Worth splitting out or at least calling out separately so it isn't buried under a "status rename" changelog entry.

pkg/sbomscanner/v1/server.go
```diff
@@ -121,7 +121,7 @@ func (s *scannerServer) CreateSBOM(ctx context.Context, req *pb.CreateSBOMReques
 			helpers.Int("maxImageSize", int(req.MaxImageSize)),
 			helpers.String("imageID", imageID))
 		return &pb.CreateSBOMResponse{
-			Status: helpersv1.Incomplete,
+			Status: helpersv1.TooLarge,
```
matthyx: 🟡 The gRPC change is right in spirit but note the ambiguity it creates downstream: `adapters/v1/sidecar.go:103` copies `result.Status` verbatim into the domain SBOM, and the sidecar adapter already sets `TooLarge` itself for crash-exhausted images (`sidecar.go:139`), disambiguated only by substring-matching a `"scanner OOM"` annotation (`scan_failure_reasons.go:78-84`).

With this change `TooLarge` means three different things — image too large, SBOM too large, scanner OOM-killed — behind one string tiebreaker. It works today; please at least document it here.

Also: `pkg/sbomscanner/v1/server_test.go` doesn't cover this branch. The updated `syft_test.go` case pulls a real image from Docker Hub, so it may not run in CI — a table entry for the gRPC path would actually lock this behaviour in.

matthyx (commented):
## Review

The direction is right — `Incomplete` genuinely is the wrong bucket for a deterministic size failure, and the `Fatal` → `Warning` downgrade fixes a real daemon-kill bug (`logger.L().Ctx(ctx).Fatal` bottoms out in `zap.Logger.Fatal`, which calls `os.Exit(1)`; the temp dir is already cleaned by the outer `t.Cleanup()` defer at `adapters/v1/syft.go:180-187`, so nothing leaks by downgrading).

But the one-line status swap has consequences beyond the switch statement. Two blockers below, then some non-blocking notes.

---

### 🔴 Blocker 1 — `TooLarge` is a one-way door in the storage layer

`kubescape/storage` (v0.0.258, the version pinned in `go.mod`) short-circuits **every** update to an object already annotated `TooLarge`:

```go
// pkg/registry/file/storage.go:731-737 — GuaranteedUpdate
annotations := origState.obj.(metav1.Object).GetAnnotations()
if annotations != nil && annotations[helpersv1.StatusMetadataKey] == helpersv1.TooLarge {
    logger.L().Debug("GuaranteedUpdate - already too large object, skipping update", ...)
    v.Set(reflect.ValueOf(origState.obj).Elem())
    return nil   // silently succeeds, writes nothing
}
```

`APIServerStore.StoreSBOM` persists via Create → on `AlreadyExists` → Update (`repositories/apiserver.go:1141-1156`). So the first time an oversized image is stored under this PR, its `SBOMSyft` is frozen permanently — and the update path logs `"updated SBOM in storage"` because `retryErr` is `nil`. A false success.

Concretely:

1. `maxImageSize: 2Gi`, image A is 5 GB → `SBOMSyft` pinned `TooLarge`.
2. Operator raises `maxImageSize` to `10Gi`.
3. `GetSBOM` returns the frozen object with non-nil `Content` (it's always `&manifest.Spec.Syft`, `apiserver.go:1060`), so `scan.go:432`'s `sbom.Content == nil` gate skips regeneration → status check at `scan.go:472` → `ErrIncompleteSBOM`.
4. Image A is **never scanned again**, at any limit.

Even the syft-version-bump escape hatch doesn't recover it: `GetSBOM` discards the manifest on an outdated tool version (`apiserver.go:1049`), kubevuln re-generates the SBOM and re-stores it — and that write is silently dropped, leaving the object `TooLarge` at the old version. Every subsequent scan then redoes the full pull + catalog and fails to persist, forever.

With `Incomplete` there is no short-circuit, so today the image recovers on the next scanner-version bump.

This is not hypothetical — node-agent hit exactly this and documented it:

> `pkg/sbommanager/v1/sbom_manager.go:378-386`
> "unlike `Incomplete`, `TooLarge` is a one-way door in the storage layer — `GuaranteedUpdate` silently drops every future write once status=too-large is set … `Incomplete` has no such short-circuit and stays safely retryable, so it's used instead."

node-agent pairs its `TooLarge` pin with a `ScannerMemoryLimitAnnotation` and re-opens the scan when the recorded limit differs from the current one (`sbom_manager.go:402-412`). kubevuln has no equivalent for `maxImageSize`.

**Ask:** either record the `maxImageSize` that produced the verdict as an annotation and re-scan when the configured limit changes (mirroring node-agent), or confirm with the maintainers that a permanent, config-change-proof pin is the intended semantic. As written, raising `maxImageSize` silently stops working as a remediation.

---

### 🔴 Blocker 2 — the user-facing failure reason regresses

`classifySBOMStatus` maps the status straight through:

```go
// core/services/scan_failure_reasons.go:65-72
case helpersv1.TooLarge:
    return scanfailure.ReasonSBOMTooLarge
```

`ReasonSBOMTooLarge`'s customer-visible text is *"Generated software inventory (SBOM) exceeds the maximum size limit"* (`armoapi-go/scanfailure/types.go:44`). After this PR an image that was **never even downloaded** reports that to the user.

`scanfailure.ReasonImageTooLarge` — *"Image exceeds the maximum size limit for vulnerability scanning"* — already exists for exactly this case, and node-agent already emits it for the same error (`sbom_manager.go:570`).

So the PR fixes the status but regresses the reason: `ReasonSBOMIncomplete` at least said *"the scan may have timed out or the image exceeded size limits"*; `ReasonSBOMTooLarge` is now affirmatively wrong.

**Ask:** carry the distinction through. Setting a marker annotation in the too-large branch and extending `classifySBOMStatusWithAnnotation` to return `ReasonImageTooLarge` would match the existing `"scanner OOM"` disambiguation pattern (use a dedicated key, not `StatusMetadataKey` — `StoreSBOM:1104` overwrites that one with the status).

---

### Non-blocking notes

- **`TooLarge` now carries three meanings.** `adapters/v1/sidecar.go:103` copies `result.Status` verbatim into the domain SBOM, and the sidecar itself already sets `TooLarge` for crash-exhausted images (`sidecar.go:139`), disambiguated only by string-sniffing a `"scanner OOM"` annotation. Adding image-too-large makes three distinct conditions behind one status with a substring tiebreaker. It works today, but it's worth a comment at minimum.
- **No coverage for the `server.go` change.** `pkg/sbomscanner/v1/server_test.go` doesn't exercise the image-too-large branch, and the updated `syft_test.go` case pulls a real image from Docker Hub, so it may not run in CI. A table entry for the gRPC path would lock this in.
- **Prefer `errors.Is`.** `strings.Contains(err.Error(), image.ErrImageTooLarge.Error())` is pre-existing, not introduced here — but node-agent uses `errors.Is(srcErr, syftutil.ErrImageTooLarge)` and it's more robust.
- **Scope.** The `Fatal` → `Warning` change is a good fix and unrelated to #391 — it also aligns `syft.go` with `server.go:145`, which already used `Warning`. Consider calling it out separately so it isn't lost in a "status rename" changelog entry.

---

### Does node-agent need the same fix?

Checked both paths in `kubescape/node-agent`:

- **In-process path — already correct.** `pkg/sbommanager/v1/sbom_manager.go:562-570` returns `helpersv1.TooLarge` + `scanfailure.ReasonImageTooLarge`, guarded so a content-bearing SBOM never reaches the one-way door. No change needed.
- **Sidecar path — has the same bug as #391.** node-agent's scanner server returns a gRPC `codes.FailedPrecondition` for `ErrImageTooLarge` (`pkg/sbomscanner/v1/server.go:57-61`), but the client only maps `Unavailable`/`Aborted` to a sentinel (`client.go:88-94`) and passes everything else through raw. `sbom_manager.go:501-522` therefore falls into the generic branch: `handleGenericFailure` retries this **deterministic** failure up to `maxScanRetries` (3) before pinning `Incomplete`, and reports `ReasonSBOMGenerationFailed`. `codes.FailedPrecondition` is never consumed anywhere outside tests.

  Fix there: map `FailedPrecondition` to an `ErrImageTooLarge` sentinel in the client, and handle it in `sbom_manager` with the same `wipSbomHadContent` guard the in-process path uses.

---

*Verified locally on `pr392`: `go build ./adapters/... ./pkg/sbomscanner/... ./core/...` passes; `go vet` is clean apart from pre-existing findings in `adapters/v1/backend_test.go`; `go test ./core/services/...` passes.*

singhshresth26:
Hi @matthyx,

Thanks for the feedback. To address both blocker issues and the other notes, here is the technical implementation I have designed. Please let us know if this aligns with your expectations:

### 1. Blocker 1 (TooLarge Storage Recovery on Configuration Changes)
Since `kubernetes/storage` ignores updates to any object with a `TooLarge` annotation, we need to bypass this one-way door by deleting the outdated object first when the limit changes.

* We will extend `ports.SBOMCreator` with a `GetMaxImageSize() int64` method.
* We will extend `ports.SBOMRepository` with a `DeleteSBOM(ctx, name)` method.
* When saving a `TooLarge` status due to the image limit, the adapter will set a custom annotation mapping the limit:
  - `kubescape.io/status-reason: "image-too-large"`
  - `kubescape.io/max-image-size: "<limit-value>"`
* In `ScanService`, we will introduce a wrapper helper method `getSBOM` to replace direct `GetSBOM` queries. 
* Inside `getSBOM`, if a retrieved SBOM has a `TooLarge` status with reason `image-too-large`, we will compare its stored limit annotation with `sbomCreator.GetMaxImageSize()`. 
* If the configured limit has increased, we will issue an explicit `DeleteSBOM` call to remove the old SBOM (both unfiltered and filtered) from the storage API server, and return an empty result to trigger a fresh scan.

### 2. Blocker 2 (User-Facing Failure Reason Disambiguation)
* Because `StoreSBOM` overwrites the `StatusMetadataKey` with the status name (`TooLarge`), we will use the custom annotation `kubescape.io/status-reason: "image-too-large"` (established in Blocker 1).
* We will update `classifySBOMStatusWithAnnotation` inside `core/services/scan_failure_reasons.go` to inspect this custom annotation. If present and matching, we will return `scanfailure.ReasonImageTooLarge` instead of mapping it directly to `ReasonSBOMTooLarge`.

### 3. Non-Blocking Notes
* **Avoid String Matching:** We will update the logic in `server.go` and `syft.go` to use `errors.Is(err, image.ErrImageTooLarge)` instead of string parsing.
* **Sidecar Error Mapping:** We will update `CreateSBOMResponse` in the gRPC server (`server.go`) to populate the `ErrorMessage` field when it returns `TooLarge` due to the image limit, so the sidecar adapter can read it and apply the correct annotations.
* **Test Coverage:** We will write a dedicated unit test in `server_test.go` to cover the gRPC server-side logic when `max_image_size` is exceeded, avoiding reliance on external image downloads in CI.
* **Scope Cleanliness:** We will revert/split the `Fatal -> Warning` change on `Close()` in `syft.go` out into a separate commit/PR to keep this PR's changes focused purely on the status rename and recovery flow.

Let me know if you have any feedback or suggestions on this.

matthyx:
Thanks, this is a good plan — the delete-then-rescan approach for Blocker 1 is the right call, and I confirmed it works: `StorageImpl.Delete` (`pkg/registry/file/storage.go:315-326`) has no `TooLarge` short-circuit; only `GuaranteedUpdate` (`:731-737`) does. So the escape hatch is real.

Four things before you start:

**1. The sidecar `ErrorMessage` mapping will break the flow — please don't do that one as written.**

`adapters/v1/sidecar.go:110-112`:

```go
if result.ErrorMessage != "" && result.Status != helpersv1.Learning {
    return domainSBOM, fmt.Errorf("%s", result.ErrorMessage)
}
```

Populating `ErrorMessage` alongside `Status: TooLarge` makes `SidecarSBOMAdapter.CreateSBOM` return a **non-nil error**. `scan.go:433` then takes the error branch, `classifySBOMError` falls through to `ReasonSBOMGenerationFailed` (none of its string matches hit), and it returns `fmt.Errorf("creating SBOM: %w", err)` — so `StoreSBOM` is never reached and the SBOM gets **no terminal status at all** on the sidecar path. That's worse than what we have today.

Add a dedicated field to `CreateSBOMResponse` instead — `status_reason` (or a small enum) — and have the sidecar adapter translate that into the annotation. Overloading `ErrorMessage` collides with existing control flow.

**2. `errors.Is` works today, but please pin it with a test.**

I ran it against a real oversized pull: for `syft.GetSource(..., WithSources("registry"))`, `errors.Is(err, image.ErrImageTooLarge)` is `true`. So the swap is safe.

The caveat is that syft's `sourceError` (`syft/get_source.go:76-89`) only uses `%w` when **exactly one** provider errored — with two or more it falls back to `fmt.Sprintf("%s")` and the chain breaks silently. `WithSources("registry")` can select more than one provider (the stereoscope registry provider plus `ocimodelsource`). If that ever happens, `errors.Is` returns `false`, detection silently disappears, and every oversized image quietly becomes a generic scan failure.

So: add a regression test asserting `errors.Is` on the real error, or keep `|| strings.Contains(...)` as a fallback. Also worth knowing that `image.ErrImageTooLarge` comes from the `matthyx/stereoscope` fork pinned via the `replace` directive in `go.mod`, not upstream stereoscope.

**3. Consider making the recovery helper generic — `TooLarge` has two other producers with the same freeze.**

`maxSBOMSize` (`adapters/v1/syft.go:301`) and the sidecar OOM-exhaustion path (`adapters/v1/sidecar.go:139`) also write `TooLarge`, and both have operator-tunable remediation knobs (`maxSBOMSize`, the scanner memory limit). Same one-way door, same "raise the limit and nothing happens" symptom.

If `getSBOM` is written as *"record the limit that produced this verdict; delete when the current limit differs"* rather than special-casing image size, all three are covered for roughly the same effort. node-agent already does exactly this for the memory limit (`ScannerMemoryLimitAnnotation`, `pkg/sbommanager/v1/sbom_manager.go:402-412`), so mirroring that shape keeps the two repos consistent.

**4. Handle SBOMs that are already frozen.**

Compare limits with `!=` rather than "increased", and decide explicitly what happens when the `max-image-size` annotation is **absent** — that's every SBOM pinned before this ships. I'd treat `TooLarge` + missing annotation as stale and delete it once; otherwise the fix isn't retroactive and existing frozen SBOMs stay frozen forever.

**Smaller notes:**

- `getSBOM` needs to cover all three call sites: `scan.go:106`, `:211`, `:425`.
- Deleting the filtered SBOM too is right — just make `DeleteSBOM` tolerate `NotFound` on either object independently.
- Annotation constants: node-agent defines `kubescape.io/scanner-memory-limit` locally, so local constants are fine precedent. If node-agent should ever read these, they belong in `k8s-interface/instanceidhandler/v1/helpers/keys.go` next to `StatusMetadataKey`.
- On scope cleanliness: please **keep** the `Fatal` → `Warning` fix, just land it as its own PR rather than dropping it. It's a genuine `os.Exit(1)` in the daemon and the temp dir is already cleaned by the outer `t.Cleanup()` defer, so it's a safe standalone change.

The node-agent sidecar gap I mentioned is separate from this PR — I'll track it on its own.

adapters/v1/sidecar.go
```diff
@@ -137,9 +148,11 @@ func (s *SidecarSBOMAdapter) handleCrash(ctx context.Context, name, imageID, ima
 		s.mu.Unlock()
 
 		domainSBOM.Status = helpersv1.TooLarge
+		domainSBOM.Annotations[domain.StatusReasonAnnotationKey] = domain.ReasonScannerOOM
 		if s.memoryLimit != "" {
```
matthyx, line 153: 🔴 **This is a regression when `SCANNER_MEMORY_LIMIT` is unset.**

`domain.StatusReasonAnnotationKey = domain.ReasonScannerOOM` is written unconditionally on line 151, but `ScannerMemoryLimitAnnotationKey` only lands inside this `if`. `memoryLimit` is `os.Getenv("SCANNER_MEMORY_LIMIT")` (`cmd/http/main.go:81`), so it's `""` in any deployment that doesn't set it.

That produces an unbounded loop:

1. Sidecar OOMs `maxCrashRetries` times → `TooLarge` + `status-reason=scanner-oom`, **no** limit annotation → stored.
2. Next scan → `getSBOM` → `case domain.ReasonScannerOOM` → `limitStr, ok := …` → `!ok` → `stale = true` → delete → return empty → regenerate.
3. Sidecar OOMs `maxCrashRetries` times again → step 1.

Every scan cycle burns `maxCrashRetries` OOM-killing sidecar restarts on the same image, forever. Pre-PR the SBOM just stayed frozen — bad, but cheap.

Fix: write the limit annotation unconditionally so the comparison against `GetMemoryLimit()` is well-defined on the second pass:

```go
domainSBOM.Annotations[domain.StatusReasonAnnotationKey] = domain.ReasonScannerOOM
domainSBOM.Annotations[domain.ScannerMemoryLimitAnnotationKey] = s.memoryLimit
if s.memoryLimit != "" {
    domainSBOM.Annotations[helpersv1.StatusMetadataKey] = fmt.Sprintf(
        "scanner OOM after %d retries (memory limit: %s)", maxCrashRetries, s.memoryLimit)
}
```

`max-image-size` and `max-sbom-size` are always written, so the other two reasons don't have this problem.

core/services/scan.go
```diff
@@ -928,3 +928,69 @@ func (s *ScanService) ValidateScanRegistry(ctx context.Context, workload domain.
 func (s *ScanService) Version() string {
 	return s.sbomCreator.Version() + "-" + s.cveScanner.Version()
 }
+
+func (s *ScanService) getSBOM(ctx context.Context, name string, creatorVersion string) (domain.SBOM, error) {
+	sbom, err := s.sbomRepository.GetSBOM(ctx, name, creatorVersion)
+	if err != nil {
+		return sbom, err
+	}
+	if sbom.Content == nil {
```
matthyx, line 963: 🟡 **This early return is why the cleanup can't reach version-bumped SBOMs.**

`GetSBOM` discards the manifest and returns an empty `domain.SBOM` whenever the stored tool version is older than the current one (`repositories/apiserver.go:1091-1097`). That lands here with `Content == nil`, so the staleness check below never runs for exactly the objects that most need purging.

After any syft bump, every pre-existing `TooLarge` SBOM becomes permanently uncleanable: `GetSBOM` hides it → `getSBOM` can't delete it → kubevuln regenerates → `StoreSBOM` Updates → `GuaranteedUpdate` silently drops the write and returns `nil`, so `retryErr` is nil and it logs `"updated SBOM in storage"`. The object stays pinned at the old version and every scan cycle redoes the full pull + catalog for nothing.

Less severe than the original blocker — the in-memory scan still completes so CVE results are produced — but it means the limit-change recovery only works if the limit changes *before* the next scanner-version bump.

Worth a version-agnostic read for the purge decision (a metadata-only `Get` here, or having `GetSBOM` distinguish "found but outdated" from "not found"). Happy to track this separately if you'd rather not grow the PR.

pkg/sbomscanner/v1/server.go
```diff
@@ -116,12 +116,13 @@ func (s *scannerServer) CreateSBOM(ctx context.Context, req *pb.CreateSBOMReques
...
 	switch {
-	case err != nil && strings.Contains(err.Error(), image.ErrImageTooLarge.Error()):
+	case err != nil && (errors.Is(err, image.ErrImageTooLarge) || strings.Contains(err.Error(), image.ErrImageTooLarge.Error())):
 		logger.L().Warning("Image exceeds size limit",
 			helpers.Int("maxImageSize", int(req.MaxImageSize)),
 			helpers.String("imageID", imageID))
 		return &pb.CreateSBOMResponse{
-			Status: helpersv1.Incomplete,
+			Status:       helpersv1.TooLarge,
+			StatusReason: "image-too-large",
```
matthyx: 🟡 Please use `domain.ReasonImageTooLarge` here (and `domain.ReasonSBOMTooLarge` on line 201) rather than the string literals.

`adapters/v1/sidecar.go:107-113` compares `result.StatusReason` against those constants, so the two sides have to agree exactly. Same module and there's no import cycle — `core/domain` only pulls in `storage/…/v1beta1`. A typo on either side fails silently: the annotation is simply never written, and the recovery path degrades to the `default:` branch.

matthyx (commented):
## Re-review

This is a big improvement — both blockers are genuinely addressed and the new gRPC test with a local mock registry (`TestCreateSBOM_ImageTooLarge_LocalRegistry`) is exactly the right shape: no network dependency, asserts `Status`, `StatusReason` and the empty `ErrorMessage` together.

Confirmed fixed:

- **Blocker 2** — `classifySBOMStatusWithAnnotation` now returns `ReasonImageTooLarge` / `ReasonSBOMTooLarge` / `ReasonScannerOOMKilled` off `kubescape.io/status-reason`, with the legacy `"scanner OOM"` substring kept as a fallback. Correct, and covered by `scan_failure_reasons_test.go`.
- **The sidecar `ErrorMessage` trap** — dodged two ways: the guard now excludes `TooLarge`, and the server doesn't populate `ErrorMessage` on that path at all. Good.
- **Generalising to all three producers** — `image-too-large` / `sbom-too-large` / `scanner-oom` each record the limit that produced the verdict. This is the right shape and it's more than I asked for.

Verified locally on `3279c47`: `go build ./...` passes, `go vet` is clean apart from the pre-existing `adapters/v1/backend_test.go` findings, and `./core/services/...`, `./repositories/...`, `./pkg/sbomscanner/...` all pass.

Two things left — one is a real regression, the other is a gap the current design can't reach.

---

### 🔴 `scanner-oom` + unset `SCANNER_MEMORY_LIMIT` = unbounded delete/rescan loop

`handleCrash` writes the reason annotation unconditionally but the limit annotation only inside `if s.memoryLimit != ""` (`adapters/v1/sidecar.go:151-156`). And `memoryLimit` comes straight from `os.Getenv("SCANNER_MEMORY_LIMIT")` (`cmd/http/main.go:81`) — unset in any deployment that doesn't explicitly set it.

So with the env var unset:

1. Sidecar OOMs `maxCrashRetries` times → `Status=TooLarge`, `status-reason=scanner-oom`, **no** `max-…`/`scanner-memory-limit` annotation → stored.
2. Next scan → `getSBOM` → `case domain.ReasonScannerOOM` → annotation missing → `stale = true` → **delete** → return empty → regenerate.
3. Sidecar OOMs `maxCrashRetries` times again → back to step 1.

That repeats on every scan cycle, forever, each iteration burning `maxCrashRetries` OOM-killing sidecar restarts on the same image. Before this PR the SBOM just stayed frozen — bad, but cheap. This makes it expensive.

Simplest fix: write `ScannerMemoryLimitAnnotationKey` unconditionally (empty string included) so the comparison against `GetMemoryLimit()` is always well-defined and matches on the second pass. The `if s.memoryLimit != ""` guard should stay only around the human-readable `StatusMetadataKey` message.

Worth double-checking the same shape for the other two reasons — `max-image-size` and `max-sbom-size` are always written, so they're fine.

### 🟡 Version-bumped `TooLarge` SBOMs are invisible to the cleanup

`GetSBOM` discards the manifest before returning whenever the stored tool version is older than the current one (`repositories/apiserver.go:1091-1097`), returning an empty `domain.SBOM`. `getSBOM` then hits `if sbom.Content == nil { return sbom, nil }` (`core/services/scan.go:937`) and the staleness check never runs.

So after any syft bump, every pre-existing `TooLarge` SBOM becomes permanently uncleanable: `GetSBOM` hides it, `getSBOM` can't delete it, kubevuln regenerates, `StoreSBOM` calls Update, and `GuaranteedUpdate` silently drops the write and returns `nil` — so `retryErr` is nil and it even logs `"updated SBOM in storage"`. Net effect: the object stays pinned at the old version forever and every scan cycle redoes the full pull + catalog and silently fails to persist.

It's less severe than the original blocker (the in-memory scan still completes, so CVE results are produced), but it does mean the recovery path only works if the limit changes *before* the next scanner-version bump — and syft gets bumped fairly often here.

Fix would be to run the staleness/delete check against a version-agnostic read — e.g. a metadata-only `Get` in `getSBOM` before delegating to `GetSBOM`, or have `GetSBOM` surface "found but outdated" distinctly from "not found" so `getSBOM` can still purge it.

If you'd rather not grow this PR further, I'm fine landing it as-is once the OOM loop above is fixed, and tracking this one separately — say the word and I'll open the issue.

### Minor

- `pkg/sbomscanner/v1/server.go:125,201` use the string literals `"image-too-large"` / `"sbom-too-large"`, while `adapters/v1/sidecar.go` compares against `domain.ReasonImageTooLarge` / `domain.ReasonSBOMTooLarge`. Same module and no import cycle (`core/domain` only pulls in `storage/…/v1beta1`), so please use the constants — a typo on either side fails silently and just skips the annotation.
- `TestErrImageTooLarge_ErrorsIs_Or_Fallback` (`adapters/v1/syft_test.go`) constructs its own wrapped and unwrapped errors, so it asserts `errors.Is` semantics from the standard library rather than anything in kubevuln — it would pass with the production code deleted. The real coverage is the `maxImageSize: 1` table entry plus the new server test; consider dropping this one or repointing it at a real `syft.GetSource` error.
- `getSBOM` returns an empty SBOM even when `DeleteSBOM` fails (`core/services/scan.go:983-987`). On a delete failure that guarantees a regenerate whose write will be silently dropped. Returning the original `sbom` on delete failure would at least keep it to one wasted cycle.
- Keeping the `Fatal` → `Warning` change here is fine by me — no need to split it out.

singhshresth26:
I've addressed all the comments and pushed the fixes:

1. **Memory Limit Annotation Loop Fix**: Stored `domain.ScannerMemoryLimitAnnotationKey` unconditionally in `adapters/v1/sidecar.go` (even if empty) to ensure `GetMemoryLimit()` comparisons are well-defined on subsequent passes, avoiding the infinite delete/rescan loop.
2. **Version-Agnostic TooLarge Cleanup**: Updated `GetSBOM` in `repositories/apiserver.go` to return `domain.ErrOutdatedSBOM` along with the metadata/content on version mismatch. `getSBOM` in `core/services/scan.go` handles this by deleting the outdated SBOM to trigger a fresh scan.
3. **Graceful Delete Failures**: If `DeleteSBOM` fails in `getSBOM`, it now returns the original `sbom` and a `nil` error to prevent triggering wasted scan/regeneration cycles.
4. **Use Constants**: Swapped out the literal strings `"image-too-large"` and `"sbom-too-large"` in `pkg/sbomscanner/v1/server.go` for their corresponding `domain` constants.
5. **Test Updates**: Dropped the redundant test `TestErrImageTooLarge_ErrorsIs_Or_Fallback` in `adapters/v1/syft_test.go` (cleaning up its unused imports) and updated `TestAPIServerStore_GetSBOM` in `repositories/apiserver_test.go` to assert the new `domain.ErrOutdatedSBOM` mismatch behavior.

core/services/scan.go
```diff
@@ -942,3 +942,80 @@ func (s *ScanService) ValidateScanRegistry(ctx context.Context, workload domain.
...
+func (s *ScanService) getSBOM(ctx context.Context, name string, creatorVersion string) (domain.SBOM, error) {
+	sbom, err := s.sbomRepository.GetSBOM(ctx, name, creatorVersion)
+	if err != nil {
+		if errors.Is(err, domain.ErrOutdatedSBOM) {
+			logger.L().Ctx(ctx).Info("SBOM is outdated, deleting to trigger a fresh scan",
+				helpers.String("name", name))
+			// Delete both unfiltered and filtered SBOMs
+			if delErr := s.sbomRepository.DeleteSBOM(ctx, name); delErr != nil {
+				logger.L().Ctx(ctx).Warning("failed to delete outdated SBOM", helpers.Error(delErr), helpers.String("name", name))
+				return sbom, nil
```
matthyx: 🔴 **This returns an outdated SBOM as if it were current.**

`sbom` is what `GetSBOM` returned on the `ErrOutdatedSBOM` path, which now carries `Content: &manifest.Spec.Syft` (non-nil) and the *old* tool version. Returning it with a `nil` error makes the callers' `if sbom.Content == nil` gate false, so regeneration is skipped and the CVE scan runs against SBOM content from an older syft — while the CVE manifest is stored labelled with the current scanner version.

Pre-PR, an outdated SBOM was always discarded and regenerated, so this is a scan-correctness regression, not just wasted work.

The likely trigger is RBAC: `delete` on `sbomsyfts`/`sbomsyftfiltereds` is a new verb for kubevuln. If the ClusterRole isn't updated in lockstep, every delete returns Forbidden and every outdated SBOM is silently reused.

```go
if delErr := s.sbomRepository.DeleteSBOM(ctx, name); delErr != nil {
    logger.L().Ctx(ctx).Warning("failed to delete outdated SBOM", helpers.Error(delErr), helpers.String("name", name))
    return domain.SBOM{}, nil
}
```

Regenerating (even if the subsequent write gets silently dropped) is strictly better than scanning stale content.

Separately, worth gating this whole outdated-delete on `sbom.Status == helpersv1.TooLarge`: a non-`TooLarge` outdated SBOM was never frozen, and the existing Create → `AlreadyExists` → Update path handled it fine. As written, every syft bump issues 2 deletes + 1 create per image cluster-wide instead of 1 update.

matthyx (commented):
## Re-review (`9d09e85`)

Three of the four points are cleanly fixed:

- **OOM loop** — `ScannerMemoryLimitAnnotationKey` is now written unconditionally, with the `if s.memoryLimit != ""` guard correctly left around only the human-readable `StatusMetadataKey` message. That closes the delete/rescan loop.
- **`domain.Reason*` constants** in `server.go` instead of literals. 
- **Tautological `errors.Is` test** dropped.
- **Version-agnostic cleanup** — `GetSBOM` now returns the manifest plus `domain.ErrOutdatedSBOM` instead of swallowing it, and `getSBOM` purges on that sentinel. Nice touch adding the `wantErr` column to `TestAPIServerStore_GetSBOM` so the sentinel is actually asserted.

Verified locally on `9d09e85`: `go build ./...` passes, `go vet` clean apart from the pre-existing `adapters/v1/backend_test.go` findings, and `./core/...`, `./repositories/...`, `./pkg/sbomscanner/...` all pass.

One new problem, though — introduced by the delete-failure early return, and it's on me for not scoping my earlier suggestion.

---

### 🔴 On `DeleteSBOM` failure, an outdated SBOM is now used as if it were current

`core/services/scan.go:953-956`:

```go
if delErr := s.sbomRepository.DeleteSBOM(ctx, name); delErr != nil {
    logger.L().Ctx(ctx).Warning("failed to delete outdated SBOM", ...)
    return sbom, nil
}
```

`sbom` here is what `GetSBOM` returned on the outdated path — and that now carries `Content: &manifest.Spec.Syft` (non-nil) plus the *old* tool version. Returning it with a `nil` error means the callers' `if sbom.Content == nil` gate is false, so regeneration is skipped and the CVE scan runs against SBOM content produced by an older syft — while the resulting CVE manifest gets stored labelled with the *current* scanner version.

Before this PR an outdated SBOM was unconditionally discarded and regenerated, so this is a regression in scan correctness, not just efficiency.

The trigger is any `DeleteSBOM` failure, and the most likely one is RBAC: `delete` on `sbomsyfts`/`sbomsyftfiltereds` is a brand-new verb for kubevuln (see below). If the ClusterRole isn't updated in lockstep, *every* delete returns Forbidden and *every* outdated SBOM is silently reused — a missing permission quietly turns into stale-data scanning.

Fix is one line — fall back to the old discard-and-regenerate behaviour:

```go
if delErr := s.sbomRepository.DeleteSBOM(ctx, name); delErr != nil {
    logger.L().Ctx(ctx).Warning("failed to delete outdated SBOM", ...)
    return domain.SBOM{}, nil
}
```

To be clear about where my earlier note applies: returning the original `sbom` on delete failure is fine in the **stale `TooLarge`** branch at line 1013, because that object is terminal — it flows into the `Incomplete || TooLarge` check and yields `ErrIncompleteSBOM`, so nothing scans it. The outdated branch has no such backstop. Sorry for not drawing that line the first time.

### 🟡 RBAC prerequisite

`DeleteSBOM` is the first delete kubevuln issues against these resources. There's no RBAC manifest in this repo, so the ClusterRole change needs to land in the helm chart with or before this PR — otherwise the whole recovery path silently no-ops. Worth calling out in the PR description so it isn't missed at release time.

### 🟡 Consider scoping the outdated-delete to what actually needs it

An outdated SBOM that *isn't* `TooLarge` was never frozen — the existing Create → `AlreadyExists` → Update path handled it fine, and `GuaranteedUpdate` only short-circuits on the `TooLarge` annotation. Deleting unconditionally means every syft version bump now issues 2 deletes + 1 create per image across the whole cluster instead of 1 update, plus a window where the SBOM doesn't exist.

Gating the outdated-delete on `sbom.Status == helpersv1.TooLarge` keeps the fix and drops the churn. Not a blocker, but on a large cluster it's a meaningful difference.

### Nit

`pkg/sbomscanner/v1/server.go:28` — `core/domain` sorts before `pkg/sbomscanner/v1/proto`, so `gofmt -l` now flags this file. It was clean on `main` (a few other files in the repo aren't, so CI clearly doesn't gate on it — just keeping the delta clean).

---

Fix the outdated-delete fallback and confirm the RBAC change is queued, and this is good to go from my side.

core/services/scan.go
```diff
@@ -942,3 +942,80 @@ func (s *ScanService) ValidateScanRegistry(ctx context.Context, workload domain.
...
+		}
+
+		if stale {
+			logger.L().Ctx(ctx).Info("SBOM is stale/frozen due to changed limits, deleting to trigger a fresh scan",
+				helpers.String("name", name),
+				helpers.String("reason", reason))
+			// Delete both unfiltered and filtered SBOMs
+			if delErr := s.sbomRepository.DeleteSBOM(ctx, name); delErr != nil {
+				logger.L().Ctx(ctx).Warning("failed to delete stale SBOM", helpers.Error(delErr), helpers.String("name", name))
+				return sbom, nil
```
matthyx, line 1015: ✅ This one is correct as written — no change needed.

Returning the original `sbom` here is safe because the object is terminal: it carries `Status == TooLarge`, so it flows into the `Incomplete || TooLarge` check in the callers and yields `ErrIncompleteSBOM`. Nothing scans it.

That's the distinction I should have drawn in my earlier note — the outdated branch at line 955 has no such backstop, which is why it needs `domain.SBOM{}` instead.

matthyx (commented):
## Re-review (`8410433`) — blockers resolved

Both outstanding items are fixed correctly:

- **Delete-failure fallback** — the outdated path now returns `domain.SBOM{}, nil` instead of the stale manifest, so a failed delete falls back to discard-and-regenerate rather than silently scanning content from an older syft.
- **Gated on `TooLarge`** — the outdated-delete only fires for objects that were actually frozen. A non-`TooLarge` outdated SBOM still returns empty (so it regenerates) but goes through the cheap Create → `AlreadyExists` → Update path instead of delete+create. That's exactly the right split, and the new `TestScanService_getSBOM_Outdated` table pins all three combinations.
- The `server.go` import ordering nit is fixed too.

Verified locally on `8410433`: `go build ./...` passes, `go vet` is clean apart from the pre-existing `adapters/v1/backend_test.go` findings, and `./core/...`, `./repositories/...`, `./pkg/sbomscanner/...` all pass.

Nothing blocking left from me. Three small things, none of which need to hold the merge:

### 1. `gofmt` — `core/services/scan_test.go`

`gofmt -l` flags this file; three struct-field alignment spots (lines 455, 1163-1166, 1180-1187). A plain `gofmt -w core/services/scan_test.go` fixes it. The rest of your changed files are clean.

### 2. The headline recovery path still has no unit test

`TestScanService_getSBOM_Outdated` covers the `ErrOutdatedSBOM` branch nicely, but nothing exercises the limit-comparison path — no test anywhere references `MaxImageSizeAnnotationKey`, `MaxSBOMSizeAnnotationKey`, or `ScannerMemoryLimitAnnotationKey`. That's the actual headline behaviour of this PR ("raise `maxImageSize` and the SBOM gets re-scanned").

Extending the same table with `getErr: nil` rows would cover it — `TooLarge` + matching limit → no delete, + differing limit → delete, + missing annotation → delete. It needs a stub `sbomCreator` on the `ScanService` since that path calls `GetMaxImageSize()` / `GetMaxSBOMSize()` / `GetMemoryLimit()`; `schemaUnsupportedSBOMAdapter` in this file is already close to what's needed.

### 3. RBAC still needs confirming (not a code change here)

`DeleteSBOM` is the first delete kubevuln issues against `sbomsyfts` / `sbomsyftfiltereds`. The ClusterRole lives in the helm chart, not this repo, so that change needs to land with or before this — otherwise every delete returns Forbidden and the recovery path silently no-ops. Now that the fallback is fixed the failure mode is benign (wasted rescans, not stale data), but the feature won't actually work. Worth a line in the PR description.

Thanks for working through all of this — the final shape is a good deal more robust than a status rename.

core/services/scan.go
```diff
@@ -942,3 +942,82 @@ func (s *ScanService) ValidateScanRegistry(ctx context.Context, workload domain.
...
+	sbom, err := s.sbomRepository.GetSBOM(ctx, name, creatorVersion)
+	if err != nil {
+		if errors.Is(err, domain.ErrOutdatedSBOM) {
+			if sbom.Status == helpersv1.TooLarge {
+				logger.L().Ctx(ctx).Info("SBOM is outdated and too large, deleting to trigger a fresh scan",
+					helpers.String("name", name))
+				// Delete both unfiltered and filtered SBOMs
+				if delErr := s.sbomRepository.DeleteSBOM(ctx, name); delErr != nil {
+					logger.L().Ctx(ctx).Warning("failed to delete outdated SBOM", helpers.Error(delErr), helpers.String("name", name))
+					return domain.SBOM{}, nil
```
matthyx, line 956: Tiny cleanup, take it or leave it: this `return` is identical to the fall-through on line 959, so the branch can just log and continue:

```go
if errors.Is(err, domain.ErrOutdatedSBOM) {
    if sbom.Status == helpersv1.TooLarge {
        logger.L().Ctx(ctx).Info("SBOM is outdated and too large, deleting to trigger a fresh scan",
            helpers.String("name", name))
        if delErr := s.sbomRepository.DeleteSBOM(ctx, name); delErr != nil {
            logger.L().Ctx(ctx).Warning("failed to delete outdated SBOM", helpers.Error(delErr), helpers.String("name", name))
        }
    }
    return domain.SBOM{}, nil
}
```

Behaviour is identical — it just makes it obvious that every outdated path regenerates regardless of whether the delete worked.

core/services/scan_test.go
```diff
@@ -1152,3 +1157,90 @@ func TestOptionsFromWorkload(t *testing.T) {
 		})
 	}
 }
+
+type mockSBOMRepository struct {
+	ports.SBOMRepository
+	getSBOMSBOM domain.SBOM
+	getSBOMErr  error
+	deleteErr   error
+	deleteCalled bool
```
matthyx, line 1166: `gofmt -l` flags this file — this struct plus the test-table struct just below (lines ~1180-1187) and the one-line method at line 455. `gofmt -w core/services/scan_test.go` sorts all three; your other changed files are clean.

While you're in here: consider extending the table below with `getErr: nil` rows to cover the limit-comparison path — `TooLarge` + matching limit → no delete, differing limit → delete, missing annotation → delete. Nothing currently references `MaxImageSizeAnnotationKey` / `MaxSBOMSizeAnnotationKey` / `ScannerMemoryLimitAnnotationKey` in any test, so the PR's headline behaviour is untested. Those rows need a stub `sbomCreator` on the `ScanService` (that path calls `GetMaxImageSize()` and friends) — `schemaUnsupportedSBOMAdapter` above is nearly it already.

matthyx (approved):
LGTM — both blockers resolved. The delete-failure fallback and the `TooLarge` gate are correct, and verified locally on `8410433`: build passes, `go vet` clean apart from the pre-existing `adapters/v1/backend_test.go` findings, and `./core/...`, `./repositories/...`, `./pkg/sbomscanner/...` all pass.

Approving. The three items from my last pass are non-blocking and can be follow-ups: the `gofmt` fix on `core/services/scan_test.go`, test coverage for the limit-comparison path, and confirming the helm-chart ClusterRole grants `delete` on `sbomsyfts`/`sbomsyftfiltereds` before release — without that the recovery path silently no-ops.

The node-agent sidecar gap is tracked separately. Thanks for the thorough iteration on this one.

PR #394, opened 2026-07-29
Fix VEX update path swapping vulnerability ID and name for appended statements

### LINKED ISSUE #393: Bug: VEX update path swaps vulnerability ID and name for newly appended statements
# Description
When `APIServerStore.updateVEX` appends a statement for a CVE that was not already present in the stored `OpenVulnerabilityExchangeContainer`, it writes `Vulnerability.ID` from `v.Vulnerability.DataSource` and `Vulnerability.Name` from `v.Vulnerability.ID`. The create path (`createVEX`) uses the opposite mapping, so the same document's field mapping changes depending on whether a given statement was written during the first create or a later update.

## Summary
`createVEX` and `updateVEX` populate `Vulnerability.ID`/`Vulnerability.Name` from opposite source fields for newly appended statements, producing internally inconsistent VEX documents after an update.

## Affected Area
`repositories/apiserver.go` — `createVEX` and `updateVEX`.

# Environment
OS: `Any`

## Steps to Reproduce
2. Compare `createVEX` (`ID: v.Vulnerability.ID, Name: v.Vulnerability.DataSource`, around lines 811-817) with `updateVEX` (`ID: v.Vulnerability.DataSource, Name: v.Vulnerability.ID`, around lines 885-891).
3. Store a VEX document once via `StoreVEX`, then call `StoreVEX` again with a manifest containing an additional CVE not present in the first store.
4. Observe that the newly appended statement from the second call has its `Vulnerability.ID`/`Vulnerability.Name` fields swapped relative to statements written by the first call.

# Expected behavior
The create and update paths should emit the same `vulnerability.id`/`vulnerability.name` mapping for every statement, regardless of whether the statement was written on create or on a later update.

# Actual Behavior
Statements appended during an update mix a different field mapping than statements written during create, so the stored VEX document is internally inconsistent and can break consumers that key off `vulnerability.id`.

## Technical Evidence
The existing test `TestAPIServerStore_storeVEX` in `repositories/apiserver_test.go` exercises the update path but only asserts on the count of `affected` statements — it does not check field identity, so it does not catch this mismatch. Manually tracing the two code paths confirms the swap described above.

## Suspected Root Cause
`updateVEX`'s statement-append branch (the `if !found` block, roughly lines 885-891) was written with the `ID`/`Name` source fields transposed relative to the equivalent block in `createVEX`.

## Proposed Direction
Align `updateVEX`'s appended-statement mapping with `createVEX`'s (`ID: v.Vulnerability.ID`, `Name: v.Vulnerability.DataSource`), and add a regression test that stores a subset of matches, then updates with an additional match, and asserts the newly appended statement's ID/Name fields match the create-path mapping.

# Additional context
The repository already has a VEX update test in `repositories/apiserver_test.go`, but it only checks the count of affected statements and does not validate statement identity or field integrity.

## Acceptance Criteria
- [ ] `updateVEX` uses the same `Vulnerability.ID`/`Vulnerability.Name` mapping as `createVEX` for appended statements.
- [ ] A regression test demonstrates that a statement appended via the update path has the same field mapping as one written via create.
- [ ] Existing VEX tests continue to pass.

rootp1:
## Overview
Fixes a field-mapping bug where `updateVEX` wrote `Vulnerability.ID`/`Vulnerability.Name` from the opposite source fields compared to `createVEX`, causing newly appended statements during an update to have swapped ID/Name values relative to statements written during the initial create.

## Summary
- `repositories/apiserver.go`: aligned `updateVEX`'s appended-statement mapping (`ID: v.Vulnerability.ID`, `Name: v.Vulnerability.DataSource`) with `createVEX`'s existing mapping.
- `repositories/apiserver_test.go`: added `TestAPIServerStore_storeVEX_updatePreservesFieldMapping`, which stores a subset of matches, then updates with an additional match, and asserts the appended statement's `Vulnerability.ID`/`Name` match the create-path mapping. Verified this test fails against the pre-fix code and passes after the fix.

## Linked Issue
Fixes #393

## What Changed
- `repositories/apiserver.go` — swap corrected in `updateVEX`'s statement-append branch.
- `repositories/apiserver_test.go` — new regression test.

## Verification
- `go build ./...` — passed
- `go test ./repositories/...` — passed
- `go vet ./...` — passed (pre-existing warnings in `adapters/v1/backend_test.go` unrelated to this change)
- `gofmt -l .` — no newly introduced formatting issues (pre-existing unformatted files unrelated to this change)
- `go test ./...` — `core/services` and `pkg/sbomscanner/v1` fail in this sandbox due to missing Docker (testcontainers) and a macOS unix-socket path-length limitation; both are pre-existing environment constraints unrelated to this change, and were failing identically prior to it.

## Scope
No files outside `repositories/apiserver.go` and `repositories/apiserver_test.go` were changed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed vulnerability identity and data source mapping in VEX statement updates.
  * Normalized legacy statements with swapped fields to prevent duplicate entries.
  * Improved handling of non-URL data sources while preserving correct vulnerability mappings.
  * Prevented update errors when product or subcomponent details are unavailable.

* **Tests**
  * Added regression coverage for appended statements, legacy field normalization, and non-URL data sources.

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -749,7 +749,7 @@ func markRelevantVulnerabilitiesAsAffectedInVex(vexDoc *v1beta1.VEX, cvep *domai
 	// Now change the status of the filtered vulnerabilities to "Affected"
 	for _, v := range cvep.Content.Matches {
 		for i, s := range vexDoc.Statements {
-			if s.Vulnerability.ID == v.Vulnerability.ID {
+			if s.Vulnerability.Name == v.Vulnerability.ID {
 				foundProduct := false
 				for _, p := range s.Products {
 					for _, sc := range p.Subcomponents {
@@ -810,8 +810,8 @@ func (a *APIServerStore) createVEX(ctx context.Context, cve domain.CVEManifest, 
 
 		vexDoc.Statements = append(vexDoc.Statements, v1beta1.Statement{
 			Vulnerability: v1beta1.VexVulnerability{
-				ID:          v.Vulnerability.ID,
-				Name:        v.Vulnerability.DataSource,
+				ID:          v.Vulnerability.DataSource,
+				Name:        v.Vulnerability.ID,
 				Description: v.Vulnerability.Description,
 				Aliases:     aliases,
 			},
@@ -862,12 +862,29 @@ func (a *APIServerStore) updateVEX(ctx context.Context, cve domain.CVEManifest, 
 
 	// Extend the VEX document with vulnerability data from full vulnerability manifest
 	vexDoc := vexContainer.Spec
+
+	// Statements written before the ID/Name mapping was corrected to match createVEX
+	// carry the CVE identifier in ID and the data source URL in Name. Normalize them in
+	// place so the dedup below (which now keys on Name) also finds these older entries,
+	// instead of re-appending a duplicate for every one of them.
+	for i, s := range vexDoc.Statements {
+		if !strings.Contains(s.Vulnerability.ID, "://") && strings.Contains(s.Vulnerability.Name, "://") {
+			vexDoc.Statements[i].Vulnerability.ID, vexDoc.Statements[i].Vulnerability.Name = s.Vulnerability.Name, s.Vulnerability.ID
+		}
+	}
+
 	for _, v := range cve.Content.Matches {
 		found := false
 		for _, s := range vexDoc.Statements {
-			if s.Vulnerability.ID == v.Vulnerability.ID && v.Artifact.PURL == s.Products[0].Subcomponents[0].ID {
-				found = true
+			if s.Vulnerability.Name != v.Vulnerability.ID {
 				continue
+			}
+			if len(s.Products) == 0 || len(s.Products[0].Subcomponents) == 0 {
+				continue
+			}
+			if v.Artifact.PURL == s.Products[0].Subcomponents[0].ID {
+				found = true
+				break
 			}
 		}
 		if !found {
@@ -884,8 +901,8 @@ func (a *APIServerStore) updateVEX(ctx context.Context, cve domain.CVEManifest, 
 
 			vexDoc.Statements = append(vexDoc.Statements, v1beta1.Statement{
 				Vulnerability: v1beta1.VexVulnerability{
-					ID:          v.Vulnerability.ID,
-					Name:        v.Vulnerability.DataSource,
+					ID:          v.Vulnerability.DataSource,
+					Name:        v.Vulnerability.ID,
 					Description: v.Vulnerability.Description,
 					Aliases:     aliases,
 				},
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -868,7 +868,7 @@ func (a *APIServerStore) updateVEX(ctx context.Context, cve domain.CVEManifest, 
 	// place so the dedup below (which now keys on Name) also finds these older entries,
 	// instead of re-appending a duplicate for every one of them.
 	for i, s := range vexDoc.Statements {
-		if s.Vulnerability.ID != "" && !strings.Contains(s.Vulnerability.ID, "://") {
+		if !strings.Contains(s.Vulnerability.ID, "://") && (s.Vulnerability.Name == "" || strings.Contains(s.Vulnerability.Name, "://")) {
 			vexDoc.Statements[i].Vulnerability.ID, vexDoc.Statements[i].Vulnerability.Name = s.Vulnerability.Name, s.Vulnerability.ID
 		}
 	}
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -884,8 +884,8 @@ func (a *APIServerStore) updateVEX(ctx context.Context, cve domain.CVEManifest, 
 
 			vexDoc.Statements = append(vexDoc.Statements, v1beta1.Statement{
 				Vulnerability: v1beta1.VexVulnerability{
-					ID:          v.Vulnerability.DataSource,
-					Name:        v.Vulnerability.ID,
+					ID:          v.Vulnerability.ID,
+					Name:        v.Vulnerability.DataSource,
 					Description: v.Vulnerability.Description,
 					Aliases:     aliases,
 				},
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -868,7 +868,7 @@ func (a *APIServerStore) updateVEX(ctx context.Context, cve domain.CVEManifest, 
 	// place so the dedup below (which now keys on Name) also finds these older entries,
 	// instead of re-appending a duplicate for every one of them.
 	for i, s := range vexDoc.Statements {
-		if !strings.Contains(s.Vulnerability.ID, "://") && strings.Contains(s.Vulnerability.Name, "://") {
+		if s.Vulnerability.ID != "" && !strings.Contains(s.Vulnerability.ID, "://") {
 			vexDoc.Statements[i].Vulnerability.ID, vexDoc.Statements[i].Vulnerability.Name = s.Vulnerability.Name, s.Vulnerability.ID
 		}
 	}
--- a/repositories/apiserver_test.go
+++ b/repositories/apiserver_test.go
@@ -458,6 +458,48 @@ func TestAPIServerStore_updateVEX_normalizesLegacyStatements(t *testing.T) {
 	}
 }
 
+func TestAPIServerStore_updateVEX_doesNotNormalizeCorrectShapeWithNonURLDataSource(t *testing.T) {
+	cveManifestFull := tools.FileToCVEManifest("testdata/nginx-cve.json")
+	cveManifestFiltered := tools.FileToCVEManifest("testdata/nginx-cve-filtered.json")
+	require.NotEmpty(t, cveManifestFull.Content.Matches)
+
+	a := NewFakeAPIServerStorage("kubescape")
+	ctx := context.TODO()
+	workload := domain.ScanCommand{
+		ImageHash:     "sha256:32fdf92b4e986e109e4db0865758020cb0c3b70d6ba80d02fe87bad5cc3dc228",
+		InstanceID:    "apiVersion-apps/v1/namespace-kubescape/kind-ReplicaSet/name-kubevuln-65bfbfdcdd/containerName-kubevuln",
+		Wlid:          "wlid://cluster-aaa/namespace-anyNamespaceJob/job-anyJob",
+		ImageTag:      "registry.k8s.io/coredns/coredns:v1.10.1",
+		ContainerName: "anyJobContName",
+	}
+	ctx = context.WithValue(ctx, domain.WorkloadKey{}, workload)
+
+	require.NoError(t, a.StoreVEX(ctx, cveManifestFull, cveManifestFiltered, false))
+
+	vexContainer, err := a.StorageClient.OpenVulnerabilityExchangeContainers(a.Namespace).Get(context.Background(), cveManifestFull.Name, metav1.GetOptions{})
+	require.NoError(t, err)
+	statementCountAfterCreate := len(vexContainer.Spec.Statements)
+
+	// A correct-shape statement whose data source happens not to be a URL (ID holds
+	// it verbatim) must not be mistaken for a legacy statement and swapped.
+	for i := range vexContainer.Spec.Statements {
+		vexContainer.Spec.Statements[i].Vulnerability.ID = "nvd"
+	}
+	_, err = a.StorageClient.OpenVulnerabilityExchangeContainers(a.Namespace).Update(context.Background(), vexContainer, metav1.UpdateOptions{})
+	require.NoError(t, err)
+
+	require.NoError(t, a.StoreVEX(ctx, cveManifestFull, cveManifestFiltered, false))
+
+	vexContainerAfterUpdate, err := a.StorageClient.OpenVulnerabilityExchangeContainers(a.Namespace).Get(context.Background(), cveManifestFull.Name, metav1.GetOptions{})
+	require.NoError(t, err)
+	assert.Equal(t, statementCountAfterCreate, len(vexContainerAfterUpdate.Spec.Statements))
+
+	for _, s := range vexContainerAfterUpdate.Spec.Statements {
+		assert.Equal(t, "nvd", s.Vulnerability.ID)
+		assert.NotContains(t, s.Vulnerability.Name, "://")
+	}
+}
+
 func TestAPIServerStore_StoreCVESummaryStub(t *testing.T) {
 	workload := domain.ScanCommand{
 		Wlid:          "wlid://cluster-kind/namespace-local-path-storage/deployment-local-path-provisioner",
--- a/repositories/apiserver_test.go
+++ b/repositories/apiserver_test.go
@@ -398,15 +398,64 @@ func TestAPIServerStore_storeVEX_updatePreservesFieldMapping(t *testing.T) {
 	var appended *v1beta1.Statement
 	for i := range vexContainer.Spec.Statements {
 		s := &vexContainer.Spec.Statements[i]
-		if s.Vulnerability.ID == lastMatch.Vulnerability.ID && len(s.Products) > 0 && len(s.Products[0].Subcomponents) > 0 &&
+		if s.Vulnerability.Name == lastMatch.Vulnerability.ID && len(s.Products) > 0 && len(s.Products[0].Subcomponents) > 0 &&
 			s.Products[0].Subcomponents[0].ID == lastMatch.Artifact.PURL {
 			appended = s
 			break
 		}
 	}
-	require.NotNil(t, appended, "expected the statement appended during update to have ID == Vulnerability.ID and matching PURL, matching the create-path mapping")
-	assert.Equal(t, lastMatch.Vulnerability.ID, appended.Vulnerability.ID)
-	assert.Equal(t, lastMatch.Vulnerability.DataSource, appended.Vulnerability.Name)
+	require.NotNil(t, appended, "expected the statement appended during update to have Name == Vulnerability.ID and matching PURL, matching the create-path mapping")
+	assert.Equal(t, lastMatch.Vulnerability.ID, appended.Vulnerability.Name)
+	assert.Equal(t, lastMatch.Vulnerability.DataSource, appended.Vulnerability.ID)
+}
+
+// TestAPIServerStore_updateVEX_normalizesLegacyStatements guards against a regression where
+// statements written with the pre-fix, swapped ID/Name mapping would no longer match the
+// current Name-keyed dedup in updateVEX, causing every legacy statement to be duplicated
+// (and left permanently un-promotable to "affected") on the next scan.
+func TestAPIServerStore_updateVEX_normalizesLegacyStatements(t *testing.T) {
+	cveManifestFull := tools.FileToCVEManifest("testdata/nginx-cve.json")
+	cveManifestFiltered := tools.FileToCVEManifest("testdata/nginx-cve-filtered.json")
+	require.NotEmpty(t, cveManifestFull.Content.Matches)
+
+	a := NewFakeAPIServerStorage("kubescape")
+	ctx := context.TODO()
+	workload := domain.ScanCommand{
+		ImageHash:     "sha256:32fdf92b4e986e109e4db0865758020cb0c3b70d6ba80d02fe87bad5cc3dc228",
+		InstanceID:    "apiVersion-apps/v1/namespace-kubescape/kind-ReplicaSet/name-kubevuln-65bfbfdcdd/containerName-kubevuln",
+		Wlid:          "wlid://cluster-aaa/namespace-anyNamespaceJob/job-anyJob",
+		ImageTag:      "registry.k8s.io/coredns/coredns:v1.10.1",
+		ContainerName: "anyJobContName",
+	}
+	ctx = context.WithValue(ctx, domain.WorkloadKey{}, workload)
+
+	require.NoError(t, a.StoreVEX(ctx, cveManifestFull, cveManifestFiltered, false))
+
+	vexContainer, err := a.StorageClient.OpenVulnerabilityExchangeContainers(a.Namespace).Get(context.Background(), cveManifestFull.Name, metav1.GetOptions{})
+	require.NoError(t, err)
+	statementCountAfterCreate := len(vexContainer.Spec.Statements)
+
+	// Simulate a document written by the pre-fix update path: ID holds the CVE
+	// identifier and Name holds the data source URL, the opposite of the current mapping.
+	for i := range vexContainer.Spec.Statements {
+		s := &vexContainer.Spec.Statements[i]
+		s.Vulnerability.ID, s.Vulnerability.Name = s.Vulnerability.Name, s.Vulnerability.ID
+	}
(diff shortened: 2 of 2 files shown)


repositories/apiserver.go
```diff
@@ -884,8 +884,8 @@ func (a *APIServerStore) updateVEX(ctx context.Context, cve domain.CVEManifest,
 
 			vexDoc.Statements = append(vexDoc.Statements, v1beta1.Statement{
 				Vulnerability: v1beta1.VexVulnerability{
-					ID:          v.Vulnerability.DataSource,
-					Name:        v.Vulnerability.ID,
+					ID:          v.Vulnerability.ID,
+					Name:        v.Vulnerability.DataSource,
```
matthyx: **Blocker: this aligns `updateVEX` to `createVEX`, but `createVEX` looks like the side that's actually wrong.**

`v1beta1.VexVulnerability` (kubescape/storage v0.0.258, `pkg/apis/softwarecomposition/v1beta1/types.go:363`) defines the two fields as:

```go
//  ID is an IRI to reference the vulnerability in the statement.
ID string `json:"@id,omitempty"`

// Name is the main vulnerability identifier.
Name string `json:"name,omitempty"`
```

That is the OpenVEX `vulnerability` object, and the document these statements land in declares `@context: "https://openvex.dev/ns/v0.2.0"` (`apiserver.go:788`). By that contract the CVE ID belongs in `name` and the data-source URL — an IRI — belongs in `@id`, which is exactly the mapping this PR *removes*. After this change every emitted statement serialises as:

```json
"vulnerability": {
  "@id": "CVE-2022-34903",
  "name": "https://security-tracker.debian.org/tracker/CVE-2022-34903"
}
```

`@id` is not an IRI and `name` is not a vulnerability identifier.

A second signal that `Name` is meant to hold the CVE ID: `sortVexStatements` (`apiserver.go:983`) orders statements by `Vulnerability.Name` before the canonical hash is computed. That mirrors the OpenVEX canonicalization algorithm, which sorts by vulnerability *name*. With this mapping the document is hashed in data-source-URL order, so `Metadata.ID` won't agree with a spec-conformant implementation's hash of the same document.

I'm not asking you to just flip the constant — the reason `createVEX`'s mapping is load-bearing is that two lookups key off it:

- `apiserver.go:868` — `s.Vulnerability.ID == v.Vulnerability.ID` (this dedup)
- `apiserver.go:752` — `s.Vulnerability.ID == v.Vulnerability.ID` in `markRelevantVulnerabilitiesAsAffectedInVex`

So the spec-correct fix is the mirror image of this PR: keep `ID: v.Vulnerability.DataSource, Name: v.Vulnerability.ID` here, apply the same to `createVEX` (`apiserver.go:813-814`), and change those two comparisons to `s.Vulnerability.Name == v.Vulnerability.ID`. Same size of change, same regression test shape.

The one argument for the direction you took is if downstream consumers (kubescape backend / UI) already read `@id` expecting a bare CVE ID, in which case create's mapping is the de-facto contract and the schema comment is the thing that's stale. I can't verify that from this repo. Could a maintainer confirm which of the two is canonical before this lands? Whichever way it goes, it should be decided rather than inherited from whichever function was written first — and the field comments in `kubescape/storage` should be corrected if `@id` really is meant to carry a CVE ID.

repositories/apiserver.go
```diff
@@ -884,8 +884,8 @@ func (a *APIServerStore) updateVEX(ctx context.Context, cve domain.CVEManifest,
 
 			vexDoc.Statements = append(vexDoc.Statements, v1beta1.Statement{
```
matthyx, line 902: **Blocker: VEX containers already in clusters double in size and end up self-contradictory after this change.**

Any `OpenVulnerabilityExchangeContainer` in the field that has been through the old update path contains statements shaped `ID: <dataSource URL>, Name: <CVE-ID>`. The dedup at line 868 compares `s.Vulnerability.ID == v.Vulnerability.ID`, so after this fix those legacy statements no longer match anything, and the loop appends a fresh statement for every one of them.

I measured it on your branch: store `testdata/nginx-cve.json` (456 matches), rewrite the stored statements into the legacy shape to simulate a pre-fix document, then run the same scan again through `StoreVEX`:

```
statements after create=456, after update over legacy-shaped doc=912
affected statements=432
```

Every vulnerability now appears twice. Worse, `markRelevantVulnerabilitiesAsAffectedInVex` also matches on `.ID`, so only the new-format copies get promoted to `affected` while the legacy copies stay pinned at `not_affected` with `"Vulnerable component is not loaded into the memory"`. The document now asserts both statuses for the same vulnerability on the same product — a consumer reading it gets a coin flip. This is not self-healing: the duplicates persist across all subsequent updates, since from then on the dedup finds the new-format copy and stops.

Two ways out, either is fine:

1. Normalise on read — before the dedup loop, walk `vexDoc.Statements` and rewrite any statement whose `ID` doesn't look like the expected identifier into the current mapping. One-time migration, no duplicates, no doubling.
2. Widen the dedup to tolerate both shapes: `s.Vulnerability.ID == v.Vulnerability.ID || s.Vulnerability.Name == v.Vulnerability.ID`. Cheaper, but leaves legacy statements permanently in the old shape and `markRelevantVulnerabilitiesAsAffectedInVex` still won't promote them, so option 1 is the one I'd take.

Worth a test either way — a fixture in the legacy shape run through `StoreVEX`, asserting the statement count is unchanged. That case is what will actually hit users on upgrade, and neither the existing tests nor the new one covers it.

matthyx (changes requested):
Reviewed the change and reproduced both directions locally.

**The bug is real and the regression test is genuine.** I reverted only `repositories/apiserver.go` on your branch and re-ran the new test: it fails with `expected the statement appended during update to have ID == Vulnerability.ID`, and passes with the fix. The impact is also worse than "inconsistent document" — because both `updateVEX`'s dedup (`apiserver.go:868`) and `markRelevantVulnerabilitiesAsAffectedInVex` (`apiserver.go:752`) match on `s.Vulnerability.ID == v.Vulnerability.ID`, statements written by the old update path were re-appended on every subsequent scan and could never be promoted to `affected`.

That said, I don't think this is mergeable as-is. Two blockers, both inline.

**Blocker 1 — the fix may be aligned in the wrong direction** (inline on `ID`/`Name`). Per the schema, `ID` is `@id` "an IRI to reference the vulnerability" and `Name` is "the main vulnerability identifier". By that definition the *pre-fix* `updateVEX` mapping was the correct one and `createVEX` is the buggy side. This needs a maintainer call, not a coin flip — details inline.

**Blocker 2 — already-stored VEX containers double in size and become self-contradictory** (inline). Verified empirically: 456 statements → 912. Details inline.

Two smaller things, both in the block immediately above the change (out of diff range, so noting them here):

- `apiserver.go:870` — `continue` inside `for _, s := range vexDoc.Statements` should be `break`. After a match it keeps scanning every remaining statement, making the loop unconditionally O(n²) (456 × 456 on the test fixture alone). Behaviour is correct, cost is not.
- `apiserver.go:868` — `s.Products[0].Subcomponents[0].ID` indexes both slices unguarded. Any statement reaching `updateVEX` without a product, or with a product carrying no subcomponents, panics the scan. Pre-existing, but this PR makes the surrounding branch the one that matters, and a `len()` guard is two lines.

I'd also second CodeRabbit's point on the test — matching the appended statement by `Vulnerability.ID` alone is fragile. `testdata/nginx-cve.json` has 456 matches with many CVE IDs repeated across packages; the test only works today because the last match (`CVE-2022-34903`) happens to be unique in that file. Matching on ID **and** subcomponent PURL, plus asserting the statement count grew by exactly one, would also cover the duplicate-append regression, which is the more damaging half of this bug and is currently untested.

repositories/apiserver.go
rootp1: You're right, flipped the mapping the other way instead. createVEX now writes ID as the data source URL and Name as the CVE id, same as updateVEX did before this PR touched it, and I updated the two dedup checks to compare Name. Also normalized statements written with the old swapped fields so they get picked up by the new comparison instead of duplicating.

repositories/apiserver.go
rootp1: Good catch, added normalization for this. On update I now walk the existing statements first and swap any that look legacy, basically if ID doesn't look like a URL and Name does, before running the dedup. That way old statements land in the same shape as new ones and stop being duplicated. Added a test that seeds a legacy-shaped doc and checks the statement count stays flat after another scan.

repositories/apiserver_test.go
```diff
@@ -347,6 +347,62 @@ func TestAPIServerStore_storeVEX(t *testing.T) {
...
+	for i := range vexContainer.Spec.Statements {
+		s := &vexContainer.Spec.Statements[i]
+		if s.Vulnerability.ID == lastMatch.Vulnerability.ID {
+			appended = s
+			break
+		}
+	}
+	require.NotNil(t, appended, "expected the statement appended during update to have ID == Vulnerability.ID, matching the create-path mapping")
+	assert.Equal(t, lastMatch.Vulnerability.ID, appended.Vulnerability.ID)
+	assert.Equal(t, lastMatch.Vulnerability.DataSource, appended.Vulnerability.Name)
```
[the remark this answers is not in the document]
rootp1: Fixed, the appended statement is now matched by Vulnerability.Name (the CVE id under the new mapping) plus the subcomponent PURL, and I assert the statement count grows by one after the update. Thanks for catching this.

rootp1:
Also picked up the two smaller points from the review body. The dedup loop now breaks once it finds a match instead of continuing through the rest of the statements, and there's a length check before indexing into Products and Subcomponents so a statement without a product doesn't panic the scan. Pushed everything together in the latest commit.

repositories/apiserver.go
```diff
@@ -862,13 +862,30 @@ func (a *APIServerStore) updateVEX(ctx context.Context, cve domain.CVEManifest,
 
 	// Extend the VEX document with vulnerability data from full vulnerability manifest
 	vexDoc := vexContainer.Spec
+
+	// Statements written before the ID/Name mapping was corrected to match createVEX
+	// carry the CVE identifier in ID and the data source URL in Name. Normalize them in
+	// place so the dedup below (which now keys on Name) also finds these older entries,
+	// instead of re-appending a duplicate for every one of them.
+	for i, s := range vexDoc.Statements {
+		if !strings.Contains(s.Vulnerability.ID, "://") && strings.Contains(s.Vulnerability.Name, "://") {
```
matthyx: Non-blocking: this requires `Name` to contain `://` before it will treat a statement as legacy. A legacy statement whose `dataSource` was empty is shaped `ID: "CVE-x", Name: ""` — no `://` on either side, so it falls through un-normalized and still gets duplicated once by the loop below.

In the corrected mapping `ID` is always either a data-source URL or empty, so anything non-empty that isn't a URL is legacy by definition:

```go
if s.Vulnerability.ID != "" && !strings.Contains(s.Vulnerability.ID, "://") {
```

Same behaviour on every case your version handles, one condition shorter, and no blind spot on empty `dataSource`.

To be clear about how speculative this is: I could not produce a real instance. All 1,129 matches across `nginx-cve.json`, `nginx-cve-filtered.json` and `nginx-cve-filtered-2.json` have URL data sources, and grype populates `dataSource` from the DB provider, so an empty one may not be reachable in practice. Purely defensive — fine to leave as-is if you'd rather not churn the diff.

matthyx (dismissed):
Re-checked at `6e901f8`. Both blockers are resolved, and I found independent confirmation that the direction you picked is the right one. Approving.

**Blocker 1 — resolved, and the direction is now corroborated by something better than the field comments.** I went looking for a canonical reference and found it in the schema's own repo: `kubescape/storage`, `artifacts/openvulnerabilityexchangecontainer/example-01.yaml`, emits exactly

```yaml
vulnerability:
  "@id": https://nvd.nist.gov/vuln/detail/CVE-2020-8908
  name: CVE-2020-8908
```

which is precisely what this PR now produces. So the flip isn't just satisfying a doc comment — it matches the reference artifact for this CRD. `sortVexStatements` now also sorts by the CVE ID via `Name`, which lines up with the OpenVEX canonicalization algorithm.

I also checked the compat risk I raised, since this now changes the **create** path for every new document. Searched the kubescape org for consumers of `OpenVulnerabilityExchangeContainer`: the only real client is `kubescape/headlamp-plugin`, which declares the `VexVulnerability` type but never reads `@id` or `name` off statements. No consumer breakage that I can find.

**Blocker 2 — resolved.** The normalization does what's needed, and I checked the case I care about most: a document created by the old create path and then extended by the old update path contains *both* shapes at once, which is the realistic production state. Your heuristic converges both onto the new shape in one pass, and it can't false-positive on a correct statement since a CVE identifier never contains `://`. Verified the test is a genuine guard — deleting just the normalization loop makes `TestAPIServerStore_updateVEX_normalizesLegacyStatements` fail, and the flat-statement-count assertion is what would have caught the 456→912 doubling.

Also confirmed the two smaller things came along: `break` instead of `continue`, and the `len(s.Products)` / `len(Subcomponents)` guards.

On my box at `6e901f8`: `go build ./...`, `go vet ./repositories/`, `gofmt -l repositories/` clean, `go test ./repositories/` passes.

Two non-blocking notes, neither worth holding the merge:

**1. The `://` heuristic has one blind spot.** It requires `Name` to contain `://` to recognise a legacy statement. A legacy statement whose `dataSource` was empty is shaped `ID: "CVE-x", Name: ""` — no `://` anywhere, so it isn't normalised and gets duplicated once anyway. A tighter rule with no blind spot: in the corrected mapping `ID` is *always* either a data-source URL or empty, so anything non-empty without `://` is legacy by definition:

```go
if s.Vulnerability.ID != "" && !strings.Contains(s.Vulnerability.ID, "://") {
```

One condition shorter and it also catches the empty-`dataSource` case. Worth saying I could not produce a real instance of this — all 1,129 matches across `nginx-cve.json`, `nginx-cve-filtered.json` and `nginx-cve-filtered-2.json` have URL data sources, and grype populates `dataSource` from the DB provider — so this is defensive, not a known break.

**2. Downgrade hazard worth a release note.** The normalization is one-directional. If a cluster rolls back to a pre-fix kubevuln after this ships, the old update path dedups on `s.Vulnerability.ID == v.Vulnerability.ID`, which won't match the new-shape documents this version writes, and it will duplicate them on the next scan. Nothing to fix here — the old code is the buggy one — but operators rolling back should know they'll want to delete the affected containers and let them regenerate.

Nice work on the turnaround, and thanks for taking the harder of the two fix directions rather than the smaller diff.

matthyx:
Separate from the review — heads up that **DCO is now failing** and will block the merge. The two new commits are missing the sign-off trailer that `ac5efe6` has:

```
6e901f8  fix: align VEX Vulnerability ID/Name to OpenVEX schema and dedup on Name   (no Signed-off-by)
d92ddc0  fix: match appended VEX statement by ID and PURL in regression test        (no Signed-off-by)
ac5efe6  fix: correct swapped vulnerability ID/name in VEX update path              Signed-off-by: rootp1 <arnav.iitr@gmail.com>
```

Quickest fix:

```
git rebase --signoff ac5efe6~1
git push --force-with-lease
```

Every other check on the branch is green.

repositories/apiserver.go
rootp1: @matthyx good point, switched to checking ID alone, non empty and not a URL means legacy. Simpler and covers the empty dataSource case too. Pushed the change.

repositories/apiserver.go
```diff
@@ -862,13 +862,30 @@ func (a *APIServerStore) updateVEX(ctx context.Context, cve domain.CVEManifest,
 
 	// Extend the VEX document with vulnerability data from full vulnerability manifest
 	vexDoc := vexContainer.Spec
+
+	// Statements written before the ID/Name mapping was corrected to match createVEX
+	// carry the CVE identifier in ID and the data source URL in Name. Normalize them in
+	// place so the dedup below (which now keys on Name) also finds these older entries,
+	// instead of re-appending a duplicate for every one of them.
+	for i, s := range vexDoc.Statements {
+		if s.Vulnerability.ID != "" && !strings.Contains(s.Vulnerability.ID, "://") {
```
matthyx: **This is the line to change — and it's my fault, I suggested it.** The version this replaced was safe in a way this one isn't.

`6e901f8` required `Name` to contain `://` before swapping. In the corrected mapping `Name` holds a CVE identifier, which can never contain `://`, so a correct statement was *structurally* unswappable — the guarantee didn't depend on what `dataSource` happens to hold. This version keys only on `ID` not looking like a URL, so a correct statement whose `dataSource` is non-empty and not a URL now matches and gets swapped.

Measured on this branch, fixture with `dataSource = "nvd"`:

```
after create:   n=456  ID="nvd" Name="CVE-2005-2541"     <- correct shape
after update 1: n=912  ID="nvd" Name="CVE-2005-2541"     <- doubled
after update 2: n=912
```

The swapped statements no longer match the `Name`-keyed dedup below, so all 456 are re-appended. And it never settles: each scan swaps both halves again, so the half that `markRelevantVulnerabilitiesAsAffectedInVex` can promote to `affected` alternates every time. That is blocker 2 all over again, fired by ordinary data rather than legacy data.

Suggested replacement, which keeps the structural guarantee *and* covers the empty-`dataSource` legacy case I was originally reaching for:

```go
if !strings.Contains(s.Vulnerability.ID, "://") && (s.Vulnerability.Name == "" || strings.Contains(s.Vulnerability.Name, "://")) {
```

A correct statement has a CVE identifier in `Name` — neither empty nor a URL — so it can never be swapped, regardless of `dataSource`. Verified locally across all four shapes:

```
non-URL dataSource, correct shape:     456 -> 456   (this commit: 912)
empty dataSource, correct shape:       456 -> 456
legacy shape, empty dataSource:        456 -> 456   (6e901f8: 912)
legacy shape, URL dataSource:          456 -> 456
```

Both of your regression tests pass unchanged, and `go test ./repositories/` is green.

Worth adding a test for the first row — a manifest whose `dataSource` isn't a URL, stored then re-stored, asserting the count is flat. That's the case neither predicate's test currently covers, and it's what would have caught this.

Reverting to `6e901f8` is equally acceptable if you'd rather not churn again; the gap I was chasing needs a legacy document *and* an empty `dataSource` simultaneously, which nobody has shown is reachable.

matthyx (changes requested):
Re-checked at `92415b2`. **My previous suggestion was wrong and you should not ship it — please revert or replace that line.** Apologies, this one is on me: I proposed the predicate in `92415b2` to close a speculative blind spot, and in doing so I traded a harmless false-negative for a false-positive that re-creates the exact doubling blocker we spent this review closing. Details and a predicate that's actually safe are inline.

Short version. `6e901f8`'s condition was safe *by construction*: it required `Name` to contain `://` before swapping, and in the corrected mapping `Name` holds a CVE identifier, which can never contain `://`. So it could never mangle a correct statement — its only weakness was declining to normalize a legacy statement whose `dataSource` was empty. `92415b2` drops the `Name` side and keys purely on `ID` not looking like a URL. That now depends on an assumption about `dataSource`'s *content* rather than on a structural invariant, and when the assumption doesn't hold it swaps correct statements.

Measured on your branch, feeding the fixture through with `dataSource` set to a non-URL, non-empty value:

```
after create:   n=456  ID="nvd" Name="CVE-2005-2541"     <- correct shape
after update 1: n=912  ID="nvd" Name="CVE-2005-2541"     <- doubled
after update 2: n=912
```

Correct, current-shape statements get swapped by the normalization, the `Name`-keyed dedup below then can't find them, and all 456 are re-appended. It also doesn't settle: on each subsequent scan the two halves swap shape again, so which half `markRelevantVulnerabilitiesAsAffectedInVex` can promote to `affected` alternates per scan. That's blocker 2 again, except now it fires on ordinary data instead of legacy data.

The predicate that handles every case:

```go
if !strings.Contains(s.Vulnerability.ID, "://") && (s.Vulnerability.Name == "" || strings.Contains(s.Vulnerability.Name, "://")) {
```

It keeps `6e901f8`'s structural guarantee — a correct statement has a CVE identifier in `Name`, which is neither empty nor a URL, so it can never be swapped — while still catching the legacy-plus-empty-`dataSource` case that motivated the change. I patched it in locally and ran all four scenarios:

```
non-URL dataSource, correct shape:     456 -> 456   (92415b2 gives 912)
empty dataSource, correct shape:       456 -> 456
legacy shape, empty dataSource:        456 -> 456   (normalized, 6e901f8 gives 912)
legacy shape, URL dataSource:          456 -> 456
```

Your two regression tests pass unchanged with it, and the whole `./repositories/` suite is green.

Reverting `92415b2` outright is also a perfectly fine outcome — `6e901f8` was already correct for every case anyone has demonstrated, and the gap I was pointing at needs a legacy document *and* an empty `dataSource` at the same time. I'd take the combined predicate since it costs one clause and closes both directions, but I don't want my speculation driving churn twice. Your call; either way I'll re-approve.

Also still outstanding, unchanged from my last pass: **DCO is failing.** `d92ddc0`, `6e901f8` and `92415b2` are all missing the sign-off trailer that `ac5efe6` has — `git rebase --signoff ac5efe6~1 && git push --force-with-lease` covers all three.

matthyx (approved):
Re-checked at `44f8df0`. This is right — approving, and this time I've mutation-tested the predicate rather than reasoning about it, which is what I should have done before making the suggestion that broke it last round.

**The regression I introduced is fixed.** The normalization now reads:

```go
if !strings.Contains(s.Vulnerability.ID, "://") && (s.Vulnerability.Name == "" || strings.Contains(s.Vulnerability.Name, "://")) {
```

which restores the structural guarantee: a correct statement carries a CVE identifier in `Name`, which is neither empty nor a URL, so it cannot be swapped no matter what `dataSource` holds.

**The new test is a genuine guard**, not decoration. I swapped the predicate back to `92415b2`'s version and `TestAPIServerStore_updateVEX_doesNotNormalizeCorrectShapeWithNonURLDataSource` fails; restore it and the suite is green. That's the case that would have caught the 456→912 doubling on ordinary data.

**DCO passes** — all five commits now carry the sign-off trailer.

On my box at `44f8df0`: `go build ./...`, `go vet ./repositories/`, `gofmt -l repositories/` clean, `go test ./repositories/` passes.

Everything from the earlier rounds still holds: the `@id`/`name` flip matches `kubescape/storage`'s own reference artifact, no consumer in the org reads those fields off statements, the dedup and `markRelevantVulnerabilitiesAsAffectedInVex` both key on `Name` consistently, and the `break` and `len()` guards are in place.

One non-blocking gap, **test-only — no code change wanted here**:

Reverting the predicate all the way back to `6e901f8`'s version (`!Contains(ID, "://") && Contains(Name, "://")`) still passes the entire suite. So the `s.Vulnerability.Name == ""` clause — the part that handles a legacy statement whose `dataSource` was empty — isn't pinned by anything. If you want it locked in, the case is a legacy-shaped document built from a manifest with `DataSource` cleared: `ID` holds the CVE identifier, `Name` is empty, and the assertion is that the statement count stays flat across the next `StoreVEX`. That's the same shape as your `normalizesLegacyStatements` test with one field blanked.

And one thing that is **deliberately not actionable**, flagging it only so nobody tries to close it later: a legacy statement whose `dataSource` was non-empty and not a URL is genuinely indistinguishable from a correct one by field shape alone, so it will still be duplicated once. Closing that would need a version marker on the document, which is not worth it for a case nobody has shown is reachable. The current predicate is the right stopping point.

Sorry again for the detour on `92415b2` — thanks for turning it around quickly, and for adding the regression test rather than just reverting.

rootp1:
@matthyx I reverted the previous condition and used the combined predicate you gave, requiring Name to be empty or a URL alongside the ID check, so a correct statement can no longer be swapped. Added a regression test for the non-URL dataSource false positive you found. Also rebased with signoff so DCO is clean on all commits now.

PR #395, opened 2026-07-30
fix: log real error instead of preceding sentinel/nil at 5 sites

### LINKED ISSUE #389: fix: wrong error variable captured in log statements (5 sites across apiserver.go and backend.go)
## Description

Two related bugs with the same root cause: a log statement captures the wrong Go variable — one that's guaranteed `nil` or semantically meaningless — instead of the actual error that just occurred. In incident response, this actively misdirects operators.

- **Bug A — `apiserver.go` (4 sites):** In `StoreCVE` (L351), `StoreCVESummary` (L588), and both `StoreSBOM` branches (L1127, L1160), when a Create-on-conflict triggers a `RetryOnConflict` update, the retry-exhaustion log uses `helpers.Error(err)` — the original `IsAlreadyExists` sentinel. The string it prints is literally `"vulnerabilitymanifests.spdx.softwarecomposition.kubescape.io \"<name>\" already exists"`. That sounds benign. The actual failure — etcd timeout, RBAC denial, connection reset — is in `retryErr` and gets dropped entirely.
- **Bug B — `backend.go` (1 site):** In `SubmitCVE` (L280-283), when `ParseImageManifest` fails, the parse error is stored in a local `e`. But the warning logs `helpers.Error(err)` — not `e`. At that point in the function, `err` is guaranteed `nil` (both prior `err` checks returned early). So every image manifest parse failure logs `nil` instead of the actual parse error.

---

## Environment

- **OS:** Any
- **Version:** Any (both bugs exist on `main` as of July 2026)

---

## Steps To Reproduce

2. Search for `Warning("failed to update` — each of the 4 instances uses `helpers.Error(err)` where `err` is the Create sentinel, not the retry result.
3. Open `adapters/v1/backend.go`.
4. Read around line 280 — `helpers.Error(err)` is logged when the actual parse error is `e`.

---

## Expected Behavior

Every warning log should contain the real failure error — not a preceding sentinel, and not `nil`.

---

## Actual Behavior

- Retry-exhaustion warnings show `"already exists"` instead of the actual update failure (`apiserver.go` ×4).
- Image manifest parse failures show `<nil>` instead of the JSON parse error (`backend.go` ×1).

---

## Proposed Solution

The fix is mechanical — 5 one-line changes:

In `repositories/apiserver.go`:
- **Line 351:** `helpers.Error(err)` → `helpers.Error(retryErr)`
- **Line 588:** `helpers.Error(err)` → `helpers.Error(retryErr)`
- **Line 1127:** `helpers.Error(err)` → `helpers.Error(retryErr)`
- **Line 1160:** `helpers.Error(err)` → `helpers.Error(retryErr)`

*(The correct pattern already exists at line 673 (`StoreCVESummaryStub`) in the same file — proof that `retryErr` was the intended variable).*

In `adapters/v1/backend.go`:
- **Line 280:** `helpers.Error(err)` → `helpers.Error(e)`

*(The correct pattern is in the same function — `e` is the variable immediately assigned from `ParseImageManifest` on the line above, and the `if e != nil` guard proves it was meant to be logged).*

---

## Additional Context

I can submit a PR fixing all 5 sites if you agree this is worth cleaning up. It's five one-line changes — mechanical, but it fixes a real incident-response blind spot. If you'd prefer a different approach, I'm open to suggestions!

AdeshDeshmukh:
## Overview

Two related log-correctness bugs, same root cause: a `Warning` log captures the wrong Go variable — one that's guaranteed `nil` or semantically meaningless — instead of the actual error that just occurred.

In `repositories/apiserver.go` (StoreCVE L351, StoreCVESummary L588, StoreSBOM filtered L1127, StoreSBOM unfiltered L1160) the retry-exhaustion warning logs `helpers.Error(err)` — the original `IsAlreadyExists` sentinel — instead of `helpers.Error(retryErr)` — the actual Get/Update failure. `StoreCVESummaryStub` L673 in the same file proves the intended pattern.

In `adapters/v1/backend.go` SubmitCVE L282 the manifest parse failure warning logs `helpers.Error(err)` (guaranteed `nil` at that point) instead of `helpers.Error(e)` (the actual parse error).

## Additional Information

- 5 one-line changes total (4 in apiserver.go, 1 in backend.go)
- No new imports needed
- No behavioral change — only the correct error variable is now logged

## How to Test

1. `go build ./...`
2. `go test ./repositories/...`
3. `go test ./adapters/v1/... -run TestBackend`
4. Verify no regressions: `go test ./...`

The 3 `Test_syftAdapter_CreateSBOM` failures are pre-existing (`arm64` vs `amd64` fixture mismatch, fail on clean `main` too).

## Related issues/PRs

Fixes #389

### DIFF
--- a/adapters/v1/backend.go
+++ b/adapters/v1/backend.go
@@ -279,7 +279,7 @@ func (a *BackendAdapter) SubmitCVE(ctx context.Context, cve domain.CVEManifest, 
 
 	imageManifest, e := ParseImageManifest(cve.Content)
 	if e != nil {
-		logger.L().Ctx(ctx).Warning("failed to parse image manifest from grype document", helpers.Error(err))
+		logger.L().Ctx(ctx).Warning("failed to parse image manifest from grype document", helpers.Error(e))
 	}
 
 	// merge cve and cvep
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -348,7 +348,7 @@ func (a *APIServerStore) StoreCVE(ctx context.Context, cve domain.CVEManifest, w
 			return updateErr
 		})
 		if retryErr != nil {
-			logger.L().Ctx(ctx).Warning("failed to update CVE manifest in storage", helpers.Error(err),
+			logger.L().Ctx(ctx).Warning("failed to update CVE manifest in storage", helpers.Error(retryErr),
 				helpers.String("name", cve.Name),
 				helpers.String("relevant", strconv.FormatBool(withRelevancy)))
 		} else {
@@ -585,7 +585,7 @@ func (a *APIServerStore) StoreCVESummary(ctx context.Context, cve domain.CVEMani
 			return updateErr
 		})
 		if retryErr != nil {
-			logger.L().Ctx(ctx).Warning("failed to update CVE summary manifest in storage", helpers.Error(err),
+			logger.L().Ctx(ctx).Warning("failed to update CVE summary manifest in storage", helpers.Error(retryErr),
 				helpers.String("name", cve.Name),
 				helpers.String("relevant", strconv.FormatBool(withRelevancy)))
 		} else {
@@ -1124,7 +1124,7 @@ func (a *APIServerStore) StoreSBOM(ctx context.Context, sbom domain.SBOM, isFilt
 				return updateErr
 			})
 			if retryErr != nil {
-				logger.L().Ctx(ctx).Warning("failed to update filtered SBOM in storage", helpers.Error(err),
+				logger.L().Ctx(ctx).Warning("failed to update filtered SBOM in storage", helpers.Error(retryErr),
 					helpers.String("name", sbom.Name))
 			} else {
 				logger.L().Debug("updated filtered SBOM in storage",
@@ -1157,7 +1157,7 @@ func (a *APIServerStore) StoreSBOM(ctx context.Context, sbom domain.SBOM, isFilt
 				return updateErr
 			})
 			if retryErr != nil {
-				logger.L().Ctx(ctx).Warning("failed to update SBOM in storage", helpers.Error(err),
+				logger.L().Ctx(ctx).Warning("failed to update SBOM in storage", helpers.Error(retryErr),
 					helpers.String("name", sbom.Name))
 			} else {
 				logger.L().Debug("updated SBOM in storage",


matthyx (commented):
## Review: no blockers

I verified all 5 sites against the source at `71b34db` rather than against the description. Every change is correct, and the set is complete.

**`adapters/v1/backend.go:282`** — confirmed `err` is provably `nil` at this point: the only two assignments (`GetCVEExceptions` L270, `DomainToArmo` L275) each return early on failure, so control can only reach L282 with `err == nil`. Logging `e` — the value the enclosing `if e != nil` guard tests — is the correct variable. This site was silently logging `<nil>` for every manifest parse failure.

**`repositories/apiserver.go:351, 588, 1127, 1160`** — confirmed each is inside `case errors.IsAlreadyExists(err):`, so `err` is by definition the `AlreadyExists` sentinel, and the real Get/Update failure is in `retryErr`. Logging `retryErr` under the message `"failed to update ... in storage"` is right, and it now matches `StoreCVESummaryStub:673`, which already had the correct pattern.

**Completeness** — `grep -n retryErr repositories/apiserver.go` returns 5 `retryErr != nil` log sites; all 5 now log `retryErr`. No instance of this pattern is left behind.

**No compile risk** — `err` remains live at all 4 apiserver sites via `errors.IsAlreadyExists(err)` and the sibling `case err != nil:` branch, so nothing becomes unused. Same in `backend.go`, where `err` is used before L282.

**Behavior** — log content only. No control flow, no new imports, nothing that changes what is stored or returned. CI is green (DCO, FOSSA license/security/dependency, GitGuardian).

One out-of-scope observation, explicitly **not** a blocker for this PR: at all 5 retry sites the function logs the failure and then still `return nil`, so a caller cannot distinguish "stored" from "retries exhausted". That is pre-existing behavior and orthogonal to the logging fix in #389 — worth its own issue if the maintainers agree, but it should not hold this up.

LGTM.

AdeshDeshmukh:
Thanks for the thorough review , matthyx — And thank you for the catch on the retry sites still returning nil after retries are exhausted — that's a good observation about callers not being able to tell "stored" from "failed".

I'd like to take that up as a separate issue and PR if you and the team think it's worth addressing. I'll file it as its own issue so it's tracked properly , and I'm happy to work on it whenever it's convenient...!

matthyx (approved):

Issue #390, opened 2026-07-29
fix: GetSBOM/GetCVE swallow non-NotFound apiserver errors, triggering 30s+ SBOM regeneration on transient flakes


AdeshDeshmukh:
## Description

In `repositories/apiserver.go`, both `GetCVE` (L229–237) and `GetSBOM` (L1040–1046) have the same error-handling pattern: when the K8s API call fails with any error other than `IsNotFound`, they log a warning and return `(zero, nil)` — byte-for-byte identical to the "manifest not present" path.

```go
case err != nil:
    logger.L().Ctx(ctx).Warning("failed to get CVE manifest from apiserver", helpers.Error(err),
        helpers.String("name", name))
    return domain.CVEManifest{}, nil   // ← indistinguishable from "not found"
```

The caller in `core/services/scan.go` (lines 425–436 for `ScanCVE`, lines 198–211 for `ScanCP`) consumes this exactly as a cache miss — if `cve.Content == nil` — and proceeds to re-create the SBOM from scratch. That means a full Syft image pull + cataloging: 30+ seconds per affected image, per container, for every transient apiserver timeout, throttle, or connection reset.

`MemoryStore.GetCVE` returns `domain.ErrMockError` on its `getError` branch — so the test suite proves that error-surfacing works, while production silently falls back to the expensive regenerate path. This is a classic "tests pass, prod silently wastes resources" gap.

The sibling `GetContainerProfile` (lines 213–216) in the same file already does it correctly:

```go
return ..., fmt.Errorf("failed to get container profile from apiserver: %w", err)
```

---

## Environment

- **OS:** Any
- **Version:** `main` (Jul 2026)

---

## Steps To Reproduce

1. In `repositories/apiserver.go`, read `GetCVE` lines 229–237 and `GetSBOM` lines 1040–1046.
2. Confirm the `case err != nil:` branch returns `(zero, nil)` — swallowing the transient error.
3. In `core/services/scan.go`, read how `GetCVE`'s return is consumed at lines 425–436 (`if cve.Content == nil { ... regenerate ... }`).
4. Compare against `GetContainerProfile` at lines 213–216, which correctly wraps the error.

---

## Expected Behavior

[...]

PR #396, opened 2026-07-30
fix: surface transient apiserver errors in GetCVE/GetSBOM

AdeshDeshmukh:
Refs #390

## Overview
GetCVE and GetSBOM in repositories/apiserver.go swallowed non-NotFound apiserver errors by returning (zero, nil) — byte-for-byte identical to the manifest-not-present path. This PR surfaces those errors to the caller so a transient flake is distinguishable from a genuine miss.

Note: the headline harm in #390 — a transient flake still triggers a full 30s+ SBOM regeneration because callers currently fall through on error — is NOT changed by this PR. It stays tracked in #390 for a follow-up.

### repositories/apiserver.go (2 sites)
- GetCVE (L235): returns domain.CVEManifest{}, fmt.Errorf("failed to get CVE manifest from apiserver: %w", err)
- GetSBOM (L1042): returns domain.SBOM{}, fmt.Errorf("failed to get SBOM from apiserver: %w", err)

The now-redundant repository-level Warning was removed to avoid double-logging. The removed record is not lost: all five callers in scan.go (L108/200/213/415/427) already log the returned error at Warning level with imageSlug context.

## Why this is safe
- IsNotFound branch unchanged
- Empty-name early-return unchanged
- All 5 callers already treat non-nil errors as non-fatal warnings
- Wrapped-error pattern matches GetContainerProfile (L213-216)

## Tests added
- TestAPIServerStore_GetCVE_transientError — fake clientset reactor returning apierrors.NewInternalError; fails on base 2131f54, passes here (verified)
- TestAPIServerStore_GetSBOM_transientError — same, on "sbomsyfts"

## Scope note
GetCVESummary (L282-285) has the same swallow but is intentionally left for a follow-up to keep this PR scoped. Happy to extend if preferred.

## How to Test
1. go build ./...
2. go test ./repositories/...
3. go test ./core/services/...
4. gofmt -l repositories/                       # clean
5. go vet ./repositories/...                    # clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling when CVE or SBOM data cannot be retrieved.
  * Retrieval failures now report an error instead of appearing as successful requests with empty results.
* **Tests**
  * Added coverage for transient storage failures affecting CVE and SBOM retrieval.

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -232,9 +232,7 @@ func (a *APIServerStore) GetCVE(ctx context.Context, name, SBOMCreatorVersion, C
 			helpers.String("name", name))
 		return domain.CVEManifest{}, nil
 	case err != nil:
-		logger.L().Ctx(ctx).Warning("failed to get CVE manifest from apiserver", helpers.Error(err),
-			helpers.String("name", name))
-		return domain.CVEManifest{}, nil
+		return domain.CVEManifest{}, fmt.Errorf("failed to get CVE manifest from apiserver: %w", err)
 	}
 	// discard the manifest if it was created by an older version of the scanner
 	if manifest.Annotations[helpersv1.ToolVersionMetadataKey] != SBOMCreatorVersion ||
@@ -1041,9 +1039,7 @@ func (a *APIServerStore) GetSBOM(ctx context.Context, name, SBOMCreatorVersion s
 			helpers.String("name", name))
 		return domain.SBOM{}, nil
 	case err != nil:
-		logger.L().Ctx(ctx).Warning("failed to get SBOM from apiserver", helpers.Error(err),
-			helpers.String("name", name))
-		return domain.SBOM{}, nil
+		return domain.SBOM{}, fmt.Errorf("failed to get SBOM from apiserver: %w", err)
 	}
 	// discard the manifest if it was created by an older version of the scanner
 	if semver.Compare(manifest.Spec.Metadata.Tool.Version, SBOMCreatorVersion) == -1 {
--- a/repositories/apiserver_test.go
+++ b/repositories/apiserver_test.go
@@ -2,16 +2,21 @@ package repositories
 
 import (
 	"context"
+	"fmt"
 	"testing"
 
 	helpersv1 "github.com/kubescape/k8s-interface/instanceidhandler/v1/helpers"
 	"github.com/kubescape/kubevuln/core/domain"
 	"github.com/kubescape/kubevuln/internal/tools"
 	"github.com/kubescape/storage/pkg/apis/softwarecomposition/v1beta1"
+	"github.com/kubescape/storage/pkg/generated/clientset/versioned/fake"
 	"github.com/openvex/go-vex/pkg/vex"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	k8stesting "k8s.io/client-go/testing"
 )
 
 const name = "k8s.gcr.io-kube-proxy-sha256-c1b13"
@@ -702,4 +707,26 @@ func TestMergeMaps(t *testing.T) {
 			assert.Equal(t, tt.expected, tt.existing)
 		})
 	}
+}
+
+func TestAPIServerStore_GetCVE_transientError(t *testing.T) {
+	clientset := fake.NewSimpleClientset()
+	injectedErr := apierrors.NewInternalError(fmt.Errorf("etcd timeout"))
+	clientset.PrependReactor("get", "vulnerabilitymanifests", func(k8stesting.Action) (bool, runtime.Object, error) {
+		return true, nil, injectedErr
+	})
+	a := &APIServerStore{StorageClient: clientset.SpdxV1beta1(), Namespace: "kubescape"}
+	_, err := a.GetCVE(context.TODO(), name, "", "", "")
+	require.ErrorIs(t, err, injectedErr)
+}
+
+func TestAPIServerStore_GetSBOM_transientError(t *testing.T) {
+	clientset := fake.NewSimpleClientset()
+	injectedErr := apierrors.NewInternalError(fmt.Errorf("etcd timeout"))
+	clientset.PrependReactor("get", "sbomsyfts", func(k8stesting.Action) (bool, runtime.Object, error) {
+		return true, nil, injectedErr
+	})
+	a := &APIServerStore{StorageClient: clientset.SpdxV1beta1(), Namespace: "kubescape"}
+	_, err := a.GetSBOM(context.TODO(), name, "")
+	require.ErrorIs(t, err, injectedErr)
 }


repositories/apiserver.go
```diff
@@ -234,7 +234,7 @@ func (a *APIServerStore) GetCVE(ctx context.Context, name, SBOMCreatorVersion, C
 	case err != nil:
 		logger.L().Ctx(ctx).Warning("failed to get CVE manifest from apiserver", helpers.Error(err),
 			helpers.String("name", name))
-		return domain.CVEManifest{}, nil
+		return domain.CVEManifest{}, fmt.Errorf("failed to get CVE manifest from apiserver: %w", err)
```
matthyx: Wrapping the error is the right contract change, but note this branch **already** logs the error today (`Warning` + `helpers.Error(err)`), and the not-found branch above logs at `Debug` — so a transient flake is already distinguishable from a genuine miss in the logs. After this change the same failure is logged twice per occurrence: once here, and again by the caller (`core/services/scan.go:414` / `:425` / `:199` / `:212` / `:107` all log `getting CVE` / `getting SBOM` with the returned error).

Suggest dropping this now-redundant `Warning` and letting the wrapped error be the single record — the caller has more context (imageSlug, jobID) anyway.

**This branch also needs a test.** It is testable with a reactor on the fake clientset — I verified this passes on `0c96105` and fails on base `2131f54`:

```go
func TestAPIServerStore_GetCVE_transientError(t *testing.T) {
	clientset := fake.NewSimpleClientset()
	clientset.PrependReactor("get", "vulnerabilitymanifests", func(k8stesting.Action) (bool, runtime.Object, error) {
		return true, nil, apierrors.NewInternalError(fmt.Errorf("etcd timeout"))
	})
	a := &APIServerStore{StorageClient: clientset.SpdxV1beta1(), Namespace: "kubescape"}
	_, err := a.GetCVE(context.TODO(), name, "", "", "")
	require.Error(t, err)
}
```

(imports: `k8stesting "k8s.io/client-go/testing"`, `apierrors "k8s.io/apimachinery/pkg/api/errors"`, `"k8s.io/apimachinery/pkg/runtime"`, and the storage `fake` clientset. `NewFakeAPIServerStorage` doesn't expose the clientset, hence constructing `APIServerStore` directly.)

repositories/apiserver.go
```diff
@@ -1043,7 +1043,7 @@ func (a *APIServerStore) GetSBOM(ctx context.Context, name, SBOMCreatorVersion s
 	case err != nil:
 		logger.L().Ctx(ctx).Warning("failed to get SBOM from apiserver", helpers.Error(err),
 			helpers.String("name", name))
-		return domain.SBOM{}, nil
+		return domain.SBOM{}, fmt.Errorf("failed to get SBOM from apiserver: %w", err)
```
matthyx: Same two points as `GetCVE` above: the pre-existing `Warning` here becomes a duplicate of the caller's log once the error is returned, and this branch needs the matching test (`PrependReactor("get", "sbomsyfts", ...)` — verified it passes on this branch and fails on base).

One behavioural note specific to `GetSBOM`: on a flake the caller now warns and still falls through to `sbom.Content == nil` → full Syft regeneration → `StoreSBOM`, which lands on the `IsAlreadyExists` → `Update` path and overwrites an SBOM that was actually present. That is unchanged by this PR, but it is the cost #390 is about — see the review summary.

matthyx (commented):
Reviewed at `0c96105`.

**The change itself is safe.** Matching `GetContainerProfile` (L200-219) is the right contract, and I checked every consumer: all five call sites (`core/services/scan.go` L106, L198, L211, L413, L425) only log a warning and fall through, and none of them re-tests a stale `err` afterwards — so nothing new short-circuits. Locally on this branch: `go build ./...` clean, `go test ./repositories/... ./core/services/...` pass. (Worth noting the repo's test workflow hasn't actually run on the PR — only DCO and GitGuardian have.)

Two blockers before merge, both about what the PR claims rather than what it breaks:

**1. `Fixes #390` would auto-close an issue this doesn't fix.** #390's headline harm is "transient flake ⇒ full 30s+ SBOM regeneration". That is completely unchanged here: every caller still warns and proceeds, so the flake path still runs Syft end-to-end, rescans, and then `StoreCVE`/`StoreSBOM` hits `IsAlreadyExists` → `Update` (L332-348), overwriting a manifest that was there all along. Either:
- (a) make one caller act on the error — e.g. in `ScanCVE`, retry the Get with backoff before falling through to regeneration, or return the error so the job is retried instead of paying for a rescan; or
- (b) change `Fixes #390` to `Refs #390` and keep the issue open, describing this PR as the observability half only.

**2. The stated observability gain already exists.** The PR body says the flake happens "with no log evidence of why", but both changed branches already log `Warning` with `helpers.Error(err)`, while the not-found branch logs at `Debug` — operators can distinguish the two today. The net effect of this PR on logs is that each failure is now recorded **twice** (repository + caller). If the goal is a single clear record, drop the repository-level `Warning` in the two changed branches; see the inline comments.

**3. No test for the changed contract** (inline comments have a verified snippet — fails on base `2131f54`, passes on `0c96105`). The PR description's "existing unit tests pass" is true but vacuous here: no existing test can reach the `err != nil` branch, so nothing guards the new behaviour against a future revert.

Non-blocking: `GetCVESummary` (L280-284) still has the byte-identical swallow, and #390 explicitly asked whether to include it. Leaving it out is fine, but say so in the PR body or open a follow-up, otherwise the inconsistency looks accidental.

AdeshDeshmukh:
Thanks for the thorough review, and for checking the callers too — that's a big help.

All three points are addressed: I switched to `Refs #390` since this is really just the observability half, dropped the redundant `Warning` in both branches (the log isn't lost — the callers in `scan.go` already log the error at `Warning` level), and added the two tests you suggested. I confirmed they fail on the base version and pass here, so the behavior won't silently regress.

On `GetContainerProfile` — I left its `Warning` in place because its caller just wraps and returns the error instead of logging it, so it would've been lost there. Kept `GetCVESummary` out of scope and noted it in the PR body; happy to do a follow-up if you'd like.

One question — the repo test workflow didn't run on this PR (only DCO and GitGuardian). Is there a label or step to trigger it, or is that expected?

matthyx (commented):
Re-checked at `b196e89` — all three points addressed, thanks for the quick turnaround.

Verified locally on this head:

- `go build ./...`, `gofmt -l repositories/` and `go vet ./repositories/...` are clean; `go test ./repositories/... ./core/services/...` pass.
- The two new tests are genuine regression guards, not just green: with the new test file in place I reverted `repositories/apiserver.go` to base `2131f54` and both `TestAPIServerStore_GetCVE_transientError` and `TestAPIServerStore_GetSBOM_transientError` fail with *"An error is expected but got nil"*, then pass again on `b196e89`.
- Removing the repository-level `Warning` loses no log record: all five call sites (`core/services/scan.go` L108, L200, L213, L415, L427) log the returned error at `Warning` with `imageSlug`, and the value they pass is the same string the repository was logging as `name`.
- `Refs #390` with the still-unfixed regeneration cost called out explicitly, plus the `GetCVESummary` scope note — that resolves my main objection to the previous revision.

No further blockers from me. Resolving my two (now outdated) threads.

One process note, unrelated to the code: the Go test workflow still hasn't run on this PR — the only checks are DCO, GitGuardian, FOSSA and CodeRabbit — so the CI evidence here is still my local run rather than the repo's own.

AdeshDeshmukh:
Applied the CodeRabbit nitpick locally — the transient-error tests now assert the wrapped cause via `require.ErrorIs(t, err, injectedErr)` instead of just `require.Error`. Verified: fails on base `2131f54`, passes here, full `./repositories/...` suite green, gofmt + vet clean.

I haven't pushed it to avoid churning the PR you've already cleared. Happy to push it as a follow-up commit if you'd prefer it included before merge — otherwise leaving as-is.

matthyx:
if you have it already, push it

AdeshDeshmukh:
Pushed — the tests now assert the wrapped cause via require.ErrorIs.

matthyx (approved):

PR #399, opened 2026-07-31
fix: surface storage failures in StoreCVE/StoreCVESummary/StoreSBOM

### LINKED ISSUE #398: fix: StoreCVE/StoreCVESummary/StoreSBOM always return nil, hiding storage failures from callers
# Description

`APIServerStore.StoreCVE`, `APIServerStore.StoreCVESummary`, `APIServerStore.StoreCVESummaryStub` and `APIServerStore.StoreSBOM` in `repositories/apiserver.go` always returned `nil`, even when the initial Create failed or the Create-on-conflict retry was exhausted. Storage failures were only logged as warnings, so callers could not distinguish "stored" from "storage failed".

This was flagged by @matthyx during review of #395:

> One out-of-scope observation, explicitly **not** a blocker for this PR: at all 5 retry sites the function logs the failure and then still `return nil`, so a caller cannot distinguish "stored" from "retries exhausted". That is pre-existing behavior and orthogonal to the logging fix in #389 — worth its own issue if the maintainers agree.

The concrete effect is that `GenerateSBOM` in `core/services/scan.go:127-131` has a branch that reports `scanfailure.ScanFailureBackendPost` with `ReasonSBOMStorageFailed` and returns the error — but it was dead code in production because `StoreSBOM` never returned a non-nil error. The same applied to the warn-and-continue error branches of all other Store callers in `scan.go` (L244/285/291/302/307/325/351/359/458/494/499/509/514), and to the `StoreCVESummaryStub` callers at L235 and L449.

# Environment

- `kubescape/kubevuln` at `main` (`22be3a5`)
- `repositories/apiserver.go`

# Steps To Reproduce

1. Run kubevuln against an apiserver.
2. Make the storage write fail (e.g. RBAC denies `create` on the relevant resource, or the apiserver is temporarily unavailable).
3. Observe: a `Warning` is logged (`"failed to store ... in storage"` / `"failed to update ... in storage"`), but the Store method returns `nil` and the caller proceeds as if storage succeeded — `GenerateSBOM` does **not** report the scan failure, and schema-unsupported workloads silently get no summary stub.

# Expected behavior

The Store methods return a wrapped error when the Create (or the retry-update) fails, matching the propagation pattern already used by `StoreVEX` (`apiserver.go:691-724`) and by the in-memory siblings (`repositories/memory.go:123-125, 142-144`, which return `ErrMockError`). Callers already handle non-nil errors, so `GenerateSBOM`'s `ReasonSBOMStorageFailed` branch becomes live and the other callers log the real failure.

# Actual Behavior

All methods swallow the error and `return nil`:
- `StoreCVE` — `apiserver.go:366`
- `StoreCVESummary` — `apiserver.go:603`
- `StoreCVESummaryStub` — `apiserver.go:692`
- `StoreSBOM` — `apiserver.go:1170`

Each has two return-nil-after-log sites (retry-exhausted arm and `case err != nil` arm), for a total of **10 sites**:
- `StoreCVE`: retry arm (`retryErr`) + err arm (`err`)
- `StoreCVESummary`: retry arm (`retryErr`) + err arm (`err`)
- `StoreCVESummaryStub`: retry arm (`retryErr`) + err arm (`err`)
- `StoreSBOM` filtered block: retry arm (`retryErr`) + err arm (`err`)
- `StoreSBOM` unfiltered block: retry arm (`retryErr`) + err arm (`err`)

# Additional context

- **Behavioral decision for maintainers:** making `StoreSBOM` return an error activates the `GenerateSBOM` abort branch at `scan.go:127-131` (returns the error and reports `ScanFailureBackendPost`/`ReasonSBOMStorageFailed`). The branch was written to do exactly this, so returning the error matches its intent — but if maintainers prefer the scan to continue, that single caller can be switched to warn-and-continue.
- All remaining Store callers in `scan.go` already warn-and-continue on error, so no caller changes are required for the fix.
- `StoreCVESummary` and `StoreCVESummaryStub` require a `domain.WorkloadKey{}`- and `domain.TimestampKey{}`-carrying context for their enrichment steps, which matters for tests.
- `IsNotFound` / empty-name branches are unchanged by the fix.

Proposed fix: return `fmt.Errorf("failed to ... in storage: %w", err)` / `...%w", retryErr)` at the 10 sites above, mirroring the `StoreVEX` pattern.

AdeshDeshmukh:
Fixes #398

## Overview

`APIServerStore.StoreCVE`, `APIServerStore.StoreCVESummary`, `APIServerStore.StoreCVESummaryStub`
and `APIServerStore.StoreSBOM` in `repositories/apiserver.go` always returned `nil`, even when the
initial Create failed or the Create-on-conflict retry was exhausted. Storage failures were only
logged as warnings, so callers could not distinguish "stored" from "storage failed".

This was raised by @matthyx during review of #395:

> at all 5 retry sites the function logs the failure and then still `return nil`, so a caller
> cannot distinguish "stored" from "retries exhausted". ... worth its own issue.

The concrete consequence: `GenerateSBOM` in `core/services/scan.go` (L127-131) has a live branch
that reports `scanfailure.ScanFailureBackendPost` with `ReasonSBOMStorageFailed` and returns the
error — but it was dead code in production because `StoreSBOM` never returned a non-nil error.
The same applied to the warn-and-continue error branches of all other Store callers, and to the
`StoreCVESummaryStub` callers at `scan.go:235` and `:449` (schema-unsupported workloads silently
got no stub).

### `repositories/apiserver.go` — 10 return sites across 4 functions

Each failure arm now returns a wrapped error instead of falling through to `return nil`,
matching the propagation pattern `StoreVEX` already uses (apiserver.go:691-724) and the
in-memory siblings (`MemoryStore` returns `ErrMockError`, memory.go:123-125, 142-144).

| Function | retry-exhausted arm | `case err != nil` arm |
|---|---|---|
| `StoreCVE` | `retryErr` | `err` |
| `StoreCVESummary` | `retryErr` | `err` |
| `StoreCVESummaryStub` | `retryErr` | `err` |
| `StoreSBOM` (filtered) | `retryErr` | `err` |
| `StoreSBOM` (unfiltered) | `retryErr` | `err` |

Each returns `fmt.Errorf("failed to <update/store> ... in storage: %w", err|retryErr)`.

The repository-side failure logs are demoted from `Warning` to `Debug` (no `Ctx`): the error is
now returned, so the caller owns the reported log — one record per failure instead of two.

### Callers — no change required

All Store callers in `core/services/scan.go` already handle non-nil errors:
- `GenerateSBOM` L127-131: aborts via `return err` + `ReportScanFailure(... ReasonSBOMStorageFailed)` — this branch is now live, which is the intended fix.
- All others (L235, 244, 285, 291, 302, 307, 325, 351, 359, 449, 458, 494, 499, 509, 514): warn-and-continue.

## Tests added — 14 transientError tests (`repositories/apiserver_test.go`)

Each of the 4 Store methods now covers the initial-create failure, the get-inside-update-closure
failure, and the real conflict-exhaustion path:

| Test | Arm exercised |
|---|---|
| `TestAPIServerStore_StoreCVE_transientError` | create fails |
| `TestAPIServerStore_StoreCVE_updateGetFailure_transientError` | create → AlreadyExists, get inside retry fails |
| `TestAPIServerStore_StoreCVE_retryExhausted_transientError` | create → AlreadyExists, update always Conflict → `RetryOnConflict` exhausts (`retry.DefaultRetry.Steps` attempts) |
| `TestAPIServerStore_StoreCVESummary_transientError` | create fails |
| `TestAPIServerStore_StoreCVESummary_updateGetFailure_transientError` | create → AlreadyExists, get inside retry fails |
| `TestAPIServerStore_StoreCVESummary_retryExhausted_transientError` | update always Conflict → retries exhausted |
| `TestAPIServerStore_StoreCVESummaryStub_transientError` | create fails |
| `TestAPIServerStore_StoreCVESummaryStub_retryExhausted_transientError` | update always Conflict → retries exhausted |
| `TestAPIServerStore_StoreSBOM_transientError` | create fails (unfiltered) |
| `TestAPIServerStore_StoreSBOM_updateGetFailure_transientError` | create → AlreadyExists, get inside retry fails (unfiltered) |
| `TestAPIServerStore_StoreSBOM_retryExhausted_transientError` | update always Conflict → retries exhausted (unfiltered) |
| `TestAPIServerStore_StoreSBOMFiltered_transientError` | create fails (filtered) |
| `TestAPIServerStore_StoreSBOMFiltered_updateGetFailure_transientError` | create → AlreadyExists, get inside retry fails (filtered) |
| `TestAPIServerStore_StoreSBOMFiltered_retryExhausted_transientError` | update always Conflict → retries exhausted (filtered) |

The `_retryExhausted_` tests seed the object via `fake.NewSimpleClientset(seeded)` (so the
tracker returns `AlreadyExists` on Create and the seeded object on Get), stub the `update`
reactor to always return `apierrors.NewConflict(...)`, and assert both `apierrors.IsConflict(err)`
and that the update reactor was called exactly `retry.DefaultRetry.Steps` (5) times.

All 14 tests were proven to **fail on base** (error swallowed, `nil` returned) and **pass with
this change**.

## Why this is safe

- `IsNotFound` and empty-name branches are unchanged.
- No caller changes needed — all already handle errors; the only behavioral activation is the intended `GenerateSBOM` abort branch.
- Pattern proven by `StoreVEX` and `MemoryStore`.
- No new runtime imports (`fmt` was already imported).

## How to Test

```
go build ./...
go test ./repositories/... -count=1
go test ./core/... -count=1
gofmt -l repositories/
go vet ./repositories/...
```

Local results: `repositories` 52 passed, `core` 108 passed, gofmt clean, vet clean, build clean.

## Notes for reviewers

- **Behavioral decision:** I kept `GenerateSBOM` aborting (`return err`) on storage failure,
  since that is what the existing branch at scan.go:127-131 was written to do. Happy to switch
  it to warn-and-continue if the maintainers prefer a non-fatal path.
- `StoreCVESummary` / `StoreCVESummaryStub` tests require a `domain.WorkloadKey{}`- and
  `domain.TimestampKey{}`-carrying context (already handled in the test fixtures).

## Checklist

- [x] Code style followed
- [x] Self-review completed
- [x] Tests added and passing locally
- [x] DCO signed off

## Summary by CodeRabbit

* **Bug Fixes**
  * Storage failures when creating or updating CVE and SBOM data are now reported instead of being silently ignored.
  * Error details are preserved for temporary failures and exhausted retries, improving troubleshooting and reliability.
  * Conflict retries now honor the configured retry limit.

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -346,18 +346,19 @@ func (a *APIServerStore) StoreCVE(ctx context.Context, cve domain.CVEManifest, w
 			return updateErr
 		})
 		if retryErr != nil {
-			logger.L().Ctx(ctx).Warning("failed to update CVE manifest in storage", helpers.Error(retryErr),
+			logger.L().Debug("failed to update CVE manifest in storage", helpers.Error(retryErr),
 				helpers.String("name", cve.Name),
 				helpers.String("relevant", strconv.FormatBool(withRelevancy)))
-		} else {
-			logger.L().Debug("updated CVE manifest in storage",
-				helpers.String("name", cve.Name),
-				helpers.String("relevant", strconv.FormatBool(withRelevancy)))
+			return fmt.Errorf("failed to update CVE manifest in storage: %w", retryErr)
 		}
-	case err != nil:
-		logger.L().Ctx(ctx).Warning("failed to store CVE manifest in storage", helpers.Error(err),
+		logger.L().Debug("updated CVE manifest in storage",
 			helpers.String("name", cve.Name),
 			helpers.String("relevant", strconv.FormatBool(withRelevancy)))
+	case err != nil:
+		logger.L().Debug("failed to store CVE manifest in storage", helpers.Error(err),
+			helpers.String("name", cve.Name),
+			helpers.String("relevant", strconv.FormatBool(withRelevancy)))
+		return fmt.Errorf("failed to store CVE manifest in storage: %w", err)
 	default:
 		logger.L().Debug("stored CVE manifest in storage",
 			helpers.String("name", cve.Name),
@@ -583,18 +584,19 @@ func (a *APIServerStore) StoreCVESummary(ctx context.Context, cve domain.CVEMani
 			return updateErr
 		})
 		if retryErr != nil {
-			logger.L().Ctx(ctx).Warning("failed to update CVE summary manifest in storage", helpers.Error(retryErr),
+			logger.L().Debug("failed to update CVE summary manifest in storage", helpers.Error(retryErr),
 				helpers.String("name", cve.Name),
 				helpers.String("relevant", strconv.FormatBool(withRelevancy)))
-		} else {
-			logger.L().Debug("updated CVE summary manifest in storage",
-				helpers.String("name", cve.Name),
-				helpers.String("relevant", strconv.FormatBool(withRelevancy)))
+			return fmt.Errorf("failed to update CVE summary manifest in storage: %w", retryErr)
 		}
-	case err != nil:
-		logger.L().Ctx(ctx).Warning("failed to store CVE summary manifest in storage", helpers.Error(err),
+		logger.L().Debug("updated CVE summary manifest in storage",
 			helpers.String("name", cve.Name),
 			helpers.String("relevant", strconv.FormatBool(withRelevancy)))
+	case err != nil:
+		logger.L().Debug("failed to store CVE summary manifest in storage", helpers.Error(err),
+			helpers.String("name", cve.Name),
+			helpers.String("relevant", strconv.FormatBool(withRelevancy)))
+		return fmt.Errorf("failed to store CVE summary manifest in storage: %w", err)
 	default:
 		logger.L().Debug("stored CVE summary manifest in storage",
 			helpers.String("name", manifest.Name),
@@ -668,18 +670,19 @@ func (a *APIServerStore) StoreCVESummaryStub(ctx context.Context, status string)
 			return updateErr
 		})
 		if retryErr != nil {
-			logger.L().Ctx(ctx).Warning("failed to update CVE summary stub in storage", helpers.Error(retryErr),
+			logger.L().Debug("failed to update CVE summary stub in storage", helpers.Error(retryErr),
 				helpers.String("name", manifest.Name),
 				helpers.String("status", status))
-		} else {
-			logger.L().Debug("updated CVE summary stub in storage",
-				helpers.String("name", manifest.Name),
-				helpers.String("status", status))
+			return fmt.Errorf("failed to update CVE summary stub in storage: %w", retryErr)
 		}
-	case err != nil:
-		logger.L().Ctx(ctx).Warning("failed to store CVE summary stub in storage", helpers.Error(err),
+		logger.L().Debug("updated CVE summary stub in storage",
 			helpers.String("name", manifest.Name),
 			helpers.String("status", status))
+	case err != nil:
+		logger.L().Debug("failed to store CVE summary stub in storage", helpers.Error(err),
+			helpers.String("name", manifest.Name),
+			helpers.String("status", status))
+		return fmt.Errorf("failed to store CVE summary stub in storage: %w", err)
 	default:
 		logger.L().Debug("stored CVE summary stub in storage",
 			helpers.String("name", manifest.Name),
@@ -1137,15 +1140,16 @@ func (a *APIServerStore) StoreSBOM(ctx context.Context, sbom domain.SBOM, isFilt
 				return updateErr
 			})
 			if retryErr != nil {
-				logger.L().Ctx(ctx).Warning("failed to update filtered SBOM in storage", helpers.Error(retryErr),
+				logger.L().Debug("failed to update filtered SBOM in storage", helpers.Error(retryErr),
 					helpers.String("name", sbom.Name))
-			} else {
-				logger.L().Debug("updated filtered SBOM in storage",
-					helpers.String("name", sbom.Name))
+				return fmt.Errorf("failed to update filtered SBOM in storage: %w", retryErr)
 			}
-		case err != nil:
-			logger.L().Ctx(ctx).Warning("failed to store filtered SBOM in storage", helpers.Error(err),
+			logger.L().Debug("updated filtered SBOM in storage",
 				helpers.String("name", sbom.Name))
+		case err != nil:
+			logger.L().Debug("failed to store filtered SBOM in storage", helpers.Error(err),
+				helpers.String("name", sbom.Name))
+			return fmt.Errorf("failed to store filtered SBOM in storage: %w", err)
 		default:
 			logger.L().Debug("stored filtered SBOM in storage",
 				helpers.String("name", sbom.Name))
@@ -1170,15 +1174,16 @@ func (a *APIServerStore) StoreSBOM(ctx context.Context, sbom domain.SBOM, isFilt
 				return updateErr
 			})
 			if retryErr != nil {
-				logger.L().Ctx(ctx).Warning("failed to update SBOM in storage", helpers.Error(retryErr),
+				logger.L().Debug("failed to update SBOM in storage", helpers.Error(retryErr),
 					helpers.String("name", sbom.Name))
-			} else {
-				logger.L().Debug("updated SBOM in storage",
-					helpers.String("name", sbom.Name))
+				return fmt.Errorf("failed to update SBOM in storage: %w", retryErr)
 			}
-		case err != nil:
-			logger.L().Ctx(ctx).Warning("failed to store SBOM in storage", helpers.Error(err),
+			logger.L().Debug("updated SBOM in storage",
 				helpers.String("name", sbom.Name))
+		case err != nil:
+			logger.L().Debug("failed to store SBOM in storage", helpers.Error(err),
+				helpers.String("name", sbom.Name))
+			return fmt.Errorf("failed to store SBOM in storage: %w", err)
 		default:
 			logger.L().Debug("stored SBOM in storage",
 				helpers.String("name", sbom.Name))
--- a/repositories/apiserver_test.go
+++ b/repositories/apiserver_test.go
@@ -16,7 +16,9 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	k8stesting "k8s.io/client-go/testing"
+	"k8s.io/client-go/util/retry"
 )
 
 const name = "k8s.gcr.io-kube-proxy-sha256-c1b13"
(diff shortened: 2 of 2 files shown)


repositories/apiserver.go
```diff
@@ -595,6 +598,7 @@ func (a *APIServerStore) StoreCVESummary(ctx context.Context, cve domain.CVEMani
 		logger.L().Ctx(ctx).Warning("failed to store CVE summary manifest in storage", helpers.Error(err),
 			helpers.String("name", cve.Name),
 			helpers.String("relevant", strconv.FormatBool(withRelevancy)))
+		return fmt.Errorf("failed to store CVE summary manifest in storage: %w", err)
```
matthyx, line 599: **Blocker (scope): `StoreCVESummaryStub` is the 5th retry site and still swallows.**

`StoreCVESummaryStub` (`repositories/apiserver.go:622-692`) has byte-for-byte the same bug this PR fixes, at two sites, and is untouched:

- `apiserver.go:674-677` — retry arm: logs `"failed to update CVE summary stub in storage"`, falls through to `return nil` at :692
- `apiserver.go:683-686` — `case err != nil`: logs `"failed to store CVE summary stub in storage"`, falls through to `return nil` at :692

That is the 5th of the "all 5 retry sites" my original comment in #395 referred to — #398's body enumerates only the 3 methods, so the PR is complete against the issue text, but not against the observation the issue quotes.

It matters concretely, not just for symmetry: **both callers already check the error and it is dead code**, exactly like `GenerateSBOM` was.

```go
// core/services/scan.go:235 and :449
if stubErr := s.cveRepository.StoreCVESummaryStub(ctx, helpersv1.UnsupportedSchema); stubErr != nil {
    logger.L().Ctx(ctx).Warning("storing schema-unsupported summary stub", helpers.Error(stubErr), ...)
}
```

For `APIServerStore`, `stubErr` can only ever be non-nil from the four enrichment/name helpers that return early (`enrichSummaryManifestObjectAnnotations`, `enrichSummaryManifestObjectLabels`, `GetCVESummaryK8sResourceName`, `GetCVESummaryK8sResourceNamespace`). A genuine storage failure returns `nil` — so a schema-unsupported workload silently gets no stub, which defeats the whole point of the stub ("so UIs do not show the workload as silently unscanned", per the comment at scan.go:443-445).

Two lines, mirroring what you already did here:

```go
if retryErr != nil {
    logger.L().Ctx(ctx).Warning("failed to update CVE summary stub in storage", ...)
    return fmt.Errorf("failed to update CVE summary stub in storage: %w", retryErr)
}
...
case err != nil:
    logger.L().Ctx(ctx).Warning("failed to store CVE summary stub in storage", ...)
    return fmt.Errorf("failed to store CVE summary stub in storage: %w", err)
```

Note the existing `TestAPIServerStore_StoreCVESummaryStub` (`apiserver_test.go:509`) has no transient-error subtest, so this would need one of the same shape as the ones added here.

If you'd rather keep this PR scoped exactly to #398's enumeration, that's fine — but then please open the follow-up before merging so the inconsistency doesn't sit unexplained in the file.

repositories/apiserver_test.go
```diff
@@ -883,3 +884,123 @@ func TestAPIServerStore_GetSBOM_transientError(t *testing.T) {
...
+		ContainerName: "anyJobContName",
+	}
+	ctx := context.WithValue(context.TODO(), domain.WorkloadKey{}, workload)
+	ctx = context.WithValue(ctx, domain.TimestampKey{}, int64(1734957372))
+	cveManifest := tools.FileToCVEManifest("testdata/nginx-cve.json")
+	err := a.StoreCVESummary(ctx, cveManifest, domain.CVEManifest{}, false)
+	require.ErrorIs(t, err, injectedErr)
+}
+
+func TestAPIServerStore_StoreCVE_retryExhausted_transientError(t *testing.T) {
```
matthyx: **Should fix: these four `_retryExhausted_` tests do not exhaust retries — they don't retry at all.**

Confirming CodeRabbit's finding with a measurement rather than by inspection. I instrumented the `get` reactor with a counter and ran this exact test shape:

```
GET-FAILURE attempts=1 err=failed to update CVE manifest in storage: Internal error occurred: etcd timeout
```

**One** attempt, not `retry.DefaultRetry.Steps` (5). The reason: `retry.RetryOnConflict` is `retry.OnError(backoff, apierrors.IsConflict, fn)`, and the injected `apierrors.NewInternalError(...)` is not a conflict — so `OnError` returns it on the first pass without ever looping. The four tests are really "Get inside the update closure failed", which is a different thing from retry exhaustion.

This is a fidelity issue, not a coverage hole — the arm you changed (`if retryErr != nil { return ... }`) *is* executed, and all four tests do fail on base and pass here, so the change is genuinely covered. But:

1. The names claim something the tests don't do, which will mislead the next person who reads them.
2. The actual conflict-exhaustion path — `Update` returning `Conflict` five times until the backoff is spent — has no coverage anywhere.

Suggested fix, both parts:

- Rename to something honest, e.g. `TestAPIServerStore_StoreCVE_updateGetFailure_transientError`.
- Add one real exhaustion test per method: let `get` fall through to the tracker (`return false, nil, nil`) after seeding the object, and have the `update` reactor always return `apierrors.NewConflict(...)`. Then assert `apierrors.IsConflict(err)` and that the update reactor was called exactly `retry.DefaultRetry.Steps` times — that pins both the propagation *and* the retry policy.

Heads up if you write it: seeding via a first `StoreCVE` call won't work once the `create` reactor is stubbed to return `AlreadyExists`, so seed the object through `clientset.Tracker().Add(...)` (or `NewSimpleClientset(obj)`) instead. I hit exactly that when probing this — my throwaway conflict test ended up asserting a `not found` from the tracker rather than a conflict.

repositories/apiserver.go
```diff
@@ -349,6 +349,7 @@ func (a *APIServerStore) StoreCVE(ctx context.Context, cve domain.CVEManifest, w
 			logger.L().Ctx(ctx).Warning("failed to update CVE manifest in storage", helpers.Error(retryErr),
 				helpers.String("name", cve.Name),
 				helpers.String("relevant", strconv.FormatBool(withRelevancy)))
+			return fmt.Errorf("failed to update CVE manifest in storage: %w", retryErr)
```
matthyx, line 352: **Nit (non-blocking): the failure is now reported twice to the telemetry backend.**

Each of the 8 sites keeps its `logger.L().Ctx(ctx).Warning(...)` *and* returns the error — and every caller then logs it again with `Ctx(ctx)` (`scan.go:246`, `287`, `293`, ...). `Ctx()` binds the record to the OTEL span, so one storage failure now emits two log records with near-identical text instead of one. `GenerateSBOM` makes it three: repository warning + `ReportScanFailure` + `controllers/http.go:69` error log.

Harmless per-event, but during an apiserver or etcd outage every in-flight scan multiplies its telemetry, which is when you least want the extra volume.

Since the message is already carried verbatim in the wrapped error, the repository-side `Ctx` warning is now redundant — worth demoting these to `logger.L().Debug(...)` (no `Ctx`) and letting the caller own the reported log. Your call; it's a separate cleanup and I would not hold the PR for it.

**Second nit, same line:** now that this branch ends in `return`, the `} else {` at :353 is dead weight — `revive`'s `indent-error-flow` would want the `else` dropped and the debug log de-indented. Applies to all four retry arms. Cosmetic only; the repo has no `.golangci.yml`, so nothing will fail on it.

matthyx (commented):
Reviewed the change against `main` (`22be3a5`) and ran the full gate locally on the head commit (`5059e30`):

| Check | Result |
|---|---|
| `go build ./...` | clean |
| `go test ./repositories/... -count=1` | ok (0.29s) |
| `go test ./core/... -count=1` | ok (7.04s) |
| `go vet ./repositories/... ./core/...` | clean |
| `gofmt -l repositories/` | clean |

(`gofmt -l core/` flags `core/ports/providers.go`, but that is already unformatted on `main` — pre-existing, not this PR.)

## Assessment

The production change is correct and minimal: 8 additive lines, zero deletions, `IsNotFound` and empty-name early returns untouched. It matches the `StoreVEX` propagation pattern.

I independently verified the two claims the PR rests on, rather than taking the description at face value:

- **Caller inventory is complete.** `StoreCVE` / `StoreCVESummary` / `StoreSBOM` have no production callers outside `core/services/scan.go`, and all of them except `GenerateSBOM` warn-and-continue. Confirmed.
- **The `GenerateSBOM` activation is contained.** `controllers/http.go:63` writes `200 OK` to the client *before* `workerPool.Submit`, so the newly live `return err` at `scan.go:127-131` cannot change any HTTP response — it adds one `ReportScanFailure(ScanFailureBackendPost, ReasonSBOMStorageFailed)` plus an error log on a background worker. Keeping the abort is the right call; no need to switch it to warn-and-continue.

I also checked for a stale-`err` hazard, since `err` is reused across the `ScanCP` / `ScanCVE` loop bodies and `Store*` previously always reset it to `nil`. Every subsequent read of `err` is preceded by a reassignment (`ScanSBOM`, `filterSBOM`, `SubmitCVE`), so no downstream `if err != nil` can misfire on a leftover storage error. No regression there.

## Findings

**1. Blocker (scope) — `StoreCVESummaryStub` still swallows, and it is the 5th retry site.** See inline comment. This is the same bug in the same file with a caller that already checks the error; leaving it means #398 is only 4/5 fixed and the next reader has to rediscover why one of the five is different. Either fold it in here or open the follow-up explicitly.

**2. Should fix — the four `_retryExhausted_transientError` tests never retry.** CodeRabbit flagged this; I confirmed it empirically (instrumented reactor: **1 attempt**, not 5). Details inline. The changed lines *are* covered, so this is test-fidelity, not a coverage hole.

Two non-blocking nits inline as well.

matthyx (commented):
Re-reviewed at `21ceb151` (force-pushed over `5059e30`). **All four points from my previous review are addressed, and the fixes are correct.** Verified rather than taken on trust:

| Check | Result |
|---|---|
| `go build ./...` | clean |
| `go vet ./repositories/... ./core/...` | clean |
| `gofmt -l repositories/` | clean |
| `go test ./repositories/... -count=1` | ok (0.53s) |
| `go test ./core/... -count=1` | ok (4.36s) |

### 1. Blocker resolved — `StoreCVESummaryStub` folded in

Now 10 propagating sites across 4 methods (was 8 across 3). The stub's retry arm and `case err != nil` arm both return wrapped errors, so the `stubErr != nil` checks at `scan.go:236` and `:450` are live and a schema-unsupported workload no longer silently loses its stub.

### 2. Test fidelity resolved — and the exhaustion tests are the real thing

The four misnamed tests are now `_updateGetFailure_transientError`, which is what they actually exercise. Five genuine conflict-exhaustion tests were added, and they're built the right way: seeded via `fake.NewSimpleClientset(seeded)` with no `create` reactor, so `AlreadyExists` comes from the tracker and `Get` falls through naturally — that sidesteps the seeding trap I flagged. Asserting both `apierrors.IsConflict(err)` and `updateCalls == retry.DefaultRetry.Steps` pins the retry policy as well as the propagation, which is more than I asked for.

I checked the suite actually guards the change rather than just passing: **all 14 new tests pass at `21ceb151`, and all 14 fail** when `repositories/apiserver.go` is reverted to `22be3a5` underneath them. No free-riders.

### 3 & 4. Nits resolved

Redundant `else` gone at all five retry arms; the eight `Ctx().Warning` calls demoted to `Debug`.

On the demotion I checked for a telemetry hole, since `Debug` is off in production and a silently-dropped failure would be worse than the double-log it replaces. There isn't one: all 15 `Store*` call sites in `scan.go` log the returned error with `Ctx().Warning`, and `GenerateSBOM`'s path reports through `ReportScanFailure` plus the error log at `controllers/http.go:69`. The only thing lost from production telemetry is the `name` / `relevant` / `status` fields the repository-side log carried; the caller logs `imageSlug` / `instanceID` instead, which is sufficient to locate the object. Fine by me.

**LGTM.** Nothing blocking left.

---

### Two things for later, neither blocking this PR

**a. Trivial gap:** `StoreCVESummaryStub` has `_transientError` and `_retryExhausted_` but no `_updateGetFailure_` variant, so it's 4/5 on that shape. The arm is already covered by its exhaustion test — purely cosmetic symmetry, ignore if you like.

**b. Pre-existing panic in `mergeMaps` — worth its own issue.** Not introduced here and not this PR's job to fix, but I found it while checking why the new tests seed `Annotations: map[string]string{}` explicitly. `mergeMaps` (`apiserver.go:1207-1211`) writes into `existing` without a nil check:

```go
func mergeMaps(existing, new map[string]string) {
	for k, v := range new {
		existing[k] = v   // panics if existing == nil
	}
}
```

Every retry closure calls `mergeMaps(result.Annotations, manifest.Annotations)` on an object fetched from storage. If that stored object has nil `Annotations` (or nil `Labels`) and the incoming manifest has any, it panics. Confirmed on this branch with a throwaway probe — seed a `VulnerabilityManifest` with nil `ObjectMeta.Annotations`, call `StoreCVE` with non-empty `cve.Annotations`:

```
PANICKED: assignment to entry in nil map
```

`StoreCVE` is the exposed one, since `manifest.Annotations = cve.Annotations` is passed through unguarded (unlike `StoreSBOM`, which always seeds at least `StatusMetadataKey`). Reachable for any manifest written with no annotations — by an older kubevuln, or by another component. The fix is a one-liner guard in `mergeMaps`, but it wants its own issue and its own test rather than being smuggled into this one.

Happy to open that issue if you'd like.

matthyx (approved):

AdeshDeshmukh:
Sure Sir I would like to !

matthyx:
Opened the follow-up for the pre-existing `mergeMaps` nil-map panic I mentioned above: #402. Independent of this PR — `mergeMaps` is untouched here, and #399 stays as-is.

PR #401, opened 2026-07-31
fix: reset VEX statements to not_affected before reapplying update

### LINKED ISSUE #400: Bug: VEX update path never restores statements from affected back to not_affected
# Description
`markRelevantVulnerabilitiesAsAffectedInVex` only flips matching statements from `not_affected` to `affected`. During updates, `APIServerStore.updateVEX` reuses the previously stored VEX document (`vexDoc := vexContainer.Spec`) and never reinitializes statement status before applying the latest filtered manifest. A CVE/package pair that was relevant in an earlier scan but is no longer relevant in a later scan stays permanently marked `affected`, even though the vulnerable component is no longer loaded.

# Environment
OS: `Any`
Version: `upstream/main @ 22be3a5`

# Steps To Reproduce
2. Read `updateVEX` and `markRelevantVulnerabilitiesAsAffectedInVex`.
3. Observe that `updateVEX` starts from `vexContainer.Spec` (the previously stored document) and calls `markRelevantVulnerabilitiesAsAffectedInVex` without first resetting existing statements to a baseline `not_affected` state.
4. Compare that with `createVEX`, which always starts from a freshly built, all-`not_affected` document.
5. Store a VEX document where a CVE/package pair is relevant (`affected`), then store an update where the filtered manifest no longer contains that CVE/package pair; the statement remains `affected`.

# Expected behavior
Each VEX update should reflect the current filtered manifest only. Statements that are no longer relevant should return to `not_affected` with the corresponding justification.

# Actual Behavior
Once a statement becomes `affected`, later updates do not clear it back to `not_affected` when the CVE/package pair disappears from the filtered manifest.

# Additional context
The existing update test in `repositories/apiserver_test.go` (`TestAPIServerStore_storeVEX`) only exercises the case where the second filtered manifest contains one additional relevant CVE. It does not cover the reverse transition (a CVE becoming no longer relevant).

rootp1:
## Overview
Fixes the VEX update path so statements are reset to `not_affected` before the current filtered manifest is reapplied. Previously, `updateVEX` reused the stored document's statement status and `markRelevantVulnerabilitiesAsAffectedInVex` only ever promoted matches to `affected`, so a CVE/package pair that stopped being relevant in a later scan stayed marked `affected` forever.

Resolved #400

## Summary by CodeRabbit

* **Bug Fixes**
  * VEX statements now correctly reset to “not affected” when previously relevant vulnerabilities are no longer present in subsequent scans.
  * Prevents outdated “affected” statuses from persisting after vulnerability data changes.

* **Tests**
  * Added coverage to verify status restoration during successive VEX updates.

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -916,6 +916,16 @@ func (a *APIServerStore) updateVEX(ctx context.Context, cve domain.CVEManifest, 
 		}
 	}
 
+	// Reset every statement back to the baseline "not affected" status before
+	// reapplying the current filtered manifest. Without this, a statement that
+	// was marked "affected" during an earlier update stays "affected" forever,
+	// even after the corresponding CVE/package pair is no longer relevant.
+	for i := range vexDoc.Statements {
+		vexDoc.Statements[i].Status = v1beta1.Status(vex.StatusNotAffected)
+		vexDoc.Statements[i].Justification = v1beta1.Justification(vex.VulnerableCodeNotPresent)
+		vexDoc.Statements[i].ImpactStatement = "Vulnerable component is not loaded into the memory"
+	}
+
 	// Now change the status of the filtered vulnerabilities to "Affected"
 	err := markRelevantVulnerabilitiesAsAffectedInVex(&vexDoc, &cvep)
 	if err != nil {
--- a/repositories/apiserver_test.go
+++ b/repositories/apiserver_test.go
@@ -352,6 +352,61 @@ func TestAPIServerStore_storeVEX(t *testing.T) {
 	assert.Equal(t, relevant+1, relevant2)
 }
 
+// TestAPIServerStore_storeVEX_updateRestoresNotAffected guards against a regression where
+// updateVEX only ever promoted statements to "affected" and never reset them back to
+// "not_affected" once the corresponding CVE/package pair stopped being relevant.
+func TestAPIServerStore_storeVEX_updateRestoresNotAffected(t *testing.T) {
+	cveManifest := tools.FileToCVEManifest("testdata/nginx-cve.json")
+	cveManifestFiltered := tools.FileToCVEManifest("testdata/nginx-cve-filtered.json")
+	require.NotEmpty(t, cveManifestFiltered.Content.Matches)
+
+	a := NewFakeAPIServerStorage("kubescape")
+
+	ctx := context.TODO()
+	workload := domain.ScanCommand{
+		ImageHash:     "sha256:32fdf92b4e986e109e4db0865758020cb0c3b70d6ba80d02fe87bad5cc3dc228",
+		InstanceID:    "apiVersion-apps/v1/namespace-kubescape/kind-ReplicaSet/name-kubevuln-65bfbfdcdd/containerName-kubevuln",
+		Wlid:          "wlid://cluster-aaa/namespace-anyNamespaceJob/job-anyJob",
+		ImageTag:      "registry.k8s.io/coredns/coredns:v1.10.1",
+		ContainerName: "anyJobContName",
+	}
+	ctx = context.WithValue(ctx, domain.WorkloadKey{}, workload)
+
+	// First store makes the filtered CVEs "affected".
+	require.NoError(t, a.StoreVEX(ctx, cveManifest, cveManifestFiltered, false))
+
+	vexContainer, err := a.StorageClient.OpenVulnerabilityExchangeContainers(a.Namespace).Get(context.Background(), cveManifest.Name, metav1.GetOptions{})
+	require.NoError(t, err)
+
+	affectedBefore := 0
+	for _, stmt := range vexContainer.Spec.Statements {
+		if stmt.Status == v1beta1.Status(vex.StatusAffected) {
+			affectedBefore++
+		}
+	}
+	assert.Equal(t, len(cveManifestFiltered.Content.Matches), affectedBefore)
+
+	// Second store with an empty filtered manifest: none of the CVEs are relevant anymore,
+	// so every previously "affected" statement should revert to "not_affected".
+	emptyFiltered := cveManifestFiltered
+	emptyContent := *cveManifestFiltered.Content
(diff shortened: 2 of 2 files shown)


rootp1:
hello @matthyx Please have a look whenever you have a chance to

repositories/apiserver.go
```diff
@@ -916,6 +916,16 @@ func (a *APIServerStore) updateVEX(ctx context.Context, cve domain.CVEManifest,
...
 
+	// Reset every statement back to the baseline "not affected" status before
+	// reapplying the current filtered manifest. Without this, a statement that
+	// was marked "affected" during an earlier update stays "affected" forever,
+	// even after the corresponding CVE/package pair is no longer relevant.
+	for i := range vexDoc.Statements {
+		vexDoc.Statements[i].Status = v1beta1.Status(vex.StatusNotAffected)
+		vexDoc.Statements[i].Justification = v1beta1.Justification(vex.VulnerableCodeNotPresent)
+		vexDoc.Statements[i].ImpactStatement = "Vulnerable component is not loaded into the memory"
+	}
```
matthyx, line 927: **Non-blocking — scope of the reset vs. statements that are no longer in the image at all.**

`updateVEX` only ever *appends* statements; nothing in this function (or anywhere else I could find) ever removes a statement whose CVE/package pair has dropped out of `cve.Content.Matches`. Since the VEX document is keyed on the instance-ID slug (`filterSBOM` -> `instanceID.GetSlug(false)`), a workload whose pod name is stable across an image change (StatefulSet/DaemonSet) keeps accumulating statements from every image it has ever run.

Before this PR those orphaned statements just kept whatever stale status they had. Now they get affirmatively re-stamped every scan with `not_affected` + `"Vulnerable component is not loaded into the memory"` — a positive claim about a package that is not in the scanned image at all, refreshed with a new `LastUpdated`/`Version` each time.

Resetting them is still *less* wrong than leaving them `affected`, so I would not hold the PR for this. But it may be worth either scoping this loop to statements present in the current `cve.Content.Matches`, or opening a follow-up to prune orphaned statements.

**Nit:** the baseline triple is now written out three times in this file (`createVEX` ~L821, the append branch above ~L912, and here). A tiny helper would keep them from drifting:

```go
func resetStatementToNotAffected(s *v1beta1.Statement) {
	s.Status = v1beta1.Status(vex.StatusNotAffected)
	s.Justification = v1beta1.Justification(vex.VulnerableCodeNotPresent)
	s.ImpactStatement = notLoadedImpactStatement
}
```

repositories/apiserver_test.go
```diff
@@ -352,6 +352,61 @@ func TestAPIServerStore_storeVEX(t *testing.T) {
...
+	vexContainer, err = a.StorageClient.OpenVulnerabilityExchangeContainers(a.Namespace).Get(context.Background(), cveManifest.Name, metav1.GetOptions{})
+	require.NoError(t, err)
+
+	affectedAfter := 0
+	for _, stmt := range vexContainer.Spec.Statements {
+		if stmt.Status == v1beta1.Status(vex.StatusAffected) {
+			affectedAfter++
+		}
+	}
+	assert.Equal(t, 0, affectedAfter, "statements that are no longer relevant should be reset to not_affected")
```
matthyx, line 407: **Non-blocking — consider covering the partial transition too.**

This case ("*all* relevance disappears") is the right primary guard and, as noted, it does fail without the fix. The scenario in #400 that is closest to production is the partial one though: CVE/package pair A stays relevant while pair B stops being relevant, and the assertion is that A is still `affected` **and** B has reverted.

That variant is strictly stronger — it pins down that the reset happens *before* `markRelevantVulnerabilitiesAsAffectedInVex` rather than after, without relying on `TestAPIServerStore_storeVEX` to catch the ordering indirectly. Cheap to add here: build the second filtered manifest from a single element of `cveManifestFiltered.Content.Matches` instead of `nil`.

matthyx (commented):
### Review summary — no blockers found

I reproduced the bug and verified the fix locally on `47b1fc6` (Go 1.25.8):

- `go build ./...`, `go vet ./repositories/... ./core/...` clean; `go test ./...` fully green.
- **The new test is a genuine regression guard.** I deleted only the 9-line reset block from `updateVEX` and re-ran: `TestAPIServerStore_storeVEX_updateRestoresNotAffected` fails at `apiserver_test.go:407` while `TestAPIServerStore_storeVEX` and `TestAPIServerStore_storeVEX_updatePreservesFieldMapping` still pass. So the guard is load-bearing and not redundant with existing coverage.
- The reset restores exactly the three fields `markRelevantVulnerabilitiesAsAffectedInVex` mutates (`Status`, `Justification`, `ImpactStatement`), and it is correctly placed *after* the append loop and *before* the marking call, so re-marking still works — confirmed by `TestAPIServerStore_storeVEX` (`relevant+1 == relevant2`) continuing to pass.

The change is minimal, matches `createVEX`'s baseline, and does what #400 asks for. Nothing here should block merge. Two things I would like a maintainer to consciously ACK, plus two nits, below.

#### Worth an explicit maintainer ACK: VEX status is no longer sticky

This is a real behavioural change beyond "fix the stuck `affected`", and it is visible to VEX consumers:

`GetContainerRelevancyScans` accepts partial container profiles when `PartialRelevancy` is enabled (`config/config.go:53`, `adapters/v1/container_profile.go:38` — profiles in `Learning` status are scanned). Relevancy grows monotonically *within* one container profile, so the reset is a no-op there. But when a container profile is recreated — pod restart, node-agent restart — relevancy restarts near-empty for the same instance-ID slug, i.e. the same VEX document name. With this change, statements that were `affected` flip back to `not_affected` and then flip forward again as the profile refills, bumping `Metadata.Version` and `LastUpdated` each time.

That is arguably exactly what #400 asks for ("each VEX update should reflect the current filtered manifest only"), and I am not suggesting a code change — just flagging that downstream consumers diffing VEX documents will now see oscillation where the old behaviour was monotonic. Better to decide that on purpose than discover it in the field.

matthyx (approved):

PR #404, opened 2026-07-31
fix: set action_statement on affected VEX statements

### LINKED ISSUE #403: Bug: generated affected VEX statements omit action_statement required by the local VEX schema comments
# Description
When kubevuln marks a VEX statement as `affected`, it clears `justification` and sets `impact_statement`, but it never sets `action_statement`. The storage API type comments in the vendored `github.com/kubescape/storage` module explicitly state that an affected VEX statement must include an action statement.

# Environment
OS: `Any`

# Steps To Reproduce
1. Open `github.com/kubescape/storage` module, `pkg/apis/softwarecomposition/v1beta1/types.go`, and read the `Statement` comment describing the `affected` status requirements.
2. Open `repositories/apiserver.go` and inspect `markRelevantVulnerabilitiesAsAffectedInVex`.
3. Observe that kubevuln sets `Status = affected` but leaves `ActionStatement` empty.

# Expected behavior
Affected statements should carry an `action_statement` so the emitted VEX is self-consistent with the local schema contract and provides remediation context.

# Actual Behavior
Affected statements are emitted without any action statement.

# Additional context
This is especially important if kubevuln starts ingesting and joining external VEX, because downstream consumers will compare internally generated and externally supplied VEX documents.

rootp1:
## Overview
markRelevantVulnerabilitiesAsAffectedInVex set a VEX statement's status to affected but never set ActionStatement, even though the vendored github.com/kubescape/storage API type comments require an action statement for that status. This PR sets ActionStatement (and clears ImpactStatement) when a statement is marked affected, and resets ActionStatement back to empty in the same reset-to-not_affected loop introduced by #401 so a statement that stops being relevant does not keep a stale action statement.

Resolved #403

## How to Test
- go test ./repositories/...

## Related issues/PRs:
* Resolved #403

## Summary by CodeRabbit

* **New Features**
  * Added remediation guidance to affected vulnerability statements, using known fixed versions when available or a default upgrade recommendation.
  * Existing impact details are cleared when remediation guidance is provided.
  * Remediation guidance is automatically cleared when a statement returns to a not-affected status.

* **Bug Fixes**
  * Improved VEX statement updates to keep remediation and impact information consistent with the current assessment.

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -746,6 +746,20 @@ func createProductStructForImageAndPackage(imagePullable string, packagePURL str
 	return &product, nil
 }
 
+// defaultActionStatement is used when a match does not carry enough fix data to build a
+// more specific remediation string.
+const defaultActionStatement = "Upgrade the vulnerable component to a version that is not affected"
+
+// buildActionStatement returns a remediation string for an affected VEX statement. When the
+// match reports a fixed state with known fix versions, it names them; otherwise it falls back
+// to defaultActionStatement.
+func buildActionStatement(v v1beta1.Match) string {
+	if v.Vulnerability.Fix.State == "fixed" && len(v.Vulnerability.Fix.Versions) > 0 {
+		return fmt.Sprintf("Upgrade %s to version %s", v.Artifact.PURL, strings.Join(v.Vulnerability.Fix.Versions, " or "))
+	}
+	return defaultActionStatement
+}
+
 func markRelevantVulnerabilitiesAsAffectedInVex(vexDoc *v1beta1.VEX, cvep *domain.CVEManifest) error {
 	// Now change the status of the filtered vulnerabilities to "Affected"
 	for _, v := range cvep.Content.Matches {
@@ -757,8 +771,8 @@ func markRelevantVulnerabilitiesAsAffectedInVex(vexDoc *v1beta1.VEX, cvep *domai
 						if sc.ID == v.Artifact.PURL {
 							vexDoc.Statements[i].Status = v1beta1.Status(vex.StatusAffected)
 							vexDoc.Statements[i].Justification = ""
-							vexDoc.Statements[i].ImpactStatement = "Vulnerable component is loaded into the memory"
-							vexDoc.Statements[i].ActionStatement = "Upgrade the vulnerable component to a version that is not affected"
+							vexDoc.Statements[i].ImpactStatement = ""
+							vexDoc.Statements[i].ActionStatement = buildActionStatement(v)
 							foundProduct = true
 						}
 						if foundProduct {
@@ -770,6 +784,16 @@ func markRelevantVulnerabilitiesAsAffectedInVex(vexDoc *v1beta1.VEX, cvep *domai
 					}
 				}
 			}
+		}
+	}
+
+	// Backfill statements that were already marked "affected" by a version of kubevuln
+	// predating action_statement support, or that were left stale because their CVE fell
+	// out of the current relevancy set before the backfill could run on a prior update.
+	for i, s := range vexDoc.Statements {
+		if s.Status == v1beta1.Status(vex.StatusAffected) && s.ActionStatement == "" {
+			vexDoc.Statements[i].ImpactStatement = ""
+			vexDoc.Statements[i].ActionStatement = defaultActionStatement
 		}
 	}
 	return nil
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -758,6 +758,7 @@ func markRelevantVulnerabilitiesAsAffectedInVex(vexDoc *v1beta1.VEX, cvep *domai
 							vexDoc.Statements[i].Status = v1beta1.Status(vex.StatusAffected)
 							vexDoc.Statements[i].Justification = ""
 							vexDoc.Statements[i].ImpactStatement = "Vulnerable component is loaded into the memory"
+							vexDoc.Statements[i].ActionStatement = "Upgrade the vulnerable component to a version that is not affected"
 							foundProduct = true
 						}
 						if foundProduct {
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -786,16 +786,6 @@ func markRelevantVulnerabilitiesAsAffectedInVex(vexDoc *v1beta1.VEX, cvep *domai
 			}
 		}
 	}
-
-	// Backfill statements that were already marked "affected" by a version of kubevuln
-	// predating action_statement support, or that were left stale because their CVE fell
-	// out of the current relevancy set before the backfill could run on a prior update.
-	for i, s := range vexDoc.Statements {
-		if s.Status == v1beta1.Status(vex.StatusAffected) && s.ActionStatement == "" {
-			vexDoc.Statements[i].ImpactStatement = ""
-			vexDoc.Statements[i].ActionStatement = defaultActionStatement
-		}
-	}
 	return nil
 }
 
@@ -952,6 +942,7 @@ func (a *APIServerStore) updateVEX(ctx context.Context, cve domain.CVEManifest, 
 		vexDoc.Statements[i].Status = v1beta1.Status(vex.StatusNotAffected)
 		vexDoc.Statements[i].Justification = v1beta1.Justification(vex.VulnerableCodeNotPresent)
 		vexDoc.Statements[i].ImpactStatement = "Vulnerable component is not loaded into the memory"
+		vexDoc.Statements[i].ActionStatement = ""
 	}
 
 	// Now change the status of the filtered vulnerabilities to "Affected"
--- a/repositories/apiserver_test.go
+++ b/repositories/apiserver_test.go
@@ -405,6 +405,7 @@ func TestAPIServerStore_storeVEX_updateRestoresNotAffected(t *testing.T) {
 		if stmt.Status == v1beta1.Status(vex.StatusAffected) {
 			affectedAfter++
 		}
+		assert.Empty(t, stmt.ActionStatement, "not_affected statement %q must not carry an action_statement", stmt.Vulnerability.Name)
 	}
 	assert.Equal(t, 0, affectedAfter, "statements that are no longer relevant should be reset to not_affected")
 }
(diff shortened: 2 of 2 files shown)


rootp1:
hello @matthyx 
Please have a look whenever you have a chance to

repositories/apiserver.go
```diff
@@ -758,6 +758,7 @@ func markRelevantVulnerabilitiesAsAffectedInVex(vexDoc *v1beta1.VEX, cvep *domai
 							vexDoc.Statements[i].Status = v1beta1.Status(vex.StatusAffected)
 							vexDoc.Statements[i].Justification = ""
 							vexDoc.Statements[i].ImpactStatement = "Vulnerable component is loaded into the memory"
+							vexDoc.Statements[i].ActionStatement = "Upgrade the vulnerable component to a version that is not affected"
```
matthyx: ```Validate()``` never reaches this line.

In go-vex's `StatusAffected` case the `ImpactStatement` check runs first and returns, so as long as line 760 sets `ImpactStatement = "Vulnerable component is loaded into the memory"`, the statement fails validation with the same error before and after this change. Clearing `ImpactStatement` here (or moving that text to `StatusNotes`) is what makes the added `ActionStatement` observable. See the review comment for the before/after numbers.

matthyx (commented):
Thanks for picking this up — the diagnosis in #403 is correct, the call site you patched is the only place that sets `StatusAffected` (verified: `grep -n "StatusAffected" repositories/` → one hit at `apiserver.go:758`), and the new test is a genuine regression guard (I deleted the production line and it fails with `affected statement "CVE-2022-29824" must carry an action_statement`).

That said, I don't think this PR achieves the stated goal yet. Two blockers.

## Blocker 1 — the emitted VEX is exactly as invalid after this PR as before it

`markRelevantVulnerabilitiesAsAffectedInVex` clears `Justification` (correct — it's a `not_affected`-only field) but leaves `ImpactStatement` set on the affected statement. The type comment this PR cites says the same thing about that field:

> For "not_affected" status, a VEX statement MAY include an ImpactStatement that contains a description why the vulnerability cannot be exploited.

And this isn't just a comment — `github.com/openvex/go-vex`, which kubevuln already imports for `vex.StatusAffected`, enforces it in `Statement.Validate()`, and it checks `ImpactStatement` **before** `ActionStatement` in the `StatusAffected` case:

```go
case StatusAffected:
    if v := stmt.Justification; v != "" { ... }
    if v := stmt.ImpactStatement; v != "" {
        return fmt.Errorf("impact statement should not be set when using status %q (was set to %q)", s, v)
    }
    // action statement is now required
    if v := stmt.ActionStatement; v == "" { ... }
```

I ran `go-vex`'s own `Statement.Validate()` over everything `StoreVEX` produces for the `nginx-cve.json` / `nginx-cve-filtered.json` fixtures, on this branch and on `main`:

| | invalid statements | error |
|---|---|---|
| `main` | 216 / 456 | `impact statement should not be set when using status "affected" (was set to "Vulnerable component is loaded into the memory")` |
| this PR | 216 / 456 | *(identical)* |

`Validate()` returns on the first failure, so it never reaches the action-statement check. The document is byte-for-byte as non-conformant as it was; the field this PR adds is unreachable by the validator. If the point is "self-consistent with the local schema contract" for downstream consumers joining external VEX, this needs to clear `ImpactStatement` in the same block that clears `Justification`:

```go
vexDoc.Statements[i].Status = v1beta1.Status(vex.StatusAffected)
vexDoc.Statements[i].Justification = ""
vexDoc.Statements[i].ImpactStatement = ""
vexDoc.Statements[i].ActionStatement = "..."
```

If the "Vulnerable component is loaded into the memory" text is surfaced in a UI and can't just be dropped, `StatusNotes` is the spec-legal home for it (`MAY convey information about how [status] was determined`, no status restriction). Either way that's a call for a maintainer, but shipping `action_statement` while leaving `impact_statement` in place buys nothing measurable.

The new test should then assert the whole contract rather than one field — ideally by calling `Validate()` on each statement, which gets you the `not_affected` side for free too.

## Blocker 2 — already-stored affected statements are never backfilled

`markRelevantVulnerabilitiesAsAffectedInVex` only touches statements whose CVE is in the *current* `cvep.Content.Matches`. `updateVEX` never revisits the rest. So a statement that was marked `affected` by a pre-fix kubevuln and has since dropped out of the relevancy set keeps `status: affected` with `action_statement: ""` forever — no scan will ever fix it.

Reproduced against this branch: store with `nginx-cve-filtered-2.json`, blank out every `ActionStatement` to simulate a document written by an older version, store again with `nginx-cve-filtered.json` (which doesn't contain `CVE-2005-2541`):

```
still affected without action_statement: CVE-2005-2541
affected statements missing action_statement after update: 1
```

There's already precedent for exactly this in the same function — the ID/Name normalization loop at `apiserver.go:871-875` added for the same class of problem. A few lines next to it would cover it:

```go
for i, s := range vexDoc.Statements {
    if s.Status == v1beta1.Status(vex.StatusAffected) && s.ActionStatement == "" {
        vexDoc.Statements[i].ActionStatement = defaultActionStatement
    }
}
```

Worth a test on the update path — the current test only exercises `createVEX` (first store).

## Non-blocking

- **Use the fix data you already have.** The spec says the action statement SHOULD *describe actions* to remediate. `v.Vulnerability.Fix.Versions` is right there at the call site and is populated in the fixtures (`{"versions": ["232-25+deb9u11"], "state": "fixed"}`), so something like `Upgrade <purl> to <version>` is both more useful and closer to the SHOULD. The generic string is a reasonable fallback when `Fix.State != "fixed"`.
- **Extract the literal to a package const** rather than inlining a fourth hardcoded English string in this file — you'll need to reference it from the backfill loop above anyway.

## What's good

- Correct and complete on the call-site question — one place sets `affected`, and it's patched.
- The test genuinely fails without the production change, and `require.True(t, foundAffected, ...)` guards against the fixture silently producing zero affected statements and the assertion vacuously passing. That guard is often missing in tests like this.
- `go vet ./repositories/...` clean, `go test ./repositories/...` passes on this branch.

rootp1:
Thanks for the thorough review, @matthyx. Both blockers are addressed in the latest commit:

- `markRelevantVulnerabilitiesAsAffectedInVex` now clears `ImpactStatement` in the same block that clears `Justification` and sets `ActionStatement`, so `go-vex`'s `Statement.Validate()` no longer trips on the impact-statement check for affected statements.
- Added a backfill pass over all statements already marked `affected` with an empty `ActionStatement` (regardless of whether their CVE is in the current filtered manifest), so pre-fix documents and statements that fell out of relevancy before the fix landed get repaired on the next update.
- `ActionStatement` now uses `Vulnerability.Fix.Versions`/`State` when available (`Upgrade <purl> to version <versions>`), falling back to the generic string (now a package const, `defaultActionStatement`) when the match has no fixed version.
- Extended the existing test to also assert `ImpactStatement` is empty on affected statements, and added a new test reproducing your backfill repro (store with `nginx-cve-filtered-2.json`, blank out `ActionStatement` on `CVE-2005-2541`, store again with `nginx-cve-filtered.json`, assert it gets backfilled).

`go build ./...`, `go vet ./repositories/...`, and `go test ./repositories/...` all pass locally. Please have another look whenever you get a chance.

rootp1:
hey @matthyx, please have a look whenever you have a chance to

repositories/apiserver.go
```diff
@@ -771,6 +786,16 @@ func markRelevantVulnerabilitiesAsAffectedInVex(vexDoc *v1beta1.VEX, cvep *domai
 			}
 		}
 	}
+
+	// Backfill statements that were already marked "affected" by a version of kubevuln
+	// predating action_statement support, or that were left stale because their CVE fell
+	// out of the current relevancy set before the backfill could run on a prior update.
+	for i, s := range vexDoc.Statements {
```
matthyx: This loop is unreachable once you rebase onto `main`.

#401 (`a5e45e3`) now resets every statement to `not_affected` in `updateVEX` before this function runs, and `createVEX` builds them as `not_affected` too — so `s.Status == affected` is never true when this loop is reached. Verified with a `panic` in the body against the full `repositories` suite: zero hits. Its test also fails on the merged tree, because it asserts the stale statement stays `affected` and #401 deliberately made it revert.

Suggest dropping both, and instead clearing `ActionStatement` in #401's reset loop — see the review comment.

matthyx (commented):
Re-checked `0bbf1fe`. Both blockers from my earlier review are fixed, and both non-blocking suggestions were taken (`buildActionStatement` off `Fix.Versions`, `defaultActionStatement` const). Re-running `go-vex`'s `Statement.Validate()` over everything `StoreVEX` emits for the `nginx-cve` fixtures, **on this branch as it stands**:

| | invalid statements |
|---|---|
| `main` @ `eeba982` (the base this PR was cut from) | 216 / 456 |
| `b0d58ab` (first commit) | 216 / 456 |
| `0bbf1fe` (now) | **0 / 456** |

`go vet ./repositories/...` and `go test ./repositories/...` are clean here too. That's the right outcome.

**But this branch is now out of date, and rebasing breaks it.** #401 merged to `main` as `a5e45e3` about an hour after you opened this, and GitHub already reports the PR as `CONFLICTING`. #401 added a reset loop to `updateVEX` that sends every statement back to the `not_affected` baseline before `markRelevantVulnerabilitiesAsAffectedInVex` reapplies the current manifest. That interacts badly with this change in two ways.

I merged `origin/main` into `0bbf1fe` locally (only `apiserver_test.go` conflicts — `apiserver.go` auto-merges) and resolved it the mechanical way, keeping both sides' tests. Results below.

### 1. Blocker — the reset loop doesn't clear `ActionStatement`, so the bug comes back mirrored

`a5e45e3` resets `Status`, `Justification` and `ImpactStatement`, but not `ActionStatement`:

```go
for i := range vexDoc.Statements {
    vexDoc.Statements[i].Justification = v1beta1.Justification(vex.VulnerableCodeNotPresent)
    vexDoc.Statements[i].ImpactStatement = "Vulnerable component is not loaded into the memory"
}
```

Before this PR that was harmless, because nothing ever set `ActionStatement`. With this PR it means every statement that stops being relevant reverts to `not_affected` while keeping the action statement from the previous scan — and `Validate()` rejects that just as hard, from the `StatusNotAffected` case:

```
action statement should not be set when using status "not_affected"
  (was set to "Upgrade pkg:deb/debian/libssl1.1@1.1.0f-3+deb9u2?... to version 1.1.0l-1~deb9u3")
```

Same fixtures, store once with `nginx-cve-filtered.json` then again with an empty filtered manifest, on the merged tree: **216 / 456 invalid** — the identical count, just pointing the other way. One line in that loop fixes it, and I verified it takes the merged tree to **0 / 456**:

```go
    vexDoc.Statements[i].ImpactStatement = "Vulnerable component is not loaded into the memory"
    vexDoc.Statements[i].ActionStatement = ""
```

I'd put that line in this PR rather than a follow-up — it's this PR that makes the field non-empty in the first place, and the two changes are only in conflict because they're in flight together.

### 2. Blocker — the backfill loop and its test contradict what #401 just landed

`TestAPIServerStore_updateVEX_backfillsActionStatementOnStaleAffectedStatements` asserts `"stale affected statement should remain affected"`. That is exactly the behavior #401 removed on purpose. On the merged tree it fails:

```
Error: Not equal:
  expected: "affected"
  actual  : "not_affected"
Test: TestAPIServerStore_updateVEX_backfillsActionStatementOnStaleAffectedStatements
Messages: stale affected statement should remain affected
```

The loop it guards is also unreachable after the rebase — every statement now enters `markRelevantVulnerabilitiesAsAffectedInVex` as `not_affected` (reset by #401 on the update path, freshly constructed on the create path), so the `s.Status == affected && s.ActionStatement == ""` condition never holds. I confirmed by putting a `panic` in the loop body and running the whole `repositories` suite: zero hits.

So the scenario I flagged in my last review is real, but #401 solved it a better way — by not leaving statements stale at all. With fix 1 above in place I'd just drop the backfill loop and that test; the "legacy document written by a pre-fix kubevuln" case is covered, since those statements get reset to `not_affected` on the next update like any other.

### Summary

- Rebase onto `main` (required anyway — the PR is conflicting).
- Add `ActionStatement = ""` to `updateVEX`'s reset loop.
- Drop the backfill loop and `TestAPIServerStore_updateVEX_backfillsActionStatementOnStaleAffectedStatements`; keep `TestAPIServerStore_storeVEX_affectedStatementsHaveActionStatement`, which is a good guard and passes on the merged tree.
- Optional but worth it now that there are two symmetric rules to keep straight: assert `stmt.Validate()` per statement in that test instead of hand-checking individual fields. It would have caught both directions of this on its own.

repositories/apiserver.go
rootp1: @matthyx dropped the backfill loop and its test, and added ActionStatement clearing to the reset loop instead. Tests pass on the rebased branch.

rootp1:
@matthyx pushed both required fixes: reset loop now clears ActionStatement, and I dropped the unreachable backfill loop along with its now-contradicted test. go test ./repositories/... and go vet ./repositories/... are clean on the rebased branch.

I skipped the stmt.Validate() suggestion for now. v1beta1.Statement and go-vex's Statement are different types with no conversion helper in this repo, so reusing Validate() directly would mean adding a mapping layer just for test assertions, which felt like more surface area than this fix needs. Happy to revisit if you'd rather have that in place.

rootp1:
@matthyx could you take another look when you get a chance?

matthyx (commented):
Re-checked `53b2adf`. All three blockers are resolved and I have nothing left that should hold this up.

Rebased onto `a5e45e3`, `ActionStatement = ""` added to the reset loop, backfill loop and its test dropped. Nice touch extending `TestAPIServerStore_storeVEX_updateRestoresNotAffected` with the `not_affected` assertion instead of adding a parallel test — that's where the invariant belongs.

### Verification

`go-vex`'s `Statement.Validate()` over every statement `StoreVEX` emits, across the paths that can change a status:

| scenario | statements | affected | invalid |
|---|---|---|---|
| create | 456 | 216 | **0** |
| update, relevancy set widens (`filtered` → `filtered-2`) | 456 | 217 | **0** |
| update, relevancy set narrows (`filtered-2` → `filtered`) | 456 | 216 | **0** |
| update, nothing relevant anymore | 456 | 0 | **0** |

And the legacy-document case from my first review, which the dropped backfill loop used to handle — seed a stored document the way a pre-fix kubevuln would have written it (217 statements `affected`, `impact_statement` set, no `action_statement`), then rescan:

```
legacy before rescan   statements=456 affected=217 INVALID=217
legacy after rescan    statements=456 affected=216 INVALID=0
```

Healed in a single update by #401's reset loop, so dropping the backfill was right — it wasn't just unreachable, it was redundant.

Merged `origin/main` (`fede5c2`) in locally: clean, no conflicts. `go build ./...`, `go vet ./repositories/...`, `gofmt -l`, and the full `go test ./...` all pass on the merged tree. CI is green on the PR head.

### Two non-blocking observations

**`wont-fix` matches get a fallback that can't be acted on.** `buildActionStatement` falls through to `defaultActionStatement` — *"Upgrade the vulnerable component to a version that is not affected"* — whenever `Fix.State != "fixed"`. But grype distinguishes three states, and in this fixture alone they split:

| `fix.state` | matches |
|---|---|
| `fixed` (with versions) | 82 |
| `wont-fix` | 80 |
| `not-fixed` | 54 |

For the 80 `wont-fix` ones, telling a reader to upgrade to a non-affected version is advice that by definition cannot be followed — the vendor has declared there won't be one. The spec asks the action statement to describe actions to *remediate or mitigate*; for `wont-fix` the honest action is closer to "no upstream fix is available — remove or replace the component, or restrict its exposure". `not-fixed` is fine on the generic string. Cheap to split, and it's the difference between a field that's populated and a field that's useful.

**The relevancy signal is now gone from affected statements.** Pre-PR, an affected statement carried *"Vulnerable component is loaded into the memory"* — kubevuln's whole differentiator, in the document. Clearing `ImpactStatement` was the right call for validity, but nothing replaced it, so that information no longer survives into the VEX. `StatusNotes` is the spec-legal home (`MAY convey information about how [status] was determined`, no status restriction) and `Validate()` doesn't constrain it. Worth a maintainer's call on whether anything downstream was reading `impact_statement` off affected statements — if so, this is a silent behavior change for them; if not, `status_notes` is still a cheap way to keep the provenance.

Neither blocks merge. LGTM from my side.

matthyx (approved):

PR #405, opened 2026-07-31
fix: mergeMaps panics on nil map, crashing the process when storing over an object without labels or annotations

### LINKED ISSUE #402: fix: mergeMaps panics on nil map, crashing the process when storing over an object without labels/annotations
# Description

`mergeMaps` (`repositories/apiserver.go:1200-1204`) writes into `existing` without checking whether it is nil:

```go
func mergeMaps(existing, new map[string]string) {
	for k, v := range new {
		existing[k] = v   // panics if existing == nil and new is non-empty
	}
}
```

It is called on both `result.Annotations` and `result.Labels` of an object just fetched from storage, inside the `RetryOnConflict` closure of every `Store*` update path — 10 call sites across 5 blocks:

| Caller | Call sites |
|---|---|
| `StoreCVE` | `apiserver.go:341-342` |
| `StoreCVESummary` | `apiserver.go:578-579` |
| `StoreCVESummaryStub` | `apiserver.go:665-666` |
| `StoreSBOM` (filtered) | `apiserver.go:1132-1133` |
| `StoreSBOM` (unfiltered) | `apiserver.go:1165-1166` |

A Kubernetes object with no `.metadata.labels` / `.metadata.annotations` decodes to a **nil** map, not an empty one. So if the object already in storage is missing either, and the incoming manifest has the corresponding map populated, the store panics.

In every one of the five blocks at least one of the two incoming maps is **unconditionally non-empty**, so there is no case where the nil map is harmlessly skipped:

- `StoreCVE` — `manifest.Labels` always carries `helpersv1.ContextMetadataKey` (`apiserver.go:520-528`)
- `StoreSBOM` (both blocks) — `manifest.Annotations` always carries `helpersv1.StatusMetadataKey` (`apiserver.go:1122`)
- `StoreCVESummary` / `StoreCVESummaryStub` — `enrichSummaryManifestObjectAnnotations` and `enrichSummaryManifestObjectLabels` (`apiserver.go:443-462`, `464-500`) both allocate and always write at least one key, so *both* maps are always non-empty

The panic is **not recovered anywhere**. There is no `recover()` in the repository, every `Store*` caller runs inside a `h.workerPool.Submit(...)` goroutine (`controllers/http.go:66, 123, 170, 240`), and `gammazero/workerpool@v1.1.3` invokes `task()` bare (`workerpool.go:120, 237`). Gin's recovery middleware does not apply, since the work happens off the request goroutine. A single malformed stored object therefore takes down the whole kubevuln process.

# Environment

- `kubescape/kubevuln` at `main` (`22be3a5`)
- `repositories/apiserver.go`

# Steps To Reproduce

Seed a stored object with nil `ObjectMeta.Annotations` / `ObjectMeta.Labels` — which is what an object with no `metadata.labels` / `metadata.annotations` in storage decodes to — then call the corresponding `Store*` method so it takes the `IsAlreadyExists` → update path:

```go
seeded := &v1beta1.VulnerabilityManifest{
	ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: "kubescape"}, // nil maps
}
a := &APIServerStore{
	StorageClient: fake.NewSimpleClientset(seeded).SpdxV1beta1(),
	Namespace:     "kubescape",
}
_ = a.StoreCVE(context.TODO(), domain.CVEManifest{Name: name}, false)
```

I ran this against `22be3a5` for all five blocks. Every one panics:

```
[StoreCVE]            PANIC: assignment to entry in nil map
[StoreSBOM]           PANIC: assignment to entry in nil map
[StoreSBOMFiltered]   PANIC: assignment to entry in nil map
[StoreCVESummary]     PANIC: assignment to entry in nil map
[StoreCVESummaryStub] PANIC: assignment to entry in nil map
```

In a cluster, the equivalent trigger is any `VulnerabilityManifest` / `VulnerabilityManifestSummary` / `SBOMSyft` / `SBOMSyftFiltered` in storage without those metadata fields, followed by a re-scan of the same image or workload.

# Expected behavior

Storing over an existing object with no labels or annotations should merge normally and succeed. A missing optional metadata map is valid Kubernetes input, not a crash condition.

# Actual Behavior

`panic: assignment to entry in nil map`, unrecovered, terminating the kubevuln process. The offending object stays in storage, so the next scan of the same image or workload hits it again.

# Additional context

**How reachable is this in practice?** kubevuln's own `Create` calls populate the relevant maps, so a storage lifecycle driven purely by a current kubevuln will not produce a triggering object on its own. The realistic sources are an object written by an **older kubevuln**, by **another component** writing these CRDs, or a hand-applied / `kubectl edit`-ed manifest. That makes it a latent crash rather than an everyday one — but the blast radius when it does fire (whole-process panic, no recovery, re-triggered on the next scan) argues for fixing it rather than leaving it to chance.

**Suggested fix.** Guard inside `mergeMaps`, which fixes all 10 call sites at once and needs no caller changes:

```go
func mergeMaps(existing, new map[string]string) map[string]string {
	if existing == nil {
		existing = make(map[string]string, len(new))
	}
	for k, v := range new {
		existing[k] = v
	}
	return existing
}
```

Returning the map means the call sites become `result.Annotations = mergeMaps(result.Annotations, manifest.Annotations)`, since a freshly allocated map has to be assigned back. A non-returning variant that instead sets `result.Annotations`/`result.Labels` at each call site works too, but touches ten places instead of five pairs.

Worth adding a table-driven test over the five blocks with a nil-map seed, so a future `Store*` method with the same shape does not silently reintroduce it.

**Relationship to #399.** #399 neither introduces nor fixes this — `mergeMaps` is untouched there. I noticed it while reviewing that PR: its new retry-exhaustion tests have to seed `Annotations: map[string]string{}` and `Labels: map[string]string{}` explicitly, and dropping those seeds turns the test panic-red. Keeping this out of #399 so it gets its own fix and its own regression test.

AdeshDeshmukh:
Fixes #402

## What's going on

`mergeMaps` in `repositories/apiserver.go` wrote straight into the destination map without checking whether it was `nil`. That's fine when the object we're updating already has labels and annotations — but a Kubernetes object without `metadata.labels` / `metadata.annotations` decodes to a **nil** map, not an empty one. In that case the merge panicked with `assignment to entry in nil map`, and since nothing recovers panics in the store layer, the whole kubevuln process went down. The offending object stayed in storage, so the next scan of the same image hit it again.

Every `Store*` update path was affected — 10 call sites across 5 blocks. That's how we found it: while reviewing #399, @matthyx noticed our retry-exhaustion tests had to seed `Annotations: map[string]string{}` explicitly, and dropping those seeds turned the tests panic-red. He filed this issue with a clear diagnosis, and it's a real crash waiting for the right stored object.

## The fix

`mergeMaps` now guards against a nil destination and returns the merged map:

```go
func mergeMaps(existing, new map[string]string) map[string]string {
	if existing == nil {
		existing = make(map[string]string, len(new))
	}
	for k, v := range new {
		existing[k] = v
	}
	return existing
}
```

The 10 call sites (StoreCVE, StoreCVESummary, StoreCVESummaryStub, StoreSBOM filtered and unfiltered) assign the returned map back to the object being updated, so the freshly-allocated map is actually used.

## Tests

- `TestMergeMaps` updated for the new signature, with two new cases covering a nil `existing` map.
- The five retry-exhaustion tests now seed their objects **without** labels/annotations, so they drive the exact update path a malformed stored object would. On the old code they panic; with this change they pass.

## Verification

- `go build ./...` — clean
- `go test ./repositories/...` — 54 passed
- `go test ./core/...` — 108 passed
- `gofmt -l repositories/` and `go vet ./repositories/...` — clean
- Reverting the guard reproduces the exact panic from the issue (`assignment to entry in nil map`) on the retry-exhaustion tests; with the fix they pass.

No behavior change for objects that already have labels/annotations — only the nil case is handled, and callers already assign the result.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed updates to CVE and SBOM records so merged labels and annotations are consistently saved.
  * Improved handling when existing metadata is missing, preventing update failures and ensuring metadata is preserved correctly.

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -338,8 +338,8 @@ func (a *APIServerStore) StoreCVE(ctx context.Context, cve domain.CVEManifest, w
 				return getErr
 			}
 			// update the vulnerability manifest
-			mergeMaps(result.Annotations, manifest.Annotations)
-			mergeMaps(result.Labels, manifest.Labels)
+			result.Annotations = mergeMaps(result.Annotations, manifest.Annotations)
+			result.Labels = mergeMaps(result.Labels, manifest.Labels)
 			result.Spec = manifest.Spec
 			// try to send the updated vulnerability manifest
 			_, updateErr := a.StorageClient.VulnerabilityManifests(a.Namespace).Update(context.Background(), result, metav1.UpdateOptions{})
@@ -576,8 +576,8 @@ func (a *APIServerStore) StoreCVESummary(ctx context.Context, cve domain.CVEMani
 				return getErr
 			}
 			// update the vulnerability manifest
-			mergeMaps(result.Annotations, manifest.Annotations)
-			mergeMaps(result.Labels, manifest.Labels)
+			result.Annotations = mergeMaps(result.Annotations, manifest.Annotations)
+			result.Labels = mergeMaps(result.Labels, manifest.Labels)
 			result.Spec = manifest.Spec
 			// try to send the updated vulnerability manifest
 			_, updateErr := a.StorageClient.VulnerabilityManifestSummaries(workloadNamespace).Update(context.Background(), result, metav1.UpdateOptions{})
@@ -664,8 +664,8 @@ func (a *APIServerStore) StoreCVESummaryStub(ctx context.Context, status string)
 				return nil
 			}
 			// refresh annotations/labels only, keep any existing Spec
-			mergeMaps(result.Annotations, manifest.Annotations)
-			mergeMaps(result.Labels, manifest.Labels)
+			result.Annotations = mergeMaps(result.Annotations, manifest.Annotations)
+			result.Labels = mergeMaps(result.Labels, manifest.Labels)
 			_, updateErr := a.StorageClient.VulnerabilityManifestSummaries(workloadNamespace).Update(context.Background(), result, metav1.UpdateOptions{})
 			return updateErr
 		})
@@ -1132,8 +1132,8 @@ func (a *APIServerStore) StoreSBOM(ctx context.Context, sbom domain.SBOM, isFilt
 					return getErr
 				}
 				// update the filtered sbom
-				mergeMaps(result.Annotations, filtered.Annotations)
-				mergeMaps(result.Labels, filtered.Labels)
+				result.Annotations = mergeMaps(result.Annotations, filtered.Annotations)
+				result.Labels = mergeMaps(result.Labels, filtered.Labels)
 				result.Spec = filtered.Spec
 				// try to send the updated filtered sbom
 				_, updateErr := a.StorageClient.SBOMSyftFiltereds(a.Namespace).Update(context.Background(), result, metav1.UpdateOptions{})
@@ -1166,8 +1166,8 @@ func (a *APIServerStore) StoreSBOM(ctx context.Context, sbom domain.SBOM, isFilt
 					return getErr
 				}
 				// update the sbom
-				mergeMaps(result.Annotations, manifest.Annotations)
-				mergeMaps(result.Labels, manifest.Labels)
+				result.Annotations = mergeMaps(result.Annotations, manifest.Annotations)
+				result.Labels = mergeMaps(result.Labels, manifest.Labels)
 				result.Spec = manifest.Spec
 				// try to send the updated sbom
 				_, updateErr := a.StorageClient.SBOMSyfts(a.Namespace).Update(context.Background(), result, metav1.UpdateOptions{})
@@ -1201,9 +1201,15 @@ func convertToFilteredSBOM(sbom *v1beta1.SBOMSyft) *v1beta1.SBOMSyftFiltered {
 	}
 }
 
-// mergeMaps merges new into existing, overwriting existing keys with new values
-func mergeMaps(existing, new map[string]string) {
-	for k, v := range new {
+// mergeMaps merges incoming into existing, overwriting existing keys with new values,
+// and returns the merged map. Callers MUST use the returned value: when existing is
+// nil a fresh map is allocated, so ignoring the return silently drops the merge.
+func mergeMaps(existing, incoming map[string]string) map[string]string {
+	if existing == nil {
+		existing = make(map[string]string, len(incoming))
+	}
+	for k, v := range incoming {
 		existing[k] = v
 	}
+	return existing
 }
--- a/repositories/apiserver_test.go
+++ b/repositories/apiserver_test.go
@@ -854,14 +854,49 @@ func TestMergeMaps(t *testing.T) {
 			new:      map[string]string{},
 			expected: map[string]string{},
 		},
+		{
+			name:     "merge with nil existing map",
+			existing: nil,
+			new:      map[string]string{"key1": "value1"},
+			expected: map[string]string{"key1": "value1"},
+		},
+		{
+			name:     "merge with nil existing and nil new maps",
+			existing: nil,
+			new:      nil,
+			expected: map[string]string{},
+		},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			mergeMaps(tt.existing, tt.new)
-			assert.Equal(t, tt.expected, tt.existing)
+			got := mergeMaps(tt.existing, tt.new)
+			assert.Equal(t, tt.expected, got)
 		})
 	}
+}
+
+func TestAPIServerStore_StoreCVE_mergesMetadataOnUpdate(t *testing.T) {
+	seeded := &v1beta1.VulnerabilityManifest{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: "kubescape",
+		},
+	}
+	clientset := fake.NewSimpleClientset(seeded)
+	a := &APIServerStore{StorageClient: clientset.SpdxV1beta1(), Namespace: "kubescape"}
+
+	cve := domain.CVEManifest{
+		Name:        name,
+		Annotations: map[string]string{"k1": "v1"},
+		Labels:      map[string]string{"l1": "v2"},
+	}
+	require.NoError(t, a.StoreCVE(context.TODO(), cve, false))
+
+	got, err := a.StorageClient.VulnerabilityManifests("kubescape").Get(context.TODO(), name, metav1.GetOptions{})
+	require.NoError(t, err)
+	assert.Equal(t, "v1", got.Annotations["k1"])
+	assert.Equal(t, "v2", got.Labels["l1"])
 }
 
 func TestAPIServerStore_GetCVE_transientError(t *testing.T) {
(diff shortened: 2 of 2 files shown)


matthyx (commented):
Reviewed at 7ee6e196. The diagnosis and the fix are both correct, and I reproduced the crash independently. One blocker before merge, plus a few smaller things.

### What I verified

- `go build ./...`, `gofmt -l repositories/`, `go vet ./repositories/...` — all clean. `go test ./repositories/...` and `go test ./core/...` — pass.
- Removing the nil guard and running each retry-exhaustion test **individually**: all five panic with `assignment to entry in nil map` at `apiserver.go:1207`. So the crash is real and every one of the five genuinely reaches the path (not just the first one to abort the binary).
- No `recover()` and no `HandleCrash` anywhere in the repo, so the panic really does take the process down as the description says.
- Audited the rest of the store layer for the same shape: `createVEX`/`updateVEX` only read from maps, and `StoreCVE` (L300) and `StoreSBOM` (L1118) already nil-guard the maps they build locally. `mergeMaps` was the only remaining site — the fix scope is complete.

### Blocker: the `result.X =` half of the fix has no test

Revert all 10 call sites to bare `mergeMaps(result.Annotations, manifest.Annotations)` while **keeping** the new nil guard, and the whole `./repositories/...` suite still passes. That configuration silently drops every label and annotation on update for any stored object with nil metadata — precisely the "metadata is preserved correctly" behavior this PR claims to deliver. Nothing fails.

The five repurposed tests can't catch it by construction: they inject `Conflict` on every update, so `Update` never succeeds and no persisted object is ever read back. They prove "doesn't panic", not "merged and saved". `TestMergeMaps` covers the function in isolation, not the 10 call sites — which are 10 of the 12 changed production lines.

Please add one test where the update **succeeds**: seed a `VulnerabilityManifest` with no `Annotations`/`Labels`, call `StoreCVE` with a manifest carrying both, then `Get` from the fake clientset and assert the merged keys landed on the stored object. One call site is enough — the other four are the same shape.

### Non-blocking

1. The `mergeMaps` doc comment still reads "merges new into existing" with no mention of the return value or the nil case. The function now has a mixed contract — mutates in place when non-nil, allocates and returns when nil — so a caller who ignores the result is silently wrong. Worth stating that callers must use the returned map.
2. The five retry-exhaustion tests now do double duty (retry-count assertion + nil-metadata regression) with nothing in their names or assertions saying so. If anyone re-adds `Annotations: map[string]string{}` for an unrelated reason, the regression coverage vanishes and no test fails.
3. `new` shadows the builtin. Pre-existing, but the signature line is already being touched, so `incoming` or `updates` is free.
4. The branch is cut from `eeba982`; `main` is now at `a5e45e3` (#401 not included). Mergeable and CI is green, so this only matters if you want a rebase first.

### On urgency

The production change is correct and this fixes a real process crash, so I would not hold it if you want the fix in now. But it should not go in without the follow-up test — as it stands, nothing prevents a future "simplify" from reverting 10 of the 12 changed production lines with a green suite.

repositories/apiserver.go
```diff
@@ -1202,8 +1202,12 @@ func convertToFilteredSBOM(sbom *v1beta1.SBOMSyft) *v1beta1.SBOMSyftFiltered {
 }
 
 // mergeMaps merges new into existing, overwriting existing keys with new values
-func mergeMaps(existing, new map[string]string) {
+func mergeMaps(existing, new map[string]string) map[string]string {
```
matthyx: The contract changed in a way worth documenting: this mutates `existing` in place when it is non-nil, but returns a **fresh** map when it is nil. That makes the return value load-bearing, and a caller that ignores it is silently wrong. The doc comment above still just says "merges new into existing" — please extend it to say callers must assign the result.

That is not hypothetical: revert the 10 call sites to bare `mergeMaps(...)` while keeping this guard, and it compiles and passes the entire suite while dropping all metadata on update (details in the review summary).

Minor while you are here: `new` shadows the builtin, and this signature line is already changing — `incoming` would be free.

repositories/apiserver_test.go
```diff
@@ -1009,10 +1021,8 @@ func TestAPIServerStore_StoreCVESummary_updateGetFailure_transientError(t *testi
 func TestAPIServerStore_StoreCVE_retryExhausted_transientError(t *testing.T) {
 	seeded := &v1beta1.VulnerabilityManifest{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:        name,
-			Namespace:   "kubescape",
-			Annotations: map[string]string{},
-			Labels:      map[string]string{},
+			Name:      name,
```
matthyx, line 1052: **Blocker.**

Dropping the map seeds here does drive the panic path — I removed the nil guard and ran each of the five retry-exhaustion tests individually, and all five panic at `apiserver.go:1207`. So this genuinely covers the crash.

What it cannot cover is the other half of the fix. Every update in these tests returns `Conflict`, so `Update` never succeeds and no stored object is ever read back — they assert "didn't panic", not "merged and saved".

Concretely: revert all 10 call sites to bare `mergeMaps(result.Annotations, manifest.Annotations)` but keep the nil guard, and `go test ./repositories/...` still passes — while every label and annotation is now silently dropped on update for any object with nil metadata.

Please add one test where the update **succeeds**: seed a `VulnerabilityManifest` with no `Annotations`/`Labels`, call `StoreCVE` with a manifest carrying both, then `Get` from the fake clientset and assert the merged keys landed on the stored object. One call site is enough — the other four are the same shape.

Secondary: these tests now do double duty (retry count + nil-metadata regression) with nothing in their names saying so, so a future unrelated re-seed of `Annotations: map[string]string{}` would silently delete the regression coverage.

AdeshDeshmukh:
Thanks for reproducing the crash and walking through the store layer — that was genuinely helpful. I've pushed the fix with the test you asked for: seeded a manifest with no labels/annotations, stored one with both, read it back and confirmed the keys survive. It fails without the assignments, so it's actually catching the regression.

 Also updated the mergeMaps comment and renamed new to incoming. New head is 84f7c89 — happy to adjust further if needed....!

matthyx (commented):
Re-reviewed at `84f7c89`. **The blocker is resolved** — and I verified it the same way I found it, rather than taking the new test's word for it.

### Blocker closed

`TestAPIServerStore_StoreCVE_mergesMetadataOnUpdate` is exactly the missing coverage. I re-ran the experiment that motivated the finding: reverting all 10 call sites to bare `mergeMaps(result.Annotations, manifest.Annotations)` while keeping the nil guard now **fails**, on both halves:

```
--- FAIL: TestAPIServerStore_StoreCVE_mergesMetadataOnUpdate
    apiserver_test.go:898  expected: "v1"  actual: ""
    apiserver_test.go:899  expected: "v2"  actual: ""
```

Previously that same edit passed the whole suite. The assign-back is now load-bearing in a way a test enforces, which was the entire ask.

I also confirmed the original panic coverage did not regress in the process: with the nil guard removed, all five retry-exhaustion tests still panic with `assignment to entry in nil map` when run individually. Both halves of the fix now have independent regression coverage.

### Other points, all addressed

- The `mergeMaps` doc comment now states the contract explicitly, including "ignoring the return silently drops the merge" — which is precisely the failure mode above.
- `new` → `incoming`.
- The load-bearing-seeds NOTE above the retry-exhaustion tests fixes the silent-coverage-loss concern, and points forward to the new test. Good call putting it there rather than in a commit message.

### Verification on this head

`go build ./...`, `gofmt -l repositories/`, `go vet ./repositories/...` clean; `go test ./repositories/...` and `go test ./core/...` pass. CI is green and GitHub reports MERGEABLE.

### Two non-blocking notes

1. The new test seeds an object with no metadata, so it locks in the nil case only. It does not assert the *merge* semantic on a non-nil map — that pre-existing keys survive alongside the new ones. That path was already working and is unchanged by this PR, so it is not a regression risk; just a cheap extra assertion if you ever touch this again.
2. The branch is still cut from `eeba982` while `main` has moved to `fede5c2`. Merges cleanly, so this only matters if you prefer a rebase before merge.

Nothing blocking left from my side. I have deliberately left this as a comment rather than an approval — the merge call is the maintainers'.

matthyx (approved):

PR #407, opened 2026-07-31
fix: send actual lifecycle status in BackendAdapter.SendStatus

### LINKED ISSUE #406: Bug: BackendAdapter.SendStatus always sends JobSuccess even for started and done lifecycle steps
# Description
`BackendAdapter.SendStatus` computes the correct `report.Status` from the `statuses` table (`JobStarted`, `JobSuccess`, or `JobDone` depending on the step), but the actual call to `a.sendStatusFunc` always passes `sysreport.JobSuccess` as the status argument instead of `report.Status`. This means the transport-level status sent to the backend never reflects the actual lifecycle step being reported.

# Environment
OS: `Any`
Version: `upstream/main`

# Steps To Reproduce
1. Open `adapters/v1/backend.go`.
2. Read the `statuses` table at lines 74-79.
3. Read `SendStatus` at lines 205-219.
4. Observe that line 217 always calls `a.sendStatusFunc(sender, sysreport.JobSuccess, true)` instead of using `report.Status` (which was set from `statuses[step]` on line 214).

# Expected behavior
The sender should transmit the same lifecycle status (`JobStarted`, `JobSuccess`, or `JobDone`) that was chosen for the current step.

# Actual Behavior
Every call to `sendStatusFunc` goes through the `JobSuccess` path, even when `report.Status` was set to `JobStarted` or `JobDone`.

# Additional context
The existing unit test in `adapters/v1/backend_test.go` (`TestBackendAdapter_SendStatus`) does not assert the `s` (status) argument passed to `sendStatusFunc`, so this mismatch is not covered today.

rootp1:
## Overview
`BackendAdapter.SendStatus` computed the correct lifecycle status (`JobStarted`, `JobSuccess`, `JobDone`) into `report.Status`, but the call to `sendStatusFunc` always passed the hardcoded `sysreport.JobSuccess` constant instead of `report.Status`. This fix passes `report.Status` so the transport-level status argument matches the step being reported.

## How to Test
Run `go test ./adapters/v1/... -run TestBackendAdapter_SendStatus -v`. The updated test asserts that the status passed to `sendStatusFunc` matches the expected status for the given step.

## Related issues/PRs:
Resolved #406

## Summary by CodeRabbit

* **Bug Fixes**
  * Status reports now accurately reflect the computed result of each job step instead of always reporting success.

* **Tests**
  * Added coverage to verify that emitted statuses match the expected step results.

### DIFF
--- a/adapters/v1/backend.go
+++ b/adapters/v1/backend.go
@@ -214,7 +214,7 @@ func (a *BackendAdapter) SendStatus(ctx context.Context, step int) error {
 	report.Status = statuses[step]
 
 	sender := backendClientV1.NewBaseReportSender(a.eventReceiverRestURL, &http.Client{}, a.getRequestHeaders(), report)
-	a.sendStatusFunc(sender, sysreport.JobSuccess, true)
+	a.sendStatusFunc(sender, report.Status, true)
 	return nil
 }
 
--- a/adapters/v1/backend.go
+++ b/adapters/v1/backend.go
@@ -214,7 +214,7 @@ func (a *BackendAdapter) SendStatus(ctx context.Context, step int) error {
 	report.Status = statuses[step]
 
 	sender := backendClientV1.NewBaseReportSender(a.eventReceiverRestURL, &http.Client{}, a.getRequestHeaders(), report)
-	a.sendStatusFunc(sender, report.Status, true)
+	a.sendStatusFunc(sender, statuses[step], true)
 	return nil
 }
 
--- a/adapters/v1/backend_test.go
+++ b/adapters/v1/backend_test.go
@@ -312,6 +312,7 @@ func TestBackendAdapter_SendStatus(t *testing.T) {
 				sendStatusFunc: func(sender *beClientV1.BaseReportSender, s string, b bool) {
 					report := sender.GetBaseReport()
 					assert.NotEqual(t, *report, tt.report) //nolint:govet
+					assert.Equal(t, statuses[tt.step], s)
 				},
 			}
 			ctx := context.TODO()
--- a/adapters/v1/backend_test.go
+++ b/adapters/v1/backend_test.go
@@ -289,7 +289,6 @@ func TestBackendAdapter_SendStatus(t *testing.T) {
 	tests := []struct {
 		name       string
 		step       int
-		report     sysreport.BaseReport
 		wantStatus string
 		wantErr    bool
 	}{
@@ -297,15 +296,6 @@ func TestBackendAdapter_SendStatus(t *testing.T) {
 			name:       "send status",
 			step:       1,
 			wantStatus: sysreport.JobStarted,
-			report: sysreport.BaseReport{
-				Reporter:   ReporterName,
-				Target:     "vuln scan:: scanning wlid: wlid , container: container imageTag: imageTag imageHash: imageHash",
-				Status:     "Dequeueing",
-				ActionName: ActionName,
-				ActionID:   "1",
-				ActionIDN:  1,
-				Details:    "started",
-			},
 		},
 		{
 			name:       "inqueueing step reports started",
@@ -323,12 +313,12 @@ func TestBackendAdapter_SendStatus(t *testing.T) {
 			wantStatus: sysreport.JobDone,
 		},
 	}
-	for _, tt := range tests { //nolint:govet
+	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			var called bool
 			a := &BackendAdapter{
 				sendStatusFunc: func(sender *beClientV1.BaseReportSender, s string, b bool) {
-					report := sender.GetBaseReport()
-					assert.NotEqual(t, *report, tt.report) //nolint:govet
+					called = true
 					assert.Equal(t, tt.wantStatus, s)
 				},
 			}
@@ -344,6 +334,7 @@ func TestBackendAdapter_SendStatus(t *testing.T) {
 			if err := a.SendStatus(ctx, tt.step); (err != nil) != tt.wantErr {
 				t.Errorf("SendStatus() error = %v, wantErr %v", err, tt.wantErr)
 			}
+			assert.True(t, called, "sendStatusFunc was never invoked")
 		})
 	}
 }
--- a/adapters/v1/backend_test.go
+++ b/adapters/v1/backend_test.go
@@ -287,14 +287,16 @@ func TestNewBackendAdapter(t *testing.T) {
 
 func TestBackendAdapter_SendStatus(t *testing.T) {
 	tests := []struct {
-		name    string
-		step    int
-		report  sysreport.BaseReport
-		wantErr bool
+		name       string
+		step       int
+		report     sysreport.BaseReport
+		wantStatus string
+		wantErr    bool
 	}{
 		{
-			name: "send status",
-			step: 1,
+			name:       "send status",
+			step:       1,
+			wantStatus: sysreport.JobStarted,
 			report: sysreport.BaseReport{
 				Reporter:   ReporterName,
 				Target:     "vuln scan:: scanning wlid: wlid , container: container imageTag: imageTag imageHash: imageHash",
@@ -305,6 +307,21 @@ func TestBackendAdapter_SendStatus(t *testing.T) {
 				Details:    "started",
 			},
 		},
+		{
+			name:       "inqueueing step reports started",
+			step:       0,
+			wantStatus: sysreport.JobStarted,
+		},
+		{
+			name:       "second dequeueing step reports success",
+			step:       2,
+			wantStatus: sysreport.JobSuccess,
+		},
+		{
+			name:       "final dequeueing step reports done",
+			step:       3,
+			wantStatus: sysreport.JobDone,
+		},
 	}
 	for _, tt := range tests { //nolint:govet
 		t.Run(tt.name, func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


rootp1:
@matthyx whenever you have a chance, could you take a look at this one?

adapters/v1/backend_test.go
```diff
@@ -312,6 +312,7 @@ func TestBackendAdapter_SendStatus(t *testing.T) {
 				sendStatusFunc: func(sender *beClientV1.BaseReportSender, s string, b bool) {
 					report := sender.GetBaseReport()
 					assert.NotEqual(t, *report, tt.report) //nolint:govet
+					assert.Equal(t, statuses[tt.step], s)
```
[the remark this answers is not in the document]
rootp1: Added independent cases for JobStarted, JobSuccess, and JobDone with explicit wantStatus values instead of deriving from statuses[tt.step].

adapters/v1/backend.go
```diff
@@ -214,7 +214,7 @@ func (a *BackendAdapter) SendStatus(ctx context.Context, step int) error {
 	report.Status = statuses[step]
 
 	sender := backendClientV1.NewBaseReportSender(a.eventReceiverRestURL, &http.Client{}, a.getRequestHeaders(), report)
-	a.sendStatusFunc(sender, sysreport.JobSuccess, true)
+	a.sendStatusFunc(sender, report.Status, true)
```
matthyx: Nit, take it or leave it: `statuses[step]` reads slightly better here than `report.Status`, since it's the same value and makes the "send the status we just computed" intent local to the line.

There's also a small technical reason to prefer it: `report.Status` is read here *without* holding `report.Mutex`, while the `DoSetStatus` inside `BaseReportSender.SendStatus` writes that same field *under* the mutex. It's benign today — `reportFromContext` mints a fresh report per call and nothing else holds a reference — so this is not a real race, just an inconsistency that `statuses[step]` sidesteps for free.

Either way the fix is correct.

adapters/v1/backend_test.go
```diff
@@ -305,13 +307,29 @@ func TestBackendAdapter_SendStatus(t *testing.T) {
...
+			wantStatus: sysreport.JobDone,
+		},
 	}
 	for _, tt := range tests { //nolint:govet
 		t.Run(tt.name, func(t *testing.T) {
 			a := &BackendAdapter{
 				sendStatusFunc: func(sender *beClientV1.BaseReportSender, s string, b bool) {
 					report := sender.GetBaseReport()
 					assert.NotEqual(t, *report, tt.report) //nolint:govet
+					assert.Equal(t, tt.wantStatus, s)
```
matthyx, line 322: The whole value of this test lives inside the `sendStatusFunc` callback, so if the callback ever stops being invoked, all four subtests pass silently while asserting nothing.

Right now that's covered indirectly — if `reportFromContext` fails, `SendStatus` returns an error and the `wantErr` check below trips — but that's a coincidence of the current control flow, not a guarantee. A three-line guard makes it explicit:

```go
var called bool
a := &BackendAdapter{
    sendStatusFunc: func(sender *beClientV1.BaseReportSender, s string, b bool) {
        called = true
        assert.Equal(t, tt.wantStatus, s)
    },
}
// ... after SendStatus:
assert.True(t, called, "sendStatusFunc was never invoked")
```

adapters/v1/backend_test.go
```diff
@@ -305,13 +307,29 @@ func TestBackendAdapter_SendStatus(t *testing.T) {
...
+			step:       3,
+			wantStatus: sysreport.JobDone,
+		},
 	}
 	for _, tt := range tests { //nolint:govet
 		t.Run(tt.name, func(t *testing.T) {
 			a := &BackendAdapter{
 				sendStatusFunc: func(sender *beClientV1.BaseReportSender, s string, b bool) {
 					report := sender.GetBaseReport()
 					assert.NotEqual(t, *report, tt.report) //nolint:govet
```
matthyx: Since you're already reshaping this table: the `report` field and this `NotEqual` are now dead weight, and dropping both would be a strict improvement.

- For the three new cases `tt.report` is the zero-value `BaseReport`, so `NotEqual` against a fully-populated report can never fail — it asserts nothing.
- Even for the original `step: 1` case, an *inequality* assertion can't catch a regression; almost any change keeps it passing (`Timestamp` alone guarantees it).
- `*report` copies a `sync.Mutex`. `go vet ./adapters/v1/...` reports three lock-copy errors (`backend_test.go:326` range var, `:331` ×2) — pre-existing on `main` and papered over with `//nolint:govet`, but removing the field would delete both `nolint` directives and the vet noise along with it.

Coverage-wise you'd lose nothing: `wantStatus` is what actually pins the behaviour this PR fixes. If you'd rather keep asserting on the report contents, an `assert.Equal` on the specific fields you care about (`Status`, `Details`, `ActionID`) would be worth more than the current `NotEqual` on the whole struct — and avoids the mutex copy too.

matthyx (commented):
## Review: no blockers

The bug is real and the one-line fix is the right one. `BaseReportSender.SendStatus(status, sendReport)` calls `report.DoSetStatus(status)` *before* marshalling, so the `report.Status = statuses[step]` assignment on line 214 was unconditionally clobbered with `"success"` on the wire for every step. Passing `report.Status` restores the intent of the `statuses` table.

### Verification I ran locally (branch `58fe795`)

- **Mutation test** — reverted line 217 back to `sysreport.JobSuccess` and re-ran: 3 of the 4 subtests fail (`expected: "done" / actual: "success"`). The new test is a genuine regression guard, not a tautology.
- `go build ./...` clean; `go test ./core/services/...` green; `TestBackendAdapter_SendStatus` green (4/4).
- Only local failure is `Test_syftAdapter_CreateSBOM/registry_scan` on `no space left on device` — my sandbox, unrelated. CI on the PR is fully green.

### Behaviour-change note for maintainers (not a blocker, worth one confirmation)

`SendStatus` is only reached from `ScanRegistry` (`core/services/scan.go:555/596/609`). Registry scans previously emitted `status="success"` three times per scan; they now emit `"started"`, `"success"`, `"done"`. All three are legitimate `systemreports` constants, so the event-receiver should accept them — but if any dashboard or metric counts `status == "success"` events per scan, that count drops 3 → 1. Worth a sanity check on the backend side before merging.

Three non-blocking comments inline.

### Out of scope / pre-existing, just flagging

- `details[step]` / `statuses[step]` panic on `step >= 4`; `SendStatus(ctx, step)` takes a bare `int` with no bound check. Not this PR's job.
- `step: 0` (`domain.Accepted`) has no production caller. Harmless to cover — it keeps the table honest.

matthyx:
cc @kooomix does our backend want 3 times `status="success"` or it doesn't matter?

adapters/v1/backend.go
rootp1: @matthyx switched to statuses[step] directly, avoids the unsynchronized read too.

adapters/v1/backend_test.go
rootp1: @matthyx added a called guard so the test fails loudly if sendStatusFunc stops being invoked.

adapters/v1/backend_test.go
rootp1: @matthyx dropped the report field and the NotEqual check, the vet lock-copy warnings are gone with it.

rootp1:
@matthyx pushed the requested changes, could you take another look when you have a moment?

matthyx (commented):
Re-checked at `6f15938`. All three follow-ups are correct — thanks for the quick turnaround. Still no blockers.

**1. `statuses[step]` at the call site** ✅ — the unsynchronized read of `report.Status` is gone.

**2. `called` guard** ✅ — verified it actually bites. I stubbed out the `sendStatusFunc` call in `SendStatus` and all four subtests fail with `sendStatusFunc was never invoked`, instead of passing vacuously. That's exactly the failure mode it was meant to close.

**3. Dropped `report` + `assert.NotEqual`** ✅ — and this is a strict improvement over `main`: `go vet ./adapters/v1/...` is now **clean**. On `main` it reports three lock-copy errors (`backend_test.go:309` range var, `:314` ×2); both `//nolint:govet` directives went away with the field.

### Verification

- **Mutation test A** — reverted line 217 to the hardcoded `sysreport.JobSuccess`: 3 of 4 subtests fail (`expected: "done" / actual: "success"`). Still a real regression guard.
- **Mutation test B** — made `sendStatusFunc` unreachable: 4 of 4 fail on the new `called` assertion.
- `go build ./...` clean, `go vet ./adapters/v1/...` clean, `TestBackendAdapter_SendStatus` 4/4, `go test ./core/services/...` green.

CI on this commit: `Create cross-platform build`, `CodeQL`, `DCO`, `build-and-push-image` and CodeRabbit have passed; `Basic-Test` and the three FOSSA checks were still running when I looked (they were green on the previous commit).

Good to merge from my side once CI finishes. The only thing I'd still like a second opinion on is the behaviour note from my first review, which this round doesn't change: registry scans now emit `"started"` / `"success"` / `"done"` where they previously emitted `"success"` three times, so any backend metric counting `success` events per scan drops 3 → 1. That's the intended fix, just worth someone confirming nothing downstream keys on the old shape.

matthyx (approved):

kooomix:
@matthyx No — the backend doesn't want repeated `success`. The job-report ingester switches on status: `started` creates a pending "scan in progress" stub, `success`/`done` clears it, `failed` marks it as error. The hardcoded `success` meant kubevuln never created pending stubs at all, so this fix actually restores the intended lifecycle. Change is safe (and welcome) from our side.

PR #412, opened 2026-08-01
fix: StoreVEX write path lacks conflict-safe retry

### LINKED ISSUE #410: Bug: StoreVEX write path lacks conflict-safe retry, unlike every sibling Store* method
# Description

`StoreVEX` (and its helpers `createVEX`/`updateVEX`) in `repositories/apiserver.go` is the only write path in `APIServerStore` that is missing the conflict-safe write pattern (`retry.RetryOnConflict`, create-first-then-fall-back-to-update) that every sibling `Store*` method (`StoreCVE`, `StoreCVESummary`, `StoreCVESummaryStub`, `StoreSBOM`) now uses. As a result, VEX documents can silently fail to persist or drift out of sync with the CVE manifest/summary for the exact same scan, under a routine concurrency scenario: two goroutines racing to store VEX for the same image.

This is the same class of correctness issue already fixed for the other four storage methods in #396 / #399 (surfacing storage failures, adding `retry.RetryOnConflict`), but that fix pass never reached `StoreVEX`, which has a structurally different (and unsafe) shape.

`StoreVEX` (`repositories/apiserver.go:694-727`):

```go
func (a *APIServerStore) StoreVEX(ctx context.Context, cve domain.CVEManifest, cvep domain.CVEManifest, _ bool) error {
	...
	vexContainer, err := a.StorageClient.OpenVulnerabilityExchangeContainers(a.Namespace).Get(context.Background(), cvep.Name, metav1.GetOptions{})
	if err != nil {
		if errors.IsNotFound(err) {
			err = a.createVEX(ctx, cve, cvep)
			if err != nil {
				return err
			}
		} else {
			return err
		}
	} else {
		err = a.updateVEX(ctx, cve, cvep, vexContainer)
		if err != nil {
			return err
		}
	}
	return nil
}
```

**Defect 1 — TOCTOU race on create** (`repositories/apiserver.go:706-713`, `createVEX` L777-856): two goroutines scanning the same image concurrently (e.g. two replicas of the same workload, dispatched through the `workerpool.WorkerPool` in `controllers/http.go`) both `Get` the VEX container, both see `NotFound`, and both call `createVEX`, which ends with a bare `Create`:

```go
// repositories/apiserver.go:853-855
_, err = a.StorageClient.OpenVulnerabilityExchangeContainers(a.Namespace).Create(context.Background(), &vexContainer, metav1.CreateOptions{})
return err
```

The loser gets a raw `IsAlreadyExists` error returned all the way up through `StoreVEX`, with no fallback to convert it into an update — unlike `StoreCVE`/`StoreSBOM`, which explicitly handle `errors.IsAlreadyExists` by falling through to a get+update retry loop.

**Defect 2 — no conflict retry on update** (`repositories/apiserver.go:950-953`, end of `updateVEX`):

```go
vexContainer.Spec = vexDoc
_, err = a.StorageClient.OpenVulnerabilityExchangeContainers(a.Namespace).Update(context.Background(), vexContainer, metav1.UpdateOptions{})
return err
```

Single-shot `Update` using a `resourceVersion` captured earlier by `StoreVEX`'s `Get`. Every other update path in this file — `StoreCVE` (L333), `StoreCVESummary` (L571), `StoreCVESummaryStub` (L656), `StoreSBOM` (L1137, L1171) — wraps this exact call in `retry.RetryOnConflict(retry.DefaultRetry, func() error { ... })` to survive a concurrent writer bumping `resourceVersion` between `Get` and `Update`. `updateVEX` does not, so any concurrent modification of the same VEX object produces an unretried 409 Conflict.

```
$ grep -n "retry.RetryOnConflict" repositories/apiserver.go
333:		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
571:		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
656:		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
1137:			retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
1171:			retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
None of these hits are inside `createVEX`/`updateVEX` — confirmed on current `main` (`a5e45e3`).

**Defect 3 — failure is silently swallowed at the call site, causing cross-store drift** (`core/services/scan.go:365-371`):

```go
if s.vexGeneration {
    err = s.cveRepository.StoreVEX(ctx, cve, cvep, true)
    if err != nil {
        logger.L().Ctx(ctx).Warning("storing VEX", helpers.Error(err), ...)
        // no continue, storing the VEX is not critical
    }
}
```

`StoreCVE`/`StoreCVESummary` run immediately before this (`scan.go:349-364`) and *do* retry on conflict, so they routinely succeed under the same race. `StoreVEX` does not retry, so it fails and is only logged as a warning — the operator has no indication that the VEX document for that image is now stale relative to the CVE manifest/summary it just successfully wrote. VEX container names are derived from image identity (not per-pod), so this reproduces any time two workloads sharing an image are scanned within the same window — routine given `ScanConcurrency` (`cmd/http/main.go:118`) runs scans in parallel via `workerpool.WorkerPool`.

**Defect 4 — no test coverage:** `repositories/apiserver_test.go` has `_transientError`, `_updateGetFailure_`, and `_retryExhausted_` test variants for `StoreCVE`, `StoreSBOM`, `StoreCVESummary`, and `StoreCVESummaryStub` (added in the #396/#399 hardening pass), but none for `StoreVEX`/`createVEX`/`updateVEX` — no conflict test, no transient-Get-error test, no concurrent-create-race test.

# Environment
OS: `N/A — code-review finding, not runtime-specific`
Version: `main @ a5e45e3` (post #401)

# Steps To Reproduce
1. Deploy kubevuln with `vexGeneration` enabled and `ScanConcurrency > 1`.
2. Trigger scans for two or more workloads that reference the *same* container image within the same scan window (e.g. two replicas of one Deployment, or two different Deployments sharing a base image), so their `StoreVEX` calls race on the same `OpenVulnerabilityExchangeContainer` name.
3. Observe (via code inspection / instrumentation, since the race is timing-dependent):
   - In `StoreVEX` (`repositories/apiserver.go:706-713`): both goroutines `Get` the container, both observe `NotFound`, both call `createVEX`; the losing `Create` call returns `IsAlreadyExists`, which propagates unhandled out of `StoreVEX`.
   - Alternatively, if one goroutine's `Get` returns an existing object and a concurrent writer updates it before `updateVEX`'s `Update` call at `apiserver.go:951`, the `Update` returns a 409 Conflict with no retry.
4. In both cases, `core/services/scan.go:367` only logs `logger.L().Ctx(ctx).Warning("storing VEX", ...)` and continues — the CVE manifest/summary for that scan already succeeded (they retry on conflict), but the VEX document is left stale or unwritten, with no error surfaced beyond the warning log.

# Expected behavior
`StoreVEX` should be as conflict-safe as `StoreCVE`/`StoreSBOM`/`StoreCVESummary`/`StoreCVESummaryStub`: a concurrent create race should transparently fall back to an update instead of surfacing `AlreadyExists`, and the update path should retry on 409 Conflict via `retry.RetryOnConflict`, re-fetching the latest object on each attempt — so VEX documents stay consistent with the CVE manifest/summary written in the same scan, even under concurrent scans of a shared image.

# Actual Behavior
`StoreVEX`'s `Get`-then-branch shape has a TOCTOU race on create with no `AlreadyExists` fallback, and `updateVEX`'s single-shot `Update` has zero conflict retry — the only `Store*` method in `repositories/apiserver.go` without `retry.RetryOnConflict`. Failures are swallowed as a `Warning` log at the `scan.go` call site with no propagation, silently leaving the VEX document stale/missing while the CVE manifest and summary for the same scan succeed. There is also no test coverage for any of these paths, unlike the other four `Store*` methods.

# Additional context

**Root cause:** `StoreVEX` predates (or was written independently of) the create-first / catch-`AlreadyExists` / `RetryOnConflict` pattern that `StoreCVE`, `StoreCVESummary`, `StoreCVESummaryStub`, and `StoreSBOM` converged on. When those four methods were hardened for storage-failure surfacing and conflict handling (#396, #399), the VEX path was out of scope each time because the VEX bugs tracked concurrently were content-correctness bugs (#393 ID/name swap, #400 missing not_affected reset), not write-conflict handling — so the robustness pass never touched `StoreVEX`.

**Impact:** Silent, intermittent staleness/loss of VEX documents under the most common real-world scanning condition (multiple workloads sharing an image, scanned concurrently); storage-layer inconsistency between `VulnerabilityManifest`/`VulnerabilityManifestSummary` and `OpenVulnerabilityExchangeContainer` for the same image; no error surfaced beyond a warning log; and increasing risk as the external VEX feeds effort (#387) adds more concurrent writers to the same objects.

**Affected files:**
- `repositories/apiserver.go`: `StoreVEX` (L694-727), `createVEX` (L777-856), `updateVEX` (L858-954)
- `repositories/apiserver_test.go`: missing test coverage
- Caller: `core/services/scan.go` (L365-371)

**Not a duplicate of:** checked all open/closed issues and PRs #387-#409 in kubescape/kubevuln. Closest related: #393/#400 (VEX content-correctness bugs, already fixed, different bug class), #397/#390 (read-path error swallowing on `GetCVESummary`/`GetSBOM`/`GetCVE`, different methods), #403/#404 (VEX `action_statement` content gap, different bug). None address write-conflict retry on `StoreVEX`.

**Acceptance criteria:**
1. `StoreVEX` follows the same create-first, catch-`errors.IsAlreadyExists`, `retry.RetryOnConflict`-wrapped-update pattern already used by `StoreCVE`/`StoreSBOM`/`StoreCVESummary`/`StoreCVESummaryStub`.
2. A concurrent-create race (two callers both observe `NotFound`) no longer surfaces a raw `AlreadyExists` error from `StoreVEX` — the loser transparently falls through to an update.
3. The final `Update()` call in `updateVEX` is wrapped in `retry.RetryOnConflict`, re-fetching the latest object/`resourceVersion` on each retry attempt, and errors after retries are exhausted are returned (not silently dropped).
4. No change to VEX content/statement logic itself (ID/Name mapping, not_affected reset, action_statement) — that's tracked separately in #400/#393/#403/#404. This issue is scoped purely to the write-conflict robustness envelope.
5. New tests added in `repositories/apiserver_test.go` mirroring the existing suite: `TestAPIServerStore_StoreVEX_transientError`, `..._updateGetFailure_transientError`, `..._retryExhausted_transientError`, plus a test exercising the concurrent-create-race path.

1. Refactor `StoreVEX` to attempt `Create` first (mirroring `StoreCVE`'s shape), with the "build the initial VEX doc" logic currently in `createVEX` invoked on that path.
2. On `errors.IsAlreadyExists`, enter a `retry.RetryOnConflict` loop that re-`Get`s the latest object, applies the existing `updateVEX` merge/normalize logic, and `Update`s — retrying on conflict, same shape as `StoreCVE`'s update loop.
3. Preserve all existing statement-merge/normalize/not-affected-reset logic in `updateVEX` unchanged; only add the retry wrapper and correct error propagation on the final `Update`.
4. Add the missing test suite, reusing the fake-clientset reactor helpers already present in `apiserver_test.go` for the `_transientError`/`_retryExhausted_` tests on the other four `Store*` methods.
5. Update any relevant doc comments describing the VEX storage contract to note the retry-on-conflict behavior, for consistency with the other methods.

bhuvan-somisetty:
## Overview

`StoreVEX`/`createVEX`/`updateVEX` in `repositories/apiserver.go` were the only write path in `APIServerStore` missing the conflict-safe write pattern already used by every sibling `Store*` method (`StoreCVE`, `StoreCVESummary`, `StoreCVESummaryStub`, `StoreSBOM`): fall back to an update on `AlreadyExists`, and retry the update with `retry.RetryOnConflict`.

Previously, `StoreVEX` did a `Get`-then-branch: if two callers raced and both observed `NotFound`, the loser's `createVEX` call returned a raw `AlreadyExists` error with no fallback to an update. Separately, `updateVEX`'s final `Update()` call had no conflict retry at all, so any concurrent writer bumping the object's `resourceVersion` between `Get` and `Update` produced an unretried 409. Since `core/services/scan.go` only logs a warning when `StoreVEX` fails ("storing the VEX is not critical"), both failure modes silently left the VEX document out of sync with the CVE manifest/summary written in the same scan — a routine occurrence whenever two workloads share an image are scanned concurrently.

`StoreVEX` now checks for an existing container first (`Get`), same as before. On `NotFound` it calls `createVEX`; if that races and comes back `AlreadyExists`, it falls through into a `retry.RetryOnConflict` loop instead of surfacing the raw error. If the container already exists, it goes straight into the same retry loop, which re-`Get`s the latest object and calls `updateVEX` against it, retrying on conflict. This keeps the common "already exists" path cheap (no wasted full-document build/hash/POST via `createVEX` that gets thrown away), while still closing both race windows. Errors are wrapped and returned instead of being dropped. No change to VEX content/statement logic in `updateVEX`/`createVEX` (ID/Name mapping, not_affected reset, product/statement construction) — this PR is scoped purely to the write-conflict robustness envelope.

The retry loop's `Get` deliberately uses plain `GetOptions{}`, **not** `GetOptions{ResourceVersion: "metadata"}` like the sibling methods use — that variant returns an ObjectMeta-only object with a zero `Spec` in `kubescape/storage`, which the siblings can tolerate because they overwrite `Spec` wholesale. `updateVEX` instead merges into `Spec.Statements`, so a metadata-only read would silently drop every existing statement. This is called out in a comment at the call site since the fake clientset used in tests ignores `GetOptions` entirely and can't catch a regression here.

## Additional Information

This is the same class of correctness issue already fixed for the other four `Store*` methods in #396 / #399 (surfacing storage failures, adding `retry.RetryOnConflict`), but that fix pass never reached `StoreVEX`, which had a structurally different (and unsafe) shape.

A pre-existing, unrelated gap was found during review — `updateVEX` never refreshes `Labels`/`Annotations` from the latest scan, unlike its siblings which `mergeMaps` them before `Update`. That's tracked separately as #419 rather than folded into this PR.

## How to Test

```
go build ./...
go vet ./repositories/...
go test ./repositories/... -run TestAPIServerStore -v
```

All existing VEX tests plus the new ones added in this PR pass, including a test that a container created concurrently by another writer is correctly picked up via the update fallback, and a test that a single transient conflict is retried and recovered from (not just that retries eventually exhaust).

## Related issues/PRs:

* Resolved #410

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when saving VEX data, including scenarios where multiple operations attempt to create the same container simultaneously.
  - Existing containers are now refreshed and updated more consistently.
  - Added recovery for temporary conflicts during updates.
  - Improved error reporting when container creation, retrieval, or updates fail, making issues easier to diagnose.

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -700,31 +700,57 @@ func (a *APIServerStore) StoreVEX(ctx context.Context, cve domain.CVEManifest, c
 		return nil
 	}
 
-	err := a.createVEX(ctx, cve, cvep)
+	// Check for an existing container first so the common "VEX already exists" path
+	// (every scan after the first one for a given image) goes straight to the cheap
+	// update below, instead of paying for building the full VEX document, hashing it,
+	// and POSTing it via createVEX only to discard the result on AlreadyExists.
+	_, err := a.StorageClient.OpenVulnerabilityExchangeContainers(a.Namespace).Get(context.Background(), cvep.Name, metav1.GetOptions{})
 	switch {
-	case errors.IsAlreadyExists(err):
-		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-			// retrieve the latest version before attempting update
-			// RetryOnConflict uses exponential backoff to avoid exhausting the apiserver
-			vexContainer, getErr := a.StorageClient.OpenVulnerabilityExchangeContainers(a.Namespace).Get(context.Background(), cvep.Name, metav1.GetOptions{})
-			if getErr != nil {
-				return getErr
-			}
-			return a.updateVEX(ctx, cve, cvep, vexContainer)
-		})
-		if retryErr != nil {
-			logger.L().Debug("failed to update VEX in storage", helpers.Error(retryErr),
+	case errors.IsNotFound(err):
+		err = a.createVEX(ctx, cve, cvep)
+		switch {
+		case err == nil:
+			logger.L().Debug("stored VEX in storage", helpers.String("name", cvep.Name))
+			return nil
+		case errors.IsAlreadyExists(err):
+			// A concurrent writer created the container between our Get and our
+			// Create; fall through to the retry-on-conflict update path below
+			// instead of surfacing the raw AlreadyExists error.
+		default:
+			logger.L().Debug("failed to store VEX in storage", helpers.Error(err),
 				helpers.String("name", cvep.Name))
-			return fmt.Errorf("failed to update VEX in storage: %w", retryErr)
+			return fmt.Errorf("failed to store VEX in storage: %w", err)
 		}
-		logger.L().Debug("updated VEX in storage", helpers.String("name", cvep.Name))
 	case err != nil:
-		logger.L().Debug("failed to store VEX in storage", helpers.Error(err),
+		logger.L().Debug("failed to get VEX from storage", helpers.Error(err),
 			helpers.String("name", cvep.Name))
-		return fmt.Errorf("failed to store VEX in storage: %w", err)
-	default:
-		logger.L().Debug("stored VEX in storage", helpers.String("name", cvep.Name))
+		return fmt.Errorf("failed to get VEX from storage: %w", err)
 	}
+
+	retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		// retrieve the latest version before attempting update
+		// RetryOnConflict uses exponential backoff to avoid exhausting the apiserver
+		//
+		// NOTE: this must NOT use GetOptions{ResourceVersion: "metadata"} like the
+		// sibling Store* methods do. That option returns an ObjectMeta-only object
+		// with a zero Spec in kubescape/storage's apiserver, which is safe for the
+		// siblings because they overwrite Spec wholesale. updateVEX instead merges
+		// into vexContainer.Spec.Statements, so a metadata-only read would silently
+		// drop every previously stored statement and then fail when it tries to
+		// parse the zeroed Spec.Metadata.Timestamp. The fake clientset used in tests
+		// ignores GetOptions entirely, so no test can catch a regression here.
+		vexContainer, getErr := a.StorageClient.OpenVulnerabilityExchangeContainers(a.Namespace).Get(context.Background(), cvep.Name, metav1.GetOptions{})
+		if getErr != nil {
+			return getErr
+		}
+		return a.updateVEX(ctx, cve, cvep, vexContainer)
+	})
+	if retryErr != nil {
+		logger.L().Debug("failed to update VEX in storage", helpers.Error(retryErr),
+			helpers.String("name", cvep.Name))
+		return fmt.Errorf("failed to update VEX in storage: %w", retryErr)
+	}
+	logger.L().Debug("updated VEX in storage", helpers.String("name", cvep.Name))
 
 	return nil
 }
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -706,7 +706,7 @@ func (a *APIServerStore) StoreVEX(ctx context.Context, cve domain.CVEManifest, c
 		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
 			// retrieve the latest version before attempting update
 			// RetryOnConflict uses exponential backoff to avoid exhausting the apiserver
-			vexContainer, getErr := a.StorageClient.OpenVulnerabilityExchangeContainers(a.Namespace).Get(context.Background(), cvep.Name, metav1.GetOptions{ResourceVersion: "metadata"})
+			vexContainer, getErr := a.StorageClient.OpenVulnerabilityExchangeContainers(a.Namespace).Get(context.Background(), cvep.Name, metav1.GetOptions{})
 			if getErr != nil {
 				return getErr
 			}
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -704,7 +704,10 @@ func (a *APIServerStore) StoreVEX(ctx context.Context, cve domain.CVEManifest, c
 	// (every scan after the first one for a given image) goes straight to the cheap
 	// update below, instead of paying for building the full VEX document, hashing it,
 	// and POSTing it via createVEX only to discard the result on AlreadyExists.
-	_, err := a.StorageClient.OpenVulnerabilityExchangeContainers(a.Namespace).Get(context.Background(), cvep.Name, metav1.GetOptions{})
+	existing, err := a.StorageClient.OpenVulnerabilityExchangeContainers(a.Namespace).Get(ctx, cvep.Name, metav1.GetOptions{})
+	// Get returns a non-nil (but empty) object even on error, so success must be
+	// tracked explicitly rather than by checking the returned pointer for nil.
+	haveExisting := err == nil
 	switch {
 	case errors.IsNotFound(err):
 		err = a.createVEX(ctx, cve, cvep)
@@ -728,20 +731,29 @@ func (a *APIServerStore) StoreVEX(ctx context.Context, cve domain.CVEManifest, c
 	}
 
 	retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-		// retrieve the latest version before attempting update
-		// RetryOnConflict uses exponential backoff to avoid exhausting the apiserver
-		//
-		// NOTE: this must NOT use GetOptions{ResourceVersion: "metadata"} like the
-		// sibling Store* methods do. That option returns an ObjectMeta-only object
-		// with a zero Spec in kubescape/storage's apiserver, which is safe for the
-		// siblings because they overwrite Spec wholesale. updateVEX instead merges
-		// into vexContainer.Spec.Statements, so a metadata-only read would silently
-		// drop every previously stored statement and then fail when it tries to
-		// parse the zeroed Spec.Metadata.Timestamp. The fake clientset used in tests
-		// ignores GetOptions entirely, so no test can catch a regression here.
-		vexContainer, getErr := a.StorageClient.OpenVulnerabilityExchangeContainers(a.Namespace).Get(context.Background(), cvep.Name, metav1.GetOptions{})
-		if getErr != nil {
-			return getErr
+		// Reuse the object already fetched above on the first attempt, so the common
+		// "VEX already exists" path only costs a single Get. Any retry (after a real
+		// conflict) forces a fresh Get for the latest resourceVersion.
+		vexContainer := existing
+		alreadyHadExisting := haveExisting
+		haveExisting = false
+		if !alreadyHadExisting {
+			// retrieve the latest version before attempting update
+			// RetryOnConflict uses exponential backoff to avoid exhausting the apiserver
+			//
+			// NOTE: this must NOT use GetOptions{ResourceVersion: "metadata"} like the
+			// sibling Store* methods do. That option returns an ObjectMeta-only object
+			// with a zero Spec in kubescape/storage's apiserver, which is safe for the
+			// siblings because they overwrite Spec wholesale. updateVEX instead merges
+			// into vexContainer.Spec.Statements, so a metadata-only read would silently
+			// drop every previously stored statement and then fail when it tries to
+			// parse the zeroed Spec.Metadata.Timestamp. The fake clientset used in tests
+			// ignores GetOptions entirely, so no test can catch a regression here.
+			var getErr error
+			vexContainer, getErr = a.StorageClient.OpenVulnerabilityExchangeContainers(a.Namespace).Get(ctx, cvep.Name, metav1.GetOptions{})
+			if getErr != nil {
+				return getErr
+			}
 		}
 		return a.updateVEX(ctx, cve, cvep, vexContainer)
 	})
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -700,30 +700,33 @@ func (a *APIServerStore) StoreVEX(ctx context.Context, cve domain.CVEManifest, c
 		return nil
 	}
 
-	// Check if VEX already exists
-	// If it does, update it
-	// If it doesn't, create it
-	vexContainer, err := a.StorageClient.OpenVulnerabilityExchangeContainers(a.Namespace).Get(context.Background(), cvep.Name, metav1.GetOptions{})
-	if err != nil {
-		if errors.IsNotFound(err) {
-			// Create VEX
-			err = a.createVEX(ctx, cve, cvep)
-			if err != nil {
-				return err
+	err := a.createVEX(ctx, cve, cvep)
+	switch {
+	case errors.IsAlreadyExists(err):
+		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			// retrieve the latest version before attempting update
+			// RetryOnConflict uses exponential backoff to avoid exhausting the apiserver
+			vexContainer, getErr := a.StorageClient.OpenVulnerabilityExchangeContainers(a.Namespace).Get(context.Background(), cvep.Name, metav1.GetOptions{ResourceVersion: "metadata"})
+			if getErr != nil {
+				return getErr
 			}
-		} else {
-			return err
+			return a.updateVEX(ctx, cve, cvep, vexContainer)
+		})
+		if retryErr != nil {
+			logger.L().Debug("failed to update VEX in storage", helpers.Error(retryErr),
+				helpers.String("name", cvep.Name))
+			return fmt.Errorf("failed to update VEX in storage: %w", retryErr)
 		}
-	} else {
-		// Update VEX
-		err = a.updateVEX(ctx, cve, cvep, vexContainer)
-		if err != nil {
-			return err
-		}
+		logger.L().Debug("updated VEX in storage", helpers.String("name", cvep.Name))
+	case err != nil:
+		logger.L().Debug("failed to store VEX in storage", helpers.Error(err),
+			helpers.String("name", cvep.Name))
+		return fmt.Errorf("failed to store VEX in storage: %w", err)
+	default:
+		logger.L().Debug("stored VEX in storage", helpers.String("name", cvep.Name))
 	}
 
 	return nil
-
 }
 
 func createProductStructForImageAndPackage(imagePullable string, packagePURL string) (*v1beta1.Product, error) {
--- a/repositories/apiserver_test.go
+++ b/repositories/apiserver_test.go
@@ -1291,4 +1291,7 @@ func TestAPIServerStore_StoreVEX_concurrentCreateRace(t *testing.T) {
 	vexContainer, err := a.StorageClient.OpenVulnerabilityExchangeContainers(a.Namespace).Get(context.Background(), name, metav1.GetOptions{})
 	require.NoError(t, err)
 	require.NotNil(t, vexContainer)
+	// updateVEX bumps Metadata.Version on every successful update, so a version > 0
+	// confirms the fallback actually went through updateVEX rather than a no-op.
+	require.Greater(t, vexContainer.Spec.Metadata.Version, int64(0))
 }
--- a/repositories/apiserver_test.go
+++ b/repositories/apiserver_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"testing"
+	"time"
 
 	helpersv1 "github.com/kubescape/k8s-interface/instanceidhandler/v1/helpers"
 	"github.com/kubescape/kubevuln/core/domain"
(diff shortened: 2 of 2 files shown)


bhuvan-somisetty:
Thanks for the review! Pushed a fix for both points:
- switched the retry-loop's re-fetch to plain `GetOptions{}` instead of the synthetic `ResourceVersion: "metadata"` value
- strengthened the concurrent-create-race test to assert `Spec.Metadata.Version > 0`, so it actually verifies the fallback went through `updateVEX` and not just that the object exists

All existing + new tests still pass locally.

bhuvan-somisetty:
Hey @matthyx @rotemamsa @slashben @kooomix — could one of you take a look when you get a chance? This addresses issue #410 (StoreVEX missing the same conflict-retry safety net that StoreCVE/StoreSBOM/StoreCVESummary already have). Already responded to CodeRabbit's feedback above. Thanks!

repositories/apiserver.go
```diff
@@ -700,30 +700,33 @@ func (a *APIServerStore) StoreVEX(ctx context.Context, cve domain.CVEManifest, c
...
-			err = a.createVEX(ctx, cve, cvep)
-			if err != nil {
-				return err
+	err := a.createVEX(ctx, cve, cvep)
+	switch {
+	case errors.IsAlreadyExists(err):
+		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			// retrieve the latest version before attempting update
+			// RetryOnConflict uses exponential backoff to avoid exhausting the apiserver
+			vexContainer, getErr := a.StorageClient.OpenVulnerabilityExchangeContainers(a.Namespace).Get(context.Background(), cvep.Name, metav1.GetOptions{})
```
matthyx: **🟠 Major — this line is load-bearing, deliberately inconsistent with its four siblings, and completely unguarded. Please add a comment explaining why.**

Commit `fbe9979` was right to drop `ResourceVersion: "metadata"` here, but nothing in the tree records *why*, and no test can catch its reintroduction. This is a regression waiting to happen the next time someone tidies these five methods for consistency.

Why it matters — `ResourceVersion: "metadata"` is not a hint, it is a different read in `kubescape/storage`:

```go
// kubescape/storage@v0.0.258 pkg/registry/file/storage.go:422
if opts.ResourceVersion == softwarecomposition.ResourceVersionMetadata {
    metadata, err := ReadMetadata(conn, key)   // SQLite ObjectMeta only
    ...
    return json.Unmarshal(metadata, objPtr)    // <- returns with a ZERO Spec
}
```

`StoreCVE` / `StoreCVESummary` / `StoreCVESummaryStub` / `StoreSBOM` can all use it safely because they *overwrite* `result.Spec = manifest.Spec`. `updateVEX` is the odd one out: it reads `vexDoc := vexContainer.Spec` and **merges** into the existing statements. A metadata-only read hands it a zero `Spec`.

I confirmed the failure mode locally by feeding `updateVEX` an ObjectMeta-only container (what a `"metadata"` Get returns):

```
statements after create: 456
updateVEX with metadata-only object -> err=parsing time "" as "2006-01-02T15:04:05Z07:00": cannot parse "" as "2006", statements=0
```

All 456 statements dropped, and `calculateVexCanonicalHash` then hard-fails on the empty `Spec.Metadata.Timestamp` — so every VEX update would fail permanently. Since `core/services/scan.go:366` only logs a warning, that would be silent.

And the tests cannot catch it: `k8s.io/client-go/testing`'s fake clientset ignores `GetOptions` entirely, so `GetOptions{}` and `GetOptions{ResourceVersion: "metadata"}` are indistinguishable to every test in this file.

Suggested:

```go
// NOTE: unlike the sibling Store* methods, this Get must NOT use
// GetOptions{ResourceVersion: "metadata"} — that returns an ObjectMeta-only
// object with a zero Spec. updateVEX merges into the existing Spec.Statements
// (the siblings overwrite Spec wholesale), so a metadata-only read would drop
// every stored statement and then fail on the empty Spec.Metadata.Timestamp.
// The fake clientset ignores GetOptions, so no test will catch this.
vexContainer, getErr := a.StorageClient.OpenVulnerabilityExchangeContainers(a.Namespace).Get(context.Background(), cvep.Name, metav1.GetOptions{})
```

Minor, while you're on this line: the sibling code passes `context.Background()` too, so this is consistent — but `ctx` is right there and unused.

repositories/apiserver.go
```diff
@@ -700,30 +700,33 @@ func (a *APIServerStore) StoreVEX(ctx context.Context, cve domain.CVEManifest, c
...
-	// If it does, update it
-	// If it doesn't, create it
-	vexContainer, err := a.StorageClient.OpenVulnerabilityExchangeContainers(a.Namespace).Get(context.Background(), cvep.Name, metav1.GetOptions{})
-	if err != nil {
-		if errors.IsNotFound(err) {
-			// Create VEX
-			err = a.createVEX(ctx, cve, cvep)
-			if err != nil {
-				return err
+	err := a.createVEX(ctx, cve, cvep)
```
matthyx: **🟡 Minor / worth a maintainer decision — new cost on the hot path.**

Create-first means that on the common case (image already scanned, VEX exists), every scan now:

1. builds a full `v1beta1.Statement` per entry in `cve.Content.Matches` (456 for the `nginx-cve.json` fixture, and real images go well beyond that),
2. runs `markRelevantVulnerabilitiesAsAffectedInVex` + `calculateVexCanonicalHash` (sort + SHA-256 over the whole statement list),
3. POSTs the entire document to the apiserver,

…and throws all of it away on `AlreadyExists`, then does the full `Get` + `Update` anyway. The old Get-first shape paid none of that on the update path.

I don't think it blocks: `StoreCVE`/`StoreSBOM` already eat the same wasted full-size `Create` on every scan, so this is consistent with the established pattern rather than a new class of cost — and consistency is the explicit goal of the PR. But it *is* strictly more work than before on the most frequent path, and unlike the siblings this one also pays a full-payload `Get` (see the other comment) rather than a cheap metadata read. Flagging so it's a conscious tradeoff rather than an accident.

If you'd rather keep the fast path, `Get`-first → on `NotFound` `createVEX` → on `AlreadyExists` from that create, fall into the same `RetryOnConflict` block gets both. Happy either way — just noting it wasn't discussed.

repositories/apiserver_test.go
```diff
@@ -1211,3 +1212,86 @@ func TestAPIServerStore_StoreCVESummaryStub_retryExhausted_transientError(t *tes
...
+	var updateCalls int
+	clientset.PrependReactor("update", "openvulnerabilityexchangecontainers", func(k8stesting.Action) (bool, runtime.Object, error) {
+		updateCalls++
+		return true, nil, apierrors.NewConflict(schema.GroupResource{Resource: "openvulnerabilityexchangecontainers"}, name, fmt.Errorf("conflict"))
+	})
+	a := &APIServerStore{StorageClient: clientset.SpdxV1beta1(), Namespace: "kubescape"}
+	cve := domain.CVEManifest{Name: name, Content: &v1beta1.GrypeDocument{}}
+	err := a.StoreVEX(context.TODO(), cve, cve, false)
+	require.True(t, apierrors.IsConflict(err))
+	require.Equal(t, retry.DefaultRetry.Steps, updateCalls)
```
matthyx, line 1273: **🟡 Minor — the headline behaviour of this PR isn't actually asserted anywhere.**

The three new error tests all assert *failure* paths (create fails, get fails, retries exhausted), and `concurrentCreateRace` exercises a conflict-**free** update. Nothing proves that `RetryOnConflict` *recovers*: conflict once → retry → succeed.

That's the behaviour the PR title promises, and it's the one a future refactor is most likely to break (e.g. moving the `Get` outside the closure would still pass all four new tests, but would then retry forever against a stale `resourceVersion`).

```go
func TestAPIServerStore_StoreVEX_recoversFromTransientConflict(t *testing.T) {
	// ...seed as in retryExhausted...
	var updateCalls int
	clientset.PrependReactor("update", "openvulnerabilityexchangecontainers", func(k8stesting.Action) (bool, runtime.Object, error) {
		updateCalls++
		if updateCalls == 1 {
			return true, nil, apierrors.NewConflict(schema.GroupResource{Resource: "openvulnerabilityexchangecontainers"}, name, fmt.Errorf("conflict"))
		}
		return false, nil, nil // fall through to the tracker
	})
	require.NoError(t, a.StoreVEX(context.TODO(), cve, cve, false))
	require.Equal(t, 2, updateCalls)
}
```

Note the four sibling `Store*` methods have the exact same gap, so this is a nice-to-have rather than something I'd hold the PR on.

matthyx (commented):
## Review: `fix: StoreVEX write path lacks conflict-safe retry`

I checked this out, built it, and ran the suite locally:

```
go build ./...            # ok
go vet ./repositories/... # ok
go test ./repositories/ -run 'VEX|TestAPIServerStore' -count=1  # ok (0.7s)
```

**Verdict: no merge-blocking correctness defect.** The change is a genuine improvement and is behaviourally equivalent to the old code on both non-racing paths, while fixing the two race paths the description calls out:

| scenario | before | after |
|---|---|---|
| object absent, no race | Get→NotFound→Create | Create | 
| object present, no race | Get→update (no retry) | Create→AlreadyExists→Get→update | 
| two creators race | loser returns raw `AlreadyExists`, data dropped | loser falls back to update ✅ |
| concurrent writer bumps RV | unretried 409 | `RetryOnConflict` ✅ |

One **major** point below is about protecting a load-bearing detail that commit `fbe9979` got right but that nothing in the tree currently guards. The rest are minor / follow-up.

---

### Two things for the PR description (not the code)

1. **The description is now stale and says the opposite of the code.** It still claims the retry loop re-`Get`s "using `ResourceVersion: \"metadata\"` for a consistent read, matching the other `Store*` methods". Commit `fbe9979` correctly reversed that, and the `"metadata"` variant would have been a data-loss bug (see the inline comment on `apiserver.go`). Please update the description so the merge commit doesn't record the wrong rationale.

2. **Follow-up, out of scope here:** unlike all four siblings, the `StoreVEX` update path still never refreshes `Labels`/`Annotations` from `cvep` — the siblings all `mergeMaps(result.Annotations, manifest.Annotations)` before `Update`. `updateVEX` only touches `Spec`, so annotations on an existing VEX container are frozen at whatever the first-ever scan wrote. Pre-existing, and I agree it's outside this PR's stated scope, but now that `StoreVEX` is being aligned with the sibling pattern it's a visible loose end — worth a follow-up issue rather than scope creep here.

repositories/apiserver_test.go
```diff
@@ -1211,3 +1212,86 @@ func TestAPIServerStore_StoreCVESummaryStub_retryExhausted_transientError(t *tes
...
+}
+
+// TestAPIServerStore_StoreVEX_concurrentCreateRace guards against a regression where a
+// caller that loses a concurrent create race (the VEX container was created by a concurrent
+// writer between this caller's earlier NotFound check and its own Create call) received the
```
matthyx: **🔵 Nit — stale wording.** "between this caller's earlier NotFound check and its own Create call" describes the pre-fix code; after this PR there is no NotFound check, the `Create` *is* the first call. The regression being guarded against is still accurately described, but a reader hitting this comment in six months will go looking for a `Get` that isn't there.

Maybe: *"…where a caller whose `Create` lost the race (another writer created the container first) received the raw `AlreadyExists` error instead of falling back to an update."*

Otherwise this test is good — asserting `Spec.Metadata.Version > 0` rather than just non-nil is exactly the right call.

bhuvan-somisetty:
Thanks @matthyx, this was a great review appreciate you actually reproducing the metadata-only-Get failure instead of just flagging it.

Went back to a Get-first shape for `StoreVEX` so we're not building and hashing a full VEX doc just to throw it away on `AlreadyExists`; a concurrent writer winning the race still falls through into the retry-on-conflict block instead of surfacing the raw error. Added the comment you suggested explaining why the `Get` can't use `ResourceVersion: "metadata"`, added a test that actually proves the retry loop recovers from a conflict (not just that it eventually gives up), and rewrote `concurrentCreateRace` to genuinely race a `Create` now that we're back to Get-first. PR description is updated too.

Filed the Labels/Annotations gap as #419 rather than folding it in here.

Everything still builds and passes locally. Let me know if anything else needs another pass.

repositories/apiserver.go
```diff
@@ -700,30 +700,71 @@ func (a *APIServerStore) StoreVEX(ctx context.Context, cve domain.CVEManifest, c
...
 		}
-	} else {
-		// Update VEX
-		err = a.updateVEX(ctx, cve, cvep, vexContainer)
-		if err != nil {
-			return err
+	case err != nil:
+		logger.L().Debug("failed to get VEX from storage", helpers.Error(err),
+			helpers.String("name", cvep.Name))
+		return fmt.Errorf("failed to get VEX from storage: %w", err)
```
matthyx, line 730: **🔵 Nit — this is the one new error branch without a test.**

Every other error path in this PR has one (`transientError` covers the create failure, `updateGetFailure_transientError` the retry-loop `Get`, `retryExhausted_transientError` the conflict exhaustion). This branch is new in `7f8c423` — it didn't exist in the Create-first shape — and nothing exercises it.

Coverage confirms it's the only uncovered block in `StoreVEX` (`StoreVEX` sits at 87.1%; the other uncovered block is the pre-existing `cvep.Name == ""` guard):

```
lines 722-725  count=1
lines 727-730  count=0   <-- this branch
lines 733-740  count=1
```

It's a handful of lines to close, mirroring the existing `transientError` test but reacting on `get` instead of `create`:

```go
func TestAPIServerStore_StoreVEX_existenceCheckFailure_transientError(t *testing.T) {
	clientset := fake.NewSimpleClientset()
	injectedErr := apierrors.NewInternalError(fmt.Errorf("etcd timeout"))
	clientset.PrependReactor("get", "openvulnerabilityexchangecontainers", func(k8stesting.Action) (bool, runtime.Object, error) {
		return true, nil, injectedErr
	})
	a := &APIServerStore{StorageClient: clientset.SpdxV1beta1(), Namespace: "kubescape"}
	cve := domain.CVEManifest{Name: name, Content: &v1beta1.GrypeDocument{}}
	require.ErrorIs(t, a.StoreVEX(context.TODO(), cve, cve, false), injectedErr)
}
```

Worth having: a transient failure on the existence check is the most likely of the three to actually happen in production, and right now nothing pins that it returns rather than falling through to the update path.

repositories/apiserver_test.go
```diff
@@ -1211,3 +1212,131 @@ func TestAPIServerStore_StoreCVESummaryStub_retryExhausted_transientError(t *tes
...
+		return false, nil, nil // fall through to the tracker's default update handling
+	})
+	a := &APIServerStore{StorageClient: clientset.SpdxV1beta1(), Namespace: "kubescape"}
+	cve := domain.CVEManifest{Name: name, Content: &v1beta1.GrypeDocument{}}
+	require.NoError(t, a.StoreVEX(context.TODO(), cve, cve, false))
+	require.Equal(t, 2, updateCalls)
+
+	vexContainer, err := a.StorageClient.OpenVulnerabilityExchangeContainers(a.Namespace).Get(context.Background(), name, metav1.GetOptions{})
+	require.NoError(t, err)
+	require.Greater(t, vexContainer.Spec.Metadata.Version, int64(0))
```
matthyx, line 1341: **🔵 Nit — the doc comment above promises more than this assertion delivers.**

It says this test guards against *"a future refactor hoists the `Get` outside the closure and keeps retrying against a stale resourceVersion"*. I injected exactly that regression and **this test still passes**. The fake `ObjectTracker` enforces no `resourceVersion` precondition on `Update`, so the stale-object retry succeeds; and since `updateVEX` mutates its argument in place, the reused object just gets bumped twice (0→1→2), which `> 0` happily accepts.

The regression *is* caught — by `updateGetFailure_transientError` and `concurrentCreateRace` — just not here. So the guarantee is covered; the comment is only attributing it to the wrong test, which is the same trap as the stale comment fixed earlier in this PR.

Cheapest fix is to make the assertion exact, which makes the comment true as written:

```suggestion
	require.Equal(t, int64(1), vexContainer.Spec.Metadata.Version)
```

I verified both directions: it passes on the current code (fresh `Get` on retry → `Version == 1`) and fails under the hoisted-`Get` regression (`expected: 1, actual: 2`). Alternatively just soften the comment to what the test really asserts — that a single transient conflict is absorbed and the retry succeeds.

repositories/apiserver.go
```diff
@@ -700,30 +700,71 @@ func (a *APIServerStore) StoreVEX(ctx context.Context, cve domain.CVEManifest, c
...
-			if err != nil {
-				return err
-			}
-		} else {
-			return err
+	// Check for an existing container first so the common "VEX already exists" path
+	// (every scan after the first one for a given image) goes straight to the cheap
+	// update below, instead of paying for building the full VEX document, hashing it,
+	// and POSTing it via createVEX only to discard the result on AlreadyExists.
+	existing, err := a.StorageClient.OpenVulnerabilityExchangeContainers(a.Namespace).Get(ctx, cvep.Name, metav1.GetOptions{})
```
matthyx, line 707: **🔵 Nit — the `ctx` migration is half-done within this one operation.**

Both `Get`s in `StoreVEX` now take `ctx`, but the writes they exist to set up still don't: `createVEX` calls `Create(context.Background(), ...)` (`apiserver.go:894`) and `updateVEX` calls `Update(context.Background(), ...)` (`apiserver.go:992`).

So a single logical store now has reads that honor cancellation and writes that don't. Harmless today — I checked, every entry point reaches here through `context.WithoutCancel` — but if a deadline is ever introduced upstream, the failure mode is lopsided: the existence check aborts while the write that would have followed it would have gone through regardless.

Not worth holding the PR for, but I'd pick one: thread `ctx` into the `Create`/`Update` calls in `createVEX`/`updateVEX` too, or leave all four on `context.Background()` and keep this method uniform with its four siblings. Right now it's the only `Store*` method that mixes both.

matthyx (approved):
## Re-review of `d7c80fb` — all four points addressed, ✅ LGTM

Re-checked out at `d7c80fb` and re-verified locally:

```
go build ./...                          # ok
go vet ./repositories/...               # ok
go test ./repositories/... -count=1     # ok
```

Everything I raised is resolved, and the Get-first restructure in `7f8c423` is correct — I traced it rather than taking it on faith:

- The create-race fallback survives the rewrite: `Get`→`NotFound`→`createVEX`→`AlreadyExists` explicitly falls through to the retry-update path rather than surfacing the raw error. That was the whole point of the PR and it would have been easy to lose here.
- The `haveExisting` one-shot cache is right, and it *matters*: `updateVEX` mutates its argument in place (`vexContainer.Spec = vexDoc`, `Metadata.Version += 1`), so reusing the cached object across retries would double-bump `Version` and re-mutate the statement slice. Attempt 1 uses the cached object, every retry re-`Get`s. Correct.
- I confirmed the guard is real by injecting the exact regression (forcing the closure to never re-`Get`): `updateGetFailure_transientError` and `concurrentCreateRace` both fail. Good.
- Using `ctx` instead of `context.Background()` checks out. I verified all three scan entry points (`GenerateSBOM`, `ScanCP`, `ScanCVE`) are HTTP handlers that go through `context.WithoutCancel` in `controllers/http.go`, and that #374 (2026-06-16) post-dates the older defensive comment at `apiserver.go:113` (2026-04-17) — so that comment is pre-#374 history, not a live counterexample. The reasoning in the commit message holds.
- Description updated with the correct `GetOptions{}` rationale, and #419 filed for the labels/annotations gap. 👍

Two nits below, neither blocking. Nice work on this one — the `Tracker().Create()` trick in `concurrentCreateRace` (sidestepping the `Fake`'s non-reentrant action lock) is a genuinely good touch, and the comment explaining *why* is the kind of thing that saves the next person an hour.

Issue #411, opened 2026-08-01
fix: panic in optionsFromWorkload on non-boolean registry attribute values


AdeshDeshmukh:
## Description

`optionsFromWorkload` in `core/services/scan.go` (lines 680–685) performs unguarded type assertions on two values read from `workload.Args`:

```go
if useHTTP, ok := workload.Args[domain.AttributeUseHTTP]; ok {
    options.InsecureUseHTTP = useHTTP.(bool)   // panics if value is not a bool
}
if skipTLSVerify, ok := workload.Args[domain.AttributeSkipTLSVerify]; ok {
    options.InsecureSkipTLSVerify = skipTLSVerify.(bool)
}
```

`workload.Args` is a `map[string]interface{}` populated directly from the scan request JSON. The `ok` check only confirms the key exists — it says nothing about the underlying value's type. A client sending `"attribute.use.http": "true"` as a JSON string (or a number, or `null`) triggers a runtime panic:

```text
panic: interface conversion: interface {} is string, not bool
```

**Why this is severe:** `optionsFromWorkload` executes inside the `workerPool.Submit` goroutines for `GenerateSBOM`, `ScanCP`, `ScanCVE`, and `ScanRegistry`. While `gin.Recovery()` (cmd/http/main.go:122) catches panics on the Gin handler goroutine, it **does not reach worker-pool goroutines**. Consequently, a single malformed payload crashes the entire `kubevuln` pod and drops all concurrent in-flight scans.

The safe double comma-ok pattern (`val, ok := x.(bool)`) is already used elsewhere in this codebase (e.g., `domain_to_armo.go`:81–94); this is the sole location asserting blindly.

---

## Environment

- **OS:** Any
- **Version:** `main` (Aug 2026)

---

## Steps To Reproduce

1. Send a `ScanCP`, `ScanCVE`, or `ScanRegistry` request whose `args` map contains `"attribute.use.http": "true"` (a JSON string instead of a JSON boolean).
2. Observe the worker goroutine panic with `interface conversion: interface {} is string, not bool`.
3. Because the panic fires outside the Gin HTTP handler, the process exits and all in-flight scans are terminated.

---

## Expected Behavior

Non-boolean values should be safely ignored (ideally with a structured warning so operators can detect misconfigured clients) rather than crashing the pod. Valid boolean values should continue working as expected.

---

## Actual Behavior

A panic occurs inside a worker-pool goroutine, crashing the entire `kubevuln` process and dropping all concurrent scans.

[...]

PR #413, opened 2026-08-01
fix: don't panic on non-boolean registry options in optionsFromWorkload

AdeshDeshmukh:
**Problem**

`optionsFromWorkload` (core/services/scan.go) reads the registry options from the workload's `Args` map and asserts them as `bool` without first checking their runtime type:

```go
if useHTTP, ok := workload.Args[domain.AttributeUseHTTP]; ok {
    options.InsecureUseHTTP = useHTTP.(bool)         // panics if value is not a bool
}
if skipTLSVerify, ok := workload.Args[domain.AttributeSkipTLSVerify]; ok {
    options.InsecureSkipTLSVerify = skipTLSVerify.(bool)
}
```

`Args` is decoded from the scan request JSON into `map[string]interface{}`, and the `ok` in the lookup only confirms the key exists — not the value's type. A client sending `"attribute.use.http": "true"` as a JSON string (or a number, or `null`) triggers:

```
panic: interface conversion: interface {} is string, not bool
```

This is more than a normal panic: `optionsFromWorkload` runs inside the `workerPool.Submit` goroutines for `GenerateSBOM`, `ScanCP`, `ScanCVE`, and `ScanRegistry`, which sit **outside** the scope of `gin.Recovery()` (cmd/http/main.go:122). The panic is unrecovered, terminating the whole kubevuln process and dropping every concurrent in-flight scan.

**Change**

- `optionsFromWorkload` now takes `ctx` and uses the double comma-ok pattern already used in this codebase (e.g. `domain_to_armo.go:81-94`).
- When a value is present but not a `bool`, it is skipped and a structured warning is logged with the offending value's type (`%T`) and image slug — so operators can spot misconfigured clients instead of the pod silently changing behavior.
- All four call sites (`GenerateSBOM`, `ScanCP`, `ScanCVE`, `ScanRegistry`) updated to pass `ctx`, which was already in scope at each. In `ScanCP`, `optionsFromWorkload` now receives `subWorkload`, which copies `Args`/`CredentialsList` verbatim from `workload` (scan.go:179-192), so the warning's `imageSlug` reflects the actual image being scanned.

**Testing**

Added `TestOptionsFromWorkload` in `core/services/scan_test.go`, covering:

- `bool true` / `bool false` — applied correctly
- JSON `string` — ignored, defaults to false
- `float64` (JSON number) — ignored, defaults to false
- `null` — ignored, defaults to false
- missing keys — default false
- unrelated extra keys — ignored

The string, number, and null cases all **panic on the previous code**, so this test is a true regression test rather than a snapshot of existing behavior.

Verified locally:

```
go build ./...
go vet ./core/services/
go test ./core/services/ -run TestOptionsFromWorkload -count=1
```

**Checklist**

- [x] Bug fix (non-breaking) — addresses #411
- [x] Unit tests added that fail on the previous behavior
- [x] `go build ./...` passes
- [x] `go vet ./core/services/` clean
- [x] Full `./core/services/` test suite passes
- [x] No new imports introduced

**Fixes** #411

### DIFF
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -113,7 +113,7 @@ func (s *ScanService) GenerateSBOM(ctx context.Context) error {
 	// if SBOM is not available, create it
 	if sbom.Content == nil {
 		// create SBOM
-		sbom, err = s.sbomCreator.CreateSBOM(ctx, workload.ImageSlug, workload.ImageHash, workload.ImageTagNormalized, optionsFromWorkload(workload))
+		sbom, err = s.sbomCreator.CreateSBOM(ctx, workload.ImageSlug, workload.ImageHash, workload.ImageTagNormalized, optionsFromWorkload(ctx, workload))
 		s.checkCreateSBOM(err, workload.ImageHash)
 		if err != nil {
 			_ = s.platform.ReportScanFailure(ctx, scanfailure.ScanFailureSBOMGeneration,
@@ -220,7 +220,7 @@ func (s *ScanService) ScanCP(mainCtx context.Context) error {
 			if sbom.Content == nil {
 				if s.sbomGeneration {
 					// create SBOM
-					sbom, err = s.sbomCreator.CreateSBOM(ctx, subWorkload.ImageSlug, subWorkload.ImageHash, subWorkload.ImageTagNormalized, optionsFromWorkload(workload))
+					sbom, err = s.sbomCreator.CreateSBOM(ctx, subWorkload.ImageSlug, subWorkload.ImageHash, subWorkload.ImageTagNormalized, optionsFromWorkload(ctx, subWorkload))
 					s.checkCreateSBOM(err, scan.ImageID)
 					if err != nil {
 						logger.L().Ctx(ctx).Error("creating SBOM, skipping scan", helpers.Error(err),
@@ -433,7 +433,7 @@ func (s *ScanService) ScanCVE(ctx context.Context) error {
 		if sbom.Content == nil {
 			if s.sbomGeneration {
 				// create SBOM
-				sbom, err = s.sbomCreator.CreateSBOM(ctx, workload.ImageSlug, workload.ImageHash, workload.ImageTagNormalized, optionsFromWorkload(workload))
+				sbom, err = s.sbomCreator.CreateSBOM(ctx, workload.ImageSlug, workload.ImageHash, workload.ImageTagNormalized, optionsFromWorkload(ctx, workload))
 				s.checkCreateSBOM(err, workload.ImageHash)
 				if err != nil {
 					reason := classifySBOMError(err)
@@ -559,7 +559,7 @@ func (s *ScanService) ScanRegistry(ctx context.Context) error {
 	}
 
 	// create SBOM
-	sbom, err := s.sbomCreator.CreateSBOM(ctx, workload.ImageSlug, workload.ImageHash, workload.ImageTagNormalized, optionsFromWorkload(workload))
+	sbom, err := s.sbomCreator.CreateSBOM(ctx, workload.ImageSlug, workload.ImageHash, workload.ImageTagNormalized, optionsFromWorkload(ctx, workload))
 	s.checkCreateSBOM(err, workload.ImageTagNormalized)
 	if err != nil {
 		repErr := s.platform.ReportError(ctx, err)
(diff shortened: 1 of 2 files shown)


core/services/scan.go
```diff
@@ -113,7 +113,7 @@ func (s *ScanService) GenerateSBOM(ctx context.Context) error {
 	// if SBOM is not available, create it
 	if sbom.Content == nil {
 		// create SBOM
-		sbom, err = s.sbomCreator.CreateSBOM(ctx, workload.ImageSlug, workload.ImageHash, workload.ImageTagNormalized, optionsFromWorkload(workload))
+			sbom, err = s.sbomCreator.CreateSBOM(ctx, workload.ImageSlug, workload.ImageHash, workload.ImageTagNormalized, optionsFromWorkload(ctx, workload))
```
matthyx: **Blocker (gofmt):** this line is indented one tab too deep — three tabs, but the enclosing `if sbom.Content == nil {` block is at two (compare `s.checkCreateSBOM(...)` on the next line, which is unchanged and correctly at two).

This is what puts `core/services/scan.go` in `gofmt -l` output on this branch; the file is clean on `main`. Run `gofmt -w core/services/scan.go`.

core/services/scan.go
```diff
@@ -433,7 +433,7 @@ func (s *ScanService) ScanCVE(ctx context.Context) error {
 		if sbom.Content == nil {
 			if s.sbomGeneration {
 				// create SBOM
-				sbom, err = s.sbomCreator.CreateSBOM(ctx, workload.ImageSlug, workload.ImageHash, workload.ImageTagNormalized, optionsFromWorkload(workload))
+		sbom, err = s.sbomCreator.CreateSBOM(ctx, workload.ImageSlug, workload.ImageHash, workload.ImageTagNormalized, optionsFromWorkload(ctx, workload))
```
matthyx: **Blocker (gofmt):** the mirror of the line 116 issue — this one lost two tabs. It sits inside `if sbom.Content == nil { if s.sbomGeneration {`, so it should be at four tabs like the `s.checkCreateSBOM(...)` line directly below it, not two.

Both of these disappear with `gofmt -w core/services/scan.go`, which will also reduce this hunk to just the added `ctx` argument.

matthyx (changes requested):
Reviewed at `3436e5d`.

The core fix is correct, and the test is a genuine regression test rather than a snapshot — I verified locally that `TestOptionsFromWorkload` panics on the pre-fix `optionsFromWorkload` (`panic: interface conversion: interface {} is string, not bool`) and passes after. `go build ./...`, `go vet ./core/services/`, and the full `./core/services/` suite are green on this branch.

One blocker before merge.

## Blocker — this PR breaks `gofmt` on `core/services/scan.go`

`gofmt -l ./core/` reports `core/services/scan.go` on this branch but **not** on `main`, so the regression is introduced here. Two of the four touched call sites were re-indented by hand: line 116 gained a tab, line 436 lost two. Inline comments below.

`gofmt -w core/services/scan.go` fixes both, and it also shrinks the diff down to the four added `ctx` arguments — which is what a reviewer should actually be looking at. Worth flagging that CI will not catch this for you: `pr-created.yaml` runs build and tests but has no format gate, so all checks are green despite the file being unformatted.

## Non-blocking notes

**1. Undocumented behavior change in `ScanCP` (line 223).** The argument changed from `workload` to `subWorkload`, which the description doesn't mention — it only says the call sites were updated to pass `ctx`. I checked and it is safe: `subWorkload` copies `CredentialsList` and `Args` verbatim from `workload` (`core/services/scan.go:179-192`), so the returned `RegistryOptions` are identical. The only field that differs is `ImageSlug`, which is now the per-image slug and therefore makes the new warning log *more* accurate. Please call it out in the description so it isn't mistaken for a stray edit.

**2. The `Args` panic class is not fully closed.** Four sibling unguarded assertions on the same untrusted map remain: `controllers/http.go:105-106` and `core/services/scan.go:154-155` and `880-881`, all `Args[ArgsName].(string)` / `Args[ArgsNamespace].(string)`. These are less severe than what you fixed, but the reason is worth knowing: the two inside `ScanCP` (154-155) are shielded only because `controllers/http.go:105-106` asserts the same two keys first, inside the Gin handler where `gin.Recovery()` does apply — so a payload with a non-string `name` 500s instead of crashing the pod. That protection is incidental, not designed. Any future non-HTTP entrypoint into `ScanCP`, or a reorder of the handler, puts an unrecovered `.(string)` back inside a worker goroutine with exactly the blast radius you describe. Not asking you to widen this PR — a follow-up issue would be the right home.

**3. The warning logs an unbounded client-controlled value.** `fmt.Sprintf("%v", useHTTP)` formats an arbitrary JSON value: `Args` is `map[string]interface{}`, so a client can put a deeply nested object or a multi-megabyte string under that key and it lands verbatim in the log, once per scan request. `logger.L().Ctx(ctx)` also forwards upstream to the reporting backend, so a single misconfigured client turns into sustained alert noise. Consider truncating the value, or dropping the `value` field and keeping only `%T` — the type is the actionable part for an operator.

**4. Minor.** The two branches are identical apart from the key and target field; a small `boolArg(ctx, workload, key) (bool, bool)` helper would remove the duplication. Take it or leave it.

AdeshDeshmukh:
Thanks @matthyx . I've pushed the fixes on the updated commit (c0b61c9):

- gofmt is fixed — gofmt -l core/ is clean now (only the pre-existing providers.go remains, which is already unformatted on main).
- I removed the value field from both warnings and kept just the type, as you suggested.
- Added the subWorkload note to the PR description — you were right that it deserved to be called out rather than left as a silent change.

I left the boolArg helper out for now to keep the diff focused, but happy to add it if you'd prefer.

And per your suggestion, I'll open a follow-up issue for the remaining unguarded Args[...].(string) assertions in http.go and ScanCP/ValidateScanCVE — since as you noted, they're only incidentally protected by gin.Recovery().

If you could resolve the threads when you get a chance, that'd be great.

matthyx (approved):
Re-reviewed at `c0b61c9`. The blocker is resolved — approving.

**Blocker fixed.** `gofmt -l ./core/` no longer reports `core/services/scan.go`; the only remaining entry is `core/ports/providers.go`, which is unformatted on `main` too and is not this PR's concern. Lines 116 and 436 are back at the correct indentation, and the diff is now exactly the four added `ctx` arguments plus the guarded assertions — much easier to read.

Also picked up from the non-blocking notes:

- The `subWorkload` switch in `ScanCP` is now documented in the description, including why the `imageSlug` in the warning is the more accurate one.
- The unbounded `value` field is gone from both warning logs. `key` + `%T` + `imageSlug` is the right amount to log — an operator gets the actionable part without client-controlled content flowing into the reporting backend.

**Verified locally at this commit:**

- `gofmt -l ./core/` — clean for the touched files
- `go build ./...` — pass
- `go vet ./core/services/` — clean
- `go test ./core/services/ -count=1` — full package passes
- Regression check: reverting `optionsFromWorkload` to the blind assertions while keeping the new test still produces `panic: interface conversion: interface {} is string, not bool`, so the test genuinely covers the reported bug rather than the new behavior.

All CI checks are green.

The remaining sibling assertions on `Args` (`controllers/http.go:105-106`, `core/services/scan.go:154-155` and `880-881`) are out of scope here as discussed — worth a follow-up issue, not a change to this PR. LGTM.

PR #415, opened 2026-08-01
fix: prevent panic on non-string ScanCP args.name/namespace

### LINKED ISSUE #414: Bug: application profile scan path panics on missing or non-string args.name / args.namespace
# Description
Both the HTTP controller and the service-layer validation for the container/application profile scan endpoint (`ScanCP`) perform unchecked type assertions on `Args["name"]` and `Args["namespace"]`. When a request payload sends a non-string value for either field (or omits it in a way that leaves a non-string type), the type assertion panics instead of kubevuln returning a structured validation error.

# Environment
OS: `Any`
Version: `upstream/main @ a5e45e3`

# Steps To Reproduce
1. Open `controllers/http.go` lines 104-106 and `core/services/scan.go` lines 863-865.
2. Observe the direct assertions `newScan.Args[domain.ArgsName].(string)` and `workload.Args[domain.ArgsNamespace].(string)` in both locations (and again at `core/services/scan.go` lines 154-155 inside `ScanCP`).
3. Send `POST /v1/applicationProfileScan` (`ScanCP`) with a JSON body whose `args.name` or `args.namespace` is a non-string JSON type (e.g. a number or boolean).
4. The unchecked type assertion panics before a validation error can be returned; gin's recovery middleware converts this into a generic 500 response.

# Expected behavior
Malformed or incomplete `args` values should be rejected with a normal validation error (`domain.ErrMissingCpInfo`), without panicking.

# Actual Behavior
The `ScanCP` handler and its validation path can panic on missing or wrongly typed `args` keys.

# Additional context
This is reachable directly from client-supplied JSON and does not require any special internal state to trigger.

rootp1:
## Overview
Both the HTTP controller and the service-layer validation for the `ScanCP` (application profile scan) path perform unchecked type assertions on `Args["name"]` and `Args["namespace"]`. A request with a non-string value for either field causes a panic instead of a validation error.

This PR replaces the unchecked type assertions in `controllers/http.go`, `core/services/scan.go` (`ValidateScanCP`), and `core/services/scan.go` (`ScanCP`) with comma-ok assertions, so malformed values fall through to the existing `domain.ErrMissingCpInfo` validation error path instead of panicking.

## How to Test
`go test ./controllers/... ./core/services/...`

Added `TestScanService_ValidateScanCP`, covering missing args and non-string `name`/`namespace` values, which previously would panic.

## Related issues/PRs:
Resolved #414

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented scan requests with missing, invalid, or incomplete workload information from causing application errors.
  * Such requests are now handled through standard validation and return the appropriate missing-information response.

* **Tests**
  * Added coverage for invalid and valid scan inputs, including missing fields and non-text values.
  * Added regression coverage to ensure incomplete requests are handled safely.

### DIFF
--- a/controllers/http.go
+++ b/controllers/http.go
@@ -56,7 +56,7 @@ func (h HTTPController) GenerateSBOM(c *gin.Context) {
 			helpers.String("imageSlug", newScan.ImageSlug),
 			helpers.String("imageTagNormalized", newScan.ImageTagNormalized),
 			helpers.String("imageHash", newScan.ImageHash))
-		_, _ = problem.Of(http.StatusInternalServerError).Append(details).WriteTo(c.Writer)
+		_, _ = problem.Of(validationStatusCode(err)).Append(details).WriteTo(c.Writer)
 		return
 	}
 
@@ -113,7 +113,7 @@ func (h HTTPController) ScanCP(c *gin.Context) {
 			helpers.String("wlid", newScan.Wlid),
 			helpers.String("name", name),
 			helpers.String("namespace", namespace))
-		_, _ = problem.Of(http.StatusInternalServerError).Append(details).WriteTo(c.Writer)
+		_, _ = problem.Of(validationStatusCode(err)).Append(details).WriteTo(c.Writer)
 		return
 	}
 
@@ -160,7 +160,7 @@ func (h HTTPController) ScanCVE(c *gin.Context) {
 			helpers.String("imageSlug", newScan.ImageSlug),
 			helpers.String("imageTagNormalized", newScan.ImageTagNormalized),
 			helpers.String("imageHash", newScan.ImageHash))
-		_, _ = problem.Of(http.StatusInternalServerError).Append(details).WriteTo(c.Writer)
+		_, _ = problem.Of(validationStatusCode(err)).Append(details).WriteTo(c.Writer)
 		return
 	}
 
@@ -177,6 +177,16 @@ func (h HTTPController) ScanCVE(c *gin.Context) {
 				helpers.String("imageHash", newScan.ImageHash))
 		}
 	})
+}
+
+// validationStatusCode maps a validation error to the HTTP status code that best
+// describes it. ErrTooManyRequests reflects transient registry back-pressure, not
+// a malformed request, so it maps to 429 rather than 400.
+func validationStatusCode(err error) int {
+	if errors.Is(err, domain.ErrTooManyRequests) {
+		return http.StatusTooManyRequests
+	}
+	return http.StatusBadRequest
 }
 
 func websocketScanCommandToScanCommand(c wssc.WebsocketScanCommand) domain.ScanCommand {
@@ -230,7 +240,7 @@ func (h HTTPController) ScanRegistry(c *gin.Context) {
 			helpers.String("imageSlug", newScan.ImageSlug),
 			helpers.String("imageTagNormalized", newScan.ImageTagNormalized),
 			helpers.String("imageHash", newScan.ImageHash))
-		_, _ = problem.Of(http.StatusInternalServerError).Append(details).WriteTo(c.Writer)
+		_, _ = problem.Of(validationStatusCode(err)).Append(details).WriteTo(c.Writer)
 		return
 	}
 
--- a/controllers/http.go
+++ b/controllers/http.go
@@ -102,8 +102,8 @@ func (h HTTPController) ScanCP(c *gin.Context) {
 	}
 
 	newScan := websocketScanCommandToScanCommand(websocketScanCommand)
-	name := newScan.Args[domain.ArgsName].(string)
-	namespace := newScan.Args[domain.ArgsNamespace].(string)
+	name, _ := newScan.Args[domain.ArgsName].(string)
+	namespace, _ := newScan.Args[domain.ArgsNamespace].(string)
 
 	details := problem.Detailf("Wlid=%s, Name=%s, Namespace=%s", newScan.Wlid, name, namespace)
 
--- a/core/services/scan_test.go
+++ b/core/services/scan_test.go
@@ -602,12 +602,12 @@ func TestScanService_ValidateScanCP(t *testing.T) {
 	tests := []struct {
 		name     string
 		workload domain.ScanCommand
-		wantErr  bool
+		wantErr  error
 	}{
 		{
 			name:     "missing args",
 			workload: domain.ScanCommand{},
-			wantErr:  true,
+			wantErr:  domain.ErrMissingCpInfo,
 		},
 		{
 			name: "non-string name and namespace",
@@ -617,7 +617,45 @@ func TestScanService_ValidateScanCP(t *testing.T) {
 					domain.ArgsNamespace: true,
 				},
 			},
-			wantErr: true,
+			wantErr: domain.ErrMissingCpInfo,
+		},
+		{
+			name: "non-string name only",
+			workload: domain.ScanCommand{
+				Args: map[string]interface{}{
+					domain.ArgsName:      123,
+					domain.ArgsNamespace: "kube-system",
+				},
+			},
+			wantErr: domain.ErrMissingCpInfo,
+		},
+		{
+			name: "non-string namespace only",
+			workload: domain.ScanCommand{
+				Args: map[string]interface{}{
+					domain.ArgsName:      "daemonset-kube-proxy",
+					domain.ArgsNamespace: true,
+				},
+			},
+			wantErr: domain.ErrMissingCpInfo,
+		},
+		{
+			name: "missing name only",
+			workload: domain.ScanCommand{
+				Args: map[string]interface{}{
+					domain.ArgsNamespace: "kube-system",
+				},
+			},
+			wantErr: domain.ErrMissingCpInfo,
+		},
+		{
+			name: "missing namespace only",
+			workload: domain.ScanCommand{
+				Args: map[string]interface{}{
+					domain.ArgsName: "daemonset-kube-proxy",
+				},
+			},
+			wantErr: domain.ErrMissingCpInfo,
 		},
 		{
 			name: "with name and namespace",
@@ -627,16 +665,17 @@ func TestScanService_ValidateScanCP(t *testing.T) {
 					domain.ArgsNamespace: "kube-system",
 				},
 			},
-			wantErr: false,
+			wantErr: nil,
 		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			s := NewScanService(adapters.NewMockSBOMAdapter(false, false, false), repositories.NewMemoryStorage(false, false), adapters.NewMockCVEAdapter(), repositories.NewMemoryStorage(false, false), adapters.NewMockPlatform(false, nil), adapters.NewMockRelevancyAdapter(), false, false, true, false, false)
 			_, err := s.ValidateScanCP(context.TODO(), tt.workload)
-			if (err != nil) != tt.wantErr {
-				t.Errorf("ValidateScanCP() error = %v, wantErr %v", err, tt.wantErr)
-				return
+			if tt.wantErr != nil {
+				assert.ErrorIs(t, err, tt.wantErr)
+			} else {
+				assert.NoError(t, err)
 			}
 		})
 	}
--- a/core/services/scan_test.go
+++ b/core/services/scan_test.go
@@ -598,6 +598,50 @@ func TestScanService_ValidateGenerateSBOM(t *testing.T) {
 	}
 }
 
+func TestScanService_ValidateScanCP(t *testing.T) {
+	tests := []struct {
+		name     string
+		workload domain.ScanCommand
+		wantErr  bool
+	}{
+		{
+			name:     "missing args",
+			workload: domain.ScanCommand{},
+			wantErr:  true,
+		},
+		{
+			name: "non-string name and namespace",
+			workload: domain.ScanCommand{
+				Args: map[string]interface{}{
+					domain.ArgsName:      123,
+					domain.ArgsNamespace: true,
+				},
+			},
+			wantErr: true,
+		},
+		{
+			name: "with name and namespace",
+			workload: domain.ScanCommand{
+				Args: map[string]interface{}{
+					domain.ArgsName:      "daemonset-kube-proxy",
+					domain.ArgsNamespace: "kube-system",
+				},
+			},
+			wantErr: false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			s := NewScanService(adapters.NewMockSBOMAdapter(false, false, false), repositories.NewMemoryStorage(false, false), adapters.NewMockCVEAdapter(), repositories.NewMemoryStorage(false, false), adapters.NewMockPlatform(false, nil), adapters.NewMockRelevancyAdapter(), false, false, true, false, false)
+			_, err := s.ValidateScanCP(context.TODO(), tt.workload)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("ValidateScanCP() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+		})
+	}
+}
+
 func TestScanService_ValidateScanCVE(t *testing.T) {
 	tests := []struct {
 		name     string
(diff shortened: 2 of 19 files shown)


rootp1:
@matthyx can you please have a look whenever you have a chance to

matthyx (commented):
Reviewed at `cc7daf0`.

**The core fix is right.** `Args` is a `map[string]interface{}` populated straight from the request JSON (`controllers/http.go:194`, `Args: c.Args`), so `args.name: 123` — or no `args` at all — reaches `newScan.Args[domain.ArgsName].(string)` at line 105 before any validation runs. I put your new test in front of `main`'s `core/services/scan.go` and it reproduces exactly what #414 describes:

```
--- FAIL: TestScanService_ValidateScanCP/missing_args
panic: interface conversion: interface {} is nil, not string
```

With the comma-ok assertions the malformed value falls through to `ErrMissingCpInfo`, which is the behaviour the issue asks for, and `go test ./controllers/... ./core/services/...` is clean here.

**One blocker: this PR re-introduces the VEX backfill loop we dropped in #404 on purpose.** `repositories/apiserver.go` and `repositories/apiserver_test.go` have nothing to do with #414, and the code they carry is the unreachable loop from the pre-rebase version of that PR. Details inline. Everything else is non-blocking.

### Follow-up, not for this PR

The same bug class is still live one layer down, and there it is worse. `adapters/v1/backend.go:322-340` does an unchecked `val.(string)` on seven `workload.Args[...]` values, fed by the same client-supplied `args` map (`registryScanCommandToScanCommand` also does `Args: c.Args`). Confirmed by calling `SubmitCVE` with `Args: {"registryName": 123}`:

```
panic: interface conversion: interface {} is int, not string
	.../adapters/v1/backend.go:322
```

`SubmitCVE` runs inside a `h.workerPool.Submit(...)` task, and `gammazero/workerpool` has no `recover()` anywhere in its worker loop — so unlike the panic this PR fixes, which `gin.Recovery()` (`cmd/http/main.go:121-122`) downgrades to a 500, that one takes the process down. `backend.go:174` in the same file already does it the safe way, so the fix is mechanical. Worth its own issue rather than growing this PR.

repositories/apiserver.go
```diff
@@ -786,6 +786,16 @@ func markRelevantVulnerabilitiesAsAffectedInVex(vexDoc *v1beta1.VEX, cvep *domai
...
+
+	// Backfill statements that were already marked "affected" by a version of kubevuln
+	// predating action_statement support, or that were left stale because their CVE fell
+	// out of the current relevancy set before the backfill could run on a prior update.
+	for i, s := range vexDoc.Statements {
+		if s.Status == v1beta1.Status(vex.StatusAffected) && s.ActionStatement == "" {
+			vexDoc.Statements[i].ImpactStatement = ""
+			vexDoc.Statements[i].ActionStatement = defaultActionStatement
+		}
+	}
```
matthyx: **Blocker** — this is the backfill loop from the pre-rebase version of #404, which we dropped there on purpose, and none of it belongs to #414.

Recap from that thread: I asked for this loop in my first review, then #401 (`a5e45e3`) landed the reset pass in `updateVEX` that sends every statement back to `not_affected` with `ActionStatement = ""` before `markRelevantVulnerabilitiesAsAffectedInVex` reapplies the manifest. That made the loop unreachable, so `53b2adf` dropped it — merged #404 (`33b467b`) has no backfill.

It is unreachable on this branch too, on both callers:

- `createVEX` (`apiserver.go:856`) — statements are constructed `not_affected`, and the mark loop then sets `ActionStatement = buildActionStatement(v)`, which never returns `""` (it falls back to `defaultActionStatement`).
- `updateVEX` (`apiserver.go:952-957`) — the reset loop clears `Status` **and** `ActionStatement` immediately before the call.

Same probe as last time: replace the loop body with `panic("BACKFILL LOOP REACHED")` and run the suite.

```
$ go test ./repositories/...
ok  	github.com/kubescape/kubevuln/repositories	2.761s
```

Zero hits — including from the test this PR adds below.

How it got here: the branch was cut before that rework (`35619a3` and `2f04c5a` predate it) and the `cc7daf0` merge of `upstream/main` auto-merged `apiserver.go` — `main` never had the loop, so there was nothing for the merge to remove and your side was kept verbatim. `git checkout upstream/main -- repositories/apiserver.go repositories/apiserver_test.go` should get this PR back to the three files #414 is about.

repositories/apiserver_test.go
```diff
@@ -448,6 +448,66 @@ func TestAPIServerStore_storeVEX_affectedStatementsHaveActionStatement(t *testin
...
+	found = false
+	for _, stmt := range vexContainer.Spec.Statements {
+		if stmt.Vulnerability.Name == "CVE-2005-2541" {
+			found = true
+			assert.Equal(t, v1beta1.Status(vex.StatusNotAffected), stmt.Status, "statement no longer relevant should be reset to not_affected")
+			assert.Empty(t, stmt.ActionStatement, "not_affected statement must not carry an action_statement")
+		}
+	}
+	require.True(t, found, "expected CVE-2005-2541 statement to still be present after the update")
+}
```
matthyx: **Blocker** — goes away with the same revert as `apiserver.go`, but flagging separately because this test is misleading on its own terms.

The name says `backfillsActionStatementOnStaleAffectedStatements`; the assertions say the opposite — `not_affected` and an empty `ActionStatement`, i.e. that the backfill does *not* happen. Your own doc comment on lines 453-455 says as much. It passes, but it passes against the reset behaviour from #401, not against the loop it is named for: delete that loop and this test is still green.

The invariant it does check is already covered, and more broadly. `TestAPIServerStore_storeVEX_updateRestoresNotAffected` (line 360) walks *every* statement in the document rather than just `CVE-2005-2541`, and already asserts `assert.Empty(t, stmt.ActionStatement, ...)` at line 408. So there is no coverage lost by dropping these 60 lines.

controllers/http.go
```diff
@@ -102,8 +102,8 @@ func (h HTTPController) ScanCP(c *gin.Context) {
 	}
 
 	newScan := websocketScanCommandToScanCommand(websocketScanCommand)
-	name := newScan.Args[domain.ArgsName].(string)
-	namespace := newScan.Args[domain.ArgsNamespace].(string)
+	name, _ := newScan.Args[domain.ArgsName].(string)
+	namespace, _ := newScan.Args[domain.ArgsNamespace].(string)
```
matthyx, line 106: Non-blocking, but the handler is the half of this fix with no test behind it.

#414's repro hits these two lines before `ValidateScanCP` is ever reached, yet the only test added is at the service layer — and there is no `TestHTTPController_ScanCP` at all today (`http_test.go` has `Alive`, `GenerateSBOM`, `Ready`, `ScanCVE`, `ScanRegistry`, `registryScanCommandToScanCommand`, `ContextCancellationIsDetached`). Drop either `, _` here and the suite stays green.

A small handler test closes it: `POST /v1/scanCP` with `"args": {"name": 123, "namespace": true}`, asserting the validation error is returned and no job reaches the pool. One thing to watch — register the route on a bare `gin.New()` rather than `gin.Default()`, otherwise `gin.Recovery()` converts a regression into the same 500 the validation path returns and the test can't tell the two apart. The `contextSpyScanService` harness at line 289 already gives you the "did the worker run" signal.

core/services/scan_test.go
```diff
@@ -598,6 +598,50 @@ func TestScanService_ValidateGenerateSBOM(t *testing.T) {
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			s := NewScanService(adapters.NewMockSBOMAdapter(false, false, false), repositories.NewMemoryStorage(false, false), adapters.NewMockCVEAdapter(), repositories.NewMemoryStorage(false, false), adapters.NewMockPlatform(false, nil), adapters.NewMockRelevancyAdapter(), false, false, true, false, false)
+			_, err := s.ValidateScanCP(context.TODO(), tt.workload)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("ValidateScanCP() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
```
matthyx: Nit: `wantErr bool` is satisfied by *any* error, but the contract in #414 and in your description is specifically that malformed values fall through to `domain.ErrMissingCpInfo`. A `wantErr error` field with `assert.ErrorIs(t, err, tt.wantErr)` pins the behaviour you actually claim, and would still trip if someone later swaps the sentinel or short-circuits earlier in the function.

rootp1:
@matthyx you're right, that was leftover from a merge conflict resolution and doesn't belong in this PR. I removed the backfill loop from apiserver.go and the accompanying test, keeping this PR scoped to the ScanCP args type assertion fix.

rootp1:
@matthyx pushed the fix, could you take another look when you have a chance?

matthyx (commented):
Re-checked `a8e2c82`. The blocker is resolved.

`git diff main...HEAD` is now exactly the three files #414 is about:

```
 controllers/http.go        |  4 ++--
 core/services/scan.go      |  8 ++++----
 core/services/scan_test.go | 44 ++++++++++++++++++++++++++++++++++++++++++++
```

`git diff main...HEAD -- repositories/` is empty, so `apiserver.go` and `apiserver_test.go` are back to byte-identical with `main` and the unreachable backfill loop is gone. `go build ./...`, `go vet` and `go test ./controllers/... ./core/services/... ./repositories/...` are all clean here, and CI is green on the new head.

I mutation-checked the guard while I was in there — reverting either comma-ok in `ValidateScanCP` brings the panic straight back through your test, so it is a real regression guard:

```
--- FAIL: TestScanService_ValidateScanCP/missing_args
panic: interface conversion: interface {} is nil, not string
```

Nothing left that should hold this up. Two things stay open from the last round, both non-blocking — take either, neither, or a follow-up:

1. `controllers/http.go:105-106` is still the untested half. Reverting both lines to the unchecked form leaves `go test ./controllers/...` green, and that is the exact line #414's repro hits first.
2. CodeRabbit's note on the test table is fair, and it turns out to be the same gap as my sentinel nit: the invalid case makes `name` and `namespace` bad at the same time, so nothing pins the `namespace` half of `if name == "" || namespace == ""`. I checked — deleting `|| namespace == ""` keeps `TestScanService_ValidateScanCP` green. One invalid-`name` case, one invalid-`namespace` case, plus `assert.ErrorIs(t, err, domain.ErrMissingCpInfo)` instead of `wantErr bool`, closes both at once.

controllers/http.go
rootp1: Added a small handler test with a bare gin.New() router asserting no panic on missing args, per your suggestion.

core/services/scan_test.go
rootp1: Switched to wantErr error with assert.ErrorIs and split the case into independent name-only and namespace-only variants, so each half of the || pins separately.

core/services/scan_test.go
```diff
@@ -598,6 +598,50 @@ func TestScanService_ValidateGenerateSBOM(t *testing.T) {
...
+		{
+			name: "non-string name and namespace",
+			workload: domain.ScanCommand{
+				Args: map[string]interface{}{
+					domain.ArgsName:      123,
+					domain.ArgsNamespace: true,
+				},
+			},
+			wantErr: true,
+		},
```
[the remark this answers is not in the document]
rootp1: Fixed. Added separate cases for invalid name only and invalid namespace only, alongside the existing combined case.

rootp1:
@matthyx addressed both non-blocking items from the last round: added a handler-level test with a bare router confirming no panic on missing args, and split the ScanCP validation test into independent name-only and namespace-only cases using assert.ErrorIs. This has been through a couple of review rounds now and checks are green, so let me know if it looks ready to merge whenever you have a moment.

matthyx (commented):
Re-checked `e0756de`. Both items from the last round are addressed, and the mutations that were silent before now fail the way they should.

`TestScanService_ValidateScanCP` pins each half of the guard independently and asserts the sentinel. Deleting `|| namespace == ""` from `ValidateScanCP`:

```
--- FAIL: TestScanService_ValidateScanCP/non-string_namespace_only
--- FAIL: TestScanService_ValidateScanCP/missing_namespace_only
```

That mutation was green last round. Reverting either comma-ok in the same function still panics through the test.

`TestHTTPController_ScanCP_MissingArgsDoesNotPanic` closes the handler gap — with both `, _` in `http.go:105-106` reverted:

```
--- FAIL: TestHTTPController_ScanCP_MissingArgsDoesNotPanic
    Error:  func (assert.PanicTestFunc)(...) should not panic
    panic: interface conversion: interface {} is nil, not string
```

Also green last round. Good call on the bare `gin.New()` — that's what makes the assertion mean anything.

`go build ./...`, `go vet ./...`, `gofmt` over the touched files and `go test ./controllers/... ./core/services/... ./repositories/...` are all clean here, and all 11 checks are green on this head.

Two observations, neither needs a change:

- The comment at `http_test.go:178-180` says `gin.Recovery()` would mask the panic "into the same 400 the validation path returns" — on this branch that path is a 500 (`problem.Of(http.StatusInternalServerError)` at `http.go:116`). It becomes a 400 with #417, so the comment lands correctly once that merges; the reasoning holds either way.
- The third site, `scan.go:154-155` inside `ScanCP`, isn't pinned by anything — reverting it leaves the suite green. That's expected rather than a gap, since `ScanCP` only ever runs after `ValidateScanCP` has proved both values are non-empty strings, so it's defence in depth. Noting it only so nobody reads the suite as covering all three sites.

Ready to go in from my side.

matthyx:
rebase and I'll merge

rootp1:
@matthyx I fixed the merge conflict

matthyx (commented):
Re-checked `2659780`. Everything still passes, but the merge changed what this PR *is*: #417 landed on `main` as `31d1b03` in the meantime and took two of the three fixes with it, so the diff has shrunk to one production line plus tests.

Against `main` today:

| file | what's left |
|---|---|
| `core/services/scan.go` | +2/-2 — the `ScanCP` worker site at `154-155`, the only production change remaining |
| `core/services/scan_test.go` | +83 — new `TestScanService_ValidateScanCP` |
| `controllers/http_test.go` | +24/-28 — moves and re-implements a test `main` already has |

`git log -S 'name, _ := newScan.Args[domain.ArgsName].(string)' main` → `31d1b03`, same for the `ValidateScanCP` site, so `controllers/http.go` has dropped out of the diff entirely and `ValidateScanCP` is already fixed upstream.

**The test table is the real contribution now, and it earns its place.** `main` has no direct coverage of `ValidateScanCP`. Breaking the version that is already on `main` fails it two different ways:

```
# name comma-ok reverted
--- FAIL: TestScanService_ValidateScanCP/missing_args
panic: interface conversion: interface {} is nil, not string

# `|| namespace == ""` deleted
--- FAIL: TestScanService_ValidateScanCP/non-string_namespace_only
--- FAIL: TestScanService_ValidateScanCP/missing_namespace_only
```

**`controllers/http_test.go` is now churn and I would drop it.** #417 already added `TestHTTPController_ScanCP_MissingArgsDoesNotPanic`; this moves it up the file and swaps `contextSpyScanService` for `NewMockScanService(true)`. Comparing the test function names either side of the change:

```
$ git show main:controllers/http_test.go | grep '^func Test' | sort > a
$ grep '^func Test' controllers/http_test.go | sort > b
$ diff a b        # (no output)
```

Same set, and reverting the handler assertions on `main` fails either copy identically, so nothing is gained or lost by dropping the file from this PR. The one thing worth keeping is the comment fix: `main`'s copy still says `gin.Recovery()` would mask the panic "into the same 500", which stopped being true when `validationStatusCode` started returning `http.StatusBadRequest`. That's a one-line correction, not a move.

**On `scan.go:154-155`** — worth keeping, worth describing accurately. Reverting it leaves `go test ./controllers/... ./core/services/...` green, because `ScanCP` only runs after `ValidateScanCP` has already proved both values are non-empty strings. It is defence in depth rather than a reachable panic, and the PR description still presents all three sites as live.

Housekeeping: the branch is one commit behind `main` (`178f253`, #413, which also touches `core/services/scan_test.go`) — no conflict, GitHub still reports mergeable. And the description is stale now that `controllers/http.go` and `ValidateScanCP` are upstream; worth a refresh so the history reads right.

`go build ./...`, `go vet`, `gofmt` and `go test ./controllers/... ./core/services/... ./repositories/...` are clean on this head. Nothing blocking from me — trimming `http_test.go` and updating the description would leave a much clearer PR.

matthyx (approved):

PR #417, opened 2026-08-01
fix: return 400 Bad Request for validation failures in HTTP controller

### LINKED ISSUE #416: Bug: request validation failures are surfaced as 500 Internal Server Error instead of 4xx client errors
# Description
All four HTTP handlers map business-validation failures to `500 Internal Server Error`, even when the error is caused by a malformed or incomplete client request such as missing image identifiers or missing application-profile arguments.

# Environment
OS: `Any`

# Steps To Reproduce
1. Open `controllers/http.go`.
2. Inspect the validation failure branches in `GenerateSBOM` (lines 53-60), `ScanCP` (lines 110-117), `ScanCVE` (lines 157-164), and `ScanRegistry` (lines 227-234).
3. Confirm each branch writes `problem.Of(http.StatusInternalServerError)`.
4. Compare that with the validation helpers in `core/services/scan.go`, which return input-related errors such as `ErrMissingImageInfo` and `ErrMissingCpInfo`.

# Expected behavior
Input validation failures should return a 4xx status, typically `400 Bad Request` or `422 Unprocessable Entity`.

# Actual Behavior
The API reports client-side validation problems as server-side faults.

# Additional context
The current controller tests in `controllers/http_test.go` explicitly assert these 500 responses, so the mismatch is locked in by test expectations.

rootp1:
## Overview
Validation failures in `GenerateSBOM`, `ScanCP`, `ScanCVE`, and `ScanRegistry` were mapped to `500 Internal Server Error`, even though they are caused by malformed or incomplete client requests (missing image identifiers, missing application-profile arguments). This PR changes those four branches to return `400 Bad Request` instead, matching the client-error nature of the failure.

## Related issues/PRs:
* Resolved #416

### DIFF
--- a/controllers/http.go
+++ b/controllers/http.go
@@ -56,7 +56,7 @@ func (h HTTPController) GenerateSBOM(c *gin.Context) {
 			helpers.String("imageSlug", newScan.ImageSlug),
 			helpers.String("imageTagNormalized", newScan.ImageTagNormalized),
 			helpers.String("imageHash", newScan.ImageHash))
-		_, _ = problem.Of(http.StatusBadRequest).Append(details).WriteTo(c.Writer)
+		_, _ = problem.Of(validationStatusCode(err)).Append(details).WriteTo(c.Writer)
 		return
 	}
 
@@ -102,8 +102,8 @@ func (h HTTPController) ScanCP(c *gin.Context) {
 	}
 
 	newScan := websocketScanCommandToScanCommand(websocketScanCommand)
-	name := newScan.Args[domain.ArgsName].(string)
-	namespace := newScan.Args[domain.ArgsNamespace].(string)
+	name, _ := newScan.Args[domain.ArgsName].(string)
+	namespace, _ := newScan.Args[domain.ArgsNamespace].(string)
 
 	details := problem.Detailf("Wlid=%s, Name=%s, Namespace=%s", newScan.Wlid, name, namespace)
 
@@ -113,7 +113,7 @@ func (h HTTPController) ScanCP(c *gin.Context) {
 			helpers.String("wlid", newScan.Wlid),
 			helpers.String("name", name),
 			helpers.String("namespace", namespace))
-		_, _ = problem.Of(http.StatusBadRequest).Append(details).WriteTo(c.Writer)
+		_, _ = problem.Of(validationStatusCode(err)).Append(details).WriteTo(c.Writer)
 		return
 	}
 
@@ -160,7 +160,7 @@ func (h HTTPController) ScanCVE(c *gin.Context) {
 			helpers.String("imageSlug", newScan.ImageSlug),
 			helpers.String("imageTagNormalized", newScan.ImageTagNormalized),
 			helpers.String("imageHash", newScan.ImageHash))
-		_, _ = problem.Of(http.StatusBadRequest).Append(details).WriteTo(c.Writer)
+		_, _ = problem.Of(validationStatusCode(err)).Append(details).WriteTo(c.Writer)
 		return
 	}
 
@@ -177,6 +177,16 @@ func (h HTTPController) ScanCVE(c *gin.Context) {
 				helpers.String("imageHash", newScan.ImageHash))
 		}
 	})
+}
+
+// validationStatusCode maps a validation error to the HTTP status code that best
+// describes it. ErrTooManyRequests reflects transient registry back-pressure, not
+// a malformed request, so it maps to 429 rather than 400.
+func validationStatusCode(err error) int {
+	if errors.Is(err, domain.ErrTooManyRequests) {
+		return http.StatusTooManyRequests
+	}
+	return http.StatusBadRequest
 }
 
 func websocketScanCommandToScanCommand(c wssc.WebsocketScanCommand) domain.ScanCommand {
@@ -230,7 +240,7 @@ func (h HTTPController) ScanRegistry(c *gin.Context) {
 			helpers.String("imageSlug", newScan.ImageSlug),
 			helpers.String("imageTagNormalized", newScan.ImageTagNormalized),
 			helpers.String("imageHash", newScan.ImageHash))
-		_, _ = problem.Of(http.StatusBadRequest).Append(details).WriteTo(c.Writer)
+		_, _ = problem.Of(validationStatusCode(err)).Append(details).WriteTo(c.Writer)
 		return
 	}
 
--- a/controllers/http.go
+++ b/controllers/http.go
@@ -56,7 +56,7 @@ func (h HTTPController) GenerateSBOM(c *gin.Context) {
 			helpers.String("imageSlug", newScan.ImageSlug),
 			helpers.String("imageTagNormalized", newScan.ImageTagNormalized),
 			helpers.String("imageHash", newScan.ImageHash))
-		_, _ = problem.Of(http.StatusInternalServerError).Append(details).WriteTo(c.Writer)
+		_, _ = problem.Of(http.StatusBadRequest).Append(details).WriteTo(c.Writer)
 		return
 	}
 
@@ -113,7 +113,7 @@ func (h HTTPController) ScanCP(c *gin.Context) {
 			helpers.String("wlid", newScan.Wlid),
 			helpers.String("name", name),
 			helpers.String("namespace", namespace))
-		_, _ = problem.Of(http.StatusInternalServerError).Append(details).WriteTo(c.Writer)
+		_, _ = problem.Of(http.StatusBadRequest).Append(details).WriteTo(c.Writer)
 		return
 	}
 
@@ -160,7 +160,7 @@ func (h HTTPController) ScanCVE(c *gin.Context) {
 			helpers.String("imageSlug", newScan.ImageSlug),
 			helpers.String("imageTagNormalized", newScan.ImageTagNormalized),
 			helpers.String("imageHash", newScan.ImageHash))
-		_, _ = problem.Of(http.StatusInternalServerError).Append(details).WriteTo(c.Writer)
+		_, _ = problem.Of(http.StatusBadRequest).Append(details).WriteTo(c.Writer)
 		return
 	}
 
@@ -230,7 +230,7 @@ func (h HTTPController) ScanRegistry(c *gin.Context) {
 			helpers.String("imageSlug", newScan.ImageSlug),
 			helpers.String("imageTagNormalized", newScan.ImageTagNormalized),
 			helpers.String("imageHash", newScan.ImageHash))
-		_, _ = problem.Of(http.StatusInternalServerError).Append(details).WriteTo(c.Writer)
+		_, _ = problem.Of(http.StatusBadRequest).Append(details).WriteTo(c.Writer)
 		return
 	}
 
--- a/controllers/http_test.go
+++ b/controllers/http_test.go
@@ -416,3 +416,71 @@ func TestHTTPController_ContextCancellationIsDetached(t *testing.T) {
 	}
 }
 
+func TestValidationStatusCode(t *testing.T) {
+	assert.Equal(t, http.StatusTooManyRequests, validationStatusCode(domain.ErrTooManyRequests))
+	assert.Equal(t, http.StatusBadRequest, validationStatusCode(domain.ErrMissingCpInfo))
+	assert.Equal(t, http.StatusBadRequest, validationStatusCode(domain.ErrMockError))
+}
+
+// validateErrScanService returns a fixed error from every Validate* method so
+// handler tests can exercise a specific validation error without depending on
+// MockScanService's generic sad path.
+type validateErrScanService struct {
+	*services.MockScanService
+	err error
+}
+
+func (s validateErrScanService) ValidateGenerateSBOM(ctx context.Context, _ domain.ScanCommand) (context.Context, error) {
+	return ctx, s.err
+}
+
+func (s validateErrScanService) ValidateScanCVE(ctx context.Context, _ domain.ScanCommand) (context.Context, error) {
+	return ctx, s.err
+}
+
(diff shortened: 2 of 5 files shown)


controllers/http.go
```diff
@@ -56,7 +56,7 @@ func (h HTTPController) GenerateSBOM(c *gin.Context) {
 			helpers.String("imageSlug", newScan.ImageSlug),
 			helpers.String("imageTagNormalized", newScan.ImageTagNormalized),
 			helpers.String("imageHash", newScan.ImageHash))
-		_, _ = problem.Of(http.StatusInternalServerError).Append(details).WriteTo(c.Writer)
+		_, _ = problem.Of(http.StatusBadRequest).Append(details).WriteTo(c.Writer)
```
matthyx: **Blocker.** `ValidateGenerateSBOM` does not only return input-validation errors — it also returns `domain.ErrTooManyRequests` when a previous pull of this image hit a registry `429` (`core/services/scan.go:852-855`). That is transient server-side back-pressure on a well-formed request, and it now surfaces as `400 Bad Request`, telling the caller the request is permanently broken.

Please branch on the error instead of blanket-mapping to `400`:

```go
status := http.StatusBadRequest
if errors.Is(err, domain.ErrTooManyRequests) {
    status = http.StatusTooManyRequests
}
_, _ = problem.Of(status).Append(details).WriteTo(c.Writer)
```

controllers/http.go
```diff
@@ -113,7 +113,7 @@ func (h HTTPController) ScanCP(c *gin.Context) {
 			helpers.String("wlid", newScan.Wlid),
 			helpers.String("name", name),
 			helpers.String("namespace", namespace))
-		_, _ = problem.Of(http.StatusInternalServerError).Append(details).WriteTo(c.Writer)
+		_, _ = problem.Of(http.StatusBadRequest).Append(details).WriteTo(c.Writer)
```
matthyx: **Blocker.** This branch is unreachable for the case the PR description names ("missing application-profile arguments"). Lines 105-106 do unchecked type assertions on `newScan.Args`:

```go
name := newScan.Args[domain.ArgsName].(string)
namespace := newScan.Args[domain.ArgsNamespace].(string)
```

A request without `args` panics there, and gin's `Recovery` returns a bodyless `500`. Verified on this branch:

```
interface conversion: interface {} is nil, not string
  controllers/http.go:105 HTTPController.ScanCP
=> STATUS=500 BODY=""
```

Use the comma-ok form (here and at `core/services/scan.go:864-865`) so the request actually reaches `ValidateScanCP` and gets `ErrMissingCpInfo` → `400`.

controllers/http.go
```diff
@@ -160,7 +160,7 @@ func (h HTTPController) ScanCVE(c *gin.Context) {
 			helpers.String("imageSlug", newScan.ImageSlug),
 			helpers.String("imageTagNormalized", newScan.ImageTagNormalized),
 			helpers.String("imageHash", newScan.ImageHash))
-		_, _ = problem.Of(http.StatusInternalServerError).Append(details).WriteTo(c.Writer)
+		_, _ = problem.Of(http.StatusBadRequest).Append(details).WriteTo(c.Writer)
```
matthyx: **Blocker (same as `GenerateSBOM`).** `ValidateScanCVE` returns `domain.ErrTooManyRequests` at `core/services/scan.go:899-902`, so registry rate-limit back-pressure is now reported as `400 Bad Request` on this endpoint too. Needs the same `errors.Is` branch to `429`.

controllers/http.go
```diff
@@ -230,7 +230,7 @@ func (h HTTPController) ScanRegistry(c *gin.Context) {
...
```
matthyx: **Blocker (same as `GenerateSBOM`).** `ValidateScanRegistry` returns `domain.ErrTooManyRequests` at `core/services/scan.go:921-924` — keyed on `ImageTagNormalized` here. Same fix: map that error to `429 Too Many Requests` rather than `400`.

controllers/http_test.go
```diff
@@ -51,8 +51,8 @@ func TestHTTPController_GenerateSBOM(t *testing.T) {
 		{
 			name:         "validation error",
 			scanService:  services.NewMockScanService(false),
-			expectedCode: http.StatusInternalServerError,
-			expectedBody: "{\"detail\":\"ImageHash=k8s.gcr.io/kube-proxy@sha256:c1b135231b5b1a6799346cd701da4b59e5b7ef8e694ec7b04fb23b8dbe144137\",\"status\":500,\"title\":\"Internal Server Error\"}",
+			expectedCode: http.StatusBadRequest,
+			expectedBody: "{\"detail\":\"ImageHash=k8s.gcr.io/kube-proxy@sha256:c1b135231b5b1a6799346cd701da4b59e5b7ef8e694ec7b04fb23b8dbe144137\",\"status\":400,\"title\":\"Bad Request\"}",
```
matthyx, line 55: These assertions all run through `MockScanService`, whose `Validate*` methods collapse the sad path into one generic error (`core/services/mockscan.go:52-78`). That means the `ErrTooManyRequests` branch of the real validators has no coverage, and the regression flagged above passes CI unnoticed.

Worth adding a case that primes the `tooManyRequests` cache on a real `ScanService` and asserts `429`, so the two error classes stay pinned apart.

matthyx (commented):
Reviewed against `main` and ran `go test ./controllers/... ./cmd/...` on `635c882` — green. The direction (validation failure is a client error, not a server fault) is right, but the change collapses *every* error `Validate*` can return into `400`, and two of those errors are not client errors. Three blockers below.

### Blocker 1 — `ErrTooManyRequests` now returns `400`

`ValidateGenerateSBOM`, `ValidateScanCVE` and `ValidateScanRegistry` do not only validate input. Each one ends with a registry back-pressure check (`core/services/scan.go:852-855`, `899-902`, `921-924`):

```go
// check if previous image pull resulted in TOOMANYREQUESTS error
if _, ok := s.tooManyRequests.Get(workload.ImageHash); ok {
    return ctx, domain.ErrTooManyRequests
}
```

That entry is set by `checkCreateSBOM` (`core/services/scan.go:79-86`) when the upstream registry answered `429`, and it expires after `ttl`. So the condition is **transient, server-side, and self-clearing** — the client's request was perfectly well-formed. `500` was already wrong, but `400` is wrong in a more damaging direction: `400` semantically means "don't send this again, it is broken", whereas the correct answer here is `429 Too Many Requests` (ideally with `Retry-After`), which is exactly the code kubevuln itself already recognises from upstream (`adapters/v1/backend_utils.go:103`).

Suggested shape for all four handlers:

```go
ctx, err = h.scanService.ValidateGenerateSBOM(ctx, newScan)
if err != nil {
    logger.L().Ctx(ctx).Error("validation error", ...)
    status := http.StatusBadRequest
    if errors.Is(err, domain.ErrTooManyRequests) {
        status = http.StatusTooManyRequests
    }
    _, _ = problem.Of(status).Append(details).WriteTo(c.Writer)
    return
}
```

### Blocker 2 — the `ScanCP` case the PR claims to fix still panics

The description says the change covers "missing application-profile arguments", but `ScanCP` dereferences those arguments *before* it calls the validator (`controllers/http.go:105-106`):

```go
name := newScan.Args[domain.ArgsName].(string)
namespace := newScan.Args[domain.ArgsNamespace].(string)
```

`Args` is copied verbatim from the request body, so an unchecked type assertion on a missing key panics before line 116 is ever reached. Verified on this branch — POST to `/v1/scanApplicationProfile` with `{"wlid":"wlid://cluster-x/namespace-y/deployment-z","imageTag":"nginx:latest"}` (no `args`):

```
interface conversion: interface {} is nil, not string
  controllers/http.go:105 HTTPController.ScanCP
=> STATUS=500 BODY=""
```

gin's `Recovery` turns it into a bodyless `500`, so the headline symptom from #416 is unfixed for the exact input the PR names. `ValidateScanCP` has the same unchecked assertion at `core/services/scan.go:864-865`. Please use the comma-ok form and return `ErrMissingCpInfo`:

```go
name, _ := newScan.Args[domain.ArgsName].(string)
namespace, _ := newScan.Args[domain.ArgsNamespace].(string)
```

### Blocker 3 — `docs/API.md` still documents the old contract

This PR changes a documented public API response but leaves `docs/API.md` contradicting the code in six places:

- `docs/API.md:152`, `:205`, `:255`, `:310` — `| \`500 Internal Server Error\` | Validation failed |` in each of the four endpoint tables
- `docs/API.md:409` — `| \`500\` | Internal Server Error | Validation failed or internal error |`
- `docs/API.md:427` — the example error body still shows `"status": 500, "title": "Internal Server Error"`

Whichever codes you settle on for blockers 1 and 2, please update these in the same PR.

### Non-blocking

- **No regression test for the rate-limit branch.** Every test this PR touches goes through `MockScanService`, whose `Validate*` methods return a single generic error for the sad path (`core/services/mockscan.go:52-78`), so the `ErrTooManyRequests` path has no coverage at all — which is why this regression is invisible to CI. A test that primes the `tooManyRequests` cache and asserts `429` would lock blocker 1 down.
- **`400` is now ambiguous.** The malformed-JSON branch (`controllers/http.go:45`, `100`, `149`, `219`) and the validation branch now return the same code, distinguishable only by the free-text `detail`. #416 explicitly offers `422 Unprocessable Entity` as an alternative — worth considering so bind failures and semantic-validation failures stay separable by clients.

controllers/http.go
rootp1: Fixed. Added a validationStatusCode helper that maps ErrTooManyRequests to 429 and everything else to 400, used by all four handlers.

controllers/http.go
rootp1: Fixed. Switched to comma-ok assertions in ScanCP and ValidateScanCP so malformed args fall through to ErrMissingCpInfo instead of panicking. Added a handler test with a bare gin.New() router that would fail if this regressed.

controllers/http.go
rootp1: Fixed. Same 429 mapping applied here.

controllers/http.go
rootp1: Fixed. Same 429 mapping applied here.

controllers/http_test.go
rootp1: Added a handler test that primes the mock to return ErrTooManyRequests and asserts 429, so this branch has coverage now.

rootp1:
@matthyx addressed all three blockers: registry rate-limit errors now map to 429 instead of 400, ScanCP uses comma-ok assertions so it no longer panics before validation, and docs/API.md is updated to match the new status codes. Also added handler tests for the 429 path and the no-panic case. Could you take another look when you have a chance?

controllers/http_test.go
```diff
@@ -416,3 +416,71 @@ func TestHTTPController_ContextCancellationIsDetached(t *testing.T) {
...
+	router.POST("/v1/scanCP", c.ScanCP)
+
+	payload := `{
+		"wlid": "wlid://cluster-x/namespace-y/deployment-z",
+		"imageTag": "nginx:latest"
+	}`
+	req, _ := http.NewRequest("POST", "/v1/scanCP", strings.NewReader(payload))
+	w := httptest.NewRecorder()
+
+	assert.NotPanics(t, func() {
```
matthyx, line 483: Non-blocking: the `gin.New()` comment explaining why `Recovery` is omitted is exactly right, and this does pin the panic regression.

One gap though — `contextSpyScanService.ValidateScanCP` returns `nil` (`controllers/http_test.go:332-334`), so this request ends at `200 OK` and the assertion passes as long as nothing panics. It doesn't pin the *outcome* the PR is about, i.e. that missing args now yield `400`. Adding `assert.Equal(t, http.StatusBadRequest, w.Code)` against a stub whose `ValidateScanCP` returns `domain.ErrMissingCpInfo` would cover that half too.

matthyx (commented):
Re-reviewed at `648cc3b`. **All three blockers are resolved.** Verified by checking out the branch, running `go build ./...`, `go vet`, the full `go test ./...` (green), and a few throwaway probe tests of my own to confirm the behaviour end-to-end rather than just reading the diff.

**Blocker 1 — `ErrTooManyRequests` → `429`.** `validationStatusCode` (`controllers/http.go:183-190`) is the right shape, and it's wired into all four handlers. Confirmed live: driving a validator that returns `domain.ErrTooManyRequests` through `ScanCVE` and `ScanRegistry` — the two endpoints without a dedicated test — both produce

```
{"detail":"...","status":429,"title":"Too Many Requests"}
```

**Blocker 2 — `ScanCP` panic.** Comma-ok is in place at both sites (`controllers/http.go:105-106` and `core/services/scan.go:864-865`). I exercised the *real* `ValidateScanCP` with a nil `Args` map, an empty map, `name`-only, and wrong-typed values (`int`/`bool`) — no panic in any case, all four return `ErrMissingCpInfo`, and the handler turns that into `400`. The original repro from my last review now returns `400` instead of a bodyless `500`.

**Blocker 3 — docs.** `docs/API.md` matches the code; no stale "Validation failed → 500" rows remain. Nice touch leaving `429` off the `applicationProfileScan` table — `ValidateScanCP` genuinely can't return it, so that's correct rather than an omission.

CI is green on `648cc3b`.

### Two small test-coverage notes (non-blocking, your call)

- `validateErrScanService` overrides `ValidateGenerateSBOM`/`ValidateScanCVE`/`ValidateScanRegistry` but not `ValidateScanCP`, and the `429` handler test only covers `GenerateSBOM`. `ScanCVE` and `ScanRegistry` work (I checked), they're just not pinned by CI — a table over the three endpoints would cost a few lines.
- The comma-ok fix in `core/services/scan.go:864-865` has no direct test; the existing `ValidateScanCP` tests in `core/services/scan_test.go` all pass well-formed args. One case with a nil `Args` map asserting `ErrMissingCpInfo` would lock it down.

Nothing here blocks merge from my side.

matthyx (approved):

PR #420, opened 2026-08-02
fix(repositories): synchronize labels and annotations on updateVEX

### LINKED ISSUE #418: Improvement: StoreVEX update path should synchronize Labels and Annotations metadata
### Description
Currently, when updating a VEX container, the `updateVEX` method inside `repositories/apiserver.go` only synchronizes the `Spec` field. It does not update the `Labels` and `Annotations` metadata from the incoming `CVEManifest` (which the `createVEX` path does). This leaves the stored VEX metadata frozen at whatever the first-ever scan wrote.

As noted by @matthyx in the code review on PR #412:
> "unlike all four siblings, the StoreVEX update path still never refreshes Labels/Annotations from cvep — the siblings all mergeMaps(result.Annotations, manifest.Annotations) before Update. updateVEX only touches Spec..."

### Proposed Solution
Synchronize `Labels` and `Annotations` during `updateVEX` using the `mergeMaps` helper function, aligning it with the other `Store*` sibling methods.

Here is the current `mergeMaps` usage across sibling methods:

Image

I would love to work on this and can submit a PR with the fix and unit test coverage!

NaitikRishu:
## Overview
This PR adds the missing metadata (labels and annotations) synchronization to the `StoreVEX` update path inside `repositories/apiserver.go`. 

When updating a VEX container, it now merges labels and annotations from the incoming `CVEManifest` using the `mergeMaps` helper, aligning it with the other sibling storage methods.

## Additional Information
N/A

## How to Test
A new unit test `TestAPIServerStore_updateVEX_mergesMetadataOnUpdate` has been added to verify that initial labels/annotations are created properly, and that consecutive updates to metadata are correctly merged (new keys added, existing keys updated) instead of remaining frozen.

Run the following command:
`TMPDIR=/tmp go test -count=1 -v ./repositories -run TestAPIServerStore_updateVEX_mergesMetadataOnUpdate`

## Examples/Screenshots
**Before (Failing test showing the bug):**
Screenshot 2026-08-02 at 10 30 44 AM

**After (Passing test showing the fix):**
Screenshot 2026-08-02 at 10 31 33 AM

## Related issues/PRs:
* Resolved #418

## Summary by CodeRabbit

- **Bug Fixes**
  - VEX updates now preserve existing metadata while incorporating new annotations and labels.
  - Metadata is retained consistently when VEX records are updated.

### DIFF
--- a/repositories/apiserver_test.go
+++ b/repositories/apiserver_test.go
@@ -601,6 +601,72 @@ func TestAPIServerStore_updateVEX_doesNotNormalizeCorrectShapeWithNonURLDataSour
 	}
 }
 
+func TestAPIServerStore_updateVEX_mergesMetadataOnUpdate(t *testing.T) {
+	cveManifestFull := tools.FileToCVEManifest("testdata/nginx-cve.json")
+	cveManifestFiltered := tools.FileToCVEManifest("testdata/nginx-cve-filtered.json")
+	require.NotEmpty(t, cveManifestFull.Content.Matches)
+
+	a := NewFakeAPIServerStorage("kubescape")
+	ctx := context.TODO()
+	workload := domain.ScanCommand{
+		ImageHash:     "sha256:32fdf92b4e986e109e4db0865758020cb0c3b70d6ba80d02fe87bad5cc3dc228",
+		InstanceID:    "apiVersion-apps/v1/namespace-kubescape/kind-ReplicaSet/name-kubevuln-65bfbfdcdd/containerName-kubevuln",
+		Wlid:          "wlid://cluster-aaa/namespace-anyNamespaceJob/job-anyJob",
+		ImageTag:      "registry.k8s.io/coredns/coredns:v1.10.1",
+		ContainerName: "anyJobContName",
+	}
+	ctx = context.WithValue(ctx, domain.WorkloadKey{}, workload)
+
+	// Create with initial annotations/labels preserving existing metadata
+	if cveManifestFull.Annotations == nil {
+		cveManifestFull.Annotations = make(map[string]string)
+	}
+	cveManifestFull.Annotations["keep-me"] = "init-val"
+	cveManifestFull.Annotations["k1"] = "v1"
+
+	if cveManifestFiltered.Annotations == nil {
+		cveManifestFiltered.Annotations = make(map[string]string)
+	}
+	cveManifestFiltered.Annotations["keep-me"] = "init-val"
+	cveManifestFiltered.Annotations["k1"] = "v1"
+
+	if cveManifestFull.Labels == nil {
+		cveManifestFull.Labels = make(map[string]string)
+	}
+	cveManifestFull.Labels["l1"] = "v2"
+
+	if cveManifestFiltered.Labels == nil {
+		cveManifestFiltered.Labels = make(map[string]string)
+	}
+	cveManifestFiltered.Labels["l1"] = "v2"
+
+	require.NoError(t, a.StoreVEX(ctx, cveManifestFull, cveManifestFiltered, false))
+
+	// Verify they are created correctly
+	vexContainer, err := a.StorageClient.OpenVulnerabilityExchangeContainers(a.Namespace).Get(context.Background(), cveManifestFull.Name, metav1.GetOptions{})
+	require.NoError(t, err)
+	assert.Equal(t, "v1", vexContainer.Annotations["k1"])
+	assert.Equal(t, "v2", vexContainer.Labels["l1"])
+	assert.Equal(t, "init-val", vexContainer.Annotations["keep-me"])
+
+	// Update with new annotations/labels (omitting keep-me to verify merge retention)
+	cveManifestFull.Annotations = map[string]string{"k1": "new-v1", "k2": "v2"}
+	cveManifestFull.Labels = map[string]string{"l1": "new-v2", "l2": "v3"}
+	cveManifestFiltered.Annotations = map[string]string{"k1": "new-v1", "k2": "v2"}
+	cveManifestFiltered.Labels = map[string]string{"l1": "new-v2", "l2": "v3"}
+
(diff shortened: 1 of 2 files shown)


repositories/apiserver_test.go
```diff
@@ -601,6 +601,53 @@ func TestAPIServerStore_updateVEX_doesNotNormalizeCorrectShapeWithNonURLDataSour
...
+		ContainerName: "anyJobContName",
+	}
+	ctx = context.WithValue(ctx, domain.WorkloadKey{}, workload)
+
+	// Create with initial annotations/labels
+	cveManifestFull.Annotations = map[string]string{"k1": "v1"}
+	cveManifestFull.Labels = map[string]string{"l1": "v2"}
+	cveManifestFiltered.Annotations = map[string]string{"k1": "v1"}
+	cveManifestFiltered.Labels = map[string]string{"l1": "v2"}
+
```
matthyx: This is the one I'd like fixed before merge: **the test cannot tell "merge" from "replace".**

I patched `updateVEX` to do a wholesale replace instead —

```go
vexContainer.Annotations = cvep.Annotations
vexContainer.Labels = cvep.Labels
```

— and this test still passes. Every key in the second pair of maps is either an overwrite or a new key, so nothing here exercises what `mergeMaps` actually guarantees. Please add a key that exists only in the initial maps, omit it from the update, and assert it survived (same as CodeRabbit's suggestion).

Second, smaller point on the same lines: assigning the maps wholesale throws away `kubescape.io/image-id`, which `testdata/nginx-cve.json` does carry and which both `createVEX` and `updateVEX` read into `imagePullable`. Without it every statement's product degrades to `pkg:oci/?repository_url=` (I checked — all 456 statements). Prefer adding to what's already there:

```go
cveManifestFull.Annotations["k1"] = "v1"
cveManifestFiltered.Annotations["k1"] = "v1"
cveManifestFull.Labels = map[string]string{"l1": "v2"}
cveManifestFiltered.Labels = map[string]string{"l1": "v2"}
```

(`Labels` is nil in both testdata files, so assigning those is fine.)

repositories/apiserver_test.go
```diff
@@ -601,6 +601,53 @@ func TestAPIServerStore_updateVEX_doesNotNormalizeCorrectShapeWithNonURLDataSour
...
+	require.NoError(t, a.StoreVEX(ctx, cveManifestFull, cveManifestFiltered, false))
+
+	// Verify they are merged on update
+	vexContainerAfterUpdate, err := a.StorageClient.OpenVulnerabilityExchangeContainers(a.Namespace).Get(context.Background(), cveManifestFull.Name, metav1.GetOptions{})
+	require.NoError(t, err)
+	assert.Equal(t, "new-v1", vexContainerAfterUpdate.Annotations["k1"])
+	assert.Equal(t, "v2", vexContainerAfterUpdate.Annotations["k2"])
+	assert.Equal(t, "new-v2", vexContainerAfterUpdate.Labels["l1"])
+	assert.Equal(t, "v3", vexContainerAfterUpdate.Labels["l2"])
+}
```
matthyx: Nit: both manifests get identical maps, so the test doesn't pin down *where* the metadata comes from — swapping `cvep` for `cve` in `updateVEX` would pass just as happily. `createVEX` sources labels/annotations from `cvep` and your fix correctly follows it; giving the two manifests distinct values would lock that in.

matthyx (commented):
Thanks @NaitikRishu — the change in `updateVEX` is exactly what #418 asked for, and it merges cleanly on top of the `RetryOnConflict` rework that landed in #412: the two lines end up inside the retry closure, and every retry re-`Get`s a fresh object, so there is no stale-object mutation to worry about.

I ran the merge result locally: `go build ./...`, `go vet ./repositories/...` and `gofmt` are clean, and `go test ./repositories/... ./core/...` passes. I also verified the new test is a genuine regression test — reverting just the two lines in `apiserver.go` makes it fail.

**No blocker in the production change.** One thing I'd like fixed before merge and one nit, both in the test — see inline.

CI note: `pull_request_created` and `PR Image Build` are still `action_required` on this SHA, so the build/unit-test jobs have not actually run on the PR yet. I'll approve the workflow run.

repositories/apiserver_test.go
NaitikRishu: Done. I've updated the test to assign individual keys instead of replacing the maps wholesale, preserving annotations like `kubescape.io/image-id`. 

I also added a `keep-me` key to the initial map, omitted it from the update, and verified it survived to test the actual merge logic.

matthyx:
Re-checked at `17d5f81`. Both points are addressed — thanks.

The test now actually guards the fix. I re-ran the two mutations against the merge result on top of current `main`:

- removing the two `mergeMaps` lines from `updateVEX` → **fails** (as before)
- replacing them with a wholesale `vexContainer.Annotations = cvep.Annotations` / `= cvep.Labels` → **fails** on `keep-me`

That second one passed on the previous revision, so the `keep-me` key is doing exactly the job I asked for.

Building the initial maps additively also fixed the `kubescape.io/image-id` problem: products now come out as `pkg:oci/nginx@sha256:32fdf92b...?repository_url=docker-pullable:%2F%2Fdocker.io%2Flibrary` instead of the degenerate `pkg:oci/?repository_url=`, and the container keeps the image-id annotation through the update (statement count stays at 456, no duplicates).

`go build ./...`, `go vet ./repositories/...`, `gofmt` clean; `go test ./repositories/... ./core/...` passes.

The remaining nit from my last review — giving `cveManifestFull` and `cveManifestFiltered` distinct values so the test pins down that metadata is sourced from `cvep` and not `cve` — is still open, but it's genuinely a nit and I won't hold the PR for it.

The force-push reset the workflow approval, so I've re-approved `pull_request_created` and `PR Image Build`; they're running now.

matthyx (approved):

PR #424, opened 2026-08-02
fix: port GCP Workload-Identity ADC fallback to sidecar scanner

### LINKED ISSUE #423: fix: sidecar scanner missing GCP Workload-Identity ADC fallback present in the in-process syft adapter
## Description

Private GCR / Google Artifact Registry images that rely on Workload Identity
(Application Default Credentials, ADC) scan successfully through the in-process
Syft adapter but fail with `401 Unauthorized` when kubevuln runs in sidecar
mode (`SBOM_SCANNER_SOCKET` set, per `cmd/http/main.go:72-86`).

PR #351 (`7c80e87`) added the GCP ADC fallback only to `adapters/v1/syft.go` —
the `isGCPRegistry` / `gcpCredentials` helpers (L36-51) and the GCP-attempt
step in the 401 path (L204-222). The structural twin,
`pkg/sbomscanner/v1/server.go`, was never ported: its 401 path (L110-116)
only attempts the anonymous fallback.

Confirmed via git history: `7c80e87` touched only `syft.go` / `syft_test.go` /
`go.mod`, and `git log -S isGCPRegistry -- pkg/sbomscanner/v1/server.go` is
empty across all history.

## Environment

- **OS:** Any (GKE / Google Cloud)
- **Version:** `main` — `pkg/sbomscanner/v1/server.go`

## Steps To Reproduce

1. Deploy kubevuln with `SBOM_SCANNER_SOCKET` set (sidecar / memory-isolated mode).
2. Scan a private image hosted on `gcr.io` / `*-docker.pkg.dev` using Workload
   Identity (no explicit registry credentials).
3. Observe: the sidecar returns `401 Unauthorized` and the anonymous fallback
   also fails.
4. Under the same IAM / Workload Identity setup, the in-process adapter scans
   the same image successfully.

## Expected Behavior

The sidecar server attempts GCP ADC for GCP-registry images before falling back
to anonymous access, matching the in-process adapter (syft.go:204-222).

## Actual Behavior

The sidecar only retries anonymously (server.go:110-116), so private GCR/GAR
images fail to scan in sidecar mode — a silent production inconsistency where
switching deployment topology breaks GKE registry integration.

## Proposed Solution

Port the two helpers (`isGCPRegistry`, `gcpCredentials`) and the GCP-attempt
step into `server.go`, mirroring the syft.go pattern. No new module
dependencies (`golang.org/x/oauth2/google` is already in go.mod;
`stereoscope/pkg/image` is already imported). Add `TestIsGCPRegistry` mirroring
syft_test.go's cases.

## Additional Context

- The original fallback was authored in PR #351 by the maintainer; the sidecar
  simply never received the port.
- The fix is scoped to the 401 path only; the anonymous fallback and all other
  behavior are unchanged.
- I can open a PR with the port and the helper test if this looks good.

AdeshDeshmukh:
Fixes #423

## Overview

PR #351 added a GCP Workload-Identity / ADC fallback to the in-process Syft
adapter (`adapters/v1/syft.go`), but the gRPC sidecar server
(`pkg/sbomscanner/v1/server.go`) never received the port. In sidecar mode
(`SBOM_SCANNER_SOCKET` set), private `gcr.io` / `*-docker.pkg.dev` images
relying on Workload Identity fail with 401, while the same image scans fine
through the in-process adapter. Verified via git history: `7c80e87` touched
only `syft.go`/`syft_test.go`/`go.mod`; `isGCPRegistry` has never appeared in
`server.go`.

### pkg/sbomscanner/v1/server.go
- Port `isGCPRegistry` and `gcpCredentials` (mirrors syft.go:36-51).
- Extract the retry chain (MANIFEST_UNKNOWN + 401/ADC/anonymous) into
  `resolveSource`, behind a `sourceGetter` seam so the fallback ordering is
  unit-testable with a scripted implementation.
- The anonymous fallback now runs on any ADC failure, not just a 401. The
  original 401 is stashed and restored when the anonymous retry fails with a
  non-401, so the response still classifies as `helpersv1.Unauthorize`.
- `gcpErr` is now logged so operators can see why ADC failed.
- Decision note on context: `FindDefaultCredentials` honors the request `ctx`;
  `Token()` performs the actual token fetch without a context, so only
  credential lookup is cancellable. The surrounding `context.Background()`
  calls in this file are out of scope here — tracked separately in #421.
- One new import (`golang.org/x/oauth2/google`, already in go.mod); no new
  module dependencies (`stereoscope/pkg/image` was already imported).

### adapters/v1/syft.go
- Same anonymous-fallback fix as the sidecar (parity — the in-process adapter
  had the identical structure per review): the retry now runs on any ADC
  failure, the original 401 is preserved for `Unauthorize` classification,
  and `gcpErr` is logged.

### pkg/sbomscanner/v1/server_test.go
- `TestResolveSource` covers the retry ordering with a scripted sourceGetter:
  non-GCP + 401 → anonymous retry; GCP + ADC unavailable → anonymous retry;
  GCP + ADC available + retry succeeds → no anonymous retry (credentials
  swapped); GCP + ADC available + retry fails non-401 → anonymous retry runs
  and the original 401 is still reported. Verified to fail with the buggy
  401-guard and pass with the fix.

## Verification

- `TMPDIR=/tmp go test ./pkg/sbomscanner/v1/... -count=1` — 19 passed
- `go build ./...` — clean
- `gofmt -l` / `go vet` on the changed files — clean
  (the socket-based tests need `TMPDIR=/tmp` on macOS due to Unix socket path
  length; this is a local test-harness quirk, not a code issue)

## Follow-up option

Happy to extract `isGCPRegistry`/`gcpCredentials` into a shared internal
package — and with it, scope the GCP credential's `Authority` and append rather
than replace the pull secrets — in a separate PR if you'd prefer not to keep
the two copies.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added support for pulling images from Google Container Registry and Google Artifact Registry.
  - Automatically uses available Google Cloud credentials when registry authentication is required.
  - Falls back to anonymous access when credentials are unavailable or unsuccessful.

- **Bug Fixes**
  - Improved image resolution for missing tags and authentication failures.
  - Preserves the original authorization error when fallback attempts fail, providing more accurate error reporting.

### DIFF
--- a/adapters/v1/syft.go
+++ b/adapters/v1/syft.go
@@ -202,22 +202,27 @@ func (s *SyftAdapter) CreateSBOM(ctx context.Context, name, imageID, imageTag st
 	}
 
 	if err != nil && strings.Contains(err.Error(), "401 Unauthorized") {
-		if isGCPRegistry(imageID) {
+		unauthorizedErr := err
+		if isGCPRegistry(pullRef) {
 			if gcpCreds, gcpErr := gcpCredentials(ctx); gcpErr != nil {
 				logger.L().Debug("GCP ADC unavailable, falling back to anonymous",
+					helpers.Error(gcpErr),
 					helpers.String("imageID", imageID))
 			} else {
 				registryOptions.Credentials = []image.RegistryCredentials{*gcpCreds}
-				src, err = syft.GetSource(ctxWithSize, imageID, syft.DefaultGetSourceConfig().WithRegistryOptions(&registryOptions).WithSources("registry"))
+				src, err = syft.GetSource(ctxWithSize, pullRef, syft.DefaultGetSourceConfig().WithRegistryOptions(&registryOptions).WithSources("registry"))
 			}
 		}
 		// If GCP ADC was not attempted, succeeded in auth but still got 401, or the image is not a GCP registry,
 		// fall back to anonymous access. err/src retain the result of the last attempt.
-		if err != nil && strings.Contains(err.Error(), "401 Unauthorized") {
-			logger.L().Debug("got 401, retrying without credentials",
+		if err != nil {
+			logger.L().Debug("retrying without credentials",
 				helpers.String("imageID", imageID))
 			registryOptions.Credentials = nil
-			src, err = syft.GetSource(ctxWithSize, rewriteImageRef(imageID, s.proxyRegistryMap), syft.DefaultGetSourceConfig().WithRegistryOptions(&registryOptions).WithSources("registry"))
+			src, err = syft.GetSource(ctxWithSize, pullRef, syft.DefaultGetSourceConfig().WithRegistryOptions(&registryOptions).WithSources("registry"))
+			if err != nil && !strings.Contains(err.Error(), "401 Unauthorized") {
+				err = fmt.Errorf("%w (anonymous fallback failed: %v)", unauthorizedErr, err)
+			}
 		}
 	}
 
--- a/pkg/sbomscanner/v1/server.go
+++ b/pkg/sbomscanner/v1/server.go
@@ -26,7 +26,73 @@ import (
 	helpersv1 "github.com/kubescape/k8s-interface/instanceidhandler/v1/helpers"
 	pb "github.com/kubescape/kubevuln/pkg/sbomscanner/v1/proto"
 	"github.com/kubescape/storage/pkg/apis/softwarecomposition/v1beta1"
+	"golang.org/x/oauth2/google"
 )
+
+func isGCPRegistry(imageID string) bool {
+	host, _, _ := strings.Cut(imageID, "/")
+	return host == "gcr.io" || strings.HasSuffix(host, ".gcr.io") || strings.HasSuffix(host, "-docker.pkg.dev")
+}
+
+func gcpCredentials(ctx context.Context) (*image.RegistryCredentials, error) {
+	creds, err := google.FindDefaultCredentials(ctx, "https://www.googleapis.com/auth/cloud-platform")
+	if err != nil {
+		return nil, err
+	}
+	token, err := creds.TokenSource.Token()
+	if err != nil {
+		return nil, err
+	}
+	return &image.RegistryCredentials{Username: "oauth2accesstoken", Password: token.AccessToken}, nil
+}
+
+// gcpCredsFn is an indirection over gcpCredentials so resolveSource can be
+// unit-tested without a live GCP environment.
+var gcpCredsFn = gcpCredentials
+
+// sourceGetter abstracts the syft.GetSource call so the fallback ordering in
+// resolveSource is testable with a scripted implementation.
+type sourceGetter func(ctx context.Context, ref string, opts *image.RegistryOptions) (source.Source, error)
+
+// resolveSource downloads an image source, applying the MANIFEST_UNKNOWN and
+// 401/ADC/anonymous fallbacks in order. It mirrors the retry chain in
+// adapters/v1/syft.go so the sidecar and in-process adapters behave the same.
+func resolveSource(ctx context.Context, get sourceGetter, imageID, imageTag string, opts image.RegistryOptions) (source.Source, error) {
+	pullRef := imageID
+	src, err := get(ctx, pullRef, &opts)
+	if err != nil && strings.Contains(err.Error(), "MANIFEST_UNKNOWN") {
+		logger.L().Debug("got MANIFEST_UNKNOWN, retrying with imageTag",
+			helpers.String("imageTag", imageTag),
+			helpers.String("imageID", imageID))
+		pullRef = imageTag
+		src, err = get(ctx, pullRef, &opts)
+	}
+	if err != nil && strings.Contains(err.Error(), "401 Unauthorized") {
+		unauthorizedErr := err
+		if isGCPRegistry(imageID) {
+			if gcpCreds, gcpErr := gcpCredsFn(ctx); gcpErr != nil {
+				logger.L().Debug("GCP ADC unavailable, falling back to anonymous",
+					helpers.Error(gcpErr),
+					helpers.String("imageID", imageID))
+			} else {
+				opts.Credentials = []image.RegistryCredentials{*gcpCreds}
+				src, err = get(ctx, pullRef, &opts)
+			}
+		}
+		// If GCP ADC was not attempted, succeeded in auth but still got 401, or
+		// the image is not a GCP registry, fall back to anonymous access.
+		if err != nil {
+			logger.L().Debug("retrying without credentials",
+				helpers.String("imageID", imageID))
+			opts.Credentials = nil
+			src, err = get(ctx, pullRef, &opts)
+			if err != nil && !strings.Contains(err.Error(), "401 Unauthorized") {
+				err = fmt.Errorf("%w (anonymous fallback failed: %v)", unauthorizedErr, err)
+			}
+		}
+	}
+	return src, err
+}
 
 type scannerServer struct {
 	pb.UnimplementedSBOMScannerServer
@@ -95,25 +161,12 @@ func (s *scannerServer) CreateSBOM(ctx context.Context, req *pb.CreateSBOMReques
 
 	// Download image from registry
 	logger.L().Debug("downloading image", helpers.String("imageID", imageID))
-	ctxWithSize := context.WithValue(context.Background(), image.MaxImageSize, req.MaxImageSize)
-	src, err := syft.GetSource(ctxWithSize, imageID,
-		syft.DefaultGetSourceConfig().WithRegistryOptions(&registryOptions).WithPlatform(imgPlatform).WithSources("registry"))
-
-	if err != nil && strings.Contains(err.Error(), "MANIFEST_UNKNOWN") {
-		logger.L().Debug("got MANIFEST_UNKNOWN, retrying with imageTag",
-			helpers.String("imageTag", imageTag),
-			helpers.String("imageID", imageID))
-		src, err = syft.GetSource(ctxWithSize, imageTag,
-			syft.DefaultGetSourceConfig().WithRegistryOptions(&registryOptions).WithPlatform(imgPlatform).WithSources("registry"))
-	}
-
-	if err != nil && strings.Contains(err.Error(), "401 Unauthorized") {
-		logger.L().Debug("got 401, retrying without credentials",
-			helpers.String("imageID", imageID))
-		registryOptions.Credentials = nil
-		src, err = syft.GetSource(ctxWithSize, imageID,
-			syft.DefaultGetSourceConfig().WithRegistryOptions(&registryOptions).WithPlatform(imgPlatform).WithSources("registry"))
-	}
+	src, err := resolveSource(ctx, func(_ context.Context, ref string, opts *image.RegistryOptions) (source.Source, error) {
+		// Pulls intentionally run on a detached context (see #421); the request ctx is used only for gcpCredentials inside resolveSource.
+		ctxWithSize := context.WithValue(context.Background(), image.MaxImageSize, req.MaxImageSize)
+		return syft.GetSource(ctxWithSize, ref,
+			syft.DefaultGetSourceConfig().WithRegistryOptions(opts).WithPlatform(imgPlatform).WithSources("registry"))
+	}, imageID, imageTag, registryOptions)
 
 	switch {
 	case err != nil && strings.Contains(err.Error(), image.ErrImageTooLarge.Error()):
--- a/pkg/sbomscanner/v1/server_test.go
+++ b/pkg/sbomscanner/v1/server_test.go
@@ -2,11 +2,14 @@ package v1
 
 import (
 	"context"
+	"errors"
 	"net"
 	"os"
 	"path/filepath"
 	"testing"
 
+	"github.com/anchore/stereoscope/pkg/image"
+	"github.com/anchore/syft/syft/source"
 	pb "github.com/kubescape/kubevuln/pkg/sbomscanner/v1/proto"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -70,4 +73,147 @@ func TestCreateSBOM_ContextCancelled(t *testing.T) {
 	})
 	assert.Nil(t, resp)
 	require.Error(t, err)
+}
+
+func TestIsGCPRegistry(t *testing.T) {
+	tests := []struct {
+		imageID string
+		want    bool
+	}{
+		{"gcr.io/foo/bar", true},
+		{"us.gcr.io/foo/bar", true},
+		{"us-docker.pkg.dev/foo/bar", true},
+		{"europe-west1-docker.pkg.dev/project/repo/image:tag", true},
+		{"quay.io/foo/bar", false},
+		{"quay.io/foo/bar-docker.pkg.dev/x", false},
+		{"index.docker.io/library/alpine", false},
+		{"", false},
+	}
+	for _, tt := range tests {
+		t.Run(tt.imageID, func(t *testing.T) {
+			assert.Equal(t, tt.want, isGCPRegistry(tt.imageID))
+		})
+	}
+}
+
+func TestResolveSource(t *testing.T) {
+	err401 := errors.New("401 Unauthorized")
+	err403 := errors.New("403 Forbidden")
+	manifestUnknownErr := errors.New("MANIFEST_UNKNOWN")
+
+	tests := []struct {
+		name            string
+		imageID         string
+		imageTag        string
+		results         []error
+		adcErr          error
+		wantCalls       int
+		wantLastCreds   []image.RegistryCredentials
+		wantErr         bool
+		wantErrIs       error
+		wantErrContains string
+		wantRefs        []string
+	}{
+		{
+			name:      "non-GCP image + 401 falls back to anonymous",
+			imageID:   "docker.io/library/nginx",
+			results:   []error{err401, err403},
+			wantCalls: 2,
+			wantErr:   true,
+		},
+		{
+			name:      "GCP image, ADC unavailable, falls back to anonymous",
+			imageID:   "gcr.io/project/image",
+			results:   []error{err401, err403},
+			adcErr:    errors.New("ADC unavailable"),
+			wantCalls: 2,
+			wantErr:   true,
+		},
+		{
+			name:    "GCP image, ADC available, retry succeeds",
+			imageID: "gcr.io/project/image",
+			results: []error{err401, nil},
+			wantLastCreds: []image.RegistryCredentials{
+				{Username: "oauth2accesstoken", Password: "token"},
+			},
+			wantCalls: 2,
+		},
+		{
+			name:      "GCP image, ADC succeeds but pull fails non-401, still anonymous and restores Unauthorize",
+			imageID:   "gcr.io/project/image",
+			results:   []error{err401, err403, err403},
+			wantCalls: 3,
+			wantErr:   true,
+			wantErrIs: err401,
+		},
+		{
+			name:     "MANIFEST_UNKNOWN on imageID, 401 on imageTag, retries use imageTag",
+			imageID:  "gcr.io/project/image@sha256:abc",
+			imageTag: "gcr.io/project/image:latest",
+			results:  []error{manifestUnknownErr, err401, err403, err403},
+			wantRefs: []string{
+				"gcr.io/project/image@sha256:abc",
+				"gcr.io/project/image:latest",
+				"gcr.io/project/image:latest",
+				"gcr.io/project/image:latest",
+			},
+			wantCalls: 4,
+			wantErr:   true,
+			wantErrIs: err401,
+		},
+		{
+			name:            "too-large during anonymous retry is not masked by original 401",
+			imageID:         "gcr.io/project/image",
+			adcErr:          errors.New("ADC unavailable"),
+			results:         []error{err401, image.ErrImageTooLarge},
+			wantCalls:       2,
+			wantErr:         true,
+			wantErrContains: image.ErrImageTooLarge.Error(),
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			orig := gcpCredsFn
+			defer func() { gcpCredsFn = orig }()
+			gcpCredsFn = func(ctx context.Context) (*image.RegistryCredentials, error) {
+				if tt.adcErr != nil {
+					return nil, tt.adcErr
+				}
+				return &image.RegistryCredentials{Username: "oauth2accesstoken", Password: "token"}, nil
+			}
+
+			var calls [][]image.RegistryCredentials
+			var refs []string
+			callCount := 0
+			get := func(ctx context.Context, ref string, opts *image.RegistryOptions) (source.Source, error) {
+				refs = append(refs, ref)
+				creds := append([]image.RegistryCredentials(nil), opts.Credentials...)
+				calls = append(calls, creds)
+				res := tt.results[callCount]
+				callCount++
+				return nil, res
+			}
+
+			_, err := resolveSource(context.Background(), get, tt.imageID, tt.imageTag, image.RegistryOptions{})
+			if tt.wantErr {
+				require.Error(t, err)
+			} else {
+				require.NoError(t, err)
+			}
+			if tt.wantErrIs != nil {
+				assert.ErrorIs(t, err, tt.wantErrIs)
+			}
+			if tt.wantErrContains != "" {
+				assert.Contains(t, err.Error(), tt.wantErrContains)
+			}
+			assert.Equal(t, tt.wantCalls, callCount)
+			if tt.wantLastCreds != nil {
+				assert.Equal(t, tt.wantLastCreds, calls[len(calls)-1])
+			}
+			if tt.wantRefs != nil {
+				assert.Equal(t, tt.wantRefs, refs)
+			}
+		})
+	}
 }


pkg/sbomscanner/v1/server.go
```diff
@@ -108,11 +126,25 @@ func (s *scannerServer) CreateSBOM(ctx context.Context, req *pb.CreateSBOMReques
...
+					helpers.String("imageID", imageID))
+			} else {
+				registryOptions.Credentials = []image.RegistryCredentials{*gcpCreds}
+				src, err = syft.GetSource(ctxWithSize, imageID,
+					syft.DefaultGetSourceConfig().WithRegistryOptions(&registryOptions).WithPlatform(imgPlatform).WithSources("registry"))
+			}
+		}
+		// If GCP ADC was not attempted, succeeded in auth but still got 401, or the
+		// image is not a GCP registry, fall back to anonymous access.
+		if err != nil && strings.Contains(err.Error(), "401 Unauthorized") {
```
matthyx: **Blocker — the anonymous fallback is now skipped for non-401 failures of the ADC attempt.**

We only reach this outer block because the *original* error was a 401, so an anonymous retry is always warranted from here. But line 136 overwrote `err` with the ADC attempt's error, and this guard re-tests for the `"401 Unauthorized"` substring. So any ADC-attempt failure that isn't literally a 401 falls straight through to the `switch` at line 156:

- Artifact Registry / GCR answer an *authenticated but unauthorised* request with `403 Forbidden` / `denied`, not 401. That is the normal outcome when the workload-identity SA has no IAM binding on the image's project — which is precisely the situation for public images like `gcr.io/distroless/*`, `k8s.gcr.io/*` (matches `.gcr.io`) or a public AR repo pulled from a GKE cluster. Before this PR they resolved anonymously; now they fail.
- Same for any transport-level failure of the token-bearing pull (`connection reset`, `context deadline exceeded`), regardless of what GCR's status codes happen to be.

Second effect: the original 401 is lost, so `case err != nil && strings.Contains(err.Error(), "401 Unauthorized")` at line 158 no longer matches. The response carries an empty `Status` instead of `helpersv1.Unauthorize`, and `adapters/v1/sidecar.go:110` surfaces it as a generic error — the backend loses the "this image needs credentials" signal it used to get.

Minimal fix — drop the substring re-test:

```go
		if err != nil {
			logger.L().Debug("retrying without credentials",
			registryOptions.Credentials = nil
			src, err = syft.GetSource(ctxWithSize, imageID,
				syft.DefaultGetSourceConfig().WithRegistryOptions(&registryOptions).WithPlatform(imgPlatform).WithSources("registry"))
		}
```

That is safe in every branch: ADC not attempted → `err` is still the original 401 → retry, as before. ADC succeeded → `err == nil` → skipped. ADC attempted and failed for any reason → retry, which is the case being fixed.

If you also want to preserve the `Unauthorize` classification, stash `unauthorizedErr := err` at the top of the outer block and restore it when the anonymous retry fails with a non-401 — the original 401 is the more accurate root cause to report.

Same structure exists in `adapters/v1/syft.go:216`, so the in-process adapter needs the identical change.

pkg/sbomscanner/v1/server.go
```diff
@@ -108,11 +126,25 @@ func (s *scannerServer) CreateSBOM(ctx context.Context, req *pb.CreateSBOMReques
...
 
 	if err != nil && strings.Contains(err.Error(), "401 Unauthorized") {
-		logger.L().Debug("got 401, retrying without credentials",
-			helpers.String("imageID", imageID))
-		registryOptions.Credentials = nil
-		src, err = syft.GetSource(ctxWithSize, imageID,
-			syft.DefaultGetSourceConfig().WithRegistryOptions(&registryOptions).WithPlatform(imgPlatform).WithSources("registry"))
+		if isGCPRegistry(imageID) {
+			if gcpCreds, gcpErr := gcpCredentials(ctx); gcpErr != nil {
+				logger.L().Debug("GCP ADC unavailable, falling back to anonymous",
```
matthyx: Agreeing with Copilot here: `gcpErr` is discarded, so "ADC unavailable" is indistinguishable from a missing workload-identity binding, an unreachable metadata server, or a malformed `GOOGLE_APPLICATION_CREDENTIALS`. In a sidecar container that operators can't easily attach to, that error is the whole diagnostic.

```go
				logger.L().Debug("GCP ADC unavailable, falling back to anonymous",
					helpers.Error(gcpErr),
```

No secret exposure risk — `FindDefaultCredentials`/`Token()` errors don't carry the token.

matthyx (commented):
## Review

The gap this PR describes is real and I confirmed it independently: `isGCPRegistry`/`gcpCredentials` exist only in `adapters/v1/syft.go` on `main`, so `SBOM_SCANNER_SOCKET` mode never got the ADC fallback from #351. Build, `go vet` and `go test ./pkg/sbomscanner/v1/... -count=1` are all green on `d4ba3b8` here (go1.25.8, Linux — no `TMPDIR` workaround needed).

But the ported retry chain has one behavioural regression I'd treat as a blocker, and the test that ships with it doesn't cover the part that regressed.

### Blockers

1. **The anonymous fallback becomes unreachable whenever the ADC attempt fails with anything other than a 401** (`server.go:141`). The ADC retry overwrites `err`, and the inner guard re-tests for the `"401 Unauthorized"` substring — so a 403/`denied`, a token-fetch transport error, or a timeout on the ADC-authenticated pull all skip the anonymous retry that used to run unconditionally on this path. Public GCP-hosted images (`gcr.io/distroless/*`, `k8s.gcr.io/*`, public AR repos) scanned from a GKE cluster whose workload-identity SA has no access to that project are the concrete case: they previously succeeded anonymously, and now fail. Secondary effect — because the original 401 is gone, `server.go:158` no longer classifies the result as `helpersv1.Unauthorize`, so the response goes back with an empty status and a generic `ErrorMessage`, which `adapters/v1/sidecar.go:110` turns into a plain error. Details and a minimal fix inline.

2. **The behaviour change has no test.** The only new test covers `isGCPRegistry`, a pure predicate that is already tested case-for-case in `adapters/v1/syft_test.go`. Everything this PR actually changes — retry ordering, which errors trigger which fallback, whether `Unauthorize` still gets reported — is untested, and that is exactly where (1) lives. Inline note on a seam that makes it testable without a live registry.

### Non-blocking

- Seconding Copilot on the dropped `gcpErr` — see inline, it's one `helpers.Error(gcpErr)`.
- `registryOptions.Credentials` is replaced rather than appended, and the GCP credential goes in with an empty `Authority`. Same as `syft.go`, so not a new problem, but flagged inline.
- On the offer to extract a shared package: worth taking, and the two copies have **already** drifted rather than being at risk of drifting — see the inline note on `server.go:129`. Duplication is what caused this bug report in the first place.
- The description says passing the request `ctx` means "the short token fetch respects caller cancellation". Only half true: `google.FindDefaultCredentials(ctx, …)` honours it, but `creds.TokenSource.Token()` takes no context, so the actual HTTP token fetch is uncancellable. Worth softening the wording rather than changing the code.

Note that (1) is inherited from #351 — `adapters/v1/syft.go:204-222` has the identical structure — so the in-process adapter needs the same fix. Given this PR's premise is parity, fixing both here (or in an immediate follow-up) seems right.

pkg/sbomscanner/v1/server.go
```diff
@@ -108,11 +126,25 @@ func (s *scannerServer) CreateSBOM(ctx context.Context, req *pb.CreateSBOMReques
...
+					helpers.String("imageID", imageID))
+			} else {
+				registryOptions.Credentials = []image.RegistryCredentials{*gcpCreds}
```
matthyx: Two things worth a second look, both inherited from `syft.go` rather than introduced here:

1. This **replaces** the credential slice built from `req.Credentials` (the pull secrets the adapter forwarded) instead of appending. For a multi-arch pull or an image whose manifest references another registry, those credentials are now gone for the retry.
2. `gcpCredentials` returns a `RegistryCredentials` with an empty `Authority`, so the OAuth token is offered to *any* registry contacted during that retry, not just the GCP host that 401'd. Setting `Authority` to the host from `strings.Cut(imageID, "/")` — which `isGCPRegistry` already computes — would scope it properly and let you append rather than replace.

Not a merge blocker, but if you do the shared-helper extraction it'd be a good time to fix it in one place.

pkg/sbomscanner/v1/server.go
```diff
@@ -108,11 +126,25 @@ func (s *scannerServer) CreateSBOM(ctx context.Context, req *pb.CreateSBOMReques
 	}
 
 	if err != nil && strings.Contains(err.Error(), "401 Unauthorized") {
-		logger.L().Debug("got 401, retrying without credentials",
-			helpers.String("imageID", imageID))
-		registryOptions.Credentials = nil
-		src, err = syft.GetSource(ctxWithSize, imageID,
-			syft.DefaultGetSourceConfig().WithRegistryOptions(&registryOptions).WithPlatform(imgPlatform).WithSources("registry"))
+		if isGCPRegistry(imageID) {
```
matthyx: On the "extract into a shared internal package" offer in the description — worth taking, and I'd argue the copies have *already* diverged rather than merely being at risk of it:

- `adapters/v1/syft.go:205` calls `isGCPRegistry(imageID)` on the **original** reference, then retries the ADC pull against `imageID` too — bypassing `rewriteImageRef(imageID, s.proxyRegistryMap)`, which both the initial attempt (`syft.go:193`) and the anonymous fallback (`syft.go:220`) go through.
- Here, `req.ImageId` has already been rewritten by `adapters/v1/sidecar.go:81`, so this checks the **proxied** reference.

So with a `proxyRegistryMap` configured, in-process and sidecar make opposite decisions about whether to try ADC at all, and `syft.go`'s ADC retry targets a host the operator deliberately proxied away from. The sidecar's behaviour is the more defensible of the two; a single shared implementation with an explicit "which reference do we classify" rule would settle it.

Minor, same-hunk: if the `MANIFEST_UNKNOWN` retry at line 120 pulled `imageTag` and *that* is what returned 401, both fallbacks here still target `imageID` — a reference already known to fail with `MANIFEST_UNKNOWN`. Pre-existing on the anonymous path, but the new ADC branch inherits it.

pkg/sbomscanner/v1/server_test.go
```diff
@@ -71,3 +71,24 @@ func TestCreateSBOM_ContextCancelled(t *testing.T) {
 	assert.Nil(t, resp)
 	require.Error(t, err)
 }
+
```
matthyx, line 77: **Blocker — this test doesn't cover what the PR changes.**

`isGCPRegistry` is a pure string predicate, and these are the same eight cases already asserted in `adapters/v1/syft_test.go`. It passes identically before and after the behavioural change, so it can't catch a regression in the thing being shipped. The description is candid that this is a "compile-gated helper test" — the issue is that it's the *only* test, and the retry-ordering bug flagged on `server.go:141` would sail through it.

What's untested and worth covering:

- non-GCP image + 401 → anonymous retry runs (unchanged behaviour, worth pinning)
- GCP image, ADC unavailable → anonymous retry runs
- GCP image, ADC available, retry succeeds → no anonymous retry, credentials actually swapped
- GCP image, ADC available, retry fails with a **non-401** → anonymous retry still runs, and `helpersv1.Unauthorize` is still reported

Agreed a live private GCR image is out of scope, but none of the above needs one. The blocker is that the retry chain is inlined in a ~150-line `CreateSBOM` calling `syft.GetSource` directly. Lifting it out behind a seam makes all four cases table-testable with a fake:

```go
type sourceGetter func(ctx context.Context, ref string, opts *image.RegistryOptions) (source.Source, error)

func resolveSource(ctx context.Context, get sourceGetter, imageID, imageTag string, opts image.RegistryOptions) (source.Source, error)
```

`CreateSBOM` passes the real syft call; the tests pass a stub that returns scripted errors per call and records the credentials it was handed. That covers the ADC/anonymous ordering and the `Unauthorize` classification with no network and no GCP. It'd also be reusable by `syft.go` if the shared-package extraction happens.

AdeshDeshmukh:
Thanks for the review ..!

Both blockers are fixed, in the sidecar and the in-process adapter (syft.go had the same bug, so it got the same fix): the anonymous retry now runs no matter what the ADC attempt returns, the original 401 survives so Unauthorize still comes through, and gcpErr is now logged. The new tests fail on the old guard, so they're actually doing their job.

I also went ahead and fixed the MANIFEST_UNKNOWN retry-target point CodeRabbit flagged as Major — the ADC and anonymous retries now reuse the selected reference (pullRef) instead of rebuilding from imageID, with a test covering the MANIFEST_UNKNOWN → 401 sequence. For proxy-configured setups this also aligns the ADC retry with the anonymous path, consistent with the behavior you called the more defensible of the two.

adapters/v1/syft.go
```diff
@@ -202,22 +202,27 @@ func (s *SyftAdapter) CreateSBOM(ctx context.Context, name, imageID, imageTag st
...
-		if err != nil && strings.Contains(err.Error(), "401 Unauthorized") {
-			logger.L().Debug("got 401, retrying without credentials",
+		if err != nil {
+			logger.L().Debug("retrying without credentials",
 				helpers.String("imageID", imageID))
 			registryOptions.Credentials = nil
-			src, err = syft.GetSource(ctxWithSize, rewriteImageRef(imageID, s.proxyRegistryMap), syft.DefaultGetSourceConfig().WithRegistryOptions(&registryOptions).WithSources("registry"))
+			src, err = syft.GetSource(ctxWithSize, pullRef, syft.DefaultGetSourceConfig().WithRegistryOptions(&registryOptions).WithSources("registry"))
+			if err != nil && !strings.Contains(err.Error(), "401 Unauthorized") {
+				err = unauthorizedErr
```
matthyx: **This is the one thing I'd fix before merge — and it's my fault for suggesting it without qualification.**

Restoring `unauthorizedErr` discards the anonymous retry's error entirely, and the `switch` immediately below tests `image.ErrImageTooLarge` *first*:

```go
case err != nil && strings.Contains(err.Error(), image.ErrImageTooLarge.Error()):   // line 230
case err != nil && strings.Contains(err.Error(), "401 Unauthorized"):
```

Concrete sequence: supplied pull secret is stale → 401 → ADC not applicable or unavailable → anonymous retry authenticates fine and starts downloading → image exceeds `maxImageSize` → `ErrImageTooLarge`. That error is thrown away and replaced with the 401, so line 230 never matches and the image is reported `Unauthorize` instead of `Incomplete`. Those aren't interchangeable: `Incomplete` is terminal, `Unauthorize` tells the backend the image needs credentials and it comes back for another attempt that will fail the same way every time.

Wrapping instead of replacing keeps both classifications available, and the more specific one wins because of the `switch` ordering:

```go
			if err != nil && !strings.Contains(err.Error(), "401 Unauthorized") {
				err = fmt.Errorf("%w (anonymous fallback failed: %v)", unauthorizedErr, err)
			}
```

A too-large failure now hits the `ErrImageTooLarge` case; anything else still carries the 401 substring and lands on `Unauthorize` as intended; and operators get the real downstream error in the message rather than a 401 that didn't cause the failure. `errors.Is(err, unauthorizedErr)` still holds, so `TestResolveSource`'s `assert.ErrorIs(t, err, err401)` keeps passing unchanged.

Same line in the sidecar at `pkg/sbomscanner/v1/server.go:87`. Worth a sixth `TestResolveSource` case: `results: []error{err401, errTooLarge}` asserting the returned error still reports as too-large.

adapters/v1/syft.go
```diff
@@ -202,22 +202,27 @@ func (s *SyftAdapter) CreateSBOM(ctx context.Context, name, imageID, imageTag st
 	}
 
 	if err != nil && strings.Contains(err.Error(), "401 Unauthorized") {
+		unauthorizedErr := err
 		if isGCPRegistry(imageID) {
```
matthyx: **New in this round: the GCP token can now be sent to a non-Google host.**

This still classifies `imageID` — the *original* reference — but line 213 now pulls `pullRef`, which is `rewriteImageRef(imageID, s.proxyRegistryMap)`. Before this round the ADC retry used `imageID`, so the token necessarily went to Google. Now the two references can point at different hosts.

With `proxyRegistryMap: {"gcr.io": "harbor.internal.corp"}`, `rewriteImageRef` (line 78) rewrites `gcr.io/proj/img` → `harbor.internal.corp/proj/img`. `isGCPRegistry("gcr.io/proj/img")` is `true`, so the ADC branch fetches a token and hands it to the mirror as basic auth `oauth2accesstoken:<token>`. The scope requested in `gcpCredentials` is `https://www.googleapis.com/auth/cloud-platform` — the broad one, not a registry-read scope — so that's a widely-privileged bearer token landing in a third-party registry's auth path.

Classify the reference you actually pull:

```go
		if isGCPRegistry(pullRef) {
```

That's strictly more correct in both directions: a proxied GCP image no longer leaks the token, and a non-GCP image proxied *through* Artifact Registry correctly gets ADC. It also matches the sidecar, where `req.ImageId` has already been rewritten by `adapters/v1/sidecar.go:81` — so this closes the in-process/sidecar divergence I raised last round, and makes the two logic bodies identical enough to lift into the shared helper you offered.

Separately, and pre-existing from #351 — worth a follow-up, not this PR: narrowing that scope to `https://www.googleapis.com/auth/devstorage.read_only` would limit the blast radius of a token that only ever needs to read image layers.

matthyx (commented):
## Re-review of `9508b54`

Both blockers are resolved, and the fix went further than I asked — the ordering bug is fixed in `syft.go` too, so #351's copy no longer carries it. Verified locally on `9508b54` (go1.25.8, Linux): `go build ./...`, `go vet ./pkg/sbomscanner/... ./adapters/...` and `gofmt -l` on all three changed files are clean; `go test ./pkg/sbomscanner/v1/... ./adapters/v1/... -count=1` passes (`adapters/v1` 56s).

**Blocker 1 — cleared.** `if err != nil` now guards the anonymous fallback in both files, so a 403/`denied` or a transport failure on the ADC attempt no longer strands the pull. The `pullRef` change is a genuine improvement over what I proposed: threading the manifest-fallback reference through all three attempts fixes the `imageID`-vs-`imageTag` mismatch I filed as a nit and the one CodeRabbit raised on `syft.go#L218-L225`, in one move.

**Blocker 2 — cleared.** `resolveSource` with the `sourceGetter` seam is exactly the shape I had in mind, and `TestResolveSource`'s five cases cover the branches that were untested — including `results: []error{err401, err403, err403}` asserting the anonymous retry still runs after a non-401 ADC failure, which is the regression itself. `gcpCredsFn` as a swappable var is a reasonable way to keep ADC out of the test path. I confirmed all five subtests execute and pass.

Two new things, both introduced by this round rather than pre-existing. Neither blocks in the way the originals did, but the first is a misclassification I'd rather not ship.

### Should fix

1. **`err = unauthorizedErr` is too blunt and masks `ErrImageTooLarge`** (`syft.go:224`, `server.go:87`). This is my fault — I suggested restoring the original error as an afterthought and didn't think through the interaction with the `switch` below it, which tests `ErrImageTooLarge` *before* the 401 case. Detail and a one-line fix inline.

2. **`isGCPRegistry(imageID)` now classifies a different reference than the one being pulled** (`syft.go:206`). With `proxyRegistryMap` configured, the ADC branch decides "this is GCP" from the original ref but sends the token to the proxy host — so a `cloud-platform`-scoped GCP access token goes to an operator's internal mirror. This is new: before this round the ADC retry pulled `imageID`, so the token always went to Google. Inline.

### Nits

- `resolveSource` dropped the `"got MANIFEST_UNKNOWN, retrying with imageTag"` debug line that `server.go` had (and that `syft.go:196` still has). Inline.
- The shared-package offer from the description still stands — the two implementations now agree behaviourally but are still two implementations. `resolveSource` is unexported in `pkg/sbomscanner/v1`, so `syft.go` can't use it. Fixing #2 by switching to the pulled reference would make the two logic bodies identical, which is the right moment to lift one out.

pkg/sbomscanner/v1/server.go
```diff
@@ -26,8 +26,71 @@ import (
...
+// resolveSource is testable with a scripted implementation.
+type sourceGetter func(ctx context.Context, ref string, opts *image.RegistryOptions) (source.Source, error)
+
+// resolveSource downloads an image source, applying the MANIFEST_UNKNOWN and
+// 401/ADC/anonymous fallbacks in order. It mirrors the retry chain in
+// adapters/v1/syft.go so the sidecar and in-process adapters behave the same.
+func resolveSource(ctx context.Context, get sourceGetter, imageID, imageTag string, opts image.RegistryOptions) (source.Source, error) {
+	pullRef := imageID
+	src, err := get(ctx, pullRef, &opts)
+	if err != nil && strings.Contains(err.Error(), "MANIFEST_UNKNOWN") {
```
matthyx, line 63: Nit: the extraction dropped the `"got MANIFEST_UNKNOWN, retrying with imageTag"` debug line the old `server.go` had, and which `syft.go:196` still emits. Now the only visible trace of a digest→tag fallback in the sidecar is the pull failure itself. Since the retries downstream all reuse `pullRef`, knowing which reference was selected is the first thing you'd want when reading logs from a scan that ended up `Unauthorize`. Re-adding it here keeps the two adapters' log output comparable:

```go
	if err != nil && strings.Contains(err.Error(), "MANIFEST_UNKNOWN") {
		logger.L().Debug("got MANIFEST_UNKNOWN, retrying with imageTag",
			helpers.String("imageTag", imageTag),
		pullRef = imageTag
```

Also minor, on the closure at line 161: it accepts a `ctx` parameter and then shadows it with `context.Background()` on line 162, so `resolveSource`'s `ctx` reaches only `gcpCredsFn`. That's the deliberate #421 behaviour, but the shadowing reads like a bug at a glance — either name the parameter `_` or drop a one-line comment pointing at #421.

AdeshDeshmukh:
Thanks for the re-review — the ErrImageTooLarge masking was a real catch, and fair that it came from the restore you'd suggested without qualification.

Both should-fixes are in. The anonymous-retry error is now wrapped rather than replaced (fmt.Errorf("%w (anonymous fallback failed: %v)", unauthorizedErr, err)), so a too-large failure still lands on Incomplete while anything else keeps the 401 signal and Unauthorize comes back — errors.Is(err, err401) still holds. I added the sixth TestResolveSource case you asked for, asserting against the real image.ErrImageTooLarge, and confirmed it fails with the old blunt replace.

isGCPRegistry in syft.go now classifies the pulled reference (pullRef), so the GCP token can't be handed to a proxy host. I left the sidecar's classification as-is since its imageID is already the rewritten reference per your sidecar.go:81 note. Also re-added the MANIFEST_UNKNOWN debug line in resolveSource and cleaned up the ctx shadow in the closure.

New head is 9386972.

pkg/sbomscanner/v1/server.go
```diff
@@ -26,8 +26,74 @@ import (
...
+		logger.L().Debug("got MANIFEST_UNKNOWN, retrying with imageTag",
+			helpers.String("imageTag", imageTag),
+			helpers.String("imageID", imageID))
+		pullRef = imageTag
+		src, err = get(ctx, pullRef, &opts)
+	}
+	if err != nil && strings.Contains(err.Error(), "401 Unauthorized") {
+		unauthorizedErr := err
+		if isGCPRegistry(imageID) {
```
matthyx, line 72: Last cosmetic thing, then I'm done: this still classifies `imageID`, while `syft.go:206` now classifies `pullRef`. In the sidecar the two are the same value except after a `MANIFEST_UNKNOWN` fallback, and both refs are rewritten through the same proxy map at `adapters/v1/sidecar.go:81-82`, so I can't construct a case where it actually misbehaves here.

It's worth aligning anyway:

```go
		if isGCPRegistry(pullRef) {
```

`pullRef` is the reference the retry will actually be issued against, so it's the one whose host should decide whether to attach a GCP token — that's the invariant that made the change necessary in `syft.go`, and it holds here for the same reason even though nothing currently violates it. It also leaves `resolveSource` and the `syft.go` block character-for-character equivalent, which is what turns the shared-package extraction from a merge into a move.

Not worth another round-trip on its own if you'd rather fold it into the follow-up.

matthyx (commented):
## Re-review of `9386972`

All three items from the last round are addressed, and I have nothing blocking left. Verified locally on `9386972` (go1.25.8, Linux): `go build ./...`, `go vet ./pkg/sbomscanner/... ./adapters/...`, `gofmt -l` on the three changed files — all clean; `go test ./pkg/sbomscanner/v1/... ./adapters/v1/... -count=1` passes (`adapters/v1` 63s), and all six `TestResolveSource` subtests pass individually.

- **Error masking — fixed.** `fmt.Errorf("%w (anonymous fallback failed: %v)", …)` in both files. The wrapped message carries both classifications, and since the `switch` tests `ErrImageTooLarge` before the 401 case, an oversized image behind a stale pull secret now lands on `Incomplete` rather than being retried forever as `Unauthorize`. The new `too-large during anonymous retry is not masked by original 401` case pins it, and `assert.ErrorIs(t, err, err401)` in the neighbouring case still passes, so both directions are covered.
- **Token-to-proxy — fixed.** `isGCPRegistry(pullRef)` at `syft.go:206` means the ADC decision now follows the reference actually being pulled, so a `proxyRegistryMap` that remaps `gcr.io` no longer results in a `cloud-platform`-scoped token being handed to the mirror.
- **`MANIFEST_UNKNOWN` log — restored**, and the detached-context comment on the closure with `_ context.Context` reads clearly now.

One cosmetic leftover, called out inline: `server.go:72` still classifies `imageID` while `syft.go:206` classifies `pullRef`. No behavioural consequence I can construct in the sidecar, but it's the last textual difference between two functions that are otherwise identical — worth closing while you're in here, since the shared-package extraction becomes a straight lift once they match. Entirely your call whether that's this PR or the follow-up.

Thanks for taking the whole list, including fixing #351's copy rather than just the sidecar — the in-process adapter was carrying the same defect and would have kept carrying it.

matthyx (approved):

PR #428, opened 2026-08-02
fix: non-blocking Grype DB background updates in GrypeAdapter.Ready (#427)

### LINKED ISSUE #427: Bug: GrypeAdapter.Ready holds write lock during 15-minute DB updates, causing global scan deadlocks, background goroutine leaks, and unexpected os.Exit crashes
# Description

`GrypeAdapter.Ready` handles downloading and updating the Grype vulnerability database. However, the current implementation contains critical concurrency and state management flaws that cause global service deadlocks, background goroutine leaks, and unexpected process crashes in production:

1. **Global Scan Deadlock via Exclusive Lock Retention**: When `GrypeAdapter.Ready(ctx)` detects that the Grype DB needs initialization or updating (on cold start or during the 24-hour update cycle), it acquires an exclusive write lock (`g.mu.Lock()`) and holds it for the entire duration of the DB download (up to the 15-minute timeout). Because every vulnerability scan request (`GrypeAdapter.ScanSBOM`) acquires a read lock (`g.mu.RLock()`), all incoming and in-flight scan requests across the entire service block behind `g.mu.Lock()` for up to 15 minutes, causing widespread request timeouts and API unresponsiveness.
2. **Self-Inflicted Process Termination (`os.Exit(0)`)**: When a DB update attempt encounters a network failure, HTTP 500 from the listing server, or initial download timeout, `Ready()` invokes `os.Exit(0)`. In Kubernetes environments with transient network issues, this forces `kubevuln` into an unrecoverable `CrashLoopBackOff` state even when a valid previously downloaded DB is available or when transient errors can be safely retried.
3. **Background Goroutine Leak on DB Update Timeout**: `grype.LoadVulnerabilityDB` does not accept a `context.Context` for cancellation. When `updateCtx` times out after 15 minutes, `Ready()` returns control to the caller while the spawned `go func()` continues executing `LoadVulnerabilityDB` in the background indefinitely. Subsequent readiness probes spawn additional unmanaged background goroutines, leading to memory bloat and disk I/O contention.

# Environment

- OS: Linux / Kubernetes (All supported distributions)
- Version: `kubevuln` main / latest release

# Steps To Reproduce

1. Start `kubevuln` in a cluster or local environment with a simulated slow network or a listing URL that introduces latency during DB downloads.
2. Trigger `/v1/readiness` or wait for the 24-hour update interval to fire.
3. Send a concurrent POST request to `/v1/scan/container` (`ScanCVE` / `ScanSBOM`).
4. Observe that the scan request blocks completely until the DB update completes or times out (up to 15 minutes).
5. Simulate an HTTP failure during DB download and observe `kubevuln` process termination (`os.Exit(0)`).

# Expected behavior

- `GrypeAdapter.Ready` should perform DB updates asynchronously without blocking active scan requests (`ScanSBOM`).
- Scans should continue using the active, previously loaded Grype DB (`g.store`) while a background worker updates the database (double-buffering / RCU pattern).
- DB update failures should be logged with warnings/errors without invoking `os.Exit(0)`, allowing the service to remain running and fall back gracefully.
- At most one background DB update task should run at a time to prevent goroutine leaks on timeouts.

# Actual Behavior

- `GrypeAdapter.Ready` holds an exclusive write lock (`g.mu.Lock()`) for up to 15 minutes during DB downloads, causing all concurrent scan requests (`g.mu.RLock()`) to deadlock/block.
- Failed DB updates call `os.Exit(0)`, crashing `kubevuln` into `CrashLoopBackOff`.
- Timed-out DB updates leave orphaned `grype.LoadVulnerabilityDB` goroutines running in the background indefinitely.

# Additional context

### Root Cause Analysis

In [`adapters/v1/grype.go`](file:///c:/Users/BHUVAN/OneDrive/Desktop/kubevuln/adapters/v1/grype.go#L168-L260):

```go
if g.dbStatus == nil || now.Sub(g.lastDbUpdate) > 24*time.Hour {
    g.mu.Lock()
    defer g.mu.Unlock() // Lock held across 15-minute select loop!

    updateCtx, cancel := context.WithTimeout(ctx, 15*time.Minute)
    defer cancel()

    resultCh := make(chan updateResult, 1)
    go func() {
        store, dbStatus, err := grype.LoadVulnerabilityDB(g.distCfg, g.installCfg, true)
        resultCh <- updateResult{store: store, dbStatus: dbStatus, err: err}
    }()

    select {
    case result := <-resultCh:
        if result.err != nil {
            tools.DeleteContents(g.installCfg.DBRootDir)
            if !testing.Testing() {
                os.Exit(0) // Process abruptly terminated on update failure
            }
            return false
        }
        // ...
    case <-updateCtx.Done():
        // ...
    }
}
```

Meanwhile, in [`ScanSBOM`](file:///c:/Users/BHUVAN/OneDrive/Desktop/kubevuln/adapters/v1/grype.go#L273-L274):

```go
g.mu.RLock()
defer g.mu.RUnlock()
```

- **Lock Contention**: `g.mu.Lock()` is acquired at line 168 before launching the download goroutine and is held until `Ready()` finishes (lines 238–260). Any concurrent call to `ScanSBOM` blocks at line 273 on `g.mu.RLock()`.
- **Process Exit**: Lines 225 & 256 execute `os.Exit(0)` directly upon update errors or initial download timeouts.
- **Orphan Goroutines**: Line 194 spawns a goroutine that cannot be cancelled when `updateCtx.Done()` triggers at line 239.

- [`adapters/v1/grype.go`](file:///c:/Users/BHUVAN/OneDrive/Desktop/kubevuln/adapters/v1/grype.go#L159-L264): `GrypeAdapter.Ready`, locking mechanism, DB update lifecycle, and error handling.
- [`core/services/scan.go`](file:///c:/Users/BHUVAN/OneDrive/Desktop/kubevuln/core/services/scan.go#L138-L140): `ScanService.Ready` delegation.
- [`controllers/http.go`](file:///c:/Users/BHUVAN/OneDrive/Desktop/kubevuln/controllers/http.go#L83-L90): `HTTPController.Ready` probe endpoint handler.

1. **Decouple DB Updates from Read Locks**: Refactor `GrypeAdapter` to use `singleflight.Group` or an atomic pointer / RCU pattern for `store` and `dbStatus`. Perform database downloading and loading outside of `g.mu.Lock()`, swapping `g.store` atomically once complete.
2. **Refactor Readiness and Error Handling**: Replace `os.Exit(0)` with structured error returns and fallback logic.
3. **Unit Tests**: Add concurrency unit tests in `adapters/v1/grype_test.go` verifying non-blocking scan execution during DB updates and thread safety under `go test -race`.

bhuvan-somisetty:
## Overview

`GrypeAdapter.Ready` previously held an exclusive write lock (`g.mu.Lock()`) for the entire duration of the 15-minute Grype vulnerability database download and load operation. Because vulnerability scans (`ScanSBOM`) acquire a read lock (`g.mu.RLock()`), all active and incoming scan requests across the service blocked behind `g.mu.Lock()` for up to 15 minutes during 24-hour update cycles.

Furthermore, failed update attempts or initial download timeouts invoked `os.Exit(0)`, forcing `kubevuln` into an unrecoverable `CrashLoopBackOff` on transient network errors.

This PR refactors `GrypeAdapter.Ready` to update the database asynchronously in a background goroutine without holding write locks during the download phase, allowing active scans to continue running concurrently without blocking.

## Additional Information

- **Non-Blocking Background Updates**: `GrypeAdapter.Ready` launches database updates asynchronously in a background goroutine without holding `g.mu.Lock()` during network downloads.
- **Atomic Double-Buffering**: Active scans continue using the existing valid database (`g.store`) while new database versions download. Once loaded, `g.store` is swapped atomically under a microsecond write lock.
- **Single-Flight Guarding**: Added `updating` boolean tracking and `updateChan` synchronization to prevent duplicate background update goroutines when probed by readiness checks (`/v1/readiness`).
- **Process Stability**: Removed all `os.Exit(0)` calls. If an update fails or times out while an existing DB is active, `kubevuln` logs a warning and continues serving scans using the active DB while scheduling a delayed retry.

## How to Test

Run the unit test suite for `adapters/v1`:
```bash
go test -v -count=1 -run "Test_grypeAdapter_NonBlockingReady|Test_grypeAdapter_resolveUseDefaultMatchers" ./adapters/v1/...
```

## Related issues/PRs:

* Resolved #427

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved vulnerability database update reliability and recovery from failed or timed-out updates.
  * Existing usable databases are now preserved when updates cannot be completed.
  * Improved startup readiness while database loading is in progress.
  * Scans now clearly reject requests when no valid vulnerability database is available.
  * Prevented duplicate database loads during concurrent operations.

### DIFF
--- a/adapters/v1/grype.go
+++ b/adapters/v1/grype.go
@@ -81,10 +81,10 @@ var _ ports.CVEScanner = (*GrypeAdapter)(nil)
 // NewGrypeAdapter initializes the GrypeAdapter structure
 // DB loading is done via readiness probes
 func NewGrypeAdapter(listingURL string, matchingMode config.CVEMatchingMode, trustedVendors []string) *GrypeAdapter {
+	distCfg := distribution.DefaultConfig()
+	distCfg.LatestURL = listingURL
 	g := &GrypeAdapter{
-		distCfg: distribution.Config{
-			LatestURL: listingURL,
-		},
+		distCfg: distCfg,
 		installCfg: installation.Config{
 			DBRootDir: path.Join(xdg.CacheHome, "grype", "db"),
 		},
@@ -140,10 +140,10 @@ func NewGrypeAdapterFixedDBWithMatchers(matchingMode config.CVEMatchingMode, tru
 	if err != nil {
 		return nil, nil, err
 	}
+	distCfg := distribution.DefaultConfig()
+	distCfg.LatestURL = fmt.Sprintf("http://localhost:%s/databases", port)
 	g := &GrypeAdapter{
-		distCfg: distribution.Config{
-			LatestURL: fmt.Sprintf("http://localhost:%s/databases", port),
-		},
+		distCfg: distCfg,
 		installCfg: installation.Config{
 			DBRootDir: path.Join(xdg.CacheHome, "grype-offline", "db"),
 		},
@@ -262,15 +262,25 @@ func (g *GrypeAdapter) updateDBBackground(ctx context.Context) {
 	select {
 	case <-done:
 	case <-time.After(15 * time.Minute):
+		// grype.LoadVulnerabilityDB takes no context and cannot be cancelled, and the
+		// pinned grype's HTTP client bounds only dial/TLS handshake, not response-body
+		// reads (its withUserAgent post-processor replaces the *http.Client wholesale,
+		// silently dropping any Timeout we set on distCfg). A connection that stalls
+		// mid-download therefore hangs the load goroutine forever, so finishUpdate would
+		// never run: updating would stay true and DBRootDir would never be retried.
+		// Restart here - the one case where this loses non-blocking behaviour is a
+		// download that is genuinely stuck, not merely slow, which is the same condition
+		// that used to trigger os.Exit(0) unconditionally before this PR.
 		g.mu.RLock()
 		existingVersion := g.dbVersionLocked()
 		g.mu.RUnlock()
 		if hasExistingDB {
-			logger.L().Ctx(ctx).Warning("grype DB update taking longer than 15 minutes, continuing with existing DB",
+			logger.L().Ctx(ctx).Error("grype DB update stuck past 15 minutes with an uncancellable load in flight, restarting to recover",
 				helpers.String("existingDBVersion", existingVersion))
 		} else {
-			logger.L().Ctx(ctx).Error("grype DB initial download taking longer than 15 minutes")
+			logger.L().Ctx(ctx).Error("grype DB initial download stuck past 15 minutes with an uncancellable load in flight, restarting to recover")
 		}
+		os.Exit(0)
 	}
 }
 
@@ -279,7 +289,8 @@ func (g *GrypeAdapter) updateDBBackground(ctx context.Context) {
 // gave up on it. It owns every mutation of shared state for this update - installing a
 // successful result, scheduling the next attempt, and releasing the single-flight guard -
 // so nothing else can delete DBRootDir or start a second update while this one is still
-// writing to it.
+// writing to it. State mutation happens under g.mu, but the actual I/O (closing providers,
+// deleting DBRootDir) happens after unlocking, so scans only block for the pointer swap.
 func (g *GrypeAdapter) finishUpdate(ctx context.Context, hasExistingDB bool, store vulnerability.Provider, dbStatus *vulnerability.ProviderStatus, err error) {
 	now := time.Now()
 
@@ -290,17 +301,21 @@ func (g *GrypeAdapter) finishUpdate(ctx context.Context, hasExistingDB bool, sto
 		err = dbStatus.Error
 	}
 
-	g.mu.Lock()
-	defer g.mu.Unlock()
-	defer func() {
+	releaseGuard := func() {
 		g.updating = false
 		if g.updateChan != nil {
 			close(g.updateChan)
 			g.updateChan = nil
 		}
-	}()
+	}
 
 	if err != nil {
+		g.mu.Lock()
+		// Schedule retry in 5 minutes on update failure
+		g.nextUpdateAttempt = now.Add(5 * time.Minute)
+		releaseGuard()
+		g.mu.Unlock()
+
 		logger.L().Ctx(ctx).Error("failed to update grype DB", helpers.Error(err))
 		if store != nil {
 			if closeErr := store.Close(); closeErr != nil {
@@ -309,20 +324,21 @@ func (g *GrypeAdapter) finishUpdate(ctx context.Context, hasExistingDB bool, sto
 		}
 		if !hasExistingDB {
 			if delErr := tools.DeleteContents(g.installCfg.DBRootDir); delErr != nil {
-				logger.L().Debug("cleaned up cache", helpers.Error(delErr),
+				logger.L().Ctx(ctx).Warning("failed to clean up grype DB cache", helpers.Error(delErr),
 					helpers.String("DBRootDir", g.installCfg.DBRootDir))
 			}
 		}
-		// Schedule retry in 5 minutes on update failure
-		g.nextUpdateAttempt = now.Add(5 * time.Minute)
 		return
 	}
 
+	g.mu.Lock()
 	oldStore := g.store
 	g.store = store
 	g.dbStatus = dbStatus
 	// Schedule the next routine refresh in 24 hours
 	g.nextUpdateAttempt = now.Add(24 * time.Hour)
+	releaseGuard()
+	g.mu.Unlock()
 
 	if oldStore != nil && oldStore != store {
 		if closeErr := oldStore.Close(); closeErr != nil {
--- a/adapters/v1/grype.go
+++ b/adapters/v1/grype.go
@@ -268,19 +268,22 @@ func (g *GrypeAdapter) updateDBBackground(ctx context.Context) {
 		// silently dropping any Timeout we set on distCfg). A connection that stalls
 		// mid-download therefore hangs the load goroutine forever, so finishUpdate would
 		// never run: updating would stay true and DBRootDir would never be retried.
-		// Restart here - the one case where this loses non-blocking behaviour is a
-		// download that is genuinely stuck, not merely slow, which is the same condition
-		// that used to trigger os.Exit(0) unconditionally before this PR.
+		//
+		// On a cold start there is nothing to lose, so restart - the one case where this
+		// loses non-blocking behaviour is a download that is genuinely stuck, not merely
+		// slow, which is the same condition that used to trigger os.Exit(0) unconditionally
+		// before this PR. With an existing DB, though, the pod is healthy and serving: keep
+		// it that way instead of dropping in-flight scans, and let finishUpdate install the
+		// stuck load whenever (if ever) it returns.
+		if !hasExistingDB {
+			logger.L().Ctx(ctx).Error("grype DB initial download stuck past 15 minutes with an uncancellable load in flight, restarting to recover")
+			os.Exit(0)
+		}
 		g.mu.RLock()
 		existingVersion := g.dbVersionLocked()
 		g.mu.RUnlock()
-		if hasExistingDB {
-			logger.L().Ctx(ctx).Error("grype DB update stuck past 15 minutes with an uncancellable load in flight, restarting to recover",
-				helpers.String("existingDBVersion", existingVersion))
-		} else {
-			logger.L().Ctx(ctx).Error("grype DB initial download stuck past 15 minutes with an uncancellable load in flight, restarting to recover")
-		}
-		os.Exit(0)
+		logger.L().Ctx(ctx).Error("grype DB update stuck past 15 minutes with an uncancellable load in flight, continuing with existing DB",
+			helpers.String("existingDBVersion", existingVersion))
 	}
 }
 
@@ -297,14 +300,10 @@ func (g *GrypeAdapter) updateDBBackground(ctx context.Context) {
 func (g *GrypeAdapter) finishUpdate(ctx context.Context, hasExistingDB bool, store vulnerability.Provider, dbStatus *vulnerability.ProviderStatus, err error) {
 	now := time.Now()
 
-	// A successful load whose reported status still carries an error is not usable: treat
-	// it the same as a load failure so Ready() keeps reporting not-ready and a retry gets
-	// scheduled, instead of wedging the pod NotReady for 24h with nothing to recover it.
-	if err == nil && dbStatus != nil && dbStatus.Error != nil {
-		err = dbStatus.Error
-	}
-
-	releaseGuard := func() {
+	// Released last via defer, after any I/O below, no matter which branch runs or how it
+	// returns: a leaked updating=true would permanently wedge future updates, since nothing
+	// else ever clears it, so this must not depend on every branch remembering to call it.
+	defer func() {
 		g.mu.Lock()
 		g.updating = false
 		if g.updateChan != nil {
@@ -312,6 +311,13 @@ func (g *GrypeAdapter) finishUpdate(ctx context.Context, hasExistingDB bool, sto
 			g.updateChan = nil
 		}
 		g.mu.Unlock()
+	}()
+
+	// A successful load whose reported status still carries an error is not usable: treat
+	// it the same as a load failure so Ready() keeps reporting not-ready and a retry gets
+	// scheduled, instead of wedging the pod NotReady for 24h with nothing to recover it.
+	if err == nil && dbStatus != nil && dbStatus.Error != nil {
+		err = dbStatus.Error
 	}
 
 	if err != nil {
@@ -332,7 +338,6 @@ func (g *GrypeAdapter) finishUpdate(ctx context.Context, hasExistingDB bool, sto
 					helpers.String("DBRootDir", g.installCfg.DBRootDir))
 			}
 		}
-		releaseGuard()
 		return
 	}
 
@@ -343,8 +348,6 @@ func (g *GrypeAdapter) finishUpdate(ctx context.Context, hasExistingDB bool, sto
 	// Schedule the next routine refresh in 24 hours
 	g.nextUpdateAttempt = now.Add(24 * time.Hour)
 	g.mu.Unlock()
-
-	defer releaseGuard()
 
 	if oldStore != nil && oldStore != store {
 		if closeErr := oldStore.Close(); closeErr != nil {
--- a/adapters/v1/grype.go
+++ b/adapters/v1/grype.go
@@ -50,6 +50,12 @@ const (
 	VendorTrustedMatchMetadataKey = "kubescape.io/vendor-trusted-match"
 )
 
+type loadDBFunc func(distCfg distribution.Config, installCfg installation.Config) (vulnerability.Provider, *vulnerability.ProviderStatus, error)
+
+func defaultLoadDB(distCfg distribution.Config, installCfg installation.Config) (vulnerability.Provider, *vulnerability.ProviderStatus, error) {
+	return grype.LoadVulnerabilityDB(distCfg, installCfg, true)
+}
+
 // GrypeAdapter implements CVEScanner from ports using Grype's API
 type GrypeAdapter struct {
 	lastDbUpdate   time.Time
@@ -62,6 +68,7 @@ type GrypeAdapter struct {
 	trustedVendors map[distro.Type]bool
 	updating       bool
 	updateChan     chan struct{}
+	loadDB         loadDBFunc
 }
 
 var _ ports.CVEScanner = (*GrypeAdapter)(nil)
@@ -78,6 +85,7 @@ func NewGrypeAdapter(listingURL string, matchingMode config.CVEMatchingMode, tru
 		},
 		matchingMode:   matchingMode,
 		trustedVendors: buildTrustedVendorSet(trustedVendors),
+		loadDB:         defaultLoadDB,
 	}
 	return g
 }
@@ -136,6 +144,7 @@ func NewGrypeAdapterFixedDBWithMatchers(matchingMode config.CVEMatchingMode, tru
 		},
 		matchingMode:   matchingMode,
 		trustedVendors: buildTrustedVendorSet(trustedVendors),
+		loadDB:         defaultLoadDB,
 	}
 	return g, terminate, nil
 }
@@ -227,6 +236,11 @@ func (g *GrypeAdapter) updateDBBackground(ctx context.Context) {
 	}
 	resultCh := make(chan updateResult, 1)
 
+	loadFn := g.loadDB
+	if loadFn == nil {
+		loadFn = defaultLoadDB
+	}
+
 	go func() {
 		if logger.L().GetLevel() == "debug" {
 			if distClient, err := distribution.NewClient(g.distCfg); err == nil {
@@ -242,25 +256,24 @@ func (g *GrypeAdapter) updateDBBackground(ctx context.Context) {
 				helpers.Error(err),
 				helpers.String("dbRootDir", g.installCfg.DBRootDir))
 		}
-		store, dbStatus, err := grype.LoadVulnerabilityDB(g.distCfg, g.installCfg, true)
+		store, dbStatus, err := loadFn(g.distCfg, g.installCfg)
 		resultCh <- updateResult{store: store, dbStatus: dbStatus, err: err}
 	}()
 
-	now := time.Now()
 	select {
 	case result := <-resultCh:
+		now := time.Now()
 		if result.err != nil {
 			logger.L().Ctx(ctx).Error("failed to update grype DB", helpers.Error(result.err))
+			g.mu.Lock()
 			if !hasExistingDB {
 				err := tools.DeleteContents(g.installCfg.DBRootDir)
 				logger.L().Debug("cleaned up cache", helpers.Error(err),
 					helpers.String("DBRootDir", g.installCfg.DBRootDir))
-			} else {
-				g.mu.Lock()
-				// Schedule retry in 5 minutes if update failed but existing DB is available
-				g.lastDbUpdate = now.Add(-24*time.Hour + 5*time.Minute)
-				g.mu.Unlock()
 			}
+			// Schedule retry in 5 minutes on update failure
+			g.lastDbUpdate = now.Add(-24*time.Hour + 5*time.Minute)
+			g.mu.Unlock()
 			return
 		}
 		g.mu.Lock()
@@ -277,6 +290,15 @@ func (g *GrypeAdapter) updateDBBackground(ctx context.Context) {
 		}
 		logger.L().Info("grype DB updated")
 	case <-updateCtx.Done():
+		// Drain resultCh asynchronously in case LoadVulnerabilityDB completes after timeout, preventing provider leak
+		go func() {
+			res := <-resultCh
+			if res.store != nil {
+				_ = res.store.Close()
+			}
+		}()
+
+		now := time.Now()
 		g.mu.Lock()
 		if hasExistingDB {
 			logger.L().Ctx(ctx).Warning("grype DB update timed out after 15 minutes, continuing with existing DB",
@@ -288,6 +310,8 @@ func (g *GrypeAdapter) updateDBBackground(ctx context.Context) {
 			err := tools.DeleteContents(g.installCfg.DBRootDir)
 			logger.L().Debug("cleaned up cache after timeout", helpers.Error(err),
 				helpers.String("DBRootDir", g.installCfg.DBRootDir))
+			// Back-date to retry in 5 minutes
+			g.lastDbUpdate = now.Add(-24*time.Hour + 5*time.Minute)
 		}
 		g.mu.Unlock()
 	}
--- a/adapters/v1/grype.go
+++ b/adapters/v1/grype.go
@@ -289,8 +289,11 @@ func (g *GrypeAdapter) updateDBBackground(ctx context.Context) {
 // gave up on it. It owns every mutation of shared state for this update - installing a
 // successful result, scheduling the next attempt, and releasing the single-flight guard -
 // so nothing else can delete DBRootDir or start a second update while this one is still
-// writing to it. State mutation happens under g.mu, but the actual I/O (closing providers,
-// deleting DBRootDir) happens after unlocking, so scans only block for the pointer swap.
+// writing to it. Installing the new state happens under g.mu, but the actual I/O (closing
+// providers, deleting DBRootDir) happens after unlocking, so scans only block for the pointer
+// swap. The single-flight guard is released only after that I/O completes: Ready() unblocks
+// cold-start callers as soon as updateChan closes, so closing it any earlier would let a
+// caller observe g.store before an unusable one has actually been Close()'d.
 func (g *GrypeAdapter) finishUpdate(ctx context.Context, hasExistingDB bool, store vulnerability.Provider, dbStatus *vulnerability.ProviderStatus, err error) {
 	now := time.Now()
 
@@ -302,18 +305,19 @@ func (g *GrypeAdapter) finishUpdate(ctx context.Context, hasExistingDB bool, sto
 	}
 
 	releaseGuard := func() {
+		g.mu.Lock()
 		g.updating = false
 		if g.updateChan != nil {
 			close(g.updateChan)
 			g.updateChan = nil
 		}
+		g.mu.Unlock()
 	}
 
 	if err != nil {
 		g.mu.Lock()
 		// Schedule retry in 5 minutes on update failure
 		g.nextUpdateAttempt = now.Add(5 * time.Minute)
-		releaseGuard()
 		g.mu.Unlock()
 
 		logger.L().Ctx(ctx).Error("failed to update grype DB", helpers.Error(err))
@@ -328,6 +332,7 @@ func (g *GrypeAdapter) finishUpdate(ctx context.Context, hasExistingDB bool, sto
 					helpers.String("DBRootDir", g.installCfg.DBRootDir))
 			}
 		}
+		releaseGuard()
 		return
 	}
 
@@ -337,8 +342,9 @@ func (g *GrypeAdapter) finishUpdate(ctx context.Context, hasExistingDB bool, sto
 	g.dbStatus = dbStatus
 	// Schedule the next routine refresh in 24 hours
 	g.nextUpdateAttempt = now.Add(24 * time.Hour)
-	releaseGuard()
 	g.mu.Unlock()
+
+	defer releaseGuard()
 
 	if oldStore != nil && oldStore != store {
 		if closeErr := oldStore.Close(); closeErr != nil {
--- a/adapters/v1/grype.go
+++ b/adapters/v1/grype.go
@@ -58,17 +58,22 @@ func defaultLoadDB(distCfg distribution.Config, installCfg installation.Config) 
 
 // GrypeAdapter implements CVEScanner from ports using Grype's API
 type GrypeAdapter struct {
-	lastDbUpdate   time.Time
-	dbStatus       *vulnerability.ProviderStatus
-	store          vulnerability.Provider
-	distCfg        distribution.Config
-	installCfg     installation.Config
-	mu             sync.RWMutex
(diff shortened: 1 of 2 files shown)


bhuvan-somisetty:
Hey @matthyx @slashben @kooomix - raised this PR to fix #427. Refactored GrypeAdapter.Ready to update the Grype DB asynchronously in the background so it no longer holds the write lock during 15-minute downloads and block active scans. Also removed the os.Exit(0) calls on update errors/timeouts to prevent crashloops, added single-flighting to prevent duplicate update threads from readiness probes, and added unit tests. Appreciate a review when you have a moment!

bhuvan-somisetty:
Pushed an update addressing all review points - added a pluggable loadDB function to GrypeAdapter so unit tests can stub database loading without network/disk access, added an async drain for resultCh on timeout to close any late-loaded providers, and updated retry timestamp calculations so cold start failures delay retries appropriately.

adapters/v1/grype.go
```diff
@@ -155,112 +165,156 @@ func (g *GrypeAdapter) DBVersion(context.Context) string {
...
 
-// Ready returns the status of the vulnerabilities DB
+// Ready returns the status of the vulnerabilities DB, triggering non-blocking background updates as needed.
 func (g *GrypeAdapter) Ready(ctx context.Context) bool {
-	// DB update is in progress
-	if !g.mu.TryRLock() {
-		return false
+	g.mu.RLock()
+	hasValidDB := g.store != nil && g.dbStatus != nil && g.dbStatus.Error == nil
+	needsUpdate := g.dbStatus == nil || time.Since(g.lastDbUpdate) > 24*time.Hour
```
matthyx: **Blocker — cold-start failure produces an unthrottled DB download storm.**

`needsUpdate` short-circuits on `g.dbStatus == nil`. On a cold start that has never successfully loaded, `dbStatus` stays `nil` through *both* failure paths (`grype.go:274` and `grype.go:313`), so the `lastDbUpdate` back-dating in those branches is never consulted — the first clause is already `true`. The "Schedule retry in 5 minutes" / "Back-date to retry in 5 minutes" comments describe behaviour the code does not have, in exactly the situation they were written for.

Reproduced locally against this branch:

```go
g := &GrypeAdapter{
    loadDB: func(distribution.Config, installation.Config) (vulnerability.Provider, *vulnerability.ProviderStatus, error) {
        atomic.AddInt32(&calls, 1)
        return nil, nil, errors.New("network down")
    },
}
for i := 0; i < 5; i++ { g.Ready(ctx); time.Sleep(20 * time.Millisecond) }
// loadDB invocations across 5 probes: 5  (expected 1 if the 5-minute back-off worked)
```

In production `Ready` is driven by the Kubernetes readiness probe on `/v1/readiness`, so a cold-start outage means one full Grype DB download attempt **per probe interval, indefinitely** — instead of the crash-and-back-off that `os.Exit(0)` used to give. Beyond the wasted bandwidth this is a good way to get rate-limited or blocked by the DB host.

Suggested fix: gate the launch on a dedicated attempt timestamp rather than on `dbStatus == nil` / `lastDbUpdate`, e.g. add a `nextUpdateAttempt time.Time` field, set it on every terminal outcome (success → `now.Add(24h)`, failure → `now.Add(5m)`, timeout → `now.Add(1h)`), and make `needsUpdate` simply `time.Now().After(g.nextUpdateAttempt)`. That also removes the need to back-date `lastDbUpdate`, which currently doubles as both "when the DB was last refreshed" and "when to retry" — two things that want to diverge.

adapters/v1/grype.go
```diff
@@ -155,112 +165,156 @@ func (g *GrypeAdapter) DBVersion(context.Context) string {
...
-
-		// Track if we have an existing DB to fall back on
-		hasExistingDB := g.dbStatus != nil
-
-		// Run the DB update in a goroutine to respect the timeout
-		type updateResult struct {
-			store    vulnerability.Provider
-			dbStatus *vulnerability.ProviderStatus
-			err      error
+		g.updating = false
```
matthyx: **Blocker — the single-flight guard is released while the download is still running.**

This `defer` clears `updating` and closes `updateChan` when `updateDBBackground` *returns*. On the `<-updateCtx.Done()` path that return happens at the 15-minute timeout, but the inner goroutine at `grype.go:245-261` is still executing `loadFn` → `grype.LoadVulnerabilityDB`, which takes no context and cannot be cancelled (your own comment on the drain goroutine says as much).

So after a timeout: `updating == false`, and — because of the missing back-off in the comment on `Ready` above — the very next readiness probe starts a **second** `LoadVulnerabilityDB` writing into the same `g.installCfg.DBRootDir` while the first is still writing there. On a persistently slow link these accumulate without bound: N concurrent downloaders, N goroutines, N sets of partial writes into one directory. `os.Exit(0)` previously made this impossible.

The flag needs to track the *worker*, not the waiter. Either clear `updating` / close `updateChan` from the worker goroutine (or from the existing post-timeout drain goroutine) once `loadFn` has actually returned, or hold a separate dedicated update mutex for the whole worker lifetime so a second `LoadVulnerabilityDB` can never overlap the first.

adapters/v1/grype.go
```diff
@@ -155,112 +165,156 @@ func (g *GrypeAdapter) DBVersion(context.Context) string {
...
+		now := time.Now()
+		g.mu.Lock()
+		if hasExistingDB {
+			logger.L().Ctx(ctx).Warning("grype DB update timed out after 15 minutes, continuing with existing DB",
+				helpers.String("existingDBVersion", g.dbVersionLocked()))
+			// Delay retry for 1 hour on timeout
+			g.lastDbUpdate = now.Add(-24*time.Hour + 1*time.Hour)
+		} else {
+			logger.L().Ctx(ctx).Error("grype DB initial download timed out after 15 minutes")
+			err := tools.DeleteContents(g.installCfg.DBRootDir)
```
matthyx: **Blocker — this wipes `DBRootDir` out from under a download that is still writing to it.**

We only reach here on `updateCtx.Done()`, which means `loadFn` has *not* returned — the worker at `grype.go:245-261` is still inside `grype.LoadVulnerabilityDB`, actively writing into `g.installCfg.DBRootDir`. `tools.DeleteContents` then `os.RemoveAll`s every entry in that directory (`internal/tools/tools.go:94`).

On `main` this was safe only because `os.Exit(0)` followed immediately, so no code ever observed the result. Now the process survives, and the outcome is a half-installed DB on disk plus a drain goroutine at `grype.go:293-298` that `Close()`s a provider backed by files that were just deleted. That corrupted cache is then what the next attempt starts from.

There is no safe moment to delete while the writer is live, so either drop the `DeleteContents` from the timeout branch entirely, or move it into the drain goroutine so it runs only after the worker has actually returned. (The `result.err != nil` branch at `grype.go:270` is fine — the worker has provably finished there, since we received from `resultCh`.)

adapters/v1/grype.go
```diff
@@ -155,112 +165,156 @@ func (g *GrypeAdapter) DBVersion(context.Context) string {
...
+			// Schedule retry in 5 minutes on update failure
+			g.lastDbUpdate = now.Add(-24*time.Hour + 5*time.Minute)
+			g.mu.Unlock()
+			return
+		}
+		g.mu.Lock()
+		oldStore := g.store
+		g.store = result.store
+		g.dbStatus = result.dbStatus
+		g.lastDbUpdate = now
```
matthyx: **Major — no self-heal left when the load succeeds but the status carries an error.**

If `loadFn` returns `err == nil` but `result.dbStatus.Error != nil`, we take this branch: `g.store` and `g.dbStatus` are installed and `g.lastDbUpdate = now`. From then on `Ready` returns `false` (the final check tests `g.dbStatus.Error == nil`) while `needsUpdate` is `false` for a full 24 hours — no retry, and with `os.Exit(0)` gone, no restart either. The pod sits permanently `NotReady` and `ScanSBOM` happily proceeds, since its guard at `grype.go:330` only checks `g.dbStatus == nil` and not `.Error`.

The old code had the same 24-hour wedge in principle, but the pod would be restarted by the readiness failure and re-enter the `os.Exit`-guarded path. Now nothing recovers it.

Suggest treating a non-nil `result.dbStatus.Error` as a failure: take the failure branch (short retry interval, don't set `lastDbUpdate = now`), and consider tightening the `ScanSBOM` guard to `g.store == nil || g.dbStatus == nil || g.dbStatus.Error != nil` so a known-bad DB returns `ErrInitVulnDB` rather than silently producing results.

adapters/v1/grype_test.go
```diff
@@ -121,3 +124,64 @@ func Test_grypeAdapter_resolveUseDefaultMatchers(t *testing.T) {
...
+type mockProvider struct {
+	vulnerability.Provider
+}
+
+func (m *mockProvider) Close() error { return nil }
+
+func Test_grypeAdapter_NonBlockingReady(t *testing.T) {
+	ctx := context.Background()
+	mockStore := &mockProvider{}
+	blockLoad := make(chan struct{})
```
matthyx, line 174: `gofmt -l ./adapters/` flags this file — trailing whitespace on this line and two trailing blank lines at EOF. It is clean on `main`, so this PR introduces it. `gofmt -w adapters/v1/grype_test.go` fixes both.

adapters/v1/grype_test.go
```diff
@@ -121,3 +124,64 @@ func Test_grypeAdapter_resolveUseDefaultMatchers(t *testing.T) {
...
+
+	// Verify DBVersion can acquire read locks concurrently while background update is active
+	version := g.DBVersion(ctx)
+	assert.Equal(t, "test-checksum", version)
+
+	// Verify background update is active under RLock
+	g.mu.RLock()
+	isUpdating := g.updating
+	g.mu.RUnlock()
+	assert.True(t, isUpdating, "background update should be active")
```
matthyx, line 217: **Test coverage is thin for what this PR changes.** `Test_grypeAdapter_NonBlockingReady` only covers the happy path ("`Ready` returns quickly when a valid DB already exists"). The paths this PR actually rewrote — load failure, 15-minute timeout, cold start with no DB, and concurrent probes hitting the single-flight guard — have no coverage at all, and each of them has a blocker in it (see the `grype.go` comments).

Two specifics:

- `mockStore` is returned as *both* the pre-existing `g.store` and the new store from `loadDB`, so `oldStore != result.store` never holds and the atomic swap plus `oldStore.Close()` — the central claim of the PR — is never exercised. Return a *distinct* second provider and assert `g.store` points at it afterwards, and that the first one got `Close()`d exactly once.
- The test returns immediately after `close(blockLoad)` without waiting for `updateDBBackground` to finish, leaving a goroutine running past the test boundary. Under `-race` in a full-package run that is a source of cross-test flakiness. Wait on the update completing (e.g. poll `g.updating` under `RLock`, or have the stub signal a done channel) before returning.

Now that `loadDB` is injectable, the failure and cold-start cases are cheap to write — the repro in the `Ready` comment is about ten lines. The timeout path would additionally need the hard-coded `15*time.Minute` lifted into a field.

matthyx (commented):
Reviewed the concurrency rework. The core idea is right and the **double-buffer swap itself is sound**: `ScanSBOM` holds `g.mu.RLock()` for its entire duration, so any scan that was in flight at swap time has finished before `oldStore.Close()` runs after `g.mu.Unlock()`. No use-after-close there.

But removing `os.Exit(0)` without replacing what it was implicitly guaranteeing leaves three blockers on the failure paths. `os.Exit` was doing more than "crash the pod" — it was also (a) the back-off, (b) the single-flight guard against a still-running download, and (c) what made wiping `DBRootDir` safe. All three now have gaps.

### Blockers

1. **Cold-start failure becomes an unthrottled DB download storm.** Both "retry in 5 minutes" back-dates are dead code in the exact paths they were written for. Verified with a local test — 5 readiness probes triggered 5 full `loadDB` invocations. See inline comment on `Ready`.
2. **The single-flight guard is released while the download is still running.** The 15-minute timeout returns from `updateDBBackground` and clears `updating`, but `grype.LoadVulnerabilityDB` is not cancellable and keeps going. Combined with (1), concurrent downloaders accumulate into the same `DBRootDir`.
3. **`tools.DeleteContents(DBRootDir)` on the timeout path wipes the directory out from under that still-running download.** Previously the immediately-following `os.Exit(0)` meant nothing ever observed the wiped directory.

Plus one **major**: with `os.Exit(0)` gone there is no self-heal left for a load that succeeds but returns `dbStatus.Error != nil` — `lastDbUpdate` is set to `now`, so readiness stays `false` for a full 24 h with no retry and no restart. Inline below.

### Non-blocking

- `gofmt -l` flags `adapters/v1/grype_test.go` (it is clean on `main`, so this PR introduces it).
- Test coverage is thin for a PR whose whole point is failure-path lock semantics — see inline comment.
- `loadFn := g.loadDB` (`grype.go:239`) reads a field of the mutex-protected struct outside the mutex. Harmless today since it is construction-only, but it reads as an oversight next to the carefully-locked fields around it; worth a comment or moving `loadDB` out of the guarded set.

Also note the three earlier `coderabbitai` comments look stale against the current revision — `now` is now sampled inside each `select` case, and the post-timeout provider drain exists.

_Reviewed at commit `861f323`; `go build ./...` and `go test -race ./adapters/v1/...` pass locally._

bhuvan-somisetty:
Good catch on all three — you're right that os.Exit(0) was quietly doing a lot of work.

Fixed by moving every state mutation for an update (installing the result, releasing the single-flight guard, cleaning up DBRootDir) into one place that only runs once the load actually returns, not when the 15-minute wait gives up on it. Concretely:

- `nextUpdateAttempt` replaces `lastDbUpdate` and is set on every terminal outcome regardless of `dbStatus`, so a cold start that keeps failing backs off instead of firing on every probe.
- `updating`/`updateChan` now get released by the goroutine that ran the load, so a slow download can't overlap with a second one kicked off by the next probe.
- `DBRootDir` cleanup on failure happens in that same place, so it only runs after the writer is actually done — no more deleting out from under a live download.
- A success with a non-nil `dbStatus.Error` is now treated as a failure (short retry, store not installed), and `ScanSBOM`'s guard checks `dbStatus.Error` too.

Also fixed the gofmt issue and reworked the test — the old/new store swap is now checked with two distinct providers plus close-tracking, and added tests reproducing the failing-cold-start storm, the dbStatus.Error wedge, and single-flight under concurrent Ready() calls. All green locally.

matthyx (commented):
Re-reviewed at `6acaeac`. **All four findings from my previous review are genuinely fixed**, and the fixes are the right ones rather than the minimal ones:

- **Back-off** — `nextUpdateAttempt` replaces the overloaded `lastDbUpdate`, so retry scheduling no longer depends on `dbStatus`. Splitting "when was the DB refreshed" from "when may we try again" is exactly right.
- **Single-flight** — moving every shared-state mutation into `finishUpdate`, invoked by the goroutine that actually ran the load, is a better fix than what I suggested. The guard now tracks the worker rather than the waiter.
- **`DeleteContents` racing a live download** — resolved as a consequence of the above; it can now only run once `loadFn` has returned.
- **`dbStatus.Error` wedge** — folded into the failure path, and `ScanSBOM`'s guard tightened to `store == nil || dbStatus == nil || dbStatus.Error != nil` so a known-bad DB returns `ErrInitVulnDB` instead of silently scanning against it.

The tests are now genuinely load-bearing: distinct old/new providers prove the swap and the exactly-once close, and the cold-start back-off, status-error, and concurrent single-flight paths all have direct coverage. The comments explaining *why* each invariant holds are unusually good.

Verified locally at `6acaeac`: `go build ./...` and `go vet ./adapters/...` clean; `gofmt -l` no longer flags either changed file (the remaining hits are pre-existing on `main`); `go test -race -count=5 -run Test_grypeAdapter ./adapters/v1/` passes, and all four new/updated tests run in ≤0.01s.

### One thing left, and it is the load-bearing assumption of the whole PR

Making `finishUpdate` the sole owner of shared state is correct, but it means **the adapter can only ever recover if `loadFn` returns** — and right now nothing guarantees that. The 15-minute branch was the last remaining escape hatch and it now only logs. Details inline on `grype.go:264`; the fix is small and mostly lives outside this diff.

Two minor items inline as well. Neither blocks.

adapters/v1/grype.go
```diff
@@ -155,112 +170,166 @@ func (g *GrypeAdapter) DBVersion(context.Context) string {
...
+				helpers.Error(err),
+				helpers.String("dbRootDir", g.installCfg.DBRootDir))
+		}
+		store, dbStatus, err := loadFn(g.distCfg, g.installCfg)
+		g.finishUpdate(ctx, hasExistingDB, store, dbStatus, err)
+	}()
+
+	select {
+	case <-done:
+	case <-time.After(15 * time.Minute):
```
matthyx, line 264: **A stuck download is now unrecoverable — this is the one gap the new design depends on closing.**

Making `finishUpdate` the sole owner of shared state is the right call, but it also makes it the *only* place that clears `updating`, advances `nextUpdateAttempt`, and closes `updateChan` — and it only runs when `loadFn` returns. This branch, which used to be the escape hatch, now just logs and lets `updateDBBackground` return with the guard still held.

That would be fine if `grype.LoadVulnerabilityDB` were bounded. It isn't, here:

- Both constructors build `distribution.Config{LatestURL: ...}` as a bare struct literal (`grype.go:85` and `grype.go:145`) rather than starting from `distribution.DefaultConfig()`, so `CheckTimeout` and `UpdateTimeout` are both **zero**.
- `distribution.NewClient` unconditionally applies `withClientTimeout(cfg.CheckTimeout)` / `withClientTimeout(cfg.UpdateTimeout)`, which overwrites the `httpClient.Timeout = 30 * time.Second` that `defaultHTTPClient` had just set — zero means *no* timeout.
- In grype v0.99.1 (the pinned version) it doesn't even matter, because `withUserAgent` runs afterwards and does `*(c) = *newHTTPClientWithDefaultUserAgent(c.Transport, ...)`, replacing the whole `*http.Client` with a fresh one that sets no `Timeout` field at all.
- `cleanhttp`'s transport bounds dial and TLS handshake but not response-body reads, so a connection that establishes and then stalls mid-download hangs indefinitely.

So the reachable outcomes after this branch fires:

- **Cold start** — `updating` stays `true` forever, `nextUpdateAttempt` is never advanced, `updateChan` never closes. `Ready` blocks until its caller's context is cancelled and then returns `false` permanently. No retry, and with `os.Exit(0)` gone, no restart either. On `main` this case restarted the pod.
- **Warm DB** — worse, because it's silent. `Ready` keeps returning `true`, so the pod stays in service and scans keep succeeding, but no update will ever be attempted again for the lifetime of the process. The DB just gets staler behind a single `Warning` log line. On `main` this retried on the next 24h cycle.

Both are strict regressions in recoverability versus `main`, and both are invisible to a readiness probe, which is what makes them worth fixing before merge rather than after.

The clean fix is to bound the load so `finishUpdate` is guaranteed to run: build `distCfg` from `distribution.DefaultConfig()` and override `LatestURL` (picking up the 30s/300s timeouts), or set `CheckTimeout`/`UpdateTimeout` explicitly. Worth asserting it actually takes effect given the `withUserAgent` clobber above — a quick check against a listener that accepts and never writes will tell you. If the pinned grype genuinely can't be bounded from the outside, then this branch needs a real escape hatch instead: `os.Exit(0)` here specifically, reachable only when the load is stuck past 15 minutes, can no longer crashloop the way #427 described — ordinary failures now all route through `finishUpdate` with a 5-minute back-off and never reach this branch.

adapters/v1/grype.go
```diff
@@ -155,112 +170,166 @@ func (g *GrypeAdapter) DBVersion(context.Context) string {
...
+
+	// A successful load whose reported status still carries an error is not usable: treat
+	// it the same as a load failure so Ready() keeps reporting not-ready and a retry gets
+	// scheduled, instead of wedging the pod NotReady for 24h with nothing to recover it.
+	if err == nil && dbStatus != nil && dbStatus.Error != nil {
+		err = dbStatus.Error
+	}
+
+	g.mu.Lock()
+	defer g.mu.Unlock()
```
matthyx: Minor: this now holds the **write** lock across `store.Close()` (`:306`), `tools.DeleteContents()` (`:311`, filesystem I/O) and `oldStore.Close()` (`:328`). The previous revision closed the old store after `g.mu.Unlock()`, and `main` did too.

Every scan blocks for that whole window, which cuts against the PR's own goal of keeping the write-lock hold time down to the pointer swap. It's bounded and far from the 15 minutes this PR is fixing, so not a blocker — but the swap itself is the only part that actually needs the lock. Capturing `oldStore` under the lock and doing the close/delete after unlocking gets you back to a microsecond critical section, at the cost of splitting the deferred cleanup.

(The defer ordering as written is correct, for what it's worth — the `updating`/`updateChan` reset is registered second so it runs before `g.mu.Unlock()`, i.e. still under the lock.)

adapters/v1/grype.go
```diff
@@ -155,112 +170,166 @@ func (g *GrypeAdapter) DBVersion(context.Context) string {
...
-				return true
-			} else {
-				// No existing DB to fall back on, must restart
-				logger.L().Ctx(ctx).Error("grype DB initial download timed out after 15 minutes")
-				err := tools.DeleteContents(g.installCfg.DBRootDir)
-				logger.L().Debug("cleaned up cache after timeout", helpers.Error(err),
+		}
+		if !hasExistingDB {
+			if delErr := tools.DeleteContents(g.installCfg.DBRootDir); delErr != nil {
+				logger.L().Debug("cleaned up cache", helpers.Error(delErr),
```
matthyx: Nit: this logs `"cleaned up cache"` but only fires when the cleanup **failed** (`delErr != nil`), so the one case that produces output is the one the message doesn't describe. Debug level also means it's likely invisible in the deployment where it matters.

Previous revisions logged unconditionally, which read correctly because `helpers.Error(err)` was nil on success. Either restore that, or make this a `Warning` with a message like `"failed to clean up grype DB cache"` — a `DBRootDir` that can't be cleared after a failed cold start is exactly the state that makes the *next* attempt fail too, so it's worth surfacing.

adapters/v1/grype.go
```diff
@@ -155,112 +170,182 @@ func (g *GrypeAdapter) DBVersion(context.Context) string {
...
+			close(g.updateChan)
+			g.updateChan = nil
+		}
+	}
+
+	if err != nil {
+		g.mu.Lock()
+		// Schedule retry in 5 minutes on update failure
+		g.nextUpdateAttempt = now.Add(5 * time.Minute)
+		releaseGuard()
```
matthyx: **Blocker — this ordering makes `Test_grypeAdapter_Ready_treatsStatusErrorAsFailure` fail about half the time.**

`releaseGuard()` closes `updateChan`, and that is precisely what unblocks a cold-start `Ready`. Moving the I/O out of the critical section (which was my ask, and is right) put `store.Close()` at `:321` and `tools.DeleteContents` at `:326` *after* the guard release — so `Ready` can return to its caller while the failed store is still open and `DBRootDir` is still being wiped.

`grype_test.go:274` asserts `badStore.closed == 1` immediately after `Ready` returns, so it loses that race whenever `Ready` wins it. Measured at `364dcf7` on this branch:

| run | failures |
|---|---|
| `-race -count=50`, this test alone | **26 / 50** |
| `-race -count=20`, with the other adapter tests | 9 / 20 |
| `-count=20`, no `-race` | 1 / 20 |

So this lands as a red CI check roughly every other run, not as an occasional flake.

The production impact is mild — the store was never published to `g.store`, so the only real cost is delayed file-handle release, and `nextUpdateAttempt` is five minutes out so a second load can't collide with the in-flight `DeleteContents`. But it does make the doc comment at `:289-293` literally untrue: "nothing else can delete DBRootDir or start a second update while this one is still writing to it" no longer holds once the guard is released before the delete finishes.

Cleanest fix, and it keeps the narrow critical section: in the error path nothing has been published yet, so `store.Close()` and `DeleteContents` need no lock *and* nothing can observe them — do them first, then take the lock exactly once at the end to set `nextUpdateAttempt` and release the guard. Same shape in the success path: capture `oldStore` and swap under the lock, unlock, close, then re-take the lock just to release the guard. Guard released last in both paths restores the invariant and removes the race without re-widening anything.

While you're here: `releaseGuard` is no longer `defer`red, so correctness now depends on every path calling it by hand. A leaked `updating = true` is a permanent wedge in this design — nothing else ever clears it — so an early `return` added later would silently freeze updates for the process lifetime. If the restructure above puts the lock-and-release at the tail of both paths, a single `defer func() { g.mu.Lock(); releaseGuard(); g.mu.Unlock() }()` gets that guarantee back structurally.

adapters/v1/grype.go
```diff
@@ -155,112 +170,182 @@ func (g *GrypeAdapter) DBVersion(context.Context) string {
...
+		g.mu.RLock()
+		existingVersion := g.dbVersionLocked()
+		g.mu.RUnlock()
+		if hasExistingDB {
+			logger.L().Ctx(ctx).Error("grype DB update stuck past 15 minutes with an uncancellable load in flight, restarting to recover",
+				helpers.String("existingDBVersion", existingVersion))
+		} else {
+			logger.L().Ctx(ctx).Error("grype DB initial download stuck past 15 minutes with an uncancellable load in flight, restarting to recover")
+		}
+		os.Exit(0)
```
matthyx: **Worth deciding deliberately: this now also kills a pod that is healthy and serving.**

I suggested this fallback and I still think it's right for a cold start — there's nothing to lose and it matches `main`. The part I'd reconsider is that it fires on the `hasExistingDB` path too.

In that case the pod has a valid DB, `Ready` is returning true, and scans are succeeding; the only thing wrong is that a *background refresh* stalled. On `main` that logged a warning and kept serving. Here it calls `os.Exit(0)`, so in-flight scans are dropped mid-request and the pod comes back needing a full cold-start download before it's ready again. If the stall is a property of the DB host rather than of the pod — connections accepted, bodies stalling — that's every replica with a healthy DB restarting every ~15 minutes, which is the failure mode #427 set out to remove.

Suggest gating the exit on `!hasExistingDB`. Warm-and-stuck then behaves like `main` (keep serving, log at Error), and your existing design already recovers on its own if the stuck load ever returns, since `finishUpdate` will install it whenever that happens. The residual is that a *permanently* stuck load means no further refreshes for that process lifetime — but that's an alert-on-DB-age concern, not a reason to drop live traffic.

Not asserting the current choice is wrong; preferring freshness over availability here is defensible. It just deserves to be an explicit decision rather than a side effect of collapsing both branches, because the two cases have very different blast radii.

adapters/v1/grype_test.go
```diff
@@ -121,3 +127,183 @@ func Test_grypeAdapter_resolveUseDefaultMatchers(t *testing.T) {
...
+			return badStore, &vulnerability.ProviderStatus{Error: statusErr}, nil
+		},
+	}
+
+	require.False(t, g.Ready(ctx))
+
+	g.mu.RLock()
+	defer g.mu.RUnlock()
+	assert.Nil(t, g.store, "a load whose status carries an error must not be installed as the active store")
+	assert.Equal(t, int32(1), atomic.LoadInt32(&badStore.closed), "the unusable store must be closed rather than leaked")
```
matthyx, line 274: This is the assertion that fails ~50% of the time under `-race` — see the `grype.go:316` comment for the cause. Fixing the ordering there fixes this without touching the test, which is the better outcome: the assertion is correct as written, and weakening it to a poll would hide the invariant it is checking.

Same hazard exists latently at `grype_test.go:226` (`oldStore.closed` right after `waitForNotUpdating`): the success path also releases the guard before `oldStore.Close()` at `grype.go:344`, and `waitForNotUpdating` returns as soon as `g.updating` flips. It survived 200 consecutive `-race` runs here — the 5 ms poll granularity almost always hides the window — but it's the same bug, so please apply the ordering fix symmetrically rather than only to the error path.

matthyx (commented):
Re-reviewed at `364dcf7`. All three items from the last round are addressed, but **the lock-scope fix introduced a reproducible test failure** — it is not a rare flake, it fails about half the time under `-race`. Details inline on `grype.go:316`.

**Confirmed fixed:**

- **`distribution.DefaultConfig()`** in both constructors — right hygiene. I verified it is currently *inert* for the timeout purpose in the pinned grype v0.99.1: `CheckTimeout`/`UpdateTimeout` are consumed only at `distribution/client.go:65,70`, and `withUserAgent` runs last in the post-processor chain and replaces the whole `*http.Client`, dropping `Timeout`. Your code comment says exactly that, which is the honest framing — please keep it, so nobody later assumes the timeouts are live and removes the escape hatch.
- **Write-lock scope** — provider closes and `DeleteContents` are out of the critical section, which is what I asked for. The ordering it introduced is the problem, not the idea.
- **Cleanup log** — now a `Warning` with a message that matches the condition.

**One blocker, one thing worth deciding deliberately, one nit:**

1. `grype.go:316` — releasing the single-flight guard before the cleanup I/O makes `Ready` return while the failed store is still open. Measured at `364dcf7`: **26/50** failures standalone with `-race`, 9/20 alongside the other adapter tests, 1/20 without `-race`.
2. `grype.go:283` — `os.Exit(0)` now fires on the *warm* path too, killing a pod that has a valid, serving DB. I suggested this fallback and stand behind it for the cold-start case; the warm case is the part worth a second look.
3. `grype.go:316` (same comment) — `releaseGuard` is no longer deferred.

Verified at `364dcf7`: `go build ./...` and `go vet ./...` clean; `gofmt -l` flags neither changed file (remaining hits are pre-existing on `main`); `Test_grypeAdapter_NonBlockingReady`, `_backsOffAfterFailedColdStart` and `_singleFlightUnderConcurrency` pass 5/5 under `-race` (and `NonBlockingReady` 200/200).

matthyx (commented):
Re-reviewed at `8401335`. **The blocker is fixed and I have no further findings.**

`releaseGuard` now takes `g.mu` itself and is called only after the I/O completes in both paths — explicitly at the end of the error path, via `defer` (registered after the unlock) in the success path. I checked specifically for the hazard that shape can introduce: `releaseGuard` acquires the lock, so it would deadlock if it were ever reached with `g.mu` held, and it isn't — both call sites are after an explicit `g.mu.Unlock()`, and no `defer g.mu.Unlock()` survives in `finishUpdate`. `nextUpdateAttempt` is still published before the I/O while `updating` stays true across it, so nothing can launch a second update during the window. The doc comment at `:289-296` is now accurate rather than aspirational, and it explains *why* the ordering matters, which is the part that will stop someone re-breaking it later.

Verified at `8401335`:

| check | result |
|---|---|
| `Test_grypeAdapter_Ready_treatsStatusErrorAsFailure`, `-race -count=100` | **0 failures** (was 26/50 at `364dcf7`) |
| `Test_grypeAdapter_NonBlockingReady`, `-race -count=200` | 0 failures |
| all four `Ready`/`NonBlockingReady` tests, `-race -count=30` | pass |
| `go test -race ./adapters/... ./core/...` | pass |
| `go build ./...`, `go vet ./...` | clean |
| `gofmt -l` | neither changed file (remaining hits pre-existing on `main`) |

Both of the ordering hazards I flagged last round are closed, including the latent one in the success path — good that you fixed it symmetrically rather than only where the test was red.

### One item still open, and it's yours to call

The `os.Exit(0)` at `grype.go:283` still fires on the `hasExistingDB` path, so a stalled *background refresh* restarts a pod that has a valid DB and is serving scans — where `main` logged and kept serving. My previous comment has the reasoning; gating the exit on `!hasExistingDB` is the one-line version. I'm deliberately not calling it a blocker: preferring guaranteed freshness over availability is a legitimate choice for a vulnerability scanner, and you may well want it. It just shouldn't merge as an unnoticed side effect of collapsing the two branches — a line in the PR description saying which you chose would be enough.

Nothing else from me. The final shape — single-flight owned by the goroutine that ran the load, state published under a pointer-swap-sized lock, I/O outside it, guard released last — is a genuinely clean answer to #427, and the four tests now pin every invariant that makes it work.

matthyx (approved):

bhuvan-somisetty:
Thanks for the thorough back-and-forth on this one, matthyx! Good catch on the release ordering causing that flaky failure — moving the guard release to a single deferred call at the top of finishUpdate also just makes it structurally impossible to leave updating stuck true if someone adds an early return later. And you're right about the warm-DB exit case, gated it to cold-start only so a pod with a healthy DB just logs and keeps serving instead of restarting for no reason. Appreciate the deep dive into the pinned grype client's timeout handling too, that would've been an annoying one to debug in prod.

PR #432, opened 2026-08-03
fix(securityexception): keep under-investigation vulnerabilities visible

### LINKED ISSUE #431: Bug: under_investigation SecurityException suppresses a vulnerability
# Description

A `SecurityException` with the status `under_investigation` incorrectly hides a vulnerability from the scan results.

The vulnerability is removed from `Matches` and moved to `IgnoredMatches`.

The `under_investigation` status means that the vulnerability is still being checked. It should remain visible until it is marked as `not_affected` or `fixed`.

# Environment

OS: `Ubuntu 26.04 LTS`

Version: `kubevuln main at 50254bb`

# Steps To Reproduce

1. Create a vulnerability match for `CVE-2023-44487`.
2. Create a `SecurityException` for the same CVE.
3. Set its status to `under_investigation`.
4. Convert it using `ConvertToVulnerabilityExceptionPolicies`.
5. Apply it using `ApplySecurityExceptions`.
6. Check `Matches` and `IgnoredMatches`.

I reproduced the issue with this test command:

```bash
go test ./adapters/v1 \
  -run '^TestUnderInvestigationSecurityExceptionDoesNotSuppress$' \
  -count=1 -v
```

Before the fix, the test failed:

```text
Matches after ApplySecurityExceptions: []
IgnoredMatches after ApplySecurityExceptions: [CVE-2023-44487]

Error: "[]" should have 1 item, but has 0
--- FAIL: TestUnderInvestigationSecurityExceptionDoesNotSuppress
```

# Expected behavior

The vulnerability should remain in `Matches`.

It should not be moved to `IgnoredMatches` while its status is `under_investigation`.

Only `not_affected` and `fixed` statuses should suppress the vulnerability.

# Actual Behavior

The vulnerability is removed from `Matches` and added to `IgnoredMatches`.

# Additional context

The problem happens because `ConvertToVulnerabilityExceptionPolicies` converts every vulnerability entry into an ignore policy without checking its status.

I prepared a fix that skips `under_investigation` entries in both:

* Namespaced `SecurityException`
* Cluster-scoped `ClusterSecurityException`

I also added a regression test. The relevant tests pass after applying the fix:

```text
ok github.com/kubescape/kubevuln/adapters/v1
```

I found this bug and already prepared and tested the fix. I would like to work on this issue and open the pull request. Please assign it to me if possible.

h30s:
## Overview

This PR fixes a bug where a vulnerability with the `under_investigation` status was incorrectly hidden from scan results.

### Current behavior

`ConvertToVulnerabilityExceptionPolicies` converts an `under_investigation` vulnerability into an ignore policy.

Because of this, the vulnerability is removed from `Matches` and moved to `IgnoredMatches`.

### New behavior

Vulnerabilities with the `under_investigation` status are not converted into ignore policies.

They remain visible in the scan results.

The fix works for both:

- Namespaced `SecurityException`
- Cluster-scoped `ClusterSecurityException`

The existing behavior for `not_affected` and `fixed` vulnerabilities remains unchanged.

## Additional Information

The existing `VulnerabilityExceptionPolicy` model only supports the ignore action. It cannot correctly represent the `under_investigation` status.

For this reason, the conversion now skips these entries instead of turning them into ignore policies.

Regression tests were added for both namespaced and cluster-scoped exceptions. The tests confirm that the vulnerability remains in `Matches` and is not moved to `IgnoredMatches`.

## How to Test

Run the focused regression and related tests:

```bash
go test ./adapters/v1 \
  -run '^(TestUnderInvestigation.*DoesNotSuppress|TestConvertVulnerabilityExceptions|TestApplySecurityExceptions_.*)$' \
  -count=1
```

Result:

```text
ok github.com/kubescape/kubevuln/adapters/v1
```

I also tried to run the complete `adapters/v1` test suite. The existing `Test_grypeAdapter_DBVersion` test could not run because my local user did not have permission to access the Docker socket:

```text
permission denied while trying to connect to /var/run/docker.sock
```

This Docker permission error is unrelated to the changed SecurityException code.

## Related issues/PRs

Resolves #431

- [x] I have commented on the code where needed. No complex code was added.
- [x] This is a small bug fix, and I added regression tests.
- [x] The relevant new and existing unit tests pass locally.
- [ ] The complete package test suite passes locally. It could not finish because Docker socket access was unavailable.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved vulnerability exception handling by suppressing only entries with valid IDs and supported statuses.
  * Prevented entries marked as under investigation, invalid, unknown, or missing required IDs from suppressing vulnerabilities.
  * Applied consistent filtering to both namespace-scoped and cluster-scoped exceptions.

* **Tests**
  * Added coverage for valid, invalid, mixed-status, and incomplete vulnerability exceptions.


adapters/v1/securityexception.go
```diff
@@ -34,6 +34,9 @@ func ConvertToVulnerabilityExceptionPolicies(exceptions []sev1beta1.SecurityExce
 		}
 		namespace := se.Namespace
 		for _, vuln := range se.Spec.Vulnerabilities {
+			if vuln.Status == sev1beta1.VulnerabilityStatusUnderInvestigation {
+				continue
+			}
```
matthyx: **Blocker — this is a denylist on a suppression path; it should be an allowlist.**

The filter skips exactly one literal. Every *other* status value still produces an `Ignore` policy and suppresses the CVE. I verified this against this branch with a throwaway table test:

```text
status=""                    -> policies=1 matches=0 ignored=1
status="affected"            -> policies=1 matches=0 ignored=1
status="Under_Investigation" -> policies=1 matches=0 ignored=1
status="under-investigation" -> policies=1 matches=0 ignored=1
status="typo"                -> policies=1 matches=0 ignored=1
```

Why that matters here rather than being theoretical:

1. **Empty status suppresses.** `VulnerabilityException.Status` (`pkg/securityexception/v1beta1/types.go:84`) is a plain string with no defaulting, and there is no CRD manifest in this repo — the enum/CEL validation is only *proposed* in `docs/security-exception-design.md:350` ("The CRD schema **should** include CEL validation rules ... `vulnerabilities[].status` must be one of ..."). So a manifest that omits `status:` is accepted today and silently hides the finding.
2. **`affected` is missing from the enum.** OpenVEX defines `affected`, but `types.go:43-45` only declares `not_affected`, `fixed`, `under_investigation`. The day `affected` is added, this denylist suppresses it — i.e. #431 recurs in a worse form.

Suggested shape (both loops):

```go
switch vuln.Status {
case sev1beta1.VulnerabilityStatusNotAffected, sev1beta1.VulnerabilityStatusFixed:
	// only these two suppress
default:
	continue
}
```

One thing to decide before doing this, because it is a real behavior change and needs a maintainer call: every existing test in `adapters/v1/securityexception_test.go` builds `VulnerabilityException` **without** `Status` (`grep -c Status` on that file returns 0), so they currently rely on `""` meaning "ignore". Adopting the allowlist means adding `Status: sev1beta1.VulnerabilityStatusNotAffected` to those fixtures.

If you'd rather keep `""` meaning "ignore" for backward compatibility, that's defensible — but then please make it explicit and tested (a case asserting empty status still suppresses), instead of leaving it as an accident of the denylist.

adapters/v1/securityexception.go
```diff
@@ -51,6 +54,9 @@ func ConvertToVulnerabilityExceptionPolicies(exceptions []sev1beta1.SecurityExce
 			continue
 		}
 		for _, vuln := range cse.Spec.Vulnerabilities {
+			if vuln.Status == sev1beta1.VulnerabilityStatusUnderInvestigation {
+				continue
+			}
```
matthyx: **Should fix — the per-entry filter is now copy-pasted across both loops.**

This block and the one at line 37 are byte-identical, and so is the `strings.TrimSpace(vuln.Vulnerability.ID) == ""` guard right below each. That's the exact drift risk that produced #431: a filter rule that has to be remembered in two places.

Worth collapsing into one predicate so the namespaced and cluster-scoped paths cannot diverge:

```go
// shouldSuppress reports whether a VEX entry may be converted into an ignore policy.
func shouldSuppress(vuln sev1beta1.VulnerabilityException) bool {
	if strings.TrimSpace(vuln.Vulnerability.ID) == "" {
		return false
	}
	switch vuln.Status {
	case sev1beta1.VulnerabilityStatusNotAffected, sev1beta1.VulnerabilityStatusFixed:
		return true
	default:
		return false
	}
}
```

Then both loops become `if !shouldSuppress(vuln) { continue }`. This also means the allowlist fix from the other comment only has to land once.

adapters/v1/under_investigation_regression_test.go
```diff
@@ -0,0 +1,84 @@
...
+				},
+			},
+		},
+	}
+
+	ApplySecurityExceptions(doc, domain.CVEExceptions(policies))
+
+	require.Len(t, doc.Matches, 1)
+	require.Empty(t, doc.IgnoredMatches)
+}
```
matthyx: **Test coverage gap — the interesting case isn't covered.**

Both tests use an exception containing a *single* `under_investigation` entry, so they only prove "zero policies produced → nothing suppressed". They'd still pass if the conversion were broken in a way that dropped the whole exception.

The case that actually pins the fix is a **mixed** exception — one `under_investigation` entry plus one `not_affected` entry — asserting the first stays in `Matches` while the second lands in `IgnoredMatches`. I ran that against this branch and the behavior is correct (`CVE-1` remains, `CVE-2` ignored), it's just untested, so nothing stops a future refactor from regressing it.

Also worth adding once the allowlist question above is settled: a case pinning what an empty/unrecognized status does. That's currently the highest-risk untested path.

Smaller points, take or leave:

- **File placement.** `adapters/v1/securityexception_test.go` already holds `TestConvertVulnerabilityExceptions`, `TestConvertSkipsExpired`, `TestConvertSkipsEmptyAndWhitespaceIDs` — this is the same unit and the same "conversion skips X" family. A separate `under_investigation_regression_test.go` splits the suite by bug report rather than by unit.
- **Duplication.** The two tests differ only in namespaced vs cluster-scoped; a table with a `namespaced bool` (or two `ConvertToVulnerabilityExceptionPolicies` call sites in one loop) removes ~40 duplicated lines.
- **Assertion library.** This file uses `require`; the neighbouring `securityexception_test.go` uses `assert` throughout. Worth matching the file you merge into.

For the record, the change itself is correct and I verified the full package is green on this branch:

```text
go test ./adapters/v1/... -count=1 -skip 'Test_grypeAdapter_DBVersion|TestNewGrypeAdapter'
ok  github.com/kubescape/kubevuln/adapters/v1  56.419s
```

(The two skipped tests are the Docker-socket ones you called out in the PR description — unrelated to this change.)

matthyx (commented):
Reviewed the change against the branch (built, ran the package tests, and probed the conversion path with throwaway tests).

**The fix is correct for the reported bug** — `under_investigation` entries no longer become ignore policies, in both the namespaced and cluster-scoped paths, and the package suite is green (`ok github.com/kubescape/kubevuln/adapters/v1`, skipping only the two Docker-socket tests you already flagged).

**One blocker before merge:** the guard is a denylist on a security-suppression path. It skips the single literal `under_investigation`; every other value — including the empty string, `affected`, and any case/format variant — still produces an `Ignore` policy and hides the CVE. Since `Status` has no defaulting, there's no CRD in this repo, and the enum validation is still only a proposal in `docs/security-exception-design.md:350`, a manifest that simply omits `status:` suppresses the finding today. An allowlist (`not_affected`, `fixed` suppress; everything else reports) makes this fail closed and is immune to the enum growing an `affected` value later. Details and the behavioral caveat — existing fixtures rely on empty status meaning "ignore", so they'd need updating — are in the inline comment.

Two non-blocking items also inline: the filter is now duplicated verbatim across both loops (worth extracting so the two paths can't drift), and the tests only cover the single-entry case, not the mixed `under_investigation` + `not_affected` exception that actually pins the behavior.

Thanks for the clear reproduction in #431 and for adding regression tests with the fix.

h30s:
@matthyx I've addressed all the review comments, ptal

adapters/v1/securityexception_test.go
```diff
@@ -241,24 +244,41 @@ func TestApplySecurityExceptions_ExpiredOnFix(t *testing.T) {
...
+// (empty, under_investigation, unrecognised values, blank IDs) does not.
+func TestConvertShouldSuppressAllowlist(t *testing.T) {
+	tests := []struct {
+		name        string
+		id          string
+		status      sev1beta1.VulnerabilityStatus
+		wantPolicies int
+	}{
+		{name: "empty ID is skipped", id: "", status: sev1beta1.VulnerabilityStatusNotAffected, wantPolicies: 0},
+		{name: "whitespace ID is skipped", id: "   ", status: sev1beta1.VulnerabilityStatusNotAffected, wantPolicies: 0},
```
matthyx, line 259: **Coverage regression from the consolidation — the ID-trimming guard was dropped.**

The `TestConvertSkipsEmptyAndWhitespaceIDs` that this table replaced had a third case the table doesn't carry over:

```go
{Vulnerability: sev1beta1.VulnerabilityRef{ID: "  CVE-2024-1234  "}},
...
assert.Equal(t, "CVE-2024-1234", policies[0].VulnerabilityPolicies[0].Name)
```

The new table asserts only policy *counts*, so a padded-but-valid ID is never exercised, and the `strings.TrimSpace(vuln.Vulnerability.ID)` inside `buildPolicy` is now completely unguarded. That normalization was itself a bug fix — commit 6784e9b, "fix(exceptions): normalize vulnerability ID casing and trim whitespace" — so this is a live regression guard being removed.

I confirmed it by mutation rather than by reading: deleting the trim from `buildPolicy`

```go
-  {Name: strings.TrimSpace(vuln.Vulnerability.ID)},
+  {Name: vuln.Vulnerability.ID},
```

leaves the whole package green:

```text
ok  github.com/kubescape/kubevuln/adapters/v1  1.733s
```

Cheapest fix is a `wantName` field on the table:

```go
tests := []struct {
	name         string
	id           string
	status       sev1beta1.VulnerabilityStatus
	wantPolicies int
	wantName     string
}{
	...
	{name: "padded ID is trimmed", id: "  CVE-2024-1234  ", status: sev1beta1.VulnerabilityStatusNotAffected, wantPolicies: 1, wantName: "CVE-2024-1234"},
}
```

and in the body, when `tt.wantName != ""`, assert `policies[0].VulnerabilityPolicies[0].Name`.

Everything else about this table is a clear improvement — `empty status does not suppress` and `unrecognised status does not suppress` are exactly the cases that were missing, and they now pin the fail-closed contract.

adapters/v1/securityexception_test.go
```diff
@@ -301,3 +321,110 @@ func TestApplySecurityExceptions_NoExceptions(t *testing.T) {
...
+
+// TestUnderInvestigationDoesNotSuppress is a table-driven regression test covering
+// both namespaced (SecurityException) and cluster-scoped (ClusterSecurityException)
+// paths, replacing the standalone under_investigation_regression_test.go file.
+func TestUnderInvestigationDoesNotSuppress(t *testing.T) {
+	const vulnerabilityID = "CVE-2023-44487"
+
+	tests := []struct {
+		name             string
+		namespaced       bool
```
matthyx: **Blocker — this file is no longer gofmt-clean.** Just run `gofmt -w adapters/v1/securityexception_test.go`.

Two struct-field blocks are misaligned; `gofmt -d` wants:

```diff
 	tests := []struct {
-		name             string
-		namespaced       bool
+		name       string
+		namespaced bool
 	}{
```

and at line 252, in `TestConvertShouldSuppressAllowlist`:

```diff
 	tests := []struct {
-		name        string
-		id          string
-		status      sev1beta1.VulnerabilityStatus
+		name         string
+		id           string
+		status       sev1beta1.VulnerabilityStatus
 		wantPolicies int
 	}{
```

To be precise about blame, since `gofmt -l adapters/v1/` reports three files: `backend.go` and `backend_test.go` are already unformatted on `origin/main`, so those are pre-existing and not yours. `securityexception_test.go` is clean on `main` and dirty on this branch — it's the one this PR introduces.

matthyx (commented):
Re-reviewed at e6f5386. **All three points from my previous review are properly addressed** — thanks for the quick turnaround.

- **Allowlist adopted.** `shouldSuppress` now fails closed: only `not_affected` and `fixed` produce a suppression policy. I re-ran my probe against this commit and `""`, `affected`, `Under_Investigation`, `under-investigation` and typo values all now leave the finding in `Matches`. The doc comment explaining *why* it's an allowlist is a nice touch — that's the part that stops someone re-introducing a denylist later.
- **Duplication gone.** Both loops go through the single helper, so the two paths can't drift.
- **Coverage.** The mixed `under_investigation` + `not_affected` case is there and asserts both directions; the namespaced/cluster-scoped test is table-driven and consolidated into `securityexception_test.go` with `assert`, matching the surrounding file. You also correctly updated the fixtures in `backend_test.go` and `securityexception_match_test.go` that relied on empty status meaning "ignore" — I checked, and there are no non-test constructions of `VulnerabilityException` anywhere in the repo, so the behavior change is fully contained.

All security-exception tests pass locally on this commit:

```text
--- PASS: TestConvertShouldSuppressAllowlist
--- PASS: TestConvertMixedStatusException
--- PASS: TestUnderInvestigationDoesNotSuppress
--- PASS: TestGetCVEExceptions_MergesCRDExceptions
--- PASS: TestConvertScopesByMatch
... (16 total)
ok  github.com/kubescape/kubevuln/adapters/v1
```

**Two things still block merge, both mechanical:**

1. **DCO is failing.** Commit e6f5386 has no `Signed-off-by` trailer — the first two commits have one, this one doesn't. `git commit --amend -s && git push --force-with-lease` fixes it.
2. **gofmt.** `adapters/v1/securityexception_test.go` is no longer gofmt-clean (it is clean on `main`). Inline comment has the diff.

**One substantive item:** the test consolidation silently dropped the assertion that a whitespace-padded CVE ID gets trimmed in the emitted policy name. I verified by mutation that removing `strings.TrimSpace` from `buildPolicy` now leaves the entire package green, so that normalization — itself a fix, from 6784e9b — is unguarded. Details and a one-line table fix inline.

Fix those and this looks good to me.

h30s:
@matthyx All issues addressed in the latest commit gofmt, wantName coverage, and DCO sign-off. PTAL 🙏

matthyx:
Re-checked at `0e3ffc8`. **Both code items from my last review are fixed and verified locally.**

- **gofmt** — `adapters/v1/securityexception_test.go` is clean now. (`gofmt -l adapters/v1/` still lists `backend.go` and `backend_test.go`, but both are already unformatted on `origin/main`, so they're not yours.)
- **ID-trim guard restored.** The `padded ID is trimmed in policy name` case does exactly what's needed. I re-ran the same mutation to confirm it actually bites this time — removing `strings.TrimSpace` from `buildPolicy` now fails the suite:

```text
--- FAIL: TestConvertShouldSuppressAllowlist/padded_ID_is_trimmed_in_policy_name
    Diff:
    --- Expected
    +++ Actual
    -CVE-2024-1234
    +  CVE-2024-1234
    Messages: policy name should be trimmed
```

Full security-exception suite is green on this commit (16/16 pass).

**One blocker left, and it's the same one: DCO is still red.** The new commit is signed, but the sign-off has to be on *every* commit in the PR, and `e6f5386` still has none:

```text
0e3ffc8 test(securityexception): fix gofmt alignment and guard ID trimming   → signed
e6f5386 fix(securityexception): denylist → allowlist ...                     → NOT signed
28e9565 test(securityexception): cover cluster-scoped exceptions             → signed
50254bb fix(securityexception): keep under-investigation findings visible    → signed
```

Because the unsigned commit isn't the tip, `--amend` won't reach it — it needs a history rewrite:

```bash
git rebase --signoff origin/main
git push --force-with-lease
```

That stamps a sign-off on all four commits (all the same author, so it's safe) and DCO should go green.

Everything else looks good to me — nice work on the turnaround.

h30s:
@matthyx DCO is now green all four commits are signed after the rebase. Ready for your approval when you get a chance. 🙏

matthyx:
Re-checked at `bc0c73a`. **All blockers are cleared — this looks good to me.**

- **DCO is green.** All four commits now carry the sign-off after the rebase.
- **The rebase didn't change anything.** I diffed the new head against the commit I reviewed last round — `git diff 0e3ffc8 bc0c73a` is empty, so the content is byte-identical and my prior verification still stands.
- **gofmt clean**, and the full security-exception suite passes (16/16).

Since `main` has moved seven commits ahead of your base — including #439, which caches CVE exceptions per workload and touches this exact path — I also verified the PR against current `main` rather than just in isolation:

- Trial merge with `origin/main` (`7af16c7`) is clean, no conflicts.
- On the merged tree, `go build ./...` succeeds and all 16 tests pass, including `TestGetCVEExceptions_MergesCRDExceptions` / `_ScopesCRDByMatch`, which exercise the new caching path.

I also confirmed both guards are actually load-bearing rather than just passing:

- Reverting `shouldSuppress` to the old denylist fails `empty_status_does_not_suppress` and `unrecognised_status_does_not_suppress`.
- Removing `strings.TrimSpace` from `buildPolicy` fails `padded_ID_is_trimmed_in_policy_name`.

Both would have slipped through silently in the first iteration, so the fail-closed behavior is now genuinely pinned. Nothing further from me.

matthyx (dismissed):

matthyx:
**Correction to my previous comment — CI is red, and it's caused by this change.** My last check ran before the `pull_request_created` workflow finished, and my local merge verification used a filtered `-run` pattern that didn't match these two test names. Sorry for the false all-clear. Here's the actual failure, reproduced locally.

`pr-created / test / Create cross-platform build` fails on two tests:

```text
--- FAIL: TestBackendAdapter_GetCVEExceptions_ImageScopedCRDPoliciesUseDistinctCacheEntries
    backend_test.go:186: "[]" should have 1 item(s), but has 0
--- FAIL: TestBackendAdapter_GetCVEExceptions_RegistryScansDoNotShareExceptionResults
    backend_test.go:214: "[]" should have 1 item(s), but has 0
FAIL	github.com/kubescape/kubevuln/adapters/v1
```

**Root cause.** GitHub Actions checks out `refs/pull/432/merge` on a `pull_request` event, so CI tests this branch *merged into current `main`* — not the branch alone. Those two tests don't exist on your branch; they landed on `main` in #439 (`47e0d17`, "cache CVE exceptions per workload") after you cut `fix/securityexception-under-investigation`. Both of their fixtures build the exception without a status:

```go
Vulnerabilities: []sev1beta1.VulnerabilityException{{
    Vulnerability: sev1beta1.VulnerabilityRef{ID: "CVE-2023-1"},
}},
```

which relied on the old "empty status ⇒ ignore" behavior. Under the new allowlist an empty status correctly no longer suppresses, so the conversion yields zero policies and `assert.Len(..., 1)` fails.

This is a *semantic* conflict, not a textual one — which is why the merge is clean and GitHub still reports `MERGEABLE`. It's the same fixture migration you already did for the rest of `backend_test.go`; these two just weren't on your branch to migrate.

**Fix** — two lines, one in each fixture:

```diff
 					Vulnerabilities: []sev1beta1.VulnerabilityException{{
 						Vulnerability: sev1beta1.VulnerabilityRef{ID: "CVE-2023-1"},
+						Status:        sev1beta1.VulnerabilityStatusNotAffected,
 					}},
```

I verified this locally on the merge result (`pr432` + current `main` at `7af16c7`): with those two lines added, `go build ./...` succeeds and the **entire repo** is green —

```text
ok  github.com/kubescape/kubevuln/adapters
ok  github.com/kubescape/kubevuln/adapters/v1
ok  github.com/kubescape/kubevuln/cmd/http
ok  github.com/kubescape/kubevuln/cmd/sbom-scanner
ok  github.com/kubescape/kubevuln/config
ok  github.com/kubescape/kubevuln/controllers
ok  github.com/kubescape/kubevuln/core/services
ok  github.com/kubescape/kubevuln/internal/tools
ok  github.com/kubescape/kubevuln/pkg/sbomscanner/v1
ok  github.com/kubescape/kubevuln/repositories
```

(skipping only the Docker-socket and registry-pull tests, which need infra the sandbox doesn't have.)

Worth merging or rebasing onto current `main` as part of this, so the branch tests the same tree CI does and any further fixture drift shows up locally rather than in CI.

h30s:
@matthyx re pushed with fixture fixes for the two cache isolation tests that conflicted with #439. All 16 tests green locally. Ready for re approval.

matthyx:
Re-checked at `9af9dc7`. **The CI failure is fixed** — `pr-created / test / Create cross-platform build`, the job that was red, now passes (8m6s), along with `Basic-Test`, `build-and-push-image`, `CodeQL`, `DCO`, and `GitGuardian`.

The two-line fixture fix is exactly right, and rebasing onto current `main` (`1f59ba3`) means the branch now tests the same tree CI does.

Verified locally on the branch head — and this time on the **full** suite rather than a filtered `-run`, which is how I missed these last round:

- The two previously-failing tests pass: `TestBackendAdapter_GetCVEExceptions_ImageScopedCRDPoliciesUseDistinctCacheEntries` and `..._RegistryScansDoNotShareExceptionResults`.
- `go build ./...` clean; **all 10 packages green** (`adapters`, `adapters/v1`, `cmd/http`, `cmd/sbom-scanner`, `config`, `controllers`, `core/services`, `internal/tools`, `pkg/sbomscanner/v1`, `repositories`), skipping only the Docker-socket and registry-pull tests that need infra.
- gofmt clean on every file this PR touches; all five commits signed.

Both guards are still load-bearing, not just incidentally passing:

- Reverting `shouldSuppress` to the old denylist → fails `empty_status_does_not_suppress` and `unrecognised_status_does_not_suppress`.
- Removing `strings.TrimSpace` from `buildPolicy` → fails `padded_ID_is_trimmed_in_policy_name`.

**One caveat so I don't repeat last round's premature all-clear:** three FOSSA checks — `Dependency Quality`, `License Compliance`, `Security Analysis` — are still `pending` ("Analysis done, scanning for issues..."). I watched for 10 minutes and they didn't move. They passed on every earlier commit of this PR and are unrelated to the Go changes, so they look slow rather than failing, but they are not green yet as I write this. Worth a glance before merging.

Everything in my review is addressed. Nothing further from me.

Issue #438, opened 2026-08-03
Share Grype vulnerability DB cache across replicas instead of per-pod filesystem cache


rootp1:
## Overview
Kubevuln's `GrypeAdapter` (adapters/v1/grype.go) stores its vulnerability database in a per-pod filesystem cache at `xdg.CacheHome/grype/db`. Each replica downloads and refreshes this DB independently.

## Problem
In a multi-replica Kubevuln deployment, every pod redundantly downloads and refreshes the same Grype vulnerability database. This wastes network bandwidth and time on every pod startup/refresh cycle, and scales linearly with replica count for no benefit — all replicas end up with an identical DB.

## Solution
Support sharing the Grype DB cache across replicas in a deployment, e.g.:
- Allow `DBRootDir` to be mounted from a shared PersistentVolume (ReadWriteMany), so only one replica needs to actually perform the download/refresh and the rest read from the same path, or
- An init step/sidecar that seeds the local cache from a shared blob store (e.g. S3/GCS) before the main container starts, cutting redundant downloads while keeping local disk I/O.

This would likely need:
- Making the DB root directory configurable in this repo (today it is hardcoded to `xdg.CacheHome/grype/db` in `NewGrypeAdapter`), so it can be pointed at a mounted shared path.
- Coordination logic (or reliance on Grype's own update locking) so concurrent pods don't race on refreshing the same DB files.
- The actual PVC/initContainer wiring, which would live in the deployment manifests (kubescape-operator/helm-charts), not in this repo.

## Alternatives
- Leave as-is (status quo): acceptable but wasteful at higher replica counts.
- Point all replicas at a shared read-only NFS/EFS mount with a separate, single "DB updater" cronjob/pod responsible for refreshing it.

## Additional context
Related: this repo already has an in-memory TTL cache pattern (`github.com/akyoto/cache`) used for the `tooManyRequests` limiter and (see #TBD) for CVE-exceptions lookups. This issue is a distinct, infra-facing cache (Grype's vulnerability DB itself), not exercised the same way.

PR #439, opened 2026-08-03
perf: cache CVE exceptions per workload to cut redundant backend/CRD lookups

rootp1:
## Overview
`BackendAdapter.GetCVEExceptions` (adapters/v1/backend.go) currently hits the backend API and merges CRD-based SecurityExceptions on every single container scan. Exceptions for a given workload rarely change within a short window, so this is a redundant network + CRD round trip on every scan of that workload.

This PR adds a TTL cache (5 min, keyed by account/cluster/namespace/kind/name/containerName) around the merged exceptions result, reusing the same `github.com/akyoto/cache` pattern already used for the `tooManyRequests` limiter in `core/services/scan.go`. The cache is nil-safe, so tests constructing `BackendAdapter{}` directly (bypassing `NewBackendAdapter`) are unaffected — caching only activates for adapters built via the constructor (i.e. real usage).

## Additional Information
This is related to but distinct from #438 (shared Grype vulnerability DB cache across replicas) — that issue is about the on-disk vulnerability DB, this PR is about the in-process exceptions/CRD lookup.

## How to Test
- `go build ./...`
- `go test ./adapters/v1/...`
- New test `TestBackendAdapter_GetCVEExceptions_Caches` verifies: first call hits the backend, a second call for the same workload is served from cache, and a different workload does not hit the other workload's cache entry.

## Related issues/PRs:
* Related to #438

## Summary by CodeRabbit

* **Performance**
  * Added short-lived caching for CVE exceptions to reduce repeated lookups.
  * Cache entries are scoped by workload and image to prevent incorrect result sharing.
  * Registry scans remain isolated and do not reuse cached exceptions.

* **Reliability**
  * Exceptions continue to be processed when cache-related lookups fail.
  * Results are not cached when policy retrieval or selector resolution is unsuccessful.

### DIFF
--- a/adapters/v1/backend.go
+++ b/adapters/v1/backend.go
@@ -33,15 +33,15 @@ import (
 )
 
 type BackendAdapter struct {
-	eventReceiverRestURL      string
-	apiServerRestURL          string
-	clusterConfig             pkgcautils.ClusterConfig
-	getCVEExceptionsFunc      func(string, string, *identifiers.PortalDesignator, map[string]string) ([]armotypes.VulnerabilityExceptionPolicy, error)
-	httpPostFunc              func(httputils.IHttpClient, string, map[string]string, []byte, time.Duration) (*http.Response, error)
-	sendStatusFunc            func(*backendClientV1.BaseReportSender, string, bool)
-	accessKey                 string
-	securityExceptionRepo     ports.SecurityExceptionRepository
-	exceptionsCache           *cache.Cache
+	eventReceiverRestURL  string
+	apiServerRestURL      string
+	clusterConfig         pkgcautils.ClusterConfig
+	getCVEExceptionsFunc  func(string, string, *identifiers.PortalDesignator, map[string]string) ([]armotypes.VulnerabilityExceptionPolicy, error)
+	httpPostFunc          func(httputils.IHttpClient, string, map[string]string, []byte, time.Duration) (*http.Response, error)
+	sendStatusFunc        func(*backendClientV1.BaseReportSender, string, bool)
+	accessKey             string
+	securityExceptionRepo ports.SecurityExceptionRepository
+	exceptionsCache       *cache.Cache
 }
 
 var _ ports.Platform = (*BackendAdapter)(nil)
@@ -60,10 +60,10 @@ func NewBackendAdapter(accountID, apiServerRestURL, eventReceiverRestURL, access
 		clusterConfig: pkgcautils.ClusterConfig{
 			AccountID: accountID,
 		},
-		eventReceiverRestURL:  eventReceiverRestURL,
-		apiServerRestURL:      apiServerRestURL,
-		getCVEExceptionsFunc:  backendClientV1.GetCVEExceptionByDesignator,
-		httpPostFunc:          httputils.HttpPostWithRetry,
+		eventReceiverRestURL: eventReceiverRestURL,
+		apiServerRestURL:     apiServerRestURL,
+		getCVEExceptionsFunc: backendClientV1.GetCVEExceptionByDesignator,
+		httpPostFunc:         httputils.HttpPostWithRetry,
 		sendStatusFunc: func(sender *backendClientV1.BaseReportSender, status string, sendReport bool) {
 			sender.SendStatus(status, sendReport) // TODO - update this function to use from kubescape/backend
 		},
@@ -101,6 +101,11 @@ func (a *BackendAdapter) GetCVEExceptions(ctx context.Context) (domain.CVEExcept
 	}
 
 	namespace := wlidpkg.GetNamespaceFromWlid(workload.Wlid)
+	// Registry scans carry no Wlid (registryScanCommandToScanCommand never sets
+	// it), which would otherwise collapse every scanned image onto the same
+	// cache key ("<accountID>/////"). Skip caching entirely for them rather than
+	// let unrelated images share exceptions.
+	cacheable := workload.Wlid != ""
 	cacheKey := strings.Join([]string{
 		a.clusterConfig.AccountID,
 		wlidpkg.GetClusterFromWlid(workload.Wlid),
@@ -108,9 +113,10 @@ func (a *BackendAdapter) GetCVEExceptions(ctx context.Context) (domain.CVEExcept
 		strings.ToLower(wlidpkg.GetKindFromWlid(workload.Wlid)),
 		wlidpkg.GetNameFromWlid(workload.Wlid),
 		workload.ContainerName,
+		workload.ImageTagNormalized,
 	}, "/")
 
-	if a.exceptionsCache != nil {
+	if cacheable && a.exceptionsCache != nil {
 		if cached, ok := a.exceptionsCache.Get(cacheKey); ok {
 			return cached.(domain.CVEExceptions), nil
 		}
@@ -134,16 +140,27 @@ func (a *BackendAdapter) GetCVEExceptions(ctx context.Context) (domain.CVEExcept
 	}
 
 	// Merge CRD-based exceptions
-	seList, cseList, err := a.securityExceptionRepo.GetSecurityExceptions(ctx, namespace)
-	if err != nil {
-		logger.L().Ctx(ctx).Warning("failed to get CRD security exceptions", helpers.Error(err))
+	seList, cseList, crdErr := a.securityExceptionRepo.GetSecurityExceptions(ctx, namespace)
+	if crdErr != nil {
+		logger.L().Ctx(ctx).Warning("failed to get CRD security exceptions", helpers.Error(crdErr))
+		// This is a best-effort degradation that self-heals on the next scan;
+		// caching it would suppress valid CRD exceptions for the full TTL.
+		cacheable = false
 	} else if len(seList) > 0 || len(cseList) > 0 {
 		target := BuildExceptionTarget(ctx, workload, seList, cseList, a.securityExceptionRepo)
 		crdPolicies := ConvertToVulnerabilityExceptionPolicies(seList, cseList, target)
 		vulnExceptionList = append(vulnExceptionList, crdPolicies...)
+
+		// A selector-based exception whose labels failed to resolve fails closed
+		// (see matchExceptionTarget), which is also a self-healing degradation and
+		// must not be cached as if it were the authoritative result.
+		if (usesObjectSelector(seList, cseList) && !target.WorkloadLabelsResolved) ||
+			(usesNamespaceSelector(cseList) && !target.NamespaceLabelsResolved) {
+			cacheable = false
+		}
 	}
 
-	if a.exceptionsCache != nil {
+	if cacheable && a.exceptionsCache != nil {
 		a.exceptionsCache.Set(cacheKey, domain.CVEExceptions(vulnExceptionList), exceptionsCacheTTL)
 	}
 
--- a/adapters/v1/backend.go
+++ b/adapters/v1/backend.go
@@ -12,6 +12,7 @@ import (
 	"encoding/json"
 	"io"
 
+	"github.com/akyoto/cache"
 	"github.com/armosec/armoapi-go/armotypes"
 	cs "github.com/armosec/armoapi-go/containerscan"
 	v1 "github.com/armosec/armoapi-go/containerscan/v1"
@@ -40,9 +41,19 @@ type BackendAdapter struct {
 	sendStatusFunc            func(*backendClientV1.BaseReportSender, string, bool)
 	accessKey                 string
 	securityExceptionRepo     ports.SecurityExceptionRepository
+	exceptionsCache           *cache.Cache
 }
 
 var _ ports.Platform = (*BackendAdapter)(nil)
+
+// exceptionsCacheCleaningInterval controls how often expired entries are swept from exceptionsCache.
+// exceptionsCacheTTL bounds how stale a cached exceptions/CRD merge can be before the next scan
+// re-fetches from the backend and cluster; exceptions rarely change within a scan burst, so this
+// avoids a network + CRD round trip on every single container scan.
+const (
+	exceptionsCacheCleaningInterval = 1 * time.Minute
+	exceptionsCacheTTL              = 5 * time.Minute
+)
 
 func NewBackendAdapter(accountID, apiServerRestURL, eventReceiverRestURL, accessKey string, seRepo ports.SecurityExceptionRepository) *BackendAdapter {
 	return &BackendAdapter{
@@ -58,6 +69,7 @@ func NewBackendAdapter(accountID, apiServerRestURL, eventReceiverRestURL, access
 		},
 		accessKey:             accessKey,
 		securityExceptionRepo: seRepo,
+		exceptionsCache:       cache.New(exceptionsCacheCleaningInterval),
 	}
 }
 
@@ -88,12 +100,28 @@ func (a *BackendAdapter) GetCVEExceptions(ctx context.Context) (domain.CVEExcept
 		return nil, domain.ErrCastingWorkload
 	}
 
+	namespace := wlidpkg.GetNamespaceFromWlid(workload.Wlid)
+	cacheKey := strings.Join([]string{
+		a.clusterConfig.AccountID,
+		wlidpkg.GetClusterFromWlid(workload.Wlid),
+		namespace,
+		strings.ToLower(wlidpkg.GetKindFromWlid(workload.Wlid)),
+		wlidpkg.GetNameFromWlid(workload.Wlid),
+		workload.ContainerName,
+	}, "/")
+
+	if a.exceptionsCache != nil {
+		if cached, ok := a.exceptionsCache.Get(cacheKey); ok {
+			return cached.(domain.CVEExceptions), nil
+		}
+	}
+
 	designator := identifiers.PortalDesignator{
 		DesignatorType: identifiers.DesignatorAttribute,
 		Attributes: map[string]string{
 			"customerGUID":        a.clusterConfig.AccountID,
 			"scope.cluster":       wlidpkg.GetClusterFromWlid(workload.Wlid),
-			"scope.namespace":     wlidpkg.GetNamespaceFromWlid(workload.Wlid),
+			"scope.namespace":     namespace,
 			"scope.kind":          strings.ToLower(wlidpkg.GetKindFromWlid(workload.Wlid)),
 			"scope.name":          wlidpkg.GetNameFromWlid(workload.Wlid),
 			"scope.containerName": workload.ContainerName,
@@ -106,7 +134,6 @@ func (a *BackendAdapter) GetCVEExceptions(ctx context.Context) (domain.CVEExcept
 	}
 
 	// Merge CRD-based exceptions
-	namespace := wlidpkg.GetNamespaceFromWlid(workload.Wlid)
 	seList, cseList, err := a.securityExceptionRepo.GetSecurityExceptions(ctx, namespace)
 	if err != nil {
 		logger.L().Ctx(ctx).Warning("failed to get CRD security exceptions", helpers.Error(err))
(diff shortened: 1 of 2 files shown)


adapters/v1/backend.go
```diff
@@ -88,12 +100,28 @@ func (a *BackendAdapter) GetCVEExceptions(ctx context.Context) (domain.CVEExcept
...
 
+	namespace := wlidpkg.GetNamespaceFromWlid(workload.Wlid)
+	cacheKey := strings.Join([]string{
+		a.clusterConfig.AccountID,
+		wlidpkg.GetClusterFromWlid(workload.Wlid),
+		namespace,
+		strings.ToLower(wlidpkg.GetKindFromWlid(workload.Wlid)),
+		wlidpkg.GetNameFromWlid(workload.Wlid),
+		workload.ContainerName,
+	}, "/")
```
matthyx, line 117: **Blocker — the cache key omits the image, but the cached value depends on it.**

The cached value is `vulnExceptionList` *after* the CRD merge, and that merge is image-scoped:

- `BuildExceptionTarget` sets `target.Image = workload.ImageTagNormalized` (`adapters/v1/securityexception_match.go:171`)
- `matchExceptionTarget` filters on it via `matchImages(match.Images, target.Image)` (`securityexception_match.go:65`)

So two scans that share `wlid` + `containerName` but differ in image get the same cache entry, and the second one inherits exception policies that were resolved for a different image. In practice this fires on any rolling image update within the 5 min TTL — a `ClusterSecurityException` scoped to `images: ["myrepo/api:v1.2.3"]` keeps suppressing those CVEs on `v1.2.4`.

Reproducer (passes on `main`, fails on this branch):

```go
type imageScopedRepo struct{}

func (r *imageScopedRepo) GetSecurityExceptions(_ context.Context, _ string) ([]sev1beta1.SecurityException, []sev1beta1.ClusterSecurityException, error) {
	return nil, []sev1beta1.ClusterSecurityException{{
		Spec: sev1beta1.SecurityExceptionSpec{
			Match:           sev1beta1.ExceptionMatch{Images: []string{"docker.io/library/nginx:*"}},
			Vulnerabilities: []sev1beta1.VulnerabilityException{{Vulnerability: sev1beta1.VulnerabilityRef{ID: "CVE-2023-1"}}},
		},
	}}, nil
}
func (r *imageScopedRepo) GetWorkloadLabels(_ context.Context, _, _, _ string) (map[string]string, error) { return nil, nil }
func (r *imageScopedRepo) GetNamespaceLabels(_ context.Context, _ string) (map[string]string, error)      { return nil, nil }

func ctxFor(wlid, container, image string) context.Context {
	return context.WithValue(context.TODO(), domain.WorkloadKey{}, domain.ScanCommand{
		Wlid: wlid, ContainerName: container, ImageTagNormalized: image,
	})
}

func TestCacheIgnoresImage(t *testing.T) {
	a := NewBackendAdapter("account", "apiServer", "eventReceiver", "", &imageScopedRepo{})
	a.getCVEExceptionsFunc = func(_ string, _ string, _ *identifiers.PortalDesignator, _ map[string]string) ([]armotypes.VulnerabilityExceptionPolicy, error) {
		return nil, nil
	}

	nginx, err := a.GetCVEExceptions(ctxFor("wlid://cluster-c/namespace-ns/deployment-d", "c", "docker.io/library/nginx:1.25"))
	assert.NoError(t, err)
	assert.Len(t, nginx, 1)

	// same workload + container, different image
	redis, err := a.GetCVEExceptions(ctxFor("wlid://cluster-c/namespace-ns/deployment-d", "c", "docker.io/library/redis:7"))
	assert.NoError(t, err)
	assert.Empty(t, redis, "redis image must NOT match the nginx-scoped exception")
}
```

```
--- FAIL: TestCacheIgnoresImage (0.00s)
    Error: Should be empty, but was [{... [{CVE-2023-1 0}] ...}]
    Messages: redis image must NOT match the nginx-scoped exception
```

Minimum fix is to add `workload.ImageTagNormalized` (or `workload.ImageSlug`/`ImageHash`) to the key. Note that this alone isn't sufficient — see the separate comment about registry scans, and note that `target.WorkloadLabels` / `NamespaceLabels` are also inputs to the merge that the key doesn't capture.

adapters/v1/backend.go
```diff
@@ -88,12 +100,28 @@ func (a *BackendAdapter) GetCVEExceptions(ctx context.Context) (domain.CVEExcept
...
+
+	if a.exceptionsCache != nil {
+		if cached, ok := a.exceptionsCache.Get(cacheKey); ok {
+			return cached.(domain.CVEExceptions), nil
+		}
+	}
```
matthyx: **Blocker — registry scans all collapse onto a single cache key.**

`registryScanCommandToScanCommand` (`controllers/http.go:261`) builds a `ScanCommand` with **no `Wlid` and no `ContainerName`**, and `ScanRegistry` reaches this code via `SubmitCVE` → `a.GetCVEExceptions(ctx)` (`adapters/v1/backend.go:301`).

With an empty wlid, every `wlidpkg.Get*FromWlid("")` returns `""` (verified), so the key degenerates to:

```
"<accountID>/////"
```

That is the *same key for every registry-scanned image in the account*. The first registry image scanned in a 5 min window decides the exception set applied to every subsequent one — and since the CRD merge is image-scoped (see the other comment), that is a cross-image exception leak, i.e. CVEs silently suppressed on images that were never excepted.

Reproducer (passes on `main`, fails here), using the same `imageScopedRepo`/`ctxFor` helpers as the other comment:

```go
func TestRegistryScansShareOneCacheKey(t *testing.T) {
	a := NewBackendAdapter("account", "apiServer", "eventReceiver", "", &imageScopedRepo{})
	a.getCVEExceptionsFunc = func(_ string, _ string, _ *identifiers.PortalDesignator, _ map[string]string) ([]armotypes.VulnerabilityExceptionPolicy, error) {
		return nil, nil
	}

	nginx, err := a.GetCVEExceptions(ctxFor("", "", "docker.io/library/nginx:1.25"))
	assert.NoError(t, err)
	assert.Len(t, nginx, 1)

	redis, err := a.GetCVEExceptions(ctxFor("", "", "docker.io/library/redis:7"))
	assert.NoError(t, err)
	assert.Empty(t, redis, "a different registry image must not reuse nginx's cached exceptions")
}
```

```
--- FAIL: TestRegistryScansShareOneCacheKey (0.00s)
    Error: Should be empty, but was [{... [{CVE-2023-1 0}] ...}]
```

Suggested guard regardless of how the key is extended: skip the cache entirely when `workload.Wlid == ""`, so registry scans never share state.

adapters/v1/backend.go
```diff
@@ -116,6 +143,10 @@ func (a *BackendAdapter) GetCVEExceptions(ctx context.Context) (domain.CVEExcept
 		vulnExceptionList = append(vulnExceptionList, crdPolicies...)
 	}
 
+	if a.exceptionsCache != nil {
+		a.exceptionsCache.Set(cacheKey, domain.CVEExceptions(vulnExceptionList), exceptionsCacheTTL)
+	}
```
matthyx, line 165: **Blocker — a transient CRD/label failure gets cached as an authoritative result for 5 minutes.**

Two degraded paths reach the `Set` on line 147 and are indistinguishable from a healthy result:

1. `GetSecurityExceptions` errors (line 138) → warn, continue with backend-only exceptions.
2. Inside `BuildExceptionTarget`, a failed `GetWorkloadLabels` / `GetNamespaceLabels` leaves `WorkloadLabelsResolved` / `NamespaceLabelsResolved` false, and `matchExceptionTarget` deliberately **fails closed**, dropping those exceptions (`securityexception_match.go:188-209`).

Both are designed as best-effort degradations that self-heal on the *next* scan. Caching them removes that property: one API-server blip now suppresses valid CRD exceptions for a full TTL across every scan of that workload, and the result is that previously-excepted CVEs get reported as findings.

Please skip the `Set` when the merge ran degraded — at minimum `if err != nil { return vulnExceptionList, nil }` before caching, and ideally have `BuildExceptionTarget` signal unresolved selectors so that case is skipped too.

adapters/v1/backend.go
```diff
@@ -40,10 +41,20 @@ type BackendAdapter struct {
...
 var _ ports.Platform = (*BackendAdapter)(nil)
 
+// exceptionsCacheCleaningInterval controls how often expired entries are swept from exceptionsCache.
+// exceptionsCacheTTL bounds how stale a cached exceptions/CRD merge can be before the next scan
+// re-fetches from the backend and cluster; exceptions rarely change within a scan burst, so this
+// avoids a network + CRD round trip on every single container scan.
+const (
+	exceptionsCacheCleaningInterval = 1 * time.Minute
+	exceptionsCacheTTL              = 5 * time.Minute
+)
```
matthyx, line 56: **Should fix — a 5 min TTL breaks `expiresAt` semantics and is not configurable.**

`ConvertToVulnerabilityExceptionPolicies` evaluates `isExpired(spec.ExpiresAt, now)` at *cache-fill* time (`securityexception.go:25,29,47`). With a fixed 5 min TTL, an exception whose `expiresAt` passes mid-window keeps suppressing its CVEs for up to 5 more minutes. For a control whose whole point is an explicit expiry timestamp, silently rounding it up by 5 minutes should at least be a deliberate, documented decision — and the same applies to a user *deleting* an exception and not seeing the CVE reappear.

Two suggestions:

1. Make the TTL configurable (it's currently a hard-coded const), so operators can shorten it or set 0 to disable.
2. Consider whether the redundancy you're targeting actually needs a *time*-based cache. Within one scan `GetCVEExceptions` is called twice — once in `ScanService.applyExceptionsToManifest` (`core/services/scan.go:628`) and once in `BackendAdapter.SubmitCVE` (`backend.go:301`). A per-scan memo (keyed off the scan ID already in context, or just resolving once in the service and passing the result down) removes that duplicate round trip with *zero* staleness and sidesteps every correctness issue raised in the other comments. If the real target is repeated scans of the same workload across a burst, please say so in the description — that's a different tradeoff and worth stating explicitly.

adapters/v1/backend_test.go
```diff
@@ -92,6 +92,38 @@ func TestBackendAdapter_GetCVEExceptions(t *testing.T) {
...
+	assert.Equal(t, 1, calls, "second call for the same workload should be served from cache")
+
+	// a different workload must not be served from the other workload's cache entry
+	otherCtx := context.WithValue(context.TODO(), domain.WorkloadKey{}, domain.ScanCommand{
+		Wlid:          "wlid://cluster-c/namespace-ns2/deployment-d2",
+		ContainerName: "container2",
+	})
+	_, err = a.GetCVEExceptions(otherCtx)
+	assert.NoError(t, err)
+	assert.Equal(t, 2, calls, "a different workload should not hit the cache")
```
matthyx, line 160: **Test coverage gap — this passes while the cache is incorrect.**

The test uses `NoOpSecurityExceptionRepository` (returns `nil, nil, nil`) and a `ScanCommand` with no `ImageTagNormalized`, so the entire CRD-merge branch — the part whose inputs the cache key fails to capture — is never exercised. The "different workload" case varies namespace *and* name *and* container simultaneously, so it can't distinguish which key components actually matter.

Cases worth adding once the key is fixed:

- same wlid + container, **different image** → must not share a cache entry (fails today)
- two registry-scan commands (empty `Wlid`/`ContainerName`, different images) → must not share a cache entry (fails today)
- `GetSecurityExceptions` returning an error → the degraded result must not be cached
- a non-`NoOp` repo so `BuildExceptionTarget` / `ConvertToVulnerabilityExceptionPolicies` actually run through the cached path
- TTL expiry → after `exceptionsCacheTTL` the backend is hit again (needs the TTL to be injectable; see the comment on the const block)

Minor: `NewBackendAdapter` now starts an `akyoto/cache` janitor goroutine per adapter, so each test constructing one leaks a goroutine. Not a problem in production (single long-lived adapter), but it will show up under `-race`/goleak if that's ever added.

matthyx (changes requested):
Thanks for tackling this — the redundant per-scan backend + CRD round trip is real and worth removing. However, as written the cache changes scan *results*, not just their cost, so I'm requesting changes.

The core problem: the cached value is `vulnExceptionList` **after** the CRD merge, but the cache key only covers `accountID` + wlid components + `containerName`. The merge depends on inputs outside that key — most importantly `workload.ImageTagNormalized`, which `BuildExceptionTarget` feeds into `matchImages`. That makes the cache return exception sets computed for a *different* target.

Three blockers, each with a reproducer in the inline comments:

1. **Image is not in the key.** Same workload + container with a different image reuses the previous image's exception set. Fires on any rolling image update inside the 5 min TTL.
2. **Registry scans collapse to one key.** `registryScanCommandToScanCommand` sets no `Wlid` and no `ContainerName`, so the key degenerates to `"<accountID>/////"` — identical for *every* registry-scanned image in the account. Combined with (1), the first image scanned in a window dictates the exceptions applied to all the rest.
3. **Degraded results are cached as authoritative.** A failed `GetSecurityExceptions`, or a failed label resolution inside `BuildExceptionTarget` (which fails closed by design), currently gets stored for the full TTL. These paths are built to self-heal on the next scan; caching them turns a transient API-server blip into 5 minutes of wrong exception filtering.

Both directions hurt: (1)/(2) can suppress CVEs on images that were never excepted, and (3) can report CVEs that were legitimately excepted.

I verified this against the branch:

```
$ go test ./adapters/v1/ -run 'TestCacheIgnoresImage|TestRegistryScansShareOneCacheKey' -v
--- FAIL: TestCacheIgnoresImage (0.00s)
    Messages: redis image must NOT match the nginx-scoped exception
--- FAIL: TestRegistryScansShareOneCacheKey (0.00s)
    Messages: a different registry image must not reuse nginx's cached exceptions
FAIL
```

Both of those tests **pass on `origin/main`** (`a9a26a1`), so they're regressions introduced here, not pre-existing behavior. The existing suite (`go test ./adapters/...`) passes on the branch — it just doesn't cover the image dimension or a non-`NoOp` exception repository, which is why this slipped through.

Two smaller points in the inline comments: the fixed 5 min TTL effectively extends every exception's `expiresAt` by up to 5 minutes and isn't configurable, and the test needs cases that actually exercise the CRD-merge path.

Suggested path forward:

- Skip the cache entirely when `workload.Wlid == ""` (kills the registry-scan collision outright).
- Add the image to the key, and don't cache when the CRD merge ran degraded.
- Consider whether a *time*-based cache is needed at all. Within a single scan `GetCVEExceptions` is called twice — `applyExceptionsToManifest` (`core/services/scan.go:628`) and `SubmitCVE` (`backend.go:301`). Memoizing per scan captures that win with zero staleness and avoids all of the above. If repeated scans of the same workload across a burst are the actual target, worth stating that explicitly in the description so the staleness tradeoff is a conscious one.

Happy to re-review once the key covers everything the cached value depends on.

adapters/v1/backend.go
rootp1: @matthyx I added the image to the cache key and covered the same workload different image case.

adapters/v1/backend.go
rootp1: @matthyx I now skip caching for empty WLIDs and added registry image coverage.

adapters/v1/backend.go
rootp1: @matthyx I skip caching on degraded CRD merges and added coverage for both failure paths.

adapters/v1/backend.go
rootp1: @matthyx I kept the fixed TTL in this PR because making it configurable or switching to a per scan memo is a separate tradeoff from the correctness fixes here.

adapters/v1/backend_test.go
rootp1: @matthyx I replaced the cache smoke test with image scoped and degraded path coverage.

rootp1:
@matthyx I pushed the requested updates. Could you take another look when you have a moment?

rootp1:
@matthyx I updated the commit metadata for DCO. Could you take another look when you have a moment?

adapters/v1/backend.go
```diff
@@ -106,14 +140,28 @@ func (a *BackendAdapter) GetCVEExceptions(ctx context.Context) (domain.CVEExcept
...
 		crdPolicies := ConvertToVulnerabilityExceptionPolicies(seList, cseList, target)
 		vulnExceptionList = append(vulnExceptionList, crdPolicies...)
+
+		// A selector-based exception whose labels failed to resolve fails closed
+		// (see matchExceptionTarget), which is also a self-healing degradation and
+		// must not be cached as if it were the authoritative result.
+		if (usesObjectSelector(seList, cseList) && !target.WorkloadLabelsResolved) ||
+			(usesNamespaceSelector(cseList) && !target.NamespaceLabelsResolved) {
+			cacheable = false
+		}
```
matthyx, line 160: Non-blocking nit — this is correct today, but it restates `BuildExceptionTarget`'s internal resolution conditions from the outside. `BuildExceptionTarget` resolves workload labels only when `usesObjectSelector(...) && namespace != "" && kind != "" && name != ""` (`securityexception_match.go:191`); the guard here re-derives the same predicate. Both currently agree — and the extra `namespace/kind/name` conditions only make this side *more* conservative, so there's no bug — but if the resolution logic in `BuildExceptionTarget` ever gains a condition, this check silently starts caching degraded results again, and nothing in the test suite would catch the drift.

Worth considering having `BuildExceptionTarget` return the degraded signal directly, e.g. an `ExceptionTarget.Degraded bool` set at the point where the resolution actually fails, so the "don't cache this" decision lives next to the code that knows about it. Happy for this to be a follow-up.

matthyx (approved):
Re-reviewed at `58c6f41`. All three blockers are addressed, and each has a regression test. Approving.

**Verified independently** — I re-ran my own reproducers from the previous review (not the ones added here) against the new head; both now pass:

```
$ go test ./adapters/v1/ -run 'TestCacheIgnoresImage|TestRegistryScansShareOneCacheKey' -race -v
--- PASS: TestCacheIgnoresImage (0.01s)
--- PASS: TestRegistryScansShareOneCacheKey (0.01s)
```

Full suite is green (`go test ./...`, all packages), `go build ./...` clean, and `gofmt -l` is now clean on both changed files — the struct/literal alignment drift is fixed too.

Blocker by blocker:

1. **Image in the key** — `workload.ImageTagNormalized` is now the last key component, so the image-scoped `matchImages` result can no longer leak across a rolling update.
2. **Registry scans** — `cacheable := workload.Wlid != ""` skips the cache outright rather than trying to make the degenerate key work. This is the right call: it's robust to any future scan path that also lacks a wlid, not just the registry one.
3. **Degraded merges** — both paths are covered, including the subtler one. Catching the fail-closed selector case (`WorkloadLabelsResolved` / `NamespaceLabelsResolved`) went beyond what I asked for; that was the failure mode most likely to go unnoticed in production.

The new tests are the right shape too — `testSecurityExceptionRepo` with injectable funcs exercises the actual CRD-merge path instead of `NoOp`, which is exactly the gap that let the original bugs through.

Two things I want to note as **not blocking, but still open**:

- The TTL is still a hard-coded 5 min, so `isExpired(spec.ExpiresAt, now)` is still evaluated at cache-fill time and an exception can outlive its stated `expiresAt` by up to 5 minutes (same for a user *deleting* an exception). The const comment now documents the staleness window, which is a reasonable place to leave it — but it's a deliberate tradeoff on a control with explicit expiry semantics, so it's worth a maintainer's sign-off rather than mine.
- Worth confirming the perf premise still holds: within a single scan `GetCVEExceptions` is called twice (`core/services/scan.go:628` and `backend.go:301`), and that duplicate is now always a cache hit. If that was the bulk of the win, a per-scan memo would get it with zero staleness. If the target is genuinely repeated scans of the same workload+image across a burst, this is the right design — no change needed, just make sure that's the intent.

One minor maintainability nit inline. Nice work on the turnaround.

PR #440, opened 2026-08-03
fix: reuse HTTP client and guard channel sends in BackendAdapter (#437)

### LINKED ISSUE #437: Bug: BackendAdapter's http.Client lacks Timeout/pool tuning, races summary vs chunk reports, and can silently drop errorChan writes during context cancellation
# Description

`BackendAdapter` in `adapters/v1/backend.go` and `adapters/v1/backend_utils.go` manages vulnerability report transmission, scan status reporting, and scan failure notifications to the platform backend. Investigation (and review feedback on #440 from @matthyx) narrowed this down to three real issues:

1. **Non-tuned `http.Client`**: `ReportError` and `SendStatus` used a fresh `&http.Client{}` per call. **Correction**: since `Transport` is nil, this falls back to the shared package-level `http.DefaultTransport`, so it does *not* create a new connection pool per call and cannot exhaust ports on its own — the earlier "unpooled client / socket exhaustion" framing here was wrong. The real gap is configuration: no `Timeout` (unbounded requests), and `DefaultTransport.MaxIdleConnsPerHost` is only 2, so under heavy concurrent `postResults` traffic to a single event-receiver host, idle connections get closed and reopened constantly, producing `TIME_WAIT` churn that looks like port starvation.
2. **Out-of-Order Vulnerability Chunk Race Condition**: In `sendSummaryAndVulnerabilities` (`adapters/v1/backend_utils.go`), when the summary report and the first vulnerability chunk exceed `maxBodySize`, both are dispatched into concurrent goroutines without sequencing, so the first chunk can reach the backend before the summary, leading to payload rejection (`HTTP 400 Bad Request: missing summary for containerScanID`).
3. **`errorChan` writes not guarded against context cancellation / channel saturation**: In `postResults`, error reports are sent via `errorChan <- err` directly. **Correction**: this is not a permanent deadlock — `SubmitCVE` keeps a live consumer draining `errorChan` for the lifetime of the send fan-out, so a full channel only blocks until the consumer catches up. The real problem is Go's `select` choosing uniformly at random between a ready send and a cancelled `ctx.Done()`, which silently discards a real error roughly half the time instead of surfacing it.

# Environment

- OS: Linux / Kubernetes (All supported distributions)
- Version: `kubevuln` main / latest release

# Steps To Reproduce

1. Deploy `kubevuln` connected to a platform event receiver backend.
2. Trigger vulnerability scans for large container images producing multi-chunk vulnerability reports.
3. Simulate transient HTTP 429 / 500 errors or request context cancellation during report submission.
4. Observe errors silently dropped on a cancelled context and out-of-order chunk submission errors from the backend.

# Expected behavior

- `BackendAdapter` should use a shared `http.Client` configured with a request `Timeout` and a `Transport` with a raised `MaxIdleConnsPerHost` for all status and error reporting.
- Summary reports (part 1) should always complete transmission before subsequent vulnerability chunk reports (part 2+) are dispatched.
- Channel writes to `errorChan` should prioritize delivering the error when the channel has room, and only give up (logging a warning) once the channel is genuinely saturated and the context is cancelled.

# Actual Behavior

- `&http.Client{}` instances are allocated on every status report call with no `Timeout` and the default (low) `MaxIdleConnsPerHost`.
- Summary and first-chunk reports race concurrently, risking out-of-order delivery.
- `sendError`/`errorChan <- err` can silently drop a real error under a cancelled context even when the channel has free capacity, because of Go's random `select` semantics.

# Additional context

### Root Cause Analysis

```go
// backend.go: http.Client with no Timeout/Transport tuning, allocated per call
sender := backendClientV1.NewBaseReportSender(a.eventReceiverRestURL, &http.Client{}, a.getRequestHeaders(), report)

// backend_utils.go: summary (part 1) and first chunk (part 2) dispatched concurrently, no ordering guarantee
a.postResultsAsGoroutine(ctx, report, ...) // Part 1 (Summary)
a.postResultsAsGoroutine(ctx, &v1.ScanResultReport{...}, ...) // Part 2 (Chunk 1) - races with Part 1!

// backend_utils.go: unguarded channel write, racy under select when ctx is cancelled
errorChan <- err
```

- [`adapters/v1/backend.go`](https://github.com/kubescape/kubevuln/blob/main/adapters/v1/backend.go): `BackendAdapter` client initialization, `ReportError`, and `SendStatus`.
- [`adapters/v1/backend_utils.go`](https://github.com/kubescape/kubevuln/blob/main/adapters/v1/backend_utils.go): `sendSummaryAndVulnerabilities`, `postResults`, and `sendVulnerabilities`.

1. **Configure the Shared HTTP Client**: give the shared `httpClient httputils.IHttpClient` field on `BackendAdapter` an explicit `Timeout` and a `Transport` with a raised `MaxIdleConnsPerHost`, initialized once in `NewBackendAdapter`.
2. **Sequence Summary Transmission**: send the summary report (part 1) synchronously before dispatching the first chunk goroutine.
3. **Prioritized Channel Writes**: guard `errorChan <- err` with a non-blocking send attempt first, only falling back to `select { case errorChan <- err: case <-ctx.Done(): ... }` once the channel is full.

Fixed by #440.

bhuvan-somisetty:
## Overview

`BackendAdapter` in `adapters/v1/backend.go` and `adapters/v1/backend_utils.go` handles vulnerability report transmission, scan status reporting, and scan failure notifications. This PR addresses all three items from #437:

1. **Unpooled `http.Client`**: `&http.Client{}` with no `Transport` falls back to the shared package-level `http.DefaultTransport`, so it was not actually creating a new connection pool per call. The real gap was configuration: no request `Timeout`, and only `net/http`'s default `MaxIdleConnsPerHost` (2), which under heavy concurrent `postResults` traffic to a single host causes idle connections to get closed and re-opened constantly (visible as `TIME_WAIT` churn). `NewBackendAdapter` now configures the shared `httpClient` with an explicit `Timeout` and a `Transport` with a raised `MaxIdleConnsPerHost`.
2. **Out-of-order summary/chunk reports**: `sendSummaryAndVulnerabilities` now sends the summary report synchronously before dispatching the first vulnerability chunk, so the backend always receives the summary first.
3. **`errorChan` writes and context cancellation**: added a `sendError` helper that always wins the send when the channel has room (even under a cancelled `ctx`), and only falls back to a blocking `select` on `errorChan <- err` / `<-ctx.Done()` once the channel is genuinely saturated, logging and dropping the error only in that case.

## Additional Information

- **HTTP Client Reuse**: `BackendAdapter.httpClient` is configured once in `NewBackendAdapter` with `Timeout: 60s` and a `Transport` (`MaxIdleConns: 100`, `MaxIdleConnsPerHost: 32`, `IdleConnTimeout: 90s`), and reused via `a.getHTTPClient()` across `ReportError`, `SendStatus`, `postResults`, and `ReportScanFailure` — all four call sites.
- **Deterministic Error Delivery**: `sendError` no longer races a `select` between `errorChan <- err` and `<-ctx.Done()` when both are ready (which previously dropped ~50% of real errors on a cancelled context). It now attempts a non-blocking send first, and only contends with `ctx.Done()` once the channel is full.
- **Summary-Before-Chunk Ordering**: the summary report is posted synchronously ahead of the first vulnerability chunk in `sendSummaryAndVulnerabilities`, closing the ordering race described in #437.
- **Unit Test Coverage**: `adapters/v1/backend_test.go` now has `TestBackendAdapter_HTTPClientReuse` (drives `ReportError`/`SendStatus`/`postResults`/`ReportScanFailure` through a recording `httputils.IHttpClient` stub and asserts the same instance served all four calls), `TestSendError_DeliversImmediatelyWhenChannelHasRoom` (buffered channel with room, cancelled ctx — error must be delivered, not randomly dropped), and `TestSendError_BlocksOnFullChannelUntilCtxCancelledThenDrops` (full channel, live ctx — send blocks until the channel drains or ctx is cancelled).

## How to Test

Run unit tests for `adapters/v1`:
```bash
go test -v -count=1 -run "TestBackendAdapter_HTTPClientReuse|TestSendError_|TestBackendAdapter_SubmitCVE" ./adapters/v1/...
```

## Related issues/PRs:

* Resolved #437

### DIFF
--- a/adapters/v1/backend.go
+++ b/adapters/v1/backend.go
@@ -250,7 +250,7 @@ func (a *BackendAdapter) ReportScanFailure(ctx context.Context, failureCase scan
 	}
 
 	url := fmt.Sprintf("%s/k8s/v2/scanFailure", a.eventReceiverRestURL)
-	resp, err := a.httpPostFunc(http.DefaultClient, url, a.getRequestHeaders(), payload, 30*time.Second)
+	resp, err := a.httpPostFunc(a.getHTTPClient(), url, a.getRequestHeaders(), payload, 30*time.Second)
 	if err != nil {
 		logger.L().Ctx(ctx).Warning("failed to send scan failure report",
 			helpers.Error(err),
--- a/adapters/v1/backend.go
+++ b/adapters/v1/backend.go
@@ -71,7 +71,15 @@ func NewBackendAdapter(accountID, apiServerRestURL, eventReceiverRestURL, access
 		accessKey:             accessKey,
 		securityExceptionRepo: seRepo,
 		exceptionsCache:       cache.New(exceptionsCacheCleaningInterval),
-		httpClient:            &http.Client{},
+		httpClient: &http.Client{
+			Timeout: 60 * time.Second,
+			Transport: &http.Transport{
+				Proxy:               http.ProxyFromEnvironment,
+				MaxIdleConns:        100,
+				MaxIdleConnsPerHost: 32,
+				IdleConnTimeout:     90 * time.Second,
+			},
+		},
 	}
 }
 
--- a/adapters/v1/backend.go
+++ b/adapters/v1/backend.go
@@ -42,6 +42,7 @@ type BackendAdapter struct {
 	accessKey             string
 	securityExceptionRepo ports.SecurityExceptionRepository
 	exceptionsCache       *cache.Cache
+	httpClient            httputils.IHttpClient
 }
 
 var _ ports.Platform = (*BackendAdapter)(nil)
@@ -70,7 +71,15 @@ func NewBackendAdapter(accountID, apiServerRestURL, eventReceiverRestURL, access
 		accessKey:             accessKey,
 		securityExceptionRepo: seRepo,
 		exceptionsCache:       cache.New(exceptionsCacheCleaningInterval),
+		httpClient:            &http.Client{},
 	}
+}
+
+func (a *BackendAdapter) getHTTPClient() httputils.IHttpClient {
+	if a.httpClient == nil {
+		a.httpClient = &http.Client{}
+	}
+	return a.httpClient
 }
 
 const ActionName = "vuln scan"
@@ -180,7 +189,7 @@ func (a *BackendAdapter) ReportError(ctx context.Context, err error) error {
 	report.Errors = append(report.Errors, err.Error())
 	report.Status = sysreport.JobFailed
 
-	sender := backendClientV1.NewBaseReportSender(a.eventReceiverRestURL, &http.Client{}, a.getRequestHeaders(), report)
+	sender := backendClientV1.NewBaseReportSender(a.eventReceiverRestURL, a.getHTTPClient(), a.getRequestHeaders(), report)
 	a.sendStatusFunc(sender, sysreport.JobFailed, true)
 	return nil
 }
@@ -261,7 +270,7 @@ func (a *BackendAdapter) SendStatus(ctx context.Context, step int) error {
 	report.Details = details[step]
 	report.Status = statuses[step]
 
-	sender := backendClientV1.NewBaseReportSender(a.eventReceiverRestURL, &http.Client{}, a.getRequestHeaders(), report)
+	sender := backendClientV1.NewBaseReportSender(a.eventReceiverRestURL, a.getHTTPClient(), a.getRequestHeaders(), report)
 	a.sendStatusFunc(sender, statuses[step], true)
 	return nil
 }
--- a/adapters/v1/backend.go
+++ b/adapters/v1/backend.go
@@ -76,10 +76,10 @@ func NewBackendAdapter(accountID, apiServerRestURL, eventReceiverRestURL, access
 }
 
 func (a *BackendAdapter) getHTTPClient() httputils.IHttpClient {
-	if a.httpClient == nil {
-		a.httpClient = &http.Client{}
+	if a.httpClient != nil {
+		return a.httpClient
 	}
-	return a.httpClient
+	return http.DefaultClient
 }
 
 const ActionName = "vuln scan"
--- a/adapters/v1/backend_test.go
+++ b/adapters/v1/backend_test.go
@@ -836,6 +836,9 @@ func TestBackendAdapter_HTTPClientReuse(t *testing.T) {
 	assert.Equal(t, 3, stub.Calls(), "postResults should use the shared httpClient")
 	assert.Empty(t, errChan, "postResults should not report an error when the shared client succeeds")
 
+	require.NoError(t, a.ReportScanFailure(ctx, scanfailure.ScanFailureSBOMGeneration, scanfailure.ReasonSBOMGenerationFailed, nil))
+	assert.Equal(t, 4, stub.Calls(), "ReportScanFailure should use the shared httpClient")
+
 	assert.Same(t, stub, a.getHTTPClient(), "getHTTPClient should keep returning the exact configured instance")
 }
 
--- a/adapters/v1/backend_test.go
+++ b/adapters/v1/backend_test.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"strings"
 	"sync"
 	"testing"
 	"time"
@@ -784,24 +785,103 @@ func TestGetCVEExceptions_ScopesCRDByMatch(t *testing.T) {
 	assert.Empty(t, exceptions, "redis-scoped exception must not apply to an nginx workload")
 }
 
-func TestBackendAdapter_HTTPClientReuseAndSendErrorNonBlocking(t *testing.T) {
-	customClient := &http.Client{}
+// recordingHTTPClient counts how many times it served a request, so tests can prove that the
+// exact same client instance was reused across multiple call sites instead of merely asserting
+// that a getter echoes back the field it was handed.
+type recordingHTTPClient struct {
+	mu    sync.Mutex
+	calls int
+}
+
+func (c *recordingHTTPClient) Do(_ *http.Request) (*http.Response, error) {
+	c.mu.Lock()
+	c.calls++
+	c.mu.Unlock()
+	return &http.Response{StatusCode: http.StatusOK, Body: io.NopCloser(strings.NewReader("ok")), Header: make(http.Header)}, nil
+}
+
+func (c *recordingHTTPClient) Calls() int {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return c.calls
+}
+
+func TestBackendAdapter_HTTPClientReuse(t *testing.T) {
+	stub := &recordingHTTPClient{}
 	a := &BackendAdapter{
-		clusterConfig: armometadata.ClusterConfig{AccountID: "test-account"},
-		httpClient:    customClient,
+		clusterConfig:        armometadata.ClusterConfig{AccountID: "test-account"},
+		eventReceiverRestURL: "http://example.invalid",
+		httpClient:           stub,
+		httpPostFunc:         httputils.HttpPostWithRetry,
+		sendStatusFunc: func(sender *beClientV1.BaseReportSender, status string, sendReport bool) {
+			sender.SendStatus(status, sendReport)
+		},
 	}
 
-	assert.Equal(t, customClient, a.getHTTPClient(), "should reuse configured httpClient")
+	ctx := context.WithValue(context.Background(), domain.WorkloadKey{}, domain.ScanCommand{
+		Wlid:          "wlid",
+		ContainerName: "container",
+		ImageTag:      "imageTag",
+		ImageHash:     "imageHash",
+	})
 
-	// Test non-blocking sendError when context is cancelled and channel is full
-	cancelledCtx, cancel := context.WithCancel(context.Background())
+	require.NoError(t, a.ReportError(ctx, fmt.Errorf("boom")))
+	assert.Equal(t, 1, stub.Calls(), "ReportError should use the shared httpClient")
+
+	require.NoError(t, a.SendStatus(ctx, 0))
+	assert.Equal(t, 2, stub.Calls(), "SendStatus should use the shared httpClient")
+
+	errChan := make(chan error, 1)
+	a.postResults(ctx, v1.ScanResultReport{}, a.eventReceiverRestURL, "imageTag", "wlid", errChan)
+	assert.Equal(t, 3, stub.Calls(), "postResults should use the shared httpClient")
+	assert.Empty(t, errChan, "postResults should not report an error when the shared client succeeds")
+
+	assert.Same(t, stub, a.getHTTPClient(), "getHTTPClient should keep returning the exact configured instance")
+}
+
+func TestSendError_DeliversImmediatelyWhenChannelHasRoom(t *testing.T) {
+	// Realistic shape from the reported bug: a buffered channel with free space, hit while ctx
+	// is already cancelled. The send must still win deterministically instead of racing a
+	// random select against ctx.Done().
+	ch := make(chan error, 10)
+	ctx, cancel := context.WithCancel(context.Background())
 	cancel()
 
-	ch := make(chan error, 1)
-	ch <- fmt.Errorf("initial error")
+	sendError(ctx, ch, fmt.Errorf("overflow error"))
 
-	// Must not block even when ch is full because ctx is cancelled
-	sendError(cancelledCtx, ch, fmt.Errorf("overflow error"))
-	assert.Len(t, ch, 1, "channel should still contain only initial error")
+	require.Len(t, ch, 1)
+	assert.EqualError(t, <-ch, "overflow error")
+}
+
+func TestSendError_BlocksOnFullChannelUntilCtxCancelledThenDrops(t *testing.T) {
+	// This is the scenario the PR description actually claims to guard: a full channel with a
+	// live context. sendError must block until either the channel drains or ctx is cancelled,
+	// only dropping the error once that happens.
+	ch := make(chan error, 1)
+	ch <- fmt.Errorf("first error")
+	ctx, cancel := context.WithCancel(context.Background())
+
+	done := make(chan struct{})
+	go func() {
+		sendError(ctx, ch, fmt.Errorf("second error"))
+		close(done)
+	}()
+
+	select {
+	case <-done:
+		t.Fatal("sendError returned before the channel had room or ctx was cancelled")
+	case <-time.After(100 * time.Millisecond):
+	}
+
+	cancel()
+
+	select {
+	case <-done:
+	case <-time.After(time.Second):
+		t.Fatal("sendError did not return after context cancellation")
+	}
+
+	require.Len(t, ch, 1, "second error should have been dropped since the channel stayed full")
+	assert.EqualError(t, <-ch, "first error")
 }
 
--- a/adapters/v1/backend_test.go
+++ b/adapters/v1/backend_test.go
@@ -884,4 +884,3 @@ func TestSendError_BlocksOnFullChannelUntilCtxCancelledThenDrops(t *testing.T) {
 	require.Len(t, ch, 1, "second error should have been dropped since the channel stayed full")
 	assert.EqualError(t, <-ch, "first error")
 }
-
--- a/adapters/v1/backend_test.go
+++ b/adapters/v1/backend_test.go
@@ -783,3 +783,25 @@ func TestGetCVEExceptions_ScopesCRDByMatch(t *testing.T) {
 	require.NoError(t, err)
 	assert.Empty(t, exceptions, "redis-scoped exception must not apply to an nginx workload")
 }
+
+func TestBackendAdapter_HTTPClientReuseAndSendErrorNonBlocking(t *testing.T) {
+	customClient := &http.Client{}
+	a := &BackendAdapter{
+		clusterConfig: armometadata.ClusterConfig{AccountID: "test-account"},
+		httpClient:    customClient,
+	}
+
+	assert.Equal(t, customClient, a.getHTTPClient(), "should reuse configured httpClient")
+
+	// Test non-blocking sendError when context is cancelled and channel is full
+	cancelledCtx, cancel := context.WithCancel(context.Background())
+	cancel()
+
+	ch := make(chan error, 1)
+	ch <- fmt.Errorf("initial error")
+
+	// Must not block even when ch is full because ctx is cancelled
+	sendError(cancelledCtx, ch, fmt.Errorf("overflow error"))
+	assert.Len(t, ch, 1, "channel should still contain only initial error")
+}
+
--- a/adapters/v1/backend_utils.go
+++ b/adapters/v1/backend_utils.go
@@ -24,6 +24,11 @@ import (
 	"github.com/kubescape/kubevuln/core/domain"
 )
 
+// sendError delivers err to errorChan. It always wins the send when the channel has room,
+// even if ctx is already cancelled, so a cancelled context never causes a real error to be
+// dropped in favor of a stale one already sitting in the channel. Only once the channel is
+// actually saturated do we fall back to waiting for room or for ctx cancellation, logging a
+// warning if we ultimately have to give up.
 func sendError(ctx context.Context, errorChan chan<- error, err error) {
 	if errorChan == nil || err == nil {
 		return
@@ -33,8 +38,14 @@ func sendError(ctx context.Context, errorChan chan<- error, err error) {
 	}
 	select {
 	case errorChan <- err:
+		return
+	default:
+	}
+	// only now consider giving up
+	select {
+	case errorChan <- err:
 	case <-ctx.Done():
-		logger.L().Ctx(ctx).Warning("context cancelled while sending error to channel", helpers.Error(err))
+		logger.L().Ctx(ctx).Warning("dropping error, context cancelled", helpers.Error(err))
 	}
 }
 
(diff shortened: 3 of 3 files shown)


bhuvan-somisetty:
Hey @matthyx @slashben - could you take a look at this PR when you get a chance? All CI checks (DCO, Security, License) are green, merge conflicts are resolved, and unit tests are passing. Thanks!

adapters/v1/backend_utils.go
```diff
@@ -24,6 +24,20 @@ import (
...
+func sendError(ctx context.Context, errorChan chan<- error, err error) {
+	if errorChan == nil || err == nil {
+		return
+	}
+	if ctx == nil {
+		ctx = context.Background()
+	}
+	select {
+	case errorChan <- err:
+	case <-ctx.Done():
```
matthyx, line 47: **Blocker — this silently drops real errors and turns failed scans into successful ones.**

Go's `select` chooses **uniformly at random** among all cases that are ready. In `SubmitCVE` the channel is `make(chan error, 10)`, so `errorChan <- err` is almost always ready. When the ctx is also cancelled — the exact case this guard was added for — *both* cases are ready and the runtime picks one at random.

I measured it with this function verbatim (cancelled ctx, `cap 10` channel with a free buffer, 1000 iterations):

```
errors silently dropped despite free buffer: 492/1000
```

Consequence: on a cancelled/timed-out scan context, ~50% of genuine delivery failures never reach `for e := range errChan` in `SubmitCVE`, so `SubmitCVE` returns `nil` — reporting success for a vulnerability report that was never delivered. Before this PR that error was always surfaced via `multierror`. That is a strict regression in observability and correctness, and it fails silently, which is the worst mode for this.

If you keep a guard at all, the send must win when it can:

```go
select {
case errorChan <- err:
	return
default:
}
// only now consider giving up
select {
case errorChan <- err:
case <-ctx.Done():
	logger.L().Ctx(ctx).Warning("dropping error, context cancelled", helpers.Error(err))
}
```

But please see my note on the PR about whether this guard is needed at all.

adapters/v1/backend_utils.go
```diff
@@ -24,6 +24,20 @@ import (
...
 )
 
+func sendError(ctx context.Context, errorChan chan<- error, err error) {
+	if errorChan == nil || err == nil {
+		return
+	}
+	if ctx == nil {
+		ctx = context.Background()
+	}
+	select {
```
matthyx, line 39: **Blocker — the channel-saturation guard described in the PR body doesn't exist.**

The PR says:

> If the context is cancelled **or the error channel is full**, the worker logs a warning and exits cleanly without hanging `sendWG.Wait()`.

There is no `default:` case here, so that second half is not implemented. With a full channel and a live (uncancelled) context this blocks indefinitely, exactly as the old `errorChan <- err` did. Verified with this function verbatim — full `chan error` of cap 1, `context.Background()`:

```
RESULT: sendError STILL BLOCKED after 2s on a full channel with a live ctx
```

So for the scenario #437 actually calls out (`errorChan` buffer fills during widespread backend HTTP 429/500), behaviour is unchanged. Either the code needs a `default:` branch or the PR description and the test name (`...SendErrorNonBlocking`) need to stop claiming non-blocking semantics.

adapters/v1/backend.go
```diff
@@ -58,9 +59,17 @@ func NewBackendAdapter(accountID, apiServerRestURL, eventReceiverRestURL, access
 		},
 		accessKey:             accessKey,
 		securityExceptionRepo: seRepo,
+		httpClient:            &http.Client{},
```
matthyx: **Blocker — the premise behind this change is incorrect, and the change that would actually help isn't made.**

`&http.Client{}` has a nil `Transport`. From `net/http/client.go`:

```go
func (c *Client) transport() RoundTripper {
	if c.Transport != nil {
		return c.Transport
	}
	return DefaultTransport
}
```

`http.DefaultTransport` is a single package-level value with its own connection pool, shared by every client that doesn't set `Transport`. So the old code was **not** creating a new connection pool per call — it allocated a ~small struct and reused the same pool. It cannot cause socket exhaustion or port starvation. (Note also that `postResults` and `ReportScanFailure` already passed `http.DefaultClient`, which is literally `&Client{}` — same pool.) The rationale in both this PR and #437 is wrong on this point, and as written the field is a no-op for pooling.

The change is still worth making — but only if it carries configuration the default doesn't give you. Right now `&http.Client{}` here has:

- **no `Timeout`** — an unbounded request, same as before.
- **`DefaultTransport.MaxIdleConnsPerHost` = 2** (`net/http`'s `DefaultMaxIdleConnsPerHost`). *This* is the real cause of connection churn when many `postResults` goroutines hammer a single event-receiver host: beyond 2 idle connections, sockets are closed rather than reused, producing a pile of `TIME_WAIT` entries that looks a lot like the port starvation reported in #437.

So if the reported symptom is real, this is where to fix it:

```go
httpClient: &http.Client{
	Timeout: 60 * time.Second,
	Transport: &http.Transport{
		Proxy:               http.ProxyFromEnvironment,
		MaxIdleConns:        100,
		MaxIdleConnsPerHost: 32,
		IdleConnTimeout:     90 * time.Second,
	},
},
```

Please also correct the PR description and #437, otherwise the next person reads "unpooled client" as established fact.

adapters/v1/backend_test.go
```diff
@@ -605,3 +605,25 @@ func TestGetCVEExceptions_ScopesCRDByMatch(t *testing.T) {
...
+
+	// Test non-blocking sendError when context is cancelled and channel is full
+	cancelledCtx, cancel := context.WithCancel(context.Background())
+	cancel()
+
+	ch := make(chan error, 1)
+	ch <- fmt.Errorf("initial error")
+
+	// Must not block even when ch is full because ctx is cancelled
+	sendError(cancelledCtx, ch, fmt.Errorf("overflow error"))
```
matthyx: **Blocker — this test asserts around the bugs rather than at them, so it gives false confidence.**

Three problems:

1. `ch := make(chan error, 1)` is pre-filled so the send case is **not** ready, which forces the `<-ctx.Done()` branch deterministically. That is precisely the shape that hides the random-select error loss I flagged in `backend_utils.go`. Use the realistic shape — `make(chan error, 10)` with free space and a cancelled ctx, asserting the error *is* delivered — and this test fails ~50% of runs today.
2. It never tests the non-blocking behaviour its own name claims. The scenario in the PR description (channel full, ctx **live**) is untested — and would hang, which is the actual finding.
3. `assert.Equal(t, customClient, a.getHTTPClient(), ...)` only checks that a getter returns the field it was just handed. It doesn't demonstrate reuse. A test with teeth would drive `ReportError`/`SendStatus`/`postResults` through a stub `httputils.IHttpClient` and assert the same instance served all three calls.

Also: the test bundles two unrelated concerns behind one name, and the file now ends with a stray extra blank line.

matthyx (changes requested):
Thanks for digging into this area — but I don't think this PR is mergeable as-is. I checked out the branch, built it, ran the tests (they pass, including under `-race`), and then tried to reproduce the two problems it claims to fix. Both premises don't hold up, and the `sendError` change introduces a correctness regression that's worse than the bug it's meant to prevent.

Summary of blockers (details inline):

1. **`sendError` silently drops real errors when the context is cancelled.** `select` picks uniformly at random among *ready* cases. With `errChan` buffered at 10 and a cancelled ctx, both cases are usually ready — so roughly half the time the error is discarded and `SubmitCVE` returns `nil` for a scan report that never landed. Measured: 492/1000 drops with a free buffer.
2. **The claimed channel-saturation guard isn't implemented.** There is no `default:` case. On a full channel with a live ctx, `sendError` blocks exactly as before. The PR description says otherwise.
3. **The socket-exhaustion rationale is incorrect.** `&http.Client{}` has a nil `Transport` and therefore uses the package-level `http.DefaultTransport` — one shared connection pool. Per-call `&http.Client{}` allocates a small struct; it does not create new pools and cannot exhaust ports. The change is a no-op for pooling and misses the knob that actually matters.
4. **Scope vs. `Resolved #437`.** Issue #437 lists three items; item 2 (summary/first-chunk ordering) is untouched by this diff, yet the PR is set to auto-close the issue. The commit message on `171cad6` also says "sequence report summary", which isn't in the final diff.
5. **The new test is shaped so it can't catch any of this** — see inline.

### On the deadlock premise

I can't find the path where `errorChan <- err` deadlocks. In `SubmitCVE`, the caller goroutine sits in `for e := range errChan` for the whole lifetime of the send fan-out, draining continuously until the channel is closed. Every `postResults` goroutine has a live consumer, so a send blocks only until the consumer picks it up.

If you *have* observed a hang here, I'd bet on the consumer side rather than the producer side: if `chunksChan` is never closed, `sendVulnerabilities`' `for range chunksChan` never returns, so `close(errorChan)` never runs and `SubmitCVE` hangs on `range errChan` forever. `sendError` doesn't help with that at all. A goroutine dump from the hang would settle it — please attach one if you have it.

### Cancellation is being applied to the wrong thing

`a.httpPostFunc` is `httputils.HttpPostWithRetry`, which takes no `context.Context`. So a cancelled ctx does **not** stop the retrying HTTP POST — the request work still runs to completion. This PR makes cancellation affect only *error reporting*, which is backwards: we keep doing the expensive work and throw away the record that it failed. If cancellation-awareness is the goal, it belongs on the HTTP call (`httputils.HttpPostWithContext` exists), not on the error channel.

### What I'd suggest instead

- Drop the `sendError` ctx guard entirely, or make it non-blocking and lossless in the way that matters: `default:` + log, and prioritise the send. If you want both, check `ctx.Err()` only *after* a non-blocking send attempt fails.
- If the goal is a genuinely better HTTP client, give it a `Timeout` and a `Transport` with a raised `MaxIdleConnsPerHost` — see the inline comment on `backend.go`.
- Split the ordering fix (#437 item 2) into its own PR, or implement it here and keep `Resolved #437`.

Happy to re-review once these are addressed.

bhuvan-somisetty:
Thanks for the thorough dig-in, @matthyx — you were right on all counts. Pushed a new commit that:

- Fixed `sendError` to always win the send when the channel has room (even under a cancelled ctx), only falling back to the ctx.Done()/full-channel race once actually saturated — using pretty much the exact select you sketched.
- Corrected the HTTP client rationale: you're right that `&http.Client{}` isn't creating new pools, it was falling back to `DefaultTransport`'s low `MaxIdleConnsPerHost`. Gave it a real `Timeout` and `Transport` with raised `MaxIdleConnsPerHost`, matching your suggestion.
- Sent the summary report synchronously before the first chunk goroutine, so it now actually resolves item 2 of #437 instead of leaving it untouched — updated the PR body to stop overclaiming.
- Rewrote the test: realistic buffered-channel-with-room case, a separate test that proves the full-channel/live-ctx path actually blocks then drops on cancellation, and driving ReportError/SendStatus/postResults through a recording httputils.IHttpClient stub to prove one instance serves all three.

Ran everything locally (SubmitCVE suite + the new tests, including a 200x repeat of the sendError room-available case) and it's clean. Let me know if anything still looks off.

bhuvan-somisetty:
One more pass to close out the smaller stuff I missed in the first round:

- Removed the stray trailing blank line in `backend_test.go` you flagged.
- Went back and corrected #437 itself (title + body), not just this PR's description, since I'd left the "unpooled client" framing in place there too.
- Put "Resolved #437" back in the PR description — the summary/chunk ordering fix is actually in this diff now, so that's accurate rather than overclaiming.

Everything's pushed and green locally. Appreciate you taking the time to actually run and measure this instead of just reading the diff — that 492/1000 number made it obvious I hadn't tested the failure mode I was claiming to fix.

matthyx (dismissed):
Re-reviewed at `1aa17ed`. All five blockers are genuinely fixed — I re-ran the same measurements rather than reading the diff, and also verified the ordering fix you added, which had no test of its own. LGTM.

## Verification

**1. `sendError` error loss — fixed.** Re-ran my original harness against the new implementation verbatim (cancelled ctx, `cap 10` channel with free buffer):

```
before:  errors silently dropped despite free buffer: 492/1000
after:   errors silently dropped despite free buffer: 0/100000
```

The non-blocking send now wins deterministically. I also checked the drain path: on a full channel the send is *parked*, not discarded — as soon as the channel drains, the error is delivered (`PROOF 2b delivered after drain: second`). That's the lossless behaviour, and it's the right call given `SubmitCVE` drains continuously.

**2. Saturation semantics — now honest.** `TestSendError_BlocksOnFullChannelUntilCtxCancelledThenDrops` asserts what the code actually does instead of what the description wished it did, and the PR body no longer claims non-blocking-on-full. Good.

**3. HTTP client — correct now, and the rationale is fixed everywhere.** `Timeout: 60s` plus `MaxIdleConnsPerHost: 32` is the change that actually addresses the reported symptom. Thanks for going back and correcting #437's title and body too — that matters more than the PR description, since the issue is what people will find later.

**4. Tests — real coverage.** `recordingHTTPClient` driving `ReportError`/`SendStatus`/`postResults` and asserting `assert.Same` is exactly the shape that proves reuse. Full `./adapters/v1/...` suite passes under `-race` (128s). `go vet` clean. The two `gofmt` complaints on `backend.go`/`backend_test.go` are pre-existing on `main` (struct alignment and import ordering) — not yours, and I'd leave them out of this PR.

**5. Ordering fix — independently verified.** This one arrived after my review and shipped without a test, so I wrote a throwaway one: stub `httpPostFunc` that records `PaginationInfo.ReportNumber` per call and sleeps 150ms on the summary, so an unordered implementation loses the race.

```
pre-fix  (cb3e7f8):  observed send order: [1 0]   <- chunk before summary, exactly the #437 400
post-fix (1aa17ed):  observed send order: [0 1]   <- summary first
```

So the fix works and the bug was real. **Please consider landing a version of that test** — the ordering guarantee is the one behaviour here with no regression protection, and it's the easiest of the three to silently undo later (anyone converting that call back to `postResultsAsGoroutine` for latency reasons reintroduces it with a green suite). Happy to open it as a follow-up PR if you'd rather not expand this one.

## Non-blocking follow-ups

None of these should hold up the merge.

- **A failed summary still dispatches every chunk.** `postResults` returns nothing, so `sendSummaryAndVulnerabilities` proceeds unconditionally. If the summary POST fails permanently, the chunks still go out against a scan the backend has no summary for — producing the exact `400 missing summary for containerScanID` that #437 is about. Ordering is now guaranteed; delivery isn't. Now that this call is synchronous it's cheap to close: have `postResults` return an `error` (it has exactly one synchronous caller) and skip the chunk dispatch when the summary fails.

- **The summary POST is now on `SubmitCVE`'s critical path.** `httpPostFunc` is `HttpPostWithRetry` with `maxElapsedTime: 60s`, so a flapping event receiver can now stall `SubmitCVE` for up to a minute before any chunk is dispatched, where it previously overlapped. Correct trade for ordering, but worth knowing when you next look at scan latency.

- **`sendError`'s blocking branch now has one caller with no concurrent drainer.** The synchronous summary post runs before `for e := range errChan` starts. It's provably safe today — `errChan` is empty at that point and only one error can arrive — but the margin is implicit. If anyone later shrinks the buffer or adds a pre-consumer writer, that path blocks forever with a live ctx and deadlocks `SubmitCVE`: the exact failure class this PR set out to remove. A one-line comment on `errChan := make(chan error, 10)` noting that the buffer must exceed the number of pre-consumer writers would save someone a bad afternoon.

- Separately, and outside this PR: `HttpPostWithRetry` calls `HttpPostWithContext(context.Background(), ...)`, so scan-context cancellation still can't abort an in-flight report. Worth a ticket of its own.

Nice work turning this around — corrected upstream issue included.

matthyx:
please rebase

matthyx (changes requested):
Re-checked after the rebase at `adb6496`. **The rebase dropped one of the fixes** — everything else survived intact and re-verifies clean.

## Blocker: `ReportScanFailure` lost the shared client

`adapters/v1/backend.go:253` is back to `http.DefaultClient`:

```go
resp, err := a.httpPostFunc(http.DefaultClient, url, a.getRequestHeaders(), payload, 30*time.Second)
```

Pre-rebase (`1aa17ed:205`) that line was `a.getHTTPClient()`. The hunk is simply absent from the post-rebase diff — a conflict resolution took `main`'s side. So `ReportScanFailure` is the one call site still running on the untuned default: no `Timeout`, and `MaxIdleConnsPerHost` back to 2. Three of four call sites use the shared client; this one doesn't.

Fix is the one line:

```go
resp, err := a.httpPostFunc(a.getHTTPClient(), url, a.getRequestHeaders(), payload, 30*time.Second)
```

I verified it both ways — a test asserting the client identity fails at `adb6496` and passes once that line is restored.

### Why CI didn't catch it

There's no coverage of *which* client `ReportScanFailure` uses. `TestBackendAdapter_HTTPClientReuse` exercises `ReportError`/`SendStatus`/`postResults` but not `ReportScanFailure`, and all six existing `TestBackendAdapter_ReportScanFailure_*` tests discard the client argument (`mockHTTP := func(_ httputils.IHttpClient, ...)`). So the revert was invisible to a green suite — which is exactly how it survived the rebase.

Worth extending `TestBackendAdapter_HTTPClientReuse` to cover the fourth call site, so this can't silently regress again:

```go
var gotClient httputils.IHttpClient
a.httpPostFunc = func(c httputils.IHttpClient, _ string, _ map[string]string, _ []byte, _ time.Duration) (*http.Response, error) {
	gotClient = c
	return &http.Response{StatusCode: 200, Body: io.NopCloser(bytes.NewBuffer(nil))}, nil
}
require.NoError(t, a.ReportScanFailure(ctx, scanfailure.ScanFailureSBOMGeneration, scanfailure.ReasonSBOMGenerationFailed, nil))
assert.Same(t, stub, gotClient, "ReportScanFailure must use the shared httpClient")
```

Also update the PR description while you're there — it lists reuse "across `ReportError`, `SendStatus`, and `postResults`", which omits `ReportScanFailure` and so currently matches the regressed code rather than the intent.

## Everything else survived — re-verified at `adb6496`

- **`sendError`** — byte-for-byte identical to the version I approved (diffed the function directly, not the surrounding file).
- **Summary-before-chunk ordering** — intact. Re-ran my ordering harness against the rebased head: `observed send order: [0 1]`. Still correct.
- **All three new tests present**: `TestBackendAdapter_HTTPClientReuse`, `TestSendError_DeliversImmediatelyWhenChannelHasRoom`, `TestSendError_BlocksOnFullChannelUntilCtxCancelledThenDrops`.
- **Client config** — `Timeout: 60s` / `MaxIdleConnsPerHost: 32` intact in `NewBackendAdapter`.
- `go build ./...` clean, `go vet ./adapters/v1/...` clean, full `./adapters/v1/...` suite green under `-race` (122s).
- The rebase picked up `main`'s new `exceptionsCache` work (#439) and integrated cleanly alongside the `httpClient` field — no interaction problems that I can see.

The non-blocking follow-ups from my previous review still stand unchanged (failed summary still dispatches every chunk; summary POST on `SubmitCVE`'s critical path; `HttpPostWithRetry` still uses `context.Background()`). None of those need to block this.

Restore that one line and I'll re-approve.

bhuvan-somisetty:
Rebased onto main. There was a real conflict in backend.go — the exceptions-cache change that landed on main touched the same struct/constructor as this PR, plus SendStatus itself had drifted (unpooled client, but correctly using statuses[step]) — so I merged both sides by hand instead of just taking one, keeping the cache field/init from main and the tuned client + summary-ordering fix from here. Rebuilt, reran the full adapters/v1 suite including the new exceptions-cache tests from main, all green, and force-pushed.

CI's all passing now too. Let me know if anything looks off on a fresh pass.

bhuvan-somisetty:
@matthyx good catch — you're right, the conflict resolution took main's side of that hunk and quietly reverted `ReportScanFailure` to `http.DefaultClient`. Restored the one line.

Also extended `TestBackendAdapter_HTTPClientReuse` to drive `ReportScanFailure` through the recording stub too, so all four call sites are now asserted, not just three — this exact regression would fail loudly next time instead of slipping through green. Updated the PR description to list it as well.

Pushed at a009ac5, CI's all green. Appreciate you re-diffing the function directly instead of trusting the rebase — that's exactly the kind of thing a "changes can be cleanly merged" banner doesn't catch.

matthyx (approved):
Re-checked at `a009ac5`. The rebase regression is fixed and the rest of the PR is fully intact. **LGTM** — re-approving.

## Verification

**The fix.** `ReportScanFailure` is back on `a.getHTTPClient()`. All four call sites now share the tuned client (3 in `backend.go`, 1 in `backend_utils.go`), and no `&http.Client{}` or bare `http.DefaultClient` remains at any of them.

**Nothing else was lost in the rebase.** Rather than spot-checking the fixed line, I re-verified every behaviour this PR is supposed to introduce against the head tree — the `httpClient` field; all four `Transport` settings (`Timeout: 60s`, `MaxIdleConns: 100`, `MaxIdleConnsPerHost: 32`, `IdleConnTimeout: 90s`); the `getHTTPClient()` helper; all four call-site conversions; `sendError`'s non-blocking-send-first shape; every `errorChan <- err` converted with no raw sends left; the synchronous summary post; and all three tests plus `recordingHTTPClient`. 18/18 present.

**Behaviour re-measured, not assumed:**

- `sendError` is byte-for-byte identical to the version I approved pre-rebase (diffed the function itself).
- Error loss under a cancelled ctx with a free buffer: **0/20000**. (Was 492/1000 on the original implementation.)
- Summary-before-chunk ordering: `observed send order: [0 1]`.
- `ReportScanFailure` client identity: asserted `assert.Same` against the adapter's stub — passes.

`go build ./...` clean, `go vet ./adapters/v1/...` clean, full `./adapters/v1/...` suite green under `-race` (122s).

**Your new assertion does have teeth** — I reverted the line locally and `TestBackendAdapter_HTTPClientReuse` fails, so this can't silently regress again. Thanks also for correcting the PR description to say all four call sites.

## One small suggestion (non-blocking, take it or leave it)

The way that test fails is noisier than it needs to be. With the wrong client, `ReportScanFailure` makes a *real* outbound call to `http://example.invalid` and the test dies on `require.NoError` with a DNS error after ~20s of retry backoff — before ever reaching `assert.Equal(t, 4, stub.Calls(), ...)`. So the signal is "dial tcp: no such host", not "wrong HTTP client", and it costs 20 seconds.

Capturing the client identity instead fails in 0.00s with the actual reason:

```go
var gotClient httputils.IHttpClient
a.httpPostFunc = func(c httputils.IHttpClient, _ string, _ map[string]string, _ []byte, _ time.Duration) (*http.Response, error) {
	gotClient = c
	return &http.Response{StatusCode: 200, Body: io.NopCloser(bytes.NewBuffer(nil))}, nil
}
require.NoError(t, a.ReportScanFailure(ctx, scanfailure.ScanFailureSBOMGeneration, scanfailure.ReasonSBOMGenerationFailed, nil))
assert.Same(t, stub, gotClient, "ReportScanFailure must use the shared httpClient")
```

Purely a diagnostics improvement — the current test is correct and does the job.

The three non-blocking follow-ups from earlier still stand and still shouldn't block this: a failed summary still dispatches every chunk; the summary POST now sits on `SubmitCVE`'s critical path with a 60s retry budget; and `HttpPostWithRetry` still uses `context.Background()`, so scan cancellation can't abort an in-flight report. Worth their own issues after this lands.

Good catch turnaround on the rebase drop.

Issue #436, opened 2026-08-03
Improvement: ctx-propagation tests cover only the first happy-path call — two sub-clients and every retry path are unasserted


matthyx:
## Overview

#422 propagated the caller's `ctx` to every `APIServerStore` storage call and added `_ctxPropagated` tests built on `ctxCapturing*` sub-client wrappers in `repositories/apiserver_test.go`. Those tests are sound — backing the production change out makes all six of them fail — but they only cover the *first* call each method makes on its happy path. Several storage paths that do pass `ctx` correctly have no test asserting it.

## Problem

Three gaps, all in `repositories/apiserver_test.go`:

**1. Two sub-clients have no wrapper at all.** Wrappers exist for `VulnerabilityManifests` (line 1486), `SBOMSyfts` (1506), `OpenVulnerabilityExchangeContainers` (1526) and `VulnerabilityManifestSummaries` (1546). There is none for `SBOMSyftFiltereds` or `ContainerProfiles`, so these call sites are unasserted:

- `GetContainerProfile` — `repositories/apiserver.go:207`
- `StoreSBOM(..., isFiltered: true)` — `repositories/apiserver.go:1205`, `:1214`, `:1223`

**2. Two `Store*` methods have no ctx test.** `StoreCVESummary` (`apiserver.go:571`, `:580`, `:589`) and `StoreCVESummaryStub` (`:659`, `:667`, `:678`) both go through `VulnerabilityManifestSummaries`, which *is* already wrapped — only the tests are missing. Worth noting these call it with the **workload** namespace rather than `a.Namespace`.

**3. No conflict-retry path is covered.** All seven existing assertions target `getCtx` or `createCtx` (lines 1623, 1631, 1647, 1674, 1682, 1691, 1692). `updateCtx` is captured by three wrappers and asserted by nothing, so the `Get`/`Update` inside all six `RetryOnConflict` closures (`apiserver.go:333`, `:574`, `:662`, `:742`, `:1208`, `:1245`) has zero ctx coverage — including the `ctx.Err()` guard at the top of each, which no test exercises.

Gap 3 is the one most worth closing. Every existing test hits a fresh object, so the `AlreadyExists` → retry branch — where the `ctx` plumbing is densest and easiest to break — is never entered with a ctx assertion in place.

## Solution

[...]

PR #442, opened 2026-08-03
test: key ctxCapturing sub-client wrappers by namespace

### LINKED ISSUE #435: Improvement: ctxCapturing test wrappers memoize without keying by namespace
## Overview

#422 added `ctxCapturing*` sub-client wrappers to `repositories/apiserver_test.go` so the `_ctxPropagated` tests can assert that the caller's `context.Context` actually reaches the storage client. A follow-up commit on that PR (`ea43cfd`) keyed the `VulnerabilityManifestSummaries` wrapper by namespace; the other three were left as single-instance memoization.

## Problem

`ctxCapturingClient` (line 1559) memoizes three of its four wrappers on a single field, ignoring the `namespace` argument it was called with:

```go
func (c *ctxCapturingClient) VulnerabilityManifests(namespace string) spdxv1beta1.VulnerabilityManifestInterface {
	if c.vulnManifests == nil {
		c.vulnManifests = &ctxCapturingVulnerabilityManifests{VulnerabilityManifestInterface: c.SpdxV1beta1Interface.VulnerabilityManifests(namespace)}
	}
	return c.vulnManifests
}
```

Same shape for `SBOMSyfts` (line 1579) and `OpenVulnerabilityExchangeContainers` (line 1586). Each wrapper binds to whichever namespace was requested first and then silently returns that sub-client for every subsequent namespace.

This is harmless today: those three are only ever called with `a.Namespace`. It stops being harmless the moment a wrapper is added for `ContainerProfiles`, because `GetContainerProfile(ctx, namespace, name)` takes a **caller-supplied** namespace (`repositories/apiserver.go:200` and `:207`, called from `adapters/v1/container_profile.go:29`) that is unrelated to `a.Namespace`. A test written against the wrong sub-client would still pass, which defeats the whole point of the helper — these wrappers exist precisely so that a ctx assertion cannot succeed vacuously.

## Solution

Key all four wrappers by namespace, matching what `VulnerabilityManifestSummaries` already does (line 1593):

```go
type ctxCapturingClient struct {
	spdxv1beta1.SpdxV1beta1Interface
	vulnManifests map[string]*ctxCapturingVulnerabilityManifests
	// ...
}

func (c *ctxCapturingClient) VulnerabilityManifests(namespace string) spdxv1beta1.VulnerabilityManifestInterface {
	if c.vulnManifests == nil {
		c.vulnManifests = map[string]*ctxCapturingVulnerabilityManifests{}
	}
	if _, ok := c.vulnManifests[namespace]; !ok {
		c.vulnManifests[namespace] = &ctxCapturingVulnerabilityManifests{
			VulnerabilityManifestInterface: c.SpdxV1beta1Interface.VulnerabilityManifests(namespace),
		}
	}
	return c.vulnManifests[namespace]
}
```

Existing assertions become `wrapped.vulnManifests[a.Namespace].getCtx`, consistent with the summaries wrapper. Memoization itself must stay — the comment above the accessors explains why (a single storage operation calls the accessor more than once, e.g. `StoreVEX` for its `Get` and then again for its `Create`/`Update`, so replacing the wrapper per call would clobber earlier captures).

## Alternatives

Leave the three as-is and only key wrappers whose sub-client can see more than one namespace. That works but makes the helper inconsistent, and the inconsistency is invisible at the call site — the next person to add a wrapper has to rediscover which pattern applies.

## Additional context

Verified on `main` at `33f0f67`.

Minor, and not worth acting on by itself: `ea43cfd`'s commit message states the keying was applied "as with the other wrappers", but the other three are still unkeyed. Nothing to fix in history — noting it only so the message isn't taken at face value by whoever picks this up.

NaitikRishu:
Key `vulnManifests`, `sbomSyfts`, and `ovecs` wrappers by namespace in `ctxCapturingClient`, matching the existing `VulnerabilityManifestSummaries` pattern. This prevents a future wrapper (e.g. `ContainerProfiles`) from silently binding to the wrong namespace's sub-client.

Existing assertions are updated to use `wrapped.field[a.Namespace].ctx` syntax, consistent with the summaries wrapper.

Fixes #435

## Overview

**Current behavior:** Three of the four `ctxCapturing*` sub-client wrappers (`vulnManifests`, `sbomSyfts`, `ovecs`) memoize on a single pointer field, ignoring the `namespace` argument. They bind to whichever namespace is requested first and silently return that same sub-client for every subsequent namespace. As noted in #435, this means a wrapper added for `ContainerProfiles` — where `GetContainerProfile(ctx, namespace, name)` takes a caller-supplied namespace unrelated to `a.Namespace` — would let a ctx assertion pass vacuously against the wrong sub-client.

**New behavior:** All three wrappers are keyed by namespace using `map[string]*ctxCapturing...`, matching the pattern `VulnerabilityManifestSummaries` already uses. Each accessor lazy-inits the map and keys by the `namespace` argument. Memoization is preserved — a single storage operation that calls the accessor more than once (e.g. `StoreVEX` for its `Get` and then again for its `Update`) still gets the same wrapper instance.

## How to Test

```bash
go test -v ./repositories -run ctxPropagated
```

All 7 existing `_ctxPropagated` tests pass with the updated assertions.

## Related issues/PRs:

* Resolved #435
* Prerequisite for #436

## Summary by CodeRabbit

* **Tests**
  * Improved test coverage for context propagation across CVE, SBOM, and VEX operations.
  * Added namespace-specific context tracking for retrieval, creation, and update scenarios.

### DIFF
--- a/repositories/apiserver_test.go
+++ b/repositories/apiserver_test.go
@@ -1558,9 +1558,9 @@ func (w *ctxCapturingVulnerabilityManifestSummaries) Get(ctx context.Context, na
 // everything else untouched.
 type ctxCapturingClient struct {
 	spdxv1beta1.SpdxV1beta1Interface
-	vulnManifests *ctxCapturingVulnerabilityManifests
-	sbomSyfts     *ctxCapturingSBOMSyfts
-	ovecs         *ctxCapturingOVECs
+	vulnManifests map[string]*ctxCapturingVulnerabilityManifests
+	sbomSyfts     map[string]*ctxCapturingSBOMSyfts
+	ovecs         map[string]*ctxCapturingOVECs
 	vulnSummaries map[string]*ctxCapturingVulnerabilityManifestSummaries
 }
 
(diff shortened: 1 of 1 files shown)


NaitikRishu:
Hi @matthyx, let me know if anything needs changing here — I have the follow-up for #436 ready to go once this lands.

matthyx (approved):
LGTM — this does exactly what #435 asked for, and nothing else.

I checked it out and verified locally:

- `go test ./repositories/ -count=1` — full package green; the 7 `_ctxPropagated` tests plus the new one pass.
- `go vet ./repositories/` and `gofmt -l` — clean.
- Mutation check on the new regression test: I temporarily reverted `SBOMSyfts` to first-namespace-wins memoization, and `TestCtxCapturingClient_wrappersKeyedByNamespace` failed on the `require.NotSame` for `ns-a` vs `ns-b`. So the test actually pins the behavior it claims to, rather than passing vacuously — which was the whole point of #435.

All four accessors now follow one shape, and the memoization comment above them still holds (`StoreVEX` hits the OVEC accessor twice and both captures survive — the `StoreVEX_ctxPropagated` assertions on `getCtx` *and* `createCtx` prove it).

No blockers. Two non-blocking nits, take or leave:

1. The assertions index the map directly (`wrapped.vulnManifests[a.Namespace].getCtx`). If a future refactor stops the accessor from firing at all, the missing key gives a nil-pointer panic instead of a readable assertion failure. A `require.Contains(t, wrapped.vulnManifests, a.Namespace)` first would make that failure mode legible. Same for the other five sites.
2. The PR body says to test with `go test -v ./repositories -run ctxPropagated`, which doesn't match `TestCtxCapturingClient_wrappersKeyedByNamespace`. Worth widening to `-run 'ctxPropagated|CtxCapturingClient'` so the new test isn't silently skipped by anyone following the instructions.

Approving. Go ahead with the #436 follow-up.

PR #445, opened 2026-08-03
fix: classify gRPC codes.DeadlineExceeded as ReasonScanTimeout

### LINKED ISSUE #441: fix: classify gRPC codes.DeadlineExceeded as ReasonScanTimeout instead of generic SBOM failure
## Description

`CreateSBOM` in `pkg/sbomscanner/v1/client.go` (lines 116–122) wraps two gRPC error codes into the sidecar-crashed sentinel, but lets `codes.DeadlineExceeded` bubble up as a raw gRPC `statusError`:

```go
resp, err := c.client.CreateSBOM(ctx, pbReq)
if err != nil {
    st, ok := status.FromError(err)
    if ok && (st.Code() == codes.Unavailable || st.Code() == codes.Aborted) {
        return nil, fmt.Errorf("%w: %v", ErrScannerCrashed, err)
    }
    return nil, err    // ← codes.DeadlineExceeded returns raw, unclassified
}
```

That raw error flows into `classifySBOMError` (`core/services/scan_failure_reasons.go:29`), which detects timeouts with:

```go
if errors.Is(err, context.DeadlineExceeded) {
}
```

`errors.Is(rawGrpcStatusErr, context.DeadlineExceeded)` evaluates to `false` — a gRPC status error carries a `codes.DeadlineExceeded` code, but does not wrap the stdlib sentinel. It also matches none of the string-based arms (which check 401/403, `MANIFEST_UNKNOWN`, etc.). A scan timeout that reaches the client through the sidecar therefore falls through to the generic `ReasonSBOMGenerationFailed` instead of `ReasonScanTimeout`.

**Why this matters:** `ReasonScanTimeout` exists in the `scanfailure` taxonomy specifically so scan-timeout alerts carry accurate signal. The sidecar is the production memory-isolated deployment (selected via `SBOM_SCANNER_SOCKET`, `cmd/http/main.go:73-86`), and this is the one gRPC error code the client does not wrap — the sibling of `ErrScannerCrashed` wrapping at `client.go:119-121`. Completing the wrap keeps the taxonomy correct regardless of how the client is called.

**Scope note:** In the current HTTP path, the caller's `ctx` is derived via `context.WithoutCancel` (`controllers/http.go:65,122,169,249`), so this code path is latent today — the sidecar's own timeout surfaces as `Incomplete` status, which `classifySBOMStatus` already handles. It becomes reachable for any caller that passes a deadline-bound `ctx` (future non-HTTP entrypoints, direct gRPC use). The fix completes the wrapping taxonomy so classification holds regardless of caller.

---

## Environment

- **OS:** Any
- **Version:** `main` (Aug 2026)

---

## Steps To Reproduce

1. In `pkg/sbomscanner/v1/client.go`, read `CreateSBOM` lines 116–122 — confirm `codes.DeadlineExceeded` is not wrapped, unlike `Unavailable`/`Aborted`.
2. In `core/services/scan_failure_reasons.go`, read `classifySBOMError` line 29 — observe the `errors.Is(err, context.DeadlineExceeded)` check that never matches a raw gRPC status error.
3. In `core/services/scan.go`, confirm `classifySBOMError`'s return value is what gets reported to the platform for sidecar scans.

---

## Expected Behavior

A sidecar scan that times out on the caller's deadline should be reported as `scanfailure.ReasonScanTimeout`, matching the in-process timeout classification.

---

## Actual Behavior

The timeout is reported as the generic `ReasonSBOMGenerationFailed`, so operators cannot distinguish "scan exceeded its deadline" from "SBOM generation genuinely failed" in the `failureReason` signal.

---

## Proposed Solution

Mirror the `ErrScannerCrashed` wrapping pattern so `errors.Is(err, context.DeadlineExceeded)` fires correctly:

```go
if ok && (st.Code() == codes.Unavailable || st.Code() == codes.Aborted) {
    return nil, fmt.Errorf("%w: %v", ErrScannerCrashed, err)
}
if ok && st.Code() == codes.DeadlineExceeded {
    return nil, fmt.Errorf("scan timeout: %w: %v", context.DeadlineExceeded, err)
}
return nil, err
```

The `%v` preserves the original gRPC error for debugging while `%w` keeps `errors.Is` working.

**Regression test:** Add a mock-based test in a new `pkg/sbomscanner/v1/client_test.go` that makes the gRPC client return `status.Error(codes.DeadlineExceeded, ...)` and asserts `errors.Is(err, context.DeadlineExceeded)` (and that it stays distinct from `ErrScannerCrashed`). `classifySBOMError` is unchanged — its existing test case (lines 38–41) already covers the `%w`-wrapped sentinel.

**Files to change:**
- `pkg/sbomscanner/v1/client.go` — add the `DeadlineExceeded` wrap
- `pkg/sbomscanner/v1/client_test.go` — new, mock-based regression tests

---

## Additional Context

This is the CVE-side taxonomy's missing half: PRs #395, #396, and #399 made silent failures observable, and the `scanfailure` taxonomy was designed exactly for this distinction — the SBOM side already separates `ReasonScannerOOMKilled`, `ReasonScanTimeout`, etc. The sidecar's `ErrScannerCrashed` wrap at `client.go:119-121` is proof of the intended pattern; `DeadlineExceeded` is the one code left out. I can open a PR with the fix and the regression tests if this looks good. If you'd prefer the wrapping done at a different layer (e.g., in `classifySBOMError` via `status.FromError`), I'm happy to follow your guidance.

AdeshDeshmukh:
## Overview

`sbomScannerClient.CreateSBOM` wraps gRPC `Unavailable`/`Aborted` errors as `ErrScannerCrashed` but lets `codes.DeadlineExceeded` escape as a raw gRPC status error. That raw error never satisfies `errors.Is(err, context.DeadlineExceeded)` in `classifySBOMError` (`core/services/scan_failure_reasons.go:29`), so a sidecar scan that times out on the caller's deadline is reported as `ReasonSBOMGenerationFailed` instead of `ReasonScanTimeout`.

This adds the missing sibling wrap so the `scanfailure` taxonomy holds regardless of caller. See #441 for the full analysis and scope note.

## How to Test

```bash
TMPDIR=/tmp go test ./pkg/sbomscanner/v1/ -run 'TestCreateSBOM_' -v
```

- `TestCreateSBOM_WrapsGRPCDeadlineExceeded` - asserts `errors.Is(err, context.DeadlineExceeded)` and that it stays distinct from `ErrScannerCrashed`.
- `TestCreateSBOM_WrapsGRPCCrashCodes` - confirms `Unavailable`/`Aborted` behavior is unchanged.
- `TestCreateSBOM_PassesThroughPlainDeadlineExceeded` / `TestCreateSBOM_PassesThroughUnrelatedError` - confirms no double-wrapping and pass-through of unrelated errors.

Note: `TMPDIR=/tmp` is only needed locally (this machine's temp path exceeds the Unix-socket 104-byte limit - a pre-existing, environment-only issue affecting the package's other socket tests). CI uses `/tmp`, so it's green there.

## Files changed

- `pkg/sbomscanner/v1/client.go` - add `codes.DeadlineExceeded` wrap (+3 lines)
- `pkg/sbomscanner/v1/client_test.go` - new, mock-based regression tests

Resolves #441

### DIFF
--- a/pkg/sbomscanner/v1/client.go
+++ b/pkg/sbomscanner/v1/client.go
@@ -101,6 +101,9 @@ func (c *sbomScannerClient) CreateSBOM(ctx context.Context, req ScanRequest) (*S
 		if ok && (st.Code() == codes.Unavailable || st.Code() == codes.Aborted) {
 			return nil, fmt.Errorf("%w: %v", ErrScannerCrashed, err)
 		}
+		if ok && st.Code() == codes.DeadlineExceeded {
+			return nil, fmt.Errorf("scan timeout: %w: %v", context.DeadlineExceeded, err)
+		}
 		return nil, err
 	}
 
--- a/pkg/sbomscanner/v1/client_test.go
+++ b/pkg/sbomscanner/v1/client_test.go
@@ -1,0 +1,67 @@
+package v1
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	pb "github.com/kubescape/kubevuln/pkg/sbomscanner/v1/proto"
+	"github.com/stretchr/testify/assert"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+type mockPBScannerClient struct {
+	createSBOMErr error
+}
+
+func (m *mockPBScannerClient) CreateSBOM(_ context.Context, _ *pb.CreateSBOMRequest, _ ...grpc.CallOption) (*pb.CreateSBOMResponse, error) {
+	return nil, m.createSBOMErr
+}
+
+func (m *mockPBScannerClient) Health(_ context.Context, _ *pb.HealthRequest, _ ...grpc.CallOption) (*pb.HealthResponse, error) {
+	return &pb.HealthResponse{Ready: true}, nil
+}
+
+func TestCreateSBOM_WrapsGRPCDeadlineExceeded(t *testing.T) {
+	c := &sbomScannerClient{client: &mockPBScannerClient{createSBOMErr: status.Error(codes.DeadlineExceeded, "context deadline exceeded")}}
+
+	_, err := c.CreateSBOM(context.Background(), ScanRequest{})
+	assert.Error(t, err)
+	assert.ErrorIs(t, err, context.DeadlineExceeded)
+	assert.NotErrorIs(t, err, ErrScannerCrashed)
+}
+
+func TestCreateSBOM_PassesThroughPlainDeadlineExceeded(t *testing.T) {
+	c := &sbomScannerClient{client: &mockPBScannerClient{createSBOMErr: context.DeadlineExceeded}}
+
+	_, err := c.CreateSBOM(context.Background(), ScanRequest{})
+	assert.Error(t, err)
+	assert.ErrorIs(t, err, context.DeadlineExceeded)
+	assert.NotErrorIs(t, err, ErrScannerCrashed)
+}
+
+func TestCreateSBOM_WrapsGRPCCrashCodes(t *testing.T) {
+	for _, code := range []codes.Code{codes.Unavailable, codes.Aborted} {
+		c := &sbomScannerClient{client: &mockPBScannerClient{
+			createSBOMErr: status.Error(code, "scanner crashed"),
+		}}
+		_, err := c.CreateSBOM(context.Background(), ScanRequest{})
+		assert.Error(t, err)
+		assert.ErrorIs(t, err, ErrScannerCrashed)
+		assert.NotErrorIs(t, err, context.DeadlineExceeded)
+	}
+}
+
+func TestCreateSBOM_PassesThroughUnrelatedError(t *testing.T) {
+	unrelated := errors.New("some other error")
+	c := &sbomScannerClient{client: &mockPBScannerClient{
+		createSBOMErr: unrelated,
+	}}
+
+	_, err := c.CreateSBOM(context.Background(), ScanRequest{})
+	assert.ErrorIs(t, err, unrelated)
+	assert.NotErrorIs(t, err, context.DeadlineExceeded)
+	assert.NotErrorIs(t, err, ErrScannerCrashed)
+}


AdeshDeshmukh:
@matthyx PTAL

matthyx (approved):
LGTM — approving. No blockers.

The fix is correct and minimal: `status.FromError` returns `ok=true` only for real gRPC status errors, so a plain `context.DeadlineExceeded` still passes through unwrapped (no double-wrap), and `%w` on the stdlib sentinel makes `errors.Is` fire in `classifySBOMError`, giving `ReasonScanTimeout` instead of `ReasonSBOMGenerationFailed`. Placement after the `Unavailable`/`Aborted` arm keeps the crash path untouched, so `SidecarSBOMAdapter.handleCrash` retry accounting is unaffected.

Verified locally on `d111724`:

- `go build ./...` clean, `go vet ./pkg/sbomscanner/... ./core/services/...` clean.
- `TMPDIR=/tmp go test ./pkg/sbomscanner/v1/` — whole package green, including the 4 new tests.
- `go test ./core/services/... ./adapters/` — green. (`adapters/v1` has 3 failures in `Test_syftAdapter_CreateSBOM`, all `unexpected EOF` while pulling from quay.io — network-dependent and unrelated to this change.)
- Confirmed no other caller in the repo branches on `errors.Is(err, context.DeadlineExceeded)`, so widening what matches that sentinel changes only the classification path. `sbomscanner/v1/server.go` never returns a `DeadlineExceeded` status (its own Syft deadline surfaces as `Status: Incomplete`, nil error), so this wrap can only ever mean "caller's deadline expired" — exactly `ReasonScanTimeout` semantics.

Two optional nits, neither worth another round:

1. `fmt.Errorf("scan timeout: %w: %v", context.DeadlineExceeded, err)` renders as `scan timeout: context deadline exceeded: rpc error: code = DeadlineExceeded desc = context deadline exceeded` — the phrase appears three times. Dropping the `scan timeout: ` prefix (matching the `ErrScannerCrashed` arm's `"%w: %v"`) would read better in logs.
2. The `%v` flattens the gRPC status into text, so `status.FromError` can no longer recover the code from the returned error. Nothing consumes it today, and it matches the existing crash arm, so this is consistent rather than a regression — just noting it in case a future caller wants the code back.

Also worth restating from #441 for anyone reading later: with the HTTP entrypoint deriving ctx via `context.WithoutCancel` (`controllers/http.go`), this path is latent today. That's fine — it closes the taxonomy gap for any deadline-bound caller, and the tests lock the behavior in.

PR #448, opened 2026-08-04
perf: index dynamic paths in filterSBOM by segment count

### LINKED ISSUE #447: Index dynamic paths in filterSBOM to avoid O(files × dynamic-paths) relevancy filtering
## Overview
`filterSBOM` in `core/services/scan.go` filters an SBOM's file list down to the files considered "relevant" (from runtime observation) plus files that match a "dynamic path" pattern (paths containing the `dynamicpathdetector.DynamicIdentifier` placeholder, e.g. paths with variable segments like temp/session directories).

## Problem
For every SBOM file that isn't an exact match in `relevantFiles`, the current fallback loop calls `dynamicpathdetector.CompareDynamic` against every single collected dynamic path:

```go
var dynamicPaths []string
relevantFiles.Each(func(file string) bool {
    if strings.Contains(file, dynamicpathdetector.DynamicIdentifier) {
        dynamicPaths = append(dynamicPaths, file)
    }
    return false
})

for _, f := range sbom.Content.Files {
    ...
    for _, dynamicPath := range dynamicPaths {
        if dynamicpathdetector.CompareDynamic(dynamicPath, f.Location.RealPath) {
            ...
        }
    }
}
```

This is an O(files × dynamicPaths) comparison. For images with large file counts and workloads that accumulate many dynamic paths, this becomes a real, currently-unmeasured cost on the relevancy-filtering hot path, which runs inside the scan worker pool.

Looking at `dynamicpathdetector.CompareDynamic` (`github.com/kubescape/storage/pkg/registry/file/dynamicpathdetector`), a match additionally requires both paths to have the exact same number of `/`-separated segments — the comparison loop only succeeds if both the dynamic path and the candidate path reach their end at the same time. This means the vast majority of `(dynamicPath, file)` pairs compared today can never match and are wasted work.

## Solution
Pre-index the collected `dynamicPaths` by segment count (`strings.Count(path, "/")`) when building them, and when checking a candidate file, only run `CompareDynamic` against the dynamic paths whose segment count matches the candidate file's own segment count. This preserves identical matching behavior (since a differing segment count can never match) while pruning the vast majority of comparisons.

## Alternatives
- Indexing by static path prefix instead of segment count was also considered; segment count is simpler, requires no extra parsing, and is guaranteed correct given `CompareDynamic`'s segment-count requirement.
- Leaving the linear scan in place, only adding a benchmark to measure it, was considered but does not address the actual scaling cost.

## Additional context
- Affected function: `filterSBOM` in `core/services/scan.go` (relevancy-filtering fallback loop).
- No new external dependency is required; the indexing can be implemented with the standard library.
- Existing tests in `core/services/scan_test.go` (`Test_filterSBOM`) should continue to pass unchanged since matching behavior is preserved.

rootp1:
## Overview
`filterSBOM` in `core/services/scan.go` falls back to a linear scan over all collected `dynamicPaths` for every SBOM file that isn't an exact match in `relevantFiles`, calling `dynamicpathdetector.CompareDynamic` for each pair. This is O(files x dynamicPaths). `CompareDynamic` only ever matches paths with an identical number of `/`-separated segments, so most of those comparisons can never succeed.

This PR indexes the collected dynamic paths by segment count, so the per-file fallback loop only compares against the bucket of dynamic paths with a matching segment count, instead of the full set. Matching behavior is unchanged since a differing segment count can never match under `CompareDynamic`.

## Related issues/PRs:
Resolved #447

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SBOM path filtering by matching dynamic paths with the same directory depth, reducing incorrect matches while preserving direct path matching.

* **Tests**
  * Added benchmark coverage for SBOM filtering across small, medium, and large path sets.

### DIFF
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -784,11 +784,13 @@ func filterSBOM(sbom domain.SBOM, instanceID instanceidhandler.IInstanceID, wlid
 	addedFileIDs := mapset.NewSet[string]()
 	addedRelationshipIDs := mapset.NewSet[string]()
 
-	// filter relevant files with dynamic paths
-	var dynamicPaths []string
+	// filter relevant files with dynamic paths, indexed by segment count since
+	// CompareDynamic only ever matches paths with the same number of segments
+	dynamicPathsBySegmentCount := make(map[int][]string)
 	relevantFiles.Each(func(file string) bool {
 		if strings.Contains(file, dynamicpathdetector.DynamicIdentifier) {
-			dynamicPaths = append(dynamicPaths, file)
+			segmentCount := strings.Count(file, "/")
+			dynamicPathsBySegmentCount[segmentCount] = append(dynamicPathsBySegmentCount[segmentCount], file)
 		}
 		return false
 	})
@@ -803,8 +805,8 @@ func filterSBOM(sbom domain.SBOM, instanceID instanceidhandler.IInstanceID, wlid
 				filteredSBOM.Content.Files = append(filteredSBOM.Content.Files, f)
 				continue
 			}
-			// then try dynamic match (expensive lookup)
-			for _, dynamicPath := range dynamicPaths {
+			// then try dynamic match (expensive lookup), limited to candidates with a matching segment count
+			for _, dynamicPath := range dynamicPathsBySegmentCount[strings.Count(f.Location.RealPath, "/")] {
 				if dynamicpathdetector.CompareDynamic(dynamicPath, f.Location.RealPath) {
 					addedFileIDs.Add(f.ID)
 					filteredSBOM.Content.Files = append(filteredSBOM.Content.Files, f)
--- a/core/services/scan_test.go
+++ b/core/services/scan_test.go
@@ -1074,6 +1074,41 @@ func Test_filterSBOM(t *testing.T) {
 	}
 }
 
+func BenchmarkFilterSBOM(b *testing.B) {
+	nginxSBOM := domain.SBOM{
+		Name: "nginx-sbom",
+		Annotations: map[string]string{
+			helpersv1.ImageIDMetadataKey:  "docker.io/library/nginx@sha256:04ba374043ccd2fc5c593885c0eacddebabd5ca375f9323666f28dfd5a9710e3",
+			helpersv1.ImageTagMetadataKey: "nginx:1.14.1",
+			helpersv1.StatusMetadataKey:   helpersv1.Learning,
+		},
+		Content: fileToSyftDocument("../../adapters/v1/testdata/nginx-sbom.json"),
+	}
+	instanceID, err := instanceidhandlerv1.GenerateInstanceIDFromString(
+		"apiVersion-apps/v1/namespace-default/kind-Deployment/name-nginx/containerName-nginx",
+	)
+	require.NoError(b, err)
+	labels := map[string]string{
+		helpersv1.ContainerNameMetadataKey: "nginx",
+	}
+	wlid := "wlid://cluster-test/namespace-default/deployment-nginx"
+
+	for _, dynamicPathCount := range []int{10, 100, 1000} {
+		dynamicPathCount := dynamicPathCount
+		b.Run(fmt.Sprintf("dynamicPaths=%d", dynamicPathCount), func(b *testing.B) {
+			relevantFiles := mapset.NewSet[string]()
+			for i := 0; i < dynamicPathCount; i++ {
+				relevantFiles.Add(fmt.Sprintf("/var/lib/%s/segment-%d/data", dynamicpathdetector.DynamicIdentifier, i))
+			}
+			b.ResetTimer()
+			for i := 0; i < b.N; i++ {
+				_, err := filterSBOM(nginxSBOM, instanceID, wlid, relevantFiles, labels, helpersv1.Full)
+				require.NoError(b, err)
+			}
+		})
+	}
+}
+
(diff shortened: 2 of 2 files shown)


rootp1:
HI @matthyx please have a look whenever you have a chance to

core/services/scan.go
```diff
@@ -784,11 +784,13 @@ func filterSBOM(sbom domain.SBOM, instanceID instanceidhandler.IInstanceID, wlid
...
-	// filter relevant files with dynamic paths
-	var dynamicPaths []string
+	// filter relevant files with dynamic paths, indexed by segment count since
+	// CompareDynamic only ever matches paths with the same number of segments
+	dynamicPathsBySegmentCount := make(map[int][]string)
 	relevantFiles.Each(func(file string) bool {
 		if strings.Contains(file, dynamicpathdetector.DynamicIdentifier) {
-			dynamicPaths = append(dynamicPaths, file)
+			segmentCount := strings.Count(file, "/")
+			dynamicPathsBySegmentCount[segmentCount] = append(dynamicPathsBySegmentCount[segmentCount], file)
```
matthyx, line 793: Nit / hardening: the segment-count invariant holds for every input except one — a trailing `/` on exactly one side. `CompareDynamic("/etc/nginx/⋯", "/etc/nginx/conf.d/")` returns `true` today with 3 vs 4 slashes, so the two would land in different buckets here and the file would silently stop being matched.

Not reachable in practice (dynamic paths come from `AnalyzePath`, which `path.Clean`s), but it costs one call to make the invariant unconditional instead of dependent on a caller's normalization:

```go
segmentCount := strings.Count(strings.TrimSuffix(file, "/"), "/")
```

with the matching `strings.Count(strings.TrimSuffix(f.Location.RealPath, "/"), "/")` on the lookup side. Trimming can only merge buckets, never split them, so it stays a safe over-approximation.

core/services/scan_test.go
```diff
@@ -1074,6 +1074,41 @@ func Test_filterSBOM(t *testing.T) {
...
+		"apiVersion-apps/v1/namespace-default/kind-Deployment/name-nginx/containerName-nginx",
+	)
+	require.NoError(b, err)
+	labels := map[string]string{
+		helpersv1.ContainerNameMetadataKey: "nginx",
+	}
+	wlid := "wlid://cluster-test/namespace-default/deployment-nginx"
+
+	for _, dynamicPathCount := range []int{10, 100, 1000} {
+		dynamicPathCount := dynamicPathCount
```
matthyx, line 1097: Nit: `dynamicPathCount := dynamicPathCount` is no longer needed — `go.mod` targets go 1.25, and loop variables have been per-iteration since 1.22.

matthyx (approved):
Reviewed the change, ran it locally, and verified the core claim. **Approving** — the optimization is correct in practice and the win is real. A few non-blocking notes below.

### Verification

- `go build ./...` and `go vet ./core/...` clean; `Test_filterSBOM` passes unchanged.
- Benchmark, same machine, `-benchtime 20x`, `main`'s `scan.go` vs this PR's (same `BenchmarkFilterSBOM`):

  | dynamicPaths | main | this PR | speedup |
  |---|---|---|---|
  | 10 | 562 µs | 479 µs | 1.2x |
  | 100 | 3.02 ms | 1.26 ms | 2.4x |
  | 1000 | 28.4 ms | 9.37 ms | 3.0x |

- The pruning holds up on realistic input: the `nginx-sbom.json` fixture has 2859 file entries spread over slash-depths 2–8 (134/356/268/946/604/137/414), so a bucketed lookup really does skip most of the candidate set rather than being a best-case artifact of the benchmark.

### The invariant is *almost* unconditional

I brute-forced `CompareDynamic(d, r) ⇒ strings.Count(d, "/") == strings.Count(r, "/")` over ~300k random segment combinations. It holds in every case **except when exactly one side has a trailing `/`**:

```
CompareDynamic("/etc/nginx/⋯", "/etc/nginx/conf.d/") == true   // 3 vs 4 slashes
CompareDynamic("/etc/nginx/⋯/", "/etc/nginx/conf.d") == true   // 4 vs 3 slashes
```

The final `dynamicIndex++` / `regularIndex++` steps past a trailing slash, so `dynamicIndex >= dynamicLen && regularIndex >= regularLen` can still be satisfied with unequal slash counts. Those pairs match today and would stop matching after this change — a silent drop of a relevant file, not a visible failure.

I don't think it can actually bite right now: dynamic paths only ever come from `PathAnalyzer.AnalyzePath`, which runs `path.Clean` and therefore never emits a trailing slash, and the syft fixture has 0 trailing-slash `realPath`s out of 2859. So this is hardening, not a blocker — see the inline suggestion.

### Also worth doing (not blocking)

- **No unit test pins the new behavior.** Only a benchmark was added, and it asserts `err` only, so it cannot detect a dropped match. `Test_filterSBOM` already covers the same-depth dynamic match (`scan_test.go:1006`); please add a case where a dynamic relevant path sits at a *different* depth than a candidate file and must not match. Without it, an off-by-one in the bucket key silently filters files out with a green test suite.
- **The benchmark exercises one bucket.** All generated paths have 4 slashes, so it measures "SBOM files at other depths get skipped" rather than cross-bucket distribution (CodeRabbit made the same point). The interesting adversarial case is the opposite one — every dynamic path at the same depth as the bulk of the SBOM files — where the index buys nothing; worth a subbenchmark so the ceiling and the floor are both on record.

Issue #450, opened 2026-08-04
Bug: ReportError/SendStatus still don't propagate ctx to the network layer (follow-up to #446)


bhuvan-somisetty:
# Description

Follow-up from #446: `BackendAdapter.ReportError` and `BackendAdapter.SendStatus` (`adapters/v1/backend.go`) still don't propagate the caller's `ctx` to the network layer, even after #446 fixed the same gap for `postResults` and `ReportScanFailure`.

Both go through `backendClientV1.NewBaseReportSender(...)` (`kubescape/backend@v0.0.40` `pkg/client/v1/systemreports.go:62`), whose `SendStatus`/`SendError` call `unprotectedSendAsRoutine` → `Send()` → `s.httpSender.Send(...)` (`httpreportsender.go:21`). That in turn calls `httputils.HttpPost` (no `context.Context` parameter at all — it hardcodes `context.Background()` before the request ever reaches our `httputils.IHttpClient`) and retries up to 3 times with a blocking, uncancellable `time.Sleep(RETRY_DELAY)` (`RETRY_DELAY = 5 * time.Second`).

This means there's no seam in `kubevuln` to inject `ctx` here without either:
- bypassing `NewBaseReportSender`/`sendStatusFunc` entirely and posting the status report directly through kubevuln's own ctx-aware HTTP path (duplicating some of `BaseReportSender.Send()`'s logic — URL construction, `ActionID` defaulting, `JobID` capture from the response body), or
- a change to `kubescape/backend` itself, adding a ctx-accepting `BaseReportSender` constructor/`Send` method, which `kubevuln` would then need to pull in via a version bump.

This was deliberately scoped out of #446 rather than assumed, since it's a real design decision affecting a shared dependency and changes production status-reporting retry semantics — it should be discussed with a maintainer before either option is picked.

# Environment

- OS: N/A — code-review finding
- Version: `kubevuln` main (post-#446)

# Steps To Reproduce

1. In `adapters/v1/backend.go`, read `ReportError` and `SendStatus` — both build a `sender := backendClientV1.NewBaseReportSender(...)` and call `a.sendStatusFunc(sender, ...)`.
2. In `kubescape/backend@v0.0.40` `pkg/client/v1/systemreports.go:62`, confirm `NewBaseReportSender` takes no `context.Context` parameter.
[...]

PR #451, opened 2026-08-04
fix: propagate ctx to backend HTTP retries and gate chunk dispatch on summary success (#446)

### LINKED ISSUE #446: Bug: BackendAdapter reporting never honors ctx cancellation, and a failed summary POST still lets chunks ship unguarded
# Description

`BackendAdapter`'s platform-reporting layer (`adapters/v1/backend.go`, `adapters/v1/backend_utils.go`) has two related defects left over after #437/#440: the caller's `ctx` never actually reaches the network layer at any of the four reporting entry points, and the now-synchronous summary POST introduced by #440 can fail without stopping the vulnerability chunks that depend on it from going out anyway.

Both were flagged during review of #440 (comment on PR #440, "Non-blocking follow-ups") but never turned into their own tracked issue, so they're still open on `main`.

## Defect A — `ctx` cancellation never reaches the HTTP layer

`BackendAdapter.httpPostFunc` (`adapters/v1/backend.go:40`) has the signature `func(httputils.IHttpClient, string, map[string]string, []byte, time.Duration) (*http.Response, error)` — no `context.Context` parameter. It defaults to `httputils.HttpPostWithRetry` (`backend.go:67`), which (per `armosec/utils-go@v0.0.58` `httputils/httphelpers.go:78-80`) hardcodes `context.Background()` into the underlying `HttpPostWithContext` call. So the `ctx` that `postResults` (`adapters/v1/backend_utils.go:101`) and `ReportScanFailure` (`backend.go:253`) receive is used only for logging and for the `sendError` channel guard — it's never passed to `http.NewRequestWithContext` or to the exponential-backoff retry loop wrapping the request.

The other two entry points hit the same wall through a different path: `ReportError` (`backend.go:200-201`) and `SendStatus` (`backend.go:281-282`) go through `backendClientV1.NewBaseReportSender(...)` (`kubescape/backend@v0.0.40` `pkg/client/v1/systemreports.go:62`). Its `SendStatus`/`SendError` call `unprotectedSendAsRoutine` → `Send()` → `s.httpSender.Send(...)` (`httpreportsender.go:21`), which calls `httputils.HttpPost` (no ctx parameter at all) and retries up to 3 times with a **blocking, uncancellable `time.Sleep(RETRY_DELAY)`** (`httpreportsender.go:25-48`, `RETRY_DELAY = 5 * time.Second`). Neither the request nor the inter-retry sleep observes any context.

Net effect: cancelling the caller's `ctx` never aborts an in-flight report at any of the four call sites. The `sendError` fix from #440 only stops kubevuln from *waiting* on the result — the goroutine underneath keeps doing the (now pointless) network work and retries to completion regardless.

## Defect B — a failed summary POST doesn't stop the chunks

`postResults` (`backend_utils.go:101`) returns nothing — its only way to signal failure is pushing onto `errorChan`. `sendSummaryAndVulnerabilities` (`backend_utils.go:52-85`) was changed by #440 to call `a.postResults(...)` **synchronously** for the summary (line 69) specifically so the backend always receives the summary before any chunk — but because `postResults` has no return value, it can't (and today doesn't) check whether that synchronous call actually succeeded before dispatching the first chunk (`postResultsAsGoroutine`, line 73). The same gap exists one level up: `SubmitCVE` (`backend.go:310`) only reads the `nextPartNum` return value from `sendSummaryAndVulnerabilities` and unconditionally proceeds to `sendVulnerabilitiesRoutine` for the rest of the chunks (`backend.go:437`), regardless of whether the summary landed.

A permanently-failed summary (malformed payload, 4xx from the backend, or the retry budget from Defect A simply expiring) still lets every subsequent chunk go out — reproducing the exact `HTTP 400 Bad Request: missing summary for containerScanID` failure mode #437 was originally filed to prevent, just via unconditional dispatch instead of a race.

# Environment

- OS: N/A — code-review finding
- Version: `kubevuln` main (post-#440, current as of this filing)

# Steps To Reproduce

1. In `adapters/v1/backend.go`, read the `httpPostFunc` field declaration (line 40) and its default assignment (line 67) — confirm the signature has no `context.Context` parameter and that `httputils.HttpPostWithRetry` binds `context.Background()` internally (`armosec/utils-go@v0.0.58` `httphelpers.go:78-80`).
2. In `adapters/v1/backend_utils.go`, read `postResults` (line 101) — confirm `ctx` is used only for `logger.L().Ctx(ctx)` and `sendError(ctx, ...)`, never passed into `a.httpPostFunc(...)` (line 118).
3. In `kubescape/backend@v0.0.40` `pkg/client/v1/httpreportsender.go:21-49`, confirm `HttpReportSender.Send` retries via `httputils.HttpPost` (ctx-less) with a hardcoded `time.Sleep(RETRY_DELAY)` between attempts.
4. In `adapters/v1/backend_utils.go`, read `sendSummaryAndVulnerabilities` (lines 52-85) — confirm the synchronous call to `a.postResults(...)` on line 69 is not gated on any success/failure signal before line 73 dispatches the first chunk.
5. In `adapters/v1/backend.go`, read `SubmitCVE` (around line 433-437) — confirm `nextPartNum` is the only value read off `sendSummaryAndVulnerabilities`'s return, and `sendVulnerabilitiesRoutine` (line 437) is invoked unconditionally.

# Expected behavior

- Cancelling the caller's `ctx` for `ReportError`, `SendStatus`, `postResults`, or `ReportScanFailure` should abort the in-flight HTTP request and any pending retry/backoff wait promptly, not run the full retry budget to completion regardless of cancellation.
- If the synchronous summary POST inside `sendSummaryAndVulnerabilities` fails, no vulnerability chunk for that scan should be dispatched — the failure should propagate back through `SubmitCVE` instead of being silently swallowed.

# Actual Behavior

- A cancelled scan still burns its full retry/backoff budget (up to 60s for `postResults`/`ReportScanFailure`'s `HttpPostWithRetry` call, or up to 3 × 5s plus request time for the `BaseReportSender` path) doing HTTP work nobody is waiting on anymore.
- A permanently-failed summary POST still lets every vulnerability chunk for that scan go out, reproducing the `missing summary for containerScanID` rejection #437 was meant to eliminate.

# Additional context

### Impact

- **Reliability / resource waste**: under mass scan cancellation (workload deletion, redeploys, scan timeouts), every cancelled `SubmitCVE` still holds a connection and a goroutine doing full-budget retries for work whose result is discarded. This is exactly the kind of load amplification that matters most during the failure scenario the whole subsystem exists to survive: a degraded or rate-limiting event receiver (the same "widespread HTTP 429/500" scenario #437 cited) now gets hit by both the original request *and* every cancelled retry that never got to stop.
- **Data consistency**: Defect B reopens the specific symptom #437 was filed and closed for (`400 missing summary for containerScanID`), just through a different mechanism — a real regression risk for anyone who assumed synchronous-summary alone was sufficient.
- **Latency coupling**: because the summary POST is now synchronous and on `SubmitCVE`'s critical path (a deliberate, correct trade-off from #440), a slow event receiver stalls the entire scan pipeline for up to a minute before any chunk is attempted — and today that whole wait is wasted anyway, since nothing gates on its outcome (Defect B).

- `adapters/v1/backend.go` — `BackendAdapter.httpPostFunc` field/default, `ReportError`, `SendStatus`, `ReportScanFailure`, `SubmitCVE`.
- `adapters/v1/backend_utils.go` — `postResults`, `sendSummaryAndVulnerabilities`, `postResultsAsGoroutine`.
- `adapters/v1/backend_test.go` — every test currently stubbing `httpPostFunc` with the ctx-less signature will need updating.
- Indirectly, `kubescape/backend@v0.0.40` `pkg/client/v1/systemreports.go` / `httpreportsender.go` (external dependency — see implementation plan point 2 for how to scope around this without requiring an upstream change).

### Acceptance Criteria

- [ ] Cancelling `ctx` mid-call for each of `ReportError`, `SendStatus`, `postResults`, and `ReportScanFailure` causes the call to return promptly (bounded by transport granularity, not by the full retry budget), verified by a test per call site using a controllable/slow mock `httputils.IHttpClient` plus a ctx cancelled during the call.
- [ ] `postResults` communicates its outcome to its synchronous caller; `sendSummaryAndVulnerabilities` does not dispatch the first chunk, and `SubmitCVE` does not invoke `sendVulnerabilitiesRoutine` for the remaining chunks, when the summary POST fails — verified by a test that fails the mocked summary POST and asserts zero chunk-POST calls occur.
- [ ] No behavioral change on the success path — existing tests (`TestBackendAdapter_SubmitCVE`, `TestBackendAdapter_HTTPClientReuse`, etc.) pass unmodified.
- [ ] If full ctx propagation for `ReportError`/`SendStatus` requires a change to `kubescape/backend`'s `BaseReportSender`, the PR either routes those two calls through kubevuln's own ctx-aware path instead, or explicitly documents the remaining gap with a link to a tracking issue against that dependency — this shouldn't be silently left half-done.

1. **Ctx-aware default post function**: replace the `httpPostFunc` default (`httputils.HttpPostWithRetry`, ctx-less) with a kubevuln-local wrapper around `httputils.HttpPostWithContext(ctx, client, url, headers, body, maxElapsedTime, shouldRetry)`, reimplementing the small retry predicate locally (`utils-go`'s `defaultShouldRetry` is unexported, so it can't be reused directly). Update the `httpPostFunc` field type to take `context.Context` and update both call sites (`postResults`, `ReportScanFailure`) plus every test stub.
2. **Decide the `ReportError`/`SendStatus` path**: either (a) stop routing these two through `backendClientV1.NewBaseReportSender`/`sendStatusFunc` and post directly through the ctx-aware path from step 1, or (b) keep `NewBaseReportSender` for parity with the rest of the platform-reporting ecosystem and scope a companion ctx-accepting sender/constructor change in `kubescape/backend`, taking a version bump once available. This is a real design decision affecting a shared dependency and should be confirmed with a maintainer before implementation rather than assumed.
3. **Gate chunk dispatch on summary success**: give `postResults` a return value (`error` — it already has exactly one synchronous caller via `sendSummaryAndVulnerabilities`, and the goroutine-wrapped path via `postResultsAsGoroutine` keeps using `errorChan` as-is). Have `sendSummaryAndVulnerabilities` check that return before dispatching the first chunk, and propagate the failure up through `SubmitCVE` so `sendVulnerabilitiesRoutine`/`sendVulnerabilities` are skipped for the remaining chunks when the summary never landed.
4. **Tests**: a cancellation test per call site using a slow/blocking mock `IHttpClient.Do` with ctx cancelled mid-call, asserting prompt return; a summary-failure test asserting zero chunk POSTs are attempted; update every existing `httpPostFunc` mock signature in `backend_test.go`.
5. **Documentation**: note the cancellation contract and the summary-gates-chunks invariant in the relevant function comments, since a future contributor reverting the synchronous summary call for latency reasons (a real temptation given point 3 in Impact) would silently reintroduce the #437 ordering race — this is exactly the kind of invariant that's invisible at the call site without a comment.

bhuvan-somisetty:
## Overview

`BackendAdapter`'s platform-reporting layer (`adapters/v1/backend.go`, `adapters/v1/backend_utils.go`) had two related gaps left over after #437/#440: the caller's `ctx` never reached the network layer at any of the four reporting entry points, and a failed summary POST didn't stop the vulnerability chunks that depend on it from going out anyway. Both were flagged during review of #440 but never turned into a tracked issue until #446.

## What changed

1. **`ctx` now reaches the HTTP retry loop, not just the request.** `httpPostFunc`'s signature gained a `context.Context` parameter. The new default implementation, `httpPostWithContext`, builds each request via `http.NewRequestWithContext(ctx, ...)` and drives retries through `cenkalti/backoff/v5`'s `Retry(ctx, ...)` (already a dependency — used the same way in `pkg/sbomscanner/v1/client.go`), which checks `ctx` both before each retry attempt and while sleeping between attempts. This replaces `httputils.HttpPostWithRetry`, which binds `context.Background()` to the request and retries on a plain, non-cancellable `backoff.Retry` loop — so even passing `ctx` through wouldn't have been enough on its own. `postResults` and `ReportScanFailure` now pass `ctx` through.
2. **A failed summary POST now stops the chunks.** `postResults` returns an `error`. `sendSummaryAndVulnerabilities` checks it before dispatching the first chunk, and `SubmitCVE` skips `sendVulnerabilitiesRoutine` for the rest when the summary failed — draining the (now-abandoned) `chunksChan` itself so `SplitSlice2Chunks`' producer goroutine doesn't block forever trying to hand off chunks nobody reads. A permanently-failed summary no longer lets every chunk ship and get rejected with `missing summary for containerScanID`.
3. **`ReportError`/`SendStatus` are documented, not silently left half-fixed.** Both still go through `backendClientV1.NewBaseReportSender`, which builds its own request bound to `context.Background()` inside `armosec/utils-go` before it ever reaches our `IHttpClient` — there's no seam here to inject `ctx` without either bypassing the sender or a change to `kubescape/backend`. That's a real design decision affecting a shared dependency, so it's called out in a code comment on both functions and tracked separately in #450 rather than assumed or silently dropped.

## Tests

- `TestBackendAdapter_PostResultsAbortsPromptlyOnCtxCancellation` / `TestBackendAdapter_ReportScanFailureAbortsPromptlyOnCtxCancellation`: a blocking mock `IHttpClient` that only unblocks when the request's own context is cancelled, proving the call returns promptly once `ctx` is cancelled instead of running the full retry budget.
- `TestBackendAdapter_SubmitCVE_SkipsChunksWhenSummaryFails`: fails the mocked summary POST and asserts zero chunk POSTs occur, using the same large fixture (`testdata/nginx-cve.json`) that exercises the multi-chunk `chunksChan` drain path.
- `Test_sendSummaryAndVulnerabilities` and the `httpPostFunc` mocks across `backend_test.go`/`backend_utils_test.go` updated for the new ctx-aware signature and `postResults`' new return value.

## How to Test

```bash
go test -v -count=1 -run "TestBackendAdapter|TestSendError|Test_sendSummaryAndVulnerabilities" ./adapters/v1/...
```

## Related issues/PRs:

* Resolved #446
* Follow-up tracked in #450 (ReportError/SendStatus ctx propagation, needs a maintainer call on `kubescape/backend`)

### DIFF
--- a/adapters/v1/backend.go
+++ b/adapters/v1/backend.go
@@ -1,6 +1,7 @@
 package v1
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"net/http"
@@ -21,6 +22,7 @@ import (
 	"github.com/armosec/utils-go/httputils"
 	pkgcautils "github.com/armosec/utils-k8s-go/armometadata"
 	wlidpkg "github.com/armosec/utils-k8s-go/wlid"
+	"github.com/cenkalti/backoff/v5"
 	"github.com/hashicorp/go-multierror"
 	backendClientV1 "github.com/kubescape/backend/pkg/client/v1"
 	sysreport "github.com/kubescape/backend/pkg/server/v1/systemreports"
@@ -37,7 +39,7 @@ type BackendAdapter struct {
 	apiServerRestURL      string
 	clusterConfig         pkgcautils.ClusterConfig
 	getCVEExceptionsFunc  func(string, string, *identifiers.PortalDesignator, map[string]string) ([]armotypes.VulnerabilityExceptionPolicy, error)
-	httpPostFunc          func(httputils.IHttpClient, string, map[string]string, []byte, time.Duration) (*http.Response, error)
+	httpPostFunc          func(context.Context, httputils.IHttpClient, string, map[string]string, []byte, time.Duration) (*http.Response, error)
 	sendStatusFunc        func(*backendClientV1.BaseReportSender, string, bool)
 	accessKey             string
 	securityExceptionRepo ports.SecurityExceptionRepository
@@ -64,7 +66,7 @@ func NewBackendAdapter(accountID, apiServerRestURL, eventReceiverRestURL, access
 		eventReceiverRestURL: eventReceiverRestURL,
 		apiServerRestURL:     apiServerRestURL,
 		getCVEExceptionsFunc: backendClientV1.GetCVEExceptionByDesignator,
-		httpPostFunc:         httputils.HttpPostWithRetry,
+		httpPostFunc:         httpPostWithContext,
 		sendStatusFunc: func(sender *backendClientV1.BaseReportSender, status string, sendReport bool) {
 			sender.SendStatus(status, sendReport) // TODO - update this function to use from kubescape/backend
 		},
@@ -88,6 +90,47 @@ func (a *BackendAdapter) getHTTPClient() httputils.IHttpClient {
 		return a.httpClient
 	}
 	return http.DefaultClient
+}
+
+// httpPostWithContext is the default httpPostFunc. Unlike httputils.HttpPostWithRetry (which
+// binds context.Background() to the request and retries on a plain, non-cancellable
+// backoff.Retry loop), it threads the caller's ctx through the request AND the retry loop, so
+// cancelling ctx aborts both an in-flight attempt and any further retries promptly instead of
+// running the full retry budget to completion regardless of cancellation (#446).
+func httpPostWithContext(ctx context.Context, httpClient httputils.IHttpClient, fullURL string, headers map[string]string, body []byte, maxElapsedTime time.Duration) (*http.Response, error) {
+	bo := backoff.NewExponentialBackOff()
+	return backoff.Retry(ctx, func() (*http.Response, error) {
+		req, err := http.NewRequestWithContext(ctx, http.MethodPost, fullURL, bytes.NewReader(body))
+		if err != nil {
+			return nil, backoff.Permanent(err)
+		}
+		for k, v := range headers {
+			req.Header.Set(k, v)
+		}
+
+		resp, err := httpClient.Do(req)
+		if err != nil {
+			return nil, err
+		}
+		if resp.StatusCode != http.StatusOK {
+			defer resp.Body.Close()
+			retryErr := fmt.Errorf("received status code: %d", resp.StatusCode)
+			if !shouldRetryReport(resp) {
+				return nil, backoff.Permanent(retryErr)
+			}
+			return nil, retryErr
+		}
+		return resp, nil
+	}, backoff.WithBackOff(bo), backoff.WithMaxElapsedTime(maxElapsedTime))
+}
+
+// shouldRetryReport mirrors the unexported defaultShouldRetry in armosec/utils-go: retry on any
+// response except the ones that are never going to succeed on a retry.
+func shouldRetryReport(resp *http.Response) bool {
+	return resp.StatusCode != http.StatusUnauthorized &&
+		resp.StatusCode != http.StatusForbidden &&
+		resp.StatusCode != http.StatusNotFound &&
+		resp.StatusCode != http.StatusInternalServerError
 }
 
 const ActionName = "vuln scan"
@@ -197,6 +240,12 @@ func (a *BackendAdapter) ReportError(ctx context.Context, err error) error {
 	report.Errors = append(report.Errors, err.Error())
 	report.Status = sysreport.JobFailed
 
+	// NOTE: unlike postResults/ReportScanFailure, this call is not ctx-cancellation-aware.
+	// backendClientV1.NewBaseReportSender builds its own *http.Request bound to
+	// context.Background() inside armosec/utils-go, before it ever reaches our IHttpClient, so
+	// there's no seam here to inject ctx without either bypassing the sender or a ctx-accepting
+	// constructor upstream in kubescape/backend. Tracked separately (#450) rather than silently
+	// left unaddressed.
 	sender := backendClientV1.NewBaseReportSender(a.eventReceiverRestURL, a.getHTTPClient(), a.getRequestHeaders(), report)
 	a.sendStatusFunc(sender, sysreport.JobFailed, true)
 	return nil
@@ -250,7 +299,7 @@ func (a *BackendAdapter) ReportScanFailure(ctx context.Context, failureCase scan
 	}
 
 	url := fmt.Sprintf("%s/k8s/v2/scanFailure", a.eventReceiverRestURL)
-	resp, err := a.httpPostFunc(a.getHTTPClient(), url, a.getRequestHeaders(), payload, 30*time.Second)
+	resp, err := a.httpPostFunc(ctx, a.getHTTPClient(), url, a.getRequestHeaders(), payload, 30*time.Second)
 	if err != nil {
 		logger.L().Ctx(ctx).Warning("failed to send scan failure report",
 			helpers.Error(err),
@@ -278,6 +327,8 @@ func (a *BackendAdapter) SendStatus(ctx context.Context, step int) error {
 	report.Details = details[step]
 	report.Status = statuses[step]
 
+	// NOTE: see the same comment on ReportError above — this call has the same ctx-cancellation
+	// gap for the same reason.
 	sender := backendClientV1.NewBaseReportSender(a.eventReceiverRestURL, a.getHTTPClient(), a.getRequestHeaders(), report)
 	a.sendStatusFunc(sender, statuses[step], true)
 	return nil
@@ -430,7 +481,19 @@ func (a *BackendAdapter) SubmitCVE(ctx context.Context, cve domain.CVEManifest, 
 	firstVulnerabilitiesChunk := <-chunksChan
 	firstChunkVulnerabilitiesCount := len(firstVulnerabilitiesChunk)
 	// send the summary and the first chunk in one or two reports according to the size
-	nextPartNum := a.sendSummaryAndVulnerabilities(ctx, &finalReport, a.eventReceiverRestURL, totalVulnerabilities, scanID, firstVulnerabilitiesChunk, errChan, sendWG)
+	nextPartNum, summaryErr := a.sendSummaryAndVulnerabilities(ctx, &finalReport, a.eventReceiverRestURL, totalVulnerabilities, scanID, firstVulnerabilitiesChunk, errChan, sendWG)
+	if summaryErr != nil {
+		// Nothing was sent, so no chunk goroutine was dispatched and nothing will ever drain
+		// chunksChan or close errChan on its own. Drain chunksChan ourselves so the producer
+		// goroutine started by SplitSlice2Chunks doesn't block forever trying to hand off the
+		// remaining chunks (#446), and return directly instead of ranging over an errChan
+		// nothing will close.
+		go func() {
+			for range chunksChan {
+			}
+		}()
+		return summaryErr
+	}
 	// if not all vulnerabilities got into the first chunk
 	if totalVulnerabilities != firstChunkVulnerabilitiesCount {
 		//send the rest of the vulnerabilities - error channel will be closed when all vulnerabilities are sent
--- a/adapters/v1/backend_test.go
+++ b/adapters/v1/backend_test.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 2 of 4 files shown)


bhuvan-somisetty:
@matthyx PTAL whenever u get chance

adapters/v1/backend.go
```diff
@@ -90,6 +92,47 @@ func (a *BackendAdapter) getHTTPClient() httputils.IHttpClient {
...
+			req.Header.Set(k, v)
+		}
+
+		resp, err := httpClient.Do(req)
+		if err != nil {
+			return nil, err
+		}
+		if resp.StatusCode != http.StatusOK {
+			defer resp.Body.Close()
+			retryErr := fmt.Errorf("received status code: %d", resp.StatusCode)
```
matthyx, line 117: **Non-blocking: the server's response body is lost from the error path here.**

Before this PR, a 401/403/404/500 hit `defaultShouldRetry == false` in utils-go, which does `return backoff.Permanent(err)` — and `err` is **nil** at that point, because the `err` declared by `req, err := http.NewRequestWithContext(...)` was reassigned by the *successful* `httpClient.Do`. `backoff.Permanent(nil)` returns nil (v4 `retry.go:139-142`), so `Retry` reported success and `HttpPostWithRetry` returned `(resp, nil)`. `postResults` then fell through to `httputils.HttpRespToString(resp)`, which builds `http-error: '500 ...', reason: '<server body>'` and logs it via `helpers.String("body", body)`.

Now the same response returns `received status code: 500` with the body closed and discarded, so the server's explanation — e.g. the `missing summary for containerScanID` message that kicked off #437 — never reaches the logs.

Your version is strictly more correct at the error-signalling layer (`Permanent(nil)` making a 500 look like a successful post is an upstream bug you're no longer inheriting); it just costs the message. Keeping both is cheap:

```go
if resp.StatusCode != http.StatusOK {
	defer resp.Body.Close()
	respBody, _ := io.ReadAll(io.LimitReader(resp.Body, 1024))
	retryErr := fmt.Errorf("received status code: %d, body: %s", resp.StatusCode, respBody)
	if !shouldRetryReport(resp) {
	}
	return nil, retryErr
}
```

That also makes the 429 branch in `postResults` and the `>= 400` check in `ReportScanFailure` meaningful again instead of unreachable.

matthyx (approved):
Reviewed the full diff against `main`, plus the upstream sources it leans on (`armosec/utils-go@v0.0.58` `httputils`, `cenkalti/backoff/v5@v5.0.3`, `kubescape/backend@v0.0.40`). Ran `go build ./...`, `go vet ./adapters/...` and `go test -count=1 ./adapters/v1/...` locally against `b08bf10` — all green, and CI is green too.

**No blockers — approving.**

Both defects from #446 are genuinely fixed, and the tests prove it rather than just asserting around it: `blockingHTTPClient.Do` only unblocks on `req.Context().Done()`, so the two cancellation tests would hang and fail against the old `context.Background()`-bound path. That's the right shape of test for this bug.

### Things I verified, so the next reader doesn't have to

- **Reimplementing the POST helper is justified, and the description undersells why.** `httputils.HttpPostWithContext(ctx, ...)` already exists and *does* bind ctx to the request — `httpPostDebug` right below in the same file calls it, which makes it look like an unused seam that would have solved this. It wouldn't have: its retry loop is backoff **v4**'s `backoff.Retry(operation, expBackOff)`, which takes no ctx, so after cancellation `Do` fails instantly and the loop keeps re-issuing and sleeping for the entire 60s budget. The v5 `Retry(ctx, ...)` loop is what actually buys the prompt abort. One sentence naming `HttpPostWithContext` in the doc comment would close that question for good.
- **`shouldRetryReport` faithfully mirrors `defaultShouldRetry`**, so retry classification is unchanged. (Side note, pre-existing: 400 is still retryable, so a permanently-malformed summary burns the full 60s before `SubmitCVE` bails.)
- **The `chunksChan` drain is correct.** `SplitSlice2Chunks` always closes the channel after its producer WaitGroup drains (`httphelpers.go:200-208`), so the drain goroutine terminates rather than leaking.
- **No caller-visible change in error propagation.** A failed summary already reached `SubmitCVE`'s caller via `errChan` + `multierror`. The win is that chunks nobody would accept no longer ship, and the call returns immediately instead of after the whole chunk fan-out.

### Non-blocking follow-ups (separate PR is fine)

1. **The response body is now dropped from the failure log** — inline comment below. Worth a small fix, since debuggability of backend rejections is the whole point of this issue chain.
2. `ReportScanFailure`'s `if resp.StatusCode >= http.StatusBadRequest` (`adapters/v1/backend.go:312`) and the 429 branch in `postResults` (`adapters/v1/backend_utils.go:141`) are now unreachable — `httpPostWithContext` only ever returns a 200 response. Harmless dead code, but both become live again if you take (1).
3. `httpPostDebug` (`adapters/v1/backend.go:519`) no longer matches the `httpPostFunc` field type, so it can't be dropped in as a debug override anymore. Update its signature or delete it.
4. Agree with CodeRabbit's wording nit on the `sendSummaryAndVulnerabilities` contract: `postResults` can also fail *after* a successful POST (body-read error), so "the summary was not confirmed accepted" is more precise than "nothing was sent". Behavior is right either way.

PR #457, opened 2026-08-05
fix: gate ScanCP image pulls on the too-many-requests backoff cache

### LINKED ISSUE #456: ScanCP's too-many-requests backoff never fires — wrong cache key and no read check
# Description

The registry rate-limit backoff (`ScanService.tooManyRequests`) is a shared cache that's supposed to stop kubevuln from re-hammering a registry once it's seen a `429 Too Many Requests` for an image. It works for `GenerateSBOM`, `ScanCVE` and `ScanRegistry`, but it's completely dead for `ScanCP` — the one flow that fans out multiple `CreateSBOM` calls per request (one per container in an ApplicationProfile/relevancy scan). Three things are wrong and they all live in `core/services/scan.go`:

1. `ValidateScanCP` (873-892) never reads `s.tooManyRequests` at all. Every sibling validator does:

```go
// ValidateGenerateSBOM (scan.go:867)
if _, ok := s.tooManyRequests.Get(workload.ImageHash); ok {
    return ctx, domain.ErrTooManyRequests
}
// ValidateScanCVE (scan.go:914) — same pattern, keyed by ImageHash
// ValidateScanRegistry (scan.go:936) — same pattern, keyed by ImageTagNormalized
```

`ValidateScanCP` only checks that `name`/`namespace` are non-empty — no lookup, no early return:

```go
func (s *ScanService) ValidateScanCP(ctx context.Context, workload domain.ScanCommand) (context.Context, error) {
	...
	name, _ := workload.Args[domain.ArgsName].(string)
	namespace, _ := workload.Args[domain.ArgsNamespace].(string)
	if name == "" || namespace == "" {
		return ctx, domain.ErrMissingCpInfo
	}
	...
	return ctx, nil
}
```

2. Even the write side uses the wrong key. Inside `ScanCP`'s per-container loop, the SBOM creation call records the 429 marker keyed by the *raw* `scan.ImageID`:

```go
// scan.go:224
sbom, err = s.sbomCreator.CreateSBOM(ctx, subWorkload.ImageSlug, subWorkload.ImageHash, subWorkload.ImageTagNormalized, optionsFromWorkload(ctx, subWorkload))
s.checkCreateSBOM(err, scan.ImageID)
```

but `subWorkload.ImageHash`, built two lines earlier in the same loop, is normalized:

```go
// scan.go:181
ImageHash: v1.NormalizeImageID(scan.ImageID, scan.ImageTag),
```

`NormalizeImageID` (`adapters/v1/syft.go:106`) doesn't return its input unchanged, so `scan.ImageID` and `subWorkload.ImageHash` are different strings. Even a fixed read in `ValidateScanCP` (keyed by `ImageHash`, matching every other flow) would never find the marker this write actually stores.

3. No per-container gating in the loop itself. `ScanCP` iterates every container in the ApplicationProfile (scan.go:166) and calls `CreateSBOM` for each independently. If container A's pull comes back 429, containers B, C, D sharing the same image still fire their own `CreateSBOM` calls in the same loop with no backoff check in between.

# Environment

Not a runtime/deployment-specific bug — reproducible by code inspection and unit test against current `main`. Applies to any kubevuln deployment that serves `ScanCP` (ApplicationProfile/relevancy scan) requests, regardless of OS or Kubescape version.

# Steps To Reproduce

1. Trigger a `ScanCP` request for a workload with a container whose image pull returns HTTP 429 (`transport.Error{StatusCode: http.StatusTooManyRequests}`) from `CreateSBOM`.
2. Observe `checkCreateSBOM` stores the marker under `scan.ImageID` (scan.go:224) — the raw, un-normalized image ID.
3. Send a second `ScanCP` request for the same image (or, in the same request, another container referencing the same image).
4. `ValidateScanCP` runs and never queries `s.tooManyRequests` — see scan.go:873-892 — so the request proceeds straight into another `CreateSBOM` call against the same registry.
5. Compare against `ValidateScanCVE`/`ValidateScanRegistry`: repeating the same steps for those flows returns `domain.ErrTooManyRequests` on the second call, confirming the mechanism works everywhere except `ScanCP`.
6. `core/services/scan_test.go` currently has no 429/`toomanyrequests` test case for `ScanCP` (`TestScanService_ScanCP`/`TestScanService_ValidateScanCP`), unlike `TestScanService_GenerateSBOM`, which does assert the cache round-trip — the gap isn't covered by existing tests either.

# Expected behavior

Once a container's image pull hits a 429 during `ScanCP`, subsequent `ScanCP` calls for the same image — and remaining containers referencing that image within the same call — should short-circuit with `domain.ErrTooManyRequests` instead of issuing another pull, exactly as `GenerateSBOM`/`ScanCVE`/`ScanRegistry` already do.

# Actual Behavior

`ScanCP` never checks the backoff cache before pulling, and the one write it does perform is stored under a cache key nothing ever reads (raw `scan.ImageID` vs. the normalized `ImageHash` used everywhere else). The backoff protection is fully inert for this flow: no fast-fail `429` back to the caller, and no in-loop gating between containers sharing an image. `ScanCP` is the path most likely to fan out several image pulls per request (multi-container pods), so this is exactly the flow where hitting a registry rate limit is most likely — and where nothing currently protects against amplifying a transient 429 into a longer registry ban. It also fails silently: no error or metric distinguishes "protection didn't fire" from "there was nothing to protect against."

# Additional context

Affected code:
- `core/services/scan.go` — `ValidateScanCP` (873-892), `ScanCP` (143-266, `checkCreateSBOM` call at 224), `checkCreateSBOM` (79-86)
- `adapters/v1/syft.go` — `NormalizeImageID` (106), for why the keys diverge

Acceptance criteria for a fix:
- `ValidateScanCP` checks `s.tooManyRequests` before proceeding, consistent with the other three `Validate*` methods, returning `domain.ErrTooManyRequests` on a hit.
- The cache key written by `checkCreateSBOM` inside the `ScanCP` loop matches the key read by `ValidateScanCP` (i.e. normalized `ImageHash`, not raw `scan.ImageID`).
- Within a single `ScanCP` call, once one container's pull records a 429 for a given image, later containers in the same loop referencing that image don't issue another pull.
- New test coverage: a `ScanCP` case where `CreateSBOM` returns a 429 asserts the cache is populated under the same key `ValidateScanCP` reads, and a follow-up call for the same image short-circuits instead of calling `CreateSBOM` again.

Suggested approach:
1. Standardize on `ImageHash` (post-`NormalizeImageID`) as the cache key for `ScanCP`, matching every other flow, at both the read (`ValidateScanCP`) and write (`checkCreateSBOM` call at scan.go:224) sites.
2. Add the same `s.tooManyRequests.Get(...)` fast-fail to `ValidateScanCP` that the other three validators already have.
3. Check the cache per-container inside the `ScanCP` loop (not just once at function entry), since a single request can carry multiple distinct images and a 429 on one shouldn't block unrelated images — but a repeat of the *same* image later in the loop should be skipped.
4. Extend the existing `TestScanService_ValidateScanCP`/`TestScanService_ScanCP` table tests in `scan_test.go` with a 429 case, and consider adding the same for `ValidateScanCVE`/`ValidateScanRegistry` while touching this file, since the whole cache mechanism currently has no end-to-end test coverage.

bhuvan-somisetty:
## Overview

`ScanCP` (the ApplicationProfile/relevancy scan flow) never consulted the `s.tooManyRequests` backoff cache before pulling an image, and the one place it wrote to that cache used the wrong key. Both bugs meant the "back off after a 429" protection that works for `GenerateSBOM`/`ScanCVE`/`ScanRegistry` did nothing at all for `ScanCP` — the one flow most likely to fan out several image pulls per request (one per container in an ApplicationProfile).

Fix, in `core/services/scan.go`:
- `ScanCP`'s per-container loop now records the 429 marker under `subWorkload.ImageHash` (the `NormalizeImageID`-derived key every other flow already uses) instead of the raw `scan.ImageID`, which was a structurally different string nothing ever read back.
- Added a read check right before the `CreateSBOM` call inside that same loop: if the image's `ImageHash` is already marked as rate limited, the container is skipped (with a `ReportScanFailure` call so it's still visible, same as the other skip paths in this function) instead of attempting another pull. This lives inside `ScanCP` itself rather than `ValidateScanCP`, since `ValidateScanCP` runs before the image identity is known (it only has the workload's name/namespace at that point) — the image is only discovered once `GetContainerRelevancyScans` runs inside `ScanCP`.
- This also protects across requests, not just within one loop: since the check reads the same cache `checkCreateSBOM` writes to, a later `ScanCP` request for the same image is gated too, not just the remaining containers in the current request.

## How to Test

- Extended `TestScanService_ScanCP` with a `create SBOM too many requests` case asserting the cache is populated under the correct normalized key after a 429.
- Added `TestScanService_ScanCP_SkipsPullForAlreadyRateLimitedImage`, which pre-populates the cache under the normalized key and asserts (via a small counting `SBOMCreator` wrapper) that `CreateSBOM` is never called for a container whose image is already known to be rate limited.
- `go build ./...`, `go vet ./...`: clean.
- `go test ./...`: clean, except `TestScanService_NginxTest` and `Test_grypeAdapter_DBVersion`, which fail in this environment with `panic: rootless Docker is not supported on Windows` — pre-existing `testcontainers`/Docker-dependent tests unrelated to this change (same failure reproduces on an unmodified checkout). All tests in the touched package (`core/services`) pass.

## Related issues/PRs

Resolved #456

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of SBOM generation rate limits across registry and scanning workflows.
  * Images that have reached a rate limit are skipped to avoid unnecessary processing.
  * Rate-limit failures are reported consistently as HTTP 429 errors.
  * Scanning continues for remaining images when one image is rate limited.
  * Improved image identification ensures rate-limit status is applied consistently.

### DIFF
--- a/adapters/v1/sidecar.go
+++ b/adapters/v1/sidecar.go
@@ -4,9 +4,11 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/http"
 	"sync"
 	"time"
 
+	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
 	"github.com/kubescape/go-logger"
 	"github.com/kubescape/go-logger/helpers"
 	helpersv1 "github.com/kubescape/k8s-interface/instanceidhandler/v1/helpers"
@@ -119,6 +121,14 @@ func (s *SidecarSBOMAdapter) CreateSBOM(ctx context.Context, name, imageID, imag
 	}
 
 	if result.ErrorMessage != "" && result.Status != helpersv1.Learning && result.Status != helpersv1.TooLarge {
+		if result.StatusReason == domain.ReasonTooManyRequests {
+			// StatusReason crossed the gRPC boundary as a plain string, which lost the
+			// original *transport.Error's StatusCode; reconstruct it here so
+			// ScanService.checkCreateSBOM's errors.As(...) check still recognizes this
+			// as a 429 and records the backoff, exactly as it does for the in-process
+			// syft adapter's pull errors.
+			return domainSBOM, fmt.Errorf("%s: %w", result.ErrorMessage, &transport.Error{StatusCode: http.StatusTooManyRequests})
+		}
 		return domainSBOM, fmt.Errorf("%s", result.ErrorMessage)
 	}
 
--- a/adapters/v1/sidecar.go
+++ b/adapters/v1/sidecar.go
@@ -4,11 +4,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"net/http"
 	"sync"
 	"time"
 
-	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
 	"github.com/kubescape/go-logger"
 	"github.com/kubescape/go-logger/helpers"
 	helpersv1 "github.com/kubescape/k8s-interface/instanceidhandler/v1/helpers"
@@ -122,12 +120,10 @@ func (s *SidecarSBOMAdapter) CreateSBOM(ctx context.Context, name, imageID, imag
 
 	if result.ErrorMessage != "" && result.Status != helpersv1.Learning && result.Status != helpersv1.TooLarge {
 		if result.StatusReason == domain.ReasonTooManyRequests {
-			// StatusReason crossed the gRPC boundary as a plain string, which lost the
-			// original *transport.Error's StatusCode; reconstruct it here so
-			// ScanService.checkCreateSBOM's errors.As(...) check still recognizes this
-			// as a 429 and records the backoff, exactly as it does for the in-process
-			// syft adapter's pull errors.
-			return domainSBOM, fmt.Errorf("%s: %w", result.ErrorMessage, &transport.Error{StatusCode: http.StatusTooManyRequests})
+			// StatusReason crossed the gRPC boundary as a plain string; wrap the
+			// domain.ErrTooManyRequests sentinel so ScanService.checkCreateSBOM can
+			// recognize this as a 429 (via errors.Is) and record the backoff.
+			return domainSBOM, fmt.Errorf("%s: %w", result.ErrorMessage, domain.ErrTooManyRequests)
 		}
 		return domainSBOM, fmt.Errorf("%s", result.ErrorMessage)
 	}
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -77,12 +77,36 @@ func NewScanService(sbomCreator ports.SBOMCreator, sbomRepository ports.SBOMRepo
 }
 
 func (s *ScanService) checkCreateSBOM(err error, key string) {
-	if err != nil {
-		var transportError *transport.Error
-		if errors.As(err, &transportError) && transportError.StatusCode == http.StatusTooManyRequests {
-			s.tooManyRequests.Set(key, true, ttl)
-		}
+	if isRegistryRateLimitedErr(err) {
+		s.tooManyRequests.Set(key, true, ttl)
 	}
+}
+
+// isRegistryRateLimitedErr reports whether err indicates the image pull was rate limited.
+// It recognizes three shapes:
+//   - domain.ErrTooManyRequests: the sentinel SidecarSBOMAdapter.CreateSBOM wraps its
+//     returned error with once the sidecar scanner reports a 429 (see adapters/v1/sidecar.go).
+//   - *transport.Error with StatusCode 429: what the in-process syft adapter would return
+//     if the error reached here unwrapped.
+//   - the rendered error text: stereoscope's registry provider formats the
+//     go-containerregistry pull error with %+v, not %w, which severs the errors.As chain
+//     for *transport.Error before it ever reaches here (mirrors
+//     pkg/sbomscanner/v1.isRegistryRateLimited, which has the same problem on the sidecar
+//     side of the same pull path).
+func isRegistryRateLimitedErr(err error) bool {
+	if err == nil {
+		return false
+	}
+	if errors.Is(err, domain.ErrTooManyRequests) {
+		return true
+	}
+	var transportError *transport.Error
+	if errors.As(err, &transportError) && transportError.StatusCode == http.StatusTooManyRequests {
+		return true
+	}
+	errStr := err.Error()
+	return strings.Contains(errStr, "TOOMANYREQUESTS") ||
+		strings.Contains(errStr, "429 Too Many Requests")
 }
 
 // GenerateSBOM implements the "Generate SBOM flow"
@@ -226,7 +250,7 @@ func (s *ScanService) ScanCP(mainCtx context.Context) error {
 						logger.L().Ctx(ctx).Warning("skipping SBOM creation, image pull previously rate limited",
 							helpers.String("imageSlug", slug))
 						_ = s.platform.ReportScanFailure(ctx, scanfailure.ScanFailureSBOMGeneration,
-							classifySBOMError(domain.ErrTooManyRequests), domain.ErrTooManyRequests)
+							scanfailure.ReasonSBOMGenerationFailed, domain.ErrTooManyRequests)
 						continue
 					}
 					// create SBOM
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -219,9 +219,19 @@ func (s *ScanService) ScanCP(mainCtx context.Context) error {
 			// if SBOM is not available, create it
 			if sbom.Content == nil {
 				if s.sbomGeneration {
+					// a previous container in this loop (or a previous request) may have
+					// already recorded this exact image as rate limited; skip pulling it
+					// again instead of hammering the registry with another attempt
+					if _, ok := s.tooManyRequests.Get(subWorkload.ImageHash); ok {
+						logger.L().Ctx(ctx).Warning("skipping SBOM creation, image pull previously rate limited",
+							helpers.String("imageSlug", slug))
+						_ = s.platform.ReportScanFailure(ctx, scanfailure.ScanFailureSBOMGeneration,
+							classifySBOMError(domain.ErrTooManyRequests), domain.ErrTooManyRequests)
+						continue
+					}
 					// create SBOM
 					sbom, err = s.sbomCreator.CreateSBOM(ctx, subWorkload.ImageSlug, subWorkload.ImageHash, subWorkload.ImageTagNormalized, optionsFromWorkload(ctx, subWorkload))
-					s.checkCreateSBOM(err, scan.ImageID)
+					s.checkCreateSBOM(err, subWorkload.ImageHash)
 					if err != nil {
 						logger.L().Ctx(ctx).Error("creating SBOM, skipping scan", helpers.Error(err),
 							helpers.String("imageSlug", slug))
--- a/pkg/sbomscanner/v1/server.go
+++ b/pkg/sbomscanner/v1/server.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net/http"
 	"runtime"
 	"runtime/debug"
 	"strings"
@@ -21,6 +22,7 @@ import (
 	"github.com/anchore/syft/syft/sbom"
 	"github.com/anchore/syft/syft/source"
 	"github.com/eapache/go-resiliency/deadline"
+	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
 	"github.com/kubescape/go-logger"
 	"github.com/kubescape/go-logger/helpers"
 	helpersv1 "github.com/kubescape/k8s-interface/instanceidhandler/v1/helpers"
@@ -33,6 +35,12 @@ import (
 func isGCPRegistry(imageID string) bool {
 	host, _, _ := strings.Cut(imageID, "/")
 	return host == "gcr.io" || strings.HasSuffix(host, ".gcr.io") || strings.HasSuffix(host, "-docker.pkg.dev")
+}
+
+// isRegistryRateLimited reports whether err is (or wraps) a registry 429 response.
+func isRegistryRateLimited(err error) bool {
+	var transportErr *transport.Error
+	return errors.As(err, &transportErr) && transportErr.StatusCode == http.StatusTooManyRequests
 }
 
 func gcpCredentials(ctx context.Context) (*image.RegistryCredentials, error) {
@@ -182,6 +190,15 @@ func (s *scannerServer) CreateSBOM(ctx context.Context, req *pb.CreateSBOMReques
 		return &pb.CreateSBOMResponse{
 			Status:       helpersv1.Unauthorize,
 			ErrorMessage: err.Error(),
+		}, nil
+	case err != nil && isRegistryRateLimited(err):
+		// StatusReason travels over gRPC as a plain string, so the caller (adapters/v1
+		// SidecarSBOMAdapter.CreateSBOM) reconstructs a *transport.Error from it to keep
+		// ScanService.checkCreateSBOM's errors.As(...) check working the same way it does
+		// for the in-process syft adapter.
+		return &pb.CreateSBOMResponse{
+			ErrorMessage: err.Error(),
+			StatusReason: domain.ReasonTooManyRequests,
 		}, nil
 	case err != nil:
 		return &pb.CreateSBOMResponse{
--- a/pkg/sbomscanner/v1/server.go
+++ b/pkg/sbomscanner/v1/server.go
@@ -38,9 +38,25 @@ func isGCPRegistry(imageID string) bool {
 }
 
 // isRegistryRateLimited reports whether err is (or wraps) a registry 429 response.
+//
+// The typed check alone is not enough: stereoscope's registry provider formats the
+// go-containerregistry pull error with %+v, not %w (see
+// pkg/image/oci/registry_provider.go), which severs the errors.As chain before it ever
+// reaches here. Fall back to matching the rendered text — "TOOMANYREQUESTS" is the stable
+// registry error code emitted when the response carries a JSON error body (e.g. Docker
+// Hub's rate-limit response), and "429 Too Many Requests" is *transport.Error's own
+// Error() text when the response body was empty.
 func isRegistryRateLimited(err error) bool {
+	if err == nil {
+		return false
+	}
 	var transportErr *transport.Error
-	return errors.As(err, &transportErr) && transportErr.StatusCode == http.StatusTooManyRequests
+	if errors.As(err, &transportErr) && transportErr.StatusCode == http.StatusTooManyRequests {
+		return true
+	}
+	errStr := err.Error()
+	return strings.Contains(errStr, "TOOMANYREQUESTS") ||
+		strings.Contains(errStr, "429 Too Many Requests")
 }
 
 func gcpCredentials(ctx context.Context) (*image.RegistryCredentials, error) {
--- a/adapters/mocksbom.go
+++ b/adapters/mocksbom.go
@@ -41,7 +41,12 @@ func (m MockSBOMAdapter) CreateSBOM(_ context.Context, name, imageID, imageTag s
 		return domain.SBOM{}, domain.ErrMockError
 	}
 	if m.toomanyrequests {
-		return domain.SBOM{}, fmt.Errorf("failed to get image descriptor from registry: %w",
+		// %+v, not %w, mirrors stereoscope's real registry_provider.go wrapping: the
+		// typed *transport.Error does not survive as an unwrappable error, only its
+		// rendered text does. checkCreateSBOM/isRegistryRateLimited must be able to
+		// recognize a 429 from that text alone, or this test double would mask exactly
+		// the bug it's meant to catch.
+		return domain.SBOM{}, fmt.Errorf("failed to get image descriptor from registry: %+v",
 			&transport.Error{
 				StatusCode: http.StatusTooManyRequests,
 			},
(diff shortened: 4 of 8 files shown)


core/services/scan.go
```diff
@@ -219,9 +219,19 @@ func (s *ScanService) ScanCP(mainCtx context.Context) error {
...
+						logger.L().Ctx(ctx).Warning("skipping SBOM creation, image pull previously rate limited",
+							helpers.String("imageSlug", slug))
+						_ = s.platform.ReportScanFailure(ctx, scanfailure.ScanFailureSBOMGeneration,
+							classifySBOMError(domain.ErrTooManyRequests), domain.ErrTooManyRequests)
+						continue
+					}
 					// create SBOM
 					sbom, err = s.sbomCreator.CreateSBOM(ctx, subWorkload.ImageSlug, subWorkload.ImageHash, subWorkload.ImageTagNormalized, optionsFromWorkload(ctx, subWorkload))
-					s.checkCreateSBOM(err, scan.ImageID)
+					s.checkCreateSBOM(err, subWorkload.ImageHash)
```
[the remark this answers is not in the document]
bhuvan-somisetty: Good catch, fixed in 7381e01. server.go now special-cases a 429 the same way it already does for image-too-large/401 and sets StatusReason to too-many-requests instead of letting it fall into the generic ErrorMessage-only branch. sidecar.go then rebuilds a *transport.Error from that reason before returning, so checkCreateSBOM's errors.As check works the same for both the sidecar and in-process syft paths. Added TestIsRegistryRateLimited (server.go) and TestSidecarSBOMAdapter_CreateSBOM_TooManyRequests (sidecar.go) to cover the round trip.

bhuvan-somisetty:
@matthyx PTAL, whenever u get Chance

pkg/sbomscanner/v1/server.go
```diff
@@ -35,6 +37,12 @@ func isGCPRegistry(imageID string) bool {
 	return host == "gcr.io" || strings.HasSuffix(host, ".gcr.io") || strings.HasSuffix(host, "-docker.pkg.dev")
 }
 
+// isRegistryRateLimited reports whether err is (or wraps) a registry 429 response.
+func isRegistryRateLimited(err error) bool {
+	var transportErr *transport.Error
+	return errors.As(err, &transportErr) && transportErr.StatusCode == http.StatusTooManyRequests
```
matthyx: **Blocker (see review body).** `errors.As` never matches here for a real pull: stereoscope's `registry_provider.go:69` formats the go-containerregistry error with `%+v`, so the `*transport.Error` is gone before `syft.GetSource` returns. This case is therefore dead code in production, and with it the whole `StatusReason` → reconstructed-`transport.Error` round trip added in commit 2.

Suggest keeping the typed check and adding a text fallback, mirroring the `"401 Unauthorized"` case just above:

```go
func isRegistryRateLimited(err error) bool {
	if err == nil {
		return false
	}
	var transportErr *transport.Error
	if errors.As(err, &transportErr) && transportErr.StatusCode == http.StatusTooManyRequests {
		return true
	}
	// the typed error does not survive stereoscope's %+v wrapping, so fall back to the
	// rendered text — "TOOMANYREQUESTS" is the stable registry error code (Docker Hub),
	// the status-line form appears when the response carried no JSON error body
	errStr := err.Error()
	return strings.Contains(errStr, "TOOMANYREQUESTS") ||
		strings.Contains(errStr, "429 Too Many Requests")
}
```

Worth extending `TestIsRegistryRateLimited` with the real rendered chain (`fmt.Errorf("failed to get image descriptor from registry: %+v", &transport.Error{StatusCode: 429})`) rather than only the `%w` form — that case currently asserts `false` under the guise of "plain error is not rate limiting".

core/services/scan.go
matthyx, line 258: Same root cause as the `server.go` blocker, on the in-process syft path. Fixing the key here is right, but `checkCreateSBOM` (`scan.go:82`) can't recognise a real registry 429 either — stereoscope's `%+v` wrapping strips the `*transport.Error` — so nothing ever calls `s.tooManyRequests.Set` and the gate added at line 225 reads an always-empty cache for the syft deployment.

The same text fallback needs to land in `checkCreateSBOM`, otherwise the fix only works for the sidecar deployment (and only once `isRegistryRateLimited` is fixed too).

core/services/scan.go
```diff
@@ -219,9 +219,19 @@ func (s *ScanService) ScanCP(mainCtx context.Context) error {
...
 			if sbom.Content == nil {
 				if s.sbomGeneration {
+					// a previous container in this loop (or a previous request) may have
+					// already recorded this exact image as rate limited; skip pulling it
+					// again instead of hammering the registry with another attempt
+					if _, ok := s.tooManyRequests.Get(subWorkload.ImageHash); ok {
+						logger.L().Ctx(ctx).Warning("skipping SBOM creation, image pull previously rate limited",
+							helpers.String("imageSlug", slug))
+						_ = s.platform.ReportScanFailure(ctx, scanfailure.ScanFailureSBOMGeneration,
+							classifySBOMError(domain.ErrTooManyRequests), domain.ErrTooManyRequests)
```
matthyx: Non-blocking: `classifySBOMError(domain.ErrTooManyRequests)` resolves to `scanfailure.ReasonSBOMGenerationFailed` — none of its branches match a bare `errors.New("too many requests")`. Backend-side that renders as *"Failed to generate software inventory (SBOM) for this image"*, which is misleading for a deliberate back-off skip where no pull was even attempted, and it will be emitted once per skipped container per `ScanCP` request for the full 10-minute `ttl`.

Two things worth considering:

- Write the constant directly (`scanfailure.ReasonSBOMGenerationFailed`) instead of round-tripping a sentinel through `classifySBOMError` — as written the reason actually sent isn't obvious at the call site, and it silently changes if `classifySBOMError` gains a 429 branch later.
- Longer term, `armoapi-go/scanfailure` has no rate-limit reason code; adding one (and a `classifySBOMError` branch for 429) would let both this skip path and the real failure path at line 238 report something actionable instead of a generic SBOM failure.

adapters/v1/sidecar.go
```diff
@@ -119,6 +121,14 @@ func (s *SidecarSBOMAdapter) CreateSBOM(ctx context.Context, name, imageID, imag
 	}
 
 	if result.ErrorMessage != "" && result.Status != helpersv1.Learning && result.Status != helpersv1.TooLarge {
+		if result.StatusReason == domain.ReasonTooManyRequests {
+			// StatusReason crossed the gRPC boundary as a plain string, which lost the
+			// original *transport.Error's StatusCode; reconstruct it here so
+			// ScanService.checkCreateSBOM's errors.As(...) check still recognizes this
+			// as a 429 and records the backoff, exactly as it does for the in-process
+			// syft adapter's pull errors.
+			return domainSBOM, fmt.Errorf("%s: %w", result.ErrorMessage, &transport.Error{StatusCode: http.StatusTooManyRequests})
```
matthyx: Non-blocking design note. Synthesising a `*transport.Error` purely so a downstream `errors.As` matches works, but it encodes a boundary-crossing signal as a fake HTTP error, and it renders a slightly odd message (`received status code: 429: unexpected status code 429 Too Many Requests`).

Since `checkCreateSBOM` needs touching anyway for the blocker above, a domain sentinel would be cleaner and keeps the rate-limit predicate in one place:

```go
return domainSBOM, fmt.Errorf("%s: %w", result.ErrorMessage, domain.ErrTooManyRequests)
```

with `checkCreateSBOM` accepting either `errors.Is(err, domain.ErrTooManyRequests)` or the typed/text registry check. Entirely your call — the current form is functional once detection is fixed.

matthyx (commented):
Reviewed the full change at `aa8f634`. The two bugs it targets — the write-key mismatch (`scan.ImageID` vs. the normalized `ImageHash`) and the missing read check inside `ScanCP`'s per-container loop — are correctly identified and correctly fixed. I confirmed the key convention independently: `controllers/http.go:196` builds `ImageHash` with `v1.NormalizeImageID`, so `subWorkload.ImageHash` is indeed the key `ValidateGenerateSBOM`/`ValidateScanCVE` read. `go build ./...`, `go vet ./...` and `go test ./core/services/... ./adapters/v1/... ./pkg/sbomscanner/...` are all clean here.

I also agree with the decision to skip issue #456's first acceptance criterion (a `tooManyRequests` read inside `ValidateScanCP`). That validator only ever receives `name`/`namespace`; the image identity doesn't exist until `GetContainerRelevancyScans` runs inside `ScanCP`. The criterion isn't implementable as written, so the issue should be amended rather than the code. The CodeRabbit pre-merge warning about it can be dismissed.

## Blocker: the 429 is never detected in production, so the cache this PR now reads is still never written

`checkCreateSBOM` (`core/services/scan.go:82`) and the new `isRegistryRateLimited` (`pkg/sbomscanner/v1/server.go:43`) both identify a rate limit via `errors.As(err, &transportErr)`. That type does not survive the pull path. stereoscope's registry provider formats the go-containerregistry error with `%+v`, not `%w`:

```go
// github.com/matthyx/stereoscope@v0.0.0-20250916161743-dd57158479de
// pkg/image/oci/registry_provider.go:69   (the fork this repo `replace`s to; upstream v0.1.9 is identical)
descriptor, err := remote.Get(ref, options...)
if err != nil {
    return nil, fmt.Errorf("failed to get image descriptor from registry: %+v", err)
}
```

The chain is cut there, before `syft.GetSource` returns to either adapter. Reproduced against this branch:

```go
real429 := &transport.Error{StatusCode: http.StatusTooManyRequests}
// stereoscope registry_provider.go:69
stereoscopeErr := fmt.Errorf("failed to get image descriptor from registry: %+v", real429)
// syft get_source.go sourceError(), single-error path
syftErr := fmt.Errorf("an error occurred attempting to resolve '%s': %w", "img",
	fmt.Errorf("%s: %w", "registry", stereoscopeErr))

var te *transport.Error
errors.As(syftErr, &te)   // false
s.checkCreateSBOM(syftErr, "some-image-hash")
_, ok := s.tooManyRequests.Get("some-image-hash")   // false
```

```
rendered: an error occurred attempting to resolve 'img': registry: failed to get image descriptor from registry: unexpected status code 429 Too Many Requests
errors.As matched: false
backoff recorded for a REAL registry 429: false
```

Consequences:

- **In-process syft adapter:** `checkCreateSBOM` never matches, so `ScanCP`'s new gate reads a cache nothing populates. The backoff stays inert for this flow — exactly what the PR set out to fix.
- **Sidecar adapter:** `isRegistryRateLimited` never matches, so `StatusReason` is never set to `ReasonTooManyRequests`, so `SidecarSBOMAdapter.CreateSBOM` never reconstructs the `*transport.Error`. The second commit's round trip is correct in itself but is never entered.

This is also why the case directly above the new one in `server.go` matches `"401 Unauthorized"` as a *string*, and why `adapters/v1/syft.go` does the same for `MANIFEST_UNKNOWN`: `%+v` renders `transport.Error.Error()` into the message text, and the text is all that survives.

The test suite doesn't catch this because both doubles bypass the real chain — `adapters/mocksbom.go:44` wraps with `%w` where reality uses `%+v`, and the new `server_test.go`/`sidecar_test.go` cases inject a `*transport.Error` directly.

### Suggested fix (small, and consistent with the surrounding code)

1. Give `isRegistryRateLimited` a string fallback beside the typed check. Registries render 429 two ways depending on whether the response carried a JSON error body: `unexpected status code 429 Too Many Requests` (empty `Errors`) and `TOOMANYREQUESTS: You have reached your pull rate limit...` (populated `Errors` — this is Docker Hub's form). Match both.
2. Apply the same fallback in `checkCreateSBOM`, otherwise the in-process syft deployment still records nothing and the new `ScanCP` gate is dead there.
3. Change `adapters/mocksbom.go` to wrap with `%+v` exactly as stereoscope does, so this class of regression is caught by the existing tests instead of hidden by them.

With that in place the rest of the change is, as far as I can tell, ready to merge. Two smaller non-blocking notes inline.

pkg/sbomscanner/v1/server.go
bhuvan-somisetty: Confirmed, thanks for catching this — fixed in 0f1a208. isRegistryRateLimited now falls back to matching the rendered text ("TOOMANYREQUESTS" for the Docker-Hub-style JSON-body form, "429 Too Many Requests" for transport.Error's own text when the body is empty) alongside the typed check. Extended TestIsRegistryRateLimited with cases that reproduce the actual stereoscope %+v chain (both the empty-body and TOOMANYREQUESTS-code forms) so a regression here would be caught again.

core/services/scan.go
bhuvan-somisetty: Same fix applied, 0f1a208. checkCreateSBOM now goes through a shared isRegistryRateLimitedErr helper that checks, in order: errors.Is(err, domain.ErrTooManyRequests) for the sidecar's sentinel, the typed transport.Error check, then the same text fallback as server.go's isRegistryRateLimited. Added TestIsRegistryRateLimitedErr covering all four shapes, and switched adapters/mocksbom.go's 429 test double from %w to %+v so the existing ScanCVE/ScanCP/GenerateSBOM 429 test cases actually exercise the real stereoscope wrapping instead of masking it.

core/services/scan.go
bhuvan-somisetty: Fixed too, 0f1a208 — now writes scanfailure.ReasonSBOMGenerationFailed directly instead of round-tripping through classifySBOMError, so the reason actually sent is visible at the call site and won't silently change if classifySBOMError ever grows a 429 branch. Agreed on the longer-term ask for a dedicated rate-limit reason code in armoapi-go/scanfailure — that's a change in a vendored dependency, so left it as a follow-up rather than folding it into this PR.

adapters/v1/sidecar.go
bhuvan-somisetty: Took this option, 0f1a208 — switched to fmt.Errorf("%s: %w", result.ErrorMessage, domain.ErrTooManyRequests) and dropped the transport.Error synthesis entirely (along with the now-unused transport/net-http imports in this file). checkCreateSBOM checks errors.Is(err, domain.ErrTooManyRequests) first, so the predicate stays in one place as you suggested. Updated TestSidecarSBOMAdapter_CreateSBOM_TooManyRequests to assert errors.Is instead of errors.As.

matthyx (approved):
Re-reviewed at `0f1a208`. The blocker is fixed — approving.

`isRegistryRateLimitedErr` (`core/services/scan.go`) and `isRegistryRateLimited` (`pkg/sbomscanner/v1/server.go`) now both fall back to the rendered text, so a real pull failure is recognised despite stereoscope severing the `errors.As` chain with `%+v`. I re-ran my reproduction of the actual production chain (`syft.GetSource` → `sourceError` → stereoscope `registry_provider.go:69`) against this head:

```
empty body 429                         recorded=true   ...: unexpected status code 429 Too Many Requests
docker hub TOOMANYREQUESTS body        recorded=true   ...: TOOMANYREQUESTS: You have reached your pull rate limit.
401 then 429 on anonymous fallback     recorded=true
NEGATIVE: 401 only                     recorded=false
NEGATIVE: manifest unknown             recorded=false
```

Both 429 renderings are caught, and neither of the two error shapes that flow through the *same* `%+v` wrapper false-positives. The third row is a case I didn't raise but was worth checking: when the initial pull 401s and the anonymous retry 429s, `resolveSource` merges both texts and the `401 Unauthorized` case wins the switch in `server.go`, so no `StatusReason` is set — but the service-level text check still sees the 429 in the merged message and records the backoff. It degrades in the right direction.

The other two points are addressed as well, and the choices you made on both are better than what I suggested:

- Replacing the synthesized `*transport.Error` with a wrapped `domain.ErrTooManyRequests` sentinel across the gRPC boundary is the cleaner half of the alternative I offered — the signal is now a domain concept rather than a fabricated HTTP error, and `isRegistryRateLimitedErr` keeps all three shapes in one documented place.
- `adapters/mocksbom.go` switching to `%+v` is the change that actually matters for regression safety: the double now reproduces the real error shape, so the tests would catch a revert to typed-only detection instead of hiding it. Same for the `%+v` cases added to `TestIsRegistryRateLimited` / `TestIsRegistryRateLimitedErr`, including the `TooManyRequestsErrorCode` diagnostic form.
- `scanfailure.ReasonSBOMGenerationFailed` written directly on the skip path — reason sent is now obvious at the call site.

Verification on my side at `0f1a208`: `go build ./...` and `go vet ./...` clean, and the **full** `go test ./...` is green — 10/10 packages, no failures. The two testcontainers-dependent tests you saw fail on Windows (`TestScanService_NginxTest`, `Test_grypeAdapter_DBVersion`) pass here, so those were environmental as you suspected.

Two follow-ups for later, neither blocking this PR:

1. The text matching is load-bearing only because stereoscope drops the error chain. A one-line `%+v` → `%w` fix in `pkg/image/oci/registry_provider.go` in the `matthyx/stereoscope` fork would let the typed checks work for real and make the string fallbacks pure belt-and-braces — worth an upstream PR at some point. The comments you added explain the situation well enough in the meantime.
2. `armoapi-go/scanfailure` still has no rate-limit reason code, so a back-off skip is reported to customers as "Failed to generate software inventory (SBOM) for this image". Worth adding one there and giving `classifySBOMError` a 429 branch, which would improve both this skip path and the real failure path.

Also worth noting for whoever merges: CodeRabbit's remaining pre-merge warning ("omits the required ValidateScanCP cache read") should be dismissed rather than acted on — `ValidateScanCP` only ever receives `name`/`namespace`, so that acceptance criterion in #456 isn't implementable as written. The gate belongs exactly where you put it.

bhuvan-somisetty:
Thanks @matthyx!

PR #460, opened 2026-08-05
test: expand context propagation test coverage (#436)

NaitikRishu:
## Overview
This PR expands the context propagation testing coverage in `apiserver_test.go` to ensure that context is correctly propagated through all `Get`/`Create`/`Update` calls made by the `APIServerStore` methods, addressing the gaps identified in #436. 

**Current behavior**: The tests only asserted context propagation for the initial (happy path) calls made by some `Store*` methods. Several sub-clients (`SBOMSyftFiltereds`, `ContainerProfiles`) lacked test wrappers entirely, and conflict-retry paths for all `Store*` methods were completely unasserted for context referential integrity.

**Future behavior**: 
- Added missing context capture wrappers for `SBOMSyftFiltereds` and `ContainerProfiles`.
- Expanded all `ctxCapturing*` wrappers to capture both `createCtx` and `updateCtx` (in addition to `getCtx`).
- Added new `RetryConflict_ctxPropagated` test cases for all `Store*` operations. These seed a conflicting object in the fake clientset to force the storage method into its retry loop, ensuring the `updateCtx` and `getCtx` calls inside the loop retain the original passed context.
- Implemented map-containment assertions (`require.Contains`) before accessing context wrappers in all tests to improve failure readability and prevent nil-pointer panics.

## Additional Information
This builds upon the pattern established in #422. The test setup carefully separates the initial store (which creates the object) from the second store (which is given the canary context and forced to retry over a conflict) to precisely verify propagation.

## How to Test
1. Run the tests in the `repositories` package:
   ```bash
   go test -v ./repositories
   ```
2. Observe that all `_ctxPropagated` tests, including the new `_RetryConflict_ctxPropagated` tests, pass successfully.

## Related issues/PRs:
* Partially addresses #436

```



## Summary by CodeRabbit

* **Tests**
  * Expanded coverage for vulnerability summaries, filtered software inventories, and container profiles.
  * Added validation for consistent request context handling across read, write, storage, and retry scenarios.
  * Verified reliable namespace-specific access and reuse across supported operations.
  * Strengthened checks for conflict handling, VEX operations, and related service interactions.

### DIFF
--- a/repositories/apiserver_test.go
+++ b/repositories/apiserver_test.go
@@ -1926,6 +1926,7 @@ func TestAPIServerStore_StoreVEX_RetryConflict_ctxPropagated(t *testing.T) {
 
 	requireCanaryCtx(t, wrapped.ovecs[a.Namespace].getCtx)
 	requireCanaryCtx(t, wrapped.ovecs[a.Namespace].updateCtx)
+	require.Equal(t, 1, conflicts)
 }
 
 func TestAPIServerStore_StoreCVE_RetryConflict_ctxPropagated(t *testing.T) {
@@ -1958,6 +1959,7 @@ func TestAPIServerStore_StoreCVE_RetryConflict_ctxPropagated(t *testing.T) {
 	requireCanaryCtx(t, wrapped.vulnManifests[a.Namespace].getCtx)
 	requireCanaryCtx(t, wrapped.vulnManifests[a.Namespace].createCtx)
 	requireCanaryCtx(t, wrapped.vulnManifests[a.Namespace].updateCtx)
+	require.Equal(t, 1, conflicts)
 }
 
 func TestAPIServerStore_StoreSBOM_RetryConflict_ctxPropagated(t *testing.T) {
@@ -1990,6 +1992,7 @@ func TestAPIServerStore_StoreSBOM_RetryConflict_ctxPropagated(t *testing.T) {
 	requireCanaryCtx(t, wrapped.sbomSyfts[a.Namespace].getCtx)
 	requireCanaryCtx(t, wrapped.sbomSyfts[a.Namespace].createCtx)
 	requireCanaryCtx(t, wrapped.sbomSyfts[a.Namespace].updateCtx)
+	require.Equal(t, 1, conflicts)
 }
 
 func TestAPIServerStore_StoreSBOMFiltered_RetryConflict_ctxPropagated(t *testing.T) {
@@ -2022,6 +2025,7 @@ func TestAPIServerStore_StoreSBOMFiltered_RetryConflict_ctxPropagated(t *testing
 	requireCanaryCtx(t, wrapped.sbomSyftFiltereds[a.Namespace].getCtx)
 	requireCanaryCtx(t, wrapped.sbomSyftFiltereds[a.Namespace].createCtx)
 	requireCanaryCtx(t, wrapped.sbomSyftFiltereds[a.Namespace].updateCtx)
+	require.Equal(t, 1, conflicts)
 }
 
 func TestAPIServerStore_StoreCVESummary_RetryConflict_ctxPropagated(t *testing.T) {
@@ -2065,6 +2069,7 @@ func TestAPIServerStore_StoreCVESummary_RetryConflict_ctxPropagated(t *testing.T
 	requireCanaryCtx(t, wrapped.vulnSummaries["anyNamespaceJob"].getCtx)
 	requireCanaryCtx(t, wrapped.vulnSummaries["anyNamespaceJob"].createCtx)
 	requireCanaryCtx(t, wrapped.vulnSummaries["anyNamespaceJob"].updateCtx)
+	require.Equal(t, 1, conflicts)
 }
 
 func TestAPIServerStore_StoreCVESummaryStub_RetryConflict_ctxPropagated(t *testing.T) {
@@ -2107,6 +2112,7 @@ func TestAPIServerStore_StoreCVESummaryStub_RetryConflict_ctxPropagated(t *testi
 	requireCanaryCtx(t, wrapped.vulnSummaries["anyNamespaceJob"].getCtx)
 	requireCanaryCtx(t, wrapped.vulnSummaries["anyNamespaceJob"].createCtx)
 	requireCanaryCtx(t, wrapped.vulnSummaries["anyNamespaceJob"].updateCtx)
+	require.Equal(t, 1, conflicts)
 }
 
 func TestCtxCapturingClient_wrappersKeyedByNamespace(t *testing.T) {
--- a/repositories/apiserver_test.go
+++ b/repositories/apiserver_test.go
@@ -1555,7 +1555,7 @@ func (w *ctxCapturingOVECs) Update(ctx context.Context, obj *v1beta1.OpenVulnera
 
 type ctxCapturingVulnerabilityManifestSummaries struct {
 	spdxv1beta1.VulnerabilityManifestSummaryInterface
-	getCtx context.Context
+	getCtx, createCtx, updateCtx context.Context
 }
 
 func (w *ctxCapturingVulnerabilityManifestSummaries) Get(ctx context.Context, name string, opts metav1.GetOptions) (*v1beta1.VulnerabilityManifestSummary, error) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by matthyx — CHANGES_REQUESTED** [review]
Reviewed by checking out the branch and running the tests, plus mutation testing to confirm the new assertions actually bite. The substance here is good — this closes the real gap from #436, and I verified that empirically rather than by reading. Two things to fix before merge, one of them trivial.

## Verification I ran

All 16 `_ctxPropagated` tests pass, and the package is green.

**Mutation test 1 — are the tests vacuous?** I broke ctx propagation across all 25 `StorageClient` call sites in `apiserver.go` (`ctx` → `context.Background()`). All 16 tests failed. Not vacuous.

**Mutation test 2 — what do the *new* tests uniquely catch?** I reverted, then broke ctx on `.Update(ctx,` only (6 sites). Exactly six tests failed, and they were precisely the six new ones:

```
--- FAIL: TestAPIServerStore_StoreVEX_RetryConflict_ctxPropagated
--- FAIL: TestAPIServerStore_StoreCVE_RetryConflict_ctxPropagated
--- FAIL: TestAPIServerStore_StoreSBOM_RetryConflict_ctxPropagated
--- FAIL: TestAPIServerStore_StoreSBOMFiltered_RetryConflict_ctxPropagated
--- FAIL: TestAPIServerStore_StoreCVESummary_ctxPropagated
--- FAIL: TestAPIServerStore_StoreCVESummaryStub_RetryConflict_ctxPropagated
```

Every pre-existing test still passed. That is exactly the hole #436 described — *"`updateCtx` is captured by three wrappers and asserted by nothing"* — so the new tests earn their place. Also confirmed the branch merges cleanly with current `main`, and `main` has not touched `apiserver.go` or `apiserver_test.go` since this branched.

## Blocker: the file is not gofmt-clean

`gofmt -l repositories/` flags `repositories/apiserver_test.go`. The new fields in `ctxCapturingClient` are over-aligned by one space, which pushes the whole block off:

```diff
-	vulnManifests      map[string]*ctxCapturingVulnerabilityManifests
-	sbomSyfts          map[string]*ctxCapturingSBOMSyfts
-	ovecs              map[string]*ctxCapturingOVECs
-	vulnSummaries      map[string]*ctxCapturingVulnerabilityManifestSummaries
-	sbomSyftFiltereds  map[string]*ctxCapturingSBOMSyftFiltereds
-	containerProfiles  map[string]*ctxCapturingContainerProfiles
+	vulnManifests     map[string]*ctxCapturingVulnerabilityManifests
+	sbomSyfts         map[string]*ctxCapturingSBOMSyfts
+	ovecs             map[string]*ctxCapturingOVECs
+	vulnSummaries     map[string]*ctxCapturingVulnerabilityManifestSummaries
+	sbomSyftFiltereds map[string]*ctxCapturingSBOMSyftFiltereds
+	containerProfiles map[string]*ctxCapturingContainerProfiles
```

Fix is `gofmt -w repositories/apiserver_test.go`. Flagging it as a blocker because we don't run a fmt/lint gate in CI, so nothing else will catch it and it will sit unformatted and generate diff noise on every future edit to that struct.

## Should fix: the `_RetryConflict_` names claim more than the tests do

These tests never enter a retry. I instrumented the sub-client with call counters and ran the exact scenario the tests set up (store once, then store again):

```
after 1st store: gets=0 creates=1 updates=0
after 2nd store: gets=1 creates=1 updates=1
```

One `Update` means `RetryOnConflict`'s closure ran exactly once. The second `Create` fails with `AlreadyExists`, the closure does `Get` + `Update`, the `Update` succeeds, and the loop exits. No `Conflict` is ever returned, so the retry loop never iterates. What is actually covered is the `AlreadyExists` → update branch — which is worth covering, and which #436's *Solution* section does ask for ("seed an existing object so the `AlreadyExists` → `RetryOnConflict` branch runs"). The coverage is fine; the naming is what is wrong, and a misleading test name is worse than a missing one because the next person reading it will believe the retry path is protected when it isn't.

Either rename to something like `_AlreadyExistsUpdate_ctxPropagated`, or force a genuine conflict. I checked that the reactor route works — with a `PrependReactor` returning `apierrors.NewConflict` once on `update`, the same scenario becomes `gets=2 creates=1 updates=2`, i.e. the loop really iterates:

```go
conflicts := 0
clientset.PrependReactor("update", "vulnerabilitymanifests", func(action k8stesting.Action) (bool, runtime.Object, error) {
	if conflicts < 1 {
		conflicts++
		return true, nil, apierrors.NewConflict(
			schema.GroupResource{Group: "spdx.softwarecomposition.kubescape.io", Resource: "vulnerabilitymanifests"}, name, nil)
	}
	return false, nil, nil
})
```

Renaming is perfectly acceptable if you'd rather keep this PR tight — I'd just like the names to match reality.

## Notes, not blockers

- The `ctx.Err()` guards at the top of each retry closure are still unexercised. #436 framed that as additive ("would additionally cover"), so I'm fine deferring it — but given that and the point above, "Resolved #436" slightly overstates it. Consider "Partially addresses #436" and leaving it open for the reactor + cancelled-ctx follow-up.
- In `TestAPIServerStore_StoreVEX_RetryConflict_ctxPropagated`, the asserted `getCtx` comes from the *pre-loop* `Get` at `apiserver.go:716`, not from a `Get` inside the closure — on the second store `haveExisting` is true, so the closure reuses `existing` and skips its own `Get`. The assertion is still correct and the `updateCtx` one is the load-bearing part; just noting it isn't testing what the name suggests. Not asserting `createCtx` there is correct, since `createVEX` isn't reached on that path.
- Minor: `StoreSBOM_Filtered_ctxPropagated` vs `StoreSBOMFiltered_RetryConflict_ctxPropagated` — pick one spelling.

The `require.Contains` guards, the namespace-keyed wrapper coverage in `TestCtxCapturingClient_wrappersKeyedByNamespace`, and using the workload namespace for the summaries tests are all correct. Fix the gofmt and settle the naming and I'll approve.

NaitikRishu:
I've addressed both items in the latest push:

1. **`gofmt` Blocker:** Ran `gofmt -w repositories/apiserver_test.go` to fix the struct field alignment.
2. **Retry Tests:** Added the `PrependReactor` logic to all six `_RetryConflict_` tests to force a genuine update conflict, ensuring the retry loop actually iterates and asserts context propagation across multiple calls.
3. **Naming & Description:** Fixed the spelling consistency on `StoreSBOMFiltered_ctxPropagated` and updated the PR description to "Partially addresses #436" so the issue remains open for any follow-up cancelled-context work.

All tests are passing locally.

matthyx (approved):
Both items addressed. Re-verified at `b373905` with the same checks I ran the first time, so the comparison is apples-to-apples.

**gofmt** — `gofmt -l repositories/` is now clean, and the `ctxCapturingClient` block is correctly aligned.

**The retry loop now genuinely iterates.** This was the part worth re-measuring rather than trusting. I re-ran my call-counting probe against the exact setup in `StoreCVE_RetryConflict`:

```
before (f99f523): gets=1 creates=1 updates=1 conflicts=0   <- closure ran once, never retried
after  (b373905): gets=2 creates=1 updates=2 conflicts=1   <- real conflict, real second attempt
```

Two `Update` attempts with `conflicts=1` and a `NoError` result is exactly the invariant that proves `RetryOnConflict` went around the loop. All six `_RetryConflict_` tests have the update reactor, and the `require.Equal(t, 1, conflicts)` you added is a nice touch I hadn't asked for — it stops the test from silently degrading back to the single-pass behaviour if the reactor ever stops matching the resource name. Worth keeping.

**Mutation testing, re-run on the new head:**

| mutation | result |
|---|---|
| break ctx on all 25 `StorageClient` call sites | all 16 tests fail |
| break ctx on `.Update(` only | exactly the 6 `_RetryConflict_` tests fail |
| break ctx on the retry-loop's post-conflict `Get` | `StoreCVE_RetryConflict` fails |

That third row is the one that matters most, and it's new. That `Get` only executes on a second loop iteration, so before this fix no test could have reached it — the mutation would have passed unnoticed. It now fails. That's the coverage #436 was actually asking for, and it's real rather than nominal.

Also confirmed: `go vet` and `go build ./...` clean, full `repositories` package green, branch is up to date with `main` (`main` hasn't touched either file), `MERGEABLE`.

Good call switching the wording to "Partially addresses #436" — the `ctx.Err()` guards at the top of each closure are still unexercised, which is the right thing to leave open. The comment you added explaining why `gentype` drops the ctx before a reactor can see it is genuinely useful; it documents the dead end so the next person doesn't re-derive it.

One non-blocking note on CI: `build-and-push-image` is red, but it's infrastructure, not you — it died in *Set up job* with `Failed to resolve action download info. Error: Service Unavailable`, before any of your code ran. A re-run should clear it. A few FOSSA checks and the cross-platform build were still pending when I looked; I'm not gating on those.

LGTM. Approving.

PR #461, opened 2026-08-05
refactor: simplify postResults error handling

aaa-aashna:
## Summary

This PR simplifies the error-handling flow around postResults to make the synchronous and asynchronous execution paths easier to understand.

Simplified the postResults error-handling flow.
Improved readability of the synchronous and asynchronous code paths.
Preserved existing functionality and error propagation.
No functional behavior changes intended.

This is a refactoring-only change intended to improve maintainability and make the error-handling contract easier to follow for future contributors.

 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling when posting results, with more consistent reporting for both immediate and background operations.
  * Improved cancellation behavior so interrupted requests terminate promptly.
  * Ensured communication, response-processing, and result-posting failures are surfaced more reliably.


matthyx (commented):
Reviewed the refactor. The core change is correct and I verified it is genuinely behaviour-preserving, but there are two merge blockers and the refactor stops half-way. Details below.

## Blockers

**1. DCO fails — commit `d0ea902` has no `Signed-off-by` trailer.**
Every commit merged into `main` carries one (`#474`, `#468`, `#464`, `#451`, …), and the DCO check is red on this PR. Please amend and force-push:

```bash
git commit --amend --signoff
git push --force-with-lease
```

**2. No CI evidence — the Go workflows never ran.**
`pull_request_created` and `PR Image Build` are both sitting in `action_required` (runs `31000112444` / `31000110459`, 0s), so nothing has actually built or tested this. That needs a maintainer to approve the workflow run.

For what it's worth I ran it locally against the PR head, and it is clean:

- `go build ./...` — OK
- `go vet ./adapters/... ./core/...` — OK
- `gofmt -l .` — clean
- `go test ./adapters/v1/... ./core/...` — `ok adapters/v1 66.8s`, `ok core/services 3.2s`
- `go test -race ./adapters/v1/...` — clean

So this is a process blocker, not a code one.

## Comments now contradict the code

This is a readability-only PR, so the stale comments matter more than usual — both of them now describe the *old* flow:

- `adapters/v1/backend_utils.go:96` — "failure is reported to the caller via errorChan, not the return value, for chunks sent from a goroutine". That is now backwards: the failure comes from the **return value**, and this wrapper is what forwards it to `errorChan`.
- `adapters/v1/backend_utils.go:113` — "Goroutine-wrapped callers (postResultsAsGoroutine, for chunk reports) additionally get the same failure via errorChan". No longer true; that caller passes `nil` precisely so `postResults` does *not* touch the channel.

Please rewrite both to describe the new contract, something like: *errors travel out by return value; `postResultsAsGoroutine` forwards them to `errorChan` because nothing is synchronously waiting on its return.*

## The refactor is only half-done

After this change, `postResults`' `errorChan` parameter has exactly one non-`nil` caller left — `sendSummaryAndVulnerabilities` (`backend_utils.go:75`) — and **that send is dead**. On summary failure `postResults` pushes the error into `errChan` *and* returns it; `sendSummaryAndVulnerabilities` wraps and returns it; `backend.go:485-495` then returns `summaryErr` early and never ranges over `errChan`. Nothing ever reads that value.

So the net result is a parameter that is `nil` at one call site and dead at the other. A reader hitting `postResults(..., nil)` now has to go figure out why — which is arguably *more* confusing than what we had. I'd finish the job: drop `errorChan` from `postResults` entirely, delete the four internal `sendError` calls, and let the single goroutine wrapper own the forwarding. Then the contract is uniform — "`postResults` returns errors, the goroutine wrapper channels them" — and the PR actually delivers the simplification in its title. It is a small follow-up on top of what you already have.

## Verification of behaviour-preservation

Confirmed, since the PR claims no functional change. `sendError` returns immediately when `errorChan == nil`, so passing `nil` erases the four internal sends, and the wrapper then delivers exactly one error — same value, same channel, same ctx. Delivery count and ordering are unchanged.

One incidental improvement worth calling out: `sendError` can block when `errChan` (cap 10) is saturated, until ctx is cancelled. Previously it blocked *inside* `postResults` with the HTTP response body still open (`defer resp.Body.Close()` had not run yet); now it blocks after `postResults` has returned and the body is closed, so a saturated channel no longer pins an HTTP connection. Small, but a real win.

## Minor

- The formatting churn in `adapters/mockplatform.go`, `adapters/v1/syft_test.go`, `core/ports/providers.go` and `core/services/scan_test.go` is legitimate — those four files are genuinely not `gofmt`-clean on `main` and this PR makes them clean. No objection, but it is unrelated to the stated purpose; a separate commit would keep the diff honest.
- No test covers the changed path. There is no test asserting that a chunk-goroutine failure reaches `errChan`. Given the whole claim here is "behaviour is identical", a small test pinning that contract would be worth adding — it is exactly the kind of thing a future refactor could silently break.
- The branch is 6 commits behind `main`, but `backend_utils.go` has not been touched there since your base, and GitHub reports the PR as mergeable. No rebase strictly needed.

Happy to approve once the sign-off lands, CI is green, and the comments are corrected. The `errorChan` parameter removal is the one I would really like to see, but I will not hold the PR on it if you would rather do it separately.

aaa-aashna:
@matthyx  i have removed errorChan from postResults and made it consistently return errors to the caller and worked upon other suggestions also , PTAL

matthyx (commented):
Re-reviewed at `f97d983`. Both blockers from my last pass are cleared and the main ask is done properly — thanks. One hunk needs to come out before I approve, and it is not the one you touched on purpose.

## Cleared

- **DCO** now passes; all three commits carry a `Signed-off-by`.
- **CI ran and is green**: `pr-created / test / Basic-Test`, `Create cross-platform build`, `build-and-push-image`, CodeQL, FOSSA, GitGuardian.
- **`errorChan` removed from `postResults`** — this is exactly what I was after. The function now returns errors and nothing else, the four internal `sendError` calls are gone, `sendSummaryAndVulnerabilities` just propagates the return value, and `postResultsAsGoroutine` is the single place that forwards to the channel. The contract is finally uniform, and the comment at `backend_utils.go:96-98` describes it accurately.
- **The unrelated gofmt churn is gone** from the diff. Good call.

I re-verified locally at `f97d983`: `go build ./...` OK, `go vet ./adapters/... ./core/...` OK, `gofmt -l` clean, `go test ./adapters/v1/... ./core/...` pass (`adapters/v1` 183s, `core/services` 13s), `go test -race ./adapters/v1/...` clean.

## Please revert: the `incrementCounter` change

`backend_utils.go:187-201` is unrelated to this PR's stated purpose, and all three added lines are dead:

```go
func incrementCounter(counter *int64, isGlobal, isIgnored bool) {
	if counter == nil { return }          // no caller passes nil
	if isGlobal && isIgnored { return }   // now unreachable as a distinct case
	if isIgnored { return }               // new — subsumes the check above
	*counter++
}
```

All five call sites pass `isGlobal` as a literal `true` (`backend_utils.go:261, 266, 269, 279, 282`), and no caller passes a nil counter. So today this changes nothing — I confirmed the tests behave identically. But it leaves the function in a worse state than it found it:

- The new `if isIgnored` return makes `if isGlobal && isIgnored` unreachable as a distinct branch, so **`isGlobal` is now entirely ignored** while remaining in the signature.
- That is a trap for later: the parameter exists precisely to distinguish "global/summary counter, skip ignored vulns" from "severity-level stat, count them anyway". Anyone who later calls this with `isGlobal=false, isIgnored=true` will silently get a no-op instead of the increment the signature promises.
- The `counter == nil` guard is defensive code for a case that cannot occur, and it would mask a nil-pointer bug rather than surface it.

Please drop that hunk entirely and leave `incrementCounter` as it is on `main`. If you think `isGlobal` genuinely is redundant, that is a fine cleanup — but as its own PR, where it can be argued and reviewed on its merits rather than riding along in an error-handling refactor.

## Two small things while you are in there

- **`postResults` lost its doc comment.** It had one before; now it has none. For the central function of a readability-focused PR that is a step backwards. Please add a short one, e.g. *"postResults posts a single report and returns the failure, if any. Callers that run it in a goroutine are responsible for forwarding the error to errorChan."*
- **`backend_utils.go:147`** — the stray `// return the error to the caller` sits above one of five identical `return err` statements. It restates the code and only appears on one branch. Drop it.

Once the `incrementCounter` hunk is reverted I will approve — everything else here is good.

matthyx (approved):
Re-reviewed at `a048a1a`. The `incrementCounter` hunk is gone and the two small things are addressed, so this is good to merge. Approving.

## What changed since `f97d983`

- **`incrementCounter` is back to `main`'s logic.** The `counter == nil` guard and the `if isIgnored { return }` return are both removed; the function is once again just `if isGlobal && isIgnored { return }; *counter++`. `isGlobal` means something again, and the trap for a future `isGlobal=false, isIgnored=true` caller is gone. Thanks for pulling that out — it belongs in its own PR if you still want to make the case.
- **`postResults` has a doc comment again** (`backend_utils.go:113-114`), and it states the contract accurately: returns the failure, goroutine callers forward it.
- **The stray `// return the error to the caller`** at the old line 147 is gone.

## Re-verified at `a048a1a`

CI is fully green on this head: DCO, `pr-created / test / Basic-Test`, `Create cross-platform build`, `build-and-push-image`, CodeQL, FOSSA (dependency/license/security), GitGuardian, CodeRabbit. The one CodeRabbit inline thread — which was about the `incrementCounter` change — is resolved and moot now.

Locally:

- `go build ./...` — OK
- `go vet ./adapters/... ./core/...` — OK
- `gofmt -l .` — clean
- `go test ./adapters/v1/... ./core/...` — `ok adapters/v1 58.1s`, `ok core/services 3.0s`
- `go test -race ./adapters/v1/...` — clean, 121s

I also checked the two things this refactor could plausibly have left dangling, and neither did:

- **`sendError` is not dead.** Two live callers remain: the forward in `postResultsAsGoroutine` (`backend_utils.go:101`) and the chunk-count mismatch in `sendVulnerabilities` (`:184`).
- **`sendSummaryAndVulnerabilities`' `errChan` param is not dead.** It is still passed through to `postResultsAsGoroutine` at `:86` for the first-chunk dispatch.

So the final shape is the one I was after: `postResults` returns errors and touches no channel, and `postResultsAsGoroutine` is the single place that forwards to `errorChan`.

## Non-blocking nits

None of these hold the PR — take them or leave them:

- `backend_utils.go:193` — a blank line got added inside `incrementCounter` before `*counter++`. It is the only remaining trace of that hunk in a function this PR otherwise leaves alone; dropping it would make the diff touch exactly the code it means to.
- `backend_utils.go:149` — the removed `sendError` left a blank line before `return err` in the rate-limit branch. The other three error branches return immediately, so this one now reads slightly differently for no reason.
- `backend_test.go:998` — a blank line was removed above the `ReportError` assertion. Unrelated churn.
- `postResults`' signature is now split one-param-per-line. Nothing wrong with it, but every other function in this file keeps its signature on one line, including much longer ones (`sendSummaryAndVulnerabilities`, `sendVulnerabilitiesRoutine`). Worth matching the file.
- Still no test pinning that a chunk-goroutine failure reaches `errChan`. Pre-existing gap, not something this PR introduced — I raised it on the first pass and I am not going to hold the PR on it. It would be a good small follow-up, since "behaviour is identical" is exactly the kind of claim that a later refactor can silently break.

Nice job working through three rounds of this. Approved.

PR #464, opened 2026-08-05
fix: propagate SecurityException aliases into vulnerability exception…

### LINKED ISSUE #463: bug: SecurityException aliases are ignored when converting to exception policies, so alias-configured exceptions silently fail to suppress findings
## Description

The `SecurityException` / `ClusterSecurityException` CRDs carry an `aliases` list (`pkg/securityexception/v1beta1/types.go:91-94`) so a single exception can target both the NVD ID and the GitHub Advisory ID (GHSA) of the same flaw. This is documented as a first-class feature (`docs/security-exception-design.md:122`).

However, `buildPolicy` (`adapters/v1/securityexception.go:69-92`) copies only `vuln.Vulnerability.ID` into `VulnerabilityPolicies` and drops the `Aliases` slice entirely:

```go
VulnerabilityPolicies: []armotypes.VulnerabilityPolicy{
	{Name: strings.TrimSpace(vuln.Vulnerability.ID)},
},
```

Matching runs through `getCVEExceptionMatchCVENameFromList` (`adapters/v1/backend_utils.go:275-293`), which compares only `VulnerabilityPolicies[j].Name` against the Grype-reported ID. Because aliases never reach `VulnerabilityPolicies`, an exception declared with `id: CVE-2021-44228` and `aliases: [GHSA-jfh8-c2jp-hdph]` can never suppress a finding Grype reports as `GHSA-jfh8-c2jp-hdph`.

Both consumers of the converted policies are affected:
1. **In-cluster filtering** (`ApplySecurityExceptions`, `securityexception.go:116`)
2. **Backend report generation** (`DomainToArmo`, `domain_to_armo.go:173`)

---

## Environment

- **OS:** N/A — static code bug
- **Version:** `main` (current)

---

## Steps To Reproduce

1. Create a `SecurityException` with `vulnerabilities[].vulnerability: {id: "CVE-2021-44228", aliases: ["GHSA-jfh8-c2jp-hdph"]}`.
2. Scan an image where Grype reports the finding under the primary ID `GHSA-jfh8-c2jp-hdph`.
3. Observe that the match is not moved to `IgnoredMatches` in the stored `VulnerabilityManifest`, and the backend report records `exceptionApplied=0`.

---

## Expected Behavior

An exception suppresses the finding whether Grype reports the primary ID or any of its listed aliases.

---

## Actual Behavior

The `aliases` field is silently dropped during policy conversion. Only the exact `id` value ever matches, forcing operators to duplicate exception CRDs for every possible alias representation.

---

## Additional Context

- **Root Cause:** A missing slice expansion in `buildPolicy`. `armotypes.VulnerabilityPolicy` has no dedicated aliases field (verified in `armoapi-go v0.0.718`), so appending aliases as additional policy names is the correct mechanism. The downstream matching engine already iterates over all policy names with `strings.EqualFold`.
- **Design Note:** The CRD schema's intent is that operators explicitly declare aliases. Reverse-matching Grype's reported aliases would require threading match aliases into `getCVEExceptionMatchCVENameFromList` (a signature change with a wider blast radius) and is unnecessary to honor the declared field.
- **Scope Boundary:** This fix honors explicitly declared aliases. Auto-resolving undeclared cross-ecosystem IDs (e.g., a CVE-only exception against a Grype-reported GHSA) is a separate concern and out of scope.

---

## Proposed Solution

In `adapters/v1/securityexception.go` (`buildPolicy`), expand the policy list to include non-empty trimmed aliases:

```go
vulnerabilityPolicies := []armotypes.VulnerabilityPolicy{
	{Name: strings.TrimSpace(vuln.Vulnerability.ID)},
}
for _, alias := range vuln.Vulnerability.Aliases {
	if a := strings.TrimSpace(alias); a != "" {
		vulnerabilityPolicies = append(vulnerabilityPolicies, armotypes.VulnerabilityPolicy{Name: a})
	}
}
```

---

## Tests to Add

Add unit tests in `adapters/v1/securityexception_test.go`:
1. Verify that policy conversion produces one `armotypes.VulnerabilityPolicy` per alias (handling whitespace padding and empty elements safely).
2. End-to-end regression test: Verify that `ApplySecurityExceptions` moves a `GHSA-…`-reported match to `IgnoredMatches` when the CRD declares it as an alias of a CVE.

AdeshDeshmukh:
Fixes #463

## Overview

`SecurityException` / `ClusterSecurityException` CRDs declare an `aliases` list (`pkg/securityexception/v1beta1/types.go:91-94`, documented in `docs/security-exception-design.md:122`) so one exception can target both the NVD ID and the GitHub Advisory ID of the same flaw. 

`buildPolicy` (`adapters/v1/securityexception.go:69-92`) copied only `vuln.Vulnerability.ID` into `VulnerabilityPolicies`, dropping the aliases entirely. Since `getCVEExceptionMatchCVENameFromList` (`adapters/v1/backend_utils.go:275-293`) matches solely on `VulnerabilityPolicies[].Name`, an exception configured as `id: CVE-2021-44228` + `aliases: [GHSA-jfh8-c2jp-hdph]` silently failed to suppress a finding Grype reports as `GHSA-jfh8-c2jp-hdph` — in both the in-cluster filter (`ApplySecurityExceptions`, `securityexception.go:116`) and the backend report (`DomainToArmo`, `domain_to_armo.go:173`).

---

## Change

`adapters/v1/securityexception.go` — `buildPolicy`
- Build `VulnerabilityPolicies` from the primary ID plus each trimmed, non-empty alias.
- `armotypes.VulnerabilityPolicy` has no dedicated `aliases` field (verified in `armoapi-go v0.0.718`), so expanding the name list is the only mechanism; matching is already `EqualFold`-based (post-#353).
- No changes to `backend_utils.go`; no new imports; no existing test declares aliases, so no existing assertions change.

---

## Tests

Added in `adapters/v1/securityexception_test.go`:
- `TestConvertVulnerabilityExceptions_Aliases` — ID + padded/empty aliases → one policy name per alias, empties skipped.
- `TestApplySecurityExceptions_MatchesByAlias` — end-to-end `ConvertToVulnerabilityExceptionPolicies` → `ApplySecurityExceptions`: a Grype match reported as `GHSA-…` is moved to `IgnoredMatches` when the CRD lists it as an alias of a CVE.

Both fail on `main` (alias dropped) and pass with this change. The `DomainToArmo` backend path is covered transitively: it consumes the same converted policy slice and the same matching function exercised by the conversion test.

---

## Verification

Red (unfixed main, aliases dropped):

[FAIL] TestConvertVulnerabilityExceptions_Aliases
  expected: []string{"CVE-2021-44228", "GHSA-jfh8-c2jp-5v3q", "GHSA-xxxx-xxxx-xxxx"}
  actual  : []string{"CVE-2021-44228"}
[FAIL] TestApplySecurityExceptions_MatchesByAlias

Green (with fix):

go test: 2 passed
- go build ./... — clean
- go vet ./adapters/v1/ — clean (pre-existing findings in backend_test.go, present on main)
- gofmt -l adapters/v1/securityexception.go adapters/v1/securityexception_test.go — clean
- go test ./adapters/v1/... — full package passes (with Docker running; the only otherwise-failing test, Test_grypeAdapter_DBVersion, is a Docker/testcontainers test that also fails on clean main)

---

## Checklist

- [x] DCO signed off

## Summary by CodeRabbit

* **New Features**
  * Security exceptions now recognize vulnerability aliases in addition to the main vulnerability ID.
  * Findings reported under an alias can now be matched and ignored by the corresponding exception.

* **Bug Fixes**
  * Improved matching so trimmed, non-empty alias values are handled consistently.
  * Unrelated findings continue to remain unaffected.

### DIFF
--- a/adapters/v1/securityexception.go
+++ b/adapters/v1/securityexception.go
@@ -87,13 +87,20 @@ func shouldSuppress(vuln sev1beta1.VulnerabilityException) bool {
 }
 
 func buildPolicy(spec sev1beta1.SecurityExceptionSpec, vuln sev1beta1.VulnerabilityException, namespace string) armotypes.VulnerabilityExceptionPolicy {
+	vulnerabilityPolicies := []armotypes.VulnerabilityPolicy{
+		{Name: strings.TrimSpace(vuln.Vulnerability.ID)},
+	}
+	for _, alias := range vuln.Vulnerability.Aliases {
+		if a := strings.TrimSpace(alias); a != "" {
+			vulnerabilityPolicies = append(vulnerabilityPolicies, armotypes.VulnerabilityPolicy{Name: a})
+		}
+	}
+
 	p := armotypes.VulnerabilityExceptionPolicy{
-		PolicyType: "vulnerabilityExceptionPolicy",
-		Actions:    []armotypes.VulnerabilityExceptionPolicyActions{armotypes.Ignore},
-		VulnerabilityPolicies: []armotypes.VulnerabilityPolicy{
-			{Name: strings.TrimSpace(vuln.Vulnerability.ID)},
-		},
-		Reason: spec.Reason,
+		PolicyType:            "vulnerabilityExceptionPolicy",
+		Actions:               []armotypes.VulnerabilityExceptionPolicyActions{armotypes.Ignore},
+		VulnerabilityPolicies: vulnerabilityPolicies,
+		Reason:                spec.Reason,
 	}
 
 	if spec.ExpiresAt != nil {
--- a/adapters/v1/securityexception_test.go
+++ b/adapters/v1/securityexception_test.go
@@ -335,7 +335,6 @@ func TestApplySecurityExceptions_NoExceptions(t *testing.T) {
 func TestConvertMixedStatusException(t *testing.T) {
 	const underInvestigationID = "CVE-2023-11111"
 	const notAffectedID = "CVE-2023-22222"
-
 	exceptions := []sev1beta1.SecurityException{
 		{
 			ObjectMeta: metav1.ObjectMeta{Namespace: "default"},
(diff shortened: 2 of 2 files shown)


AdeshDeshmukh:
@matthyx PTAL

matthyx:
needs rebase

AdeshDeshmukh:
@matthyx  Rebase DOne!

adapters/v1/securityexception.go
```diff
@@ -87,13 +87,20 @@ func shouldSuppress(vuln sev1beta1.VulnerabilityException) bool {
...
 
 func buildPolicy(spec sev1beta1.SecurityExceptionSpec, vuln sev1beta1.VulnerabilityException, namespace string) armotypes.VulnerabilityExceptionPolicy {
+	vulnerabilityPolicies := []armotypes.VulnerabilityPolicy{
+		{Name: strings.TrimSpace(vuln.Vulnerability.ID)},
+	}
+	for _, alias := range vuln.Vulnerability.Aliases {
+		if a := strings.TrimSpace(alias); a != "" {
+			vulnerabilityPolicies = append(vulnerabilityPolicies, armotypes.VulnerabilityPolicy{Name: a})
+		}
+	}
```
matthyx, line 97: Non-blocking nit: consider deduplicating the names (case-insensitively) before building the slice.

`getCVEExceptionMatchCVENameFromList` has no `break` in its inner loop over `VulnerabilityPolicies`, so it appends the *same* policy once per matching name. Until now that was harmless because every policy had exactly one name; with alias expansion, a CRD that repeats the primary ID in `aliases` (e.g. `id: CVE-2021-44228`, `aliases: [CVE-2021-44228, GHSA-jfh8-c2jp-5v3q]`) — or lists the same alias twice — will now produce a duplicated entry in `Vulnerability.ExceptionApplied` in `DomainToArmo`, i.e. duplicated exception objects in the backend payload.

Suppression itself is unaffected (`ApplySecurityExceptions` only checks `len(matched) > 0`, and `Summarize` looks at `ExceptionApplied[0]`), so this is cosmetic/payload noise rather than a correctness bug. Something like:

```go
seen := map[string]struct{}{}
var vulnerabilityPolicies []armotypes.VulnerabilityPolicy
for _, name := range append([]string{vuln.Vulnerability.ID}, vuln.Vulnerability.Aliases...) {
	n := strings.TrimSpace(name)
	if n == "" {
		continue
	}
	if _, ok := seen[strings.ToLower(n)]; ok {
		continue
	}
	seen[strings.ToLower(n)] = struct{}{}
	vulnerabilityPolicies = append(vulnerabilityPolicies, armotypes.VulnerabilityPolicy{Name: n})
}
```

would keep the payload clean. Fine to do in a follow-up.

matthyx (approved):
LGTM — approving.

The fix is minimal and matches the mechanism the matching engine already supports: `getCVEExceptionMatchCVENameFromList` iterates all `VulnerabilityPolicies[].Name` with `EqualFold`, so expanding the name list is the right (and only) way to honour `aliases` without an `armoapi-go` change. Both consumers — `ApplySecurityExceptions` (in-cluster filter) and `DomainToArmo` (backend `ExceptionApplied`) — go through that same function, so both paths are fixed by this one change.

Verified:
- Rebase is current (base = `main` head `0476f20`), all 7 checks green including `Basic-Test` and DCO.
- `shouldSuppress` still gates on a non-empty primary ID, so an alias-only entry stays a no-op — consistent with the CRD schema where `id` is required.
- Trimming is applied to aliases the same way it already was to the primary ID; empty entries are skipped.
- Tests cover both the conversion (whitespace-padded and empty aliases) and the end-to-end suppression of a GHSA-reported match, and they genuinely fail without the production change.
- Scope is tight — no reverse-matching of Grype's own `RelatedVulnerabilities`, which #463 explicitly ruled out of scope.

One non-blocking nit inline about deduplicating names; fine as a follow-up.

PR #466, opened 2026-08-05
fix: copy container profile labels before enriching the relevancy scan

### LINKED ISSUE #465: bug: GetContainerRelevancyScans panics on a ContainerProfile with nil labels and pollutes stored labels via map aliasing
## Description

`GetContainerRelevancyScans` (`adapters/v1/container_profile.go:66,78`) takes `containerProfile.Labels` by reference and writes `kubescape.io/workload-container-name` into it.

**Process crash (primary):** a `ContainerProfile` with no `metadata.labels` decodes to a `nil` map. `kubescape-storage` never initializes `Labels` (`ContainerProfileProcessor.PreSave` only guards `Annotations`), and `kubevuln` never writes container profiles (only `MemoryStore.StoreContainerProfile` exists). There is no `nil`-check, so the write at `container_profile.go:78` panics with:

panic: assignment to entry in nil map

`ScanCP` runs inside a `workerPool.Submit` goroutine (`controllers/http.go:123-124`) — the `gammazero` workerpool does not recover panics (`workerpool.go:235-241`), and `gin.Recovery()` (`cmd/http/main.go:122`) only covers the handler goroutine. Consequently, the panic is unrecovered and terminates the entire `kubevuln` process, dropping all concurrent in-flight scans. The same aliased map is written again downstream in `filterSBOM` (`scan.go:768`).

**Stored-state pollution (secondary):** with a reference-returning repository (`MemoryStore`, used for tests/local mode), the write mutates the stored profile — subsequent reads return runtime labels. (Production `APIServerStore` deserializes a fresh object per call.)

---

## Environment

- **OS:** N/A — static code bug
- **Version:** `main` (current)

---

## Steps To Reproduce

1. Store a `ContainerProfile` via `MemoryStore.StoreContainerProfile` with no `Labels` (or `Labels: {"foo":"bar"}`).
2. Call `GetContainerRelevancyScans(..., partialRelevancy=true)`.
3. Observed result:
   - Nil labels: runtime panic (process-fatal in the worker goroutine).
   - Non-nil labels: re-fetching the profile shows `kubescape.io/workload-container-name` injected into storage.

---

## Expected Behavior

The stored object is never mutated, and a labels-less profile scans cleanly — labels should be initialized on the returned scan instance only.

---

## Actual Behavior

- Nil labels: runtime panic terminating the worker process.
- Non-nil labels: stored profile is polluted with runtime metadata.

---

## Additional Context

- **Root Cause:** the `labels` map is assigned by reference (`container_profile.go:66`) and written to (`:78`) without a `nil`-guard or defensive copy. The write is required downstream — `filterSBOM` reads `labels[helpersv1.ContainerNameMetadataKey]` for the filtered-SBOM annotation (`scan.go:750`) — so the fix must preserve the entry.
- **Note:** this is the same panic class as #405 (`mergeMaps` nil-map panic).

---

## Proposed Solution

Shallow-copy the map before writing inside `GetContainerRelevancyScans`, which cleanly handles both isolation and the `nil` case:

```go
scanLabels := make(map[string]string, len(containerProfile.Labels)+1)
for k, v := range containerProfile.Labels {
	scanLabels[k] = v
}
scanLabels[helpersv1.ContainerNameMetadataKey] = instanceID.GetContainerName()
```
---
Tests to Add
Add unit tests in adapters/v1/container_profile_test.go (new — the adapter has no direct unit tests today):
1. Verify a nil-labels scan succeeds without panicking (fails on main).
2. Verify stored profile labels remain unchanged after a scan (mutated on main).

AdeshDeshmukh:
Fixes #465

## Overview

`GetContainerRelevancyScans` wrote `kubescape.io/workload-container-name` into a `labels` map taken by reference from the repository object (`adapters/v1/container_profile.go:66,78`).

- **Process crash:** A `ContainerProfile` with no `metadata.labels` decodes to a `nil` map — `kubescape-storage` never initializes `Labels` (`ContainerProfileProcessor.PreSave` only guards `Annotations`), and `kubevuln` never writes container profiles (only `MemoryStore.StoreContainerProfile` exists). The write at `:78` panicked with `assignment to entry in nil map`. `ScanCP` runs inside a `workerPool.Submit` goroutine (`controllers/http.go:123-124`); the `gammazero` workerpool does not recover panics (`workerpool.go:235-241`) and `gin.Recovery()` only covers the handler goroutine, so the panic was unrecovered and terminated the whole `kubevuln` process, dropping all in-flight scans. The same aliased map was written again downstream in `filterSBOM` (`scan.go:768`).
- **Stored-state pollution:** With a reference-returning repository (`MemoryStore`, tests/local), the write mutated the stored profile — subsequent reads returned runtime labels.

---

## Change

`adapters/v1/container_profile.go`: Build a shallow copy of `Labels` (`len+1` capacity) and set the container-name key on the copy, so repository-owned state is never modified and `nil` labels scan cleanly. The label remains available to `filterSBOM` for the filtered-SBOM annotation (`scan.go:750`). No new imports; no behavior change for the normal (non-nil) path.

---

## Tests

New `adapters/v1/container_profile_test.go` (the adapter had no direct unit tests):

- `TestGetContainerRelevancyScans_NilLabels` — a labels-less profile scans without panicking, and the returned scan carries the container-name label.
- `TestGetContainerRelevancyScans_DoesNotMutateStoredProfile` — stored profile labels stay `{"foo":"bar"}` after a scan.

Both fail on `main` and pass with this change.

---

## Verification

Red (unfixed `main`):

```text
[FAIL] TestGetContainerRelevancyScans_NilLabels
  panic: assignment to entry in nil map

[FAIL] TestGetContainerRelevancyScans_DoesNotMutateStoredProfile
  expected: map[string]string{"foo":"bar"}
  actual  : map[string]string{"foo":"bar", "kubescape.io/workload-container-name":"kube-proxy"}
```

Green (with fix):

```text
go test: 2 passed
```

- `go build ./...` — clean
- `go vet ./adapters/v1/` — clean (no findings in touched files)
- `gofmt -l adapters/v1/container_profile.go adapters/v1/container_profile_test.go` — clean
- `go test ./adapters/v1/...` — 60 passed, 1 failed: `Test_grypeAdapter_DBVersion` (Docker/testcontainers-dependent, passes when the Docker daemon is up; fails on clean `main` identically otherwise)
- `go test ./repositories/...` — 33 passed
- `go test ./core/services/ -run TestScanService_ScanCP` — 13 passed (no regression in the indirect integration path)

---

## Checklist

- [x] DCO signed off

### DIFF
--- a/adapters/v1/container_profile.go
+++ b/adapters/v1/container_profile.go
@@ -56,6 +56,12 @@ func (a *ContainerProfileAdapter) GetContainerRelevancyScans(ctx context.Context
 	if err != nil {
 		return nil, fmt.Errorf("failed to generate instance ID: %w", err)
 	}
+	// copy labels map so we never mutate the repository-owned profile and nil labels scan cleanly
+	scanLabels := make(map[string]string, len(containerProfile.Labels)+1)
+	for k, v := range containerProfile.Labels {
+		scanLabels[k] = v
+	}
+	scanLabels[helpersv1.ContainerNameMetadataKey] = instanceID.GetContainerName()
 	scan := ports.ContainerRelevancyScan{
 		Completion:       completionStatus,
 		ContainerName:    instanceID.GetContainerName(),
@@ -63,7 +69,7 @@ func (a *ContainerProfileAdapter) GetContainerRelevancyScans(ctx context.Context
 		ImageTag:         containerProfile.Spec.ImageTag,
 		InstanceID:       instanceID,
 		InstanceIDString: instanceIDString,
-		Labels:           containerProfile.Labels,
+		Labels:           scanLabels,
 		RelevantFiles:    mapset.NewSet[string](),
 		Wlid:             wlid,
 	}
@@ -74,8 +80,6 @@ func (a *ContainerProfileAdapter) GetContainerRelevancyScans(ctx context.Context
 	for _, f := range containerProfile.Spec.Opens {
 		scan.RelevantFiles.Add(f.Path)
 	}
-	// add container name to labels
-	scan.Labels[helpersv1.ContainerNameMetadataKey] = instanceID.GetContainerName()
 	scans = append(scans, scan)
 	return scans, nil
 }
--- a/adapters/v1/container_profile_test.go
+++ b/adapters/v1/container_profile_test.go
@@ -1,0 +1,61 @@
+package v1
+
+import (
+	"context"
+	"testing"
+
+	helpersv1 "github.com/kubescape/k8s-interface/instanceidhandler/v1/helpers"
+	"github.com/kubescape/kubevuln/repositories"
+	"github.com/kubescape/storage/pkg/apis/softwarecomposition/v1beta1"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func validContainerProfile(name, namespace string, labels map[string]string) v1beta1.ContainerProfile {
+	return v1beta1.ContainerProfile{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+			Annotations: map[string]string{
+				helpersv1.CompletionMetadataKey: helpersv1.Full,
+				helpersv1.StatusMetadataKey:     helpersv1.Learning,
+				helpersv1.InstanceIDMetadataKey: "apiVersion-apps/v1/namespace-kube-system/kind-DaemonSet/name-kube-proxy/containerName-kube-proxy",
+				helpersv1.WlidMetadataKey:       "wlid/cluster-test/namespace-kube-system/kind-DaemonSet/name-kube-proxy",
+			},
+			Labels: labels,
+		},
+		Spec: v1beta1.ContainerProfileSpec{
+			Execs:    []v1beta1.ExecCalls{{Path: "/usr/local/bin/kube-proxy"}},
+			Opens:    []v1beta1.OpenCalls{{Path: "/etc/kubernetes/kube-proxy.conf"}},
+			ImageID:  "sha256:c1b135231b5b1a6799346cd701da4b59e5b7ef8e694ec7b04fb23b8dbe144137",
+			ImageTag: "k8s.gcr.io/kube-proxy:v1.24.3",
+		},
+	}
+}
+
+func TestGetContainerRelevancyScans_NilLabels(t *testing.T) {
+	repo := repositories.NewMemoryStorage(false, false)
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Reviewed and verified locally. **No blockers — good to merge.**

### What I checked

**The bug is real and the diagnosis in the description holds.** On `main`, `GetContainerRelevancyScans` assigned `Labels: containerProfile.Labels` and then wrote `scan.Labels[ContainerNameMetadataKey] = ...` into that aliased map. Two independent failure modes, both confirmed:

- **nil-map panic.** Nothing on the read path guarantees `ContainerProfile.Labels` is non-nil — a profile CR without `metadata.labels` decodes to a nil map, and `APIServerStore.GetContainerProfile` returns it verbatim (`repositories/apiserver.go:218`).
- **Unrecovered panic → process death.** Confirmed at `controllers/http.go:124`: `ScanCP` runs inside `workerPool.Submit(func(){...})` with no `recover()` in the closure, and `gin.Recovery()` does not extend to that goroutine. So the panic takes down the whole process, not just the one scan — that is what makes this worth fixing rather than logging.

Notably every other label-enrichment site in the repo already guards this (`StoreCVE` at `repositories/apiserver.go:299`, `enrichSummaryManifestObjectLabels` at `:469`); `container_profile.go` was the one outlier.

**The fix is correct and minimal.** Building the copy with `len+1` capacity before constructing the scan means repository-owned state is never touched and the nil case degrades to an empty map. No import added, no behavior change on the non-nil path — the container-name label still reaches `filterSBOM` (`core/services/scan.go:352` → `:797`).

### Verification

I merged current `main` into the branch (clean merge — the branch is ~50 files behind, so worth confirming) and ran against that merge result:

```text
go build ./...                                          clean
gofmt -l <both touched files>                           clean
go vet ./adapters/v1/                                   clean
go test ./adapters/... -count=1                         ok (incl. the Docker-backed grype tests)
go test ./core/services/... ./repositories/... -count=1  ok
```

Both new tests are genuinely red on `main` — I reverted just `container_profile.go` to the `main` version and re-ran:

```text
--- FAIL: TestGetContainerRelevancyScans_NilLabels
panic: assignment to entry in nil map
  .../adapters/v1/container_profile.go:78
```

That is the exact production crash site, so the regression test pins the real defect rather than a proxy for it.

### Non-blocking notes

1. `maps.Clone` would be the idiomatic one-liner here, but it returns `nil` for a `nil` input and would need its own guard — the explicit loop is the better call as written. No change requested.
2. The returned `scan.Labels` map is still aliased downstream: `filterSBOM` writes `ArtifactTypeMetadataKey` into it (`scan.go:816`) and `enrichSummaryManifestObjectLabels` mutates its argument in place. That enrich-in-place pattern is pre-existing and is now confined to per-scan state, which is exactly the point of this fix — just flagging it as the next place the same class of aliasing could bite if a caller ever reuses a labels map across scans.

Approving.

PR #468, opened 2026-08-05
fix: bound graceful shutdown drain time instead of blocking indefinitely

### LINKED ISSUE #467: Graceful shutdown drains an unbounded, un-deadlined job queue — SIGKILL loses in-flight and queued scans
# Description

`cmd/http/main.go`'s graceful shutdown is built around `srv.Shutdown(ctx)` with a 5-second deadline — but that deadline bounds nothing real, because every scan handler (`GenerateSBOM`, `ScanCP`, `ScanCVE`, `ScanRegistry` in `controllers/http.go`) writes its `200 OK` response and returns *before* doing any actual work. The real scan runs asynchronously on a `workerpool.WorkerPool`, detached from the request via `context.WithoutCancel`. `srv.Shutdown` has no in-flight HTTP handlers left to wait for by the time it's called, so its 5s budget is spent waiting on nothing.

The actual drain happens afterward, in `HTTPController.Shutdown()`, via `workerPool.StopWait()` — which has no deadline, no cap, and no way to bound how long it blocks: it waits for the *entire* waiting queue to finish, not just the task currently running. Combined with `scanConcurrency` defaulting to `1` and `scanTimeout` defaulting to `5*time.Minute`, N queued jobs at shutdown time means up to `N × 5min` of blocking drain, with kubernetes' `terminationGracePeriodSeconds` (default 30s) having no relationship to any of this whatsoever.

It gets worse: `context.WithoutCancel` strips deadline propagation, not just cancellation. `ScanRegistry` (`core/services/scan.go:597-654`) unconditionally calls `platform.SendStatus`/`platform.ReportError` up to three times per scan (`Started`, then `Success`/`ReportError` on failure, then `Done`). Per the still-open #450, those two methods route through `kubescape/backend`'s `BaseReportSender`, which is hardcoded to `context.Background()` and retries up to 3 times with a blocking, uncancellable `time.Sleep(5s)` — meaning each call can block up to ~15s+ regardless of what `ctx` was passed in, because the ctx never reaches that layer at all. A single queued `ScanRegistry` job can therefore hold up `StopWait()` for 45+ seconds from `SendStatus`/`ReportError` alone, on top of whatever `CreateSBOM`/`ScanSBOM`/`SubmitCVE` take — comfortably exceeding a 30s termination grace period on its own.

Net effect: every rolling deployment, HPA scale-down, or node drain that hits kubevuln while jobs are queued or running risks the kubelet SIGKILLing the process mid-drain, silently losing scan work the client was already told succeeded (`200 OK` was written before the work even started), and potentially abandoning partial writes to CRD/backend storage mid-flight.

# Environment

Not deployment-specific — reproducible by code inspection and reasoning about `net/http.Server.Shutdown` + `gammazero/workerpool@v1.1.3` semantics against current `main`. Applies to any kubevuln deployment using the HTTP controller (`cmd/http`), i.e. any non-sidecar-only deployment, regardless of OS or Kubescape version. Worse under default config (`scanConcurrency: 1`, `scanTimeout: 5m`) but not fixed by simply raising concurrency, since `StopWait()` still has no deadline.

# Steps To Reproduce

1. Read `controllers/http.go:38-75` (`GenerateSBOM`) — note the handler writes `http.StatusOK` at line 63, *then* calls `h.workerPool.Submit(func() { err = h.scanService.GenerateSBOM(bgCtx) ... })` at lines 65-66, with `bgCtx := context.WithoutCancel(ctx)`. The same fire-then-detach pattern repeats in `ScanCP` (93-139), `ScanCVE` (142-180), and `ScanRegistry` (222-259).
2. Read `cmd/http/main.go:161-170`: `srv.Shutdown(ctx)` is given a 5-second `context.WithTimeout`, then immediately followed by `controller.Shutdown()` — a *separate*, later call with no deadline of its own.
3. Read `controllers/http.go:277-281` (`HTTPController.Shutdown`): calls `h.workerPool.StopWait()`.
4. In the vendored `gammazero/workerpool@v1.1.3/workerpool.go:74-79`, confirm `StopWait()`'s doc comment: "waits for all queued tasks tasks to complete" — the entire backlog, not just the running task — with no parameter to bound how long that takes.
5. Confirm `Submit()` (same file, lines 88-108) is documented as non-blocking regardless of queue size ("Submit will not block regardless of the number of tasks submitted"), so there is no backpressure capping how large that backlog can get before shutdown.
6. Confirm the defaults in `config/config.go:79-80`: `scanConcurrency` = `1`, `scanTimeout` = `5*time.Minute`.
7. Read `core/services/scan.go:597-654` (`ScanRegistry`): three unconditional `platform.SendStatus`/`platform.ReportError` calls per scan.
8. Cross-reference #450: `BackendAdapter.SendStatus`/`ReportError` route through `backendClientV1.NewBaseReportSender`, which is hardcoded to `context.Background()` and retries 3× with an uncancellable `time.Sleep(5s)` between attempts — so `ctx` cancellation (or its absence, post-`WithoutCancel`) makes no difference to how long these specific calls can block.

# Expected behavior

`controller.Shutdown()` should complete within a bounded, configurable window aligned with the pod's `terminationGracePeriodSeconds`, draining in-flight work best-effort and abandoning (with a clear log line, not a silent SIGKILL) anything that can't finish in time — the same way `srv.Shutdown`'s 5s budget is *supposed* to work, but actually applied to the work that matters (the queued/running scans), not to an HTTP handler that already returned.

# Actual Behavior

`srv.Shutdown`'s 5s deadline bounds an HTTP server that has nothing left in flight by the time it's called (every handler already returned after writing `200 OK`). The real work — up to `scanConcurrency` running jobs plus an unbounded waiting queue — is drained afterward by `workerPool.StopWait()`, which has no deadline, no cap, and can be extended arbitrarily by calls (`SendStatus`/`ReportError`) that ignore `ctx` entirely and retry for ~15s+ each. There is no code path that bounds total shutdown time, and no log output during the drain to indicate how much longer it will take (only a single "remaining jobs" count logged once at the start).

# Additional context

Affected code:
- `cmd/http/main.go` — shutdown sequencing (140-173)
- `controllers/http.go` — `HTTPController.Shutdown` (277-281), fire-and-forget `Submit` calls in `GenerateSBOM`/`ScanCP`/`ScanCVE`/`ScanRegistry`
- `core/services/scan.go` — `ScanRegistry`'s unconditional `SendStatus`/`ReportError` calls (597-654), compounding via #450
- `config/config.go` — `scanConcurrency`/`scanTimeout` defaults (79-80)

Acceptance criteria for a fix:
- `HTTPController.Shutdown()` (or its caller in `main.go`) bounds the total drain time to a configurable duration tied to `terminationGracePeriodSeconds`, instead of delegating to `StopWait()`'s unbounded wait.
- When the drain deadline is hit with jobs still queued or running, this is logged clearly (job count, which jobs if feasible) rather than the process being silently SIGKILLed with no trace of what was lost.
- Either the worker pool gets bounded queue depth (real backpressure on `Submit`, e.g. reject or block past a configured limit) so an unbounded backlog can't accumulate in the first place, or the shutdown path explicitly caps how many queued-but-not-started jobs it will wait for.
- `SendStatus`/`ReportError` calls that happen during a bounded shutdown drain don't independently block past the overall drain deadline — at minimum, this issue should cross-link #450 so whichever of the two gets fixed first accounts for the other.

Suggested approach:
1. Wrap `HTTPController.Shutdown()` with a `context.WithTimeout` (new config field, e.g. `shutdownTimeout`, defaulting to something safely under typical `terminationGracePeriodSeconds`) and run `workerPool.StopWait()` in a goroutine, `select`-ing between its completion and the timeout — logging explicitly which outcome occurred and the remaining/abandoned queue size in the timeout case.
2. Consider whether `Submit` should instead be `SubmitWait`-style with a bounded queue (the `workerpool` library doesn't support bounded queues directly, so this may mean tracking queue depth with a semaphore/channel ahead of `Submit`, or swapping the "always 200 OK immediately" response for a `503`/backpressure response once a depth threshold is hit — worth a maintainer decision on desired client-facing behavior under load).
3. Once #450 lands (whichever direction it takes), confirm the same ctx/deadline actually reaches `SendStatus`/`ReportError` from a bounded shutdown context, not just from a live scan's request context.
4. Add a test exercising `HTTPController.Shutdown()` against a slow/blocking task in the pool and asserting it returns within the configured deadline rather than blocking until the task finishes.

bhuvan-somisetty:
## Overview

`HTTPController.Shutdown` delegated straight to `workerpool.WorkerPool.StopWait`, which waits for the *entire* queued backlog to finish with no deadline of its own. Every scan handler (`GenerateSBOM`/`ScanCP`/`ScanCVE`/`ScanRegistry`) already writes its `200 OK` and detaches the real work onto the worker pool via `context.WithoutCancel` before returning, so by the time `srv.Shutdown`'s 5s budget in `main.go` runs, there's no in-flight HTTP handler left for it to bound — the real, unbounded wait happens afterward in `Shutdown`, invisible to that deadline.

With `scanConcurrency` defaulting to `1` and `scanTimeout` to 5 minutes, a handful of queued jobs at shutdown time could block the drain for many minutes — comfortably past Kubernetes' typical 30s `terminationGracePeriodSeconds`, with nothing logged beyond a single queue-size count at the very start. The kubelet ends up SIGKILLing the process mid-drain on every rolling deploy or scale-down that catches kubevuln with work queued, silently losing scans the caller was already told succeeded.

Fix, scoped to the shutdown/drain path only:
- New config field `shutdownTimeout` (`config/config.go`), defaulting to 25s — safely under a typical 30s termination grace period.
- `HTTPController.Shutdown` (`controllers/http.go`) now takes that timeout, runs `workerPool.StopWait()` in a goroutine, and selects between it completing and the timeout elapsing — logging explicitly which happened, including the remaining queue size in the timeout case.
- `cmd/http/main.go` passes `c.ShutdownTimeout` through.

Not touched (intentionally, per the issue's own scoping): backpressure on `Submit`/bounded queue depth, and the separate `SendStatus`/`ReportError` ctx-ignoring gap tracked in #450 — both are flagged in #467 as needing a maintainer decision on desired client-facing behavior, not something to guess at silently in this PR.

## How to Test

- Added `TestHTTPController_Shutdown_BoundedByTimeout` (`controllers/http_test.go`): submits a task that blocks forever on a channel, calls `Shutdown(100ms)`, and asserts it returns close to that timeout instead of hanging — regression guard for the exact bug this PR fixes.
- Added `TestLoadConfigShutdownTimeoutDefault` (`config/config_test.go`) asserting the 25s default.
- Updated the three existing `.Shutdown()` call sites (two in `controllers/http_test.go`, one in `cmd/http/main_test.go`) to the new signature.
- `go build ./...`, `go vet ./...`: clean.
- `go test ./...`: clean, except the pre-existing `Test_grypeAdapter_DBVersion`/`Test_grypeAdapter_ScanSBOM`/`TestScanService_NginxTest`, which fail in this environment with `panic: rootless Docker is not supported on Windows` — unrelated `testcontainers`/Docker-dependent tests, not touched by this change.

## Related issues/PRs

Resolved #467

## Summary by CodeRabbit

* **New Features**
  * Added a configurable shutdown timeout for active scans, defaulting to 20 seconds.
  * HTTP shutdown now provides an additional five-second termination window.

* **Bug Fixes**
  * Prevented shutdown from hanging indefinitely when active work remains blocked.
  * Queued scans are abandoned when shutdown begins, while in-flight scans are allowed to finish within the configured timeout.
  * Shutdown errors are logged without terminating the process prematurely.

* **Documentation**
  * Documented shutdown timing, scan-draining behavior, and configuration options.

### DIFF
--- a/cmd/http/main.go
+++ b/cmd/http/main.go
@@ -166,8 +166,9 @@ func main() {
 		logger.L().Ctx(ctx).Fatal("server forced to shutdown", helpers.Error(err))
 	}
 
-	// Purging the controller worker queue
-	controller.Shutdown()
+	// Purging the controller worker queue, bounded by ShutdownTimeout so this
+	// doesn't silently run past the pod's terminationGracePeriodSeconds
+	controller.Shutdown(c.ShutdownTimeout)
 
 	logger.L().Info("kubevuln exiting")
 }
--- a/cmd/http/main.go
+++ b/cmd/http/main.go
@@ -158,12 +158,17 @@ func main() {
 	stop()
 	logger.L().Info("shutting down gracefully")
 
-	// modify context to inform the server it has 5 seconds to finish
-	// the request it is currently handling
-	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
-	defer cancel()
-	if err := srv.Shutdown(ctx); err != nil {
-		logger.L().Ctx(ctx).Fatal("server forced to shutdown", helpers.Error(err))
+	// Give the HTTP server up to 5 seconds to finish in-flight requests. This must be
+	// derived from context.Background(), not the signal ctx above: that one is already
+	// Done() by this point (we only got here via <-ctx.Done()), so deriving from it
+	// would hand srv.Shutdown an already-expired context and return immediately.
+	srvCtx, srvCancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer srvCancel()
+	if err := srv.Shutdown(srvCtx); err != nil {
+		// Log and continue rather than exit: a failed/timed-out HTTP shutdown must not
+		// skip the worker-pool drain below, since that's where actual queued/running
+		// scan work would otherwise be silently lost (#467).
+		logger.L().Ctx(srvCtx).Error("server shutdown error", helpers.Error(err))
 	}
 
 	// Purging the controller worker queue, bounded by ShutdownTimeout so this
--- a/config/config.go
+++ b/config/config.go
@@ -80,11 +80,13 @@ func LoadConfig(path string) (Config, error) {
 	v.SetDefault("scanConcurrency", 1)
 	v.SetDefault("scanTimeout", 5*time.Minute)
 	v.SetDefault("scannerReadinessTimeout", 60*time.Second)
-	// Kept safely under Kubernetes' typical terminationGracePeriodSeconds (30s
-	// default) so the worker-pool drain (see controllers.HTTPController.Shutdown)
-	// has a chance to log an explicit abandonment before the kubelet SIGKILLs the
-	// process, instead of the drain silently running out the clock unbounded.
-	v.SetDefault("shutdownTimeout", 25*time.Second)
+	// cmd/http/main.go spends up to 5s on the HTTP server's own Shutdown before this
+	// phase (the worker-pool drain, see controllers.HTTPController.Shutdown) even
+	// starts. 20s here keeps the combined worst case at 25s, safely under Kubernetes'
+	// typical terminationGracePeriodSeconds (30s default), so the drain has a chance
+	// to log an explicit abandonment before the kubelet SIGKILLs the process, instead
+	// of silently running out the clock past the grace period.
+	v.SetDefault("shutdownTimeout", 20*time.Second)
 	v.SetDefault("vexGeneration", false)
 	v.SetDefault("namespace", "kubescape")
 	v.SetDefault("scanEmbeddedSBOMs", false)
--- a/config/config.go
+++ b/config/config.go
@@ -56,6 +56,7 @@ type Config struct {
 	ScanEmbeddedSboms       bool              `mapstructure:"scanEmbeddedSBOMs"`
 	ScanTimeout             time.Duration     `mapstructure:"scanTimeout"`
 	ScannerReadinessTimeout time.Duration     `mapstructure:"scannerReadinessTimeout"`
+	ShutdownTimeout         time.Duration     `mapstructure:"shutdownTimeout"`
 	RiskAcceptance          bool              `mapstructure:"riskAcceptance"`
 	Storage                 bool              `mapstructure:"storage"`
 	StoreFilteredSbom       bool              `mapstructure:"storeFilteredSbom"`
@@ -79,6 +80,11 @@ func LoadConfig(path string) (Config, error) {
 	v.SetDefault("scanConcurrency", 1)
 	v.SetDefault("scanTimeout", 5*time.Minute)
 	v.SetDefault("scannerReadinessTimeout", 60*time.Second)
+	// Kept safely under Kubernetes' typical terminationGracePeriodSeconds (30s
+	// default) so the worker-pool drain (see controllers.HTTPController.Shutdown)
+	// has a chance to log an explicit abandonment before the kubelet SIGKILLs the
+	// process, instead of the drain silently running out the clock unbounded.
+	v.SetDefault("shutdownTimeout", 25*time.Second)
 	v.SetDefault("vexGeneration", false)
 	v.SetDefault("namespace", "kubescape")
 	v.SetDefault("scanEmbeddedSBOMs", false)
--- a/controllers/http.go
+++ b/controllers/http.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"net/http"
 	"strconv"
+	"time"
 
 	wssc "github.com/armosec/armoapi-go/apis"
 	"github.com/gammazero/workerpool"
@@ -274,8 +275,31 @@ func registryScanCommandToScanCommand(c wssc.RegistryScanCommand) domain.ScanCom
 	return command
 }
 
-func (h HTTPController) Shutdown() {
+// Shutdown drains the worker pool's queued and running scans, but bounds how long it
+// waits. workerPool.StopWait() alone has no deadline of its own — it waits for the
+// entire queued backlog to finish, not just the task currently running — which for
+// scanConcurrency=1 with a full queue can run far longer than Kubernetes'
+// terminationGracePeriodSeconds. Without a bound here, that difference is invisible: the
+// process just gets SIGKILLed mid-drain with no log trace of what was abandoned. timeout
+// should be kept safely under the pod's terminationGracePeriodSeconds so this path has a
+// chance to log before that happens.
+func (h HTTPController) Shutdown(timeout time.Duration) {
 	logger.L().Info("purging SBOM creation queue",
-		helpers.String("remaining jobs", strconv.Itoa(h.workerPool.WaitingQueueSize())))
-	h.workerPool.StopWait()
+		helpers.String("remaining jobs", strconv.Itoa(h.workerPool.WaitingQueueSize())),
+		helpers.String("timeout", timeout.String()))
+
+	drained := make(chan struct{})
+	go func() {
+		h.workerPool.StopWait()
+		close(drained)
+	}()
+
+	select {
+	case <-drained:
+		logger.L().Info("SBOM creation queue drained")
+	case <-time.After(timeout):
+		logger.L().Warning("shutdown timeout reached, abandoning queued/running scans",
+			helpers.String("remaining jobs", strconv.Itoa(h.workerPool.WaitingQueueSize())),
+			helpers.String("timeout", timeout.String()))
+	}
 }
--- a/controllers/http.go
+++ b/controllers/http.go
@@ -275,14 +275,14 @@ func registryScanCommandToScanCommand(c wssc.RegistryScanCommand) domain.ScanCom
 	return command
 }
 
-// Shutdown drains the worker pool's queued and running scans, but bounds how long it
-// waits. workerPool.StopWait() alone has no deadline of its own — it waits for the
-// entire queued backlog to finish, not just the task currently running — which for
-// scanConcurrency=1 with a full queue can run far longer than Kubernetes'
-// terminationGracePeriodSeconds. Without a bound here, that difference is invisible: the
-// process just gets SIGKILLed mid-drain with no log trace of what was abandoned. timeout
-// should be kept safely under the pod's terminationGracePeriodSeconds so this path has a
-// chance to log before that happens.
+// Shutdown abandons queued-but-not-started scans immediately and waits only for
+// currently running scans to finish, bounded by timeout. workerPool.Stop() (as opposed
+// to StopWait()) already caps the wait at "currently running tasks", matching this
+// package's acceptance criteria of abandoning pending work rather than racing through
+// the whole backlog; timeout is the backstop for a running task that ignores ctx
+// cancellation (see #450) and would otherwise block Stop() indefinitely. timeout should
+// be kept safely under the pod's terminationGracePeriodSeconds so this path has a
+// chance to log before the kubelet SIGKILLs the process.
 func (h HTTPController) Shutdown(timeout time.Duration) {
 	logger.L().Info("purging SBOM creation queue",
 		helpers.String("remaining jobs", strconv.Itoa(h.workerPool.WaitingQueueSize())),
@@ -290,7 +290,7 @@ func (h HTTPController) Shutdown(timeout time.Duration) {
 
 	drained := make(chan struct{})
 	go func() {
-		h.workerPool.StopWait()
+		h.workerPool.Stop() // abandon the queue, wait only for in-flight scans
 		close(drained)
 	}()
 
@@ -298,8 +298,7 @@ func (h HTTPController) Shutdown(timeout time.Duration) {
 	case <-drained:
 		logger.L().Info("SBOM creation queue drained")
 	case <-time.After(timeout):
-		logger.L().Warning("shutdown timeout reached, abandoning queued/running scans",
-			helpers.String("remaining jobs", strconv.Itoa(h.workerPool.WaitingQueueSize())),
+		logger.L().Warning("shutdown timeout reached, abandoning in-flight scans",
 			helpers.String("timeout", timeout.String()))
 	}
 }
--- a/cmd/http/main_test.go
+++ b/cmd/http/main_test.go
@@ -5,6 +5,7 @@ import (
 	"net/http/httptest"
 	"os"
 	"testing"
+	"time"
 
 	"github.com/armosec/armoapi-go/apis"
 	"github.com/gin-gonic/gin"
@@ -121,7 +122,7 @@ func TestScan(t *testing.T) {
 			assert.Equal(t, test.expectedCode, w.Code, w.Code)
 			assert.Equal(t, test.expectedBody, w.Body.String(), w.Body.String())
 
-			controller.Shutdown()
+			controller.Shutdown(5 * time.Second)
 		})
 	}
 }
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -22,7 +22,7 @@ func TestLoadConfigShutdownTimeoutDefault(t *testing.T) {
 	viper.Reset()
 	c, err := LoadConfig("testdata")
 	assert.NoError(t, err)
-	assert.Equal(t, 25*time.Second, c.ShutdownTimeout)
+	assert.Equal(t, 20*time.Second, c.ShutdownTimeout)
 }
 
 func TestLoadConfigNotFound(t *testing.T) {
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+	"time"
 
 	"github.com/spf13/viper"
 	"github.com/stretchr/testify/assert"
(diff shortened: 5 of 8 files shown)


cmd/http/main.go
```diff
@@ -166,8 +166,9 @@ func main() {
 		logger.L().Ctx(ctx).Fatal("server forced to shutdown", helpers.Error(err))
 	}
 
-	// Purging the controller worker queue
-	controller.Shutdown()
+	// Purging the controller worker queue, bounded by ShutdownTimeout so this
+	// doesn't silently run past the pod's terminationGracePeriodSeconds
+	controller.Shutdown(c.ShutdownTimeout)
```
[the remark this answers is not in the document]
bhuvan-somisetty: Confirmed, real bug — fixed in bddddc5. The signal ctx was already Done() by the time we built the 5s WithTimeout from it, so srv.Shutdown got an already-expired context and its error almost always hit logger.L().Fatal → os.Exit(1), which meant controller.Shutdown(c.ShutdownTimeout) could get skipped entirely on every real shutdown, silently defeating this whole PR. Now the server-shutdown context is built from context.Background(), and a failed/timed-out server shutdown just gets logged as an error instead of exiting, so the worker-pool drain always runs.

config/config.go
```diff
@@ -79,6 +80,11 @@ func LoadConfig(path string) (Config, error) {
 	v.SetDefault("scanConcurrency", 1)
 	v.SetDefault("scanTimeout", 5*time.Minute)
 	v.SetDefault("scannerReadinessTimeout", 60*time.Second)
+	// Kept safely under Kubernetes' typical terminationGracePeriodSeconds (30s
+	// default) so the worker-pool drain (see controllers.HTTPController.Shutdown)
+	// has a chance to log an explicit abandonment before the kubelet SIGKILLs the
+	// process, instead of the drain silently running out the clock unbounded.
+	v.SetDefault("shutdownTimeout", 25*time.Second)
```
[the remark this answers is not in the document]
bhuvan-somisetty: Fair catch — fixed in bddddc5. Dropped the controller-drain default from 25s to 20s, so the combined worst case (5s HTTP server shutdown + 20s worker-pool drain) is 25s total, leaving a real 5s margin under the typical 30s terminationGracePeriodSeconds instead of exactly meeting it. Updated TestLoadConfigShutdownTimeoutDefault and the comment to match.

bhuvan-somisetty:
Thanks for catching both of these the shutdown ctx one especially, since it meant the whole point of this PR (bounding the worker-pool drain) could get skipped entirely on a real SIGTERM. Fixed both in bddddc5: server shutdown now builds its own context.Background()-derived timeout instead of reusing the already-canceled signal ctx, logs instead of exiting on a failed server shutdown, and the default drain budget dropped to 20s so there's an actual margin under the 30s grace period instead of exactly hitting it. All checks green.

bhuvan-somisetty:
@matthyx PTAL, whenever u get chance.

controllers/http.go
```diff
@@ -274,8 +275,31 @@ func registryScanCommandToScanCommand(c wssc.RegistryScanCommand) domain.ScanCom
...
+		h.workerPool.StopWait()
+		close(drained)
+	}()
+
+	select {
+	case <-drained:
+		logger.L().Info("SBOM creation queue drained")
+	case <-time.After(timeout):
+		logger.L().Warning("shutdown timeout reached, abandoning queued/running scans",
+			helpers.String("remaining jobs", strconv.Itoa(h.workerPool.WaitingQueueSize())),
```
matthyx: **Blocker (small): this count under-reports exactly the case the log exists to explain.**

`WaitingQueueSize()` returns only tasks sitting in the *waiting queue* — not the ones currently executing. In `gammazero/workerpool@v1.1.3`, `Submit` hands a task straight to a worker while `workerCount < maxWorkers` and only pushes to `waitingQueue` past that, so with the default `scanConcurrency: 1` and a single stuck scan, this line logs `remaining jobs: 0` while it is in fact abandoning a running scan.

Your own new test demonstrates it: `workerpool.New(1)` + one blocked `Submit` → `WaitingQueueSize() == 0`.

That's the one scenario an operator will actually be reading this log line for ("the pod died during rollout — what did we lose?"), and it currently says nothing was lost. Suggest relabeling to make the scope explicit and surfacing the running count too, e.g.:

```go
logger.L().Warning("shutdown timeout reached, abandoning queued/running scans",
    helpers.String("abandoned queued jobs", strconv.Itoa(h.workerPool.WaitingQueueSize())),
    helpers.Int("scanConcurrency", h.workerPool.Size()),
    helpers.String("timeout", timeout.String()))
```

Same relabel is worth doing on the "purging SBOM creation queue" line above (L288).

controllers/http.go
```diff
@@ -274,8 +275,31 @@ func registryScanCommandToScanCommand(c wssc.RegistryScanCommand) domain.ScanCom
...
+func (h HTTPController) Shutdown(timeout time.Duration) {
 	logger.L().Info("purging SBOM creation queue",
-		helpers.String("remaining jobs", strconv.Itoa(h.workerPool.WaitingQueueSize())))
-	h.workerPool.StopWait()
+		helpers.String("remaining jobs", strconv.Itoa(h.workerPool.WaitingQueueSize())),
+		helpers.String("timeout", timeout.String()))
+
+	drained := make(chan struct{})
+	go func() {
+		h.workerPool.StopWait()
```
matthyx: **Design question — should this be `Stop()` rather than `StopWait()`?**

`StopWait()` runs the *entire* backlog before returning; the timeout here doesn't stop that, it just stops us waiting for it. Concretely, during the 20s drain the pool keeps **starting brand-new queued scans**, and every scan started in that window is guaranteed to be killed mid-flight when `main` returns a moment later (`scanTimeout` is 5m — nothing started at T+15s finishes by T+20s).

So the current shape can *increase* the number of scans abandoned mid-write, which is the "partial writes to CRD/backend storage mid-flight" risk #467 calls out.

`workerpool.Stop()` has exactly the semantics the issue's acceptance criteria ask for — "waits for only currently running tasks to complete, pending tasks that are not currently running are abandoned" — which caps queued-but-not-started jobs at zero instead of racing through them. The `select`/timeout stays as the backstop for a stuck *running* task (the #450 `SendStatus`/`ReportError` case), which `Stop()` alone still doesn't bound.

```go
go func() {
    h.workerPool.Stop() // abandon the queue, wait only for in-flight scans
    close(drained)
}()
```

Not insisting — if you'd rather keep best-effort draining of the backlog, that's a legitimate call. But it's worth being deliberate about it in the doc comment, because the current comment reads as if the timeout bounds the backlog when it really only bounds *our wait for* the backlog. Happy either way as long as it's a conscious choice.

config/config.go
```diff
@@ -79,6 +80,13 @@ func LoadConfig(path string) (Config, error) {
 	v.SetDefault("scanConcurrency", 1)
 	v.SetDefault("scanTimeout", 5*time.Minute)
 	v.SetDefault("scannerReadinessTimeout", 60*time.Second)
+	// cmd/http/main.go spends up to 5s on the HTTP server's own Shutdown before this
+	// phase (the worker-pool drain, see controllers.HTTPController.Shutdown) even
+	// starts. 20s here keeps the combined worst case at 25s, safely under Kubernetes'
+	// typical terminationGracePeriodSeconds (30s default), so the drain has a chance
+	// to log an explicit abandonment before the kubelet SIGKILLs the process, instead
+	// of silently running out the clock past the grace period.
+	v.SetDefault("shutdownTimeout", 20*time.Second)
```
matthyx, line 89: **Blocker: new config option isn't documented.**

`docs/CONFIGURATION.md` is the reference for every option in this struct — `scannerReadinessTimeout`, the most recently added field, is in both the options table (L174) and the JSON schema block (L260). `shutdownTimeout` needs the same treatment:

1. A row in the options table (probably its own "Shutdown Options" subsection rather than under Scanning).
2. An entry in the JSON schema block alongside `scannerReadinessTimeout`.

Two things worth stating explicitly in that doc text, since neither is discoverable from the code:

- **The real budget is `5s + shutdownTimeout`, not `shutdownTimeout`.** The 5s in `cmd/http/main.go` is hardcoded, so anyone raising this to fit a longer `terminationGracePeriodSeconds` needs to know they're adding to a fixed 5s, not replacing it.
- **What `0` (or a negative value) does.** `time.After(0)` fires immediately, so `shutdownTimeout: 0` turns `Shutdown` into a near no-op that races the drain rather than meaning "wait forever" — the same footgun `scannerReadinessTimeout`'s doc entry already calls out explicitly, and worth mirroring here. (Alternatively, clamp a non-positive value to the default in `LoadConfig` and skip the doc caveat.)

Also `docs/ARCHITECTURE.md:76` — "Graceful shutdown | Waits for in-flight scans to complete" — is now stale; it's bounded and abandons work at the deadline.

matthyx (changes requested):
Nice, focused fix — the core change is right and the follow-up commit (`bddddc5`) fixed the two things that mattered most: deriving `srv.Shutdown`'s context from an already-canceled parent, and the `Fatal` that could skip the worker-pool drain entirely. I verified both against `net/http` and `gammazero/workerpool@v1.1.3` semantics, and built/vetted/ran the touched packages with `-race -count=3` locally — all green.

Three things before merge, one of which is a genuine blocker:

1. **`docs/CONFIGURATION.md` is missing `shutdownTimeout`** (blocker). Every other option in the `Config` struct is documented there, including `scannerReadinessTimeout` in both the table and the JSON schema. `docs/ARCHITECTURE.md:76` is also now stale. Details inline.
2. **The abandonment log under-reports** (blocker, one-line fix). `WaitingQueueSize()` excludes running tasks, so at the default `scanConcurrency: 1` this logs `remaining jobs: 0` while abandoning a running scan — the exact case the log line exists for. Details inline.
3. **`StopWait()` vs `Stop()`** (your call, not blocking). `StopWait()` keeps *starting* new queued scans during the drain window, each guaranteed to be killed mid-flight. `Stop()` matches #467's "cap queued-but-not-started jobs" criterion more directly. Details inline.

Two smaller notes, no action needed:

- The commit body for `bddddc5` says the old code "almost always returned a non-nil error". Not quite — `http.Server.Shutdown` returns `nil` immediately when `closeIdleConns()` succeeds on the first pass, which is the *common* case here precisely because every handler writes `200 OK` and returns before doing work. The bug is real but it's a "whenever a connection happens to be active at SIGTERM" bug, not an every-shutdown one. Worth softening since the commit message is the permanent record.
- `time.After(timeout)` leaks its timer until it fires. Irrelevant on a process about to exit; flagging only so nobody "fixes" it later thinking it's a real leak.

Scoping out backpressure on `Submit` and the #450 `SendStatus`/`ReportError` gap was the right call — agreed those need a maintainer decision on client-facing behavior, not a guess inside this PR.

bhuvan-somisetty:
@matthyx Thanks for the thorough pass! Fixed all three:

1. Switched to `workerPool.Stop()` instead of `StopWait()` abandons queued-but-not-started scans immediately and only waits on in-flight ones, matching #467's acceptance criteria and avoiding the "starts new scans then kills them mid-flight" issue you flagged.
2. Documented `shutdownTimeout` in CONFIGURATION.md (table + JSON schema), including the "it's on top of the hardcoded 5s server shutdown" and "0/negative doesn't mean wait forever" caveats you called out.
3. Updated the now-stale graceful shutdown line in ARCHITECTURE.md.

Ran vet/build/tests locally, all green. Ready for another look.

bhuvan-somisetty:
Good catch, fixed the schema pattern was unsigned-only so it would've rejected \`shutdownTimeout: -1s\` even though the docs say negative values are valid (immediate fire). Pattern now allows an optional leading \`-\`.

matthyx (approved):
All three points addressed — approving.

Re-verified at `1bb054e`: `go build ./...` and `go vet` clean, `go test -race -count=3 ./controllers/... ./config/...` and `go test -race ./cmd/http/...` all pass, full CI green.

- **`Stop()` over `StopWait()`** — good, this is the right semantics for a shutdown path. Queued-but-not-started scans are dropped immediately instead of being started into a window they can't finish in, so the drain no longer manufactures new mid-write abandonments. The doc comment now says what the timeout actually bounds (a stuck *running* task, per #450) rather than implying it bounds the backlog. `TestHTTPController_Shutdown_BoundedByTimeout` still exercises the real path: one task on a `New(1)` pool is *running*, not queued, so `Stop()` genuinely blocks on it and the 100ms backstop is what returns.
- **Misleading count** — removing `WaitingQueueSize()` from the timeout branch is the honest fix; that branch is now purely about in-flight scans, and the count would have been meaningless there.
- **Docs** — the `Shutdown Options` table, the JSON schema entry, and the `ARCHITECTURE.md` row all landed. Calling out `5s + shutdownTimeout` as the real budget against `terminationGracePeriodSeconds`, and the `0`-or-less footgun, is exactly what an operator sizing a pod needs.

Two nits, neither worth another round — take or leave:

- The `purging SBOM creation queue` line still labels `WaitingQueueSize()` as `remaining jobs`. Under `Stop()` that number is now precisely *the queued scans about to be dropped on the floor*, which is probably the single most actionable value in the whole shutdown path — `abandoned queued jobs` would sell it better than `remaining`.
- `1bb054e` widened the schema pattern to `^-?[0-9]+(s|m|h)$`, which now differs from `scannerReadinessTimeout`'s `^[0-9]+(s|m|h)$` even though the two document identical `0`-or-less semantics. Since the schema is descriptive docs rather than an enforced validator, I'd have left it non-negative and let the prose carry the caveat — advertising `-5s` as valid input invites someone to try it. Harmless either way.

Optionally, a small test asserting a *queued* task never runs after `Shutdown` would pin the `Stop()`-not-`StopWait()` choice against a future well-meaning revert. Not a merge condition.

Thanks for the quick turnarounds on both rounds.

bhuvan-somisetty:
Thanks @matthyx!

Issue #449, opened 2026-08-04
bug: ExpiredOnFix exceptions are applied in the stored manifest but expired in the backend report for CPE-matched CVEs


AdeshDeshmukh:
# Description

A `SecurityException` with `expiredOnFix: true` is intended to stop suppressing a CVE once a fix is available. Two code paths compute "is this fixed?" with different predicates, so for CPE-based matches (fix state `unknown`/`not-fixed` but the CPE version constraint lists a fixed version), the same exception is applied in the in-cluster `VulnerabilityManifest` but expired in the report sent to the backend. The two Kubescape surfaces then disagree about whether the finding is ignored.

---

# Environment

- **OS:** N/A (code-level bug, not environment-specific)
- **Version:** Current `main` (verified at `85c8d7a`); the divergence was introduced when `ApplySecurityExceptions` landed (`1b05cb6a80`, 2026-04) alongside the existing predicate in `DomainToArmo` (`0545618a12`, 2023-05) and has never been reconciled.

---

# Steps To Reproduce

1. Load `adapters/v1/testdata/alpine-cve.format.json` (repo-shipped real Grype output) as a `models.Document`.
2. Run `grypeToDomain` (the exact path `ScanSBOM` → `grypeToDomain` uses in production) and take the `CVE-2024-13176` match:
   `Fix.State="unknown"`, `Fix.Versions=[]`, CPE `matchDetail.versionConstraint` contains `"<"` (e.g., `">= 1.0.2, < 1.0.2zl, >= 3.0.0, < 3.0.16, ..."`).
3. Create an `expiredOnFix: true` exception for `CVE-2024-13176`.
4. Feed the same match through both `ApplySecurityExceptions` and `DomainToArmo`.

---

# Expected Behavior

One shared predicate decides whether an `ExpiredOnFix` exception has expired, used by both `ApplySecurityExceptions` (in-cluster storage) and `DomainToArmo` (backend report), so a CVE is ignored or not consistently across both surfaces.

---

# Actual Behavior

The same exception produces opposite outcomes in the two paths:

```text
ApplySecurityExceptions: inMatches=false inIgnoredMatches=true     <- CVE hidden in-cluster
DomainToArmo:            ExceptionApplied(len)=0                    <- CVE reported active
DIVERGENCE: storage ignored=true, report exceptionApplied=0
```

[...]

PR #470, opened 2026-08-05
fix(core): reconcile cached manifests with the current SecurityException set

### LINKED ISSUE #469: bug: deleting a SecurityException never restores suppression on cached scans; backend telemetry and VEX diverge by cache state
# Description

Kubevuln caches vulnerability scan results in the cluster as the **exception-filtered**
`VulnerabilityManifest` (suppressed matches moved from `Matches` to `IgnoredMatches`). On a
cache hit, `ScanCVE`/`ScanCP` load that filtered manifest and re-apply the current
`SecurityException` set. But `ApplySecurityExceptions` (`adapters/v1/securityexception.go:108-129`)
only ever moves matches from `Matches` → `IgnoredMatches` — it never restores them — and the
cache-hit store guard (`len(filteredCve.Content.IgnoredMatches) > len(cve.Content.IgnoredMatches)`
in `core/services/scan.go`) can only detect a *growing* ignore set.

As a result:

1. **Deleting an exception never un-suppresses the finding — permanently.** The cluster keeps
   showing the vulnerability as ignored, and a rescan of the same image is itself a cache hit
   that does not restore it. Only deleting/expiring the cached `VulnerabilityManifest`
   un-suppresses the finding, so a workload stays silently vulnerable after an operator revokes
   an exception.
2. **Backend telemetry diverges by cache state.** A cache miss submits the **unfiltered**
   manifest to ARMO (`SubmitCVE`); a cache hit submits the **filtered** manifest. The same
   image reports different vulnerability profiles depending on arbitrary cache state.
3. **VEX diverges too.** `createVEX` builds statements from `cve.Content.Matches`
   (`repositories/apiserver.go:799`), so on cache hits VEX silently omits exception-suppressed
   vulnerabilities that a cache miss would include.

# Environment

- **OS:** N/A — static correctness bug
- **Version:** `main` — `core/services/scan.go` (`ScanCVE`, `ScanCP`)

# Steps To Reproduce

1. Scan an image so a `VulnerabilityManifest` is stored with a `SecurityException`-suppressed
   match in `IgnoredMatches` (cache-miss path, `core/services/scan.go:480-504`).
2. Delete the `SecurityException` (or `ClusterSecurityException`) that suppressed `CVE-X`.
3. Trigger another scan of the same image (cache hit, `core/services/scan.go:505-520`).
4. Inspect the stored `VulnerabilityManifest`: `CVE-X` is still in `IgnoredMatches`.
5. Compare the manifest passed to `SubmitCVE` on a cache hit vs. a cache miss — they differ.

# Expected Behavior

A cache hit behaves identically to a cache miss: the cached manifest is reconstructed to its
unfiltered form, re-evaluated against the **current** exception set (so deleted exceptions
restore their findings), and the backend always receives unfiltered data.

# Actual Behavior

Deleted exceptions are never reflected on cache hits — the finding stays suppressed — and the
backend (and VEX) receive filtered data on cache hits vs. unfiltered data on cache misses.

# Additional Context

- The stored object remains the filtered view by design (that is what cluster UIs consume); only
  the read path is corrected, so no storage-schema change is required.
- Kubevuln has no watcher/informer on `SecurityException` that would invalidate the cached
  manifest on exception change, so this is currently unsolved in-tree.
- Known limitation (out of scope): `ExpiredOnFix` changes cannot self-heal on cache hits because
  a cached match's fix-state is a scan-time snapshot (related to #449).

# Proposed Solution

On cache hits in `ScanCVE`/`ScanCP`, restore suppressed matches before re-applying exceptions:

- Add `RestoreSuppressedMatches` (inverse of `ApplySecurityExceptions`) and `IgnoredVulnIDs` in
  `adapters/v1/securityexception.go`. `IgnoredMatch` embeds the full `Match`, so reconstruction
  is lossless. Restoring every ignored match is safe: grype produces no ignored matches here —
  `GrypeAdapter.ScanSBOM` sets neither `VulnerabilityMatcher.IgnoreRules` nor `VexProcessor`
  (`adapters/v1/grype.go:302-306`; grype `vulnerability_matcher.go:236,214`) — and it is the
  fail-open direction (findings surface rather than stay hidden).
- Replace the one-directional `>` guard with a symmetric comparison of the exception-derived
  ignored-ID set, so both additions and removals re-store the manifest.
- Have `applyExceptionsToManifest` report whether the exception fetch succeeded, and skip the
  re-store on a failed fetch so a transient error cannot wipe exception filtering from the
  stored object.

AdeshDeshmukh:
Fixes #469

## Overview

On a cache hit, `ScanCVE`/`ScanCP` loaded the previously exception-filtered `VulnerabilityManifest` and re-applied the current `SecurityException` set. Since `ApplySecurityExceptions` (`adapters/v1/securityexception.go`) only ever moves matches `Matches` → `IgnoredMatches` — it never restores them — and the cache-hit store guard (`len(filteredCve.Content.IgnoredMatches) > len(cve.Content.IgnoredMatches)` in `core/services/scan.go`) could only detect a *growing* ignore set, deleting an exception never un-suppressed the finding. Because a rescan of the same image is itself a cache hit, the suppression was effectively permanent, and the backend received **filtered** data on cache hits vs. **unfiltered** data on cache misses (VEX had the same split: `createVEX` builds statements from `cve.Content.Matches`, `repositories/apiserver.go:799`).

This makes cache hits behave like cache misses: reconstruct the unfiltered manifest, re-evaluate the current exception set, and persist only when the exception-derived ignore set actually changed.

---

### adapters/v1/securityexception.go
- `RestoreSuppressedMatches` — the inverse of `ApplySecurityExceptions`: returns a copy of the document with every ignored match moved back into `Matches` and `IgnoredMatches` cleared; the input is not mutated, mirroring `ApplySecurityExceptions`' copy semantics.
- Restoring every ignored match is safe, not heuristic: `GrypeAdapter.ScanSBOM` builds `VulnerabilityMatcher` with neither `IgnoreRules` nor `VexProcessor` (`adapters/v1/grype.go`), and grype emits ignored matches only from those two inputs (`applyIgnoreRules` short-circuits on an empty rule set, `findVEXMatches` short-circuits on a nil processor — grype `vulnerability_matcher.go:236,214`). Every stored ignored match is therefore exception-applied. It is also the fail-open direction: if that assumption ever breaks, findings surface rather than stay hidden. Lossless because `IgnoredMatch` embeds the full `Match`.
- `IgnoredVulnIDs` — the set of ignored vulnerability IDs used by the change comparison, valid because SecurityException policies match on vulnerability name only (`getCVEExceptionMatchCVENameFromList`), so a policy suppresses every match of that CVE or none.

### core/services/scan.go
- `ScanCVE`/`ScanCP` cache-hit branches: capture the cached ignored-ID set, restore suppressed matches, re-filter against the current exception set, and re-store only when the ignored-ID set changed (`maps.Equal`) — persisting both additions **and removals**, and skipping redundant writes when nothing changed.
- `applyExceptionsToManifest` now reports whether the exception fetch succeeded; on a failed fetch the cache-hit re-store is skipped, so a transient error cannot wipe previously applied exceptions from the stored manifest. The three cache-miss/`cvep` call sites are signature-only (`filteredCve, _ :=`).
- Cache-miss, `cvep`, and `ScanRegistry` paths are unchanged.

### adapters/v1/securityexception_test.go
- `TestRestoreSuppressedMatches` — restores all ignored matches (order-preserving), leaves the input document untouched, and no-ops on nil / empty documents.
- `TestIgnoredVulnIDs` — correct ID set, empty document → empty set.
- `TestIgnoredVulnIDs_MultiPackageSameCVE` — pins the match-by-CVE invariant: multiple packages sharing a CVE collapse to one entry, so the ID set is sufficient.

### core/services/scan_test.go
- New test doubles: `recordingPlatform` (configurable exceptions / fetch error, captures `SubmitCVE` input), `countingCVERepository` (counts `StoreCVE`), `fakeCVEScanner` (deterministic cache-miss output, since the shared mock emits an empty-ID match no exception can target).
- Regression tests (verified to fail on `main`): deleted-exception restore (ScanCVE and ScanCP), added-exception suppress, unchanged-exception no-rewrite, exception-fetch-failure no-wipe, cache-miss unfiltered-to-backend parity. VEX is covered transitively — the manifest handed to `StoreVEX` is the same object passed to `SubmitCVE`.

---

## Verification

- `go build ./...` — clean
- `go vet ./adapters/v1/ ./core/services/` — clean
- `gofmt -l` on the changed files — clean
- `go test ./adapters/v1/... ./core/services/...` — pass (168 tests; only the pre-existing Docker-dependent `Test_grypeAdapter_DBVersion` / `TestScanService_NginxTest` fail on a machine without Docker, identical to `main`)
- The two deleted-exception regression tests fail on the unpatched `scan.go` and pass with this change, so they are true regression tests rather than snapshots of existing behavior.

---

## Follow-up Option

`ExpiredOnFix` changes still cannot self-heal on a cache hit: a cached match's fix-state is a scan-time snapshot, so a fix published after the manifest was stored is invisible until a cache miss. That is the same stale-snapshot root cause as #449 and is intentionally out of scope here — happy to take it on in a follow-up if desired.

---

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored previously suppressed vulnerability findings when exceptions are removed or expire.
  * Improved handling of cached vulnerability results when exception retrieval is incomplete or findings remain unchanged.
  * Prevented incomplete exception data from being treated as removed exceptions.
  * Ensured consistent exception filtering across full scans and filtered vulnerability scans.
  * Improved rate-limit and unsupported SBOM schema handling.
  * Allowed report generation to continue when exception data is temporarily incomplete.

* **Tests**
  * Added comprehensive coverage for exception restoration, caching, suppression, degraded retrieval, and scan submission behavior.

### DIFF
--- a/adapters/v1/securityexception.go
+++ b/adapters/v1/securityexception.go
@@ -155,6 +155,46 @@ func ApplySecurityExceptions(doc *v1beta1.GrypeDocument, exceptions domain.CVEEx
 	doc.Matches = remaining
 }
 
+// RestoreSuppressedMatches is the inverse of ApplySecurityExceptions: it returns a copy of doc
+// with every ignored match moved back into Matches and IgnoredMatches cleared, reconstructing the
+// original unfiltered manifest so a cached (filtered) manifest can be re-evaluated against the
+// current SecurityException set. The input document is not mutated, mirroring
+// ApplySecurityExceptions' copy semantics.
+//
+// Restoring every ignored match is safe: grype produces no ignored matches here (GrypeAdapter
+// sets neither VulnerabilityMatcher.IgnoreRules nor VexProcessor — grype vulnerability_matcher.go
+// applyIgnoreRules/findVEXMatches), so each stored ignored match is exception-applied. It is also
+// the fail-open direction for a security tool: if that assumption ever breaks, findings surface
+// rather than stay hidden. Reconstruction is lossless because IgnoredMatch embeds the full Match.
+func RestoreSuppressedMatches(doc *v1beta1.GrypeDocument) *v1beta1.GrypeDocument {
+	if doc == nil {
+		return nil
+	}
+	restored := doc.DeepCopy()
+	for _, im := range restored.IgnoredMatches {
+		restored.Matches = append(restored.Matches, im.Match)
+	}
+	restored.IgnoredMatches = nil
+	return restored
+}
+
+// IgnoredVulnIDs returns the set of vulnerability IDs in a manifest's ignored matches. Keyed by
+// ID because SecurityException policies match on vulnerability name only
+// (getCVEExceptionMatchCVENameFromList), so a policy suppresses every match of that CVE or none —
+// an ID set is sufficient to detect exception-set changes on cache hits.
+func IgnoredVulnIDs(doc *v1beta1.GrypeDocument) map[string]struct{} {
+	ids := map[string]struct{}{}
+	if doc == nil {
+		return ids
+	}
+	for _, im := range doc.IgnoredMatches {
+		if im.Match.Vulnerability.ID != "" {
+			ids[im.Match.Vulnerability.ID] = struct{}{}
+		}
+	}
+	return ids
+}
+
 func buildDesignators(resources []sev1beta1.ResourceMatch, namespace string) []identifiers.PortalDesignator {
 	if len(resources) == 0 {
 		// No specific resource match — create a namespace-only designator
--- a/adapters/v1/securityexception.go
+++ b/adapters/v1/securityexception.go
@@ -156,43 +156,67 @@ func ApplySecurityExceptions(doc *v1beta1.GrypeDocument, exceptions domain.CVEEx
 }
 
 // RestoreSuppressedMatches is the inverse of ApplySecurityExceptions: it returns a copy of doc
-// with every ignored match moved back into Matches and IgnoredMatches cleared, reconstructing the
-// original unfiltered manifest so a cached (filtered) manifest can be re-evaluated against the
-// current SecurityException set. The input document is not mutated, mirroring
-// ApplySecurityExceptions' copy semantics.
+// with exception-applied ignored matches moved back into Matches, reconstructing the original
+// unfiltered manifest so a cached (filtered) manifest can be re-evaluated against the current
+// SecurityException set. The input document is not mutated, mirroring ApplySecurityExceptions'
+// copy semantics. When there is nothing to restore, the input document is returned unchanged —
+// the only caller hands the result to applyExceptionsToManifest, which copies before mutating,
+// so the shared stored object is never modified.
 //
-// Restoring every ignored match is safe: grype produces no ignored matches here (GrypeAdapter
-// sets neither VulnerabilityMatcher.IgnoreRules nor VexProcessor — grype vulnerability_matcher.go
-// applyIgnoreRules/findVEXMatches), so each stored ignored match is exception-applied. It is also
-// the fail-open direction for a security tool: if that assumption ever breaks, findings surface
-// rather than stay hidden. Reconstruction is lossless because IgnoredMatch embeds the full Match.
+// Only ignored matches carrying the signature ApplySecurityExceptions writes (a single ignore
+// rule with only the vulnerability ID set) are restored. Any other entry (e.g. one that would be
+// produced if GrypeAdapter ever wired up IgnoreRules/VexProcessor) is preserved, so the helper is
+// self-guarding instead of relying on an assumption. Reconstruction is lossless because
+// IgnoredMatch embeds the full Match.
 func RestoreSuppressedMatches(doc *v1beta1.GrypeDocument) *v1beta1.GrypeDocument {
 	if doc == nil {
 		return nil
 	}
-	restored := doc.DeepCopy()
-	for _, im := range restored.IgnoredMatches {
-		restored.Matches = append(restored.Matches, im.Match)
+	if len(doc.IgnoredMatches) == 0 {
+		return doc
 	}
-	restored.IgnoredMatches = nil
+	restored := doc.DeepCopy()
+	var kept []v1beta1.IgnoredMatch
+	for _, im := range restored.IgnoredMatches {
+		if isExceptionSourcedIgnore(im) {
+			restored.Matches = append(restored.Matches, im.Match)
+		} else {
+			kept = append(kept, im)
+		}
+	}
+	restored.IgnoredMatches = kept
 	return restored
 }
 
-// IgnoredVulnIDs returns the set of vulnerability IDs in a manifest's ignored matches. Keyed by
-// ID because SecurityException policies match on vulnerability name only
-// (getCVEExceptionMatchCVENameFromList), so a policy suppresses every match of that CVE or none —
-// an ID set is sufficient to detect exception-set changes on cache hits.
-func IgnoredVulnIDs(doc *v1beta1.GrypeDocument) map[string]struct{} {
-	ids := map[string]struct{}{}
+// isExceptionSourcedIgnore reports whether an ignored match carries the AppliedIgnoreRules
+// signature ApplySecurityExceptions writes: exactly one rule whose only field is the
+// vulnerability ID.
+func isExceptionSourcedIgnore(im v1beta1.IgnoredMatch) bool {
+	if len(im.AppliedIgnoreRules) != 1 {
+		return false
+	}
+	r := im.AppliedIgnoreRules[0]
+	return r.FixState == "" && r.Package == nil
+}
+
+// IgnoredMatchKeys returns the set of match-identity keys for a manifest's ignored matches.
+// Keyed by vulnerability ID + artifact name + version (via \x00 separators) rather than CVE ID
+// alone, because an ExpiredOnFix policy suppresses only the matches of a CVE whose
+// Fix.State != "fixed" (getCVEExceptionMatchCVENameFromList), so two matches of the same CVE can
+// have different suppression states. The manifest content is fixed across a cache hit, so these
+// keys are stable and detect both ID- and fix-state-driven changes to the ignored set.
+func IgnoredMatchKeys(doc *v1beta1.GrypeDocument) map[string]struct{} {
+	keys := map[string]struct{}{}
 	if doc == nil {
-		return ids
+		return keys
 	}
 	for _, im := range doc.IgnoredMatches {
-		if im.Match.Vulnerability.ID != "" {
-			ids[im.Match.Vulnerability.ID] = struct{}{}
+		m := im.Match
+		if m.Vulnerability.ID != "" {
+			keys[m.Vulnerability.ID+"\x00"+m.Artifact.Name+"\x00"+m.Artifact.Version] = struct{}{}
 		}
 	}
-	return ids
+	return keys
 }
 
 func buildDesignators(resources []sev1beta1.ResourceMatch, namespace string) []identifiers.PortalDesignator {
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -5,6 +5,7 @@ import (
 	"crypto/sha256"
 	"errors"
 	"fmt"
+	"maps"
 	"net/http"
 	"net/url"
 	"os"
@@ -312,7 +313,7 @@ func (s *ScanService) ScanCP(mainCtx context.Context) error {
 			}
 
 			// apply security exceptions for storage (copy — original stays intact for SubmitCVE)
-			filteredCve := s.applyExceptionsToManifest(ctx, cve)
+			filteredCve, _ := s.applyExceptionsToManifest(ctx, cve)
 
 			// store filtered CVE
 			if s.storage {
@@ -330,9 +331,16 @@ func (s *ScanService) ScanCP(mainCtx context.Context) error {
 				}
 			}
 		} else {
-			filteredCve := s.applyExceptionsToManifest(ctx, cve)
+			// A cached manifest was filtered against the exception set at store time.
+			// Reconstruct the unfiltered manifest so the CURRENT exception set is
+			// re-evaluated (deleted exceptions un-suppress findings) and
+			// SubmitCVE/StoreVEX receive the same unfiltered data a cache miss would
+			// produce.
+			prevIgnored := v1.IgnoredVulnIDs(cve.Content)
+			cve.Content = v1.RestoreSuppressedMatches(cve.Content)
+			filteredCve, exceptionsFetched := s.applyExceptionsToManifest(ctx, cve)
 
-			if s.storage && len(filteredCve.Content.IgnoredMatches) > len(cve.Content.IgnoredMatches) {
+			if s.storage && exceptionsFetched && !maps.Equal(prevIgnored, v1.IgnoredVulnIDs(filteredCve.Content)) {
 				err = s.cveRepository.StoreCVE(ctx, filteredCve, false)
 				if err != nil {
 					logger.L().Ctx(ctx).Warning("storing CVE with exceptions", helpers.Error(err),
@@ -377,7 +385,7 @@ func (s *ScanService) ScanCP(mainCtx context.Context) error {
 			}
 
 			// apply security exceptions for storage (copy — original stays intact for SubmitCVE)
-			filteredCvep := s.applyExceptionsToManifest(ctx, cvep)
+			filteredCvep, _ := s.applyExceptionsToManifest(ctx, cvep)
 
 			// store filtered CVE'
 			if s.storage {
@@ -521,7 +529,7 @@ func (s *ScanService) ScanCVE(ctx context.Context) error {
 		}
 
 		// apply security exceptions for storage (copy — original stays intact for SubmitCVE)
-		filteredCve := s.applyExceptionsToManifest(ctx, cve)
+		filteredCve, _ := s.applyExceptionsToManifest(ctx, cve)
 
 		// store filtered CVE
 		if s.storage {
@@ -537,9 +545,15 @@ func (s *ScanService) ScanCVE(ctx context.Context) error {
 			}
 		}
 	} else {
-		filteredCve := s.applyExceptionsToManifest(ctx, cve)
+		// A cached manifest was filtered against the exception set at store time.
+		// Reconstruct the unfiltered manifest so the CURRENT exception set is
+		// re-evaluated (deleted exceptions un-suppress findings) and SubmitCVE
+		// receives the same unfiltered data a cache miss would produce.
+		prevIgnored := v1.IgnoredVulnIDs(cve.Content)
+		cve.Content = v1.RestoreSuppressedMatches(cve.Content)
+		filteredCve, exceptionsFetched := s.applyExceptionsToManifest(ctx, cve)
 
-		if s.storage && len(filteredCve.Content.IgnoredMatches) > len(cve.Content.IgnoredMatches) {
+		if s.storage && exceptionsFetched && !maps.Equal(prevIgnored, v1.IgnoredVulnIDs(filteredCve.Content)) {
 			err = s.cveRepository.StoreCVE(ctx, filteredCve, false)
 			if err != nil {
 				logger.L().Ctx(ctx).Warning("storing CVE with exceptions", helpers.Error(err),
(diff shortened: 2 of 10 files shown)


matthyx:
needs rebase

core/services/scan.go
```diff
@@ -537,9 +545,15 @@ func (s *ScanService) ScanCVE(ctx context.Context) error {
...
-		if s.storage && len(filteredCve.Content.IgnoredMatches) > len(cve.Content.IgnoredMatches) {
+		// A cached manifest was filtered against the exception set at store time.
+		// Reconstruct the unfiltered manifest so the CURRENT exception set is
+		// re-evaluated (deleted exceptions un-suppress findings) and SubmitCVE
+		// receives the same unfiltered data a cache miss would produce.
+		prevIgnored := v1.IgnoredVulnIDs(cve.Content)
+		cve.Content = v1.RestoreSuppressedMatches(cve.Content)
+		filteredCve, exceptionsFetched := s.applyExceptionsToManifest(ctx, cve)
+
+		if s.storage && exceptionsFetched && !maps.Equal(prevIgnored, v1.IgnoredVulnIDs(filteredCve.Content)) {
```
matthyx: **Blocker (1)** — `exceptionsFetched` only trips on a hard error, but the CRD exception path never returns one: `APIServerStore.GetSecurityExceptions` (`repositories/apiserver.go:113-155`) swallows a failed `List` and returns `(nil, nil, nil)`, and `GetCVEExceptions` merges CRD exceptions best-effort (`adapters/v1/backend.go:203-207`). A transient API-server blip therefore looks exactly like "the operator deleted the exceptions", and this branch now **persists** that: `StoreCVE`/`StoreCVESummary` write the un-suppressed manifest, and since `crdErr` is never non-nil the degraded set is also cached for 5 min (`exceptionsCacheTTL`).

The old `>` guard made this harmless — that is what the "self-heals on the next scan" comment in `backend.go` relied on. Please propagate a *completeness* signal (not just an error) and persist removals only when the exception set is known complete; additions can still be persisted unconditionally. Same applies to the `ScanCP` branch at line 343. Details in the review summary.

adapters/v1/securityexception.go
```diff
@@ -155,6 +155,46 @@ func ApplySecurityExceptions(doc *v1beta1.GrypeDocument, exceptions domain.CVEEx
...
+	}
+	restored.IgnoredMatches = nil
+	return restored
+}
+
+// IgnoredVulnIDs returns the set of vulnerability IDs in a manifest's ignored matches. Keyed by
+// ID because SecurityException policies match on vulnerability name only
+// (getCVEExceptionMatchCVENameFromList), so a policy suppresses every match of that CVE or none —
+// an ID set is sufficient to detect exception-set changes on cache hits.
+func IgnoredVulnIDs(doc *v1beta1.GrypeDocument) map[string]struct{} {
```
matthyx: **Blocker (2)** — the invariant in this comment does not hold. `getCVEExceptionMatchCVENameFromList` also filters on the per-match fix state (`adapters/v1/backend_utils.go:320`), so an `ExpiredOnFix` policy suppresses only the matches of that CVE whose `Fix.State != "fixed"`. With one CVE on two packages of differing fix state, the ignored *ID set* can stay identical while the ignored *match set* changes, so `maps.Equal` reports "unchanged" and the manifest is never rewritten — the stale-suppression bug this PR fixes, reached via `ExpiredOnFix` instead of deletion. I reproduced it on this branch (test + output in the review summary).

Keying by match identity (`ID + "\x00" + Artifact.Name + "\x00" + Artifact.Version`, or `PURL`) fixes it; the keys are stable across a cache hit, so no spurious rewrites. Worth a mixed-fix-state test too.

adapters/v1/securityexception.go
```diff
@@ -155,6 +155,46 @@ func ApplySecurityExceptions(doc *v1beta1.GrypeDocument, exceptions domain.CVEEx
...
+// Restoring every ignored match is safe: grype produces no ignored matches here (GrypeAdapter
+// sets neither VulnerabilityMatcher.IgnoreRules nor VexProcessor — grype vulnerability_matcher.go
+// applyIgnoreRules/findVEXMatches), so each stored ignored match is exception-applied. It is also
+// the fail-open direction for a security tool: if that assumption ever breaks, findings surface
+// rather than stay hidden. Reconstruction is lossless because IgnoredMatch embeds the full Match.
+func RestoreSuppressedMatches(doc *v1beta1.GrypeDocument) *v1beta1.GrypeDocument {
+	if doc == nil {
+		return nil
+	}
+	restored := doc.DeepCopy()
```
matthyx, line 178: Non-blocking perf: this `DeepCopy` is followed immediately by another one in `applyExceptionsToManifest`, on every cache-hit scan, and large-image manifests are exactly where we are memory constrained. Most manifests have no ignored matches at all — returning `doc` unchanged when `len(doc.IgnoredMatches) == 0` skips the copy in the common case. Safe here since the only caller passes the result straight into `applyExceptionsToManifest`, which copies before mutating; just document the aliasing.

matthyx (changes requested):
Thanks for the very thorough writeup — the diagnosis in #469 is correct, the direction (restore → re-apply current set → persist only on change) is the right one, and the regression tests are real regression tests. I built and ran everything locally on `4ef88d6`: `go build ./...` clean, `go test ./adapters/v1/... ./core/services/...` all green.

Two things need to be resolved before this can go in, both about the *change-detection / persist* guard rather than the restore idea itself. The restore direction is fine — what worries me is that this PR turns "removals are never persisted" into "removals are persisted", and the two safety nets guarding that transition have holes.

---

### 1. Blocker — a *silently degraded* exception fetch now wipes suppression from storage

`applyExceptionsToManifest` returning `exceptionsFetched` is the right instinct, but it only trips on a hard error, and in production the CRD path never produces one:

* `APIServerStore.GetSecurityExceptions` (`repositories/apiserver.go:113-155`) logs a failed `List` and returns `(nil, nil, nil)`. A transient API-server hiccup is indistinguishable from "no exceptions exist".
* `BackendAdapter.GetCVEExceptions` (`adapters/v1/backend.go:203-207`) merges the CRD set best-effort — the `crdErr != nil` branch that sets `cacheable = false` is effectively dead code because of the above.
* The selector fail-closed path (`WorkloadLabelsResolved` / `NamespaceLabelsResolved`, `backend.go:216-221`) also drops exceptions with `err == nil`.

So: API server blips while listing `SecurityException`s → `GetCVEExceptions` returns a *partial* set with `err == nil` → `exceptionsFetched == true` → on the next cache hit we restore, re-filter against the partial set, the ignored-ID set differs, and `StoreCVE` + `StoreCVESummary` **persist the un-suppressed manifest** into the cluster. And because `crdErr` is never non-nil, `cacheable` stays true, so that degraded set is cached for `exceptionsCacheTTL` (5 min) and every workload scanned in that window gets the same treatment.

Before this PR the one-directional `>` guard made that degradation harmless — the comment in `backend.go` even calls it "a best-effort degradation that self-heals on the next scan", which was true precisely because nothing was ever written back. This PR removes that property, so the degradation now writes wrong data to `VulnerabilityManifest` + `VulnerabilityManifestSummary` and re-submits it to the backend. It is the fail-open direction and it self-heals on the next successful scan, so it is not a hidden-vulnerability bug — but it is exactly the scenario `exceptionsFetched` was added for, and the guard can't see it.

Suggested shape (your call on the exact plumbing): propagate a *completeness* signal, not just an error — e.g. have `GetSecurityExceptions` return its `List` error and have `GetCVEExceptions` report "degraded" — then in `scan.go` persist **additions** freely but only persist **removals** when the exception set is known complete. That keeps the #469 fix intact (a genuinely deleted exception still un-suppresses) without letting a transient lister failure look like a deletion.

### 2. Blocker — the ignored-ID comparison misses `ExpiredOnFix` changes

`IgnoredVulnIDs`' doc comment says a policy "suppresses every match of that CVE or none". That is not quite true: `getCVEExceptionMatchCVENameFromList` also filters on the *per-match* fix state (`adapters/v1/backend_utils.go:320`), so an `ExpiredOnFix` policy suppresses only the matches of that CVE whose `Fix.State != "fixed"`. When one CVE hits two packages with different fix states, the ignored **set of IDs** can stay identical while the ignored **set of matches** changes — `maps.Equal` then reports "unchanged" and the manifest is never rewritten.

I reproduced it against this branch (test dropped into `core/services`, uses your existing helpers):

```go
func TestRepro_ExpiredOnFixNarrowing_NoRestore(t *testing.T) {
	fixedMatch := matchForTest("CVE-X")
	fixedMatch.Artifact = v1beta1.GrypePackage{Name: "pkgA", Version: "1.0.0"}
	fixedMatch.Vulnerability.Fix = v1beta1.Fix{State: "fixed", Versions: []string{"1.0.1"}}

	unfixedMatch := matchForTest("CVE-X")
	unfixedMatch.Artifact = v1beta1.GrypePackage{Name: "pkgB", Version: "2.0.0"}
	unfixedMatch.Vulnerability.Fix = v1beta1.Fix{State: "not-fixed"}

	expiredOnFix := true
	platform := &recordingPlatform{exceptions: domain.CVEExceptions{{
		PolicyType:            "vulnerabilityExceptionPolicy",
		Actions:               []armotypes.VulnerabilityExceptionPolicyActions{armotypes.Ignore},
		VulnerabilityPolicies: []armotypes.VulnerabilityPolicy{{Name: "CVE-X"}},
		ExpiredOnFix:          &expiredOnFix,
	}}}

	repo := repositories.NewMemoryStorage(false, false)
	s, sbomVer, cveVer, cveDBVer, ctx := newScanCVETestService(t, platform, repo, nil)

	// cached: both matches suppressed by the earlier, wider policy (no ExpiredOnFix)
	seedCachedCVEManifest(t, repo, "imageSlug", sbomVer, cveVer, cveDBVer, ctx, &v1beta1.GrypeDocument{
			{Match: fixedMatch, AppliedIgnoreRules: []v1beta1.IgnoreRule{{Vulnerability: "CVE-X"}}},
			{Match: unfixedMatch, AppliedIgnoreRules: []v1beta1.IgnoreRule{{Vulnerability: "CVE-X"}}},
		},
	})

	require.NoError(t, s.ScanCVE(ctx))

	stored, err := s.cveRepository.GetCVE(ctx, "imageSlug", s.sbomCreator.Version(), s.cveScanner.Version(), s.cveScanner.DBVersion(ctx))
	require.NoError(t, err)
	require.Len(t, stored.Content.Matches, 1, "pkgA (fixed) must be restored into Matches")
	assert.Equal(t, "pkgA", stored.Content.Matches[0].Artifact.Name)
}
```

```
stored Matches=0 IgnoredMatches=2
  ignored: CVE-X pkgA
  ignored: CVE-X pkgB
--- FAIL: pkgA (fixed) must be restored into Matches
```

The operator narrowed the exception (set `expiredOnFix: true`), `pkgA` should have come back, and it stays suppressed forever — the same failure mode #469 is about, just reached through `ExpiredOnFix` instead of deletion. The mirror case (widening the exception) leaves a finding visible that should be suppressed.

Fix is small: key the comparison set by match identity rather than by CVE ID — e.g. `ID + "\x00" + Artifact.Name + "\x00" + Artifact.Version` (or `PURL`). The manifest content is fixed across a cache hit, so those keys are stable and there is no risk of spurious rewrites. Worth a test for the mixed-fix-state case since the current comment asserts the invariant explicitly.

---

### Non-blocking

3. **Two full `DeepCopy`s per cache hit.** `RestoreSuppressedMatches` deep-copies, then `applyExceptionsToManifest` deep-copies again, on every cache-hit scan — and manifests for large images are exactly where we are memory constrained. The common case has no ignored matches at all; returning `doc` unchanged when `len(doc.IgnoredMatches) == 0` avoids the copy (safe here: the only caller hands the result straight to `applyExceptionsToManifest`, which copies before mutating — just document the aliasing).

4. **`ScanCP` relevancy summary totals change on cache hits.** `StoreCVESummary(ctx, cve, filteredCvep, true)` (`scan.go:401`) now receives the *restored* `cve`, and `parseSeverities` counts `cve.Content.Matches`, so severity totals on a cache hit now include exception-suppressed CVEs. That matches what a cache miss has always done, so I read it as the intended alignment rather than a regression — but it is a visible count change for relevancy users on the first rescan and it is not mentioned in the description. Please call it out in the release note.

5. **Rollout note.** First rescan of every cached image will restore suppressed matches into the stored manifests and re-submit them to the backend. I checked that `ExceptionApplied` is computed live per match in `domain_to_armo.go:173`, so the backend still sees them as excepted — good — but the volume change is worth flagging to the backend folks.

6. **CodeRabbit's VEX point is fair.** The `ScanCP` test builds the service with `vexGeneration=false`, so the "VEX is covered transitively" claim isn't actually exercised. The transitive argument is sound (`StoreVEX` gets the same `cve` object as `SubmitCVE`), but flipping the flag and asserting `CVE-A` shows up in the VEX payload is cheap insurance.

7. **Optional hardening.** `RestoreSuppressedMatches` clears `IgnoredMatches` unconditionally. If grype's `IgnoreRules`/`VexProcessor` are ever wired up in `GrypeAdapter`, those entries would be silently promoted to `Matches`. Restricting the restore to entries whose `AppliedIgnoreRules` carry the signature `ApplySecurityExceptions` writes (a single rule with only `Vulnerability` set) would make the helper self-guarding instead of relying on a comment. Your call — the fail-open argument in the comment is reasonable as-is.

Fix 1 and 2 and I'm happy to merge this.

matthyx (approved):
Re-reviewed at `2c8dae0`. Both blockers are properly fixed — not papered over. Approving.

I verified them independently rather than reading the new tests:

* **Blocker 2** — my exact repro from the last round (one CVE on `pkgA` fixed / `pkgB` not-fixed, exception narrowed with `expiredOnFix: true`) now passes: `pkgA` is restored into `Matches` and only `pkgB` stays ignored. `IgnoredMatchKeys` keying on ID + artifact name + version is the right granularity, and the mirror widening case is covered too.
* **Blocker 1** — I checked the whole chain, not just the sentinel: drove `ScanCVE` through a real platform implementation (`MockPlatform`) backed by a CRD lister that fails, and the stored manifest keeps its `IgnoredMatches`. `GetSecurityExceptions` now joins the `List` errors instead of swallowing them, `GetCVEExceptions` reports `ErrExceptionsDegraded` (including the selector fail-closed case), and `SubmitCVE` correctly treats degraded as non-fatal. The additions-persist-freely / removals-only-when-complete split is exactly the right shape, and the mixed additions+removals case conservatively persisting nothing is the correct call.

Local verification on `2c8dae0`: `go build ./...` clean, `go vet` clean, `go test ./...` fully green (the two `gofmt -l` hits, `core/ports/providers.go` and `adapters/v1/syft_test.go`, are pre-existing on `main` — not yours).

The optional hardening items I raised were picked up too — the `DeepCopy` short-circuit, restoring only exception-shaped ignored matches, and `vexGeneration` enabled in the `ScanCP` test.

**CI:** `build-and-push-image` is red on a GitHub Actions infra flake ("Failed to resolve action download info. Error: Service Unavailable"), not on your code — every real build/test check is green. Needs a re-run before merge.

### Follow-ups — none blocking, your call whether to fold them in here

1. **`crdErr != nil` now discards CRD exceptions that *did* list successfully.** `GetSecurityExceptions` joins the two `List` errors but still returns both slices, so a namespaced-list failure with a healthy cluster-scoped list yields usable `cseList` — and `backend.go`'s `else if` skips the merge entirely, dropping them for that scan. Relative to `main` that slightly widens the degraded window (previously the partial set was still applied). It is fail-open and self-healing, and the new gate stops it from being persisted as a removal on cache hits, so it is not urgent — but merging the partial set while keeping `degraded = true` is two lines and strictly better.
2. **The ~20-line persist gate is duplicated verbatim in `ScanCVE` and `ScanCP`.** Worth extracting into one helper so the two copies cannot drift.
3. **No `ScanCP` degraded-path test.** `ScanCVE` has both degraded cases covered; `ScanCP` runs the same logic through a separate copy (see 2).
4. **`isExceptionSourcedIgnore` is a heuristic, not a hard guard.** `v1beta1.IgnoreRule` has no VEX field, so a hypothetical VEX-sourced ignore would still look exception-shaped and get restored. Fine as-is — it is the fail-open direction — just don't lean on it as a guarantee later.
5. **Release note.** Still worth a line for the two behavior changes from the last round: `ScanCP` summary severity totals now include exception-suppressed CVEs on cache hits (aligning with cache-miss behavior), and the first rescan of cached images will restore suppressed findings and re-submit them to the backend.

Nice work on the turnaround — the plumbing you chose for the completeness signal is cleaner than what I sketched.

Issue #471, opened 2026-08-06
Grype-DB-container-backed tests panic instead of failing/skipping gracefully when Docker is unavailable or misconfigured


rootp1:
# Description
`TestScanService_NginxTest` (core/services/scan_test.go) calls `v1.NewGrypeAdapterFixedDB()` unconditionally. That function calls `startGrypeOfflineDBContainer` in adapters/v1/grype.go, which spins up a `quay.io/kubescape/grype-offline-db:v6-ci-only` container via testcontainers-go. When the local Docker environment does not match what testcontainers-go's internal `MustExtractDockerHost` expects (e.g. rootless Docker not detected, no Docker socket, misconfigured DOCKER_HOST), testcontainers-go panics internally. This is an unrecovered panic that crashes the whole test binary rather than failing or skipping just the one test.

# Environment
OS: any (reproducible on macOS/Linux without a fully compatible Docker setup, e.g. rootless Docker or no Docker daemon)
Version: current main branch

# Steps To Reproduce
1. On a machine without a working/expected Docker setup (e.g. no Docker daemon running, or a rootless Docker install that testcontainers-go's `MustExtractDockerHost` cannot detect).
2. Run `go test ./core/services/...` (specifically `TestScanService_NginxTest`).
3. Observe the test binary panics instead of failing/skipping only the affected test.

# Expected behavior
When the container runtime is not available/detectable, the test should call `t.Skip(...)` with a clear explanatory message, and `go test ./...` should not abort a package's entire test run because of this.

# Actual Behavior
`v1.NewGrypeAdapterFixedDB()` -> `NewGrypeAdapterFixedDBWithMatchers` (adapters/v1/grype.go) -> `startGrypeOfflineDBContainer` invokes `testcontainers.GenericContainer`, which panics inside testcontainers-go with a message like "rootless Docker not found" when it cannot determine the Docker host. This panic is unrecovered and propagates up through `TestScanService_NginxTest` (core/services/scan_test.go), crashing the entire test binary for the package instead of failing/skipping just that one test.

# Additional context
Suspected root cause: no Docker-availability probe exists before invoking `testcontainers.GenericContainer`. testcontainers-go's `MustExtractDockerHost` panics rather than returning an error in some misconfigured environments.

[...]

PR #474, opened 2026-08-06
fix: remove sidecar RPC-wide lock, fix Version() cache poisoning, and bound retryCount growth

### LINKED ISSUE #473: bug: sbom-scanner sidecar path serializes every scan behind one RPC-wide lock, permanently poisons its version cache on a single transient blip, and checks maxSBOMSize only after generation
# Description

The sidecar SBOM-scanner path (`pkg/sbomscanner/v1/server.go` + `adapters/v1/sidecar.go`, used whenever `SBOM_SCANNER_SOCKET` is set — the node-level SBOM generation deployment mode) never received the concurrency and robustness hardening that its sibling, the in-process `SyftAdapter` (`adapters/v1/syft.go`), was meant to get. Three problems in this one subsystem compound each other:

1. `scannerServer.CreateSBOM` (`pkg/sbomscanner/v1/server.go:122-137`) holds a single process-wide mutex for the *entire* RPC — registry pull **and** full Syft cataloging — so `scanConcurrency` has no effect once the sidecar is in the loop; every concurrent scan queues behind whichever one is currently running.
2. `SidecarSBOMAdapter.Version()` (`adapters/v1/sidecar.go:172-185`) uses `sync.Once` around a single, no-retry, 5s `Health()` call. One transient gRPC hiccup on process startup permanently locks the adapter's version string to `"unknown"` for the rest of the pod's lifetime — no retry, no TTL, no self-healing.
3. `maxSBOMSize` is enforced only *after* `syft.CreateSBOM` finishes building the entire in-memory package catalog, both in the sidecar (`server.go:278-290`) and in-process (`syft.go:298-312`, which even has its own `// TODO check memory usage and see if we can kill the goroutine` at line 247 acknowledging this) paths — the limit is documentation until generation has already paid the memory cost it's meant to prevent.

None of this is covered by the existing closed issues on this subsystem (#423, #425, #438, #441, #443, #381) or by #429/#430, which only ever touched `adapters/v1/syft.go`'s pull-time mutex — and that fix was never actually merged into `main`, so the in-process adapter still has the identical global-mutex problem too (flagging separately, keeping this issue scoped to the sidecar-only findings).

# Environment

OS: N/A — static correctness/architecture bug, reproducible on any OS
Version: `main` (`pkg/sbomscanner/v1/server.go`, `adapters/v1/sidecar.go`, `adapters/v1/syft.go`)

# Steps To Reproduce

**#1 — serialization:**
1. Deploy kubevuln with the sbom-scanner sidecar enabled (`SBOM_SCANNER_SOCKET` set) and `scanConcurrency` > 1.
2. Trigger several scans for different images at once (e.g. a rollout touching many workloads).
3. Watch sidecar scan latency: every scan after the first blocks on `scannerServer.mu` until the in-flight one's registry pull *and* full Syft run finish — there's no overlap regardless of `scanConcurrency`.

**#2 — version poisoning:**
1. Start kubevuln with the sidecar configured, and let the sidecar's very first `Health()` RPC fail or time out once (a Unix-socket hiccup right after the sidecar comes up is enough — this is a race every pod hits on cold start, not a contrived scenario).
2. Call `SidecarSBOMAdapter.Version()` once during that window (it's called from nearly every path in `core/services/scan.go`).
3. `versionOnce` fires, `versionStr` is set to `"unknown"`, and every later call to `Version()` returns `"unknown"` for the lifetime of the pod — even though the sidecar recovers immediately after.

**#3 — size check timing:**
1. Configure a modest `maxSBOMSize` and scan an image that stays under `maxImageSize` but is engineered to produce a huge number of packages/files (same DoS shape `maxImageSize` itself defends against, just one step later in the pipeline).
2. Observe `syft.CreateSBOM` running to completion — building the full in-memory catalog — before `size.Of(syftSBOM)` is ever compared against the configured limit.
3. Memory usage during generation is bounded only by the container's memory limit (OOM-kill), not by the configured `maxSBOMSize`.

# Expected behavior

- Concurrent scans against the sidecar should run in parallel, bounded the same way the in-process adapter bounds concurrency — not serialize behind one RPC-wide lock.
- A single transient health-check failure should not permanently corrupt the adapter's cached version string; `Version()` should recover once the sidecar is healthy again.
- `maxSBOMSize` should bound memory used *during* SBOM generation, not just get checked against a result that's already fully built.

# Actual Behavior

- `scannerServer.CreateSBOM` locks `s.mu` for the whole handler body (pull + generation), so `scanConcurrency` is silently defeated for any deployment using the sidecar.
- `SidecarSBOMAdapter.Version()` permanently caches `"unknown"` after one failed health check via `sync.Once`, with no way to recover short of restarting the pod. Because `Version()` is the SBOM/CVE storage cache key (`core/services/scan.go` lines 130, 217, 222, 235, 447, 459, 891, 912, 934, 960, 979), this corrupts shared state, not just a log line: `repositories/apiserver.go:1146` does
  ```go
  if semver.Compare(manifest.Spec.Metadata.Tool.Version, SBOMCreatorVersion) == -1 {
      // discard stored manifest, treat as outdated, regenerate
  }
  ```
  and per `golang.org/x/mod/semver`, an invalid version string always sorts below a valid one. A poisoned replica writes SBOM/CVE manifests tagged `Tool.Version: "unknown"` via `StoreSBOM` (`repositories/apiserver.go:1178`, which writes unconditionally — no staleness guard on write, only on read). Every replica that reads that manifest afterwards — including healthy ones with a real version string — sees `Compare("unknown", "v1.2.3") == -1` and discards it as outdated, forcing a full re-pull + re-scan. One pod's cold-start race can quietly turn the shared cache into a churn machine for every image it touches, for as long as that pod lives.
- `maxSBOMSize` is only compared after `syft.CreateSBOM` returns, in both `server.go` and `syft.go` — it's a post-hoc check, not a bound on the work itself.

# Additional context

- Contrast with `NewSBOMScannerClient` (`pkg/sbomscanner/v1/client.go:45-89`), which gates its own `Health()` call behind exponential backoff (`cenkalti/backoff/v5`) bounded by a configurable readiness timeout — the retry discipline already exists in this codebase, `Version()` just doesn't use it.
- `maxImageSize` *is* enforced early, via `context.WithValue(ctx, image.MaxImageSize, ...)` passed into `syft.GetSource`, so oversized images are rejected before download completes. `maxSBOMSize` has no equivalent hook into `syft.CreateSBOM` — Syft doesn't appear to expose one, so the only place left to check today is after the fact.
- Small, same-file addendum worth folding into the same PR rather than filing separately: `SidecarSBOMAdapter.retryCount` (`adapters/v1/sidecar.go:33,59`) is only cleared on success or once `maxCrashRetries` is hit. An image that crashes the scanner once or twice and is then never scanned again (common with CI-generated unique tags/digests) leaves a permanent entry — slow, unbounded map growth over the adapter's lifetime. Same "no self-healing" theme as the version cache, so it fits naturally alongside that fix.
- Affected files: `pkg/sbomscanner/v1/server.go`, `adapters/v1/sidecar.go`, `adapters/v1/syft.go` (shares the size-check-ordering issue), `repositories/apiserver.go` (where the poisoned version string actually corrupts stored state), `core/services/scan.go` (call sites of `sbomCreator.Version()`).

# Proposed Solution

1. **Serialization:** replace `scannerServer.mu` (whole-RPC mutex) with either no locking at all (each RPC already downloads into its own temp dir, nothing shared needs guarding) or, if some shared Syft/stereoscope state genuinely needs bounding, a bounded semaphore sized from config — matching whatever the in-process adapter uses so the two paths don't silently diverge again.
2. **Version cache:** stop using `sync.Once` for something that can fail. Cache the version only on success (e.g. an `atomic.Pointer[string]` written only in the non-error branch), and let the error branch return a per-call fallback without poisoning the cached slot — optionally reusing the existing backoff pattern from `client.go` for the first successful resolution.
3. **Size check:** see if Syft's cataloger API exposes any incremental hook to short-circuit generation early past `maxSBOMSize`; if not, document plainly in `docs/CONFIGURATION.md` that `maxSBOMSize` is a post-hoc check backed by the container memory limit, not a hard pre-emptive bound — so operators aren't relying on a guarantee that doesn't hold.
4. **retryCount cleanup:** cap the map or add periodic/opportunistic eviction of stale entries, alongside the existing success/exhaustion clears.
5. Add tests: two concurrent `CreateSBOM` calls actually overlapping in time; `Version()` recovering after one failed then one successful `Health()` call. Run `go test ./adapters/... ./pkg/sbomscanner/... -race` to confirm no new data races from the locking change.

bhuvan-somisetty:
## Overview

The sidecar SBOM-scanner path (used whenever `SBOM_SCANNER_SOCKET` is set) never got the concurrency/robustness hardening the in-process Syft adapter was meant to get. Three problems in that subsystem compounded each other; this PR fixes all three plus a small related cleanup in the same file.

## What changed

1. **Removed the whole-RPC mutex serializing every sidecar scan.** `scannerServer.CreateSBOM` (`pkg/sbomscanner/v1/server.go`) held `s.mu` around the *entire* handler — registry pull and full Syft cataloging — so `scanConcurrency` had no effect once the sidecar was in the loop; every concurrent scan queued behind whichever one was currently running. Each RPC already downloads into its own temp dir (`file.NewTempDirGenerator`) and builds its own SBOM, and `s.version` is set once at construction and never mutated — nothing shared actually needed guarding. The lock is gone; left a comment explaining why so it doesn't get silently reintroduced.
2. **Fixed `SidecarSBOMAdapter.Version()`'s cache poisoning.** It used `sync.Once` around a single, no-retry, 5s `Health()` call. One transient gRPC hiccup (e.g. right after the sidecar starts) permanently cached `"unknown"` as the version for the rest of the pod's life. That's not just a cosmetic string: `Version()` is the SBOM/CVE storage cache key, and `repositories/apiserver.go` compares it via `semver.Compare`, which treats an invalid version string as always older than a valid one — so a poisoned replica writes manifests tagged `"unknown"`, and every replica (including healthy ones) then discards them as outdated and regenerates on every read. Replaced `sync.Once` with a cache that only sticks on a successful `Health()` call; a failure just means the next call retries instead of latching forever.
3. **Documented the `maxSBOMSize` enforcement-timing limitation.** It's checked only after `syft.CreateSBOM` finishes building the full in-memory catalog, in both the sidecar (`server.go`) and in-process (`syft.go`) paths. Syft doesn't expose an incremental size hook, so this was already the best available check — it just wasn't written down as a post-hoc check backed by the scanner container's memory limit rather than a hard pre-emptive bound. Added that caveat to `docs/CONFIGURATION.md` (same pattern as the existing `scannerReadinessTimeout` note) and inline at both call sites.
4. **Bounded `retryCount` growth.** Crash-retry entries were only ever cleared on success or once `maxCrashRetries` was hit, so an image that crashed the scanner once and was never scanned again (common with CI-generated unique tags/digests) left a permanent entry — slow, unbounded map growth over the adapter's lifetime. Added a `crashRetry{count, lastSeen}` struct and an opportunistic TTL sweep (`retryCountTTL = 1h`) that runs each time `handleCrash` touches the map.

## Tests

- `TestSidecarSBOMAdapter_Version_RecoversAfterTransientFailure` (new): asserts `Version()` returns `"unknown"` on a failed `Health()` call, then recovers to the real version on the next call instead of staying pinned to `"unknown"`.
- `TestSidecarSBOMAdapter_CreateSBOM_CrashRetry_EvictsStaleEntries` (new): asserts a stale `retryCount` entry gets swept out the next time the map is touched by a different image's crash.
- Existing `TestSidecarSBOMAdapter_*` suite in `adapters/v1/sidecar_test.go` — unchanged behavior, all still pass.
- Full `pkg/sbomscanner/v1` package (unit + integration tests over the real gRPC/Unix-socket path) passes with the mutex removed.

## How to Test

```bash
go build ./...
go vet ./adapters/... ./pkg/sbomscanner/...
go test ./adapters/v1/... -run TestSidecarSBOMAdapter -v
go test ./pkg/sbomscanner/...
```

Not touched by this PR: `Test_grypeAdapter_DBVersion` in `adapters/v1` fails locally on Windows with a `testcontainers`/Docker panic (`rootless Docker is not supported on Windows`) — that's pre-existing and unrelated, tracked separately in #471.

## Related issues/PRs:

* Resolved #473

## Summary by CodeRabbit

* **Bug Fixes**
  * Version checks now retry after temporary health-check failures and cache only successful results.
  * Concurrent version checks are consolidated to avoid redundant health requests.
  * Crash-retry records expire after one hour, preventing stale entries from affecting processing.

* **Performance**
  * SBOM generation can now process multiple requests concurrently.

* **Documentation**
  * Clarified that SBOM size limits are checked after generation and memory usage is governed by the scanner’s container limit.

### DIFF
--- a/adapters/v1/sidecar.go
+++ b/adapters/v1/sidecar.go
@@ -83,13 +83,18 @@ func (s *SidecarSBOMAdapter) CreateSBOM(ctx context.Context, name, imageID, imag
 		imageID = NormalizeImageID(imageID, imageTag)
 	}
 
+	// Resolved once and reused for both fields below: two separate s.Version() calls could
+	// otherwise observe different results (e.g. the first hits a transient Health() failure
+	// and returns "unknown" while a concurrent call succeeds moments later), tagging the same
+	// SBOM with two different creator versions.
+	version := s.Version()
 	domainSBOM := domain.SBOM{
 		Name:               name,
-		SBOMCreatorVersion: s.Version(),
+		SBOMCreatorVersion: version,
 		SBOMCreatorName:    "syft",
 		Annotations: map[string]string{
 			helpersv1.ImageIDMetadataKey:     imageID,
-			helpersv1.ToolVersionMetadataKey: s.Version(),
+			helpersv1.ToolVersionMetadataKey: version,
 		},
 		Labels: tools.LabelsFromImageID(imageID),
 	}
--- a/adapters/v1/sidecar.go
+++ b/adapters/v1/sidecar.go
@@ -14,6 +14,7 @@ import (
 	"github.com/kubescape/kubevuln/core/ports"
 	"github.com/kubescape/kubevuln/internal/tools"
 	sbomscanner "github.com/kubescape/kubevuln/pkg/sbomscanner/v1"
+	"golang.org/x/sync/singleflight"
 )
 
 const (
@@ -47,8 +48,9 @@ type SidecarSBOMAdapter struct {
 	mu         sync.Mutex
 	retryCount map[string]*crashRetry
 
-	versionMu  sync.Mutex
-	versionStr string
+	versionMu    sync.Mutex
+	versionStr   string
+	versionGroup singleflight.Group
 }
 
 var _ ports.SBOMCreator = (*SidecarSBOMAdapter)(nil)
@@ -211,6 +213,13 @@ func (s *SidecarSBOMAdapter) evictStaleRetriesLocked() {
 // on this pod to "unknown" and starve the storage cache (see #473) — the next call simply
 // retries instead of returning a value baked in by sync.Once for the rest of the process's
 // life.
+//
+// Concurrent callers that miss the cache are coalesced through versionGroup into a single
+// in-flight Health() call: without this, two overlapping callers could each run their own
+// Health() independently, and if one happened to fail while the other succeeded, the failing
+// caller would return "unknown" even though the version was, at that same moment, already
+// known — an avoidable cache miss for whatever that caller was about to look up. Coalescing
+// means every overlapping caller observes the same outcome as the single underlying call.
 func (s *SidecarSBOMAdapter) Version() string {
 	s.versionMu.Lock()
 	if s.versionStr != "" {
@@ -219,14 +228,21 @@ func (s *SidecarSBOMAdapter) Version() string {
 	}
 	s.versionMu.Unlock()
 
-	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
-	defer cancel()
-	version, _, err := s.client.Health(ctx)
+	v, err, _ := s.versionGroup.Do("version", func() (any, error) {
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+		version, _, err := s.client.Health(ctx)
+		if err != nil {
+			return "", err
+		}
+		return version, nil
+	})
 	if err != nil {
 		logger.L().Warning("failed to get scanner version", helpers.Error(err))
 		return "unknown"
 	}
 
+	version := v.(string)
 	s.versionMu.Lock()
 	s.versionStr = version
 	s.versionMu.Unlock()
--- a/adapters/v1/sidecar.go
+++ b/adapters/v1/sidecar.go
@@ -16,7 +16,22 @@ import (
 	sbomscanner "github.com/kubescape/kubevuln/pkg/sbomscanner/v1"
 )
 
-const maxCrashRetries = 3
+const (
+	maxCrashRetries = 3
+	// retryCountTTL bounds how long a crashed image's retry count is remembered. Without
+	// this, an image that crashes the scanner once or twice and is then never scanned
+	// again (common with CI-generated unique tags/digests) would leave a permanent entry
+	// in retryCount, since the map was previously only ever cleared on success or once
+	// maxCrashRetries was hit (see #473).
+	retryCountTTL = 1 * time.Hour
+)
+
+// crashRetry tracks a per-image crash-retry count and when it was last touched, so stale
+// entries (images that crashed once and were never retried) can be evicted.
+type crashRetry struct {
+	count    int
+	lastSeen time.Time
+}
 
 // SidecarSBOMAdapter implements ports.SBOMCreator by delegating SBOM generation
 // to the sbom-scanner sidecar container via gRPC over a Unix domain socket.
@@ -30,10 +45,10 @@ type SidecarSBOMAdapter struct {
 	memoryLimit       string
 
 	mu         sync.Mutex
-	retryCount map[string]int
+	retryCount map[string]*crashRetry
 
-	versionOnce sync.Once
-	versionStr  string
+	versionMu  sync.Mutex
+	versionStr string
 }
 
 var _ ports.SBOMCreator = (*SidecarSBOMAdapter)(nil)
@@ -56,7 +71,7 @@ func NewSidecarSBOMAdapter(
 		scanTimeout:       scanTimeout,
 		scanEmbeddedSBOMs: scanEmbeddedSBOMs,
 		memoryLimit:       memoryLimit,
-		retryCount:        make(map[string]int),
+		retryCount:        make(map[string]*crashRetry),
 	}
 }
 
@@ -138,8 +153,15 @@ func (s *SidecarSBOMAdapter) CreateSBOM(ctx context.Context, name, imageID, imag
 
 func (s *SidecarSBOMAdapter) handleCrash(ctx context.Context, name, imageID, imageTag string, options domain.RegistryOptions, domainSBOM domain.SBOM, crashErr error) (domain.SBOM, error) {
 	s.mu.Lock()
-	s.retryCount[imageID]++
-	retries := s.retryCount[imageID]
+	s.evictStaleRetriesLocked()
+	entry := s.retryCount[imageID]
+	if entry == nil {
+		entry = &crashRetry{}
+		s.retryCount[imageID] = entry
+	}
+	entry.count++
+	entry.lastSeen = time.Now()
+	retries := entry.count
 	s.mu.Unlock()
 
 	logger.L().Warning("SBOM scanner sidecar crashed during scan",
@@ -169,19 +191,46 @@ func (s *SidecarSBOMAdapter) handleCrash(ctx context.Context, name, imageID, ima
 	return domainSBOM, crashErr
 }
 
-func (s *SidecarSBOMAdapter) Version() string {
-	s.versionOnce.Do(func() {
-		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
-		defer cancel()
-		version, _, err := s.client.Health(ctx)
-		if err != nil {
-			logger.L().Warning("failed to get scanner version", helpers.Error(err))
-			s.versionStr = "unknown"
-			return
+// evictStaleRetriesLocked removes retry-count entries untouched for longer than
+// retryCountTTL, bounding the map's growth for images that crash once and are never
+// scanned again. Called with s.mu held; piggybacks on handleCrash rather than a background
+// goroutine since crashes are rare and this keeps the map bounded without extra machinery.
+func (s *SidecarSBOMAdapter) evictStaleRetriesLocked() {
+	cutoff := time.Now().Add(-retryCountTTL)
+	for imageID, entry := range s.retryCount {
+		if entry.lastSeen.Before(cutoff) {
+			delete(s.retryCount, imageID)
 		}
-		s.versionStr = version
-	})
-	return s.versionStr
+	}
+}
+
+// Version returns the sidecar's Syft version, used as the SBOM/CVE storage cache key
+// (see repositories/apiserver.go's semver comparison against the stored manifest). It only
+// caches a successful Health() result: a failed/timed-out health check is never written to
+// versionStr, so a transient blip (e.g. on cold start) does not permanently pin every scan
+// on this pod to "unknown" and starve the storage cache (see #473) — the next call simply
+// retries instead of returning a value baked in by sync.Once for the rest of the process's
+// life.
+func (s *SidecarSBOMAdapter) Version() string {
+	s.versionMu.Lock()
+	if s.versionStr != "" {
+		defer s.versionMu.Unlock()
+		return s.versionStr
+	}
+	s.versionMu.Unlock()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+	version, _, err := s.client.Health(ctx)
+	if err != nil {
+		logger.L().Warning("failed to get scanner version", helpers.Error(err))
+		return "unknown"
+	}
+
+	s.versionMu.Lock()
+	s.versionStr = version
+	s.versionMu.Unlock()
+	return version
 }
 
 func (s *SidecarSBOMAdapter) GetMaxImageSize() int64 {
--- a/adapters/v1/sidecar_test.go
+++ b/adapters/v1/sidecar_test.go
@@ -18,6 +18,7 @@ import (
 // mockScannerClient implements sbomscanner.SBOMScannerClient for testing
 type mockScannerClient struct {
 	createSBOMFunc func(ctx context.Context, req sbomscanner.ScanRequest) (*sbomscanner.ScanResult, error)
+	healthFunc     func(ctx context.Context) (string, bool, error)
 	healthVersion  string
 	healthReady    bool
 	healthErr      error
@@ -31,6 +32,9 @@ func (m *mockScannerClient) CreateSBOM(ctx context.Context, req sbomscanner.Scan
 }
 
 func (m *mockScannerClient) Health(ctx context.Context) (string, bool, error) {
+	if m.healthFunc != nil {
+		return m.healthFunc(ctx)
+	}
 	return m.healthVersion, m.healthReady, m.healthErr
 }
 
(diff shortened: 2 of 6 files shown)


bhuvan-somisetty:
Fixed all three:

1. \`Version()\` now coalesces overlapping health checks through \`singleflight\` instead of letting each caller race its own \`Health()\` call good catch, the mixed success/failure case was real. Added a regression test with two goroutines deterministically synchronized around an in-flight call (no sleep-and-hope on the assertion itself).
2. Reworded the \`maxSBOMSize\` doc line to name whichever container actually runs Syft, since it's not always the sidecar.
3. Toned down the \`CreateSBOM\` comment it's accurate to say concurrent RPCs are bounded by the caller, but a timed-out scan's background goroutine can still keep chewing memory after the RPC returns, since Syft doesn't support cancellation. Comment now says that explicitly instead of implying full resource bounding.

All green locally (build/vet/tests), pushed.

bhuvan-somisetty:
Good catch fixed. \`CreateSBOM\` was calling \`s.Version()\` twice (once for \`SBOMCreatorVersion\`, once for the \`ToolVersionMetadataKey\` annotation), and with the coalescing fix from the previous round those two calls could still land on different results if a concurrent caller resolved the version in between the two. Now it's resolved once into a local var and reused for both, so a stored SBOM can't end up tagged with two different creator versions. Added a test for it too.

All green, pushed.

bhuvan-somisetty:
@matthyx whenever you get a chance, would appreciate a look at this one. Happy to make any changes you'd like.

pkg/sbomscanner/v1/server.go
```diff
@@ -132,10 +130,22 @@ func NewScannerServer() pb.SBOMScannerServer {
...
 }
 
+// CreateSBOM handles one scan per call, with no state shared across concurrent calls: each
+// invocation downloads into its own temp dir (via its own file.NewTempDirGenerator) and builds
+// its own SBOM from its own source. s.version is set once in NewScannerServer and never
+// mutated, so it's safe to read concurrently without a lock. Do not add a mutex/semaphore
+// around this method — a previous version serialized the whole RPC (pull + generation) behind
+// a single process-wide lock, which defeated the caller's scanConcurrency entirely regardless
+// of its configured value (see #473); the number of concurrent in-flight RPCs is bounded by
+// the caller instead, matching scanConcurrency.
```
matthyx, line 140: Two notes on this comment block — both about wording, the code itself is fine.

**a) The in-process adapter has the same lock.** `adapters/v1/syft.go:249-250` takes `pullMutex` and holds it via a function-scoped `defer` across the whole `dl.Run` — pull *and* full cataloging, not just the pull its comment mentions. So `SyftAdapter` serializes SBOM generation exactly the way `scannerServer` did. The PR description frames this as the sidecar missing hardening the in-process adapter already has, but the situation is the reverse: after this PR the sidecar path honours `scanConcurrency` and the in-process path still doesn't. Either narrow `pullMutex` to the actual pull with the same reasoning, or state here (and in the description) that the sidecar is deliberately the only concurrent path — otherwise the next reader hits the asymmetry with no explanation.

**b) "Do not add a mutex/semaphore" is broader than the evidence.** What defeated `scanConcurrency` was serializing to *one* — a semaphore sized above 1 wouldn't. Given the sidecar exists specifically for memory isolation (36cd416) and its OOM path is the crash-retry protocol, a configurable in-sidecar cap is the natural mitigation if operators find that `scanConcurrency` concurrent Syft catalogings don't fit the container's limit. Suggest narrowing the prohibition to "don't reintroduce a lock that serializes this to a single scan" so a future bounded semaphore isn't pre-vetoed by a comment.

docs/CONFIGURATION.md
```diff
@@ -167,7 +167,7 @@ The main configuration file. All options can be overridden via environment varia
 | Option | Type | Default | Description |
 |--------|------|---------|-------------|
 | `maxImageSize` | int | `536870912` | Maximum image size to scan in bytes (default: 512 MB) |
-| `maxSBOMSize` | int | `20971520` | Maximum SBOM size in bytes (default: 20 MB) |
+| `maxSBOMSize` | int | `20971520` | Maximum SBOM size in bytes (default: 20 MB). Enforced after SBOM generation completes, not during it — Syft doesn't expose an incremental size hook, so an oversized SBOM is rejected only once it has already been built in memory. During generation, actual memory use is bounded by the memory limit of whichever container runs Syft, not by this value: the `sbom-scanner` sidecar container when `SBOM_SCANNER_SOCKET` is configured, or the main kubevuln container otherwise (in-process `SyftAdapter`). |
```
matthyx, line 170: Good clarification, and it resolves CodeRabbit's point about naming the right container.

One gap left in this same file: the tuning guidance at line 557 — *"Set memory limits to at least `2 * maxImageSize * scanConcurrency`"* — still doesn't say **which** container. That was harmless before this PR, because the RPC-wide lock meant the `sbom-scanner` container only ever ran one Syft regardless of `scanConcurrency`, so sizing it for 1x happened to be enough. After this PR that container genuinely needs the `scanConcurrency` multiple, and getting it wrong shows up as OOM-kills that take down every concurrent scan. Since this is the operator-visible consequence of the change, it belongs in this PR:

> **Recommendation:** Set memory limits to at least `2 * maxImageSize * scanConcurrency`. When `SBOM_SCANNER_SOCKET` is configured, this applies to the `sbom-scanner` sidecar container, which runs up to `scanConcurrency` SBOM generations in parallel; otherwise it applies to the main kubevuln container.

adapters/v1/sidecar_test.go
```diff
@@ -178,3 +213,123 @@ func TestSidecarSBOMAdapter_Version(t *testing.T) {
...
+// flaky, since a goroutine that arrives after the in-flight call has already completed
+// legitimately starts a fresh call rather than a bug.
+func TestSidecarSBOMAdapter_Version_CoalescesConcurrentCalls(t *testing.T) {
+	var callCount int32
+	started := make(chan struct{})
+	release := make(chan struct{})
+	mock := &mockScannerClient{
+		healthFunc: func(ctx context.Context) (string, bool, error) {
+			atomic.AddInt32(&callCount, 1)
+			close(started)
```
matthyx, line 265: `close(started)` runs on *every* `Health()` call, so the test only survives as long as the behaviour it's testing holds. I checked what happens when it doesn't: reverting `Version()` to the plain double-checked-lock form (no singleflight) makes this test panic with `close of closed channel` from inside the mock, which aborts the whole `adapters/v1` test binary and reports the failure at `sidecar_test.go:265` instead of at the `callCount` assertion that actually describes the bug.

It does catch the regression, so this is a legibility nit rather than a correctness one — but a one-line guard keeps the failure pointing at the right place:

```go
var startOnce sync.Once
mock := &mockScannerClient{
	healthFunc: func(ctx context.Context) (string, bool, error) {
		atomic.AddInt32(&callCount, 1)
		startOnce.Do(func() { close(started) })
		<-release
		return "v0.100.0", true, nil
	},
}
```

With that, a regression fails cleanly on `assert.Equal(t, int32(1), callCount, ...)` and the rest of the package still runs.

adapters/v1/sidecar.go
```diff
@@ -169,19 +198,60 @@ func (s *SidecarSBOMAdapter) handleCrash(ctx context.Context, name, imageID, ima
...
 	})
-	return s.versionStr
+	if err != nil {
+		logger.L().Warning("failed to get scanner version", helpers.Error(err))
+		return "unknown"
+	}
+
+	version := v.(string)
+	s.versionMu.Lock()
+	s.versionStr = version
```
matthyx, line 252: Minor edge case worth a guard, since it lands in the exact failure mode this PR is fixing: the sidecar's own `packageVersion()` (`pkg/sbomscanner/v1/server.go:364-374`) falls back to the literal string `"unknown"` when `debug.ReadBuildInfo()` carries no `github.com/anchore/syft` entry. That arrives here as a **successful** `Health()`, so `"unknown"` gets written to `versionStr` and latched for the life of the process — the same poisoned cache key as before, just reached through a healthy path instead of a transient failure, and `semver.Compare` treats it the same way.

Cheap fix while you're here:

```go
version := v.(string)
if version != "" && version != "unknown" {
	s.versionMu.Lock()
	s.versionStr = version
	s.versionMu.Unlock()
}
return version
```

Rare in practice (a normally-built binary has syft in `bi.Deps`), so entirely optional.

matthyx (approved):
## Review — approving, with notes

Reviewed the three-commit stack against `main`, plus the surrounding code the change reasons about. Verdict: **good to merge.** Nothing below blocks it; item 2 (the `CONFIGURATION.md` tuning line) is the only one I'd like folded into this PR, and it's a one-liner.

### What I verified locally (Linux, go1.25.8, PR head `46cbfc7`)

- `go build ./...` and `go vet ./adapters/... ./pkg/sbomscanner/...` — clean.
- `go test -race -count=1 ./adapters/v1/... -run TestSidecarSBOMAdapter` — ok.
- `go test -race -count=1 ./pkg/sbomscanner/...` — ok (unit + the gRPC/unix-socket integration tests, with the lock removed).
- `go test -count=1 ./adapters/...` — ok, 74s, including `Test_grypeAdapter_DBVersion` (the testcontainers failure noted in the description is Windows-specific; it passes here).
- `go test -race -count=20 ./adapters/v1/... -run TestSidecarSBOMAdapter_Version` — no flakes. The `time.Sleep(50ms)` is a timing *hint*, not a correctness dependency: if the second caller arrives after the in-flight call completes, it hits the now-populated cache, so `callCount` stays 1 either way.
- Mutation check on the new coalescing test: reverting `Version()` to the pre-singleflight double-checked-lock form does make `TestSidecarSBOMAdapter_Version_CoalescesConcurrentCalls` fail — but as a panic rather than an assertion. See the inline note on `sidecar_test.go`.

### The three substantive claims, checked against the code

1. **Lock removal is safe.** `scannerServer` really has no shared mutable state across calls: `version` is set once in `NewScannerServer` and only read, each call builds its own `file.NewTempDirGenerator("stereoscope")` and its own source, and the only package-level var in `server.go` (`gcpCredsFn`) is swapped by tests only.
2. **The caller-side bound is real.** `controllers/http.go:33` builds `workerpool.New(concurrency)` from `c.ScanConcurrency` (`cmd/http/main.go:122`) and every scan entrypoint goes through `workerPool.Submit`, so concurrent in-flight `CreateSBOM` RPCs really are capped at `scanConcurrency`. The comment's claim holds.
3. **The `sync.Once` poisoning was real and this fixes it.** `Version()` feeds `SBOMCreatorVersion`, and `repositories/apiserver.go` compares it with `semver.Compare`, where an invalid string sorts below any valid one — so a latched `"unknown"` did make every replica discard those manifests. Caching only on success + `singleflight` is the right shape, and resolving the version once per `CreateSBOM` closes the last window where one stored SBOM could carry two versions.

### Follow-ups I'd track separately (not this PR)

- **Crash attribution under real concurrency.** `ErrScannerCrashed` is raised from `codes.Unavailable`/`Aborted` — i.e. the sidecar *process* died, which fails every in-flight RPC, not just the guilty one. Now that `scanConcurrency > 1` actually parallelizes, an OOM triggered by one large image increments `retryCount` for every image scanning alongside it; three such events and a perfectly fine image gets stored as `TooLarge` with `ReasonScannerOOM` and the memory-limit annotation. This is pre-existing (RPCs blocked on the old mutex died the same way), but the change makes it materially more reachable. Worth an issue: only count a crash against an image when it was the sole in-flight scan, or attribute by resident scan size.
- **The dropped TODO in `syft.go`** (`// TODO check memory usage and see if we can kill the goroutine`) describes a problem that is still open — the new `server.go` comment restates it and points at anchore/syft#3705. Fine to drop the marker, but consider leaving the syft#3705 pointer in `syft.go` too so the in-process path keeps the same breadcrumb.
- **Pre-existing latent race, unrelated to this PR:** the `dl.Run` closure assigns the outer `syftSBOM`/`err` while the caller reads `err` immediately after `deadline.ErrTimedOut` — and on timeout the abandoned goroutine is still running. Same pattern in `server.go` and `syft.go`. Not introduced here, but the abandoned-goroutine comment this PR adds is the natural place to eventually link it.

PR #476, opened 2026-08-06
fix: fetch the full summary in StoreCVESummaryStub so the data-preservation guard works

### LINKED ISSUE #475: bug: StoreCVESummaryStub's metadata-only Get returns a zero-Spec summary, defeating the data-preservation guard and erasing real summary data
## Description

`StoreCVESummaryStub` (`repositories/apiserver.go:626-701`) is designed to never overwrite a real summary's `Spec` — `summaryHasVulnerabilityData` (`:672`) and the "keep any existing Spec" intent (`:675`) exist for exactly that. However, its retry-path `Get` (`:667`) requests `metav1.GetOptions{ResourceVersion: "metadata"}`. That option is honored by the pinned storage backend: `kubescape/storage@v0.0.258` (`pkg/registry/file/storage.go:422-436`) returns an `ObjectMeta`-only object (SQLite metadata blob, no `Spec`) whenever `resourceVersion == "metadata"` (`ResourceVersionMetadata = "metadata"`, `pkg/apis/softwarecomposition/register.go:27`). `VulnerabilityManifestSummary` is served by exactly this file-based backend.

With `Spec` zeroed, `summaryHasVulnerabilityData` (`apiserver.go:614-624`) always returns `false`, so the guard never fires, and the `Update` (`:676-678`) writes the zeroed `Spec` back — silently erasing a real summary's `Severities` and `Vulnerabilities`.

The safe siblings are unaffected only because they overwrite `Spec` wholesale:
- `StoreCVE` (`:339`), `StoreCVESummary` (`:580`), `StoreSBOM` (`:1224/:1261`) use the same metadata-Get but set `result.Spec = manifest.Spec` before `Update` (`StoreCVESummary:587`).
- `StoreVEX` documents this exact pitfall (`:756-763`) and deliberately uses plain `GetOptions{}` (`:765`) because it merges into `Spec`.

`StoreCVESummaryStub` is the one method that both uses the metadata-Get and preserves the fetched `Spec` — so the documented-safe pattern is unsafe here, and the guard is dead in production.

---

## Environment

- **OS:** N/A — storage-layer code bug
- **Version:** `main` (current); storage `kubescape/storage v0.0.258`

---

## Steps To Reproduce

1. A workload/image scans successfully → a real `VulnerabilityManifestSummary` is stored with severities + vulnerabilities.
2. A later scan of the same workload hits `ReasonImageSchemaUnsupported` (scanner/schema drift, or the workload's image becomes an unsupported format) → `core/services/scan.go:269/:483` calls `StoreCVESummaryStub`.
3. `Create` returns `AlreadyExists` → the retry-path `Get` with `ResourceVersion: "metadata"` returns an object with a zeroed `Spec` → `summaryHasVulnerabilityData` returns `false` → the `Update` writes the zeroed `Spec` back.
4. The workload's summary silently loses its `Severities`/`Vulnerabilities` until a fresh successful scan rewrites it.

---

## Expected Behavior

A stub must never modify the `Spec` of a summary that already holds real scan data — the `summaryHasVulnerabilityData` guard must actually fire and leave the real summary untouched.

---

## Actual Behavior

The guard is unreliable in production: the metadata-only `Get` returns a zero-`Spec` object, so the check always fails and the `Update` erases the real summary's data. The existing behavioral test (`repositories/apiserver_test.go:730-757`, "update over a real summary preserves Spec") passes on both the buggy and the fixed code, because the fake clientset ignores `GetOptions` and always returns the full object — so no existing test can catch this.

---

## Additional Context

- **Root Cause:** `GetOptions{ResourceVersion: "metadata"}` is only safe for the sibling `Store*` methods because they overwrite `Spec` wholesale — the maintainers documented this in the `StoreVEX` NOTE (`apiserver.go:756-763`). `StoreCVESummaryStub` was grouped with those siblings while behaving like `StoreVEX` (`Spec`-preserving). Introduced by #370, whose stated intent was precisely "never overwrites an existing summary's `Spec` with the zeroed stub."

---

## Proposed Solution

Use plain `metav1.GetOptions{}` at `repositories/apiserver.go:667`, matching the `StoreVEX` pattern (`:765`):

```go
// repositories/apiserver.go:667
result, getErr := a.StorageClient.VulnerabilityManifestSummaries(workloadNamespace).Get(ctx, manifest.Name, metav1.GetOptions{})
```

The guard then sees the real `Spec` (real data → skip update), and when no real data exists the update preserves the existing `Spec`. Correct-by-construction regardless of storage-server behavior; the only cost is a slightly heavier read on the `AlreadyExists` path.

---

## Tests to Add

Add a mechanism test asserting the stub's `Get` no longer requests `ResourceVersion: "metadata"` (extend the existing `ctxCapturingVulnerabilityManifestSummaries` wrapper to capture `GetOptions`, `apiserver_test.go:1556-1564`) — red on `main` (captures `"metadata"`), green with the fix. Because the fake clientset ignores `GetOptions`, no behavioral test can reproduce the wipe, so this is the only reliable regression guard.

AdeshDeshmukh:
Fixes #475

## Overview

`StoreCVESummaryStub` (`repositories/apiserver.go:626-701`) is meant to never overwrite a real summary's `Spec` — `summaryHasVulnerabilityData` (`:672`) and the "keep any existing Spec" intent (`:675`). The retry-path `Get` (`:667`) used `metav1.GetOptions{ResourceVersion: "metadata"}`, which the pinned storage backend (`kubescape/storage@v0.0.258`) returns as an `ObjectMeta`-only object with a zeroed `Spec` (`pkg/registry/file/storage.go:422-436`; `ResourceVersionMetadata = "metadata"`, `pkg/apis/softwarecomposition/register.go:27`). The guard therefore sees an empty `Spec` and never fires, and the subsequent `Update` writes the zeroed `Spec` back — silently erasing a real summary's `Severities`/`Vulnerabilities`.

The siblings are safe with the metadata-`Get` only because they overwrite `Spec` wholesale (`StoreCVE:339`, `StoreCVESummary:580`, `StoreSBOM:1224/:1261`, e.g. `result.Spec = manifest.Spec` at `StoreCVESummary:587`). `StoreVEX` documents this exact pitfall (`:756-763`) and deliberately uses plain `GetOptions{}` (`:765`) because it merges into `Spec`. `StoreCVESummaryStub` is the one method that both uses the metadata-`Get` and preserves the fetched `Spec`. This change is correct-by-construction regardless of storage-server behavior.

---

## Change

`repositories/apiserver.go:667`:

```go
result, getErr := a.StorageClient.VulnerabilityManifestSummaries(workloadNamespace).Get(ctx, manifest.Name, metav1.GetOptions{})
```

One-token change: `{ResourceVersion: "metadata"}` → `{}`, matching the `StoreVEX` pattern (`:765`). The guard now sees the real `Spec` (real data → skip update); with no real data, the update preserves the existing `Spec`. Only cost is a slightly heavier read on the `AlreadyExists` path.

---

## Tests

`TestAPIServerStore_StoreCVESummaryStub_DoesNotUseMetadataGet` (`repositories/apiserver_test.go`).

**Why a mechanism test:** The existing behavioral test ("update over a real summary preserves Spec", `apiserver_test.go:730-757`) passes on both the buggy and the fixed code, because the fake clientset ignores `GetOptions` and always returns the full object. Only an assertion on the requested `GetOptions.ResourceVersion` can catch the regression. The test seeds a real summary (forcing the `AlreadyExists` → retry-path `Get`) and asserts the stub's `Get` uses an empty `ResourceVersion`.

Red on `main` / green with this change (verified by reverting the fix):

```text
# fix reverted (buggy code):
[FAIL] TestAPIServerStore_StoreCVESummaryStub_DoesNotUseMetadataGet
  Should be empty, but was metadata

# fix applied:
Go test: 1 passed in 1 packages
```

---

## Verification

- `go build ./...` — clean
- `go vet ./repositories/` — no issues
- `gofmt -l repositories/apiserver.go repositories/apiserver_test.go` — clean
- `go test ./repositories/ -run 'TestAPIServerStore_StoreCVESummaryStub' -count=1` — 5 passed (4 existing + new)
- `go test ./repositories/... -count=1` — 34 passed

---

## Checklist

- [x] DCO signed off

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed vulnerability summary retrieval to use complete resource data when updating existing summaries.
  * Existing vulnerability summaries are now preserved correctly during retries and updates.

* **Tests**
  * Added regression coverage to verify correct resource retrieval behavior.

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -664,7 +664,7 @@ func (a *APIServerStore) StoreCVESummaryStub(ctx context.Context, status string)
 				return err
 			}
 			// retrieve the latest version before attempting update
-			result, getErr := a.StorageClient.VulnerabilityManifestSummaries(workloadNamespace).Get(ctx, manifest.Name, metav1.GetOptions{ResourceVersion: "metadata"})
+			result, getErr := a.StorageClient.VulnerabilityManifestSummaries(workloadNamespace).Get(ctx, manifest.Name, metav1.GetOptions{})
 			if getErr != nil {
 				return getErr
 			}
--- a/repositories/apiserver_test.go
+++ b/repositories/apiserver_test.go
@@ -757,6 +757,49 @@ func TestAPIServerStore_StoreCVESummaryStub(t *testing.T) {
 	})
 }
 
+// TestAPIServerStore_StoreCVESummaryStub_DoesNotUseMetadataGet guards the
+// data-preservation guard in StoreCVESummaryStub: its retry-path Get must fetch
+// the full object, not request ResourceVersion "metadata" (which kubescape/
+// storage returns as an ObjectMeta-only object with a zeroed Spec). The fake
+// clientset ignores GetOptions, so a behavioral test cannot reproduce the
+// resulting data loss; only an assertion on the requested GetOptions can.
+func TestAPIServerStore_StoreCVESummaryStub_DoesNotUseMetadataGet(t *testing.T) {
+	workload := domain.ScanCommand{
+		Wlid:          "wlid://cluster-kind/namespace-local-path-storage/deployment-local-path-provisioner",
+		ContainerName: "local-path-provisioner",
+	}
+	ctx := context.WithValue(context.Background(), domain.WorkloadKey{}, workload)
+	ctx = context.WithValue(ctx, domain.TimestampKey{}, int64(1734957372))
+
+	clientset := fake.NewSimpleClientset()
+	wrapped := &ctxCapturingClient{SpdxV1beta1Interface: clientset.SpdxV1beta1()}
+	a := &APIServerStore{StorageClient: wrapped, Namespace: "kubescape"}
+
+	ns, err := GetCVESummaryK8sResourceNamespace(ctx)
+	require.NoError(t, err)
+	resourceName, err := GetCVESummaryK8sResourceName(ctx)
+	require.NoError(t, err)
+
+	// Seed a real summary so the stub's Create returns AlreadyExists and the
+	// retry-path Get executes.
+	real := &v1beta1.VulnerabilityManifestSummary{
+		ObjectMeta: metav1.ObjectMeta{Name: resourceName},
+		Spec: v1beta1.VulnerabilityManifestSummarySpec{
+			Severities: v1beta1.SeveritySummary{
+				High: v1beta1.VulnerabilityCounters{All: 3, Relevant: 1},
+			},
+		},
+	}
+	_, err = wrapped.VulnerabilityManifestSummaries(ns).Create(context.Background(), real, metav1.CreateOptions{})
+	require.NoError(t, err)
+
+	require.NoError(t, a.StoreCVESummaryStub(ctx, helpersv1.Incomplete))
+
+	require.NotNil(t, wrapped.vulnSummaries[ns], "summary sub-client must have been exercised")
+	require.Empty(t, wrapped.vulnSummaries[ns].getResourceVersion,
+		"StoreCVESummaryStub must not request the metadata-only Get")
+}
+
 func TestAPIServerStore_enrichSummaryManifestObjectLabels(t *testing.T) {
 	ctx := context.Background()
 
@@ -1555,11 +1598,13 @@ func (w *ctxCapturingOVECs) Update(ctx context.Context, obj *v1beta1.OpenVulnera
 
 type ctxCapturingVulnerabilityManifestSummaries struct {
 	spdxv1beta1.VulnerabilityManifestSummaryInterface
-	getCtx context.Context
+	getCtx             context.Context
+	getResourceVersion string
 }
 
 func (w *ctxCapturingVulnerabilityManifestSummaries) Get(ctx context.Context, name string, opts metav1.GetOptions) (*v1beta1.VulnerabilityManifestSummary, error) {
 	w.getCtx = ctx
+	w.getResourceVersion = opts.ResourceVersion
 	return w.VulnerabilityManifestSummaryInterface.Get(ctx, name, opts)
 }
 


matthyx (approved):
Reviewed and independently verified against the pinned dependency and a local build. **No blockers — approving.**

## Verification

I re-derived the bug from source rather than taking the description on faith:

- `kubescape/storage@v0.0.258`, `pkg/registry/file/storage.go:421-436` — `get()` short-circuits on `opts.ResourceVersion == softwarecomposition.ResourceVersionMetadata` and returns `json.Unmarshal(metadata, objPtr)` from the SQLite metadata blob, never touching the payload file. Confirmed `ResourceVersionMetadata = "metadata"` (`pkg/apis/softwarecomposition/register.go:27`). So the retry-path `Get` really did hand back a zero-`Spec` object.
- The write-back really is destructive: `GuaranteedUpdate` (`storage.go:678-838`) runs `tryUpdate` and then `saveObject(conn, key, ret, ...)` — the client-supplied object is persisted wholesale. There is no server-side spec-merge that would have masked this. Zeroed `Spec` in → zeroed `Spec` on disk.
- The "safe siblings" claim holds for all four sites: `StoreCVE:339`, `StoreCVESummary:580`, `StoreSBOM:1224`, `StoreSBOM:1261` each assign `result.Spec = manifest.Spec` before `Update`, so the metadata-only read is genuinely harmless there. `StoreCVESummaryStub` was the only `Spec`-preserving method still using the metadata `Get` — the fix is complete, not a one-off patch that leaves siblings latent.
- Reachability is real, not theoretical: `core/services/scan.go:269` / `:483` call the stub on `ReasonImageSchemaUnsupported`, which is exactly the case where a workload already has a real summary from an earlier successful scan.

## Local results

- `go build ./...`, `gofmt -l repositories/`, `go vet ./repositories/` — all clean
- `go test ./repositories/... -count=1` — pass
- Reverted the one-line fix and re-ran: `TestAPIServerStore_StoreCVESummaryStub_DoesNotUseMetadataGet` fails with `Should be empty, but was metadata`. Genuinely red on `main`, green with the change — the regression guard works.
- All GitHub checks green.

The mechanism-test justification is sound: the fake clientset ignores `GetOptions` entirely, so a behavioral assertion on `Spec` preservation passes on both the buggy and the fixed code. An assertion on the requested `GetOptions` is the only thing that can catch this.

---

## Non-blocking notes

**1. The functional change is larger than "one token" — a dormant guard becomes live.** Because `summaryHasVulnerabilityData` never fired in production, every stub call previously wrote its status annotation (while wiping `Spec`). Now, when a real summary exists, the retry path returns early and writes *nothing* — including no `UnsupportedSchema` status marker. That is consistent with the stated intent at `apiserver.go:670` and with the existing assertion at `apiserver_test.go:756`, so I'm not treating it as a defect. But it is worth being explicit that a workload which scanned successfully and later hits an unsupported schema will now keep its stale-but-real summary with no status marker, which slightly cuts against the "so UIs do not show the workload as silently unscanned" rationale at `scan.go:478-480`. With this fix in place, merging annotations/labels while leaving `Spec` untouched would now be safe — if surfacing the status is desired, that's the follow-up. Fine to leave as-is for this PR.

**2. Optional test hardening.** The `ctxCapturingVulnerabilityManifestSummaries` wrapper could also *simulate* the backend — zero the returned `Spec` when `opts.ResourceVersion == "metadata"` — which would let a behavioral assertion ("real `Severities` survive the stub") catch the same regression without coupling the test to a specific option string. Slightly more robust against future refactors, but the current mechanism test is correct and demonstrably red-on-main, so this is purely optional.

**3. Re: the CodeRabbit `fake.NewSimpleClientset` (SA1019) nit — I'd skip it.** `apiserver_test.go` uses `fake.NewSimpleClientset` in 31 places and `fake.NewClientset` in zero. The new test matches the file's convention; migrating one call site would make the file inconsistent for no behavioral gain. If the deprecation matters, it should be a separate repo-wide change.

PR #479, opened 2026-08-06
fix: propagate K8s API list failures from GetSecurityExceptions instead of swallowing them

### LINKED ISSUE #477: bug: GetSecurityExceptions swallows K8s API errors, silently defeating the exceptions-cache's own cacheable=false safety guard for 5 minutes on every apiserver hiccup
# Description

`APIServerStore.GetSecurityExceptions` (`repositories/apiserver.go:113-155`) never returns a non-nil error, even when the underlying Kubernetes API calls fail. Its only caller that matters in production, `BackendAdapter.GetCVEExceptions` (`adapters/v1/backend.go:197-225`), was written with an explicit safety guard for exactly this failure mode — but because the guard's precondition (`crdErr != nil`) can never actually become true, the guard is dead code, and the exact outcome its own comment says it's trying to prevent happens anyway: a wrongly-empty CRD exception result gets cached for 5 minutes, silently un-suppressing previously-suppressed CVEs.

```go
// repositories/apiserver.go:113-155
func (a *APIServerStore) GetSecurityExceptions(ctx context.Context, namespace string) ([]sev1beta1.SecurityException, []sev1beta1.ClusterSecurityException, error) {
	listCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), 30*time.Second)
	defer cancel()

	var exceptions []sev1beta1.SecurityException
	if namespace != "" {
		seList, err := a.DynamicClient.Resource(securityExceptionGVR).Namespace(namespace).List(listCtx, metav1.ListOptions{})
		if err != nil {
			logger.L().Ctx(ctx).Warning("failed to list SecurityExceptions", helpers.Error(err), helpers.String("namespace", namespace))
			// error logged and dropped — falls through with exceptions == nil
		} else {
			// ... append to exceptions
		}
	}

	var clusterExceptions []sev1beta1.ClusterSecurityException
	cseList, err := a.DynamicClient.Resource(clusterSecurityExceptionGVR).List(listCtx, metav1.ListOptions{})
	if err != nil {
		logger.L().Ctx(ctx).Warning("failed to list ClusterSecurityExceptions", helpers.Error(err))
		// same pattern — error dropped
	} else {
		// ... append to clusterExceptions
	}

	return exceptions, clusterExceptions, nil // always nil, even after either failure above
}
```

```go
// adapters/v1/backend.go:203-217
seList, cseList, crdErr := a.securityExceptionRepo.GetSecurityExceptions(ctx, namespace)
if crdErr != nil {
	logger.L().Ctx(ctx).Warning("failed to get CRD security exceptions", helpers.Error(crdErr))
	// This is a best-effort degradation that self-heals on the next scan;
	// caching it would suppress valid CRD exceptions for the full TTL.
	cacheable = false
} else if len(seList) > 0 || len(cseList) > 0 {
	target := BuildExceptionTarget(ctx, workload, seList, cseList, a.securityExceptionRepo)
	crdPolicies := ConvertToVulnerabilityExceptionPolicies(seList, cseList, target)
	// ...
}

if cacheable && a.exceptionsCache != nil {
	a.exceptionsCache.Set(cacheKey, domain.CVEExceptions(vulnExceptionList), exceptionsCacheTTL) // 5 minutes
}
```

The `crdErr != nil` branch's comment shows the author was fully aware of the risk and wrote the guard specifically to prevent it — the guard just never fires, because `GetSecurityExceptions`'s contract (an `error` return value that's part of the `ports.SecurityExceptionRepository` interface it implements) is violated by its own implementation.

# Environment

OS: N/A — static correctness bug, reproducible on any OS
Version: `main` (`repositories/apiserver.go`, `adapters/v1/backend.go`)

# Steps To Reproduce

1. Create a `SecurityException` or `ClusterSecurityException` that suppresses `CVE-X` for a given workload/image.
2. Scan that workload once so `CVE-X` is correctly suppressed in the report.
3. Simulate a transient Kubernetes API failure during the *next* scan's `GetSecurityExceptions` call — e.g. briefly revoke RBAC on `securityexceptions`/`clustersecurityexceptions` for kubevuln's service account, or trigger apiserver throttling/a brief network blip during a rolling apiserver-store restart, right as `a.DynamicClient.Resource(...).List(...)` is in flight.
4. Observe: `GetSecurityExceptions` logs `"failed to list SecurityExceptions"` (or the cluster-scoped equivalent) but still returns `(nil, nil, nil)`.
5. `GetCVEExceptions` sees `crdErr == nil` and `len(seList) == 0 && len(cseList) == 0`, so it skips the CRD merge entirely and — critically — leaves `cacheable == true`.
6. The now-incomplete exception list (missing the CRD-based suppression of `CVE-X`) is written into `exceptionsCache` for `exceptionsCacheTTL` = 5 minutes (`adapters/v1/backend.go:58`), keyed by `accountID/cluster/namespace/kind/name/containerName/imageTag`.
7. Any scan of that same workload+image within the next 5 minutes — even after the apiserver has fully recovered — reuses the cached, incomplete result. `CVE-X` reappears as an active finding.

# Expected behavior

A Kubernetes API failure while listing `SecurityException`/`ClusterSecurityException` CRDs should propagate as an error from `GetSecurityExceptions`, so `GetCVEExceptions` takes the `crdErr != nil` branch, sets `cacheable = false`, and the degraded result is never cached — matching the behavior the code's own comment describes and the behavior `TestBackendAdapter_GetCVEExceptions_DoesNotCacheWhenCRDLookupFails` already asserts (see below).

# Actual Behavior

`GetSecurityExceptions` always returns a `nil` error regardless of whether either underlying `List()` call failed, so the `cacheable = false` guard in `GetCVEExceptions` is unreachable in production. A transient CRD-listing failure is indistinguishable from "this workload genuinely has zero exceptions," and that wrong result gets cached for the full 5-minute TTL — silently un-suppressing previously-suppressed CVEs for every scan of the same workload+image during that window.

# Additional context

- This is why the bug wasn't caught: `TestBackendAdapter_GetCVEExceptions_DoesNotCacheWhenCRDLookupFails` (`adapters/v1/backend_test.go:223-241`) does correctly assert the `cacheable = false` behavior — but it does so against a **fake** `testSecurityExceptionRepo` that returns `errors.New("boom")` directly at the interface level:
  ```go
  a := NewBackendAdapter("account", "apiServer", "eventReceiver", "", &testSecurityExceptionRepo{
      getSecurityExceptions: func(context.Context, string) ([]sev1beta1.SecurityException, []sev1beta1.ClusterSecurityException, error) {
          return nil, nil, errors.New("boom")
      },
  })
  ```
  It never exercises the real `APIServerStore.GetSecurityExceptions`, which is the only implementation that actually talks to the Kubernetes API and is the one with the bug. There is no test in `repositories/apiserver_test.go` for `GetSecurityExceptions` at all — searched, none exists. So the consumer-side contract is well-tested in isolation, but the concrete implementation silently violates that contract, and nothing catches the mismatch.
- Second caller, lower severity: `adapters/mockplatform.go:47-49` also calls `GetSecurityExceptions` and correctly checks `err != nil` — but since `MockPlatform` has no caching layer, the same underlying defect (API failure treated as "zero exceptions") only affects a single scan's result there, not a 5-minute cached window across every subsequent scan. Worth fixing in the same PR since it shares the root cause, not worth a separate issue.
- `ports.SecurityExceptionRepository.GetSecurityExceptions` (`core/ports/repositories.go:35`) is the interface both implementations satisfy; `NoOpSecurityExceptionRepository.GetSecurityExceptions` (`repositories/noop_security_exception.go:18`) trivially returns `(nil, nil, nil)` by design (there's no CRD backing it) and is not part of this bug.
- Affected files: `repositories/apiserver.go` (root cause), `adapters/v1/backend.go` (where the dead guard lives and the cache write happens), `adapters/mockplatform.go` (secondary caller, same fix benefits it), `repositories/apiserver_test.go` (missing coverage).

# Proposed Solution

1. Change `GetSecurityExceptions` to actually propagate the `List()` failures instead of only logging them — e.g. collect both errors with `errors.Join` (Go 1.20+, already used elsewhere in this codebase's error handling style) and return that instead of a hard-coded `nil`. Partial results (e.g. cluster-scoped list succeeded but namespaced list failed) should still return whatever was successfully collected alongside the non-nil error, since `GetCVEExceptions` already treats a non-nil `crdErr` as "don't trust this result, don't cache it" regardless of what's in the slices.
2. Add a test in `repositories/apiserver_test.go` for `GetSecurityExceptions` itself — using a fake/erroring dynamic client — asserting that a `List()` failure surfaces as a non-nil returned error. This is the coverage gap that let the bug ship.
3. Extend `TestBackendAdapter_GetCVEExceptions_DoesNotCacheWhenCRDLookupFails` (or add a sibling test) that goes through the real `APIServerStore` (with a failing fake dynamic client) rather than only the `testSecurityExceptionRepo` fake, so the consumer-side test actually exercises the real implementation's error contract end-to-end.
4. No change needed to `GetCVEExceptions` itself — its `crdErr != nil` handling is already correct; it just needs `crdErr` to ever actually be non-nil.

bhuvan-somisetty:
## Overview

`APIServerStore.GetSecurityExceptions` (`repositories/apiserver.go`) logged a Kubernetes `List()` failure on either `SecurityExceptions` or `ClusterSecurityExceptions` but always returned a `nil` error. Its production caller, `BackendAdapter.GetCVEExceptions` (`adapters/v1/backend.go`), has an explicit guard written specifically for this failure mode — but since `crdErr` could never actually be non-nil, the guard was dead code, and a transient apiserver hiccup got silently cached as "this workload has zero exceptions" for the full 5-minute TTL.

## What changed

1. **`GetSecurityExceptions` now propagates `List()` failures instead of swallowing them.** Both the namespaced and cluster-scoped list calls collect their errors via `stderrors.Join` (stdlib `errors` aliased to `stderrors` since this file already imports `k8s.io/apimachinery/pkg/api/errors` unaliased as `errors`) instead of only logging and moving on. Individual item-conversion failures are still just logged and skipped, since a single malformed object isn't the same class of problem as a listing-wide API failure.
2. **This makes the existing `cacheable = false` guard in `GetCVEExceptions` actually fire.** No change was needed there — its handling of a non-nil `crdErr` was already correct, it just never had a real chance to run.
3. **`NewFakeAPIServerStorage` now registers list-kinds for the SecurityException GVRs.** The fake dynamic client (`k8s.io/client-go/dynamic/fake`) hard-panics on any `List()` call if the GVR's list-kind isn't registered up front — that's the reason `GetSecurityExceptions` had zero test coverage before this PR, and the fix is a prerequisite for the new tests below to even run.

## Tests

- `TestAPIServerStore_GetSecurityExceptions_PropagatesListErrors` (new, `repositories/apiserver_test.go`): table test with a fake dynamic client + `PrependReactor`, covering a namespaced-list failure, a cluster-scoped-list failure, and a cluster-scoped-list failure when no namespace is given. Asserts the failure surfaces via `errors.Is`.
- `TestAPIServerStore_GetSecurityExceptions_NoErrorOnSuccess` (new): guards the other direction — joining a nil error slice must still return `nil`, not a non-nil "empty" error, so successful calls behave exactly as before.
- `TestBackendAdapter_GetCVEExceptions_DoesNotCacheWhenRealCRDListFails` (new, `adapters/v1/backend_test.go`): the important one. The existing `TestBackendAdapter_GetCVEExceptions_DoesNotCacheWhenCRDLookupFails` only proves `GetCVEExceptions` handles a non-nil error correctly — it injects that error directly at the interface level via a fake `securityExceptionRepo`, so it never actually exercised the real `APIServerStore.GetSecurityExceptions`, which is where the bug lived. This new test wires a real `APIServerStore` backed by a failing fake dynamic client end-to-end, so it would have caught this exact bug.

## How to Test

```bash
go build ./...
go vet ./repositories/... ./adapters/...
go test ./repositories/... -count=1
go test ./adapters/v1/... -run "TestBackendAdapter|TestGetCVEExceptions|TestApplySecurityExceptions|TestAPIServerStore" -count=1
```

Not touched by this PR: `Test_grypeAdapter_DBVersion` in `adapters/v1` fails locally on Windows with a `testcontainers`/Docker panic (`rootless Docker is not supported on Windows`) — pre-existing and unrelated, tracked separately in #471.

## Related issues/PRs:

* Resolved #477

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of security-exception retrieval failures.
  * Errors from namespaced and cluster-wide lookups are now reported instead of silently ignored.
  * Preserved successfully retrieved exceptions when part of a lookup fails.
  * Prevented failed or incomplete CVE exception lookups from being cached, allowing later requests to retry.

* **Tests**
  * Added regression coverage for failed, partially successful, and successful security-exception lookups.


bhuvan-somisetty:
@matthyx PTAL whenever u get chance.

matthyx (dismissed):
Approving — the diagnosis is right, the fix is minimal and correct, and the test that matters (`TestBackendAdapter_GetCVEExceptions_DoesNotCacheWhenRealCRDListFails`) genuinely closes the gap that let this bug live behind a green suite. Verified locally on Linux at `860569d`: `go build ./...`, `go vet ./repositories/... ./adapters/...`, and `go test ./adapters/... ./cmd/... ./core/... ./repositories/... -count=1` all pass (including `Test_grypeAdapter_DBVersion`, which only fails on your Windows box — #471).

No blockers. A few things worth your attention, one of which I'd like resolved before or right after merge.

---

### 1. Heads-up: this collides with #470

#470 (`fix(core): reconcile cached manifests with the current SecurityException set`, @AdeshDeshmukh) makes **the same change to the same lines** of `GetSecurityExceptions` — also `stderrors "errors"` + `stderrors.Join` on both list failures — and then builds `domain.ErrExceptionsDegraded` on top of it. Whichever lands second will conflict.

Suggested order: merge this one first (it's the smaller, focused fix and it's the one carrying the repository-level tests), then #470 rebases and drops its duplicate hunk in `repositories/apiserver.go`. @AdeshDeshmukh FYI.

### 2. Partial results are dropped by the caller — the PR description says otherwise

`GetCVEExceptions` merges CRD exceptions in an `else if`:

```go
if crdErr != nil {
    cacheable = false
} else if len(seList) > 0 || len(cseList) > 0 {
    // merge
}
```

So once `crdErr` is non-nil, the successfully-listed side is discarded too. Concretely: the namespaced `SecurityException` list fails (RBAC blip in one namespace) but the cluster-scoped list succeeds — before this PR the `ClusterSecurityException`s were still applied; after it, they are not. CVEs suppressed cluster-wide become visible again for the duration of the outage.

That's still a net improvement over the old behaviour (5-minute cached wrong answer → one wrong scan that self-heals), so I'm not blocking on it. But it's a one-line restructure to get both properties:

```go
if crdErr != nil {
    logger.L().Ctx(ctx).Warning("failed to get CRD security exceptions", helpers.Error(crdErr))
    cacheable = false
}
if len(seList) > 0 || len(cseList) > 0 {
    // merge whatever did come back
}
```

Worth noting the CodeRabbit pre-merge check and the PR body both claim the change "preserves partial results" — end to end, it doesn't. Either take the restructure above, or drop that claim and say the fail-closed-for-one-scan behaviour is deliberate. `adapters/mockplatform.go:48-52` has the identical shape and would want the same treatment. Fine as a follow-up if you'd rather keep this PR surgical, since #470 also touches this exact block.

### 3. `backend_test.go` copies the two GVRs across the package boundary

The comment is honest about why, but the copies will silently drift if `repositories` ever changes the group/version, and the test would keep passing against the wrong GVR. Since this PR *also* fixes `NewFakeAPIServerStorage` to register those list kinds, you can just use it and delete the 17-line `var` block plus three imports:

```go
store := repositories.NewFakeAPIServerStorage("kubescape")
store.DynamicClient.(*fakedynamic.FakeDynamicClient).PrependReactor("list", "securityexceptions",
    func(k8stesting.Action) (bool, runtime.Object, error) {
        return true, nil, fmt.Errorf("etcd timeout")
    })
```

I ran this against your branch — it compiles and produces the same failure path (`failed to list SecurityExceptions: etcd timeout`). Bonus: it makes the `NewFakeAPIServerStorage` registration fix load-bearing in a second package, so a future regression there breaks a test instead of going unnoticed. Nit, take it or leave it.

### 4. Every list failure now logs twice

`GetSecurityExceptions` still logs `Warning("failed to list SecurityExceptions", ...)` *and* returns the error, which both `GetCVEExceptions` and `MockPlatform.GetCVEExceptions` log again. Now that the error propagates, the repository-level warnings are redundant — the caller has the full `errors.Join` chain with the wrapped context. Dropping them would keep the signal clean during exactly the apiserver-hiccup storms this PR is about. Minor.

---

Points 2–4 are all optional. The core fix — propagate the error, register the fake's list kinds so the method is testable at all, and prove the guard fires against the real `APIServerStore` — is right, and the commit message explaining *why* the guard was dead code is the kind of thing I wish more fixes carried. Good catch on #477.

matthyx:
Update: #470 landed first (`9abb561`), so this is now `CONFLICTING`. My approval stands — but the rebase isn't purely mechanical, there's one real test failure hiding behind it. I ran the rebase locally so you don't have to discover it the hard way.

### What's left of this PR after #470

#470 shipped the same `stderrors.Join` propagation in `GetSecurityExceptions`, so the core fix is already on `main`. What is **still unique to this PR and still worth merging**:

1. **`NewFakeAPIServerStorage` list-kind registration** — `main` is still on `fakedynamic.NewSimpleDynamicClient(runtime.NewScheme())`, which panics on any `List()`. Not fixed by #470.
2. **All three tests** — #470 shipped zero coverage for the repository-level propagation it added. Right now `main` has the fix with nothing guarding it.
3. Wrapped error messages (`failed to list SecurityExceptions: %w` vs. the bare `err` on `main`) — small, but it makes the joined error readable in the log line the caller emits.

So please don't close this as superseded; rebase it.

### Resolving the conflicts

Four hunks in `repositories/apiserver.go`, all the same shape — `main`'s `var listErr error` / `listErr = stderrors.Join(listErr, err)` vs. your `var listErrs []error` / `append`. **Keep your side** for all four; it's equivalent plus the wrapped messages. Plus one trivial add-location conflict in `repositories/apiserver_test.go` (both sides appended new tests at the same spot — keep both).

### The part that isn't mechanical

After a clean rebase, `TestBackendAdapter_GetCVEExceptions_DoesNotCacheWhenRealCRDListFails` **fails**:

```
--- FAIL: TestBackendAdapter_GetCVEExceptions_DoesNotCacheWhenRealCRDListFails
    backend_test.go:293: Received unexpected error: exception set is incomplete
```

#470 changed `GetCVEExceptions` to return `domain.ErrExceptionsDegraded` on the degraded path, so your `assert.NoError(t, err)` is now wrong. Two-line fix, and it makes the test strictly stronger — it now also proves the degraded signal reaches the caller:

```go
_, err := a.GetCVEExceptions(ctx)
require.ErrorIs(t, err, domain.ErrExceptionsDegraded)
_, err = a.GetCVEExceptions(ctx)
require.ErrorIs(t, err, domain.ErrExceptionsDegraded)
assert.Equal(t, 2, calls, "a real CRD list failure must disable caching, not just a faked one")
```

(`domain` is already imported in `backend_test.go`.)

With those changes the effective diff against `main` is 3 files, +125/−6, and `go build ./...`, `go vet ./repositories/... ./adapters/...` and `go test ./adapters/... ./repositories/... ./core/... ./cmd/... -count=1` are all green on my Linux box.

### Still open from my earlier review

Point 2 — `GetCVEExceptions` discarding the successfully-listed side via the `else if` — got *more* relevant, not less: #470 built its `degraded` flag on that same branch, so a namespaced-list blip now both drops the cluster-scoped exceptions and marks the whole set degraded. Still a follow-up, not this PR's job. Points 3 and 4 are unchanged and still optional.

Happy to push the rebase to your branch myself if you'd rather not deal with it (`maintainer_can_modify` is on) — say the word.

bhuvan-somisetty:
Rebased on main and resolved merge conflicts in 
epositories/apiserver.go and 
epositories/apiserver_test.go.

- Updated TestBackendAdapter_GetCVEExceptions_DoesNotCacheWhenRealCRDListFails to use NewFakeAPIServerStorage and assert domain.ErrExceptionsDegraded.
- Restructured GetCVEExceptions and MockPlatform.GetCVEExceptions so any successfully fetched CRD policies are still appended during degraded execution paths.
- Simplified ackend_test.go setup by leveraging 
epositories.NewFakeAPIServerStorage.

matthyx (approved):
Re-approving `1116184`. My previous approval was auto-dismissed by the push, which is why the PR showed `BLOCKED` / `REVIEW_REQUIRED` — that clears now.

Checked the update against current `main` (`9582842`, which has moved on 3 commits since your merge base). I merged your head into it locally: no conflicts, `gofmt` clean on all five changed files, `go build ./...`, `go vet ./...`, and `go test ./... -count=1` all green — full suite this time, not just the touched packages.

**The `ErrExceptionsDegraded` assertion is right**, and taking point 2 was the right call. Worth spelling out why it's more than a nicety: `applyExceptionsToManifest` in `core/services/scan.go:702-728` already documents "a degraded fetch (`ErrExceptionsDegraded`) still applies the partial set but reports `complete=false`". Before this change that sentence couldn't actually be true — the `else if` meant a degraded fetch always yielded an *empty* set, so there was never a partial set to apply. Your restructure is what makes #470's own documented contract hold. Doing it in `mockplatform.go` too keeps the two platform implementations honest with each other.

One cleanup left, non-blocking:

**`adapters/v1/backend_test.go:43-58` is now dead.** You switched the test body to `repositories.NewFakeAPIServerStorage` (👍 — that's exactly what I meant), but the duplicated `securityExceptionGVR` / `clusterSecurityExceptionGVR` block stayed behind. Nothing references it anymore — Go doesn't flag unused package-level vars, so it compiles and CI stays green while the duplication I flagged is still sitting there, along with a doc comment that now describes something the file no longer does. Delete lines 43-58 and the `"k8s.io/apimachinery/pkg/runtime/schema"` import on line 38 (`schema` is used nowhere else in the file — `runtime`, `fakedynamic` and `k8stesting` are all still needed).

Two smaller notes, neither worth a round trip:

- The branch has a merge commit (`Merge branch 'upstream/main'`), not a rebase. The repo allows squash-merge so it'll disappear on the way in — just flagging it since the description says "rebased".
- `MockPlatform.GetCVEExceptions` now returns a nil `domain.CVEExceptions` on the empty path where it used to return `domain.CVEExceptions{}`. Every consumer goes through `len()`, so it's inert — noting it only so it isn't a surprise later.

Squash and merge once the dead block is gone (or as-is if you'd rather sweep it separately — I won't hold the PR for it).

PR #480, opened 2026-08-06
fix(adapters): key relevancy merge by (cve, package)

### LINKED ISSUE #478: bug : Relevancy merge keys by CVE id only, marking IsRelevant on packages that were never executed
## Description

When the same CVE affects multiple packages in a container image, the relevancy merge in the backend adapter marks the vulnerability relevant on every one of those packages, even when the relevancy (CVEp) scan only found it on a single executed package.

The relevancy index is keyed by CVE id only, dropping the package dimension that the data model treats as the record identity:

```go
cvepIndices := map[string]struct{}{}
for _, v := range relevantVulnerabilities {
	cvepIndices[v.Name] = struct{}{}      // keyed by CVE id only
}
for i, v := range vulnerabilities {
	_, isRelevant := cvepIndices[v.Name]
	vulnerabilities[i].IsRelevant = &isRelevant
}
```

(`adapters/v1/backend.go:411-418`)

`(CVE, package)` is the record identity everywhere else: `DomainToArmo` emits `RelatedPackageName` (`adapters/v1/domain_to_armo.go:159`), and the uniqueness assertion in `adapters/v1/backend_test.go:362` keys by `Name + "+" + RelatedPackageName`.

---

## Environment

- **OS:** N/A — pure report-generation logic
- **Version:** `main` (`kubevuln`, `SubmitCVE` with non-nil `cvep.Content`)

---

## Steps To Reproduce

A common real-world case (e.g., a `libc` CVE affecting `libc6`, `libc-bin`, and `multiarch-support`, where only `libc6` is executed):

1. Full scan finds `CVE-X` on packages A and B.
2. Relevancy scan — produced from the executed-files-filtered SBOM (`filterSBOM`, `core/services/scan.go:842`) — finds `CVE-X` on package A only.
3. Call `SubmitCVE(ctx, cve, cvep)` with the full scan as `cve` and the filtered scan as `cvep`.
4. Inspect the sent report: both `(CVE-X, A)` and `(CVE-X, B)` are marked `IsRelevant = true`.

Multiple-package CVEs already exist in the test fixture `adapters/v1/testdata/nginx-cve.json` (e.g., `CVE-2007-5686`, `CVE-2009-5155`, `CVE-2016-10228`), so a reproduction is a copy of that fixture with the CVE removed from the filtered fixture for one package.

---

## Expected Behavior

`(CVE-X, A)` → `IsRelevant = true`; `(CVE-X, B)` → `IsRelevant = false`, because package B was never executed. `RelevantCount` counts the finding once, not once per package.

---

## Actual Behavior

Both `(CVE-X, A)` and `(CVE-X, B)` are marked relevant: the CVE is attributed to packages that were never executed, and `RelevantCount` / `RelevancyScanCount` are inflated.

---

## Additional Context

- **Impact (scoped):** Report data accuracy — wrong package attribution and inflated relevant counts. Not a security exposure; the CVE is still genuinely relevant via the executed package. The defect is which package is reported as affected.
- **Why this was missed:** The integration fixtures `testdata/nginx-cve.json` and `testdata/nginx-filtered-cve.json` are content-identical at the `(CVE, package)` level, so the golden test cannot observe a subset divergence. A proper subset is the normal case in production, not an exception.

---

## Proposed Solution

Key the index by `Name + "+" + RelatedPackageName` on both build and lookup:

```go
cvepIndices := map[string]struct{}{}
for _, v := range relevantVulnerabilities {
	cvepIndices[v.Name+"+"+v.RelatedPackageName] = struct{}{}
}
for i, v := range vulnerabilities {
	_, isRelevant := cvepIndices[v.Name+"+"+v.RelatedPackageName]
	vulnerabilities[i].IsRelevant = &isRelevant
}
```

This is lossless — CVE/GHSA/EUVD IDs never contain `+`, so the separator is unambiguous even when a package name does (e.g., `libstdc++6`) — and it matches the existing uniqueness key in `backend_test.go:362`.

AdeshDeshmukh:
Closes #478

## Overview

Fixed the relevancy merge in `SubmitCVE` that marked a CVE as relevant on every package it affects, even when the relevancy (CVEp) scan only found it on a single executed package.

The merge was keyed by CVE ID only (`cvepIndices[v.Name]`), dropping `RelatedPackageName` — even though `(CVE, package)` is the record identity: `DomainToArmo` emits `RelatedPackageName`, and the uniqueness assertion in `backend_test.go` keys by `Name + "+" + RelatedPackageName`.

The merge is now extracted into `markRelevantVulnerabilities`, keyed by the composite `Name + "+" + RelatedPackageName` on both build and lookup. The key is lossless: CVE/GHSA/EUVD IDs never contain `+`, so the separator is unambiguous even when a package name does (e.g., `libstdc++6`).

**Impact (scoped):** Report data accuracy — wrong package attribution and inflated `RelevantCount`/`RelevancyScanCount`. Not a security exposure; the CVE is still genuinely relevant via the executed package.

---

## Additional Information

- **Why this went unnoticed:** The integration fixtures (`testdata/nginx-cve.json` vs `testdata/nginx-filtered-cve.json`) are content-identical at the `(CVE, package)` level, so the golden test cannot observe a subset divergence. A proper subset is the normal case in production — the CVEp scan is produced from the executed-files-filtered SBOM (`filterSBOM`, `core/services/scan.go:842`).
- **Testing approach:** A new end-to-end test (`TestBackendAdapter_SubmitCVE_RelevancySubset`) reproduces the exact scenario through `SubmitCVE` — `CVE-2024-0001` on `libssl1.1` + `openssl`, with CVEp containing only `libssl1.1`. It failed before the fix (`openssl` marked relevant) and passes after. A focused unit test (`TestMarkRelevantVulnerabilities`) covers the merge logic (same-CVE multi-package, disjoint, empty CVEp, mixed).

---

## How to Test

```bash
go test ./adapters/v1/ -run 'TestMarkRelevantVulnerabilities|TestBackendAdapter_SubmitCVE' -v
go build ./...
```

Full suite: 438 passed, 2 failed — both pre-existing Docker/local-registry environment failures (`Test_grypeAdapter_DBVersion`, `TestCreateSBOM_ImageTooLarge_LocalRegistry`), verified to fail on clean `main` as well. `golangci-lint` reports 18 issues, identical to the clean baseline (0 new).

---

## Examples/Screenshots

N/A

---

## Related Issues/PRs

- Resolved #478

---

## Checklist Before Requesting a Review

### DIFF
--- a/adapters/v1/backend.go
+++ b/adapters/v1/backend.go
@@ -407,16 +407,8 @@ func (a *BackendAdapter) SubmitCVE(ctx context.Context, cve domain.CVEManifest, 
 		if err != nil {
 			return fmt.Errorf("failed to convert filtered vulnerabilities to report: %w", err)
 		}
-		// index relevantVulnerabilities
-		cvepIndices := map[string]struct{}{}
-		for _, v := range relevantVulnerabilities {
-			cvepIndices[v.Name] = struct{}{}
-		}
 		// mark common vulnerabilities as relevant
-		for i, v := range vulnerabilities {
-			_, isRelevant := cvepIndices[v.Name]
-			vulnerabilities[i].IsRelevant = &isRelevant
-		}
+		markRelevantVulnerabilities(vulnerabilities, relevantVulnerabilities)
 	}
 
 	finalReport := v1.ScanResultReport{
@@ -519,4 +511,20 @@ func (a *BackendAdapter) SubmitCVE(ctx context.Context, cve domain.CVEManifest, 
 func httpPostDebug(httpClient httputils.IHttpClient, fullURL string, headers map[string]string, body []byte) (*http.Response, error) {
 	logger.L().Debug("httpPostDebug", helpers.String("fullURL", fullURL), helpers.Interface("headers", headers), helpers.String("body", string(body)))
 	return httputils.HttpPostWithContext(context.Background(), httpClient, fullURL, headers, body, -1, func(resp *http.Response) bool { return true })
+}
+
+// markRelevantVulnerabilities annotates each vulnerability with IsRelevant=true iff the same
+// (CVE, package) pair also appeared in the relevancy (CVEp) scan. Keying by CVE id alone would
+// mark a CVE relevant on every package it affects even when only one of those packages was
+// executed; the pair is the record identity (see RelatedPackageName in DomainToArmo and the
+// uniqueness assertion in backend_test.go).
+func markRelevantVulnerabilities(vulnerabilities, relevantVulnerabilities []cs.CommonContainerVulnerabilityResult) {
+	cvepIndices := make(map[string]struct{}, len(relevantVulnerabilities))
+	for _, v := range relevantVulnerabilities {
+		cvepIndices[v.Name+"+"+v.RelatedPackageName] = struct{}{}
+	}
+	for i, v := range vulnerabilities {
+		_, isRelevant := cvepIndices[v.Name+"+"+v.RelatedPackageName]
+		vulnerabilities[i].IsRelevant = &isRelevant
+	}
 }
--- a/adapters/v1/backend_test.go
+++ b/adapters/v1/backend_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	_ "embed"
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Reviewed the change and verified it locally. **No blockers — LGTM.**

## What I verified

- **The key match is sound, not just plausible.** The argument rests on `RelatedPackageName` being identical between the full and the CVEp scan. It is: `filterSBOM` (`core/services/scan.go`) copies the surviving artifacts verbatim (`filteredSBOM.Content.Artifacts = append(filteredSBOM.Content.Artifacts, artifact)`), so `m.Artifact.Name` — the only input to `RelatedPackageName` in `DomainToArmo` — is byte-identical for a package present in both scans. Filtering can only drop packages, never rename them, so the composite key can't produce false negatives.
- **The tests are genuine regression tests.** I reverted `markRelevantVulnerabilities` to the old CVE-id-only keying and re-ran: `TestBackendAdapter_SubmitCVE_RelevancySubset` fails on `RelevantCount` (expected 1, got 2) and `TestMarkRelevantVulnerabilities` fails on the `openssl` / `p2` cases. Restored and re-ran — all pass, including the pre-existing `TestBackendAdapter_SubmitCVE` golden tests.
- **The separator argument holds** for the current key, and `markRelevantVulnerabilities` is the only remaining relevancy merge site — grep found no other CVE-id-only keyed index.
- **`build-and-push-image` failed for infra reasons**, not code: `The job was not acquired by Runner of type hosted even after multiple attempts`, job concluded `cancelled` after 2h26m. Needs a re-run, not a fix.

## Non-blocking notes

1. **Prefer a struct key over string concatenation.** `map[struct{ name, pkg string }]struct{}` (or `[2]string`) removes the "CVE IDs never contain `+`" argument entirely rather than documenting it, and avoids building a concatenated string per entry and per lookup — this runs over every match in the full scan, which is thousands of entries on a large image.

2. **`(CVE, package name)` is still not unique in multi-version ecosystems.** Nested npm deps and shaded/duplicated JARs routinely put the same package name at two versions in one image, producing two grype matches. If only one is executed, both still get `IsRelevant = true` — the same class of over-marking, one level finer. This matches the repo's existing identity convention (`backend_test.go`'s uniqueness assertion keys the same way), so it's out of scope here, but `PackageVersion` in the key would be worth a follow-up.

3. **Worth a release note.** `RelevantCount` / `RelevancyScanCount` will drop for existing customers on images with multi-package CVEs. That's the point of the fix, but anything keyed on those numbers (dashboards, alert thresholds, trend comparisons) will show a step change at rollout rather than a gradual one.

Approving.

PR #487, opened 2026-08-07
fix: allow retries on 500 Internal Server Error in shouldRetryReport

### LINKED ISSUE #486: Bug: shouldRetryReport does not retry on transient 500 Internal Server Error
# Description

In `adapters/v1/backend.go`, the `shouldRetryReport` function explicitly prevents the adapter from retrying when the backend platform returns a `500 Internal Server Error`.

A `500` status code is commonly caused by transient failures, such as temporary API Gateway issues, backend pod restarts, or unexpected server-side errors. Treating it as a non-retriable error means that a single temporary backend failure can permanently drop a scan result.

The function's inline comment states that it is intended to **"retry on any response except the ones that are never going to succeed."** Since a `500 Internal Server Error` is generally considered retriable, excluding it from retries contradicts the documented behavior and appears to be a bug.

---

# Environment

- **OS:** macOS 26.5.2 (Build 25F84)
- **Version:** `fb16373` (Git commit hash)

---

# Steps To Reproduce

1. Create a Go unit test file at `adapters/v1/should_retry_test.go`:

```go
package v1

import (
	"net/http"
	"testing"
)

func TestShouldRetryReport_BugReproducer(t *testing.T) {
	resp := &http.Response{
		StatusCode: http.StatusInternalServerError,
	}

	shouldRetry := shouldRetryReport(resp)

	if !shouldRetry {
		t.Fatalf("BUG REPRODUCED: shouldRetryReport returned false for 500 Internal Server Error, meaning it will NOT retry on transient 500 errors.")
	}
}
```

2. Navigate to the `kubevuln` repository root.

3. Run the test:

```bash
go test -v -run TestShouldRetryReport_BugReproducer ./adapters/v1
```

4. Observe the test failure.

---

# Expected Behavior

`shouldRetryReport` should return `true` for a `500 Internal Server Error` so that the HTTP client can perform its configured exponential backoff and retry the request, as `500` responses are commonly transient.

---

# Actual Behaviour / Test logs

The function currently returns `false` for `500 Internal Server Error`, preventing the retry mechanism from being triggered.

**Test output:**

```text
=== RUN   TestShouldRetryReport_BugReproducer
    should_retry_test.go:18: BUG REPRODUCED: shouldRetryReport returned false for 500 Internal Server Error, meaning it will NOT retry on transient 500 errors.
--- FAIL: TestShouldRetryReport_BugReproducer (0.00s)
FAIL
FAIL    github.com/kubescape/kubevuln/adapters/v1    1.018s
FAIL
```

---

# Additional Context

The exclusion list in `shouldRetryReport` currently treats `http.StatusInternalServerError` (`500`) as a non-retriable status code.

Removing `http.StatusInternalServerError` from the exclusion list (around line 132 of `adapters/v1/backend.go`) restores the intended retry behavior, allowing transient backend failures to be retried while preserving the existing behavior for status codes that are genuinely non-retriable.

doraem-on:
## Overview

**Current behavior:** The `shouldRetryReport` function prevents the adapter from retrying when the backend returns a `500 Internal Server Error`, treating it as a fatal error even though `500` responses are often transient failures.

**New behavior:** `500 Internal Server Error` responses are now eligible for the existing HTTP retry mechanism with exponential backoff, aligning with the intent of `defaultShouldRetry` to retry responses that may succeed on subsequent attempts.

---

## Additional Information

A new unit test, `TestShouldRetryReport_RetriesOnInternalServerError`, has been added to verify this behavior and prevent future regressions.

---

## How to Test

1. Run the new unit test from the repository root:

   ```bash
   go test -v -run TestShouldRetryReport_RetriesOnInternalServerError ./adapters/v1
   ```

2. Verify that the test passes, confirming that `shouldRetryReport` now returns `true` for HTTP `500 Internal Server Error`.

---

## Output Logs

```text
=== RUN   TestShouldRetryReport_RetriesOnInternalServerError
--- PASS: TestShouldRetryReport_RetriesOnInternalServerError (0.00s)
PASS
ok      github.com/kubescape/kubevuln/adapters/v1    1.352s
```

---

## Related Issues

- Closes #486

---

## Checklist

- [x] I have added or updated tests where appropriate.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability by retrying requests that receive an HTTP 500 Internal Server Error response.

### DIFF
--- a/adapters/v1/backend.go
+++ b/adapters/v1/backend.go
@@ -124,13 +124,15 @@ func httpPostWithContext(ctx context.Context, httpClient httputils.IHttpClient, 
 	}, backoff.WithBackOff(bo), backoff.WithMaxElapsedTime(maxElapsedTime))
 }
 
-// shouldRetryReport mirrors the unexported defaultShouldRetry in armosec/utils-go: retry on any
-// response except the ones that are never going to succeed on a retry.
+// shouldRetryReport is derived from the unexported defaultShouldRetry in armosec/utils-go, but
+// intentionally diverges from it: upstream treats 500 as fatal, we retry it. A 500 from the event
+// receiver is usually transient (gateway blip, backend pod restart), so dropping the scan result
+// on the first one loses data that a retry would have delivered (#486). The retry budget is
+// bounded by maxElapsedTime, so a genuinely broken backend still fails fast enough.
 func shouldRetryReport(resp *http.Response) bool {
 	return resp.StatusCode != http.StatusUnauthorized &&
 		resp.StatusCode != http.StatusForbidden &&
-		resp.StatusCode != http.StatusNotFound &&
-		resp.StatusCode != http.StatusInternalServerError
+		resp.StatusCode != http.StatusNotFound
 }
 
 const ActionName = "vuln scan"
--- a/adapters/v1/backend_test.go
+++ b/adapters/v1/backend_test.go
@@ -1000,3 +1000,23 @@ func TestBackendAdapter_SubmitCVE_SkipsChunksWhenSummaryFails(t *testing.T) {
 	require.Error(t, err, "SubmitCVE should surface the summary post failure")
 	assert.Equal(t, int32(0), atomic.LoadInt32(&chunkPosts), "no vulnerability chunk should be posted once the summary failed")
 }
+
+func TestShouldRetryReport(t *testing.T) {
+	for _, tc := range []struct {
+		status int
+		want   bool
+	}{
+		{http.StatusUnauthorized, false},
+		{http.StatusForbidden, false},
+		{http.StatusNotFound, false},
+		{http.StatusInternalServerError, true}, // #486
+		{http.StatusTooManyRequests, true},
+		{http.StatusBadGateway, true},
+		{http.StatusServiceUnavailable, true},
+	} {
+		t.Run(http.StatusText(tc.status), func(t *testing.T) {
+			got := shouldRetryReport(&http.Response{StatusCode: tc.status})
+			assert.Equal(t, tc.want, got)
+		})
+	}
+}


matthyx (commented):
The behavior change itself is correct and I agree with the premise of #486: a `500` from the event receiver (gateway blip, pod restart, transient server-side error) is exactly the class of failure the exponential backoff exists for, and the retry budget is bounded (`60s` for vulnerability reports in `backend_utils.go:130`, `30s` for scan-failure reports in `backend.go:311`), so this can't turn into an unbounded stall.

I verified locally on `02d71e3`: `go build ./...`, `go vet ./adapters/...`, and `go test ./adapters/v1/... ./core/...` all pass.

One must-fix before merge, plus two nits — details inline.

<hr>

**Non-blocking, for maintainer awareness:** retrying a `500` means a report the receiver may have *partially* ingested before erroring can be posted again. Reports carry `ContainerScanID` and `PaginationInfo.ReportNumber`, so this is safe as long as the receiver upserts by `(scanID, reportNumber)`. If it appends, retries could duplicate vulnerability rows. Worth a sanity check on the receiver side, but it isn't a reason to hold this PR — the same exposure already exists for every other retryable status (`429`, `502`, `503`, …).

The branch is based on `fb16373` and is a few commits behind `main`; GitHub still reports it mergeable, so this is purely informational.

adapters/v1/backend.go
```diff
@@ -129,8 +129,7 @@ func httpPostWithContext(ctx context.Context, httpClient httputils.IHttpClient,
 func shouldRetryReport(resp *http.Response) bool {
```
matthyx, line 132: **Blocker (must fix before merge): the doc comment two lines above is now false.**

Lines 127-128 claim this function *"mirrors the unexported `defaultShouldRetry` in armosec/utils-go"*. I checked the pinned version (`armosec/utils-go v0.0.58`, per `go.mod:16`) and upstream still reads:

```go
func defaultShouldRetry(resp *http.Response) bool {
	// If received codes 401/403/404/500 should return false
		resp.StatusCode != http.StatusNotFound &&
		resp.StatusCode != http.StatusInternalServerError
}
```

So after this PR the two no longer match, and the comment is the only place the relationship is documented. That's a real hazard, not a style point: the next person who reads "mirrors upstream", diffs the two, and "restores parity" silently reverts this fix — with the comment vouching for them.

Please make the divergence explicit and intentional. Suggested replacement for lines 127-128:

```go
// shouldRetryReport is derived from the unexported defaultShouldRetry in armosec/utils-go, but
// intentionally diverges from it: upstream treats 500 as fatal, we retry it. A 500 from the event
// receiver is usually transient (gateway blip, backend pod restart), so dropping the scan result
// on the first one loses data that a retry would have delivered (#486). The retry budget is
// bounded by maxElapsedTime, so a genuinely broken backend still fails fast enough.
```

Adjust the wording as you like — the point is that the divergence should be stated rather than left to look like drift.

adapters/v1/should_retry_test.go
```diff
@@ -0,0 +1,16 @@
...
+	"net/http"
+	"testing"
+)
+
+func TestShouldRetryReport_RetriesOnInternalServerError(t *testing.T) {
+	resp := &http.Response{
+		StatusCode: http.StatusInternalServerError,
+	}
+
+	if !shouldRetryReport(resp) {
```
matthyx: Two nits on this file, neither blocking:

**1. The assertion is one-sided.** The regression this guards against isn't really "500 stops being retryable" — it's "the exclusion list drifts". A table-driven test pins the whole contract for about the same number of lines, and also catches the reverse regression (someone dropping `401`/`403`/`404`):

```go
func TestShouldRetryReport(t *testing.T) {
	for _, tc := range []struct {
		status int
		want   bool
	}{
		{http.StatusUnauthorized, false},
		{http.StatusForbidden, false},
		{http.StatusNotFound, false},
		{http.StatusInternalServerError, true}, // #486
		{http.StatusTooManyRequests, true},
		{http.StatusBadGateway, true},
		{http.StatusServiceUnavailable, true},
	} {
		t.Run(http.StatusText(tc.status), func(t *testing.T) {
			got := shouldRetryReport(&http.Response{StatusCode: tc.status})
			assert.Equal(t, tc.want, got)
		})
	}
}
```

**2. File placement.** `adapters/v1` keeps tests in a `_test.go` file named after the source file (`backend_test.go`, `backend_utils_test.go`, `sidecar_test.go`, …). Since `shouldRetryReport` lives in `backend.go`, this test would sit more naturally in `backend_test.go` than in a new `should_retry_test.go`.

(The bare `&http.Response{StatusCode: ...}` with a nil `Body` is fine here — `shouldRetryReport` only reads `StatusCode`.)

doraem-on:
@matthyx  please review once again , I have made the suggested changes , thanks.

matthyx (approved):
All three points from the previous review are addressed — thanks for the quick turnaround.

- The doc comment now states the divergence from `armosec/utils-go`'s `defaultShouldRetry` as intentional, with the reasoning and the `#486` reference. That was the blocker.
- The test is table-driven and pins the whole exclusion list (`401`/`403`/`404` → false, `500`/`429`/`502`/`503` → true), so it also guards the reverse regression.
- It now lives in `backend_test.go`, matching the package convention, and the standalone `should_retry_test.go` is gone.

Verified locally on `52abad6`: `go build ./...` and `go vet ./adapters/...` clean, `go test ./adapters/v1/... ./core/...` passes (including the 7 new `TestShouldRetryReport` subtests). All 6 PR checks are green and the branch is mergeable.

LGTM 🚀

PR #491, opened 2026-08-07
feat: sweep stale stereoscope temp dirs at startup

### LINKED ISSUE #483: Feature: sweep stale stereoscope-* temp dirs at startup
## Overview
kubevuln downloads container images through stereoscope, which creates a
`/tmp/stereoscope-<random>` temp dir per scan and removes it via
`defer t.Cleanup()` in two places:

- `adapters/v1/syft.go` (~line 180) — the in-process Syft adapter
- `pkg/sbomscanner/v1/server.go` (~line 189) — the SBOM-scanner sidecar

## Problem
When the process is killed by SIGKILL — OOM kill, or Kubernetes forced
termination after `terminationGracePeriodSeconds` expires (the scenario
addressed by #467/#468) — deferred cleanup never runs, and
`/tmp/stereoscope-*` directories are orphaned on disk.

They accumulate across restarts in two deployment patterns:

1. Container restart within a surviving pod (`restartPolicy: Always`):
   the pod's `emptyDir` volume survives container restarts, so orphaned
   dirs persist across each restart cycle.
2. PVC-backed `/tmp`: orphaned dirs persist across full pod recreations.

Accumulated directories consume disk until scans fail with "no space
left on device". #467/#468 fixed graceful queue draining on SIGTERM —
SIGKILL bypasses that drain entirely, and this gap remains.

## Solution
A best-effort startup sweep in `cmd/http/main.go`, before the server
accepts scan requests, calling a new helper:

```go
tools.CleanupStaleTempDirs(os.TempDir(), "stereoscope-", time.Hour)
```

The helper removes `/tmp` directories matching `stereoscope-*` whose
mtime is older than 1 hour. The age threshold prevents removing temp
dirs belonging to a live scan from the SBOM-scanner sidecar process —
a separate process that shares the same `/tmp`. One sweep in the main
process covers both in-process and sidecar orphans.

The sweep is non-fatal: on error it logs a warning and startup continues.

## Alternatives
- Periodic in-process timer: more moving parts; startup is the only
  point where no in-process scans are running.
- Disk-pressure-triggered sweep: more complex, hard to test.
- Sweep on graceful shutdown: doesn't help — SIGKILL is exactly when
  shutdown hooks don't run.
- Do nothing: in the default `emptyDir` + pod-recreation pattern the
  dirs vanish anyway; the gap only materializes on
  container-restart-in-pod or PVC-backed `/tmp`, but those are real
  configurations and the failure mode (disk exhaustion) is silent and
  hard to diagnose.

## Additional context
- Verified code path: `file.NewTempDirGenerator("stereoscope")` creates
  `os.MkdirTemp("", "stereoscope-")` (`matthyx/stereoscope` fork,
  `pkg/file/temp_dir_generator.go:24`). Both call sites rely on
  `defer t.Cleanup()` exclusively.
- Repro: `kill -9` the kubevuln process mid-scan, then
  `ls /tmp/stereoscope-*` — the directory survives; restarting the
  process reclaims nothing.
- No new dependencies, no config knobs, no changes to the sidecar
  binary. Scope: 3 files, ~60 lines.

AdeshDeshmukh:
Closes #483

## Overview
Stereoscope creates a `/tmp/stereoscope-*` temp dir per scan and removes it via
`defer t.Cleanup()` (`adapters/v1/syft.go`, `pkg/sbomscanner/v1/server.go`). SIGKILL
(OOM kill, or Kubernetes termination after `terminationGracePeriodSeconds` expires)
skips deferred cleanup, orphaning these dirs across container restarts in a surviving
pod (`emptyDir` survives) or on PVC-backed `/tmp`.

This PR adds a best-effort startup sweep in `cmd/http/main.go`, before the server
accepts scan requests, that deletes `/tmp` dirs matching `stereoscope-*` whose mtime
is older than `ScanTimeout`.

## Additional Information
- Threshold is `ScanTimeout` (default 5m): the main process and the SBOM-scanner
  sidecar run in the same Pod on the same machine (same kernel clock), so no
  clock-skew margin is needed; a dir idle longer than `ScanTimeout` can only belong
  to a dead or already-timed-out scan.
- One sweep covers both processes: the in-process Syft adapter and the sidecar write
  to the same mounted `/tmp`, so a single sweep in the main process reclaims orphans
  from both.
- Best-effort: on error it logs a warning and startup continues. No config knob, no
  new dependencies.
- Known tradeoff: on a main-only container restart, a surviving sidecar mid-scan (it
  keeps pulling on a detached context, #421) could have its temp dir swept — worst
  case is one failed scan, which is retried.
- Untested-by-design: mid-sweep removal failure aggregation (`errors.Join`) is not
  exercised — portable failure injection would require an fs abstraction.

## How to Test
```bash
go test ./internal/tools/ ./cmd/http/ -v
go build ./...
```

Full suite: 438 passed, 2 failed — both pre-existing Docker/local-registry
environment failures (`Test_syftAdapter_CreateSBOM`, `TestCreateSBOM_ImageTooLarge_LocalRegistry`),
verified to fail on clean `main` as well. No new golangci-lint issues.

## Examples/Screenshots
N/A

## Related issues/PRs
- Resolved #483

## Summary by CodeRabbit

* **Bug Fixes**
  * Automatically removes stale temporary scan directories during service startup.
  * Keeps recent and unrelated temporary directories intact.
  * Logs cleanup warnings when removal errors occur and reports successful cleanup activity.

### DIFF
--- a/cmd/http/main.go
+++ b/cmd/http/main.go
@@ -21,6 +21,7 @@ import (
 	"github.com/kubescape/kubevuln/core/ports"
 	"github.com/kubescape/kubevuln/core/services"
 	"github.com/kubescape/kubevuln/internal/metrics"
+	"github.com/kubescape/kubevuln/internal/tools"
 	sbomscanner "github.com/kubescape/kubevuln/pkg/sbomscanner/v1"
 	"github.com/kubescape/kubevuln/repositories"
 	"go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin"
@@ -129,6 +130,21 @@ func main() {
 	controller, err = controller.WithMetrics(m)
 	if err != nil {
 		logger.L().Ctx(ctx).Fatal("metrics registration error", helpers.Error(err))
+	}
+
+	// Best-effort startup sweep: reclaim stereoscope temp dirs orphaned by previous process
+	// invocations killed by SIGKILL (OOM kill, or terminationGracePeriodSeconds exceeded)
+	// before their defer-based cleanup could run. The threshold is ScanTimeout: the main
+	// process and the SBOM-scanner sidecar run in the same Pod on the same machine (same
+	// kernel clock), so no clock-skew margin is needed, and a dir idle longer than
+	// ScanTimeout can only belong to a dead or already-timed-out scan.
+	if removed, err := tools.CleanupStaleTempDirs(os.TempDir(), "stereoscope-", c.ScanTimeout); err != nil {
+		logger.L().Ctx(ctx).Warning("startup temp dir sweep error",
+			helpers.Error(err),
+			helpers.Int("removed", removed))
+	} else if removed > 0 {
+		logger.L().Ctx(ctx).Info("startup temp dir sweep complete",
+			helpers.Int("removed", removed))
 	}
 
 	gin.SetMode(gin.ReleaseMode)
--- a/internal/tools/tools.go
+++ b/internal/tools/tools.go
@@ -2,11 +2,13 @@ package tools
 
 import (
 	"encoding/json"
+	"errors"
 	"os"
 	"path"
 	"regexp"
 	"runtime/debug"
 	"strings"
+	"time"
 
 	helpersv1 "github.com/kubescape/k8s-interface/instanceidhandler/v1/helpers"
 
@@ -103,6 +105,42 @@ func DeleteContents(dir string) error {
 		}
 	}
 	return nil
+}
+
+// CleanupStaleTempDirs removes directories under dir whose names start with prefix and whose
+// mtime is older than olderThan. It is a best-effort startup sweep to reclaim disk space left
+// by processes killed before their defer-based cleanup could run (e.g. SIGKILL from OOM or
+// Kubernetes terminationGracePeriodSeconds expiry).
+//
+// It returns the number of directories removed and a joined error of all failures encountered.
+// A non-nil error never blocks startup — callers should log and continue.
+func CleanupStaleTempDirs(dir, prefix string, olderThan time.Duration) (int, error) {
+	entries, err := os.ReadDir(dir)
+	if err != nil {
+		return 0, err
+	}
+	cutoff := time.Now().Add(-olderThan)
+	removed := 0
+	var allErrs error
+	for _, e := range entries {
+		if !e.IsDir() || !strings.HasPrefix(e.Name(), prefix) {
+			continue
+		}
+		info, err := e.Info()
+		if err != nil {
+			allErrs = errors.Join(allErrs, err)
+			continue
+		}
+		if !info.ModTime().Before(cutoff) {
+			continue // young enough — may belong to a live sidecar scan
+		}
+		if err := os.RemoveAll(path.Join(dir, e.Name())); err != nil {
+			allErrs = errors.Join(allErrs, err)
+			continue
+		}
+		removed++
+	}
+	return removed, allErrs
 }
 
 func NormalizeReference(ref string) string {
--- a/internal/tools/tools_test.go
+++ b/internal/tools/tools_test.go
@@ -1,10 +1,14 @@
 package tools
 
 import (
+	"os"
+	"path"
 	"testing"
+	"time"
 
 	helpersv1 "github.com/kubescape/k8s-interface/instanceidhandler/v1/helpers"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestPackageVersion(t *testing.T) {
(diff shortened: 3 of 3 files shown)


cmd/http/main.go
```diff
@@ -131,6 +132,21 @@ func main() {
 		logger.L().Ctx(ctx).Fatal("metrics registration error", helpers.Error(err))
 	}
 
+	// Best-effort startup sweep: reclaim stereoscope temp dirs orphaned by previous process
+	// invocations killed by SIGKILL (OOM kill, or terminationGracePeriodSeconds exceeded)
+	// before their defer-based cleanup could run. The threshold is ScanTimeout: the main
+	// process and the SBOM-scanner sidecar run in the same Pod on the same machine (same
+	// kernel clock), so no clock-skew margin is needed, and a dir idle longer than
+	// ScanTimeout can only belong to a dead or already-timed-out scan.
+	if removed, err := tools.CleanupStaleTempDirs(os.TempDir(), "stereoscope-", c.ScanTimeout); err != nil {
```
matthyx, line 141: Non-blocking, but worth a guard: `c.ScanTimeout` is operator-tunable and `docs/CONFIGURATION.md` accepts any `^[0-9]+(s|m|h)$`, so a deployment running `scanTimeout: 30s` silently shrinks this sweep's safety window from 5m to 30s — which widens the "swept a live sidecar's dir" case the PR description calls out as a known tradeoff. #483 proposed a flat `time.Hour`.

```go
// Floor the sweep threshold: scanTimeout is operator-tunable and can be far
// below the default 5m, which would shrink the guard against sweeping a
// live sidecar scan's temp dir.
sweepThreshold := max(c.ScanTimeout, time.Hour)
```

Same for the doc side: the `scanTimeout` row in `docs/CONFIGURATION.md` still reads "Timeout for SBOM generation" only, and this PR gives the option a second, non-obvious effect. The `scannerReadinessTimeout` and `shutdownTimeout` rows already document their edge cases in that style, so one sentence here would match convention.

internal/tools/tools.go
```diff
@@ -105,6 +107,42 @@ func DeleteContents(dir string) error {
...
+	for _, e := range entries {
+		if !e.IsDir() || !strings.HasPrefix(e.Name(), prefix) {
+			continue
+		}
+		info, err := e.Info()
+		if err != nil {
+			allErrs = errors.Join(allErrs, err)
+			continue
+		}
+		if !info.ModTime().Before(cutoff) {
```
matthyx, line 134: Agreeing with CodeRabbit that root-dir `mtime` is a weak liveness signal, but disagreeing that it needs a lock.

Confirmed against the fork (`matthyx/stereoscope`, `pkg/file/temp_dir_generator.go`): the root `/tmp/stereoscope-XXXX` is created lazily by `getOrCreateRootLocation`, and `NewDirectory` then creates children *inside* it. So the root's mtime is bumped when a child dir is created — but a long layer download that appends to an already-created file does not touch it. A sidecar pulling a large image for longer than the threshold can therefore look idle. Real, and matched by the tradeoff already documented in the PR description.

A held lock is a heavy fix for a best-effort sweep. Much cheaper, and it closes most of the gap: take the newest mtime in the candidate's immediate children rather than the root's own, since an active pull keeps creating per-layer entries.

```go
newest := info.ModTime()
if children, err := os.ReadDir(path.Join(dir, e.Name())); err == nil {
    for _, c := range children {
        if ci, err := c.Info(); err == nil && ci.ModTime().After(newest) {
            newest = ci.ModTime()
        }
    }
}
if !newest.Before(cutoff) {
    continue
}
```

One extra `readdir` per candidate at startup, no new abstraction, and it stays best-effort. Fine as a follow-up.

internal/tools/tools_test.go
```diff
@@ -228,3 +232,83 @@ func TestReferenceMatchForms(t *testing.T) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			dir := t.TempDir()
+			tt.setup(t, dir)
+			n, err := CleanupStaleTempDirs(dir, tt.prefix, tt.age)
+			if tt.wantErr {
+				require.Error(t, err)
+			} else {
+				require.NoError(t, err)
+			}
+			assert.Equal(t, tt.wantN, n)
```
matthyx, line 306: Seconding CodeRabbit's nitpick, with the concrete failure it lets through: in the "old matching dirs removed, fresh and unrelated kept" case, an implementation that ignored the prefix and deleted `other-xyz` + one `stereoscope-*` dir would also return `2` and pass. Reading the dir back and asserting the surviving names would pin the prefix and freshness filters rather than just the count.

matthyx (approved):
### Review: `feat: sweep stale stereoscope temp dirs at startup`

**Verdict: approved — no blockers.** The change is small, correctly scoped to #483, non-fatal on error, adds no deps, and the helper is well tested.

What I verified locally on `85cad0e`:
- `go build ./...`, `go vet ./internal/tools/ ./cmd/http/` clean; `go test ./internal/tools/ ./cmd/http/ ./config/` all pass. All CI checks green.
- The premise that one sweep covers both processes holds: in `kubescape-operator`'s `kubevuln/deployment.yaml`, the `kubevuln` and `sbom-scanner` containers both mount the same `tmp-dir` volume at `/tmp` (emptyDir, or a PVC when `kubevuln.tmpDirPersistence.enabled`). So the main process really does see the sidecar's orphans, and the PVC case is exactly the accumulation scenario #483 describes.
- Placement in `cmd/http/main.go` is correct: after the sidecar readiness wait, before the router is built and before any request can be served, so no in-process scan can be swept.
- Prefix and lifecycle match `file.NewTempDirGenerator("stereoscope")` in the `matthyx/stereoscope` fork (`pkg/file/temp_dir_generator.go`): root is `os.MkdirTemp("", "stereoscope-")`, created lazily, children created inside it. `DirEntry.IsDir()` also means a symlink named `stereoscope-*` is skipped rather than followed — good.

Four **non-blocking** follow-ups below, none of which need to hold up the merge. The threshold one (1) is the one I'd most like to see addressed, here or in a follow-up.

**1. The sweep threshold is now a user-tunable value that can be much smaller than 5m.** `scanTimeout` is documented as user-configurable (`docs/CONFIGURATION.md` schema allows any `^[0-9]+(s|m|h)$`), and operators do tune it. At `scanTimeout: 30s`, the sweep guard window shrinks to 30s and the collateral-damage window in (2) gets much wider. #483 originally proposed a flat `time.Hour`. A floor (`max(c.ScanTimeout, time.Hour)`) keeps the "derived from real scan semantics" property while bounding the downside.

**2. `scanTimeout` gains a second meaning that isn't documented.** This repo documents these subtleties carefully — the `scannerReadinessTimeout` and `shutdownTimeout` rows both spell out their edge-case behaviour. The `scanTimeout` row still says only "Timeout for SBOM generation"; it now also controls the startup temp-dir sweep threshold. Worth one sentence in `docs/CONFIGURATION.md`.

**3. Coverage gap: only a *main*-container restart sweeps.** The `sbom-scanner` container is the memory-heavy one (it exists precisely so Syft's memory is isolated, and the chart sets `GOMEMLIMIT` from `sbomScanner.resources.limits.memory`), so it is the likeliest OOM victim — but a sidecar OOM restarts *only that container*, and nothing sweeps. With `tmpDirPersistence.enabled`, those orphans then survive until the main container happens to restart, which may be days. This PR is still a strict improvement and I would not block on it, but the dominant orphan producer isn't covered by the dominant restart event. A sweep at `cmd/sbom-scanner/main.go` startup would close it (that binary loads no config today, so it would need an env var or a constant).

**4. Test asserts counts but not identities** — seconding CodeRabbit's nitpick. In the first case, an implementation that wrongly deleted `other-xyz` plus one `stereoscope-*` dir still returns `2` and passes. Asserting the surviving entry names would pin the prefix and freshness filters, not just the arithmetic.

I've also left an inline note on the `mtime`-as-liveness question CodeRabbit raised — I agree it's real but disagree that it needs a lock; see there.

PR #494, opened 2026-08-07
fix: resolve TOCTOU race condition in sidecar version cache

### LINKED ISSUE #493: Bug: TOCTOU race condition in SidecarSBOMAdapter.Version() bypasses cache under load
# Description

In `adapters/v1/sidecar.go`, the `SidecarSBOMAdapter.Version()` method caches the scanner version to avoid repeatedly querying the sidecar container. It uses `singleflight.Group` to coalesce concurrent cache misses.

However, there is a **time-of-check to time-of-use (TOCTOU)** race condition.

When the `singleflight.Do` call completes, the `"version"` key is removed from the internal `singleflight` map before the caller acquires `s.versionMu.Lock()` to update `s.versionStr`. During this small window, another goroutine can enter `Version()`, observe that `s.versionStr` is still empty, and start a completely new `singleflight.Do` call because the previous key has already been removed.

As a result, multiple concurrent goroutines can perform redundant `Health()` network requests, defeating the purpose of both the cache and `singleflight` under high concurrency.

---

# Environment

- **OS:** macOS 26.5.2 (Build 25F84)
- **Version:** `fb16373` (Git commit hash)

---

# Steps To Reproduce

1. Create a Go unit test at `adapters/v1/sidecar_race_test.go`:

```go
package v1

import (
	"context"
	"sync"
	"sync/atomic"
	"testing"

	sbomscanner "github.com/kubescape/kubevuln/pkg/sbomscanner/v1"
)

type mockScannerClientRace struct {
	sbomscanner.SBOMScannerClient
	healthCalls int32
}

func (m *mockScannerClientRace) Health(ctx context.Context) (string, bool, error) {
	atomic.AddInt32(&m.healthCalls, 1)
	return "test-version", true, nil
}

func TestSidecarVersion_RaceCondition(t *testing.T) {
	mockClient := &mockScannerClientRace{}

	adapter := &SidecarSBOMAdapter{
		client: mockClient,
	}

	var wg sync.WaitGroup
	workers := 100

	for i := 0; i < workers; i++ {
		wg.Add(1)

		go func() {
			defer wg.Done()

			for j := 0; j < 1000; j++ {
				version := adapter.Version()

				if version != "test-version" {
					t.Errorf("expected test-version, got %s", version)
				}
			}
		}()
	}

	wg.Wait()

	calls := atomic.LoadInt32(&mockClient.healthCalls)

	if calls > 1 {
		t.Fatalf("BUG REPRODUCED: Health was called %d times! It should only be called exactly 1 time because of the cache.", calls)
	}
}
```

2. Navigate to the repository root.

3. Run the test:

```bash
go test -v -run TestSidecarVersion_RaceCondition ./adapters/v1
```

4. Observe the test failure.

---

# Expected Behavior

Because `Version()` uses both caching and `singleflight`, `Health()` should be invoked only once, regardless of how many concurrent goroutines call `Version()`.

---

# Actual Behaviour / Output Logs

Under heavy concurrency, the race window allows multiple goroutines to bypass the cached value and execute redundant `Health()` requests.

**Observed test output:**

```text
=== RUN   TestSidecarVersion_RaceCondition
    sidecar_race_test.go:49: BUG REPRODUCED: Health was called 12 times! It should only be called exactly 1 time because of the cache.
--- FAIL: TestSidecarVersion_RaceCondition (0.01s)
FAIL
FAIL    github.com/kubescape/kubevuln/adapters/v1    1.377s
FAIL
```

---

# Additional Context

A potential fix is to apply the standard **double-checked locking** pattern within the `singleflight.Do` closure.

If a goroutine starts a new `singleflight.Do`, it should first acquire `s.versionMu` and check whether `s.versionStr` has already been populated by another goroutine before issuing a new `Health()` request. This eliminates the race window while preserving the existing caching and request coalescing behavior.

doraem-on:
## Overview

**Current behavior:** The `SidecarSBOMAdapter.Version()` method uses `singleflight.Group` to coalesce concurrent requests. However, there is a time-of-check to time-of-use (TOCTOU) race condition: when `singleflight.Do` completes, it removes its internal key before the executing goroutine writes the result to `s.versionStr`. A new concurrent caller entering during this small window can observe an empty cache, bypass `singleflight`, and trigger a redundant `Health()` request.

**New behavior:** A standard double-checked locking pattern has been added inside the `singleflight.Do` closure. Before issuing a `Health()` request, the closure now rechecks `s.versionStr` under `versionMu`. If another goroutine has already populated the cache, the cached value is returned immediately, eliminating redundant network calls under concurrent access.

---

## Additional Information

A new concurrency-focused unit test, `TestSidecarVersion_CachingUnderConcurrency`, has been added to verify that the cached scanner version is shared correctly across concurrent callers and to guard against future regressions.

---

## How to Test

1. Run the new unit test from the repository root:

   ```bash
   go test -v -run TestSidecarVersion_CachingUnderConcurrency ./adapters/v1
   ```

2. Verify that the test passes, confirming that `Health()` is invoked only once even under heavy concurrent access.

---

## Example Output

```text
=== RUN   TestSidecarVersion_CachingUnderConcurrency
--- PASS: TestSidecarVersion_CachingUnderConcurrency (0.01s)
PASS
ok      github.com/kubescape/kubevuln/adapters/v1    1.338s
```

---

## Related Issues

- Closes #493

---

## Checklist

- [x] I have commented on my code where appropriate.
- [x] I have added tests to cover the change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved version detection during concurrent requests.
  * Prevented redundant health checks when version information is already being retrieved or cached.
  * Preserved the fallback behavior when health checks fail.

* **Tests**
  * Added coverage for high-concurrency version requests and caching behavior.

### DIFF
--- a/adapters/v1/sidecar.go
+++ b/adapters/v1/sidecar.go
@@ -250,14 +250,14 @@ func (s *SidecarSBOMAdapter) Version() string {
 		if err != nil {
 			return "", err
 		}
-		
+
 		s.versionMu.Lock()
 		defer s.versionMu.Unlock()
 		s.versionStr = version
-		
+
 		return version, nil
 	})
-	
+
 	if err != nil {
 		logger.L().Warning("failed to get scanner version", helpers.Error(err))
 		return "unknown"
--- a/adapters/v1/sidecar.go
+++ b/adapters/v1/sidecar.go
@@ -234,24 +234,36 @@ func (s *SidecarSBOMAdapter) Version() string {
 	s.versionMu.Unlock()
 
 	v, err, _ := s.versionGroup.Do("version", func() (any, error) {
+		// Another goroutine may have populated versionStr after the
+		// singleflight key was released but before this closure executes.
+		// Recheck the cache to avoid issuing a redundant Health() request.
+		s.versionMu.Lock()
+		if s.versionStr != "" {
+			s.versionMu.Unlock()
+			return s.versionStr, nil
+		}
+		s.versionMu.Unlock()
+
 		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 		defer cancel()
 		version, _, err := s.client.Health(ctx)
 		if err != nil {
 			return "", err
 		}
+		
+		s.versionMu.Lock()
+		defer s.versionMu.Unlock()
+		s.versionStr = version
+		
 		return version, nil
 	})
+	
 	if err != nil {
 		logger.L().Warning("failed to get scanner version", helpers.Error(err))
 		return "unknown"
 	}
 
-	version := v.(string)
-	s.versionMu.Lock()
-	s.versionStr = version
-	s.versionMu.Unlock()
-	return version
+	return v.(string)
 }
 
 func (s *SidecarSBOMAdapter) GetMaxImageSize() int64 {
--- a/adapters/v1/sidecar_race_test.go
+++ b/adapters/v1/sidecar_race_test.go
@@ -1,0 +1,63 @@
+package v1
+
+import (
+	"context"
+	"sync"
+	"sync/atomic"
+	"testing"
+
+	sbomscanner "github.com/kubescape/kubevuln/pkg/sbomscanner/v1"
+)
+
+type mockScannerClientRace struct {
+	sbomscanner.SBOMScannerClient
+	healthCalls int32
+}
+
+func (m *mockScannerClientRace) Health(ctx context.Context) (string, bool, error) {
+	atomic.AddInt32(&m.healthCalls, 1)
+	return "test-version", true, nil
+}
+
+func TestSidecarVersion_CachingUnderConcurrency(t *testing.T) {
+	mockClient := &mockScannerClientRace{}
+	
+	adapter := &SidecarSBOMAdapter{
+		client: mockClient,
+	}
+
+	var wg sync.WaitGroup
+	var readyWg sync.WaitGroup
+	startCh := make(chan struct{})
+	workers := 100
+	
+	// Launch many goroutines that constantly call Version() to hit the tiny race window
+	// between singleflight.Do returning and versionMu being locked.
+	for i := 0; i < workers; i++ {
+		wg.Add(1)
+		readyWg.Add(1)
+		go func() {
+			defer wg.Done()
+			readyWg.Done()
+			<-startCh
+			
+			for j := 0; j < 1000; j++ {
+				version := adapter.Version()
+				if version != "test-version" {
+					t.Errorf("expected test-version, got %s", version)
+				}
+			}
+		}()
+	}
+
+	readyWg.Wait()
+	close(startCh)
+	wg.Wait()
+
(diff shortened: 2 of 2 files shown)


bhuvan-somisetty:
Went through this against the actual `singleflight` semantics to double-check the fix (key is only forgotten *after* the closure returns, per the package source) moving the `versionStr` write inside the `Do` closure genuinely closes the race from #493.

The recheck inside the closure is worth keeping too it's not redundant with the outer check. It catches the narrower window where an independent `Do()` call finishes and populates the cache between this goroutine's outer check and its own `Do()` call:

```go
v, err, _ := s.versionGroup.Do("version", func() (any, error) {
	s.versionMu.Lock()
	if s.versionStr != "" {
		s.versionMu.Unlock()
		return s.versionStr, nil   // <- catches that narrower window
	}
	s.versionMu.Unlock()
	...
```

One small thing — `gofmt -l` flags a few trailing-whitespace blank lines in both changed files:

```diff
--- adapters/v1/sidecar.go
@@
 		if err != nil {
 			return "", err
 		}
-		
+
 		s.versionMu.Lock()
...
 	})
-	
+
 	if err != nil {
```

```diff
--- adapters/v1/sidecar_race_test.go
@@
 	mockClient := &mockScannerClientRace{}
-	
+
 	adapter := &SidecarSBOMAdapter{
```

Doesn't affect behavior, just a `gofmt -w` away.

doraem-on:
@bhuvan-somisetty Yes, exactly. I specifically moved the versionStr write inside the Do closure because, as you noted, singleflight purges the key immediately after the closure returns. Without placing the write inside, there is an unavoidable window where the key is gone but the cache isn't populated yet, meaning a high influx of concurrent requests would bypass the singleflight group entirely.

The double-checked locking (the inner check) was also an intentional design choice on my part to catch the narrow edge case where a delayed goroutine enters Do() right after a previous one finishes populating the cache. It ensures we return instantly without making a redundant Health() network call under heavy concurrency.I would like to have @matthyx views on these.

But thanks for catching those trailing whitespaces from gofmt and appreciate your efforts and time of trying to review another contributor's PR.

bhuvan-somisetty:
Glad it lines up with what you intended the `Do()`-purges-key-immediately detail was the part I wasn't 100% sure of until I actually pulled the singleflight source, so good to have it confirmed. Makes sense to get @matthyx's take too. Good catch on the original race btw, that's a nasty one to spot.

matthyx (approved):
Verified: the double-checked locking correctly closes the TOCTOU window between `singleflight.Do` releasing its key and `versionStr` being written. Built, vetted, and ran `go test -race ./adapters/v1/...` (including `TestSidecarVersion_CachingUnderConcurrency`) — all pass, no races detected. CodeRabbit's nitpick about the workers racing before a synchronized start was already addressed via the `readyWg`/`startCh` barrier in the latest commit. CI is green. LGTM.

bhuvan-somisetty:
Nice, matches what we found glad it holds up under `-race` too.

PR #500, opened 2026-08-07
refactor(sbomscanner): make registry auth fallback pluggable

### LINKED ISSUE #492: Refactor registry auth fallback into pluggable providers before adding ECR/ACR
## Overview
The current sidecar Syft registry-auth path in `pkg/sbomscanner/v1/server.go` handles GCP as a special-case fallback inside `resolveSource()`.

Today the flow is:
- pull with provided credentials
- if the pull returns `MANIFEST_UNKNOWN`, retry with `imageTag`
- if the pull returns `401 Unauthorized` and `isGCPRegistry(imageID)` matches, call `gcpCredentials()` via ADC and retry
- if that still fails, or if the registry is not GCP, retry anonymously

This works for GCP, but it hard-codes one cloud-specific auth strategy directly into the generic pull retry logic.

## Problem
This makes the code harder to extend to additional cloud registries such as ECR and ACR:
- registry detection is coupled to inline string matching (`isGCPRegistry`)
- token acquisition is coupled to the retry branch (`gcpCredentials` / `gcpCredsFn`)
- adding each new provider will grow `resolveSource()` into a chain of registry-specific conditionals
- tests currently validate GCP behavior, but there is no abstraction boundary for provider-level behavior

As a result, the current implementation is reactive and registry-specific rather than pluggable.

## Solution
Extract registry-specific auth fallback into a small provider interface and let `resolveSource()` consult a provider list when a `401 Unauthorized` occurs.

A possible shape:
- `Matches(host string) bool`
- `Credentials(ctx context.Context) (*image.RegistryCredentials, error)`

Then:
- parse the registry host once from the image reference
- iterate providers in order
- if a provider matches, attempt to fetch credentials and retry once with those credentials
- keep anonymous fallback behavior unchanged when no provider applies or provider credential acquisition fails

This would let GCP become the first implementation instead of a hard-coded branch, and would provide a clean place to add ECR and ACR support later.

## Alternatives
Keep the current ad hoc logic and add ECR/ACR as additional branches in `resolveSource()`.

That is simpler short-term, but it makes the control flow progressively harder to reason about and test as more registries are added.

## Additional context
Relevant code today:
- `pkg/sbomscanner/v1/server.go`: `isGCPRegistry()`, `gcpCredentials()`, `gcpCredsFn`, and the `401 Unauthorized` branch inside `resolveSource()`
- `pkg/sbomscanner/v1/server_test.go`: `TestIsGCPRegistry` and `TestResolveSource`

This appears related to, but narrower than, #254. That issue tracks broader registry support; this issue focuses specifically on refactoring the auth fallback path so ECR/ACR support can be added without further entangling provider-specific logic in `resolveSource()`.

rootp1:
## Overview
Extracts the sidecar's GCP-specific 401 auth fallback in `resolveSource` (`pkg/sbomscanner/v1/server.go`) into a small `registryAuthProvider` interface (`Matches(imageID) bool`, `Credentials(ctx) (*image.RegistryCredentials, error)`), with GCP as the first implementation via `registryAuthProviders`. This gives ECR/ACR a clean extension point instead of adding more inline conditionals to `resolveSource`.

Behavior is unchanged: on a 401, the first matching provider's credentials are tried once, then anonymous fallback applies exactly as before.

## How to Test
- `go test ./pkg/sbomscanner/v1/... -run 'TestResolveSource|TestIsGCPRegistry'`

## Related issues/PRs:

Resolved #492

## Summary by CodeRabbit

- **New Features**
  - Registry scanning now supports ordered authentication providers for broader private registry compatibility.
  - Credentials are retrieved and retried automatically after an unauthorized response.
  - Anonymous access is attempted when providers return no credentials and the registry permits it.

- **Bug Fixes**
  - Improved registry source matching when resolving image references by digest or fallback tag.
  - Improved handling of unauthorized registry responses during source resolution.

### DIFF
--- a/pkg/sbomscanner/v1/server.go
+++ b/pkg/sbomscanner/v1/server.go
@@ -119,10 +119,10 @@ func resolveSource(ctx context.Context, get sourceGetter, imageID, imageTag stri
 	if err != nil && strings.Contains(err.Error(), "401 Unauthorized") {
 		unauthorizedErr := err
 		for _, provider := range registryAuthProviders {
-			if !provider.Matches(imageID) {
+			if !provider.Matches(pullRef) {
 				continue
 			}
-			if creds, credErr := provider.Credentials(ctx); credErr != nil {
+			if creds, credErr := provider.Credentials(ctx); credErr != nil || creds == nil {
 				logger.L().Debug("registry auth provider credentials unavailable, falling back to anonymous",
 					helpers.Error(credErr),
 					helpers.String("imageID", imageID))
--- a/pkg/sbomscanner/v1/server.go
+++ b/pkg/sbomscanner/v1/server.go
@@ -74,12 +74,37 @@ func gcpCredentials(ctx context.Context) (*image.RegistryCredentials, error) {
 // unit-tested without a live GCP environment.
 var gcpCredsFn = gcpCredentials
 
+// registryAuthProvider supplies registry credentials for hosts it recognizes, so
+// resolveSource can consult cloud-specific auth fallbacks without hard-coding each
+// one into its retry logic.
+type registryAuthProvider interface {
+	// Matches reports whether this provider handles the given pull reference.
+	Matches(imageID string) bool
+	// Credentials fetches credentials for a host this provider matches.
+	Credentials(ctx context.Context) (*image.RegistryCredentials, error)
+}
+
+// gcpAuthProvider resolves credentials for GCR/Artifact Registry hosts via
+// Application Default Credentials.
+type gcpAuthProvider struct{}
+
+func (gcpAuthProvider) Matches(imageID string) bool { return isGCPRegistry(imageID) }
+
+func (gcpAuthProvider) Credentials(ctx context.Context) (*image.RegistryCredentials, error) {
+	return gcpCredsFn(ctx)
+}
+
+// registryAuthProviders is the ordered list of cloud registry auth fallbacks resolveSource
+// consults on a 401 Unauthorized, before falling back to anonymous access. Add ECR/ACR
+// providers here as they're implemented.
+var registryAuthProviders = []registryAuthProvider{gcpAuthProvider{}}
+
 // sourceGetter abstracts the syft.GetSource call so the fallback ordering in
 // resolveSource is testable with a scripted implementation.
 type sourceGetter func(ctx context.Context, ref string, opts *image.RegistryOptions) (source.Source, error)
 
 // resolveSource downloads an image source, applying the MANIFEST_UNKNOWN and
-// 401/ADC/anonymous fallbacks in order. It mirrors the retry chain in
+// 401/provider/anonymous fallbacks in order. It mirrors the retry chain in
 // adapters/v1/syft.go so the sidecar and in-process adapters behave the same.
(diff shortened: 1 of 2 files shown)


pkg/sbomscanner/v1/server.go
```diff
@@ -93,18 +118,22 @@ func resolveSource(ctx context.Context, get sourceGetter, imageID, imageTag stri
 	}
 	if err != nil && strings.Contains(err.Error(), "401 Unauthorized") {
 		unauthorizedErr := err
-		if isGCPRegistry(imageID) {
-			if gcpCreds, gcpErr := gcpCredsFn(ctx); gcpErr != nil {
-				logger.L().Debug("GCP ADC unavailable, falling back to anonymous",
-					helpers.Error(gcpErr),
+		for _, provider := range registryAuthProviders {
+			if !provider.Matches(imageID) {
```
matthyx: Pass `pullRef` instead of `imageID` to `provider.Matches(pullRef)` so credential selection evaluates the reference that returned 401 Unauthorized (which may have fallen back to `imageTag`).

pkg/sbomscanner/v1/server.go
```diff
@@ -93,18 +118,22 @@ func resolveSource(ctx context.Context, get sourceGetter, imageID, imageTag stri
...
+				continue
+			}
+			if creds, credErr := provider.Credentials(ctx); credErr != nil {
+				logger.L().Debug("registry auth provider credentials unavailable, falling back to anonymous",
+					helpers.Error(credErr),
 					helpers.String("imageID", imageID))
 			} else {
-				opts.Credentials = []image.RegistryCredentials{*gcpCreds}
+				opts.Credentials = []image.RegistryCredentials{*creds}
```
matthyx, line 130: Check for `creds == nil` before dereferencing `*creds`. If a provider returns `(nil, nil)`, dereferencing `*creds` will panic. Treat `creds == nil` as credentials unavailable and fall back to anonymous access.

matthyx (changes requested):
Thanks for this refactoring! I reviewed the PR and found two blocker issues that should be fixed:

1. **Nil Pointer Dereference Risk**: `provider.Credentials(ctx)` may return `(nil, nil)`. Line 130 dereferences `*creds` without checking if `creds` is `nil`, causing a panic.
2. **Pass `pullRef` to `Matches`**: Line 122 calls `provider.Matches(imageID)`. If `MANIFEST_UNKNOWN` triggered a retry with `pullRef = imageTag`, credential matching should evaluate `pullRef` (the actual reference that received the 401).

pkg/sbomscanner/v1/server.go
rootp1: Switched to matching on pullRef so the fallback tied to the actual 401'd reference.

pkg/sbomscanner/v1/server.go
rootp1: Added the nil check, credentials now only get assigned when non-nil and anonymous fallback covers the rest.

rootp1:
Pushed a fix for both blockers: provider matching now uses pullRef and credential handling guards against nil. Added regression tests for both cases, and go vet plus the package tests pass locally. @matthyx could you take another look when you get a chance?

matthyx (approved):
Both blockers (nil credentials check & matching against pullRef) have been fixed and tested. Looks good to merge!

PR #508, opened 2026-08-08
feat: add metric for degraded exception fetching 

### LINKED ISSUE #507: No metric for how often exception fetching degrades in applyExceptionsToManifest
## Description

`applyExceptionsToManifest` in `core/services/scan.go` returns a `degraded`
bool indicating whether CVE exception fetching partially failed (tracked
via `domain.ErrExceptionsDegraded`), and this is correctly used at 3 call
sites to avoid treating a transient fetch failure as an actual exception
removal.

However, this degraded state is never recorded anywhere. `internal/metrics/metrics.go`
(added recently) already tracks scan duration, completions, and rejections,
but has no counter for how often exception fetching degrades - so there's
currently no way to tell, from metrics alone, whether this is a rare
transient blip or happening on every scan.

## Why this matters

The VEX Ingestion project (kubevuln#387) will add a second, similar
external dependency (fetching VEX feeds), which will need the same kind
of "did this degrade" tracking. Establishing the metric pattern for the
existing exceptions-degraded case now gives the VEX work a pattern to
follow, rather than needing to invent one from scratch.

## Proposed fix

- Add `ExceptionsDegradedCounter` to `Metrics`, following the same pattern
  as the existing `RejectCounter`.
- Add an optional `metrics *metrics.Metrics` field to `ScanService` with a
  `SetMetrics()` method (not a constructor parameter, to avoid touching
  the ~18 existing `NewScanService` call sites, most of them in tests).
- Increment the counter inside `applyExceptionsToManifest` when it
  degrades, guarded by a nil check so it's safe when metrics aren't set.

Shreya2005-2005:
## Overview

`applyExceptionsToManifest` already detects when CVE exception fetching
degrades (via `domain.ErrExceptionsDegraded`) and handles it correctly
at the 3 call sites that check for it (avoiding treating a transient
fetch failure as an actual exception removal). But this degraded state
was never recorded anywhere, so there was no way to tell from metrics
alone whether this is a rare blip or happening on every scan.

## Changes

- Add `ExceptionsDegradedCounter` to `internal/metrics`, following the
  same pattern as the existing `RejectCounter`.
- Add an optional `metrics *metrics.Metrics` field + `SetMetrics()` to
  `ScanService`, rather than a constructor parameter, so the ~18
  existing `NewScanService` call sites (mostly tests) are unaffected.
- Increment the counter inside `applyExceptionsToManifest` when
  exceptions degrade, guarded by a nil check so it's a no-op if
  metrics were never set.
- Wire `SetMetrics()` into `cmd/http/main.go` alongside the existing
  `metrics.New()` call.

## Why this matters

The VEX Ingestion project (kubevuln#387) will add a second, similar
external dependency (fetching VEX feeds), which will need the exact
same kind of "did this degrade" tracking. This establishes the metric
pattern now, on the existing exceptions case, rather than needing to
invent one from scratch later.

## Testing

- `internal/metrics/metrics_test.go`: extended to confirm the new
  counter is exposed via the `/metrics` endpoint.
- `core/services/scan_metrics_test.go` (new): confirms the counter is
  a no-op with no panic when metrics aren't set, and confirms it
  actually increments when exception fetching degrades.
- Full `go test ./...` passes except a pre-existing, unrelated flaky
  test in `adapters/v1` (`Test_syftAdapter_CreateSBOM`) that depends
  on live Docker Hub network access — confirmed via `git stash` that
  this fails identically with these changes removed.

## Related issues/PRs:

* Resolved #507

## Summary by CodeRabbit

* **New Features**
  * Added monitoring for cases where CVE exception retrieval is degraded.
  * Exposed a Prometheus counter for degraded exception handling.

* **Bug Fixes**
  * Improved scan behavior during degraded CVE exception retrieval, including safe operation when metrics are unavailable.

* **Tests**
  * Added coverage confirming degraded processing and metric reporting work as expected.

### DIFF
--- a/cmd/http/main.go
+++ b/cmd/http/main.go
@@ -127,6 +127,7 @@ func main() {
 	if err != nil {
 		logger.L().Ctx(ctx).Fatal("metrics initialization error", helpers.Error(err))
 	}
+	service.SetMetrics(m)
 	controller, err = controller.WithMetrics(m)
 	if err != nil {
 		logger.L().Ctx(ctx).Fatal("metrics registration error", helpers.Error(err))
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -27,6 +27,7 @@ import (
 	v1 "github.com/kubescape/kubevuln/adapters/v1"
 	"github.com/kubescape/kubevuln/core/domain"
 	"github.com/kubescape/kubevuln/core/ports"
+	"github.com/kubescape/kubevuln/internal/metrics"
 	"github.com/kubescape/kubevuln/internal/tools"
 	"github.com/kubescape/storage/pkg/apis/softwarecomposition/v1beta1"
 	"github.com/kubescape/storage/pkg/registry/file/dynamicpathdetector"
@@ -55,6 +56,14 @@ type ScanService struct {
 	storage           bool
 	vexGeneration     bool
 	tooManyRequests   *cache.Cache
+	metrics           *metrics.Metrics
+}
+
+// SetMetrics attaches an optional Metrics instance to the service. It is not
+// a constructor parameter so existing callers (including the many test call
+// sites) are unaffected; metric recording is a no-op until this is called.
+func (s *ScanService) SetMetrics(m *metrics.Metrics) {
+	s.metrics = m
 }
 
 var _ ports.ScanService = (*ScanService)(nil)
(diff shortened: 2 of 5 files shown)


matthyx (approved):
LGTM! Verified changes:
-  () added to .
-  added to  and called in .
-  safely increments the counter when CVE exception retrieval is degraded ().
- Unit tests added in  and  pass as expected.

Approved for merge.

PR #511, opened 2026-08-08
fix(repositories): cache SecurityException/ClusterSecurityException List() results

### LINKED ISSUE #510: SecurityException/ClusterSecurityException exception lookups scale with scan volume, not CRD churn - no shared cache or watch behind GetSecurityExceptions
# Description

`GetSecurityExceptions` (repositories/apiserver.go:128-173) is the only place kubevuln reads `SecurityException`/`ClusterSecurityException` CRDs, and it does it with a plain `List()` call straight to the K8s API server through `dynamic.Interface` - no informer, no watch, no local cache. The one caching layer that exists sits a level up, in `BackendAdapter.GetCVEExceptions` (adapters/v1/backend.go:156-240), and it doesn't match the shape of the data it's caching, so the List() calls end up scaling with scan volume instead of with how often the exceptions actually change.

Three things compound here, same root cause, same fix surface:

1. Registry scans never use the cache at all. `cacheable := workload.Wlid != ""` (backend.go:171) is false for every `ScanRegistry` call, by design (registry scans carry no Wlid), so every single image in a registry scan does its own fresh, uncached List() of both CRD types.
2. Even on the cached path, the cache key is `account/cluster/namespace/kind/name/containerName/image` (backend.go:172-180) - per workload identity. But the data being fetched is per-namespace (SecurityException) or cluster-wide (ClusterSecurityException), not per-workload. So N distinct workloads scanned inside the same namespace within the same 5-minute TTL window trigger N separate List() calls that all return the same underlying CRDs, just filtered differently afterward.
3. All of this traffic - plus the `GetWorkloadLabels`/`GetNamespaceLabels` Get() calls that objectSelector/namespaceSelector matching needs (repositories/apiserver.go:181-216) - shares one `dynamic.Interface` client rate-limited to QPS=50/Burst=100 (repositories/apiserver.go:92-95). That's a small, easily-exhausted budget once you're issuing multiple List/Get calls per scan.

# Environment

Not version-specific - this is a structural gap in the current exception-resolution path (`GetSecurityExceptions` + `GetCVEExceptions`), present on main.

# Steps To Reproduce

1. Run kubevuln with SecurityException CRD integration enabled (`storage != nil && c.RiskAcceptance`, see cmd/http/main.go:96-102).
2. Create even a handful of `SecurityException`/`ClusterSecurityException` CRs in the cluster.
3. Trigger a `ScanRegistry` call against a registry with, say, 50 images - or just send `ScanCP`/`ScanCVE` for 50 distinct workloads/containers in the same namespace within a 5-minute window.
4. Watch API server audit logs (or client-go's rest client metrics) for `list`/`get` requests to `securityexceptions.kubescape.io` and `clustersecurityexceptions.kubescape.io`.

You'll see one List() of *both* CRD types per image/workload, not one List() shared across the batch - even though the underlying CRD set hasn't changed between those calls.

# Expected behavior

Exception data for a given namespace/cluster should be fetched roughly as often as it changes (or on a cache TTL sized around that), independent of how many workloads or images happen to be scanned against it in that window.

# Actual Behavior

List() traffic scales with scan volume, not with CRD churn. A registry scan of N images issues up to N pairs of List() calls (namespaced + cluster-scoped) for exception data that's identical across the whole batch. A namespace with N distinct workloads scanned within the same TTL window behaves the same way. Every one of those List()/Get() calls draws from the same QPS=50/Burst=100 dynamic client, so a large-enough batch will start client-side throttling and add latency to every scan sharing that client - not just the exceptions lookup.

# Additional context

**Why this matters in practice:** this isn't a rare edge case - `ScanRegistry` explicitly opts every image out of the cache (that's intentional and documented, and correct on its own terms), and a namespace with real workload diversity will naturally miss the per-workload cache key on every distinct container/image even though nothing about the exceptions changed. The failure mode is API server load plus self-inflicted client-side throttling, which shows up as slower scans across the board during exactly the kind of burst (registry scan, rollout-triggered scan wave) where you'd want kubevuln to be keeping up.

**Affected files:**
- `repositories/apiserver.go` - `GetSecurityExceptions`, `GetWorkloadLabels`, `GetNamespaceLabels`, the dynamic client construction (QPS/Burst)
- `adapters/v1/backend.go` - `GetCVEExceptions`, `exceptionsCache`, `exceptionsCacheTTL`, the cache key construction

**Acceptance criteria:**
- SecurityException/ClusterSecurityException data is fetched at a rate driven by CRD state (a shared, namespace/cluster-scoped cache or a watch-based local cache), not by per-workload/per-image scan volume
- Registry scans stop paying one uncached List() pair per image
- Existing behavior for degraded/selector-unresolved results (never caching those, self-healing on next call) is preserved
- No regression to the existing exception-matching semantics (resources/images/objectSelector/namespaceSelector) - only the fetch/cache layer changes

**Rough implementation shape** (open to whatever the maintainers prefer here):
- Split the cache: one layer keyed by namespace (for SecurityException) and one cluster-wide singleton (for ClusterSecurityException), separate from the existing per-workload merged-policy cache - so the raw CRD List() is shared across every workload in that scope, and only the per-workload match/filter step still runs per scan
- Or move to an informer/lister-backed read for both CRD types, so reads come from a local, watch-maintained cache instead of a live List() call
- Either way, `GetSecurityExceptions`'s signature and the degraded-result handling in `GetCVEExceptions` stay intact - this is a caching/fetch-path change, not a matching-semantics change

Happy to send a PR for this if it's welcome - wanted to flag the shape of the fix first since it touches both the repository and adapter layers.

bhuvan-somisetty:
## Overview

`GetSecurityExceptions` issued a fresh, uncached `List()` call to the K8s API for both `SecurityException` and `ClusterSecurityException` on every invocation. The result is identical for every workload sharing a namespace (SecurityException) or the whole cluster (ClusterSecurityException), but registry scans skip the outer per-workload policy cache entirely, and distinct workloads in the same namespace each miss it too - so List() traffic scaled with scan volume instead of with how often the exceptions actually change, all sharing one QPS=50/Burst=100 dynamic client.

This adds a lower-level cache inside `GetSecurityExceptions` itself: the namespaced list is cached per namespace, and the cluster-scoped list is cached once for the whole cluster, so every workload/image in that scope reuses the same List() instead of triggering its own.

## Additional Information

- `securityExceptionListCache` is a 30s-TTL `*cache.Cache` on `APIServerStore`. That's short enough to keep exception data close to live, but long enough to collapse a burst of many workloads/images scanned back-to-back into a single List().
- A failed List() is never cached, so a transient apiserver hiccup still self-heals on the next call instead of being pinned as "no exceptions" for the TTL - this preserves the behavior #477 relies on.
- The cache field is nil-safe, so it falls back to always listing if unset. That matters because several existing tests construct `APIServerStore{}` literals directly instead of going through the constructors.
- `GetSecurityExceptions`'s signature, matching semantics, and error propagation are unchanged - this is purely a caching/fetch-path change.

## How to Test

```
go test ./repositories/... -run GetSecurityExceptions -v
```

New tests:
- `TestAPIServerStore_GetSecurityExceptions_CachesAcrossWorkloads` - two calls for the same namespace collapse into one List() call; a new namespace still triggers its own namespaced list, but the cluster-scoped list stays shared across all of them.
- `TestAPIServerStore_GetSecurityExceptions_DoesNotCacheListFailures` - a failed List() is never cached, so the next call retries instead of replaying the failure.

`go build ./...`, `go vet ./repositories/...`, and the full `repositories`/`adapters`/`core` suites all pass.

## Related issues/PRs:

* Resolved #510

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved SecurityException listing performance by caching recent namespaced and cluster-wide results.
  * Reduced repeated data retrieval while keeping results current through time-based expiration.

* **Reliability**
  * Failed SecurityException list operations are not cached, allowing subsequent requests to retry normally.
  * Preserved handling for individual items that cannot be converted.

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/akyoto/cache"
 	"github.com/armosec/utils-k8s-go/wlid"
 	"github.com/kubescape/go-logger"
 	"github.com/kubescape/go-logger/helpers"
@@ -40,11 +41,29 @@ const (
 	vulnSummaryContNameFormat              string = "%s-%s-%s" // "<kind>-<name>-<container-name>"
 )
 
+// securityExceptionListCacheCleaningInterval/TTL bound how stale the raw SecurityException/
+// ClusterSecurityException List() results served by GetSecurityExceptions can be. Unlike
+// BackendAdapter's exceptionsCacheTTL (5m, keyed per workload), this cache sits below the
+// per-workload match/filter step and is keyed per namespace (SecurityException) or cluster-wide
+// (ClusterSecurityException), so every workload sharing that scope reuses the same List() —
+// see #510. 30s keeps the data close to live while still collapsing a burst of many
+// workloads/images scanned back-to-back into a single List() call.
+const (
+	securityExceptionListCacheCleaningInterval = 30 * time.Second
+	securityExceptionListCacheTTL              = 30 * time.Second
+	clusterSecurityExceptionListCacheKey       = "cse"
+)
+
 // APIServerStore implements both CVERepository and SBOMRepository with in-cluster storage (apiserver) to be used for production
 type APIServerStore struct {
 	StorageClient spdxv1beta1.SpdxV1beta1Interface
 	DynamicClient dynamic.Interface
 	Namespace     string
+
+	// securityExceptionListCache caches the raw List() results GetSecurityExceptions reads;
+	// nil is safe (falls back to always listing, e.g. in tests that construct APIServerStore
+	// literals directly instead of through the constructors below).
+	securityExceptionListCache *cache.Cache
 }
 
 var (
@@ -97,9 +116,10 @@ func NewAPIServerStorage(namespace string) (*APIServerStore, error) {
 		return nil, err
 	}
 	return &APIServerStore{
-		StorageClient: clientset.SpdxV1beta1(),
-		DynamicClient: dynClient,
-		Namespace:     namespace,
+		StorageClient:              clientset.SpdxV1beta1(),
+		DynamicClient:              dynClient,
+		Namespace:                  namespace,
+		securityExceptionListCache: cache.New(securityExceptionListCacheCleaningInterval),
 	}, nil
 }
 
@@ -114,7 +134,8 @@ func NewFakeAPIServerStorage(namespace string) *APIServerStore {
 			securityExceptionGVR:        "SecurityExceptionList",
 			clusterSecurityExceptionGVR: "ClusterSecurityExceptionList",
 		}),
-		Namespace: namespace,
+		Namespace:                  namespace,
+		securityExceptionListCache: cache.New(securityExceptionListCacheCleaningInterval),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed PR #511. The caching implementation for namespaced and cluster-scoped security exceptions is clean, safe against nil cache pointers, properly avoids caching List() errors, and includes comprehensive regression tests. Approved!

Issue #512, opened 2026-08-08
SBOM scan platform handling is inconsistent, silently drops requested arch, and has no end-to-end platform param


rootp1:
# Description
SBOM generation does not reliably honor the requested container platform (OS/arch), and there's no way to tell after the fact which platform was actually scanned.

- The in-process adapter (`adapters/v1/syft.go`) computes a default platform from `options.Platform`, but never passes it to `syft.GetSource()` via `.WithPlatform()` — the requested platform is silently dropped and Syft falls back to the host arch.
- The sidecar path (`pkg/sbomscanner/v1/server.go`) does wire `.WithPlatform()` correctly, so the same scan request produces different (and potentially wrong) results depending on which path handles it.
- Neither path records the platform that was actually used anywhere (no SBOM annotation, no log line), so a silently-wrong-arch SBOM is undetectable afterward.
- There is currently no end-to-end way for any caller to request a specific platform — the field exists locally in options but nothing upstream ever sets it.
- Separately, when a requested platform doesn't exist in an image's manifest, stereoscope/syft already detects this (`validatePlatform()` returns `image.ErrPlatformMismatch` — confirmed present in upstream `anchore/stereoscope@v0.1.9`, not something added by our fork), but kubevuln never checks for this error type, so it falls into the generic failure path instead of a clear "platform not found" reason.

# Environment
OS: N/A (server-side scan service, not host-specific)
Version: main

# Steps To Reproduce
1. Request an SBOM scan for a multi-arch image via the in-process adapter, specifying a platform other than the host arch.
2. Request the same scan via the sidecar path.
3. Compare the resulting SBOMs / scanned architecture.

# Expected behavior
Both scan paths honor the requested platform consistently, callers have an end-to-end way to request a specific platform, the platform actually used is recorded and inspectable, and a requested-but-unavailable platform produces a clear, distinguishable failure reason.

# Actual Behavior
The in-process adapter drops the requested platform and falls back to host arch; the sidecar path does not. Nothing records which platform was actually scanned. There is no way to request a platform end-to-end. Platform-not-found errors from stereoscope are swallowed into the generic error path.

# Sub-tasks
- [x] Fix `syft.go` to wire `.WithPlatform()` into `syft.GetSource()` so both paths behave consistently.
- [ ] Add a `platform` field to the scan command, populated by the operator from the Pod's node arch when available.
[...]

PR #513, opened 2026-08-08
fix(adapters): wire WithPlatform() into syft.GetSource() calls

rootp1:
## Summary
- The in-process adapter (`adapters/v1/syft.go`) computed a default platform from `options.Platform` but never passed it to `syft.GetSource()` via `.WithPlatform()`, so the requested platform was silently dropped and Syft fell back to whatever manifest it found.
- The sidecar path (`pkg/sbomscanner/v1/server.go`) already wires `.WithPlatform()` correctly, so the exact same scan request could produce a different result (and a different SBOM) depending on which path handled it.
- This brings the in-process adapter to parity with the sidecar path only. It does not change the default-to-host-arch fallback behavior, which is a separate, larger piece of work tracked as a follow-up sub-task.

First sub-task of #512.

## Test plan
- [x] `go build ./adapters/...`
- [x] `go vet ./adapters/...` (no new findings)
- [x] `go test ./adapters/v1/... -run Test_syftAdapter_CreateSBOM` — note: on a non-amd64 dev machine (e.g. Apple Silicon), several cases now fail because the test fixtures are amd64-only images and platform is no longer silently ignored; this reproduces the exact bug being fixed. CI runs on `ubuntu-latest` (amd64), matching the fixtures' architecture, so these should pass there.

### DIFF
--- a/adapters/v1/syft.go
+++ b/adapters/v1/syft.go
@@ -161,6 +161,14 @@ func (s *SyftAdapter) CreateSBOM(ctx context.Context, name, imageID, imageTag st
 	if options.Platform == "" {
 		options.Platform = runtime.GOARCH
 	}
+	platformStr := options.Platform
+	if !strings.Contains(platformStr, "/") {
+		platformStr = "linux/" + platformStr
+	}
+	imgPlatform, err := image.NewPlatform(platformStr)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


matthyx (approved):
LGTM! Verified that `WithPlatform(imgPlatform)` is correctly passed to all `syft.GetSource()` calls when specified, and leaves `imgPlatform` as `nil` when unconstrained. All unit and integration tests pass.

PR #515, opened 2026-08-08
fix(services): walk filterSBOM's relationship graph to a fixed point

### LINKED ISSUE #514: filterSBOM's relevancy filter silently drops transitively-relevant artifacts beyond 2 hops (order-dependent)
# Description

`filterSBOM` (core/services/scan.go) builds the relevancy-filtered SBOM' used for the "SBOM'/CVE'" flow in ScanCP - it's supposed to keep every artifact that transitively owns a file the runtime profile actually touched. The relationship-graph walk it does to figure out "transitively relevant" artifacts is two fixed passes over `sbom.Content.ArtifactRelationships`, not a real graph traversal. Depending on the order Syft happens to emit relationships in, an artifact nested 3+ levels deep (an archive inside an archive inside an archive - exactly the case `scanEmbeddedSboms`/`EnableEmbeddedSboms` exists to catalog) can be silently dropped from the filtered SBOM, along with any CVEs found in it.

# Environment

Not version-specific - structural bug in `filterSBOM`, present on main. Confirmed with a local unit test against the current implementation (see reproduction below), not just by reading the code.

# Steps To Reproduce

I wrote a minimal test that constructs a synthetic SBOM with a 3-level containment chain and calls `filterSBOM` directly:

- One file `file-F`, marked relevant (as if the runtime profile observed it being opened).
- Three packages: `pkg-A` directly contains `file-F`, `pkg-B` contains `pkg-A`, `pkg-C` contains `pkg-B`.
- `ArtifactRelationships` listed **outermost-first**: `{C→B}, {B→A}, {A→F}` (Syft doesn't document or guarantee any relationship ordering, so this is a legitimate input, not a contrived one).

```go
sbom := domain.SBOM{
	Content: &v1beta1.SyftDocument{
		Files: []v1beta1.SyftFile{
			{ID: "file-F", Location: v1beta1.Coordinates{RealPath: "/app/relevant.class"}},
		},
		Artifacts: []v1beta1.SyftPackage{
			{PackageBasicData: v1beta1.PackageBasicData{ID: "pkg-A", Name: "A"}},
			{PackageBasicData: v1beta1.PackageBasicData{ID: "pkg-B", Name: "B"}},
			{PackageBasicData: v1beta1.PackageBasicData{ID: "pkg-C", Name: "C"}},
		},
		ArtifactRelationships: []v1beta1.SyftRelationship{
			{Parent: "pkg-C", Child: "pkg-B", Type: "contains"},
			{Parent: "pkg-B", Child: "pkg-A", Type: "contains"},
			{Parent: "pkg-A", Child: "file-F", Type: "contains"},
		},
	},
}
relevantFiles := mapset.NewSet[string]()
relevantFiles.Add("/app/relevant.class")

filtered, _ := filterSBOM(sbom, instanceID, wlid, relevantFiles, labels, helpersv1.Full)
// filtered.Content.Artifacts contains pkg-A and pkg-B, but NOT pkg-C
```

Run it and check `filtered.Content.Artifacts` - `pkg-C` is missing. Reordering the same three relationships innermost-first (`{A→F}, {B→A}, {C→B}`) makes the bug disappear for this exact input, which is what makes it easy to miss in ad hoc testing - it depends on relationship order, not on the logic being correct.

# Expected behavior

Every artifact that transitively contains a relevant file should end up in the filtered SBOM, regardless of the order `ArtifactRelationships` happens to be in.

# Actual Behavior

`filterSBOM` does exactly two passes over `ArtifactRelationships`:
1. First pass: add `relationship.Parent` to `relationshipsArtifacts` when `relationship.Child` is a relevant file.
2. Second pass: add `relationship.Parent` to `relationshipsArtifacts` when `relationship.Child` is already in `relationshipsArtifacts`.

Pass 2 only catches one additional hop, and even that one hop is order-dependent: since `relationshipsArtifacts` is mutated in place while pass 2 iterates the same slice, whether a given relationship's `Child` is "already in the set" depends on whether an earlier entry in that same pass happened to add it first. There's no third pass, and no loop-until-no-more-additions, so anything beyond 2 hops from a relevant file - or even a 2-hop chain in the "wrong" order - is silently dropped.

# Additional context

**Why this matters in practice:** this directly affects `ScanCP`'s relevancy-filtered SBOM' and, downstream, CVE'/StoreVEX - the actual "what's really running and vulnerable" result kubescape's relevancy engine is built to produce. A dropped artifact isn't just missing from an internal list; its CVEs never show up as relevant findings for that workload. The bug's own trigger condition (multi-level nested archives) lines up with a feature this codebase already has a config flag for (`scanEmbeddedSboms`), so it's not a hypothetical edge case, it's the case that feature exists to catalog.

**Affected files:**
- `core/services/scan.go` - `filterSBOM` (the two-pass relationship walk, roughly lines 1031-1048 today)

**Acceptance criteria:**
- An artifact is included in the filtered SBOM if it transitively contains a relevant file, at any depth, regardless of the order `ArtifactRelationships` appears in
- No regression to the existing behavior for the 0-hop (direct file owner) and typical 1-hop cases, which already work today
- A test covering 3+ hop chains in at least one non-trivial ordering (not just the "lucky" order) is added, since that's exactly what let this go unnoticed

- Replace the two fixed passes with a proper closure: repeat the "does this relationship's Child belong to the relevant set" pass until a pass adds nothing new (bounded by the number of relationships, so it terminates), or build an adjacency map from Child→Parent once and do a single BFS/DFS from the relevant-file-owning artifacts
- Either approach is a small, local change confined to `filterSBOM` - it doesn't need to touch the relationship *data* Syft produces, just how this function walks it

bhuvan-somisetty:
## Overview

`filterSBOM` (core/services/scan.go) is supposed to keep every artifact that transitively owns a file the runtime profile actually touched. It computed that with exactly two fixed passes over `ArtifactRelationships` - the second pass only ever caught one hop past the direct file owner, and even that one hop depended on the order `ArtifactRelationships` happened to be in, since the pass mutates the same set it's checking against while iterating the same slice once.

Syft doesn't document or guarantee any relationship ordering, and this codebase already has `scanEmbeddedSboms` for exactly the multi-level nested-archive case (archive-in-archive) that needs more than one hop of transitivity. An artifact nested deeper than that, or even a 2-hop chain in the "wrong" order, was silently dropped from the filtered SBOM along with any CVEs found in it.

This replaces the two fixed passes with a worklist that walks the relationship graph to a fixed point - correct at any depth, independent of ordering.

## Additional Information

- Relationships are indexed by child once (`childToRelationships`), then a breadth-first walk starts from the relevant files and adds each newly-discovered relevant artifact to the next frontier, repeating until nothing new is found.
- Existing dedup (`addedRelationshipIDs`) and output-ordering for the already-working 0-hop and 1-hop cases is unchanged.
- The "filter relevant artifacts" step downstream is untouched - it already just filters by membership in `relationshipsArtifacts`, however that set was computed.

## How to Test

```
go test ./core/services/... -run TestFilterSBOM_TransitiveClosure -v
```

The new test constructs a 3-level containment chain (file F owned by pkg-A, pkg-A contained in pkg-B, pkg-B contained in pkg-C) with the relationships listed **outermost-first** (C->B, B->A, A->F) - the ordering that reproduces the bug against the old two-pass code. Before this fix, `pkg-C` was missing from the filtered artifacts; after, all three are present.

`go build ./...`, `go vet ./core/services/...`, and the full `core/services` suite all pass.

## Related issues/PRs:

* Resolved #514

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SBOM filtering to retain artifacts across deeply nested containment relationships.
  * Ensured related artifacts are preserved regardless of relationship ordering.
  * Preserved the source SBOM’s relationship order in filtered results.
  * Prevented valid relationships from being omitted in multi-level artifact hierarchies.

* **Tests**
  * Added coverage for three-level transitive artifact relationships and relationship ordering.

### DIFF
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -1040,23 +1040,36 @@ func filterSBOM(sbom domain.SBOM, instanceID instanceidhandler.IInstanceID, wlid
 		}
 	}
 
-	// filter relevant relationships. A relationship is relevant if the child is a relevant file
-	relationshipsArtifacts := mapset.NewSet[string]()
+	// filter relevant relationships and their transitively-relevant ancestor artifacts. A
+	// relationship is relevant if its child is a relevant file, or transitively, if its child
+	// is an artifact already known to be relevant (e.g. a package nested inside another
+	// package's archive). This walks the relationship graph to a fixed point instead of a
+	// fixed number of passes: Syft does not document or guarantee any particular relationship
+	// ordering, and a fixed number of passes can silently drop artifacts nested more than one
+	// level below the direct file owner depending on that order (see #514).
+	childToRelationships := make(map[string][]v1beta1.SyftRelationship, len(sbom.Content.ArtifactRelationships))
 	for _, relationship := range sbom.Content.ArtifactRelationships {
-		if addedFileIDs.Contains(relationship.Child) && !addedRelationshipIDs.Contains(getRelationshipID(relationship)) { // if the child is a relevant file
-			relationshipsArtifacts.Add(relationship.Parent)
-			addedRelationshipIDs.Add(getRelationshipID(relationship))
-			filteredSBOM.Content.ArtifactRelationships = append(filteredSBOM.Content.ArtifactRelationships, relationship)
-		}
+		childToRelationships[relationship.Child] = append(childToRelationships[relationship.Child], relationship)
 	}
 
-	// Add children of relevant relationships (that the parent is not relevant)
-	for _, relationship := range sbom.Content.ArtifactRelationships {
-		if relationshipsArtifacts.Contains(relationship.Child) && !addedRelationshipIDs.Contains(getRelationshipID(relationship)) {
-			relationshipsArtifacts.Add(relationship.Parent)
-			addedRelationshipIDs.Add(getRelationshipID(relationship))
-			filteredSBOM.Content.ArtifactRelationships = append(filteredSBOM.Content.ArtifactRelationships, relationship)
+	relationshipsArtifacts := mapset.NewSet[string]()
+	frontier := addedFileIDs.ToSlice()
+	for len(frontier) > 0 {
+		var next []string
+		for _, childID := range frontier {
+			for _, relationship := range childToRelationships[childID] {
+				relationshipID := getRelationshipID(relationship)
+				if addedRelationshipIDs.Contains(relationshipID) {
+					continue
+				}
+				addedRelationshipIDs.Add(relationshipID)
+				filteredSBOM.Content.ArtifactRelationships = append(filteredSBOM.Content.ArtifactRelationships, relationship)
+				if relationshipsArtifacts.Add(relationship.Parent) {
+					next = append(next, relationship.Parent)
+				}
+			}
 		}
+		frontier = next
 	}
 
 	// filter relevant artifacts. An artifact is relevant if it is in the relevant relationships
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -1052,6 +1052,12 @@ func filterSBOM(sbom domain.SBOM, instanceID instanceidhandler.IInstanceID, wlid
 		childToRelationships[relationship.Child] = append(childToRelationships[relationship.Child], relationship)
 	}
 
+	// Compute the closure first (which relationships/artifacts are relevant) without emitting
+	// anything: addedFileIDs.ToSlice() has no stable order, so a traversal that appends
+	// relationships as it discovers them would make the output order vary between calls on
+	// the same input whenever there's more than one relevant file. Emitting afterward, in a
+	// single pass over sbom.Content.ArtifactRelationships, keeps the output in the same
+	// deterministic source order the pre-fix code always had.
 	relationshipsArtifacts := mapset.NewSet[string]()
 	frontier := addedFileIDs.ToSlice()
 	for len(frontier) > 0 {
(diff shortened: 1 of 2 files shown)


bhuvan-somisetty:
Good catch, that was a real regression from the closure fix - emitting relationships as the BFS discovered them meant the output order rode on `ToSlice()`'s randomized iteration once there was more than one relevant file.

Split it into two passes now: compute the closure first (no emitting), then walk `sbom.Content.ArtifactRelationships` once in its original order and emit whatever the closure marked relevant. Output order is back to matching source order, deterministically. Added a regression test with two independent relevant-file chains interleaved in the source list, ran it x5 to make sure it's not just passing by luck of the map iteration - all green, plus the full suite and vet.

bhuvan-somisetty:
@matthyx this one's ready for review whenever you're free. Let me know if anything needs tweaking, happy to take another pass.

matthyx (dismissed):
LGTM! The BFS graph traversal correctly computes the transitive closure for nested SBOM artifact relationships of arbitrary depth. The 2-pass approach also ensures deterministic output order matching the source relationship slice, addressing CodeRabbit's feedback. Unit tests cover both transitive closure and ordering preservation.

matthyx:
rebase needed

bhuvan-somisetty:
@matthyx Rebased on main, conflict in scan_test.go resolved (just two independent test additions landing next to each other). Build, vet, and the full suite pass.

matthyx (approved):
Re-approving after rebase/merge with main. All CI status checks are green and tests pass cleanly.

PR #517, opened 2026-08-09
fix: prevent panic on non-string ScanCVE args in SubmitCVE

### LINKED ISSUE #516: Bug: Unchecked type assertions on external JSON crash the whole process
# Description
Unchecked type assertions on external JSON (`workload.Args`) cause a fatal panic inside `SubmitCVE`. If a non-string JSON value arrives for registry identifiers, it forces a panic. A similar unchecked type assertion in `ScanCP` was fixed previously (via commit `16131f2`), but the `SubmitCVE` adapter was missed.

 
# Steps To Reproduce
1. Start `kubevuln` process or mock a standard test environment.
2. Send a `ScanCommand` payload to `ScanCVE` where any of the registry/image identifiers (e.g. `identifiers.AttributeRegistryName`) inside `workload.Args` is an integer or non-string (e.g., `"registryName": 12345`).
3. The malicious or malformed payload progresses through the queue to the worker pool goroutines.
4. The goroutine running `SubmitCVE` executes `val.(string)` without checking.

# Expected behavior
The adapter should safely parse or gracefully handle the arguments without panicking (e.g., using comma-ok type assertions like `val, isStr := val.(string); if isStr { ... }`), keeping the main process and other inflight scans unaffected.

# Actual Behavior
A fatal panic kills the goroutine running `SubmitCVE`. Since `SubmitCVE` runs inside a `gammazero/workerpool` goroutine which does not inherently recover from panics, this unrecovered panic takes down the entire Go process and drops all in-flight scans.

# Additional context
Here is the raw test output trace showing the panic crash at `backend.go:438` when `identifiers.AttributeRegistryName` is passed as an `int` instead of a `string`:

```text
=== RUN   TestSubmitCVE_PanicsOnNonStringArgs
[warning] failed to parse image manifest from grype document. error: empty grype document
--- FAIL: TestSubmitCVE_PanicsOnNonStringArgs (0.00s)
panic: interface conversion: interface {} is int, not string [recovered, repanicked]

goroutine 58 [running]:
testing.tRunner.func1.2({0x10ab6f500, 0x4d93cad7f770})
	/opt/homebrew/Cellar/go/1.26.5/libexec/src/testing/testing.go:1974 +0x1a0
testing.tRunner.func1()
	/opt/homebrew/Cellar/go/1.26.5/libexec/src/testing/testing.go:1977 +0x318
panic({0x10ab6f500?, 0x4d93cad7f770?})
	/opt/homebrew/Cellar/go/1.26.5/libexec/src/runtime/panic.go:860 +0x12c
github.com/kubescape/kubevuln/adapters/v1.(*BackendAdapter).SubmitCVE(_, {_, _}, {0x4d93caa46fc0, 0x0, 0x0, {0x0, 0x0}, {0x0, 0x0}, ...}, ...)
	/Users/lalit/kubevuln/adapters/v1/backend.go:438 +0x1504
github.com/kubescape/kubevuln/adapters/v1.TestSubmitCVE_PanicsOnNonStringArgs(0x4d93cad946c8?)
	/Users/lalit/kubevuln/adapters/v1/backend_test.go:1030 +0x274
testing.tRunner(0x4d93cad946c8, 0x10b273720)
	/opt/homebrew/Cellar/go/1.26.5/libexec/src/testing/testing.go:2036 +0xc4
created by testing.(*T).Run in goroutine 1
	/opt/homebrew/Cellar/go/1.26.5/libexec/src/testing/testing.go:2101 +0x3a8

### PR DESCRIPTION
**Pull request by doraem-on**
## Overview

This PR fixes a bug in `adapters/v1/backend.go` where unchecked type assertions (e.g. `val.(string)`) on external JSON values from `workload.Args` could cause a fatal panic inside `SubmitCVE`.

If a client sent a non-string JSON value (such as an integer or boolean) for registry/image identifiers (e.g. `identifiers.AttributeRegistryName`), the goroutine executing `SubmitCVE` would crash.

Because it runs inside a worker pool outside the standard web request lifecycle, the unrecovered panic could kill the entire Go process and drop all in-flight scans.

This has been resolved by using safe, comma-ok type assertions.

## Additional Information

A very similar unchecked type assertion was previously fixed in `ScanCP` validation (Issue #415), but this particular path in `SubmitCVE` was missed.

## How to Test

 1. Start the `kubevuln` process locally.
 2. Submit a malformed `ScanCommand` payload to the `ScanCVE` endpoint where `workload.Args` contains an integer value for an expected string attribute (e.g. `"registryName": 12345`).
 3. Verify that the process safely handles or ignores the malformed input without panicking or crashing.
 4. Alternatively, run the newly added unit test:

    `go test -v -run TestSubmitCVE_NoPanicOnNonStringArgs ./adapters/v1`

## Examples/Screenshots

> Here are the logs for the new unit test confirming the fix works and no longer panics:

```text
=== RUN   TestSubmitCVE_NoPanicOnNonStringArgs
[warning] failed to parse image manifest from grype document. error: empty grype document
--- PASS: TestSubmitCVE_NoPanicOnNonStringArgs (0.00s)
PASS
ok  	github.com/kubescape/kubevuln/adapters/v1	1.327s
```

## Related issues/PRs

- Resolved #516

### DIFF
--- a/adapters/v1/backend.go
+++ b/adapters/v1/backend.go
@@ -438,26 +438,26 @@ func (a *BackendAdapter) SubmitCVE(ctx context.Context, cve domain.CVEManifest, 
 	finalReport.Designators.Attributes[identifiers.AttributeSBOMToolVersion] = cve.SBOMCreatorVersion
 	finalReport.Designators.Attributes[identifiers.AttributeSBOMToolName] = cve.SBOMCreatorName
 
-	if val, ok := workload.Args[identifiers.AttributeRegistryName]; ok {
-		finalReport.Designators.Attributes[identifiers.AttributeRegistryName] = val.(string)
+	if val, ok := workload.Args[identifiers.AttributeRegistryName].(string); ok {
+		finalReport.Designators.Attributes[identifiers.AttributeRegistryName] = val
 	}
-	if val, ok := workload.Args[identifiers.AttributeRepository]; ok {
-		finalReport.Designators.Attributes[identifiers.AttributeRepository] = val.(string)
+	if val, ok := workload.Args[identifiers.AttributeRepository].(string); ok {
+		finalReport.Designators.Attributes[identifiers.AttributeRepository] = val
 	}
-	if val, ok := workload.Args[identifiers.AttributeTag]; ok {
-		finalReport.Designators.Attributes[identifiers.AttributeTag] = val.(string)
+	if val, ok := workload.Args[identifiers.AttributeTag].(string); ok {
+		finalReport.Designators.Attributes[identifiers.AttributeTag] = val
 	}
-	if val, ok := workload.Args[identifiers.AttributeSensor]; ok {
-		finalReport.Designators.Attributes[identifiers.AttributeSensor] = val.(string)
+	if val, ok := workload.Args[identifiers.AttributeSensor].(string); ok {
+		finalReport.Designators.Attributes[identifiers.AttributeSensor] = val
 	}
-	if val, ok := workload.Args[identifiers.AttributeRegistryID]; ok {
-		finalReport.Designators.Attributes[identifiers.AttributeRegistryID] = val.(string)
+	if val, ok := workload.Args[identifiers.AttributeRegistryID].(string); ok {
+		finalReport.Designators.Attributes[identifiers.AttributeRegistryID] = val
 	}
-	if val, ok := workload.Args[identifiers.AttributeRegistryScanID]; ok {
-		finalReport.Designators.Attributes[identifiers.AttributeRegistryScanID] = val.(string)
+	if val, ok := workload.Args[identifiers.AttributeRegistryScanID].(string); ok {
+		finalReport.Designators.Attributes[identifiers.AttributeRegistryScanID] = val
 	}
-	if val, ok := workload.Args[identifiers.AttributeRegistryScanImagesCount]; ok {
-		finalReport.Designators.Attributes[identifiers.AttributeRegistryScanImagesCount] = val.(string)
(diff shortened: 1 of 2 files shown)


bhuvan-somisetty:
Checked out this branch and actually reproduced the bug to make sure the claim holds up - reverted just backend.go back to main and reran the new test:

panic: interface conversion: interface {} is int, not string
	.../adapters/v1.(*BackendAdapter).SubmitCVE(...)
		adapters/v1/backend.go:442

Confirms it, and it's worse than "a bad response gets logged" - SubmitCVE runs inside the HTTP controller's workerpool.Submit() goroutine (controllers/http.go), not inside a request handler, so gin's Recovery() middleware never sees it. gammazero/workerpool doesn't recover panics either (checked the vendored source, zero recover() calls in the library). An unrecovered panic in any goroutine takes down the whole Go process - so a single malformed registryName in one workload's Args really would kill kubevuln and drop every in-flight scan, not just this one.

Also checked whether any sibling val.(string) assertions on workload.Args were left unfixed elsewhere - they weren't, the two other Args-reading spots (scan.go's optionsFromWorkload) already use the safe two-step comma-ok pattern, so this closes out the whole class of it, not just one instance.

Ran the full adapters/v1 suite against the fix too, all green. Nice catch.

doraem-on:
@bhuvan-somisetty Yeah , It was basically a remote DoS vulnerability just waiting to happen.

It feels great to finally track this down and completely wipe out this class of crashes across the module before it could cause real chaos in production. Really appreciate you double-checking the rest of the codebase to make sure we got them all.

@matthyx  please have a look and appreciate your views, thanks.

matthyx (approved):
LGTM! The safe type assertion  cleanly prevents panics when  contains non-string values. Unit tests added cover non-string arguments and pass cleanly. All CI checks are passing.

PR #519, opened 2026-08-09
fix(repositories): enforce nil CVEManifest.Content consistently in VEX generation

### LINKED ISSUE #518: VEX generation (createVEX/updateVEX/markRelevantVulnerabilitiesAsAffectedInVex) inconsistently guards nil CVEManifest.Content
# Description

# Environment

# Steps To Reproduce

# Expected behavior

# Actual Behavior

# Additional context

**Affected files:**

**Acceptance criteria:**

bhuvan-somisetty:
## Overview

`createVEX`, `updateVEX`, and `markRelevantVulnerabilitiesAsAffectedInVex` (repositories/apiserver.go) each read `CVEManifest.Content`, but the nil-guarding around that was inconsistent - sometimes checked, sometimes not, within the same function. None of the current `StoreVEX` callers ever pass a nil `Content`, so nothing panics today, but the functions themselves had no defense of their own against it.

This enforces the guard consistently in all three places, and removes a dead unguarded duplicate loop in `updateVEX` along the way.

## Additional Information

- `markRelevantVulnerabilitiesAsAffectedInVex` now nil-checks `cvep` the same way its sibling `markIgnoredVulnerabilitiesInVex` already did - that sibling was the reference pattern.
- `createVEX`'s `Matches` loop is now under the same `cve.Content != nil` guard that already covered the `IgnoredMatches` loop right next to it, instead of being unguarded while its neighbor was guarded.
- `updateVEX` had a second, unguarded `IgnoredMatches` loop after the guarded block - dead code today since the guarded copy already adds everything it would, but still an unguarded dereference. Removed it, and guarded the `ignoredMap` build that follows (it reads the same `cve.Content`).
- No behavior change for the existing non-nil-`Content` path.

## How to Test

```
go test ./repositories/... -run TestAPIServerStore_StoreVEX_NilContentDoesNotPanic -v
```

The new test calls `StoreVEX` with a `CVEManifest{}` (nil `Content`) for both `cve` and `cvep`, twice - once to exercise the `createVEX` path (container doesn't exist yet) and once to exercise `updateVEX` (container already exists) - and asserts neither panics or errors.

`go build ./repositories/...`, `go vet ./repositories/...`, and the full `repositories` suite all pass.

## Related issues/PRs:

* Resolved #518

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when processing vulnerability records with missing content.
  * Prevented errors or crashes during initial creation and updates of incomplete vulnerability data.
  * Ensured no invalid VEX statements are generated when vulnerability content is unavailable.

* **Tests**
  * Added coverage for handling incomplete vulnerability records across creation and update workflows.

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -903,6 +903,9 @@ func buildActionStatement(v v1beta1.Match) string {
 }
 
 func markRelevantVulnerabilitiesAsAffectedInVex(vexDoc *v1beta1.VEX, cvep *domain.CVEManifest) error {
+	if cvep == nil || cvep.Content == nil {
+		return nil
+	}
 	// Now change the status of the filtered vulnerabilities to "Affected"
 	for _, v := range cvep.Content.Matches {
 		for i, s := range vexDoc.Statements {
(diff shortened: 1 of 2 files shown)


bhuvan-somisetty:
Good catch - the test only checked that updateVEX didn't panic/error, never actually confirmed statements stayed empty after it. Added the same re-fetch-and-assert-empty check after the second call that already existed after the first. All green.

bhuvan-somisetty:
@matthyx whenever you get a moment, would appreciate a look on this one. Happy to make any changes if you spot something.

matthyx (approved):
LGTM! The nil guards for `CVEManifest.Content` in `createVEX`, `updateVEX`, and `markRelevantVulnerabilitiesAsAffectedInVex` are implemented cleanly, dead code in `updateVEX` was removed, and the new unit test thoroughly verifies both create and update paths with nil Content. All CI checks are green.

PR #520, opened 2026-08-09
fix(repositories): guard nil CVEManifest.Content in the VEX path

### LINKED ISSUE #518: VEX generation (createVEX/updateVEX/markRelevantVulnerabilitiesAsAffectedInVex) inconsistently guards nil CVEManifest.Content

Lakshya77089:
## Overview

`createVEX` and `markRelevantVulnerabilitiesAsAffectedInVex` dereference `cve.Content` / `cvep.Content` without checking, while other loops in the same functions do check. No current caller passes a nil `Content`, so nothing panics today — but the functions carry no defence of their own, which is what #518 describes.

Two changes:

**Guard `Content` once at the top** of `createVEX` and `markRelevantVulnerabilitiesAsAffectedInVex`, rather than adding more inline checks alongside the existing ones.

**Drop the second `cve.Content.IgnoredMatches` loop in `updateVEX`.** It sat outside the surrounding `if cve.Content != nil` block and rebuilt the same statements the guarded loop above had already appended — same status, same justification, same `"Vulnerability was ignored by a SecurityException"` impact statement. Its own `found` check searches `vexDoc.Statements` for the same vulnerability/PURL pair, so it could only ever re-find what the first loop added. Dead code, and an unguarded dereference of the variable the function had just finished being careful about.

The inner `if cve.Content != nil` blocks are left alone. They are trivially true now that the top-level guard exists, but removing them means dedenting large blocks for no behavioural change, which makes the diff harder to review than it needs to be.

## How to Test

```
go test ./repositories/ -run storeVEX
```

Two tests added:

* `TestAPIServerStore_storeVEX_nilContent` — passes a zero-value `CVEManifest` as each of `cve` and `cvep`. **Fails on main** (panic), passes here.
* `TestAPIServerStore_storeVEX_ignoredMatchesNotDuplicated` — stores twice so the second call takes the update path where the removed loop lived, then asserts no statement appears more than once. Passes both before and after, which is the point: it demonstrates the removal changes nothing.

All six pre-existing `storeVEX` tests still pass.

## Related issues/PRs:

* Resolved #518

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented errors when vulnerability content is missing during VEX creation.
  * Improved handling of empty or unavailable filtered manifests.
  * Prevented ignored vulnerability entries from being duplicated during repeated VEX updates.
  * Ensured existing vulnerability statements retain accurate affected and not-affected statuses.

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -903,6 +903,9 @@ func buildActionStatement(v v1beta1.Match) string {
 }
 
 func markRelevantVulnerabilitiesAsAffectedInVex(vexDoc *v1beta1.VEX, cvep *domain.CVEManifest) error {
+	if cvep == nil || cvep.Content == nil {
+		return nil
+	}
 	// Now change the status of the filtered vulnerabilities to "Affected"
 	for _, v := range cvep.Content.Matches {
 		for i, s := range vexDoc.Statements {
@@ -934,6 +937,11 @@ func markRelevantVulnerabilitiesAsAffectedInVex(vexDoc *v1beta1.VEX, cvep *domai
 func (a *APIServerStore) createVEX(ctx context.Context, cve domain.CVEManifest, cvep domain.CVEManifest) error {
 	_, span := otel.Tracer("").Start(ctx, "APIServerStore.createVEX")
 	defer span.End()
+
+	if cve.Content == nil {
+		logger.L().Debug("no CVE content, skipping VEX creation")
+		return nil
+	}
 
 	imagePullable := cve.Annotations[helpersv1.ImageIDMetadataKey]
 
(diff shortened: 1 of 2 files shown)


Lakshya77089:
@bhuvan-somisetty apologies — you called this one first on the issue and I missed your comment before opening this. Your call: happy to close this in favour of yours, or leave it up if it saves you the work. No attachment either way.

One thing from it that may be useful regardless of which PR lands: the two `IgnoredMatches` loops in `updateVEX` build the same statement and share the same `found` dedupe against `vexDoc.Statements`, so the unguarded second one can only ever re-find what the guarded one already appended. That makes it dead code rather than a duplicate-statement bug, which is why removing it needs a test that passes both before and after — `TestAPIServerStore_storeVEX_ignoredMatchesNotDuplicated` here does that.

bhuvan-somisetty:
No worries, glad you caught it. You're right on the dedup - the second loop just re-finds what the guarded one already added, so it's dead code not an active dup, which is why I added a before/after test instead of just deleting it. Go ahead and close this one - #519 already covers it and is up to date. Thanks for the sanity check either way.

matthyx (commented):
### PR Review: Blocker Identified

Thank you for addressing the nil `CVEManifest.Content` handling! The approach is clean, but there is one remaining blocker where a nil `cve.Content` will cause a panic on the VEX update path:

#### 1. Nil pointer dereference in `updateVEX` (`repositories/apiserver.go`)
In `updateVEX`, `cve.Content.IgnoredMatches` is iterated without checking if `cve.Content` is non-nil:
```go
ignoredMap := make(map[string]bool)
for _, v := range cve.Content.IgnoredMatches {
	ignoredMap[v.Vulnerability.ID+v.Artifact.PURL] = true
}
```
If `StoreVEX` is called on an existing VEX document with a nil `cve.Content`, this will panic.

**Suggested Fix:**
Guard the loop with `if cve.Content != nil`:
```go
ignoredMap := make(map[string]bool)
if cve.Content != nil {
	for _, v := range cve.Content.IgnoredMatches {
		ignoredMap[v.Vulnerability.ID+v.Artifact.PURL] = true
	}
}
```

---

#### 2. Extend unit test for `updateVEX` nil-content path (`repositories/apiserver_test.go`)
`TestAPIServerStore_storeVEX_nilContent` currently only hits `createVEX` (because the container does not exist yet when called). Since `createVEX` returns early on `cve.Content == nil`, the test passes without exercising `updateVEX`.

**Suggested Fix:**
Seed an existing VEX container first, then call `StoreVEX` with `cve.Content == nil` using the same VEX name to ensure the update path handles nil content safely.

Lakshya77089:
Closing in favour of #519 — it was there first and you'd already reached the same conclusion about the dedup. Thanks for the quick reply.

bhuvan-somisetty:
For reference, #519 already guards that exact `ignoredMap` loop in `updateVEX`, and the test there does two `StoreVEX` calls with the same name so the second one exercises the update path with nil content directly (confirmed no panic).

PR #525, opened 2026-08-09
fix(repositories): guard nil CVEManifest.Content in parseSeverities

### LINKED ISSUE #524: parseSeverities/StoreCVESummary dereference CVEManifest.Content unguarded on the default (every-scan) write path
# Description

`parseSeverities` (repositories/apiserver.go), which computes the Critical/High/Medium/Low/Negligible/Unknown counts stored in every `VulnerabilityManifestSummary`, dereferences `cve.Content.Matches` completely unguarded, and `cvep.Content.Matches` guarded only by the caller-supplied `withRelevancy` bool - not by an actual nil check on `cvep.Content` itself. `StoreCVESummary`, which calls it, is part of the public `ports.CVERepository` interface, so this precondition ("Content is never nil") is not documented, not enforced by the type signature, and not visible to anyone implementing or calling the interface - it only holds today because of how `core/services/scan.go` happens to construct its arguments.

This is the same class of issue as #518 (unguarded `CVEManifest.Content` access in VEX generation), but on a materially bigger surface: `StoreCVESummary` runs on every successful scan with storage enabled - the default production configuration - across all four scan flows (`GenerateSBOM`, `ScanCVE`, `ScanCP`, `ScanRegistry`), not just when `vexGeneration` is opted into.

# Environment

Not version-specific - structural gap in the CVE-summary generation path, present on main.

# Steps To Reproduce

Read `repositories/apiserver.go`:

1. `parseSeverities(cve domain.CVEManifest, cvep domain.CVEManifest, withRelevancy bool) v1beta1.SeveritySummary` (~line 477): the first loop, `for i := range cve.Content.Matches`, has no nil check on `cve.Content` at all. The second loop, `for i := range cvep.Content.Matches`, is wrapped in `if withRelevancy { ... }` - that's a caller-intent flag, not a nil check on `cvep.Content`.
2. `StoreCVESummary` (~line 618) calls `parseSeverities(cve, cvep, withRelevancy)` directly, with no nil-check of its own beforehand.
3. `StoreCVESummary` is declared on `ports.CVERepository` (core/ports/repositories.go) - a public interface with two implementations (`APIServerStore`, `MemoryStore`). Nothing in the interface signature or its callers' contract says `cve.Content`/`cvep.Content` must be non-nil.
4. `MemoryStore.StoreCVESummary` (repositories/memory.go) never touches `.Content` at all - it's naturally nil-safe by construction, purely because the mock doesn't compute severities. That means any test written against `MemoryStore` (which is what most of this codebase's own tests use) cannot catch this class of bug; only the real `APIServerStore` implementation, used in production, has the unguarded dereference.

I traced all 7 current call sites of `StoreCVESummary` in `core/services/scan.go` and confirmed `cve.Content` is always non-nil by construction there (each call site is downstream of either a successful `ScanSBOM` or an already-non-nil cached manifest), and every call that passes a `cvep` with nil `Content` also passes `withRelevancy=false`, which skips the unguarded loop. So this doesn't panic today - but that's an invariant maintained entirely by `scan.go`'s calling discipline, not by `parseSeverities` or `StoreCVESummary` themselves, and it isn't visible anywhere in the interface those functions are declared on.

# Expected behavior

`parseSeverities` (and `StoreCVESummary`, as the entry point on a public interface) should handle nil `Content` on either argument without panicking, regardless of what any caller - current or future - passes.

# Actual Behavior

`parseSeverities` will panic with a nil pointer dereference the moment any caller passes `cve.Content == nil`, or passes `cvep.Content == nil` together with `withRelevancy == true`. Nothing in the code, the interface, or the type system prevents that.

# Additional context

**Why this matters beyond "not reachable today":** the exact same shape of bug was just found and fixed in the VEX generation path (#518/#519) - this repo has now hit this specific mistake (unguarded `CVEManifest.Content` access, safe only by caller convention) in two separate subsystems written by the same general pattern. `StoreCVESummary` is the higher-traffic one: it's on the default (storage-enabled) path for every scan type, not gated behind an opt-in config flag like VEX is, and it produces the actual severity counts surfaced to users. It's also on a public interface (`ports.CVERepository`), meaning the "safe" invariant depends on every future caller - anywhere in this codebase or a downstream consumer of the interface - independently rediscovering and preserving a constraint that isn't written down anywhere.

**Affected files:**
- `repositories/apiserver.go` - `parseSeverities`, `StoreCVESummary` (and worth auditing `parseVulnerabilitiesComponents` and `StoreCVE` alongside them while touching this area, since they're called with the same arguments from the same call sites)
- `core/ports/repositories.go` - `CVERepository.StoreCVESummary`'s doc comment, if any, should state the actual contract once this is fixed

**Acceptance criteria:**
- `parseSeverities` returns a sane zero-value `SeveritySummary` (or otherwise doesn't panic) when `cve.Content` and/or `cvep.Content` is nil, regardless of `withRelevancy`
- `StoreCVESummary` behaves correctly (no panic, no corrupted summary) when called directly with a `CVEManifest{}` for either argument - a test should call it this way, not just exercise it indirectly through `scan.go`
- No behavior change for the existing non-nil-`Content` path, and existing severity counts are unaffected
- The gap between what `MemoryStore` exercises and what `APIServerStore` actually does is closed - the new test must exercise `APIServerStore`, not just the mock

- Add a nil check at the top of `parseSeverities` for `cve.Content` (return an empty `SeveritySummary` if nil), and check `cvep.Content != nil` explicitly rather than relying on `withRelevancy` alone before the second loop
- Same treatment for `parseVulnerabilitiesComponents` if it has the same implicit assumption (it currently only reads `.Name`, not `.Content`, so likely fine, but worth confirming while in this code)
- Add a direct unit test against `APIServerStore.StoreCVESummary` with nil-`Content` manifests for both `cve` and `cvep`, in both `withRelevancy` states

bhuvan-somisetty:
## Overview

`parseSeverities` (repositories/apiserver.go) dereferenced `cve.Content.Matches` unconditionally, and `cvep.Content.Matches` guarded only by the caller-supplied `withRelevancy` bool - not by an actual nil check on `cvep.Content`. `StoreCVESummary`, which calls it, is part of the public `ports.CVERepository` interface, so this precondition was never documented or enforced anywhere; it only held because of how `core/services/scan.go` happens to construct its arguments today.

Unlike VEX generation (gated behind the `vexGeneration` config flag), `StoreCVESummary` runs on every successful scan with storage enabled - the default production configuration - across all four scan flows, and produces the severity counts shown to users. This guards both loops with an explicit nil check, the same pattern already applied to VEX generation in #519.

## Additional Information

- `parseVulnerabilitiesComponents` (the sibling function `StoreCVESummary` also calls) only reads `.Name` off `cve`/`cvep`, never `.Content` - confirmed it has no equivalent gap, so it's untouched.
- Checked `MemoryStore.StoreCVESummary` (the test-double implementation of the same interface) - it never touches `.Content` either, so it was naturally nil-safe by accident. That's part of why this had gone unnoticed: tests written against the mock can't exercise this, only the real `APIServerStore` implementation has the dereference.
- No behavior change for the existing non-nil-`Content` path - severity counts are computed identically when both are non-nil.

## How to Test

```
go test ./repositories/... -run TestAPIServerStore_StoreCVESummary_NilContentDoesNotPanic -v
```

The new test calls `StoreCVESummary` directly (not just indirectly through `scan.go`) with a `CVEManifest{}` (nil `Content`) for both `cve` and `cvep`, in both `withRelevancy` states, and calls it twice per state to exercise both the initial `Create` and the `RetryOnConflict` update path - asserts none of it panics or errors.

`go build ./repositories/...`, `go vet ./repositories/...`, and the full `repositories` suite all pass.

## Related issues/PRs:

* Resolved #524

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue that could cause CVE summary processing to fail when vulnerability content is missing.
  * CVE manifests with empty content are now handled safely during both creation and updates.

* **Tests**
  * Added coverage for empty-content CVE manifests across supported relevancy settings and processing paths.

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -488,23 +488,25 @@ func parseSeverities(cve domain.CVEManifest, cvep domain.CVEManifest, withReleva
 	var unknown int64
 	var unknownRelevant int64
 
-	for i := range cve.Content.Matches {
-		switch cve.Content.Matches[i].Vulnerability.Severity {
-		case domain.CriticalSeverity:
-			critical += 1
-		case domain.HighSeverity:
-			high += 1
-		case domain.MediumSeverity:
-			medium += 1
-		case domain.LowSeverity:
-			low += 1
-		case domain.NegligibleSeverity:
-			negligible += 1
-		case domain.UnknownSeverity:
-			unknown += 1
+	if cve.Content != nil {
+		for i := range cve.Content.Matches {
+			switch cve.Content.Matches[i].Vulnerability.Severity {
+			case domain.CriticalSeverity:
+				critical += 1
+			case domain.HighSeverity:
+				high += 1
+			case domain.MediumSeverity:
+				medium += 1
+			case domain.LowSeverity:
+				low += 1
+			case domain.NegligibleSeverity:
+				negligible += 1
+			case domain.UnknownSeverity:
+				unknown += 1
+			}
 		}
 	}
-	if withRelevancy {
+	if withRelevancy && cvep.Content != nil {
 		for i := range cvep.Content.Matches {
 			switch cvep.Content.Matches[i].Vulnerability.Severity {
 			case domain.CriticalSeverity:
--- a/repositories/apiserver_test.go
+++ b/repositories/apiserver_test.go
@@ -2484,3 +2484,39 @@ func TestAPIServerStore_StoreVEX_NilContentDoesNotPanic(t *testing.T) {
 	require.NoError(t, err)
 	assert.Empty(t, vexContainerUpdated.Spec.Statements, "updateVEX must not fabricate statements out of a nil CVEManifest.Content")
 }
+
+// TestAPIServerStore_StoreCVESummary_NilContentDoesNotPanic is a regression test for #524:
+// parseSeverities dereferenced cve.Content.Matches unguarded, and cvep.Content.Matches guarded
+// only by the caller-supplied withRelevancy bool rather than an actual nil check on
+// cvep.Content. StoreCVESummary is part of the public ports.CVERepository interface, so
+// nothing enforced that precondition on it. This calls StoreCVESummary directly (not just
+// indirectly through scan.go) with a CVEManifest{} - Content is nil - for both cve and cvep,
+// in both withRelevancy states, and asserts none of the four combinations panic or error.
+func TestAPIServerStore_StoreCVESummary_NilContentDoesNotPanic(t *testing.T) {
+	for _, withRelevancy := range []bool{false, true} {
+		t.Run(fmt.Sprintf("withRelevancy=%v", withRelevancy), func(t *testing.T) {
+			a := NewFakeAPIServerStorage("kubescape")
+			ctx := context.TODO()
+			workload := domain.ScanCommand{
+				Wlid:          "wlid://cluster-x/namespace-y/deployment-z",
+				ContainerName: "container",
+			}
+			ctx = context.WithValue(ctx, domain.WorkloadKey{}, workload)
+			ctx = context.WithValue(ctx, domain.TimestampKey{}, int64(123456))
+
+			empty := domain.CVEManifest{Name: name}
+
+			require.NotPanics(t, func() {
+				err := a.StoreCVESummary(ctx, empty, empty, withRelevancy)
+				assert.NoError(t, err, "StoreCVESummary must handle a nil CVEManifest.Content without erroring")
+			})
+
+			// Second call with the same (still nil-Content) manifests exercises the
+			// RetryOnConflict update path, not just the initial Create.
+			require.NotPanics(t, func() {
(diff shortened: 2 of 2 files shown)


bhuvan-somisetty:
Good catch - the original test only covered "both nil," which never actually reaches the cvep loop when withRelevancy is true. Added a case with a real match on cve and nil cvep.Content under withRelevancy=true, then read the stored summary back and asserted the real count shows up while the relevant count stays zero. All green.

matthyx (approved):
LGTM! The nil guards for  and  in  properly prevent nil pointer dereferences in . The added unit tests cover both nil-content cases and mixed nil-content cases across both create and conflict-retry update paths. All CI checks are passing.

bhuvan-somisetty:
Thanks for the review, matthyx!

PR #530, opened 2026-08-09
fix(sbom): resolve platform end-to-end, fix pod-less fallback, record resolved platform, classify platform mismatches

### LINKED ISSUE #512: SBOM scan platform handling is inconsistent, silently drops requested arch, and has no end-to-end platform param

rootp1:
## Overview

Follows on from #513 (which fixed the in-process adapter's missing `WithPlatform()` wiring) to close out the remaining sub-tasks of #512:

- Added `domain.ArgsPlatform` (`"platform"`), read by `optionsFromWorkload` into `RegistryOptions.Platform`. Any caller — an operator populating it from the scanned Pod's node architecture, or a direct API caller via `args.platform` (see updated `docs/API.md`) — now has an end-to-end way to request a platform.
- Fixed the sidecar path's pod-less fallback: `pkg/sbomscanner/v1/server.go` previously defaulted an empty `platform` to `runtime.GOARCH`, forcing a host-arch pull and breaking single-arch images on mismatched-arch nodes (registry rescans, periodic CRD-based rescans have no node context to derive a platform from). It now mirrors the in-process adapter's existing behavior: leave the platform unset and let Syft resolve whatever the manifest provides.
- Both adapters now record the platform actually resolved (`kubescape.io/resolved-platform` annotation) on every successful scan, whether or not one was explicitly requested, so a silently-wrong-arch SBOM is inspectable after the fact. Threaded through the sidecar's gRPC response (new `resolved_platform` proto field) into `domain.SBOM.Annotations`.
- `stereoscope`'s `*image.ErrPlatformMismatch` (returned once a requested platform doesn't exist in an image's manifest) is now classified as a distinct `scanfailure.ReasonImageNotFound` instead of falling into the generic `ReasonSBOMGenerationFailed` fallback — detected via `errors.As` in the in-process path, and via a string match on the sidecar path (the typed error doesn't survive the gRPC boundary). armoapi-go has no dedicated "platform not found" code yet, so this reuses the closest existing one; documented in `docs/scan-failure-reporting.md`.

## How to Test

```
go build ./...
go test ./... -skip 'Test_syftAdapter_CreateSBOM'   # that one test has a pre-existing, unrelated arm64-fixture mismatch (see TOP_ISSUES.md #1)
```

New/updated coverage:
- `core/services/scan_test.go::TestOptionsFromWorkload` — platform Args plumbing
- `core/services/scan_failure_reasons_test.go::TestClassifySBOMError` — typed + string-based platform-mismatch classification
- `pkg/sbomscanner/v1/server_test.go` — `TestIsPlatformMismatch`, `TestFormatResolvedPlatform`, `TestCreateSBOM_PlatformMismatch_LocalRegistry` (also a regression test for the pod-less fallback bug: `TestCreateSBOM_ImageTooLarge_LocalRegistry` now passes on an arm64 host without setting `Platform`)
- `adapters/v1/syft_test.go::TestFormatResolvedPlatform`

## Related issues/PRs

* Resolved #512
* Builds on #513

## Summary by CodeRabbit

- **New Features**
  - Request a specific image platform using OCI platform strings or architecture-only values.
  - SBOM results now report the platform that was actually resolved and used.
  - Platform details remain available across successful and failed scan results.

- **Bug Fixes**
  - Requests for unavailable platforms now receive a distinct, actionable “image not found” classification instead of a generic generation failure.

- **Documentation**
  - Added API guidance and examples for platform selection, resolved-platform annotations, and related failure reporting.


rootp1:
@matthyx please have a look whenever you have a chance to

matthyx (commented):
In `adapters/v1/syft.go` and `pkg/sbomscanner/v1/server.go`, `formatResolvedPlatform` returns partial strings such as `"/amd64"` when `os` is empty or `"linux/"` when `arch` is empty.

To adhere to the OCI platform spec (`os/arch[/variant]`) and avoid publishing malformed annotation values like `kubescape.io/resolved-platform: "/amd64"`, `formatResolvedPlatform` should return `""` unless both `os` and `arch` are present:

```go
func formatResolvedPlatform(os, arch, variant string) string {
	if os == "" || arch == "" {
		return ""
	}
	parts := []string{os, arch}
	if variant != "" {
		parts = append(parts, variant)
	}
	return strings.Join(parts, "/")
}
```

Please update the test expectations in `adapters/v1/syft_test.go` and `pkg/sbomscanner/v1/server_test.go` accordingly.

rootp1:
I updated the default platform handling, tightened the resolved-platform formatting, and refreshed the related tests. The focused regressions pass, and go test ./... still hits the existing arm64 Syft fixture mismatch in adapters/v1.

matthyx (dismissed):
Approving PR #530: the `formatResolvedPlatform` logic and test expectations have been updated to ensure complete `os/arch` pairs before returning a platform string.

matthyx (approved):

PR #531, opened 2026-08-09
docs+fix: document per-pod cache tradeoff and tighten exceptionsCache TTL

### LINKED ISSUE #528: In-process rate-limit and exceptions caches don't share state across replicas (the #438 follow-up that was never filed)
# Description

Kubevuln has three independent in-process, per-pod TTL caches that back correctness- and reliability-relevant behavior: the registry rate-limit backoff (`tooManyRequests`, core/services/scan.go), the merged CVE-exceptions/VEX-policy cache (`exceptionsCache`, adapters/v1/backend.go), and the raw SecurityException/ClusterSecurityException List() cache (`securityExceptionListCache`, repositories/apiserver.go). None of them share state across replicas - each pod in a multi-replica Kubevuln deployment keeps its own independent copy, with no coordination between them.

This is a known, explicitly-flagged gap, not a new discovery: #438 (closed, "Share Grype vulnerability DB cache across replicas") describes "a multi-replica Kubevuln deployment" as the real, intended-to-be-supported deployment topology, and its own "Additional context" section says: *"this repo already has an in-memory TTL cache pattern... used for the tooManyRequests limiter and (see #TBD) for CVE-exceptions lookups. This issue is a distinct, infra-facing cache... not exercised the same way."* That "(see #TBD)" follow-up was never filed. This issue is that follow-up.

# Environment

Not version-specific - architectural gap present on main, relevant whenever Kubevuln runs with more than one replica (the deployment topology #438 explicitly designs around).

# Steps To Reproduce

This is a cross-pod consistency problem, so reproducing it needs two Kubevuln replicas rather than a single-process test:

1. Run two Kubevuln pods (or two `ScanService`/`BackendAdapter`/`APIServerStore` instances in a test, to simulate two replicas) behind the same load-balanced scan-request path.
2. **Rate-limit cache**: send enough scan requests for the same image that pod A records a 429 in its `tooManyRequests` cache (`ttl = 10m`, core/services/scan.go:41). Immediately send a scan request for the same image that gets routed to pod B. Pod B has no record of the 429 and will attempt the pull again - the backoff this cache exists to provide only protects requests that happen to land on the same pod.
3. **Exceptions cache**: create a `SecurityException` that suppresses CVE-X for a workload, scan it via pod A (which now caches the merged policy for `exceptionsCacheTTL = 5m`, adapters/v1/backend.go:59). Delete the `SecurityException`. Scan the same workload again, this time routed to pod B (cache miss, sees the deletion immediately) versus pod A again (cache hit, still suppresses CVE-X for up to 5 more minutes). The two pods disagree on whether CVE-X is currently suppressed for the same workload, for up to the TTL window.
4. **SecurityException list cache**: same shape as (3) but at the raw List() layer (`securityExceptionListCacheTTL = 30s`, repositories/apiserver.go:53) - a shorter window, same problem.

# Expected behavior

Either these caches should share state across replicas (so a 429 recorded by one pod, or an exception change observed by one pod, is visible to all pods within a bounded, consistent window), or the actual per-pod consistency guarantee should be a deliberate, documented tradeoff - not an accidental byproduct of `cache.New()` being process-local with no comment anywhere noting the multi-replica implication.

# Actual Behavior

All three caches are constructed as plain in-memory maps (`github.com/akyoto/cache`) inside `NewScanService`, `NewBackendAdapter`, and `NewAPIServerStorage` respectively, with no shared backing store, no cross-pod invalidation, and no mention in `docs/ARCHITECTURE.md` of this being a bounded, accepted tradeoff. Two concrete consequences:

- **Rate-limit backoff gets weaker as you scale out.** The whole point of `tooManyRequests` is to stop hammering a registry that just rate-limited you. With N replicas splitting scan traffic, up to N pods can independently discover the *same* 429 before enough of them have individually cached it - the opposite of what you'd want when scaling up specifically to handle more scan volume against the same registries.
- **CVE suppression is inconsistent across pods for up to 5 minutes after a SecurityException changes.** The same workload can show a CVE as suppressed or not depending purely on which pod happened to handle that particular scan request - a real correctness/security-relevant inconsistency, not just a performance quirk.

# Additional context

**Why this is one issue, not three:** all three caches share the identical root cause (a process-local `cache.New()` instance with no cross-replica awareness) and the identical fix shape (either externalize the backing store, or make the per-pod-only behavior an explicit, bounded, documented tradeoff). Fixing them piecemeal would mean re-deriving the same design decision three times in three files.

**Affected files:**
- `core/services/scan.go` - `tooManyRequests` cache (`NewScanService`, `checkCreateSBOM`, `rateLimitCacheKey` usages)
- `adapters/v1/backend.go` - `exceptionsCache` (`NewBackendAdapter`, `GetCVEExceptions`)
- `repositories/apiserver.go` - `securityExceptionListCache` (`NewAPIServerStorage`, `listSecurityExceptions`, `listClusterSecurityExceptions`)
- `docs/ARCHITECTURE.md` - should document whichever tradeoff is chosen

**Acceptance criteria:**
- A documented decision (in `docs/ARCHITECTURE.md` or equivalent) on whether these caches are meant to be per-pod-eventually-consistent or fully shared, with the actual behavior matching that decision
- If shared: a mechanism (e.g. an external cache such as Redis, or a lighter-weight coordination primitive already available in-cluster) that lets a 429 or exception change observed by one pod become visible to others within a bounded window meaningfully shorter than today's worst case (10m / 5m / 30s of full pod-to-pod disagreement)
- If intentionally per-pod: the TTLs should be re-examined with that tradeoff in mind (e.g. is 5 minutes of possible cross-pod suppression disagreement actually acceptable for a security-relevant cache?), and it should be a documented, deliberate choice rather than silent
- No regression to single-replica behavior, which is unaffected either way

- Start with the highest-value case: `exceptionsCache`, since it's the one with actual correctness/security consequences (suppression disagreement), not just a performance-tuning knob like the rate-limit cache
- A shared cache doesn't need to be exotic - even a lightweight external KV store keyed the same way as today's in-memory cache, with the same TTLs, would close most of the gap without redesigning the caching logic itself
- `tooManyRequests` and `securityExceptionListCache` can follow the same pattern once the approach is validated on `exceptionsCache`

bhuvan-somisetty:
## Overview

`tooManyRequests`, `exceptionsCache`, and `securityExceptionListCache` are all process-local in-memory caches with no state shared across Kubevuln replicas - a gap #438 explicitly flagged as an intended follow-up ("see #TBD" in its own text) that never actually got filed until #528.

This documents that as a deliberate, bounded tradeoff rather than a silent gap, and tightens `exceptionsCacheTTL` from 5m to 1m - the one cache with actual correctness/security stakes, since a `SecurityException` change is invisible across pods until the TTL expires, meaning the same workload can show a CVE as suppressed or not depending purely on which pod handles a given scan request.

## Additional Information

- Added an "In-Process Caching" section to `docs/ARCHITECTURE.md` covering all three caches, their TTLs, and why this is an accepted tradeoff rather than an oversight - introducing a shared backing store is a real architectural change whose deployment-side wiring lives outside this repo, matching how #438 scoped the equivalent Grype-DB-cache decision.
- 1m brings `exceptionsCache`'s worst-case cross-pod disagreement window much closer to the underlying raw-List() cache's own 30s TTL (`repositories/apiserver.go`'s `securityExceptionListCache`), while still avoiding a network+CRD round trip on every container scan within a burst.
- `tooManyRequests` and `securityExceptionListCache` are intentionally left as-is in this PR, per #528's own proposed rollout: validate the approach on `exceptionsCache` first since it's the one with real correctness stakes, not just a performance cost.
- No test asserted the specific TTL value, so no test changes were needed; confirmed by grepping for `exceptionsCacheTTL` usage before making the change.

## How to Test

```
go build ./adapters/v1/...
go vet ./adapters/v1/...
go test ./adapters/v1/... -count=1
```

All pass. This is a config-value and documentation change, not new logic, so no new test was added - existing `adapters/v1` coverage (including the exceptions-cache tests) continues to pass unchanged with the new TTL.

## Related issues/PRs:

* Resolved #528

## Summary by CodeRabbit

- **Bug Fixes**
  - Reduced the exception cache refresh interval from five minutes to one minute, helping updates appear sooner across application instances.

- **Documentation**
  - Added documentation describing in-process caches, their refresh behavior, replication considerations, and tradeoffs.

### DIFF
--- a/adapters/v1/backend.go
+++ b/adapters/v1/backend.go
@@ -54,9 +54,16 @@ var _ ports.Platform = (*BackendAdapter)(nil)
 // exceptionsCacheTTL bounds how stale a cached exceptions/CRD merge can be before the next scan
 // re-fetches from the backend and cluster; exceptions rarely change within a scan burst, so this
 // avoids a network + CRD round trip on every single container scan.
+//
+// exceptionsCache is process-local (see #528): each Kubevuln replica keeps its own copy, so a
+// SecurityException change observed by one pod is not visible to the others until their own
+// cache entries expire. 1m (rather than the 5m this used to be) bounds that cross-pod worst case
+// to something closer to the underlying CRD-list cache's own 30s TTL
+// (repositories/apiserver.go's securityExceptionListCacheTTL), while still avoiding a
+// network+CRD round trip on every single container scan within a burst.
 const (
 	exceptionsCacheCleaningInterval = 1 * time.Minute
-	exceptionsCacheTTL              = 5 * time.Minute
+	exceptionsCacheTTL              = 1 * time.Minute
 )
 
 func NewBackendAdapter(accountID, apiServerRestURL, eventReceiverRestURL, accessKey string, seRepo ports.SecurityExceptionRepository) *BackendAdapter {
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -513,6 +513,27 @@ The central business logic component implementing the `ScanService` port.
 
 ---
 
+## In-Process Caching
+
+Kubevuln keeps three TTL caches in memory, all backed by `github.com/akyoto/cache`:
+
+| Cache | Location | TTL | Purpose |
+|-------|----------|-----|---------|
+| `tooManyRequests` | `core/services/scan.go` | 10m | Registry rate-limit (429) backoff, keyed by image reference |
+| `exceptionsCache` | `adapters/v1/backend.go` | 1m | Merged CVE-exceptions/VEX policy, keyed per workload |
+| `securityExceptionListCache` | `repositories/apiserver.go` | 30s | Raw `SecurityException`/`ClusterSecurityException` `List()` results, keyed per namespace/cluster |
+
+**These caches are process-local by design, not shared across replicas.** In a multi-replica Kubevuln deployment (see #438 for the equivalent discussion about the Grype vulnerability DB), each pod keeps an independent copy of each cache with no cross-pod coordination:
+
+- A 429 recorded by one pod's `tooManyRequests` cache is invisible to the others - they can each independently discover and cache the same backoff, so the mitigation this cache exists to provide only fully applies within a single pod.
+- A `SecurityException` change (create/update/delete) observed by one pod is not reflected on the others until their own cache entries naturally expire - the same workload can show a CVE as suppressed or not depending on which pod happens to handle a given scan request, for up to the relevant cache's TTL.
+
+This is an accepted tradeoff, not an oversight: introducing a shared backing store (e.g. an external KV store) is a real architectural change with its own operational cost, and the deployment-side wiring for anything like that lives outside this repo (see kubescape-operator/helm-charts), matching how #438 scoped the equivalent Grype-DB-cache decision. Given that, the TTLs above are chosen deliberately per cache's actual stakes: `securityExceptionListCache` and `exceptionsCache` are kept short (30s / 1m) because a stale read there is a suppression-correctness question, not just a performance one, while `tooManyRequests` can tolerate a longer window (10m) since its only cost is a redundant registry pull attempt.
+
+If this tradeoff ever needs to change (e.g. Kubevuln starts running at a replica count where the redundant-429 or suppression-disagreement window becomes a real operational problem), the right first step is `exceptionsCache`, since it's the one with actual correctness/security stakes rather than a pure performance cost.
+
+---
+
 ## Integration Points
 
 ### External System Integration


bhuvan-somisetty:
@matthyx small one, mostly docs plus a TTL tweak - PTAL whenever you get a chance. Happy to change anything you'd like.

matthyx (approved):
LGTM! The process-local cache tradeoffs and TTL adjustment for `exceptionsCache` (5m -> 1m) are well-reasoned, clearly documented in `docs/ARCHITECTURE.md`, and directly resolve #528.

PR #533, opened 2026-08-09
fix(adapters): share one fix-state predicate across suppression paths

### LINKED ISSUE #449: bug: ExpiredOnFix exceptions are applied in the stored manifest but expired in the backend report for CPE-matched CVEs

AnshulPatil2005:
## Overview

`ApplySecurityExceptions` and `DomainToArmo` computed "is this CVE fixed?" with different predicates: the first used `Fix.State == "fixed"`, the second used `len(Fix.Versions) != 0` or a CPE match whose `versionConstraint` contains `"<"`. Both feed `getCVEExceptionMatchCVENameFromList`, which skips an `ExpiredOnFix` policy when the CVE is considered fixed. For CPE-matched CVEs carrying no concrete fix version, the two disagreed, so the same exception was applied in the stored `VulnerabilityManifest` but expired in the report sent to the backend.

Both paths now share a single `hasKnownFix` predicate. A fix is known when the vulnerability carries concrete fix versions, when Grype reports the state as `fixed`, or when a CPE match bounds the vulnerable range from above. This is the union of the two previous predicates, so no CVE that either path already treated as fixed becomes unfixed.

## How to Test

```
go test ./adapters/v1/ -run 'TestHasKnownFix|TestExpiredOnFix'
```

`TestExpiredOnFixAgreesAcrossPaths` uses the `CVE-2024-13176` match from `adapters/v1/testdata/alpine-cve.format.json`, which has the reported shape: fix state `unknown`, no fix versions, and an upper-bounded CPE constraint. It fails on `main` with `storage and report paths disagree on whether the ExpiredOnFix exception applies` and passes with this change.

## Related issues/PRs:

* Resolved #449

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved vulnerability fix detection across security reports and exception handling.
  * Recognizes fixes identified by explicit versions, fixed status, or bounded version constraints.
  * Keeps fix recommendations and expiration behavior consistent across reporting paths.
  * Handles malformed vulnerability details without disrupting processing.

* **Tests**
  * Added coverage for fixed, unfixed, unknown, and malformed vulnerability scenarios.

### DIFF
--- a/adapters/v1/domain_to_armo.go
+++ b/adapters/v1/domain_to_armo.go
@@ -8,7 +8,6 @@ import (
 	"strings"
 
 	"github.com/Masterminds/semver/v3"
-	"github.com/anchore/grype/grype/match"
 	"github.com/anchore/syft/syft/source"
 	"github.com/armosec/armoapi-go/armotypes"
 	"github.com/armosec/armoapi-go/containerscan"
@@ -116,26 +115,11 @@ func DomainToArmo(ctx context.Context, grypeDocument v1beta1.GrypeDocument, vuln
 		// iterate over all vulnerabilities
 		for _, m := range grypeDocument.Matches {
 			var isFixed int
-			var version string
 			description := m.Vulnerability.Description
 			link := linkToVuln(m.Vulnerability.ID)
-			if len(m.Vulnerability.Fix.Versions) != 0 {
+			fixed, version := hasKnownFix(m)
+			if fixed {
 				isFixed = 1
-				version = suggestedVersion(m.Artifact.Version, m.Vulnerability.Fix.Versions)
-			} else {
-				// also check CPE matches
-				for _, detail := range m.MatchDetails {
-					var found match.CPEResult
-					err := json.Unmarshal(detail.Found, &found)
-					if err == nil {
-						// we assume that if a higher version is mentioned in the CPE, then it is fixed somewhere
-						if strings.Contains(found.VersionConstraint, "<") {
-							isFixed = 1
-							version = "unknown"
-							break
-						}
-					}
-				}
 			}
 			if description == "" && len(m.RelatedVulnerabilities) > 0 {
 				description = m.RelatedVulnerabilities[0].Description
--- a/adapters/v1/fixstate.go
+++ b/adapters/v1/fixstate.go
@@ -37,8 +37,14 @@ func hasKnownFix(m v1beta1.Match) (bool, string) {
 	if m.Vulnerability.Fix.State == fixStateFixed {
 		return true, unknownFixVersion
 	}
-	// no concrete version: fall back to CPE matches
+	// no concrete version: fall back to CPE matches. Only cpe-match details are read,
+	// since match.CPEResult is the shape Grype emits for those alone; a distro detail
+	// would also unmarshal into it (unknown fields are dropped) and its constraint would
+	// then be read as if it came from a CPE.
 	for _, detail := range m.MatchDetails {
+		if detail.Type != string(match.CPEMatch) {
+			continue
+		}
 		var found match.CPEResult
 		if err := json.Unmarshal(detail.Found, &found); err != nil {
 			continue
--- a/adapters/v1/fixstate.go
+++ b/adapters/v1/fixstate.go
@@ -1,0 +1,52 @@
+package v1
+
+import (
+	"encoding/json"
+	"strings"
+
+	"github.com/anchore/grype/grype/match"
+	"github.com/kubescape/storage/pkg/apis/softwarecomposition/v1beta1"
+)
+
+const (
+	// fixStateFixed is the Grype fix state indicating a fix has been released.
+	fixStateFixed = "fixed"
+	// unknownFixVersion is reported when a fix is known to exist but no concrete
+	// version can be derived from the match.
+	unknownFixVersion = "unknown"
+)
+
+// hasKnownFix reports whether a fix exists for a Grype match, along with the version
+// to suggest for it. The version is "unknown" when a fix exists but no concrete
+// version is available, and empty when no fix is known.
+//
+// A fix is considered known when any of the following hold:
+//  1. The vulnerability carries concrete fixed versions.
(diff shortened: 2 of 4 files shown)


bhuvan-somisetty:
Verified: built clean and ran `go test -v ./adapters/v1/... -run 'TestHasKnownFix|TestExpiredOnFix'` - all 8 subtests pass, including `TestHasKnownFixOnlyReadsCPEDetails` from the follow-up commit.

Confirmed the actual bug this closes goes both directions, not just the #449 direction: before this, `ApplySecurityExceptions` only trusted `Fix.State == "fixed"`, so a match with concrete `Fix.Versions` but a different state string would never expire an `ExpiredOnFix` exception in the manifest. `DomainToArmo` had the opposite gap - it trusted `Fix.Versions` and the CPE bound but never checked `Fix.State == "fixed"` on its own. `hasKnownFix` is the union of all three signals, so it's a strict improvement on both paths, not just a move.

Also checked that CodeRabbit's cpe-match type-filtering catch (any `MatchDetails` was being unmarshalled into `match.CPEResult`, so a non-CPE detail with a coincidental `versionConstraint` field could false-positive) was correctly closed in `750ffe9`- verified `match.CPEMatch` really is `"cpe-match"` in the vendored grype module, and the added `TestHasKnownFixOnlyReadsCPEDetails` exercises exactly that with `exact-direct-match`/`exact-indirect-match` detail types.

One thing worth a second pair of eyes before merge, not a blocker: `hasKnownFix`'s doc comment says the CPE bound "implies a fixed version exists somewhere above that range" -that's inherited from the original code's assumption, but it's still a heuristic (a `<` upper bound in a CPE constraint doesn't strictly guarantee a shipped fix, just that the vulnerable range is bounded). Might be worth a short note in the design doc for #449 that this is a known approximation, so a future reader doesn't mistake it for a Grype guarantee. Doesn't block this PR- it's carrying forward existing behavior, just consolidating where it lives.

LGTM otherwise - nice fix, and the regression tests pin the exact disagreement shape from #449 well.

matthyx (approved):
Reviewed PR 533:

- Verified  correctly unifies fix-state detection across  (storage path) and  (report path), resolving #449.
- Verified  detail type filtering prevents misinterpreting non-CPE match details.
- Verified test suite passes, including all unit and integration tests in .

No blockers found. LGTM and ready to merge!

PR #541, opened 2026-08-09
feat(securityexception): support per-vulnerability expiresAt override

### LINKED ISSUE #454: RFC: Per-vulnerability `expiresAt` on SecurityException
## Overview

Allow an optional `expiresAt` on each vulnerability entry, in addition to the optional document-level default that exists today. The change is additive and backward-compatible: documents that only use the document-level `expiresAt` continue to work unchanged.

## Problem

Risk-acceptance authoring surfaces commonly encode expiry per entry. Trivy's `.trivyignore.yaml` supports `expired_at` on each entry, and Snyk's `.snyk` policy files use `expires` on each ignore. Consolidating multiple accepted CVEs with different expiry dates into a single SecurityException today forces authors to either produce one document per vulnerability (fragmenting a naturally-grouped set of decisions) or lose per-entry expiry information when mapping their source-format entries into SecurityException.

## Solution

Add an optional `expiresAt` field on each vulnerability entry, matching the optional document-level `expiresAt` today. When present, it overrides the document-level `expiresAt` for that entry. When absent, the entry inherits the document-level `expiresAt` (existing behavior).

### Suggested shape

```yaml
# document-level default (existing): applies to all entries unless overridden
expiresAt: 2026-12-31

vulnerabilities:
  - id: CVE-2025-14505
    # inherits document-level expiresAt: 2026-12-31

  - id: CVE-2026-31808
    expiresAt: 2026-09-15    # NEW: per-entry override
```

### Runtime semantics

- An entry is considered expired when its effective `expiresAt` (per-entry if present, else document-level) is in the past.
- Expired entries no longer suppress the corresponding finding in Kubevuln's runtime output.
- Entries without any effective `expiresAt` (neither per-entry nor document-level) do not expire; behavior matches today's default when the document-level `expiresAt` is absent.

## Alternatives

**One document per vulnerability.** Fragmenting risk-accepted findings into one SecurityException per CVE lets authors keep unique expiry dates but explodes the number of Kubernetes custom resources and obscures the natural grouping of decisions made in the same review.

**Set document-level `expiresAt` to `min` of intended entry expiries.** Keep the current schema (document-level `expiresAt` only) and require authors to use the earliest intended expiry across entries. Rejected because it forces every entry to expire at the earliest chosen deadline; entries whose author-intended expiry would have been later are effectively cut short. Authors would still have to fragment into separate documents (see previous alternative).

**Encoding per-entry expiry in the free-text `impactStatement`.** Would work as a convention but is not machine-parsable, does not participate in runtime suppression decisions, and can silently drift out of sync as tooling evolves.

## Drawbacks

**Runtime consumers must resolve inheritance.** Kubevuln must compute the effective `expiresAt` for each entry (per-entry override or document-level fallback) rather than reading a single top-level field. Trivial to implement but requires a documented rule.

## References

- [SecurityException design document (kubescape/kubevuln, pinned to 85c8d7a)](https://github.com/kubescape/kubevuln/blob/85c8d7a9e42dd58a629f6378e023a090e157df98/docs/security-exception-design.md)
- [Trivy `.trivyignore.yaml` documentation](https://trivy.dev/docs/latest/guide/configuration/filtering/#trivyignoreyaml)

AnshulPatil2005:
## Overview

`expiresAt` could only be set for a whole SecurityException document. A set of CVEs accepted in the same review but with different deadlines had to be split across several CRDs, or collapsed onto the earliest deadline and cut short.

This adds an optional `expiresAt` to each vulnerability entry. An entry uses its own value when set and inherits `spec.expiresAt` otherwise, so expiry is decided per entry rather than per document:

```yaml
spec:
  expiresAt: "2026-12-31T00:00:00Z"   # document-level default
  vulnerabilities:
    - vulnerability:
        id: "CVE-2025-14505"
      status: "not_affected"           # inherits 2026-12-31

    - vulnerability:
        id: "CVE-2026-31808"
      status: "not_affected"
      expiresAt: "2026-09-15T00:00:00Z"  # overrides for this entry
```

Documents that only set `spec.expiresAt` are unaffected — every entry inherits it and expires exactly as before. The resolved expiry is also what lands on the policy's `ExpirationDate`, so a per-entry override is not dropped on the way to the backend.

## How to Test

```
go test ./adapters/v1/ -run 'TestConvertPerEntry|TestConvertSkipsExpired'
```

`TestConvertPerEntryExpiresAt` covers the inheritance matrix (entry overriding an expired or live document, entry inheriting either, and no expiry set anywhere). `TestConvertPerEntryExpiresAtWithinOneDocument` covers the case this enables: one document whose entries expire on different dates. The pre-existing `TestConvertSkipsExpired` is unchanged and still passes, pinning the document-level behavior.

## Related issues/PRs:

* Resolved #454

## Summary by CodeRabbit

* **New Features**
  * Added optional expiration dates for individual vulnerability exceptions.
  * Individual dates override the overall exception expiration; entries without one inherit it.
  * Expired entries are excluded independently, including within mixed and cluster-scoped exceptions.
  * Generated policies now reflect each entry’s effective expiration date.

* **Documentation**
  * Updated configuration examples and expiration guidance for per-entry overrides.

### DIFF
--- a/adapters/v1/securityexception.go
+++ b/adapters/v1/securityexception.go
@@ -37,15 +37,15 @@ func ConvertToVulnerabilityExceptionPolicies(exceptions []sev1beta1.SecurityExce
 
 	for i := range exceptions {
 		se := &exceptions[i]
-		if isExpired(se.Spec.ExpiresAt, now) {
-			continue
-		}
 		if !matchExceptionTarget(se.Spec.Match, target, false) {
 			continue
 		}
 		namespace := se.Namespace
 		for _, vuln := range se.Spec.Vulnerabilities {
 			if !shouldSuppress(vuln) {
+				continue
+			}
+			if isExpired(effectiveExpiresAt(se.Spec, vuln), now) {
 				continue
 			}
 			p := buildPolicy(se.Spec, vuln, namespace, suppressionSource{
@@ -59,14 +59,14 @@ func ConvertToVulnerabilityExceptionPolicies(exceptions []sev1beta1.SecurityExce
 
 	for i := range clusterExceptions {
 		cse := &clusterExceptions[i]
-		if isExpired(cse.Spec.ExpiresAt, now) {
-			continue
-		}
 		if !matchExceptionTarget(cse.Spec.Match, target, true) {
 			continue
 		}
 		for _, vuln := range cse.Spec.Vulnerabilities {
 			if !shouldSuppress(vuln) {
+				continue
+			}
+			if isExpired(effectiveExpiresAt(cse.Spec, vuln), now) {
 				continue
 			}
 			p := buildPolicy(cse.Spec, vuln, "", suppressionSource{
@@ -82,6 +82,19 @@ func ConvertToVulnerabilityExceptionPolicies(exceptions []sev1beta1.SecurityExce
 
 func isExpired(expiresAt *metav1.Time, now time.Time) bool {
 	return expiresAt != nil && expiresAt.Time.Before(now)
+}
+
+// effectiveExpiresAt resolves the expiry governing a single vulnerability entry: the
+// per-entry expiresAt when set, otherwise the document-level default.
+//
+// Expiry is therefore decided per entry rather than per document. A document-level
+// expiresAt still expires every entry that does not override it, so documents written
+// before per-entry expiry existed behave exactly as they did before.
+func effectiveExpiresAt(spec sev1beta1.SecurityExceptionSpec, vuln sev1beta1.VulnerabilityException) *metav1.Time {
+	if vuln.ExpiresAt != nil {
+		return vuln.ExpiresAt
+	}
+	return spec.ExpiresAt
 }
 
 // shouldSuppress reports whether a VulnerabilityException entry may be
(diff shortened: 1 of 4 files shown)


AnshulPatil2005:
on the cached-policy expiry point: leaving that out of this PR. the exceptions cache holds policies for the full window regardless of expiry, so a document-level `expiresAt` can already keep suppressing past its deadline on main — this change doesn't introduce it, and per-entry expiry inherits the same window. the design doc says expiry is a scan-time decision, so the cache does contradict it, but fixing that means revalidating `ExpirationDate` at apply time for cloud-sourced policies too. that's a separate behaviour change and belongs in its own issue rather than riding along here. happy to open one.

AnshulPatil2005:
note on merge order: this touches the same `VulnerabilityException` struct as #542 and #543, so whichever of the three lands second and third will need a rebase. there is no functional dependency between them and nothing has to merge first, the conflicts are all "keep both" additions: new fields on the same struct, and in some pairs test cases appended to the same end of file. i verified with trial merges. #533 is independent and merges cleanly against all three. happy to rebase as soon as the first one lands.

matthyx (approved):
LGTM! Implementation of per-vulnerability `expiresAt` override is clean, backward-compatible, and well-tested. Documentation and unit tests are complete, and all CI checks are passing.

PR #544, opened 2026-08-09
fix(services): report GenerateSBOM SBOM-status failures and add reason label to scan metrics

### LINKED ISSUE #540: Bug: GenerateSBOM silently reports success on timed-out/oversized SBOMs, and scan-failure metrics have no reason granularity across all four scan flows
# Description

Two closely related gaps in how scan failures are detected and observed, both in the same code path (`core/services/scan.go`'s scan flows → `controllers/http.go`'s outcome recording → `internal/metrics`):

**1. `GenerateSBOM` never checks for an `Incomplete`/`TooLarge` SBOM status, unlike every other scan flow.** `ScanCP` (`core/services/scan.go:332`), `ScanCVE` (`:577`), and `ScanRegistry` (`:742`) all explicitly check `sbom.Status == helpersv1.Incomplete || sbom.Status == helpersv1.TooLarge` right after SBOM creation/retrieval, call `ReportScanFailure` with the specific reason, and abort. `GenerateSBOM` (`:145-198`, the handler behind `SBOMCalculationCommandPath`) has no such check anywhere — it creates the SBOM, unconditionally stores it via `StoreSBOM`, and returns `nil`, even when the SBOM creator reported the scan as timed out or over the configured size limit. This isn't a design choice: `core/services/scan_test.go:64-69`'s own "phase 1, timeout" test case documents the regression in its own comment — `wantErr: false, // we no longer check for timeout` — meaning the check used to exist here too and was dropped while the other three flows kept it.

**2. Every scan flow collapses all failure reasons into a single opaque `outcome="error"` metric label, even though a bounded, 13-value failure-reason taxonomy (`scanfailure.Reason*` from `github.com/armosec/armoapi-go/scanfailure`) is already computed at every single failure site and handed to `ReportScanFailure` — then discarded before it reaches Prometheus.** `HTTPController.recordScan` (`controllers/http.go:65-75`) only accepts `outcome` (`"success"`/`"partial"`/`"error"`), and every call site (`:132`, `:194-207`, `:249`, `:336`) derives that 3-way value from a bare `err != nil` / `errors.Is` check, throwing away the specific reason `scan.go` already knew before returning. An operator with only `/metrics` (no access to the backend platform's `ReportScanFailure` stream or pod logs — the two places the reason currently *does* surface) cannot tell "registry auth is broken fleet-wide" from "images are hitting `maxImageSize`" from "the CVE DB match is failing" — they all look identical: one `kubevuln_scans_completed_total{outcome="error"}` count.

These interact directly: fixing (1) properly means `GenerateSBOM` needs to report a failure the same way the other three flows do, which is the natural place to also thread the already-computed reason through to the metric, instead of bolting on a fourth copy of the same coarse `outcome="error"` pattern.

# Environment
OS: N/A — structural gap in `core/services/scan.go`, `controllers/http.go`, `internal/metrics/metrics.go`, not version-specific
Version: present on `main`

# Steps To Reproduce

**Bug 1 (GenerateSBOM):**
1. Send a `SBOMCalculationCommandPath` request for an image that will produce a `helpersv1.Incomplete` or `helpersv1.TooLarge` SBOM status (e.g. one exceeding `MaxImageSize`/`MaxSBOMSize`, or one that times out mid-pull) — *without* a corresponding `ContainerScanCommandPath` request for the same image.
2. Observe the response: HTTP 200 (validation passed), and the background job completes with `err == nil`.
3. Check `kubevuln_scans_completed_total{endpoint="generateSBOM"}` — it increments `outcome="success"`.
4. Check the platform's failure-report stream / `ReportScanFailure` calls for this image — nothing was sent, because `GenerateSBOM` never reaches a `ReportScanFailure` call for this condition.
5. Compare with the exact same underlying SBOM status hitting `ScanCVE` instead (`core/services/scan.go:577`): it returns `domain.ErrIncompleteSBOM`, calls `ReportScanFailure`, and the metric records `outcome="error"` — the identical condition is silently "success" through one endpoint and a hard failure through the other three.

**Bug 2 (failure-reason metrics):**
1. Trigger any two different failures through the same endpoint, e.g. `ScanCVE` with an unreachable/auth-rejecting registry (`ReasonImageAuthFailed`) and `ScanCVE` with a corrupt/oversized image (`ReasonImageTooLarge` / `ReasonSBOMTooLarge`).
2. Check `/metrics`: both increment the exact same series, `kubevuln_scans_completed_total{endpoint="scanCVE",outcome="error"}` — indistinguishable from Prometheus alone, despite `classifySBOMError`/`classifySBOMStatusWithAnnotation` (`core/services/scan_failure_reasons.go`) having already computed the specific reason a few lines earlier at each call site.

# Expected behavior

- `GenerateSBOM` should treat an `Incomplete`/`TooLarge` SBOM exactly like `ScanCP`/`ScanCVE`/`ScanRegistry` do: report the failure via `ReportScanFailure` with the correct classified reason, and return an error (`domain.ErrIncompleteSBOM`) instead of silently storing the degraded SBOM and reporting success.
- The Prometheus `kubevuln_scans_completed_total` series (or a new series) should carry a bounded, low-cardinality `reason` label reflecting the actual `scanfailure.Reason*` value computed for the failure, for every endpoint — not just the coarse `success`/`partial`/`error` bucket — so failure-reason breakdowns are visible without needing platform-side telemetry or log access.

# Actual Behavior

- `GenerateSBOM` reports `outcome="success"` and no scan-failure telemetry for images that timed out or exceeded size limits, while the other three scan flows correctly treat the identical condition as a failure.
- All failures across all four scan endpoints collapse into `outcome="error"` with zero reason granularity in Prometheus, even though the specific reason is already computed and available at every failure site.

# Additional context

**Why this is one issue and not two trivial one-liners:** properly fixing (1) requires making `GenerateSBOM`'s failure observable the same way as the other three flows (not just adding a bare `return err`), and the cleanest way to do that without a fourth near-duplicate of the same `ReportScanFailure`-then-guess-a-coarse-outcome pattern is to also close (2) at the same time — otherwise this is the *fourth* place a maintainer would need to remember to keep three different signals (return error / `ReportScanFailure` reason / metric outcome) in sync by hand, which is exactly how (1) regressed in the first place per the stale test comment.

**Affected files:**
- `core/services/scan.go` — `GenerateSBOM` (missing status check), and the failure-reporting call sites across `ScanCP`/`ScanCVE`/`ScanRegistry`/`GenerateSBOM` that already compute a `scanfailure.Reason*` value right before returning
- `core/domain/scan.go` — sentinel errors (`ErrIncompleteSBOM`, `ErrTooManyRequests`, `ErrMissingSBOM`, etc. already exist and are already the mechanism `controllers/http.go` uses via `errors.Is`/`errors.As` to classify `ScanCP`'s `partial` outcome; the CVE-matching/store/upload failure paths currently return a bare `fmt.Errorf`-wrapped error with no equivalent sentinel to classify by)
- `controllers/http.go` — `recordScan`/`recordRejection` (coarse outcome only) and every `workerPool.Submit` call site that derives `outcome` from `err`
- `internal/metrics/metrics.go` — `ScanCounter` (no `reason` attribute today)
- `core/services/scan_test.go` — the stale "phase 1, timeout" case (`wantErr: false, // we no longer check for timeout`) needs correcting once the check is restored
- `core/services/scan_failure_reasons.go` — the existing `classifySBOMError`/`classifySBOMStatusWithAnnotation` classification this issue reuses, not modifies

**Acceptance criteria:**
- `GenerateSBOM` returns `domain.ErrIncompleteSBOM` and calls `ReportScanFailure` with the correct classified reason when the created/retrieved SBOM has `Status == Incomplete || Status == TooLarge`, mirroring `ScanCP`/`ScanCVE`/`ScanRegistry`; the degraded SBOM is not stored as if it were a successful scan (or if storage of the degraded artifact itself is intentional, the scan is still reported/metriced as a failure).
- The `core/services/scan_test.go` "phase 1, timeout" case is corrected to assert the failure is surfaced, and its stale comment is removed.
- A bounded, documented `reason` label (drawn from the existing `scanfailure.Reason*` constants, already a fixed 13-value enum, plus a default like `"none"` for successful scans) is attached to scan-outcome metrics for all four endpoints (`generateSBOM`, `scanCP`, `scanCVE`, `scanRegistry`), sourced from the same classification already computed in `core/services/scan.go` rather than re-derived in the controller.
- No behavior change for successful scans or for scans that were already correctly failing before this change (their HTTP response, `ReportScanFailure` payload, and non-reason metric labels are unaffected).
- New/updated tests cover: `GenerateSBOM` returning an error on `Incomplete`/`TooLarge` status (both storage-enabled and disabled), and the `reason` label being populated correctly for at least one failure case per endpoint.

- Add the missing `sbom.Status == helpersv1.Incomplete || sbom.Status == helpersv1.TooLarge` check to `GenerateSBOM`, following the exact pattern already used 3× elsewhere (`ReportScanFailure` + `return domain.ErrIncompleteSBOM`).
- Introduce a small, explicit way for `core/services/scan.go` to hand its already-computed `scanfailure.Reason*` value to the caller without changing every method's return signature — e.g. a typed error wrapper (`domain.ScanError{Reason string; Err error}` implementing `Unwrap`) that call sites already returning `err` can wrap once at the `ReportScanFailure` call sites, and `controllers/http.go` unwraps via `errors.As` in `recordScan`'s call sites instead of re-guessing from a bare `err`.
- Extend `HTTPController.recordScan` to accept the resolved reason (defaulting to `scanfailure.ReasonUnexpected` for a bare, unwrapped error, and a fixed `"none"`/absent value for success) and add it as a bounded `metric.WithAttributes` entry on `ScanCounter`.
- Update the stale `core/services/scan_test.go` case and add coverage for the new `GenerateSBOM` branch and reason propagation.

bhuvan-somisetty:
## Overview

Two related gaps in scan-failure detection/observability, both in the same code path (`core/services/scan.go`'s scan flows → `controllers/http.go`'s outcome recording → `internal/metrics`):

1. `GenerateSBOM` never checked for an `Incomplete`/`TooLarge` SBOM status, unlike `ScanCP`, `ScanCVE`, and `ScanRegistry`, which all check `sbom.Status == helpersv1.Incomplete || sbom.Status == helpersv1.TooLarge` right after SBOM creation/retrieval and abort with a `ReportScanFailure` call. A timed-out or oversized SBOM going through `GenerateSBOM` was silently stored via `StoreSBOM` and the scan reported as a success. `GenerateSBOM` now performs the same check, reports the failure with the classified reason, and returns `domain.ErrIncompleteSBOM` instead.
2. Every scan flow collapsed all failure reasons into a single opaque `outcome="error"` metric label, even though a bounded, 13-value failure-reason taxonomy (`scanfailure.Reason*`) was already computed at each failure site and handed to `ReportScanFailure`, then discarded before it reached Prometheus. An operator watching only `/metrics` could not distinguish "registry auth is broken fleet-wide" from "images are hitting `maxImageSize`" from "the CVE DB match is failing" — all three incremented the identical `kubevuln_scans_completed_total{outcome="error"}` series.

`ScanCP` is deliberately left unchanged: it loops over multiple images per call and `continue`s past a per-image failure instead of returning it, so there's no single dominant reason to attach to its own return value the way the other three flows have.

### Changes

- `core/domain/scan.go`: new `ScanError{Reason, Err}` type (implements `Unwrap`) that lets a scan-flow call site hand its already-computed `scanfailure.Reason*` classification to callers without changing every method's return signature. Existing `errors.Is`-based checks (e.g. `ErrPartialContainerProfile`, `ErrTooManyRequests`) keep working unchanged through `Unwrap`.
- `core/services/scan.go`:
  - `GenerateSBOM` now checks the SBOM status the same way `ScanCP`/`ScanCVE`/`ScanRegistry` do, and reports/returns a failure instead of silently storing a degraded SBOM as a success.
  - Every `return err` at a `ReportScanFailure` call site across `GenerateSBOM`, `ScanCVE`, and `ScanRegistry` now wraps the error in `&domain.ScanError{Reason: ..., Err: ...}`, using the reason already computed at that call site.
- `controllers/http.go`: `recordScan` now takes the scan's `error` and resolves a bounded `reason` attribute via `errors.As(err, *domain.ScanError)`, defaulting to `scanfailure.ReasonUnexpected` for an unclassified error and `"none"` for success/partial outcomes. The `reason` attribute is added to both the `ScanCounter` and `ScanDuration` instruments.
- `core/services/scan_test.go`: corrected the stale "phase 1, timeout" case (previously `wantErr: false, // we no longer check for timeout`, which documented this exact regression) to assert the failure is now surfaced, and added `wantReason` coverage to the `GenerateSBOM`, `ScanCVE`, and `ScanRegistry` table tests, asserting the exact `scanfailure.Reason*` value each failure case's `domain.ScanError` should carry.
- `controllers/http_test.go`: added `TestHTTPController_MetricsEndpoint_RecordsScanFailureReason`, which drives `generateSBOM`, `scanCVE`, and `scanRegistry` through a failing scan service and scrapes `/metrics` to assert the exact `kubevuln_scans_completed_total{endpoint,outcome="error",reason}` series is produced, including the unclassified-error fallback to `scanfailure.ReasonUnexpected`.
- `docs/API.md`: documented the new `reason` label on `kubevuln_scans_completed_total`/`kubevuln_scan_duration_seconds`.
- `docs/scan-failure-reporting.md`: noted the `GenerateSBOM` fix and cross-linked the new metric label to the existing failure-classification explanation.

No behavior change for successful scans or for scans that were already correctly failing before this change — their HTTP response, `ReportScanFailure` payload, and non-reason metric labels are unaffected.

## How to Test

- `go build ./...`
- `go test ./core/services/... ./controllers/... ./core/domain/... ./internal/metrics/...`
- `TestScanService_GenerateSBOM`/`TestScanService_ScanCVE`/`TestScanService_ScanRegistry` (`core/services/scan_test.go`) assert `domain.ScanError.Reason` for every classified failure case.
- `TestHTTPController_MetricsEndpoint_RecordsScanFailureReason` (`controllers/http_test.go`) proves the `reason` label reaches `/metrics` end to end for `generateSBOM`, `scanCVE`, and `scanRegistry`.
- To see it live: trigger any scan failure and check `/metrics` — `kubevuln_scans_completed_total` and `kubevuln_scan_duration_seconds` now carry a `reason` attribute (e.g. `reason="sbom_incomplete"`) alongside `endpoint`/`outcome`, instead of only the coarse `outcome="error"`.

## Related issues/PRs

Resolved #540

## Summary by CodeRabbit

* **New Features**
  * Scan failures now include classified reasons, such as rate limiting, incomplete SBOMs, and storage or scanning errors.
  * Incomplete or oversized SBOMs are reported as scan failures.
  * Scan metrics now include a bounded failure-reason label, including an `unexpected` fallback.

* **Documentation**
  * Updated metrics and scan-failure documentation to describe available reason codes and reporting behavior.

* **Tests**
  * Added coverage for classified, unclassified, timeout, and rate-limit failures across scan workflows.

### DIFF
--- a/controllers/http.go
+++ b/controllers/http.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	wssc "github.com/armosec/armoapi-go/apis"
+	"github.com/armosec/armoapi-go/scanfailure"
 	"github.com/gammazero/workerpool"
 	"github.com/gin-gonic/gin"
 	"github.com/kubescape/go-logger"
@@ -62,16 +63,36 @@ func (h *HTTPController) WithMetrics(m *metrics.Metrics) (*HTTPController, error
 // recordScan records the outcome and duration of a background scan job for the given endpoint.
 // outcome is one of "success", "partial", or "error" -- see the ScanCP call site, where a
 // domain.ErrPartialContainerProfile result is a warning-level expected outcome, not a failure.
-func (h HTTPController) recordScan(ctx context.Context, endpoint string, start time.Time, outcome string) {
+// err is the error returned by the scan (nil for success/partial) and is only consulted to
+// resolve the bounded reason label on a failed outcome; see scanFailureReason.
+func (h HTTPController) recordScan(ctx context.Context, endpoint string, start time.Time, outcome string, err error) {
 	if h.metrics == nil {
 		return
 	}
 	attrs := metric.WithAttributes(
 		attribute.String("endpoint", endpoint),
 		attribute.String("outcome", outcome),
+		attribute.String("reason", scanFailureReason(outcome, err)),
 	)
 	h.metrics.ScanCounter.Add(ctx, 1, attrs)
 	h.metrics.ScanDuration.Record(ctx, time.Since(start).Seconds(), attrs)
+}
+
+// scanFailureReason resolves the bounded scanfailure.Reason* label for a scan-outcome metric.
+// Successful/partial outcomes carry no specific failure reason ("none"). A failed outcome whose
+// error was classified via *domain.ScanError (set at the point of failure in
+// core/services/scan.go, alongside the equivalent ReportScanFailure call to the platform)
+// surfaces that reason; any other error defaults to scanfailure.ReasonUnexpected, keeping the
+// label's cardinality fixed regardless of what an unclassified error's message happens to say.
+func scanFailureReason(outcome string, err error) string {
+	if outcome != "error" {
+		return "none"
+	}
+	var scanErr *domain.ScanError
+	if errors.As(err, &scanErr) && scanErr.Reason != "" {
+		return scanErr.Reason
+	}
+	return scanfailure.ReasonUnexpected
 }
 
 // recordRejection records a validation-time rejection for the given endpoint. reason is
@@ -129,7 +150,7 @@ func (h HTTPController) GenerateSBOM(c *gin.Context) {
 		if err != nil {
 			outcome = "error"
 		}
-		h.recordScan(bgCtx, "generateSBOM", start, outcome)
+		h.recordScan(bgCtx, "generateSBOM", start, outcome, err)
 		if err != nil {
 			logger.L().Ctx(bgCtx).Error("service error - GenerateSBOM", helpers.Error(err),
 				helpers.String("imageSlug", newScan.ImageSlug),
@@ -191,20 +212,20 @@ func (h HTTPController) ScanCP(c *gin.Context) {
 		err = h.scanService.ScanCP(bgCtx)
 		if err != nil {
 			if errors.Is(err, domain.ErrPartialContainerProfile) {
-				h.recordScan(bgCtx, "scanCP", start, "partial")
+				h.recordScan(bgCtx, "scanCP", start, "partial", err)
 				logger.L().Ctx(bgCtx).Warning("service warning - ScanCP", helpers.Error(err),
 					helpers.String("wlid", newScan.Wlid),
 					helpers.String("name", name),
 					helpers.String("namespace", namespace))
 			} else {
-				h.recordScan(bgCtx, "scanCP", start, "error")
+				h.recordScan(bgCtx, "scanCP", start, "error", err)
 				logger.L().Ctx(bgCtx).Error("service error - ScanCP", helpers.Error(err),
 					helpers.String("wlid", newScan.Wlid),
 					helpers.String("name", name),
 					helpers.String("namespace", namespace))
 			}
 		} else {
-			h.recordScan(bgCtx, "scanCP", start, "success")
+			h.recordScan(bgCtx, "scanCP", start, "success", nil)
 		}
 	})
 }
@@ -246,7 +267,7 @@ func (h HTTPController) ScanCVE(c *gin.Context) {
 		if err != nil {
 			outcome = "error"
 		}
-		h.recordScan(bgCtx, "scanCVE", start, outcome)
+		h.recordScan(bgCtx, "scanCVE", start, outcome, err)
 		if err != nil {
 			logger.L().Ctx(bgCtx).Error("service error - ScanCVE", helpers.Error(err),
 				helpers.String("wlid", newScan.Wlid),
@@ -333,7 +354,7 @@ func (h HTTPController) ScanRegistry(c *gin.Context) {
 		if err != nil {
 			outcome = "error"
 		}
-		h.recordScan(bgCtx, "scanRegistry", start, outcome)
+		h.recordScan(bgCtx, "scanRegistry", start, outcome, err)
 		if err != nil {
 			logger.L().Ctx(bgCtx).Error("service error - ScanRegistry", helpers.Error(err),
 				helpers.String("imageSlug", newScan.ImageSlug),
--- a/controllers/http_test.go
+++ b/controllers/http_test.go
@@ -721,6 +721,8 @@ func TestHTTPController_MetricsEndpoint_RecordsScanFailureReason(t *testing.T) {
 			body := w.Body.String()
 			wantSeries := `kubevuln_scans_completed_total{endpoint="` + tt.endpoint + `",outcome="error",reason="` + tt.wantReason + `"} 1`
 			assert.True(t, strings.Contains(body, wantSeries), body)
+			wantDurationSeries := `kubevuln_scan_duration_seconds_count{endpoint="` + tt.endpoint + `",outcome="error",reason="` + tt.wantReason + `"} 1`
+			assert.True(t, strings.Contains(body, wantDurationSeries), body)
 		})
 	}
 }
--- a/controllers/http_test.go
+++ b/controllers/http_test.go
@@ -2,6 +2,7 @@ package controllers
 
 import (
 	"context"
+	"errors"
 	"net/http"
 	"net/http/httptest"
 	"os"
@@ -10,6 +11,7 @@ import (
 	"time"
 
 	wssc "github.com/armosec/armoapi-go/apis"
+	"github.com/armosec/armoapi-go/scanfailure"
 	"github.com/docker/docker/api/types/registry"
 	"github.com/gammazero/workerpool"
 	"github.com/gin-gonic/gin"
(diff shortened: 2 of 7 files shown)


bhuvan-somisetty:
Addressed all four CodeRabbit comments in a534a91:

- added the duration-metric assertion alongside the counter one in the HTTP metrics test
- the GenerateSBOM test loop now checks `wantReason` for every case, not just the timeout one createSBOM/rate-limit/storage failures were passing without their reason actually being verified
- added the `ErrIncompleteSBOM` check to the ScanCVE/ScanRegistry timeout cases too, matching what GenerateSBOM already did
- fixed the doc wording  `recordScan` reads the reason off the `ScanError` the scan services already built, it doesn't wrap the error itself

Ran the affected tests locally (`TestScanService_GenerateSBOM`, `TestScanService_ScanCVE`, `TestScanService_ScanRegistry`, `TestHTTPController_MetricsEndpoint_RecordsScanFailureReason`) all green. Pushed.

matthyx (approved):
LGTM! Approving PR #544.
-  wrapper in  properly preserves error chain via  while surfacing the classified .
- Metric recording in  correctly extracts  for failed outcomes and defaults to  /  cleanly.
- Unit test coverage in  and  is thorough and passes.
- Documentation in  and  updated.

PR #537, opened 2026-08-09
feat(adapters): emit K8s Events for suppressed CVEs on SecurityExceptions

bhuvan-somisetty:
## Overview

Suppression provenance - which `SecurityException`/`ClusterSecurityException` caused a CVE to be filtered out of scan results - has been tracked in memory since #495, but it only ever reached a log line. There was no `EventRecorder` anywhere in this codebase, so `kubectl describe securityexception` never showed suppression activity, even though `docs/security-exception-design.md` already documents that as the intended behavior.

This wires that up end to end: the CRD object's UID is now threaded through `suppressionSource` (needed to build a correct `ObjectReference`), `ApplySecurityExceptions`/`logSuppression` accept an optional `record.EventRecorder` and emit a `VulnerabilitySuppressed` Event alongside the existing log line when one is configured, and `cmd/http/main.go` constructs a real recorder from a typed K8s clientset, gated behind the same `storage && riskAcceptance` condition SecurityException CRD integration itself already requires.

## Additional Information

- `adapters/v1/securityexception.go`: added `uid types.UID` to `suppressionSource`, populated from `se.UID`/`cse.UID` in `ConvertToVulnerabilityExceptionPolicies`, surfaced as `attrs["sourceUID"]` in `buildSuppressionAttributes` alongside the existing `sourceKind`/`sourceNamespace`. `ApplySecurityExceptions`/`logSuppression` now take an optional `recorder record.EventRecorder`; a new `emitSuppressionEvent` builds a `*corev1.ObjectReference{Kind, APIVersion: "kubescape.io/v1beta1", Name, Namespace, UID}` from those attributes and calls `recorder.Eventf(ref, corev1.EventTypeNormal, "VulnerabilitySuppressed", ...)`. No live `Get` or scheme registration needed - `client-go/tools/reference.GetReference` short-circuits on a pre-built `*corev1.ObjectReference` anyway, and `kubescape.io/v1beta1` matches the GVR already used for these CRDs in `repositories/apiserver.go`.
- `core/services/scan.go`: added `ScanService.eventRecorder` + `SetEventRecorder`, following the exact same optional, nil-safe attachment pattern already used for `ScanService.metrics`/`SetMetrics` - existing callers and the many test call sites are unaffected, and event recording is a no-op until `SetEventRecorder` is called. `applyExceptionsToManifest` now passes `s.eventRecorder` through to `ApplySecurityExceptions`.
[...]

PR #553, opened 2026-08-10
feat(metrics): add Prometheus metrics for SecurityException matched/expired/active exceptions

### LINKED ISSUE #548: SecurityException suppressions have no Prometheus metrics: matched/expired/active exceptions are unobservable outside logs and Events
## Overview

`docs/security-exception-design-review.md` (point 10, "Missing: Observability and Dry-Run Mode") explicitly flags this as a gap and defers it:

> Metrics (Prometheus) for number of active exceptions, matched findings, expired exceptions

\#536/#537 closed the sibling gap — suppressions now emit a `VulnerabilitySuppressed` Kubernetes Event — but the metrics half of the same design-review point was never picked up, and nothing else in the codebase or issue tracker covers it. `internal/metrics/metrics.go` has a counter for *degraded* exception fetches (`ExceptionsDegradedCounter`, added for #507) but none for the exceptions pipeline's actual behavior: how many CVEs it suppresses, how many exceptions it skips because they've expired, or how many are currently in force.

This is a real observability blind spot, not just a nice-to-have: `ConvertToVulnerabilityExceptionPolicies` (`adapters/v1/securityexception.go:33-81`) silently drops any `SecurityException`/`ClusterSecurityException` whose `expiresAt` has passed (`isExpired`, called at lines 40 and 62) — no log, no event, no metric, nothing. An operator has no way to notice "12 exceptions expired this week and their CVEs are back in scope" short of diffing scan results by hand. Suppressions themselves fare slightly better — `logSuppression` (`adapters/v1/securityexception.go:258-290`) logs every suppressed CVE and, since #537, emits an Event — but there is still no counter, so answering "how many findings are currently suppressed fleet-wide, and is that number trending up" requires scraping logs/Events rather than reading `/metrics` like every other kubevuln signal.

## Problem

Three numbers a security team would reasonably want on a dashboard are computed (or computable) at existing call sites, but none of them reach Prometheus:

1. **Matched findings** — every CVE suppression already flows through `logSuppression`/`ApplySecurityExceptions` (`adapters/v1/securityexception.go:216-238`), and per-CVE provenance (`sourceKind`, `ruleId`, `sourceNamespace`) is already computed in `buildSuppressionAttributes` (`adapters/v1/securityexception.go:149-167`) for the log line and the `VulnerabilitySuppressed` Event. It stops there.
2. **Expired exceptions** — `isExpired` (`adapters/v1/securityexception.go:83-85`) silently filters out any exception whose `expiresAt` has passed, in both the namespaced and cluster-scoped loops of `ConvertToVulnerabilityExceptionPolicies`. There is no signal anywhere — not even a log line — when this happens.
3. **Active exceptions** — the number of exceptions that actually produced a policy this scan (survived expiry + target matching + `shouldSuppress`) is exactly `len(policies)` at the end of `ConvertToVulnerabilityExceptionPolicies`, but the caller (`ConvertToVulnerabilityExceptionPolicies(seList, cseList, target)` at `adapters/v1/backend.go:218`) only keeps the slice, not a count.

None of `ConvertToVulnerabilityExceptionPolicies`, `ApplySecurityExceptions`, or `BackendAdapter.GetCVEExceptions` (`adapters/v1/backend.go`) currently has access to `*metrics.Metrics` — unlike `ScanService`, which already carries `s.metrics` and increments `ExceptionsDegradedCounter` from `applyExceptionsToManifest` (`core/services/scan.go:858-885`). `record.EventRecorder` needed the same kind of plumbing in #537 (`ApplySecurityExceptions`/`logSuppression` grew an optional `recorder` parameter, `ScanService` grew `eventRecorder`/`SetEventRecorder`); metrics need an equivalent, separate threading pass, since `adapters/v1/securityexception.go` is pure/dependency-free by design and `BackendAdapter` has no metrics field today.

## Solution

Add three instruments to `internal/metrics/metrics.go`, following the exact registration pattern `ExceptionsDegradedCounter`/`ScanCounter` already use:

- `kubevuln_exceptions_matched_total` (counter, labels `sourceKind` = `SecurityException`/`ClusterSecurityException`) — incremented once per suppressed CVE, at the same point `logSuppression` already fires.
- `kubevuln_exceptions_expired_total` (counter, labels `sourceKind`) — incremented once per exception skipped by `isExpired` in `ConvertToVulnerabilityExceptionPolicies`.
- `kubevuln_exceptions_active` (gauge) — set to `len(policies)` at the end of `ConvertToVulnerabilityExceptionPolicies`, i.e. the count of exceptions that survived expiry/target-matching/`shouldSuppress` and are in force for the current call.

**Rough implementation shape**, mirroring how #537 threaded `record.EventRecorder` through the same functions:

- `ConvertToVulnerabilityExceptionPolicies` and `ApplySecurityExceptions` need a way to report these three numbers to a caller that owns `*metrics.Metrics`, without making `adapters/v1/securityexception.go` depend on the metrics package directly (it currently has zero external dependencies beyond `armotypes`/`storage` types). A small returned counts struct (e.g. `ConversionStats{Active, Expired int}` alongside the existing `[]armotypes.VulnerabilityExceptionPolicy` return, and a `SuppressionStats` return from `ApplySecurityExceptions`) keeps that separation, with `BackendAdapter.GetCVEExceptions` (`adapters/v1/backend.go:216-229`) and `ScanService.applyExceptionsToManifest` (`core/services/scan.go:858-885`) as the two places that actually hold `*metrics.Metrics` and can record them — the same shape `ExceptionsDegradedCounter` already uses at line 864-865.
- `BackendAdapter` gains a `metrics *metrics.Metrics` field (optional/nil-safe, matching every other optional dependency in this codebase — `eventRecorder`, `s.metrics` on `ScanService` itself), set via a constructor param or a `SetMetrics`-style method.

**Acceptance criteria:**

- `kubevuln_exceptions_matched_total`, `kubevuln_exceptions_expired_total`, and `kubevuln_exceptions_active` appear on `/metrics`, documented in `docs/API.md`'s Metrics table alongside the existing instruments.
- A suppressed CVE increments `kubevuln_exceptions_matched_total` with the correct `sourceKind` label; this must stay in sync with what `logSuppression`/the `VulnerabilitySuppressed` Event already report for the same suppression, since these three observability channels describe the same event and must not drift.
- An exception past its `expiresAt` increments `kubevuln_exceptions_expired_total` and does not count toward `kubevuln_exceptions_active`.
- No behavior change to `ApplySecurityExceptions`'s existing suppression logic, `ConvertToVulnerabilityExceptionPolicies`'s existing filtering, or any existing test — this is additive observability only, the same guarantee #537 held for Event emission.
- Nil `*metrics.Metrics` (or nil `BackendAdapter.metrics`) is a complete no-op, matching the established optional-dependency pattern in this codebase.

## Alternatives

- **Log-scraping / Event-watching instead of metrics.** Both channels already exist (`logSuppression`'s log line, the `VulnerabilitySuppressed` Event since #537), so this is technically possible today. Rejected as the primary answer because it doesn't give a cheap, cardinality-bounded time series for dashboards/alerting the way every other kubevuln signal (`kubevuln_scans_completed_total`, `kubevuln_scan_rejections_total`) already does — an operator would need a log pipeline or Event watcher just for this one subsystem.
- **A single combined counter with an `event` label (`matched`/`expired`) instead of two counters plus a gauge.** Would cut one instrument, but conflates a counter (monotonic, rate-queryable) with what's conceptually a point-in-time count (`active`), and loses the ability to alert on `matched_total` and `expired_total` independently without a label-based query, which is worse ergonomics for the common case ("show suppression rate") for a marginal reduction in instrument count.
- **Computing `active` as an async `ObservableGauge` sampled independently of a scan** (like `kubevuln_worker_pool_queue_depth` already does), instead of setting it synchronously from `ConvertToVulnerabilityExceptionPolicies`. Rejected because "active" here means "exceptions that matched *this* scan's target," which is inherently per-call, not a standing background value kubevuln could sample out of band without redoing the same target-matching work.

## Related

- #495 (suppression provenance tracking), #536/#537 (Events for suppressions) cover the same design-review point's Events half; this issue is its Prometheus-metrics half.
- #507 (`ExceptionsDegradedCounter`) is the closest prior art for exceptions-pipeline metrics in this codebase.
- `docs/security-exception-design-review.md`, point 10.

bhuvan-somisetty:
## Overview

SecurityException/ClusterSecurityException suppressions had no Prometheus signal. `kubevuln_exceptions_degraded_total` covers a failed exception fetch, but nothing covered how many CVE findings are actually suppressed, how many CRDs get skipped as expired, or how many exceptions are in force for a given scan. This was a known, explicitly deferred gap: `docs/security-exception-design-review.md` (point 10, "Missing: Observability and Dry-Run Mode") calls out "Metrics (Prometheus) for number of active exceptions, matched findings, expired exceptions" and defers it as a future enhancement. No open issue or PR covered it before now.

Concretely, `ConvertToVulnerabilityExceptionPolicies` (`adapters/v1/securityexception.go`) silently drops any `SecurityException`/`ClusterSecurityException` whose `expiresAt` has passed — no log, no event, no metric. And every CVE suppression already flows through `logSuppression`/`ApplySecurityExceptions`, but stopped at a log line (and, once #537 lands, a Kubernetes Event) — there was no counter, so "how many findings are currently suppressed fleet-wide, and is that trending up" required scraping logs/Events instead of reading `/metrics` like every other kubevuln signal.

## Changes

- `internal/metrics/metrics.go`: three new instruments, following the exact registration pattern `ExceptionsDegradedCounter` already uses:
  - `kubevuln_exceptions_matched_total` (counter, `sourceKind` label — `SecurityException`/`ClusterSecurityException`)
  - `kubevuln_exceptions_expired_total` (counter, `sourceKind` label)
  - `kubevuln_exceptions_active` (gauge)
- `core/domain/cve.go`: new `ExceptionStats{ExpiredBySource map[string]int}` type, so the CRD-conversion layer can report counts without depending on the metrics package.
- `adapters/v1/securityexception.go`:
  - `ConvertToVulnerabilityExceptionPolicies` now returns `(policies, ExceptionStats)`, counting each `isExpired` skip by `sourceKind` instead of silently dropping it.
  - `ApplySecurityExceptions` now returns a `map[string]int` of suppressions by `sourceKind`, computed at the same point `logSuppression` already fires.
- `core/ports/providers.go`: `Platform.GetCVEExceptions` grows a `domain.ExceptionStats` return value (best-effort — a cache hit or CRD-less call returns the zero value, not an error).
- `adapters/v1/backend.go` / `adapters/mockplatform.go`: both `GetCVEExceptions` implementations forward the stats `ConvertToVulnerabilityExceptionPolicies` computes.
- `core/services/scan.go`: `applyExceptionsToManifest` — the sole caller, already holding `*metrics.Metrics` for `ExceptionsDegradedCounter` — records all three new instruments: expired counts from `GetCVEExceptions`'s stats, `len(exceptions)` as the active gauge, and the matched-by-source map `ApplySecurityExceptions` returns.
- `docs/API.md`: documented the four exceptions-pipeline instruments in the Metrics table.
- `docs/security-exception-design-review.md`: updated point 10's resolution note — Prometheus metrics are no longer deferred.

No behavior change to suppression logic, CRD-conversion filtering, or the exceptions returned to callers — this is additive observability only, the same guarantee #537 held for Event emission.

## How to Test

- `go build ./...`
- `go test ./core/services/... ./adapters/... ./core/domain/... ./core/ports/... ./internal/metrics/... ./cmd/...` — all pass
- New coverage:
  - `TestConvertSkipsExpired` (`adapters/v1/securityexception_test.go`) asserts `ExceptionStats.ExpiredBySource` counts one expired `SecurityException` and one expired `ClusterSecurityException`.
  - `TestApplySecurityExceptions_MovesToIgnored` asserts the returned `matchedBySource` map.
  - `TestGetCVEExceptions_MergesCRDExceptions` (`adapters/v1/backend_test.go`) asserts stats flow through `BackendAdapter`.
  - `TestApplyExceptionsToManifest_RecordsExpiredMetric` and `TestApplyExceptionsToManifest_RecordsActiveAndMatchedMetrics` (`core/services/scan_metrics_test.go`) scrape `/metrics` end to end and assert the exact `kubevuln_exceptions_expired_total{sourceKind=...}`, `kubevuln_exceptions_active`, and `kubevuln_exceptions_matched_total{sourceKind=...}` series.
- To see it live: trigger a scan with an expired or suppressing `SecurityException`/`ClusterSecurityException` and check `/metrics`.

## Related issues/PRs

Resolved #548

## Summary by CodeRabbit

- **New Features**
  - Added monitoring for active, expired, suppressed, and degraded CVE exceptions.
  - Exception metrics include source-type details for clearer reporting.
  - Added statistics for expired exception policies and suppressed findings.

- **Bug Fixes**
  - Exception caches now expire with the earliest applicable policy.
  - Already-expired policies are no longer cached.
  - Suppression counts avoid double-counting findings affected by duplicate policies.

- **Documentation**
  - Documented the new Prometheus metrics and their availability at `/metrics`.


bhuvan-somisetty:
Rebased on main and sorted out the conflicts — mostly from #537 (Events) and #541 (per-vuln expiresAt) landing after I branched. Also had to fix two calls in fixstate_test.go that were missing the recorder arg #537 added, which was actually breaking main's build already, unrelated to this PR. Build and full test suite are green now.

bhuvan-somisetty:
Addressed CodeRabbit's three findings: the exceptions cache could serve a CRD policy past its own expiresAt (bounded the cache entry's TTL to the earliest expiry now), the active gauge wasn't resetting to zero on a scan with no exceptions (moved the record above the early return), and two policies suppressing the same finding were double-counted as two matches instead of one (dedupe by sourceKind per finding now). Added a regression test for each. Build and tests are green.

bhuvan-somisetty:
Fixed the last CodeRabbit finding too — cacheTTLFor could return 0 for a policy that was already expired by the time it got cached, and since the cache library only sweeps expired entries on its cleanup interval rather than instantly, that 0-TTL entry could still be served as a hit in the meantime. Now skipping the cache write entirely when the TTL isn't positive, with a test covering it.

Also rebased on main again to pick up #546's new fallback/source-resolution metrics — just an additive conflict in metrics.go, both sets of instruments now sit side by side. Build and full test suite are green.

bhuvan-somisetty:
@coderabbitai review

bhuvan-somisetty:
@matthyx PTAL whenever u get chance.

matthyx (approved):
LGTM! The Prometheus metrics implementation for SecurityException / ClusterSecurityException (matched, expired, active, degraded) is clean, properly tested, and handles cache TTL bounding, gauge resets, and deduplication as expected. All CI checks pass.

Issue #484, opened 2026-08-07
Bug: ScanRegistry bypasses SecurityException filtering, in-cluster storage, and VEX generation


bhuvan-somisetty:
# Description

In `core/services/scan.go`, `ScanRegistry` implements the image scanning pipeline for container registry scans. However, unlike `ScanCVE` and `ScanCP`, `ScanRegistry` currently bypasses the storage and security exception subsystems entirely:

1. **Bypasses SecurityException Filtering:** `ScanRegistry` does not call `s.applyExceptionsToManifest(ctx, cve)`. Any active `SecurityException` or `ClusterSecurityException` policies in the cluster are ignored during registry scans, sending raw unfiltered vulnerabilities to `s.platform.SubmitCVE`.
2. **Bypasses In-Cluster Storage (`s.storage`):** `ScanRegistry` never checks `s.storage` (`GetCVE`/`GetSBOM`) for pre-existing cached manifests, and never persists `VulnerabilityManifest`, `VulnerabilityManifestSummary`, `SBOM`, or `VEX` objects into Kubernetes storage when `s.storage` is `true`.
3. **Bypasses VEX Generation:** `ScanRegistry` does not evaluate `s.vexGeneration` or generate/store VEX documents.

This creates significant data inconsistency between registry scans and workload/image scans, exposes ignored/suppressed vulnerabilities in registry scan reports, and wastes CPU/network resources by re-scanning images whose SBOMs and CVE manifests already exist in storage.

### Affected Components
- `core/services/scan.go` (`ScanRegistry` pipeline)
- `core/services/scan_test.go` (`ScanRegistry` unit test suite)
- `adapters/v1/backend.go` & `adapters/v1/backend_utils.go`

# Environment
OS: `Linux / Kubernetes 1.28+`
Version: `v0.2.x / main`

# Steps To Reproduce

1. Deploy a `SecurityException` policy targeting a specific CVE (e.g. `CVE-2023-XXXX`).
2. Trigger a container registry scan for an image containing `CVE-2023-XXXX` via `ScanRegistry`.
3. Inspect the backend report submitted by `ScanRegistry` and the in-cluster storage resources (`VulnerabilityManifest`, `VulnerabilityManifestSummary`).
4. Observe that:
   - `CVE-2023-XXXX` is included in `SubmitCVE` despite an active `SecurityException` policy.
   - No `VulnerabilityManifest` or `VulnerabilityManifestSummary` resources are created or updated in storage.

# Expected behavior

[...]

Issue #501, opened 2026-08-07
Bug: ScanCP passes unfiltered manifests to StoreVEX and omits VEX updates on cached exception re-evaluation


bhuvan-somisetty:
# Description
In `core/services/scan.go`, `ScanCP` (container profile scan flow) exhibits two critical defects when `vexGeneration` is enabled:

1. **Unfiltered Manifests Passed to `StoreVEX` on New Scans**: When generating OpenVEX documents during a new scan (line 424), `ScanCP` passes `cve` (unfiltered main manifest) and `cvep` (unfiltered relevant manifest) to `s.cveRepository.StoreVEX(ctx, cve, cvep, true)`. Because `cvep` still contains matches that were moved to `IgnoredMatches` by `SecurityException` filtering, `markRelevantVulnerabilitiesAsAffectedInVex` erroneously marks suppressed vulnerabilities as `status: affected` with remediation action statements in the generated VEX document.
2. **Missing VEX Synchronization on Cached Exception Re-evaluation**: When a container profile scan hits the CVE cache (`cve.Content != nil`), `ScanCP` re-evaluates security exceptions against the cached manifest to update `StoreCVE` and `StoreCVESummary` when exceptions change (lines 343–370). However, `ScanCP` completely omits calling `StoreVEX` during cached exception re-evaluation. As a result, when security exceptions are added, deleted, or modified, the stored OpenVEX document is never updated to reflect the new triaged state.

# Environment
OS: `Linux / Kubernetes v1.28+`
Version: `v0.0.170+`

# Steps To Reproduce
1. Configure `kubevuln` with `storage: true` and `vexGeneration: true`.
2. Apply a `SecurityException` policy suppressing a vulnerability (e.g. `CVE-2023-1234`) present in a workload container.
3. Trigger a container profile scan (`ScanCP`).
4. Inspect the generated `OpenVulnerabilityExchangeContainer` CR in Kubernetes (`kubectl get openvulnerabilityexchangecontainers`).
5. Observe that `CVE-2023-1234` is marked with `status: affected` instead of `status: not_affected` because unfiltered `cvep` was passed to `StoreVEX`.
6. Add or delete a `SecurityException` policy and re-run `ScanCP` (hitting the CVE cache).
7. Observe that the stored VEX document is not updated at all.

# Expected behavior
[...]

PR #559, opened 2026-08-10
fix(services): generate VEX on every CVE-scan path, not two of five

### LINKED ISSUE #557: Bug: ScanCVE never generates VEX documents, and ScanRegistry's cache-hit path still doesn't either
# Description

`ScanCVE` — the handler behind the plain per-container `ContainerScanCommand` HTTP route (`controllers/http.go:234`, `POST /v1/scanImage`) and one of the three CVE-scan entry points in `core/services/scan.go` — never generates or stores a VEX document, regardless of the `vexGeneration` config flag (`config/config.go:66`, `mapstructure:"vexGeneration"`). `ScanCP` and `ScanRegistry` both call `s.cveRepository.StoreVEX` on their respective scan paths; `ScanCVE` has no such call anywhere in its ~175-line body. Separately, `ScanRegistry`'s own cache-hit branch has fallen out of sync with the comment directly above it, which explicitly says `StoreVEX` runs there — it doesn't.

`docs/ARCHITECTURE.md`'s "CVE Scanning Pipeline" diagram (lines ~344-366) shows `Generate VEX` as a standard step of CVE scanning in general, not something exclusive to two of the three scan flows. An operator who sets `vexGeneration: true` has no way to know, short of reading `core/services/scan.go` line by line, that one whole scan flow silently skips it.

This is not a duplicate of the closed #484 ("ScanRegistry bypasses SecurityException filtering, in-cluster storage, and VEX generation") or #501 ("ScanCP passes unfiltered manifests to StoreVEX and omits VEX updates on cached exception re-evaluation") — both were fixed for the flows/branches they covered (`ScanRegistry`'s cache-miss path, `ScanCP`'s both paths). Neither touched `ScanCVE`, and #484's fix appears to have only reached `ScanRegistry`'s cache-miss branch, leaving its cache-hit branch as an unfixed corner of that same original bug.

# Environment
OS: N/A — structural gap in `core/services/scan.go`, not version-specific

# Steps To Reproduce

1. Configure kubevuln with `storage: true` and `vexGeneration: true`.
2. Trigger a scan through the `ContainerScanCommand` path (`POST /v1/scanImage`, i.e. `ScanCVE`) for a workload with at least one CVE.
3. Check for an `OpenVulnerabilityExchangeContainer` CR for that image (`kubectl get openvulnerabilityexchangecontainers`).
4. Observe: none exists, and no error/log/metric indicates VEX generation was skipped.
5. Repeat via a registry scan (`ScanRegistry`) twice for the same image, so the second run hits the CVE cache; add or remove a `SecurityException` for one of its CVEs between the two runs.
6. Observe: the first (cache-miss) run produces/updates a VEX document; the second (cache-hit) run, despite the exception-set change altering the stored `IgnoredMatches`, does not touch the VEX document at all.

# Expected behavior

Every CVE-scan flow that has `vexGeneration: true` should generate/update a VEX document whenever it stores a CVE manifest with an exception-filtered result — on both a fresh scan and a cache-hit re-evaluation that changed the ignored-match set — the same way `ScanCP` already does on both its paths.

# Actual Behavior

(citing `core/services/scan.go` as of `upstream/main`)

1. **`ScanCVE` (`func (s *ScanService) ScanCVE`, lines 501-676) has zero `StoreVEX` calls.** Both its cache-miss branch (`cve.Content == nil`, lines 584-609) and cache-hit branch (`else`, lines 610-648) call `StoreCVE`/`StoreCVESummary` but never `s.cveRepository.StoreVEX`, even when `s.vexGeneration` is true. Compare:
   - `ScanCP`'s equivalent block: line 472-473 — `if s.vexGeneration && cveExceptionsComplete && cvepExceptionsComplete { err = s.cveRepository.StoreVEX(ctx, filteredCve, filteredCvep, true) }`
   - `ScanRegistry`'s cache-miss block: line 786-787 — `if s.vexGeneration { err = s.cveRepository.StoreVEX(ctx, filteredCve, filteredCve, false) }`
   - `ScanCVE`: no equivalent exists at either the storage-write site (~line 597-601) or anywhere else in the function.

2. **`ScanRegistry`'s cache-hit branch (`else`, lines 798-823) also has no `StoreVEX` call**, despite the comment immediately preceding it (line 796-798) stating: `"...and SubmitCVE/StoreVEX receive the same unfiltered data a cache miss would produce."` That branch calls `StoreCVE`/`StoreCVESummary` when the ignored-match set changed (lines 806-820) but never `StoreVEX` — the comment documents behavior the code doesn't implement. (`ScanCVE`'s own cache-hit branch, lines 610-648, has near-identical structure and the same stale-comment pattern at line 386.)

3. **`docs/ARCHITECTURE.md`** shows `Generate VEX` as a generic pipeline stage in the CVE Scanning Pipeline diagram, and separately documents that VEX documents are stored as `OpenVulnerabilityExchangeContainer` CRDs — with no caveat that this only happens for `ScanCP`/`ScanRegistry`-cache-miss, not `ScanCVE` or `ScanRegistry`-cache-hit.

# Additional context

**Likely root cause:** `ScanCVE`, `ScanCP`, and `ScanRegistry` are three independently-evolved copies of largely the same "get-or-create SBOM → get-or-create CVE → apply exceptions → store → submit" logic (this repo's issue history — #540, #484, #501, #533 — shows this triplication is a repeat source of exactly this class of bug: one flow gets a feature or fix the other two don't). `StoreVEX` was added to `ScanCP` and `ScanRegistry` independently and `ScanCVE` was simply never given the equivalent call. `ScanRegistry`'s cache-hit branch was likely written by copying `ScanCVE`'s (VEX-less) cache-hit branch rather than mirroring its own cache-miss branch's VEX handling.

**Proposed direction:**
- Add the same `if s.vexGeneration { StoreVEX(...) }` call to `ScanCVE`'s cache-miss branch (mirroring `ScanCP`/`ScanRegistry`), guarded the same way `ScanCP` guards it (only on a complete exception set, matching the existing "never persist removals from an incomplete set" rule already applied to `StoreCVE`).
- Add the missing `StoreVEX` call to both `ScanCVE`'s and `ScanRegistry`'s cache-hit branches, called only when the ignored-match set actually changed (the same condition already gating the `StoreCVE`/`StoreCVESummary` calls in those branches), so a VEX document is kept in sync with exception changes exactly like the CVE manifest already is.
- Since this is the third time (#484, #501, and now this) the same triplicated-flow shape has produced a VEX-consistency bug, consider whether the store-and-submit tail of `ScanCVE`/`ScanCP`/`ScanRegistry` is worth factoring into one shared helper so a fix/feature added to it lands in all three flows at once instead of needing to be ported by hand each time.

**Acceptance criteria:**
- `ScanCVE` calls `StoreVEX` when `vexGeneration` is true, on both its cache-miss and cache-hit paths, gated the same way the sibling flows gate it (complete exception set / changed ignored-match set).
- `ScanRegistry`'s cache-hit branch calls `StoreVEX` when the ignored-match set changed, matching its own cache-miss branch and the comment already describing this as the intended behavior.
- No behavior change when `vexGeneration` is false, or for scans that already correctly generate VEX today (`ScanCP` both paths, `ScanRegistry` cache-miss).
- New/updated tests cover: `ScanCVE` producing a VEX document on a fresh scan and on a cache-hit exception change; `ScanRegistry`'s cache-hit branch producing/updating a VEX document when the exception set changes.

AnshulPatil2005:
## Overview

`ScanCVE`, the handler behind `POST /v1/scanImage`, had no `StoreVEX` call anywhere in its body, so an operator running with `vexGeneration: true` got no VEX document from that route at all and nothing logged or reported why. `ScanRegistry`'s cache-hit branch was missing one too, directly underneath a comment stating that `StoreVEX` runs there.

Before and after, by call site:

| Flow | Path | Before | After |
|---|---|---|---|
| `ScanCP` | both | stores VEX | unchanged |
| `ScanCVE` | cache miss | nothing | stores VEX when the exception set is complete |
| `ScanCVE` | cache hit | nothing | stores VEX when the ignored-match set changed |
| `ScanRegistry` | cache miss | stores VEX | unchanged |
| `ScanRegistry` | cache hit | nothing | stores VEX when the ignored-match set changed |

Each new call is gated the way its siblings gate theirs. On a cache miss `ScanCVE` requires a complete exception set, matching `ScanCP`, because a degraded fetch would otherwise publish a document asserting fewer suppressions than the user actually configured. On a cache hit both flows store VEX under the same condition that already gates `StoreCVE`, so a re-evaluated exception set updates the manifest and the document describing it together or not at all.

Nothing changes when `vexGeneration` is false, or for the two paths that were already correct.

## Additional Information

The call moves into a shared `storeVEX` helper. All five sites were the same six lines, and having to port them by hand is why `ScanCVE` ended up without one. This is the third VEX-consistency bug of this shape after #484 and #501, so the helper is the smallest change that stops the next feature landing in two flows out of three. The larger refactor the issue floats, factoring the whole store-and-submit tail into one helper, is deliberately not attempted here: it would touch all three flows at once and is much harder to review than the bug fix.

`docs/ARCHITECTURE.md` needs no change. Its CVE pipeline diagram already shows `Generate VEX` as a generic phase of CVE scanning, which was the inaccuracy the issue flagged; with this fix the diagram becomes true rather than aspirational.

All four call sites publish only from a known-complete exception set. Two of them changed to get there, both raised in review:

`ScanRegistry`'s cache-miss branch previously stored VEX unconditionally. It now checks `exceptionsComplete` like `ScanCVE` and `ScanCP`, so a degraded fetch no longer publishes a document with partial suppressions. This is a behaviour change on a path that already generated VEX.

The two cache-hit branches originally attached the call to the manifest's persist condition, `exceptionsComplete || !hasRemovals`, which meant a degraded fetch skipped VEX on a cache miss but published it on a cache hit whose change was purely additive. The manifest uses that looser rule because a partial set can only under-suppress, never wrongly un-suppress, so additions are safe to keep. A VEX document is a published assertion about which CVEs are suppressed, and one built from a partial set understates it, so keeping the previous document beats replacing it with a weaker claim. The manifest and the document can therefore disagree briefly until a complete scan, which is the deliberate trade.

## How to Test

```
go test ./core/services/ -run VEX
```

Six tests, of which three fail on `main` (verified by reverting the fix and re-running):

- `ScanCVE_CacheMiss_GeneratesVEX` and `ScanCVE_CacheHit_ExceptionChangeUpdatesVEX` cover the flow that had no VEX generation at all.
- `ScanRegistry_CacheHit_ExceptionChangeUpdatesVEX` covers the branch whose comment already claimed this behaviour.
- `ScanCVE_CacheMiss_NoVEXWhenDisabled`, `ScanCVE_CacheMiss_DegradedFetchSkipsVEX` and `ScanCVE_CacheHit_UnchangedExceptionSkipsVEX` pin the negative cases, so the fix cannot start writing documents when it should not.

The in-memory repository's `StoreVEX` is a no-op that keeps nothing, so the tests wrap it in a small recording repository to observe the calls.

## Related issues/PRs:

* Resolved #557

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency of VEX data across vulnerability and registry scans.
  * VEX updates now reflect changes to filtered vulnerability results, including cached scans.
  * Prevented VEX generation when security-exception data is incomplete.
  * Disabled VEX configurations no longer create or store VEX records.
  * Improved handling of VEX persistence failures without interrupting scans.

### DIFF
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -667,9 +667,19 @@ func (s *ScanService) ScanCVE(ctx context.Context) error {
 							helpers.String("imageSlug", workload.ImageSlug))
 					}
 					// The stored manifest just changed, so the VEX document describing it is
-					// now stale. Kept in sync under the same condition, so a cache hit that
-					// re-evaluates exceptions updates both or neither.
-					s.storeVEX(ctx, filteredCve, filteredCve, false, workload.ImageSlug)
+					// now stale. Republish it, but only from a known-complete exception set,
+					// the same rule the cache-miss path follows.
+					//
+					// The manifest is persisted on the looser `exceptionsComplete ||
+					// !hasRemovals` condition because a partial set can only under-suppress,
+					// never wrongly un-suppress, so additions are safe to keep. A VEX document
+					// is a published assertion about which CVEs are suppressed, and one built
+					// from a partial set understates that. Leaving the previous document in
+					// place, written when the set was complete, beats replacing it with a
+					// weaker claim, so the two can briefly disagree until a complete scan.
+					if exceptionsComplete {
+						s.storeVEX(ctx, filteredCve, filteredCve, false, workload.ImageSlug)
+					}
 				}
 			}
 		}
@@ -845,9 +855,19 @@ func (s *ScanService) ScanRegistry(ctx context.Context) error {
 							helpers.String("imageSlug", workload.ImageSlug))
 					}
 					// The stored manifest just changed, so the VEX document describing it is
-					// now stale. Kept in sync under the same condition, so a cache hit that
-					// re-evaluates exceptions updates both or neither.
-					s.storeVEX(ctx, filteredCve, filteredCve, false, workload.ImageSlug)
+					// now stale. Republish it, but only from a known-complete exception set,
+					// the same rule the cache-miss path follows.
+					//
+					// The manifest is persisted on the looser `exceptionsComplete ||
+					// !hasRemovals` condition because a partial set can only under-suppress,
+					// never wrongly un-suppress, so additions are safe to keep. A VEX document
+					// is a published assertion about which CVEs are suppressed, and one built
+					// from a partial set understates that. Leaving the previous document in
+					// place, written when the set was complete, beats replacing it with a
+					// weaker claim, so the two can briefly disagree until a complete scan.
+					if exceptionsComplete {
+						s.storeVEX(ctx, filteredCve, filteredCve, false, workload.ImageSlug)
+					}
 				}
 			}
 		}
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -788,7 +788,7 @@ func (s *ScanService) ScanRegistry(ctx context.Context) error {
 		}
 
 		// apply security exceptions for storage (copy — original stays intact for SubmitCVE)
-		filteredCve, _ := s.applyExceptionsToManifest(ctx, cve)
+		filteredCve, exceptionsComplete := s.applyExceptionsToManifest(ctx, cve)
 
 		// store filtered CVE
 		if s.storage {
@@ -802,7 +802,12 @@ func (s *ScanService) ScanRegistry(ctx context.Context) error {
 				logger.L().Ctx(ctx).Warning("storing CVE summary", helpers.Error(err),
 					helpers.String("imageSlug", workload.ImageSlug))
 			}
-			s.storeVEX(ctx, filteredCve, filteredCve, false, workload.ImageSlug)
+			// Only publish VEX built from a known-complete exception set, matching how
+			// ScanCVE and ScanCP gate their own calls: a degraded fetch would otherwise
+			// emit a document asserting fewer suppressions than the user configured.
+			if exceptionsComplete {
+				s.storeVEX(ctx, filteredCve, filteredCve, false, workload.ImageSlug)
+			}
 		}
 	} else {
 		// A cached manifest was filtered against the exception set at store time.
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -483,13 +483,8 @@ func (s *ScanService) ScanCP(mainCtx context.Context) error {
 						helpers.String("imageSlug", slug))
 					// no continue, storing the CVE summary is not critical
 				}
-				if s.vexGeneration && cveExceptionsComplete && cvepExceptionsComplete {
-					err = s.cveRepository.StoreVEX(ctx, filteredCve, filteredCvep, true)
-					if err != nil {
-						logger.L().Ctx(ctx).Warning("storing VEX", helpers.Error(err),
-							helpers.String("imageSlug", slug))
-						// no continue, storing the VEX is not critical
-					}
+				if cveExceptionsComplete && cvepExceptionsComplete {
+					s.storeVEX(ctx, filteredCve, filteredCvep, true, slug)
 				}
 			}
 		}
@@ -615,7 +610,7 @@ func (s *ScanService) ScanCVE(ctx context.Context) error {
 		}
 
 		// apply security exceptions for storage (copy — original stays intact for SubmitCVE)
-		filteredCve, _ := s.applyExceptionsToManifest(ctx, cve)
+		filteredCve, exceptionsComplete := s.applyExceptionsToManifest(ctx, cve)
 
 		// store filtered CVE
 		if s.storage {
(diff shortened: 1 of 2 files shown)


matthyx:
In `core/services/scan.go`, line 777 (`ScanRegistry` cache-miss):

```go
filteredCve, _ := s.applyExceptionsToManifest(ctx, cve)
```

And line 791:
```go
s.storeVEX(ctx, filteredCve, filteredCve, false, workload.ImageSlug)
```

Notice that `exceptionsComplete` is discarded on line 777 (`_`), causing `s.storeVEX` to be called unconditionally on cache-miss during registry scans even when exception fetching was degraded (`exceptionsComplete == false`).

In contrast, `ScanCVE` (lines 599, 616) and `ScanCP` (lines 453, 472) both capture `exceptionsComplete` and check it before calling `storeVEX`, ensuring a degraded exception fetch does not publish a VEX document with partial suppressions.

Suggest updating `ScanRegistry`'s cache-miss branch to match:

```go
filteredCve, exceptionsComplete := s.applyExceptionsToManifest(ctx, cve)
// ...
if exceptionsComplete {
    s.storeVEX(ctx, filteredCve, filteredCve, false, workload.ImageSlug)
}
```

matthyx (approved):
LGTM! Thank you for addressing the feedback and adding tests for degraded exception fetches on cache miss/hit.

PR #561, opened 2026-08-10
fix(config): avoid mutating default HTTP client

### LINKED ISSUE #560: Bug: service discovery mutates http.DefaultClient globally
# Description
`LoadBackendServicesConfig` overwrites `http.DefaultClient` with a new client that has a 30-second timeout before calling startup service discovery. That mutates global process state and leaks the timeout into unrelated HTTP traffic elsewhere in the process.

# Environment
OS: `Any`
Version: `upstream/main @ 1ab2445`
 
# Steps To Reproduce
1. Open `config/config.go`.
2. Inspect `LoadBackendServicesConfig` around the service-discovery path.
3. Observe the assignment `http.DefaultClient = &http.Client{Timeout: 30 * time.Second}` before `servicediscovery.GetServices(client)`.
4. Call `LoadBackendServicesConfig` in a process that previously set `http.DefaultClient.Timeout = 0`, then inspect `http.DefaultClient.Timeout` afterward.

# Expected behavior
Service discovery should use a dedicated timeout-bound client or request path without mutating process-global HTTP defaults.

# Actual Behavior
The function changes `http.DefaultClient` for the rest of the process lifetime, including fallback cases where discovery fails and static URLs from `clusterData.json` are used instead.

# Additional context
The active code path is in `config/config.go`. The same process also uses `http.DefaultClient` elsewhere, for example through the backend adapter, so this creates a hidden initialization-order dependency and an unrelated outbound timeout side effect.

rootp1:
## Overview
Startup backend service discovery now uses a private timeout-bound HTTP client instead of replacing `http.DefaultClient`, so the discovery timeout stays scoped to config loading.

## Summary
- Stop mutating `http.DefaultClient` in `LoadBackendServicesConfig`
- Route API discovery through a local wrapper that reuses the existing v3 parser but performs the request with a dedicated `http.Client{Timeout: 30 * time.Second}`
- Add regression coverage for both timeout enforcement and the absence of global default-client mutation
- Fix stale test call sites so the current base branch compiles cleanly in PR CI

## Linked Issue
Fixes #560

## What Changed
- Added `timeoutBoundServiceDiscoveryGetter` in `config/config.go`
- Switched `LoadBackendServicesConfig` to call `loadBackendServicesFromAPI(...)` with a private timeout-bound client
- Added tests covering fallback without `http.DefaultClient` mutation and a short-timeout discovery failure path
- Updated three `adapters/v1/securityexception_test.go` call sites to ignore the stats return from `ConvertToVulnerabilityExceptionPolicies`
- Rebasing onto current `main` removed already-merged unrelated changes from the PR diff

## Verification
- `go test ./adapters/v1 -run 'TestConvertAffectedStatusRequiresAcceptedResponse|TestConvertResolvingStatusesUnchangedByAffected|TestConvertAffectedRecordsProvenance' -count=1` — passed
- `go test ./... -run '^$' -count=1` — passed
- `go test ./config/...` — passed earlier before the CI follow-up fix
- `make build` — passed earlier before the CI follow-up fix

## Scope
- `config/config.go`
- `config/config_test.go`
- `adapters/v1/securityexception_test.go` for the CI-only signature fix

### DIFF
--- a/adapters/v1/securityexception_test.go
+++ b/adapters/v1/securityexception_test.go
@@ -1012,7 +1012,7 @@ func TestConvertAffectedSuppression(t *testing.T) {
 				},
 			}
 
-			policies := ConvertToVulnerabilityExceptionPolicies(exceptions, nil, ExceptionTarget{})
+			policies, _ := ConvertToVulnerabilityExceptionPolicies(exceptions, nil, ExceptionTarget{})
 
 			if tt.wantSuppressed {
 				require.Len(t, policies, 1)
@@ -1043,7 +1043,7 @@ func TestConvertResolvingStatusesUnchangedByAffected(t *testing.T) {
 				},
 			}
 
-			policies := ConvertToVulnerabilityExceptionPolicies(exceptions, nil, ExceptionTarget{})
+			policies, _ := ConvertToVulnerabilityExceptionPolicies(exceptions, nil, ExceptionTarget{})
 
 			require.Len(t, policies, 1, "no actionStatement or response should be needed")
 		})
@@ -1067,7 +1067,7 @@ func TestConvertAffectedRecordsProvenance(t *testing.T) {
 		},
 	}
 
-	policies := ConvertToVulnerabilityExceptionPolicies(exceptions, nil, ExceptionTarget{})
+	policies, _ := ConvertToVulnerabilityExceptionPolicies(exceptions, nil, ExceptionTarget{})
 
 	require.Len(t, policies, 1)
 	assert.Equal(t, "WAF mitigation in place, ticket SEC-1234", policies[0].Attributes["actionStatement"])
--- a/config/config.go
+++ b/config/config.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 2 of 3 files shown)


matthyx (changes requested):
Reviewed the `config.go` fix — the core change is correct and does what it says:

- `LoadBackendServicesConfig` no longer mutates `http.DefaultClient` globally; the discovery call now goes through a private `timeoutBoundServiceDiscoveryGetter` with its own `http.Client{Timeout: 30s}`.
- The wrapper correctly satisfies `schema.IServiceDiscoveryServiceGetter` (embeds `IServiceDiscoveryClient` for `ParseResponse`/`GetHost`/etc., overrides `Get()`), so it type-checks fine.
- `Get()`'s handling of the response body mirrors the exact pattern used by the upstream `backend/pkg/servicediscovery/v3` client's own `Get()` (no new resource-leak regression introduced).
- New tests (`does not mutate http.DefaultClient...`, `TestLoadBackendServicesFromAPI_UsesProvidedTimeoutBoundClient`) directly exercise both the no-mutation guarantee and timeout enforcement — good coverage for issue #560.

One blocker before merge: **the branch is 6 commits behind `main`** (`ahead_by 2, behind_by 6` per compare API), so the "Files changed" tab currently shows 25 files / +554/-68, even though the PR description says only `config/config.go` and `config/config_test.go` were touched. All the extra files (sidecar/syft platform annotations, scan failure reasons, sbomscanner proto/server changes, docs) are just already-merged content from #552/#553/#558/etc. showing up because of the stale base — not new work in this PR.

Please rebase onto latest `main` (or merge `main` in) so the diff matches the stated scope of 2 files. Once that's done and the diff is clean, this is good to merge from my side — no code issues in the actual fix itself.

rootp1:
@matthyx I rebased this onto the current main branch and the diff is back to the config-only change. Could you take another look when you have a moment?

matthyx (approved):
Rebase looks good — diff is now correctly scoped to `config/config.go`, `config/config_test.go`, and one unrelated one-line test-signature fix in `adapters/v1/securityexception_test.go` (from the merge). All CI checks pass. Approving.

PR #563, opened 2026-08-10
feat(registryauth): resolve ECR credentials from the ambient AWS identity

### LINKED ISSUE #254: Integration with more image registries for vulnerability scanning

AnshulPatil2005:
## Overview

Picking up the ECR half of #254. As you noted on the issue, every registry already works through `registryx`; the painful part is authentication, and ECR is where it hurts most: an authorization token is valid for 12 hours, so a Secret holding one has to be rotated twice a day to keep working.

This adds an ECR auth provider that requests a token per scan from the ambient AWS credential chain, which on EKS is the pod's IAM role, so no Secret is needed at all. It plugs into the extension point #500 built, which left this comment:

> `// Add ECR/ACR providers here as they're implemented.`

Two details worth calling out:

- **The region comes from the registry hostname**, not from ambient AWS config. A cluster in one region can pull from a registry in another, so assuming the ambient region would break exactly the multi-region setups this is meant to help. Standard, FIPS and China partition hostnames are all matched.
- **ECR Public (`public.ecr.aws`) is deliberately not matched.** It serves anonymous pulls, so the existing anonymous fallback already covers it and requesting a token would be pointless.

## Additional Information

**The providers move to `internal/registryauth`.** Only the sidecar was pluggable. The in-process Syft adapter carried its own hardcoded GCP-only fallback, with duplicate copies of `isGCPRegistry` and `gcpCredentials` and a duplicate `TestIsGCPRegistry`. Adding ECR to the sidecar alone would have left it working on one scan path and not the other, which is the same "one flow gets a feature the others don't" shape as #484, #501 and #557. Both paths now consult the same provider list, and the duplicated test collapses into one.

**A metric bug is fixed on the way.** Both fallback loops hardcoded `gcp_adc` as the strategy label, including inside the generic provider loop the sidecar already had. An ECR fallback would have been reported as a GCP one. The label now comes from the provider, and a test asserts the strategies are distinct so a future provider cannot silently reuse one.

**ACR is not included.** The interface makes it roughly a 20 line addition now, but ACR auth is a two step exchange (AAD token, then an ACR refresh token) rather than ECR's single call, so it is worth its own PR rather than doubling this one.

**One testing gap, stated plainly.** The sidecar's fallback loop is covered end to end because `resolveSource` takes an injectable `sourceGetter`. The in-process path calls `syft.GetSource` directly with no seam, so its provider lookup is not covered by a unit test here; the substitution is mechanical and the provider logic itself is fully tested. Adding an injection point to `SyftAdapter.CreateSBOM` would be a worthwhile follow-up but is a bigger refactor than this change warrants.

**Dependencies:** `aws-sdk-go-v2/config` and `aws-sdk-go-v2/service/ecr` were already in the module graph as indirect dependencies. The only `go.mod` change is those two lines moving into the direct block; `go.sum` is untouched and no versions change.

## How to Test

```
go test ./internal/registryauth/ ./pkg/sbomscanner/v1/ ./adapters/v1/
```

`TestECRMatchesAndRegion` covers the hostname matching and region extraction, including the negative cases that matter for a credential path: ECR Public, a short account id, and a lookalike host that merely contains an ECR hostname later in the path. `TestCredentialsFromAuthorizationToken` covers the token decode, including a password containing colons, since only the first colon separates user from password. `TestECRCredentialsUsesRegionFromReference` pins that the region is taken from the image reference.

## Related issues/PRs:

* Resolved #254

## Summary by CodeRabbit

* **New Features**
  * Added automatic authentication support for GCP Artifact Registry, GCR, and Amazon ECR image sources.
  * Image retrieval now retries with matching ambient credentials before falling back to anonymous access.
* **Bug Fixes**
  * Improved handling of registry-specific credentials, including regional and specialized ECR endpoints.
  * Added validation for malformed or unavailable registry authorization data.
* **Monitoring**
  * Registry authentication attempts now record the strategy used, improving visibility into image pull behavior.

### DIFF
--- a/adapters/v1/syft.go
+++ b/adapters/v1/syft.go
@@ -26,11 +26,11 @@ import (
 	"github.com/kubescape/kubevuln/core/domain"
 	"github.com/kubescape/kubevuln/core/ports"
 	"github.com/kubescape/kubevuln/internal/metrics"
+	"github.com/kubescape/kubevuln/internal/registryauth"
 	"github.com/kubescape/kubevuln/internal/tools"
 	"github.com/kubescape/storage/pkg/apis/softwarecomposition/v1beta1"
 	"github.com/opencontainers/go-digest"
 	"go.opentelemetry.io/otel"
-	"golang.org/x/oauth2/google"
 )
 
 // formatResolvedPlatform builds an OCI-style "os/arch[/variant]" string from the platform
@@ -45,23 +45,6 @@ func formatResolvedPlatform(os, arch, variant string) string {
 		parts = append(parts, variant)
 	}
 	return strings.Join(parts, "/")
-}
-
-func isGCPRegistry(imageID string) bool {
-	host, _, _ := strings.Cut(imageID, "/")
-	return host == "gcr.io" || strings.HasSuffix(host, ".gcr.io") || strings.HasSuffix(host, "-docker.pkg.dev")
-}
-
-func gcpCredentials(ctx context.Context) (*image.RegistryCredentials, error) {
-	creds, err := google.FindDefaultCredentials(ctx, "https://www.googleapis.com/auth/cloud-platform")
-	if err != nil {
-		return nil, err
-	}
-	token, err := creds.TokenSource.Token()
-	if err != nil {
-		return nil, err
-	}
-	return &image.RegistryCredentials{Username: "oauth2accesstoken", Password: token.AccessToken}, nil
 }
 
 // SyftAdapter implements SBOMCreator from ports using Syft's API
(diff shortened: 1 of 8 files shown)


matthyx (approved):
Reviewed. Checked out the branch, built with `go build ./...` (clean), ran `go vet ./...` (clean), and ran the tests called out in the PR description:

```
ok  github.com/kubescape/kubevuln/internal/registryauth   0.004s
ok  github.com/kubescape/kubevuln/pkg/sbomscanner/v1       0.231s
ok  github.com/kubescape/kubevuln/adapters/v1              58.190s
```

Notes on the diff:
- The new `internal/registryauth` package cleanly unifies the GCP/ECR provider logic that both the in-process Syft adapter and sidecar previously duplicated (with the sidecar's copy being GCP-only), matching the description.
- ECR host regex is anchored (`^...$`) and correctly excludes `public.ecr.aws`, short account IDs, and lookalike hosts — all covered by `TestECRMatchesAndRegion`.
- Region is taken from the image reference rather than ambient AWS config, as intended for multi-region clusters (`TestECRCredentialsUsesRegionFromReference`).
- The metric-label bug fix (`gcp_adc` hardcoded in both fallback loops) is fixed, and `TestRegistryAuthStrategiesAreDistinct`-style test now guards against future providers reusing a strategy label.
- `go.mod` change is exactly the two AWS packages moving from indirect to direct, no version bumps, `go.sum` untouched, as described.
- Token-decode logic correctly splits on the first colon only, handling passwords containing colons.

No blockers. Good to merge.

Issue #359, opened 2026-05-07
proxyRegistryMap does not truly work


mkm29:
# Description

Because the default delimiter for Viper is `.`, `proxyRegistryMap: {"docker.io": "my-mirror.example.com"}` actually becomes 2 keys: `docker` and `io`

# Environment
OS: `k8s v1.34.4+k3s1 `
Version: ` v0.3.134`
 
# Steps To Reproduce

1. Deploy kubescape operator with above version of kubevuln
2. Allow operator to spin up resources
3. See error is kubevuln pod:
```
{"level":"fatal","ts":"2026-05-07T00:33:27Z","msg":"load config error","error":"decoding failed due to the following error(s):\n\n'proxyRegistryMap[docker]' expected type 'string', got unconvertible type 'map[string]interface {}'"}
```

# Expected behavior
kubevuln should be able to properly handly registry mirroring (as trivy-operator does), in my case mirroring to internal registry, and scan all images.

# Actual Behavior
kubevuln pod fails to start with above error

# Additional context

PR #361, opened 2026-05-07
fix: set viper delim key and test proxyRegistryMap

mkm29:
## Overview

Create instance of Viper with delimiter key of `::` to properly support `proxyRegistryMap` (e.g. keys will likely contain `.`'s)

## How to Test

Run tests to ensure no regression, to test particular change run `go test ./config -run=TestLoadConfigProxyRegistryMap`

## Examples/Screenshots

> Here you add related screenshots 
-->

## Related issues/PRs:

If this resolved an issue, write "Resolved #359 

## Checklist before requesting a review

put an [x] in the box to get it checked 

- [x] My code follows the style guidelines of this project
- [x] I have commented on my code, particularly in hard-to-understand areas
- [x] I have performed a self-review of my code
- [x] If it is a core feature, I have added thorough tests.
- [x] New and existing unit tests pass locally with my changes

**Please open the PR against the `dev` branch (Unless the PR contains only documentation changes)**

There is no `dev` branch.
 

## Summary by CodeRabbit

* **New Features**
  * Configuration system enhanced to support nested configuration keys for improved organization.
  * New proxy registry mapping capability enables defining container image mirrors for multiple registries (Docker Hub, Quay.io, GCR).

* **Tests**
  * Added comprehensive test coverage validating proxy registry mapping configuration.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubevuln/pull/361)

PR #564, opened 2026-08-10
fix(config): warn when storage is enabled but riskAcceptance isn't, document both undocumented fields

### LINKED ISSUE #562: Bug: riskAcceptance silently no-ops SecurityException CRD integration, and it's undocumented (like proxyRegistryMap was, per #359)
# Description

`riskAcceptance` (`config/config.go:60`, `RiskAcceptance bool`) is the single switch that gates all of kubevuln's SecurityException/ClusterSecurityException CRD integration — exception matching, VEX suppression, the `VulnerabilitySuppressed` K8s Event, and the `kubevuln_exceptions_*` Prometheus metrics all depend on it. When it's unset (its Go zero value, `false`, since it's the only bool `Config` field with no explicit `v.SetDefault(...)` call — see `config/config.go`'s defaults block), kubevuln silently falls back to `repositories.NoOpSecurityExceptionRepository`, which always returns empty results, no error. There is no log line, warning, or metric anywhere on that path — an operator who has deployed `SecurityException`/`ClusterSecurityException` CRDs per `docs/security-exception-design.md` has no way to discover, short of reading `cmd/http/main.go`, that they're being silently ignored.

Compounding this: `riskAcceptance` is completely absent from `docs/CONFIGURATION.md`, which otherwise documents every other `Config` struct field — its options tables (`Scanning Options`, `Storage Options`, `Feature Flags`) and its `### Complete Schema` JSON block list all 20 of the other 22 `mapstructure`-tagged fields in `config/config.go`. `proxyRegistryMap` (`config/config.go:55`) is the only other field missing from that schema, for the same reason (never documented) — it controls registry-mirroring for every image pull path (`v1.NewSyftAdapter`/`v1.NewSidecarSBOMAdapter` in `cmd/http/main.go`) with no mention in any `.md` file in the repo.

# Environment
OS: N/A — structural gap in `config/config.go`, `cmd/http/main.go`, `docs/CONFIGURATION.md`, not version-specific

# Steps To Reproduce

1. Follow `docs/security-exception-design.md` to create a `SecurityException` CRD suppressing a known CVE for a workload.
2. Deploy kubevuln with `storage: true` (per `docs/CONFIGURATION.md`'s Storage Options table) but without setting `riskAcceptance` (it isn't in the docs, so there's no reason to set it).
3. Trigger a scan for that workload.
4. Observe: the CVE is not suppressed — it appears in scan results as if no `SecurityException` existed.
5. Check the logs and `/metrics` for any indication of why: `cmd/http/main.go`'s `if storage != nil && c.RiskAcceptance { ... } else { seRepo = &repositories.NoOpSecurityExceptionRepository{} }` produces no log line, warning, or metric on the `else` branch — compare to the `if` branch, which logs `"SecurityException CRD integration enabled"` (`cmd/http/main.go:105`).

# Expected behavior

Either `riskAcceptance` should be documented in `docs/CONFIGURATION.md` clearly enough that an operator following the SecurityException design doc knows to set it, or kubevuln should log a warning at startup when `SecurityException` integration is disabled (`storage != nil && !c.RiskAcceptance`), so a misconfiguration is visible instead of silently discarding CRDs an operator has already created. `proxyRegistryMap` should likewise be documented.

# Actual Behavior

- `riskAcceptance` gates real, security-relevant behavior (`cmd/http/main.go:104-124`) but is mentioned in this repo's docs exactly once, in passing, inside `docs/security-exception-design.md`'s Events section (`"...enabled only when SecurityException CRD integration itself is enabled (storage and riskAcceptance both configured...)"`) — not in `docs/CONFIGURATION.md`'s Storage Options table, Feature Flags table, or Complete Schema (`docs/CONFIGURATION.md:191-304`).
- The `else` branch that results from `riskAcceptance` being unset (`cmd/http/main.go:122-124`) produces zero log output. It is indistinguishable at runtime from "no SecurityException CRDs exist in this cluster."
- `proxyRegistryMap` (`config/config.go:55`) has zero mentions in any `.md` file in the repo (`git grep -i proxyRegistryMap -- '*.md'` returns nothing), despite being wired into every SBOM-creator constructor call (`cmd/http/main.go:85,88,91`).
- Cross-checking all 22 `mapstructure`-tagged fields in `config/config.go` against `docs/CONFIGURATION.md`'s `### Complete Schema` JSON block confirms these are the *only* two fields missing — every other field is documented.
- This isn't hypothetical: `proxyRegistryMap` being undocumented has already caused a real production incident — the closed #359 is a user whose kubevuln pod crash-looped at startup (`decoding failed due to the following error(s): 'proxyRegistryMap[docker]' expected type 'string', got unconvertible type 'map[string]interface {}'`) because Viper's default `.` key delimiter silently splits a `docker.io` key. #359 fixed the underlying Viper delimiter bug, but the field is still absent from `docs/CONFIGURATION.md` today, so the next operator configuring it has no documented example of the correct shape to avoid hitting the same class of mistake again.

# Additional context

**Proposed direction:**
- Add `riskAcceptance` and `proxyRegistryMap` to `docs/CONFIGURATION.md`'s options tables and `Complete Schema` JSON block, alongside a cross-reference from the Storage Options / SecurityException sections to `docs/security-exception-design.md`.
- Add a startup log warning in `cmd/http/main.go`'s `else` branch (mirroring the existing `"SecurityException CRD integration enabled"` info log on the `if` branch) — something like `"SecurityException CRD integration disabled (storage=%v, riskAcceptance=%v); SecurityException/ClusterSecurityException CRDs will not be applied"` — so this misconfiguration is diagnosable from logs alone, the same way the repo already treats other "silently doing nothing" failure modes as bugs worth fixing (see the several closed issues in this repo's history about swallowed/silent errors).

**Acceptance criteria:**
- `riskAcceptance` and `proxyRegistryMap` appear in `docs/CONFIGURATION.md`'s tables and Complete Schema, with descriptions matching their actual runtime effect.
- `cmd/http/main.go` logs a warning when `storage` is enabled but `riskAcceptance` is not, so SecurityException CRD integration being off is visible without reading source.
- No behavior change when `riskAcceptance` is already correctly set to `true`, or when `storage` is disabled entirely (no SecurityException integration is expected either way, so no new warning is needed there).

bhuvan-somisetty:
## Overview

`riskAcceptance` (`config/config.go:60`) is the single switch that gates all of kubevuln's `SecurityException`/`ClusterSecurityException` CRD integration — exception matching, VEX suppression, the `VulnerabilitySuppressed` K8s Event, and the `kubevuln_exceptions_*` Prometheus metrics all depend on it. When it's unset (its Go zero value `false`), `cmd/http/main.go` silently falls back to `repositories.NoOpSecurityExceptionRepository`, which always returns empty results, no error — with zero log output on that path. An operator who has deployed `SecurityException`/`ClusterSecurityException` CRDs per `docs/security-exception-design.md` has no way to discover, short of reading `main.go`, that they're being silently ignored.

Compounding this: `riskAcceptance` was completely absent from `docs/CONFIGURATION.md`, which otherwise documents every other `Config` field. Cross-checking all 22 `mapstructure`-tagged fields in `config/config.go` against the doc's Complete Schema showed exactly two omissions: `riskAcceptance` and `proxyRegistryMap`. `proxyRegistryMap` being undocumented already caused a real production incident (#359, a crash-loop from Viper's key delimiter splitting a `docker.io`-style key — fixed by #361, but never documented).

## Additional Information

- `cmd/http/main.go`: logs a warning (`"SecurityException CRD integration disabled: storage is enabled but riskAcceptance is not set; SecurityException/ClusterSecurityException CRDs will not be applied"`) in the branch that results from `riskAcceptance` being unset, but only when `storage` is enabled — that's the actual misconfiguration case; when `storage` is disabled, no SecurityException integration is expected either way, so no new warning fires there. Mirrors the existing `"SecurityException CRD integration enabled"` info log on the opposite branch.
- `docs/CONFIGURATION.md`: added `riskAcceptance` to the Storage Options table (with a cross-reference to `security-exception-design.md` and a note about the now-fixed silent-no-op behavior) and `proxyRegistryMap` to the Feature Flags table (with a note on the `::` key delimiter fix from #361), plus both fields to the Complete Schema JSON block.

No behavior change for any already-correct configuration — this is additive observability (one log line) and documentation only.

## How to Test

- `go build ./...`
- `go test ./cmd/http/... ./config/...` — pass
- `gofmt -l cmd/http/main.go` — clean
- Manually: set `storage: true` without `riskAcceptance`, start kubevuln, confirm the new warning appears in the logs at startup instead of nothing.

## Related issues/PRs

Resolved #562

## Summary by CodeRabbit

- **New Features**
  - Added a startup warning when storage is enabled without risk acceptance, clarifying that SecurityException integrations will not be applied.
  - Added configuration options for risk acceptance and proxy registry mappings.

- **Documentation**
  - Documented the new configuration options, including defaults, behavior, and schema definitions.

### DIFF
--- a/cmd/http/main.go
+++ b/cmd/http/main.go
@@ -120,6 +120,13 @@ func main() {
 			logger.L().Warning("failed to get k8s config for SecurityException events")
 		}
 	} else {
+		// storage alone means an operator has SecurityException/ClusterSecurityException CRDs
+		// reachable but riskAcceptance unset -- without this log, that combination silently
+		// no-ops (see #562): the NoOpSecurityExceptionRepository below always returns empty
+		// results, indistinguishable at runtime from "no CRDs exist in this cluster".
+		if storage != nil {
+			logger.L().Warning("SecurityException CRD integration disabled: storage is enabled but riskAcceptance is not set; SecurityException/ClusterSecurityException CRDs will not be applied")
+		}
 		seRepo = &repositories.NoOpSecurityExceptionRepository{}
 	}
 
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -195,6 +195,7 @@ The main configuration file. All options can be overridden via environment varia
 | `storage` | bool | `false` | Enable Kubernetes storage backend (stores SBOMs/CVEs as CRDs) |
 | `namespace` | string | `kubescape` | Kubernetes namespace for storage |
 | `storeFilteredSbom` | bool | `false` | Store relevancy-filtered SBOMs |
+| `riskAcceptance` | bool | `false` | Enable `SecurityException`/`ClusterSecurityException` CRD integration (exception matching, VEX suppression, suppression Events/metrics) — requires `storage: true` as well. See [security-exception-design.md](security-exception-design.md). If `storage` is enabled but this is left unset, matching CRDs are silently ignored: a warning is logged at startup, but no error or metric flags it, so double-check this is set before relying on `SecurityException`/`ClusterSecurityException` CRDs. |
 
 #### Feature Flags
 
@@ -204,6 +205,7 @@ The main configuration file. All options can be overridden via environment varia
 | `nodeSbomGeneration` | bool | `false` | Enable node-level SBOM generation |
 | `partialRelevancy` | bool | `false` | Enable partial relevancy matching |
 | `vexGeneration` | bool | `false` | Generate VEX (Vulnerability Exploitability eXchange) documents |
+| `proxyRegistryMap` | map[string]string | `{}` | Maps a registry hostname to an internal mirror for image pulls, e.g. `{"docker.io": "my-mirror.example.com"}`. Applied to every SBOM-generation path (in-process and sidecar). Config keys are parsed with a `::` delimiter specifically so hostnames containing `.` are treated as a single map key instead of being split into nested keys (see #359/#361) — no special escaping needed in `docker.io`-style keys. |
 
 ### Complete Schema
 
@@ -246,6 +248,15 @@ The main configuration file. All options can be overridden via environment varia
       "default": false
     },
     "partialRelevancy": {
+      "type": "boolean",
+      "default": false
+    },
+    "proxyRegistryMap": {
+      "type": "object",
+      "additionalProperties": { "type": "string" },
+      "default": {}
+    },
+    "riskAcceptance": {
       "type": "boolean",
       "default": false
     },


matthyx (approved):
Reviewed. The logic is correct: the warning fires only when `storage != nil && !c.RiskAcceptance`, matching the described misconfiguration case, and doesn't fire when `storage == nil` (no SecurityException integration expected either way). Docs additions for `riskAcceptance` and `proxyRegistryMap` are accurate against `config/config.go` and correctly cross-reference the `::` delimiter fix (#361).

Note on CI: `pr-created / test / Create cross-platform build` is failing, but it's a pre-existing build error in `adapters/v1/securityexception_test.go` (`ConvertToVulnerabilityExceptionPolicies` returns 2 values, test expects 1) — unrelated to this PR's two changed files (`cmd/http/main.go`, `docs/CONFIGURATION.md`). Confirmed by building `main` directly: same `vet` failure exists there before this PR's commit. Not a blocker introduced by this change; worth a separate fix/issue for the base branch.

CodeRabbit's inline suggestion (rewording "silently ignored"/covering both false and unset) is a minor wording nitpick, not a functional issue — optional to address.

LGTM, approving.

PR #543, opened 2026-08-09
feat(securityexception): add affected status with actionStatement and typed response

AnshulPatil2005:
## Overview

SecurityException implemented three of OpenVEX's four statuses, so there was no way to say a finding is real and the risk has been accepted. Authors had to reach for `not_affected` with a stretched justification, which suppresses the finding but publishes a claim that does not survive audit: the vulnerability is present and the code path is reachable.

This adds the three pieces the RFC asks for:

1. `affected` as the fourth `status` value, completing the OpenVEX vocabulary.
2. An optional `actionStatement` free-text field, mapping to OpenVEX `action_statement` and parallel to the existing `impactStatement`.
3. An optional typed `response[]` list, aligned with CycloneDX VEX `analysis.response[]`.

```yaml
vulnerabilities:
  - vulnerability:
      id: "CVE-2021-44228"
    status: "affected"
    actionStatement: "Risk accepted: WAF blocks the exploit vector. Reviewed 2026-07-30 by security lead. Ticket SEC-1234."
    response:
      - will_not_fix
```

### Suppression semantics

Suppression is driven by `response[]` rather than by the status. `affected` asserts the vulnerability applies, which is the opposite of an assertion an author would expect to hide a finding, so the finding stays visible unless the entry also carries an `actionStatement` and every stated response says no remediation is coming:

| `response[]` | Finding |
|---|---|
| `can_not_fix`, `will_not_fix` | Suppressed. No fix is coming, the risk is accepted. |
| `update`, `rollback`, `workaround_available` | Visible. A remediation is planned or in place, so it stays trackable until confirmed. |
| A mix, such as `will_not_fix` plus `update` | Visible. The entry is still tracking work, which is the fail-closed reading. |
| Empty, or no `actionStatement` | Visible. |

Documents using the existing three statuses are unaffected: `not_affected` and `fixed` still suppress on the status alone, with no `actionStatement` or `response[]` needed.

## Additional Information

**On requiring `actionStatement`.** The issue body proposes "at least one of `actionStatement` or `response[]`" and lists strict OpenVEX conformance under Alternatives. The discussion settled the other way: @slashben argued for requiring `actionStatement` unconditionally so every `affected` entry is a valid OpenVEX statement by construction rather than needing `response[]` serialized into free text on degradation, and @pjungermann agreed. This PR implements that, so `response[]` stays a purely additive typed extension layered on top.

[...]

PR #556, opened 2026-08-10
chore(deps): bump kubescape/storage to v0.0.302, unblocking #554

matthyx:
## Overview

Bumps `github.com/kubescape/storage` from `v0.0.258` to `v0.0.302`, which adds the `SourceKind`/`SourceName`/`SourceNamespace`/`Justification`/`ImpactStatement` fields to `IgnoreRule` needed by #554. This PR only unblocks the build — it does not wire the new fields into `buildSuppressionAttributes`/`logSuppression`, which is the actual scope of #554.

## The blocker

As reported in [#554 (comment)](https://github.com/kubescape/kubevuln/issues/554#issuecomment-5238669752), bumping `storage` alone doesn't build:

```
containerd@v1.7.33/oci/spec_opts.go:1548:34: cannot use limit (variable of type int64) as *int64 value in assignment
```

Root cause: `storage`'s dependency tree now pulls in `opencontainers/runtime-spec` v1.3.0 (via `containers/common`), which changed `LinuxPids.Limit` from `int64` to `*int64`. `containerd/containerd` v1.7.x (still on the old field type) doesn't build against that — this affects any consumer of both modules, not just kubevuln, and containerd's own fix for it (containerd/containerd#13087) is still unmerged on `release/1.7`.

kubevuln pulls in `containerd/containerd` v1.7.x transitively through our `github.com/anchore/stereoscope` replace, which pins a fork (`matthyx/stereoscope`) carrying two small local patches (schema v1 registry fallback, `maxImageSize` support used by `adapters/v1/syft.go` and `pkg/sbomscanner/v1/server.go`) on top of an old base.

## Fix

Upstream `anchore/stereoscope` solved this the same way containerd's v2 line does: it migrated to `containerd/containerd/v2` (done as of v0.1.22, matching what `anchore/syft` v1.42.3 itself depends on). I rebased our two fork patches onto that release — both cherry-picked cleanly with no conflicts — and tagged it as [`matthyx/stereoscope@v0.1.22-ks.1`](https://github.com/matthyx/stereoscope/releases/tag/v0.1.22-ks.1).

The `anchore/syft` fork replace (`kubescape/syft v1.32.0-ks.2`) needed no changes — despite `storage` nominally requiring syft v1.42.3, the replace fully substitutes it and the existing fork still satisfies everything actually used.

[...]

PR #565, opened 2026-08-10
fix(adapters): restore the test build after the exception stats change

AnshulPatil2005:
## Overview

`main` currently fails to build its tests:

```
$ go test ./adapters/v1/
adapters/v1/securityexception_test.go:1015:16: assignment mismatch: 1 variable but ConvertToVulnerabilityExceptionPolicies returns 2 values
adapters/v1/securityexception_test.go:1046:16: assignment mismatch: 1 variable but ConvertToVulnerabilityExceptionPolicies returns 2 values
adapters/v1/securityexception_test.go:1070:14: assignment mismatch: 1 variable but ConvertToVulnerabilityExceptionPolicies returns 2 values
FAIL    github.com/kubescape/kubevuln/adapters/v1 [build failed]
```

`ConvertToVulnerabilityExceptionPolicies` gained a second return value in #553, which updated every call site that existed when it was written. #543 merged immediately before it (`5674f1c`, then `6c4e519`) and added three more call sites, so the two together leave the package uncompilable.

Neither PR is at fault and both were green on their own branches. The conflict is semantic rather than textual, so git merged them without complaint and nothing caught it until they were on `main` together.

This discards the stats value in the three affected tests, matching how every other call site in the file already ignores it.

## How to Test

```
go test ./adapters/v1/
```

Fails to build on `main`, passes here. The full suite (`go test ./...`) is green with this change.

### DIFF
--- a/adapters/v1/securityexception_test.go
+++ b/adapters/v1/securityexception_test.go
@@ -1012,7 +1012,7 @@ func TestConvertAffectedSuppression(t *testing.T) {
 				},
 			}
 
-			policies := ConvertToVulnerabilityExceptionPolicies(exceptions, nil, ExceptionTarget{})
+			policies, _ := ConvertToVulnerabilityExceptionPolicies(exceptions, nil, ExceptionTarget{})
 
 			if tt.wantSuppressed {
 				require.Len(t, policies, 1)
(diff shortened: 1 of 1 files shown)


AnshulPatil2005:
closing this, #556 landed the same three-line fix while this was open. no longer needed.

PR #571, opened 2026-08-10
fix(registryauth): cache and dedupe GCP/ECR credential fetches across both SBOM paths

### LINKED ISSUE #569: Bug: registry auth credentials (GCP/ECR) are fetched fresh on every scan, uncached and undeduplicated, across two duplicated call sites
# Description

`internal/registryauth` (added in PR #563, resolving #492's pluggable-provider refactor) resolves ambient cloud credentials — GCP Application Default Credentials and AWS ECR tokens — for private-registry pulls. Neither provider caches what it fetches: `GCP.Credentials()` (`internal/registryauth/registryauth.go:71-73`, delegating to `gcpCredentials`, lines 81-90) calls `google.FindDefaultCredentials` and takes a fresh token on every invocation; `ECR.Credentials()` (`:120-122`, delegating to `ecrCredentials`, lines 135-143) calls `config.LoadDefaultConfig` and `ecr.NewFromConfig(cfg).GetAuthorizationToken(...)` — a live AWS STS + ECR API round trip — on every invocation. The package's own doc comment states an ECR authorization token is valid for **12 hours** (`:112`), yet it's re-fetched synchronously in the hot path of every single scan of a private ECR image.

This isn't a single call site — the exact same retry chain that calls `provider.Credentials(ctx, pullRef)` is duplicated across two independent files: `adapters/v1/syft.go:205-236` (in-process SBOM path) and `pkg/sbomscanner/v1/server.go:85-136` (`resolveSource`, sidecar path). `server.go`'s own comment (line 85-86) says it "mirrors the retry chain in `adapters/v1/syft.go` so the sidecar and in-process adapters behave the same" — confirming this is deliberately kept in sync by hand, with no shared caching layer added when `internal/registryauth` was created specifically to be that shared dependency.

# Environment
OS: N/A — structural gap in `internal/registryauth`, `adapters/v1/syft.go`, `pkg/sbomscanner/v1/server.go`, not version-specific

# Steps To Reproduce

1. Deploy kubevuln scanning private images from an ECR repository, using IRSA (ambient AWS credentials, no static pull secret) — the scenario `internal/registryauth`'s own doc comment cites as the reason this package exists.
2. Trigger concurrent scans of several images from that same ECR repository (the worker pool in `controllers/http.go`, sized by `scanConcurrency`, runs scan jobs in parallel).
3. Observe (via AWS CloudTrail or ECR/STS API metrics) that each scan independently calls `sts:AssumeRole`/`ecr:GetAuthorizationToken`, once per scan, regardless of how recently another concurrent or prior scan already fetched a still-valid token for the same account+region.
4. Under sustained scan volume, or a burst against a freshly deployed cluster with many workloads pointing at the same private registry, this can trip AWS API rate limiting on `GetAuthorizationToken`/STS, surfacing as spurious scan failures (`401`/anonymous-fallback-failed) that look like a registry outage but are self-inflicted request volume.

# Expected behavior

A credential fetched from GCP ADC or AWS ECR should be reused for its actual validity window (provider-reported expiry — ECR's is in the `GetAuthorizationTokenOutput`, GCP's `oauth2.Token` has `Expiry`) instead of being re-fetched on every scan. Concurrent scans of the same registry/account should not each independently hit the cloud provider's token endpoint for credentials that are, or are about to be, fetched by another in-flight scan.

# Actual Behavior

- Every scan of a private image that hits a `401` calls `provider.Credentials(ctx, pullRef)` fresh — no cache, no TTL, no reuse — confirmed at both call sites (`adapters/v1/syft.go:208-209`, `pkg/sbomscanner/v1/server.go:102-106`).
- `internal/registryauth/registryauth.go` has no cache, no singleflight/dedup mechanism, and no test in `internal/registryauth/registryauth_test.go` exercising repeated or concurrent calls — only single-call success/failure paths are covered (confirmed by reading the test file's structure).
- The two call sites are hand-synchronized duplicates of the same retry chain, so any caching fix applied to only one (e.g., wrapping just `adapters/v1/syft.go`'s call) would leave the sidecar path — the one explicitly recommended for memory-intensive scans per `docs/scan-failure-reporting.md`'s OOM-resilience section — still making an uncached call on every scan.

# Additional context

**Likely root cause:** `internal/registryauth` was created (PR #563) specifically to deduplicate the *retry-chain logic* between the in-process and sidecar paths (per #492's stated goal: a pluggable provider interface so "a registry added for one path is available to the other"). Caching the *result* of that logic was out of scope for that refactor and was never added as a follow-up, even though the package's own doc comment explicitly reasons about token lifetime (the 12-hour ECR window) as the motivation for ambient-credential resolution in the first place — the caching gap directly undercuts that stated rationale, since a per-scan network fetch reintroduces a cost the design was meant to avoid.

**Proposed direction (why this needs a coordinated, multi-module fix, not a one-file patch):**
- Add a TTL-aware cache inside `internal/registryauth` itself (not in either caller), keyed appropriately per provider — GCP by registry host (ADC identity is process-wide, but Artifact Registry hosts can differ), ECR by account+region (the same credential-cache-key-blindness class of bug already fixed once for the rate-limit cache in #527, which is worth reusing the lesson from) — honoring each provider's own reported expiry rather than a single fixed TTL, since ECR (~12h) and GCP ADC (~1h) tokens have materially different lifetimes.
- Add request coalescing (e.g. `golang.org/x/sync/singleflight`, already a common Go idiom for exactly this "concurrent callers, one in-flight fetch" shape) so concurrent scans of the same registry during a cache miss collapse to one upstream call instead of a thundering herd.
- Wire both `adapters/v1/syft.go` and `pkg/sbomscanner/v1/server.go` through the same cached provider, since fixing only one leaves the other still hammering the cloud provider on every scan — this is the "cross-module" requirement: the fix has to land in the shared package and be consumed identically by both callers to actually close the gap.
- Extend `internal/registryauth/registryauth_test.go` with concurrency tests (concurrent callers during a cache miss produce one upstream call; a cached, still-valid credential is served without a new call; an expired one triggers a re-fetch), and add cache hit/miss observability alongside the existing `FallbackStrategyECR`/`FallbackStrategyGCPADC` fallback metrics in `internal/metrics/metrics.go`, matching this repo's existing precedent for instrumenting caches (`kubevuln_exceptions_degraded_total`, the SecurityException list cache, etc.).
- Document the caching/TTL tradeoff in the package's own doc comment, matching the precedent PR #531 set for documenting a similar per-cache tradeoff.

**Acceptance criteria:**
- A credential fetched from GCP or ECR is reused across scans until its provider-reported expiry, instead of being re-fetched every scan.
- Concurrent scans hitting a cache miss for the same registry/account collapse to a single upstream credential fetch.
- Both `adapters/v1/syft.go` and `pkg/sbomscanner/v1/server.go` go through the same cached path — no independent, uncached credential fetch remains in either.
- New tests cover cache-hit reuse, expiry-driven re-fetch, and concurrent-caller coalescing.
- No behavior change to the existing anonymous-fallback semantics when a provider doesn't match or credentials are genuinely unavailable.

bhuvan-somisetty:
## Overview

Neither `GCP.Credentials` nor `ECR.Credentials` (`internal/registryauth/registryauth.go`) cached what they fetched: every scan of a private image that hit a `401` re-ran `google.FindDefaultCredentials` or `config.LoadDefaultConfig`+`GetAuthorizationToken` from scratch, even though an ECR authorization token is valid for 12 hours and a GCP ADC token for roughly an hour. Concurrent scans of the same registry — the HTTP controller's worker pool runs scans in parallel — each independently hit the cloud provider's token endpoint for the same still-valid credential, risking STS/ECR rate limiting under load and adding avoidable per-scan latency. The retry chain that calls this is also duplicated across two independent files — `adapters/v1/syft.go` (in-process) and `pkg/sbomscanner/v1/server.go` (sidecar) — so fixing only one call site would have left the other still uncached.

## Additional Information

- `internal/registryauth/cache.go` (new): a `credentialCache` that reuses a fetched credential until the provider's own reported expiry (GCP's `oauth2.Token.Expiry`, ECR's `AuthorizationData.ExpiresAt` — never a guessed TTL), keyed appropriately per provider — a single key for GCP, since ADC isn't registry-specific, and one key per AWS region for ECR, since a token is only valid for the region it was issued in (mirroring how `ECRCredsFn` is already parameterized by region). Concurrent callers racing a cache miss for the same key collapse into a single upstream fetch via `golang.org/x/sync/singleflight` (already a dependency, no new one added). Fetch errors are never cached, so a transient STS/ADC hiccup doesn't poison the cache for the next call.
- `internal/registryauth/registryauth.go`: `GCP.Credentials`/`ECR.Credentials` now go through this cache. `adapters/v1/syft.go` and `pkg/sbomscanner/v1/server.go` needed **no changes** — both already call `provider.Credentials(...)`, so caching is transparent to them. `GCPCredsFn`/`ECRCredsFn` grew a `time.Time` return (the fetched credential's expiry) to feed the cache.
- `internal/registryauth/registryauth_test.go` / `pkg/sbomscanner/v1/server_test.go`: existing test overrides of `GCPCredsFn`/`ECRCredsFn` updated for the new signature, and now call the new `registryauth.ResetCaches()` so each test's override actually reaches `Credentials()` instead of serving a value cached by an earlier test in the same binary.
- `internal/registryauth/cache_test.go` (new): cache-hit reuse, expiry-driven refetch, distinct cache keys not sharing entries, concurrent-miss coalescing (20 goroutines racing one key resolve to exactly one upstream fetch), fetch errors not being cached, and a zero-expiry result correctly not being cached.
- `internal/metrics/metrics.go`: new `kubevuln_registry_auth_cache_total` counter (labels `strategy`, `result` = `hit`/`miss`), matching this repo's existing convention for instrumenting caches (`kubevuln_exceptions_degraded_total`, the SecurityException list cache, etc.).

No behavior change to the existing anonymous-fallback semantics when a provider doesn't match or credentials are genuinely unavailable — this is purely additive caching in front of the existing fetch calls.

## How to Test

- `go build ./...`
- `go test ./internal/registryauth/... ./internal/metrics/... ./pkg/sbomscanner/v1/...` — all pass
- `go vet ./internal/registryauth/... ./internal/metrics/... ./pkg/sbomscanner/v1/... ./adapters/v1/...` — clean
- `gofmt -l` on all touched files — clean
- New tests specifically prove the two things this issue is about: `TestCredentialCache_ReusesUnexpiredEntry`/`RefetchesAfterExpiry` (no re-fetch until expiry) and `TestCredentialCache_ConcurrentMissesCollapseToOneFetch` (concurrent scans don't each hit the cloud provider).

## Related issues/PRs

Resolved #569

## Summary by CodeRabbit

* **New Features**
  * Added credential caching for GCP and ECR registry authentication.
  * Reuses valid credentials until expiration and refreshes them when needed.
  * Deduplicates simultaneous credential requests.
  * Keeps credentials isolated by registry and provider context.
  * Added cache reset support.

* **Bug Fixes**
  * Prevents failed or non-expiring credentials from being cached.
  * Preserves ECR credential expiration information.

* **Monitoring**
  * Added metrics for registry authentication cache hits and misses.

### DIFF
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -45,11 +45,15 @@ const (
 	SourceResolutionFallbackAssistedSuccess = "fallback_assisted_success"
 	SourceResolutionFallbackFailed          = "fallback_failed"
 	SourceResolutionFirstPassFailure        = "first_pass_failure"
+
+	RegistryAuthCacheHit  = "hit"
+	RegistryAuthCacheMiss = "miss"
 )
 
 var recorderMu sync.RWMutex
 var fallbackCounter metric.Int64Counter
 var sourceResolutionCounter metric.Int64Counter
+var registryAuthCacheCounter metric.Int64Counter
 
 // Metrics wraps the OTel meter and Prometheus exporter used to expose kubevuln's
 // operational metrics (worker-pool backlog, scan outcomes, rejection counts) on
@@ -168,9 +172,18 @@ func New() (*Metrics, error) {
 		return nil, err
 	}
 
+	registryAuthCacheMetric, err := meter.Int64Counter(
+		"kubevuln_registry_auth_cache_total",
+		metric.WithDescription("Total number of registry auth credential lookups, by provider strategy and cache result (hit/miss)"),
+	)
+	if err != nil {
+		return nil, err
+	}
+
 	recorderMu.Lock()
 	fallbackCounter = scanFallbackCounter
 	sourceResolutionCounter = sourceResolutionMetric
+	registryAuthCacheCounter = registryAuthCacheMetric
 	recorderMu.Unlock()
 
 	return &Metrics{
@@ -217,6 +230,22 @@ func RecordScanFallback(ctx context.Context, component, category, strategy, outc
 		attribute.String("category", category),
 		attribute.String("strategy", strategy),
 		attribute.String("outcome", outcome),
+	))
+}
+
+// RecordRegistryAuthCache reports a registry-auth credential lookup's cache outcome
+// (RegistryAuthCacheHit/RegistryAuthCacheMiss), labeled by the provider strategy
+// (FallbackStrategyGCPADC/FallbackStrategyECR) that served it.
+func RecordRegistryAuthCache(ctx context.Context, strategy, result string) {
+	recorderMu.RLock()
+	counter := registryAuthCacheCounter
+	recorderMu.RUnlock()
+	if counter == nil {
+		return
+	}
+	counter.Add(ctx, 1, metric.WithAttributes(
+		attribute.String("strategy", strategy),
+		attribute.String("result", result),
 	))
 }
 
--- a/internal/metrics/metrics_test.go
+++ b/internal/metrics/metrics_test.go
@@ -22,6 +22,7 @@ func TestNewAndHandler(t *testing.T) {
 	m.ExceptionsDegradedCounter.Add(context.Background(), 1, metric.WithAttributes())
 	RecordScanFallback(context.Background(), ComponentInProcess, FallbackCategoryRegistryAuth, FallbackStrategyAnonymous, FallbackOutcomeSucceeded)
 	RecordSourceResolution(context.Background(), ComponentInProcess, true, true)
+	RecordRegistryAuthCache(context.Background(), FallbackStrategyECR, RegistryAuthCacheHit)
 
 	req := httptest.NewRequest("GET", "/metrics", nil)
 	w := httptest.NewRecorder()
@@ -35,6 +36,7 @@ func TestNewAndHandler(t *testing.T) {
 	assert.True(t, strings.Contains(body, "kubevuln_exceptions_degraded_total"), body)
 	assert.True(t, strings.Contains(body, `kubevuln_scan_fallbacks_total{category="registry_auth",component="in_process",outcome="succeeded",strategy="anonymous"} 1`), body)
 	assert.True(t, strings.Contains(body, `kubevuln_scan_source_resolution_total{component="in_process",outcome="fallback_assisted_success"} 1`), body)
+	assert.True(t, strings.Contains(body, `kubevuln_registry_auth_cache_total{result="hit",strategy="ecr"} 1`), body)
 }
 
 func TestMeterRegistersObservableGauge(t *testing.T) {
--- a/internal/registryauth/cache.go
+++ b/internal/registryauth/cache.go
@@ -1,0 +1,103 @@
+package registryauth
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	"github.com/anchore/stereoscope/pkg/image"
+	"github.com/kubescape/kubevuln/internal/metrics"
+	"golang.org/x/sync/singleflight"
+)
+
+// expiryLeeway is subtracted from a credential's reported expiry, so a token that's about
+// to expire mid-scan is refreshed proactively instead of being handed out and failing the
+// pull moments later.
+const expiryLeeway = 1 * time.Minute
+
+// credentialFetch is a provider's underlying fetch: it returns credentials plus the time
+// they stop being valid. A zero expiry means "no reported expiry" -- the credential is
+// still returned, but not cached, since there's nothing to bound the cache entry's
+// lifetime by.
+type credentialFetch func(ctx context.Context) (*image.RegistryCredentials, time.Time, error)
+
+// credentialCache reuses a fetched credential until its provider-reported expiry, and
+// collapses concurrent callers racing a cache miss for the same key into a single upstream
+// fetch. Without this, every scan of a private image independently re-fetches credentials
+// -- an ECR authorization token is valid for 12 hours, yet was being re-requested on every
+// single scan -- and a burst of concurrently running scans against the same registry (the
+// HTTP controller's worker pool runs many scans in parallel) each fired their own upstream
+// request, risking cloud provider rate limiting. See #569.
+type credentialCache struct {
+	mu       sync.Mutex
+	entries  map[string]cacheEntry
+	group    singleflight.Group
+	strategy string
+	// now is overridable in tests so expiry can be exercised without a real time.Sleep.
+	now func() time.Time
+}
+
+type cacheEntry struct {
+	creds  *image.RegistryCredentials
+	expiry time.Time
+}
+
+func newCredentialCache(strategy string) *credentialCache {
+	return &credentialCache{entries: map[string]cacheEntry{}, strategy: strategy, now: time.Now}
+}
+
+// get returns a cached, still-valid credential for key if one exists, otherwise calls
+// fetch -- deduplicated across concurrent callers racing the same key via singleflight --
+// and caches the result when fetch reports a non-zero expiry. Fetch errors are never
+// cached: a transient failure (e.g. a momentary STS/ADC hiccup) must not poison the cache
+// for other callers or the next scan.
+func (c *credentialCache) get(ctx context.Context, key string, fetch credentialFetch) (*image.RegistryCredentials, error) {
+	if creds, ok := c.lookup(key); ok {
+		metrics.RecordRegistryAuthCache(ctx, c.strategy, metrics.RegistryAuthCacheHit)
+		return creds, nil
+	}
+
+	v, err, _ := c.group.Do(key, func() (interface{}, error) {
+		// Re-check under the singleflight key: a caller that lost the race to become the
+		// leader for this key may have been given a result already cached by the winner
+		// of an earlier, now-resolved Do call for the same key.
+		if creds, ok := c.lookup(key); ok {
+			return creds, nil
+		}
+		creds, expiry, ferr := fetch(ctx)
+		if ferr != nil {
+			return nil, ferr
+		}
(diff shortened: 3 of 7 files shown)


matthyx (changes requested):
## PR Review: Changes Requested

Thanks for putting this together! The credential caching and request coalescing design in `internal/registryauth` is a great improvement. However, there are a few key blocking issues around cache scoping and context propagation that need to be addressed before merging:

### 1. 🚨 Blocker: ECR Cache Key Collision Across AWS Accounts
In `internal/registryauth/registryauth.go`:
```go
func (ECR) Credentials(ctx context.Context, imageID string) (*image.RegistryCredentials, error) {
	region := ecrRegion(imageID)
	return ecrCache.get(ctx, region, func(ctx context.Context) (*image.RegistryCredentials, time.Time, error) {
		return ECRCredsFn(ctx, region)
	})
}
```
Currently, `ecrCache` keys entries solely by `region` (e.g., `"us-east-1"`). If a cluster pulls images from two different AWS accounts in the same region (e.g. `111111111111.dkr.ecr.us-east-1.amazonaws.com` and `222222222222.dkr.ecr.us-east-1.amazonaws.com`), the second account will receive the authorization token cached for the first account. Because ECR authorization tokens are scoped to a specific AWS account, this causes 401 Unauthorized errors for the second account.

**Suggested Fix:** Key ECR cache entries by registry host (or `account:region`), e.g., using `host(imageID)` or `account + ":" + region`.

---

### 2. GCP Cache Key Scoping
In `internal/registryauth/registryauth.go`:
```go
const gcpCacheKey = "adc"
```
Per issue #569, GCP cache entries should be scoped by registry host (`host(imageID)`), matching the isolation design for GCR / Artifact Registry hosts.

---

### 3. Singleflight Leader Context Cancellation Risk
In `internal/registryauth/cache.go`:
```go
v, err, _ := c.group.Do(key, func() (interface{}, error) {
	...
	creds, expiry, ferr := fetch(ctx)
	...
})
```
`c.group.Do` executes `fetch(ctx)` using the `ctx` of whichever goroutine happens to win the singleflight leader race. If that leader caller's context is cancelled or times out (e.g., client HTTP disconnect), `fetch` fails and propagates `context.Canceled` to *all* concurrent callers waiting in `group.Do`, even if their contexts are still valid.

**Suggested Fix:** Use `c.group.DoChan` with a background/detached context for `fetch` (e.g., `context.WithoutCancel(ctx)`), allowing each caller to select on `ch` vs `ctx.Done()`.

---

### 4. Concurrency Test Synchronization
In `internal/registryauth/cache_test.go`:
In `TestCredentialCache_ConcurrentMissesCollapseToOneFetch`, `close(release)` is called immediately without waiting for all 20 worker goroutines to reach the cache lookup/fetch barrier. A sync barrier (e.g. `sync.WaitGroup` or atomic counter) should be used to guarantee all workers are blocked before releasing `fetch`.

---

Please address these findings and add unit tests for cross-account / cross-host cache isolation.

bhuvan-somisetty:
Thanks for the review @matthyx, good catch on the ECR one.

- Cross-account collision: fixed. The cache key was region-only, so two accounts pulling through the same region were sharing one entry and the second account's requests could get handed a token they weren't authorized for. Now keyed by registry host, which for ECR already encodes the account ID, so each account gets its own cache entry.
- GCP cache: switched from the single fixed "adc" key to keying by host as well, so it doesn't rely on the ambient identity being the same across every GCR/Artifact Registry host a pod pulls from.
- Also swapped the singleflight leader's fetch to run under its own timeout-bounded context instead of one caller's ctx, so one caller cancelling can't fail every other caller racing the same key.

Added tests for cross-account (ECR) and cross-host (GCP) isolation, plus tightened the concurrent-fetch test so it can't pass by accident without real coalescing. Full suite for registryauth/metrics/sbomscanner/adapters is green, gofmt/vet clean.

bhuvan-somisetty:
Pushed another follow-up — CodeRabbit's latest pass flagged that the concurrency test's sleep-based contention barrier was still just a heuristic, not a real guarantee, and that's fair. Replaced it with an actual synchronization barrier: an `onMiss` hook (same pattern as the existing `now` hook) that fires for every caller reaching the miss path before it joins singleflight, and the leader's fetch is gated on all n callers hitting that point. So the cache literally cannot be populated until every goroutine has genuinely raced the same miss — no timing window left to get lucky/unlucky on.

Ran it 50x in a row to check for flakiness, all green. Full suite still passing across registryauth/metrics/sbomscanner/adapters.

bhuvan-somisetty:
@coderabbitai review

bhuvan-somisetty:
@coderabbitai review

bhuvan-somisetty:
@matthyx all 4 points from your review are addressed and pushed — ECR/GCP caches are now keyed by registry host, the singleflight fetch runs on a detached context so one caller's cancellation can't kill it for everyone else, and the concurrency test uses a real join barrier instead of a sleep. All CI checks are green; this is just waiting on a re-review. Could you take another look?

matthyx (approved):
LGTM! All blocking issues have been addressed: ECR cache keys now include the registry host/account ID, GCP cache keys are scoped per host, singleflight uses  with a detached fetch timeout to avoid leader context cancellation cascading, and new unit tests verify cross-account/cross-host isolation and deterministic concurrency. Ready to merge.

PR #573, opened 2026-08-10
fix(adapters): use the shared ignore predicate in Summarize

### LINKED ISSUE #572: Bug: Summarize and ApplySecurityExceptions disagree on whether a CVE is suppressed
# Description

Two code paths answer "is this CVE suppressed?" with different predicates, so the in-cluster `VulnerabilityManifest` and the summary sent to the backend can disagree about the same finding. Same shape as #449, different predicate.

`ApplySecurityExceptions` (`adapters/v1/securityexception.go`) uses `hasIgnoreAction`, which scans every policy and every action:

```go
func hasIgnoreAction(policies []armotypes.VulnerabilityExceptionPolicy) bool {
	for _, p := range policies {
		for _, a := range p.Actions {
			if a == armotypes.Ignore {
```

`Summarize` (`adapters/v1/backend_utils.go:230`) reads only the first policy's first action:

```go
isIgnored := len(vulnerabilities[i].ExceptionApplied) > 0 &&
	len(vulnerabilities[i].ExceptionApplied[0].Actions) > 0 &&
	vulnerabilities[i].ExceptionApplied[0].Actions[0] == armotypes.Ignore
```

`ExceptionApplied` is populated by `getCVEExceptionMatchCVENameFromList`, which appends **every** policy whose `VulnerabilityPolicies` name the CVE and never filters on actions. So `ExceptionApplied[0]` is not guaranteed to be the policy carrying `Ignore`.

# Environment

- OS: N/A, code-level inconsistency
- Version: `main` (`39f669e`)

# Steps To Reproduce

Give a finding two matching policies, the first with no actions and the second carrying `Ignore`, then run both predicates over it:

```
hasIgnoreAction (manifest path)      = true    -> CVE hidden in the stored manifest
summary counts it as active (Total)  = 1       -> CVE reported active to the backend
```

The same divergence occurs with a single policy whose `Actions` are `["alert_only", "ignore"]`, since only `Actions[0]` is read.

# Expected behavior

Both surfaces decide suppression with one predicate, so a CVE is suppressed or not consistently across the stored manifest and the backend report.

# Actual Behavior

The stored manifest hides the finding while the summary counts it toward `TotalCount` and the active severity stats, so the two Kubescape surfaces disagree about the same CVE.

# Additional context

Reachability is limited today. `buildPolicy` always sets exactly `[Ignore]` for CRD-derived policies, so this needs a cloud-sourced policy with empty or non-standard `Actions` ordered ahead of an ignoring one, and `armotypes.VulnerabilityExceptionPolicyActions` currently defines only `Ignore`. So this is a latent inconsistency rather than something firing in every cluster, but it is the same failure mode #449 described and the fix is a one-line reuse of the helper that already exists in the same package.

AnshulPatil2005:
## Overview

`Summarize` decided whether a CVE was suppressed by reading only `ExceptionApplied[0].Actions[0]`, while `ApplySecurityExceptions` decides it with `hasIgnoreAction`, which scans every policy and every action.

`getCVEExceptionMatchCVENameFromList` appends every policy whose `VulnerabilityPolicies` name the CVE and never filters on actions, so `ExceptionApplied[0]` is not necessarily the policy carrying `Ignore`. When it is not, the stored `VulnerabilityManifest` hides the finding while the summary sent to the backend counts it toward `TotalCount` and the active severity stats, so the two surfaces disagree about the same CVE.

This calls `hasIgnoreAction` instead, so both paths answer the question the same way. It is a one-line change reusing a helper that already lives in the same package.

## Additional Information

Reachability is limited today, and worth stating plainly rather than overselling: `buildPolicy` always sets exactly `[Ignore]` for CRD-derived policies, and `armotypes.VulnerabilityExceptionPolicyActions` currently defines only `Ignore`. Triggering it needs a cloud-sourced policy with empty or non-standard `Actions` ordered ahead of an ignoring one. So this is a latent inconsistency rather than something firing in every cluster.

It is the same failure mode as #449, fixed in #533: two surfaces computing the same predicate differently, leaving the manifest and the backend report disagreeing about whether a finding is suppressed.

## How to Test

```
go test ./adapters/v1/ -run TestSummarize_IgnoreMatchesManifestPredicate
```

Five cases. Two of them, where `Ignore` is carried by a later policy or a later action, fail on `main` (verified by reverting the change and re-running). Each case additionally asserts that `Summarize`'s answer equals `hasIgnoreAction`'s for the same input, so the two predicates cannot drift apart again without a test failing.

## Related issues/PRs:

* Resolved #572

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected vulnerability suppression so findings are ignored whenever any applied exception policy includes an Ignore action.
  * Ensured findings remain active when no Ignore action is present.
* **Tests**
  * Added coverage for multiple policies and actions to verify consistent suppression behavior.

### DIFF
--- a/adapters/v1/backend_utils.go
+++ b/adapters/v1/backend_utils.go
@@ -227,9 +227,13 @@ func Summarize(report v1.ScanResultReport, vulnerabilities []containerscan.Commo
 	vulnsList := make([]containerscan.ShortVulnerabilityResult, 0)
 
 	for i := range vulnerabilities {
-		isIgnored := len(vulnerabilities[i].ExceptionApplied) > 0 &&
-			len(vulnerabilities[i].ExceptionApplied[0].Actions) > 0 &&
-			vulnerabilities[i].ExceptionApplied[0].Actions[0] == armotypes.Ignore
+		// Same predicate ApplySecurityExceptions uses to decide suppression for the stored
+		// manifest. Reading only ExceptionApplied[0].Actions[0] would answer a narrower
+		// question: getCVEExceptionMatchCVENameFromList appends every policy matching the
+		// CVE name without filtering on actions, so the first one is not necessarily the
+		// one carrying Ignore, and the two surfaces would then disagree about whether the
+		// finding is suppressed.
+		isIgnored := hasIgnoreAction(vulnerabilities[i].ExceptionApplied)
 
 		severitiesStats := exculdedSeveritiesStats
 		if !isIgnored {
--- a/adapters/v1/backend_utils_test.go
+++ b/adapters/v1/backend_utils_test.go
@@ -705,3 +705,77 @@ func Test_sendSummaryAndVulnerabilities(t *testing.T) {
 		})
 	}
 }
+
+// Summarize and ApplySecurityExceptions must answer "is this CVE suppressed?" the same way.
+// getCVEExceptionMatchCVENameFromList appends every policy whose VulnerabilityPolicies name
+// the CVE, without filtering on actions, so ExceptionApplied[0] is not necessarily the policy
+// carrying Ignore. Reading only the first policy's first action made the stored manifest hide
+// a finding the backend summary still counted as active.
+func TestSummarize_IgnoreMatchesManifestPredicate(t *testing.T) {
+	ignoring := armotypes.VulnerabilityExceptionPolicy{
+		PortalBase: armotypes.PortalBase{Name: "ignoring-policy"},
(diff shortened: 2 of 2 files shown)


matthyx (dismissed):
LGTM! Using  in  correctly aligns backend summary reporting with  suppression logic. Tests cover multi-policy and multi-action scenarios well, and all CI checks are passing.

matthyx (approved):

PR #577, opened 2026-08-10
fix: honor caller cancellation in security-exception and label lookups

### LINKED ISSUE #576: bug: security-exception and label lookups ignore caller cancellation for up to 30 seconds
# Description
`GetSecurityExceptions`, `GetWorkloadLabels`, and `GetNamespaceLabels` in `repositories/apiserver.go` each wrap the caller's context with `context.WithTimeout(context.WithoutCancel(ctx), 30*time.Second)` before issuing their Kubernetes API calls (lines 215, 316, and 334). `context.WithoutCancel` strips the caller's cancellation signal entirely, so when the caller (HTTP request context, worker-pool shutdown, or an upstream deadline) is canceled or expires, these three lookups keep the underlying dynamic-client `List`/`Get` call running for up to the full 30 seconds instead of aborting immediately.

This regressed a guarantee that #421 / PR #422 established: that storage-layer K8s API calls honor caller cancellation. These three call sites detach cancellation entirely rather than propagating it, so shutdown and request aborts cannot stop this work promptly.

# Environment
OS: `darwin arm64`
Version: local checkout at `upstream/main` (this behavior is present on `main` as of this report)

# Steps To Reproduce
1. In `repositories/apiserver.go`, inspect the three call sites:
   - `GetSecurityExceptions` at line 215
   - `GetWorkloadLabels` at line 316
   - `GetNamespaceLabels` at line 334
2. Note that each builds its request context with `context.WithTimeout(context.WithoutCancel(ctx), 30*time.Second)`, which discards the caller's `ctx` cancellation entirely and only bounds the call with a fixed 30-second timeout.
3. Compare this to the caller-provided `ctx`, which is still used for logging (e.g. `logger.L().Ctx(ctx)`) and higher-level scan lifecycle tracking in the same functions.
4. Observe that canceling the caller's `ctx` (request abort, shutdown, or an upstream deadline) has no effect on the in-flight `List`/`Get` call issued with `listCtx`/`getCtx` — it can only be stopped by the fixed 30-second timeout, or by it completing normally.

Reproducing the exact timing would require injecting a slow dynamic client or temporary test scaffolding not currently in the repository; the verification performed here is static analysis of the three call sites plus their surrounding logging/lifecycle code, which does use the original `ctx`.

# Expected behavior
Caller cancellation should propagate to these lookups so that when the caller's context is canceled (shutdown, request abort, or upstream deadline), the in-flight Kubernetes API call for `GetSecurityExceptions`, `GetWorkloadLabels`, or `GetNamespaceLabels` aborts promptly instead of continuing for up to 30 more seconds.

# Actual Behavior
All three lookups unconditionally detach the caller's cancellation signal via `context.WithoutCancel(ctx)` and only bound the call with a fixed 30-second timeout, so cancellation cannot stop the downstream Kubernetes API operation until that timeout expires (or the call completes).

# Additional context
This is a distinct, remaining gap from the `context.Background()` issue fixed in #421 / PR #422 — these three call sites already receive and use the caller's `ctx` for logging, but still explicitly detach it for the actual API call via `context.WithoutCancel`, per the comment at `repositories/apiserver.go:213-214` ("Use a detached context with timeout — the scan context may be canceled before the CRD listing completes due to rate limiting.").

Suggested direction: propagate the caller's cancellation by default for these three lookups (e.g. derive the call context from `ctx` directly, optionally still bounded by a shorter explicit timeout), and only keep a detached/extended context where there is a demonstrated correctness need — documented and covered by a focused test.

rootp1:
## Overview

`GetSecurityExceptions`, `GetWorkloadLabels`, and `GetNamespaceLabels` in `repositories/apiserver.go` each derived their K8s API call's context with `context.WithTimeout(context.WithoutCancel(ctx), 30*time.Second)`. `context.WithoutCancel` strips the caller's cancellation signal, so canceling the caller (shutdown, request abort, upstream deadline) could not stop the in-flight `List`/`Get` call for up to 30 seconds. This PR switches all three to `context.WithTimeout(ctx, 30*time.Second)` so cancellation propagates while still bounding the call at 30 seconds, and adds regression tests proving each lookup now honors a canceled caller context.

## How to Test

Run the new regression tests, which fail against the old `context.WithoutCancel` behavior and pass with this fix:

```bash
go test -v -run "HonorsCallerCancellation" ./repositories/...
```

Verification run against this change:

- `make build` — passed
- `go vet ./...` — passed
- `golangci-lint run ./repositories/...` — passed (two pre-existing findings remain in `repositories/apiserver.go` at unrelated lines 1345 and 1391; confirmed present on `upstream/main` before this change)
- `go test ./...` — all packages pass except `adapters/v1`, which fails on pre-existing, network/registry-platform-dependent tests (`arm64` vs `amd64` image resolution) unrelated to this change

## Related issues/PRs:

* Resolved #576

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -210,9 +210,7 @@ func NewFakeAPIServerStorage(namespace string, objects ...runtime.Object) *APISe
 // never cached, so a transient apiserver hiccup self-heals on the next call instead of being
 // pinned for the TTL.
 func (a *APIServerStore) GetSecurityExceptions(ctx context.Context, namespace string) ([]sev1beta1.SecurityException, []sev1beta1.ClusterSecurityException, error) {
-	// Use a detached context with timeout — the scan context may be canceled
-	// before the CRD listing completes due to rate limiting.
-	listCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), 30*time.Second)
+	listCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
 	defer cancel()
 
 	var listErrs []error
@@ -313,7 +311,7 @@ func (a *APIServerStore) GetWorkloadLabels(ctx context.Context, namespace, kind,
 	if err != nil {
 		return nil, fmt.Errorf("failed to resolve GroupVersionResource for kind %q: %w", kind, err)
 	}
-	getCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), 30*time.Second)
+	getCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
 	defer cancel()
 	obj, err := a.DynamicClient.Resource(gvr).Namespace(namespace).Get(getCtx, name, metav1.GetOptions{})
 	if err != nil {
@@ -331,7 +329,7 @@ func (a *APIServerStore) GetNamespaceLabels(ctx context.Context, name string) (m
 	if name == "" {
 		return nil, nil
 	}
-	getCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), 30*time.Second)
+	getCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
 	defer cancel()
 	obj, err := a.DynamicClient.Resource(namespaceGVR).Get(getCtx, name, metav1.GetOptions{})
 	if err != nil {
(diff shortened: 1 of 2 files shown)


matthyx (approved):
LGTM! Verified that removing  allows caller cancellation to properly propagate to in-flight K8s API calls in , , and , while maintaining the 30-second upper bound. Unit tests in  cover all three lookups and pass locally.

PR #579, opened 2026-08-10
fix(adapters): match an exception policy at most once per CVE

### LINKED ISSUE #578: Bug: a SecurityException with a redundant alias suppresses once but is counted twice
# Description

`getCVEExceptionMatchCVENameFromList` (`adapters/v1/backend_utils.go`) appends the whole policy once per matching `VulnerabilityPolicies` entry rather than once per policy:

```go
for i := range srcCVEList {
	for j := range srcCVEList[i].VulnerabilityPolicies {
		if strings.EqualFold(srcCVEList[i].VulnerabilityPolicies[j].Name, CVEName) {
			...
			l = append(l, srcCVEList[i])   // no break
		}
	}
}
```

`buildPolicy` (`adapters/v1/securityexception.go`) expands one vulnerability entry into a `VulnerabilityPolicy` per id **and per alias**. So an exception whose aliases repeat the id, or list the same alias twice, produces a policy with two entries naming the same CVE, and that policy is returned twice.

# Environment

- OS: N/A, code-level defect
- Version: `main` (`39f669e`)

# Steps To Reproduce

A `SecurityException` entry with a redundant alias:

```yaml
vulnerabilities:
  - vulnerability:
      id: "CVE-2021-44228"
      aliases: ["CVE-2021-44228", "GHSA-jfh8", "GHSA-jfh8"]
    status: "not_affected"
```

Converting it and matching gives:

```
policies built: 1
  VulnerabilityPolicies entries: 4
matched for CVE-2021-44228: 2   (expected 1)
matched for GHSA-jfh8:      2   (expected 1)
suppressingPolicies(matched) = 2
```

# Expected behavior

A single exception policy matches a given CVE at most once, however many of its id/alias entries name that CVE.

# Actual Behavior

The duplicate reaches two consumers of the returned slice:

- `logSuppression` iterates `suppressingPolicies(matched)`, so one suppressed finding produces two log lines and, since #537, **two Kubernetes Events**.
- `DomainToArmo` sets `ExceptionApplied` from this slice, so the report sent to the backend lists the same exception twice for one CVE.

The `matchedBySource` metric path is already safe: it collects `sourceKind` into a set before counting, so `kubevuln_exceptions_matched_total` is not inflated. That guard suggests the duplicate was already a known hazard at one call site, worked around locally rather than fixed at the source.

# Additional context

Only the duplicate is wrong; two genuinely distinct exceptions matching the same CVE should still return two policies.

The `ExpiredOnFix` check currently sits inside the inner loop even though it is a property of the policy rather than of an individual id/alias entry. Hoisting it makes the skip independent of which entry happens to name the CVE, with the same outcome.

AnshulPatil2005:
## Overview

`getCVEExceptionMatchCVENameFromList` appended the whole policy once per matching `VulnerabilityPolicies` entry rather than once per policy. Since `buildPolicy` expands one vulnerability entry into a `VulnerabilityPolicy` per id **and per alias**, an exception whose aliases repeat the id, or list the same alias twice, returned the same policy several times:

```yaml
vulnerabilities:
  - vulnerability:
      id: "CVE-2021-44228"
      aliases: ["CVE-2021-44228", "GHSA-jfh8", "GHSA-jfh8"]
    status: "not_affected"
```

```
matched for CVE-2021-44228: 2   (expected 1)
matched for GHSA-jfh8:      2   (expected 1)
```

The duplicate reaches two consumers: `logSuppression` iterates `suppressingPolicies(matched)`, so one suppressed finding emits two log lines and two Kubernetes Events (#537), and `DomainToArmo` builds `ExceptionApplied` from the same slice, so the backend is told the same exception applied twice. The `matchedBySource` metric path is already safe, since it collects `sourceKind` into a set before counting.

Breaking after the first matching entry fixes it. Two genuinely distinct exceptions matching the same CVE still return two policies.

## Additional Information

The `ExpiredOnFix` check moves out of the inner loop in the same change. It is a property of the policy, not of an individual id or alias entry, so testing it once per policy is equivalent and no longer depends on which entry happened to name the CVE. A test pins that an `ExpiredOnFix` policy is skipped whichever of its entries matches.

## How to Test

```
go test ./adapters/v1/ -run TestGetCVEExceptionMatch
```

`TestGetCVEExceptionMatch_PolicyMatchesAtMostOnce` covers an id repeated as its own alias, the same alias listed twice, case-insensitive duplicates, and the negative cases that must keep working: two distinct policies still return two, and a non-matching CVE returns none. The three duplicate cases fail without the `break` (verified by reverting it). `TestGetCVEExceptionMatch_ExpiredOnFixSkipsWholePolicy` covers the hoisted check. The pre-existing `TestGetCVEExceptionMatchCVENameFromList` is unchanged and still passes.

## Related issues/PRs:

* Resolved #578

### DIFF
--- a/adapters/v1/backend_utils.go
+++ b/adapters/v1/backend_utils.go
@@ -319,12 +319,18 @@ func getCVEExceptionMatchCVENameFromList(srcCVEList []armotypes.VulnerabilityExc
 	var l []armotypes.VulnerabilityExceptionPolicy
 
 	for i := range srcCVEList {
+		if filterFixed && srcCVEList[i].ExpiredOnFix != nil && *srcCVEList[i].ExpiredOnFix {
+			continue
+		}
 		for j := range srcCVEList[i].VulnerabilityPolicies {
 			if strings.EqualFold(srcCVEList[i].VulnerabilityPolicies[j].Name, CVEName) {
-				if filterFixed && srcCVEList[i].ExpiredOnFix != nil && *srcCVEList[i].ExpiredOnFix {
-					continue
-				}
+				// A policy contributes at most once. buildPolicy expands a vulnerability
+				// entry into one VulnerabilityPolicy per id and alias, so an exception
+				// listing an alias equal to its id, or the same alias twice, would
+				// otherwise return the same policy several times, and every consumer
+				// counts it that many times.
 				l = append(l, srcCVEList[i])
+				break
 			}
 		}
 	}
--- a/adapters/v1/backend_utils_test.go
+++ b/adapters/v1/backend_utils_test.go
@@ -705,3 +705,83 @@ func Test_sendSummaryAndVulnerabilities(t *testing.T) {
 		})
 	}
 }
+
+// buildPolicy expands one vulnerability entry into a VulnerabilityPolicy per id and alias,
+// so an exception listing an alias equal to its id, or repeating an alias, used to make the
+// same policy match several times. Every consumer then counted it that many times: duplicate
+// ExceptionApplied entries sent to the backend, and one suppression log line and Kubernetes
+// Event per duplicate.
+func TestGetCVEExceptionMatch_PolicyMatchesAtMostOnce(t *testing.T) {
+	policy := func(names ...string) armotypes.VulnerabilityExceptionPolicy {
+		vp := make([]armotypes.VulnerabilityPolicy, 0, len(names))
+		for _, n := range names {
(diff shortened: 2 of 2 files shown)


matthyx (approved):
LGTM! The fix correctly ensures that a matching vulnerability exception policy is appended at most once per CVE by breaking out of the inner loop after the first policy match, preventing duplicate suppression logs, Kubernetes events, and backend ExceptionApplied entries. Unit tests pass cleanly.

PR #581, opened 2026-08-10
fix(sbomscanner): stop the abandoned Syft goroutine racing the handler

### LINKED ISSUE #580: Bug: data race between the handler and the abandoned Syft goroutine on CreateSBOM timeout
# Description

`scannerServer.CreateSBOM` (`pkg/sbomscanner/v1/server.go`) runs Syft inside `deadline.Run`, and the work function assigns two variables the handler reads:

```go
var syftSBOM *sbom.SBOM
...
	...
	syftSBOM, err = syft.CreateSBOM(context.Background(), src, cfg)
	if err != nil {
		return fmt.Errorf("failed to generate SBOM: %w", err)
	}
	return nil
})

switch {
case errors.Is(err, deadline.ErrTimedOut):
	return &pb.CreateSBOMResponse{Status: helpersv1.Incomplete, ...}, nil
case err == nil:
	// continue
...
```

`deadline.Run` cannot stop its work function, and says so:

> It does not (and cannot) simply kill the running function, so if it doesn't respect the stopper channel then it may keep running after the deadline passes.

The work function does not respect the stopper, and cannot: the note already in `CreateSBOM` records that Syft's cataloguers do not support cancellation ([anchore/syft#3705](https://github.com/anchore/syft/issues/3705)).

So on timeout the Syft goroutine keeps running and eventually assigns `syftSBOM` and `err`, after the handler has already read `err` and returned.

# Environment

- OS: N/A, concurrency defect
- Version: `main` (`39f669e`)

# Steps To Reproduce

With a `CreateSBOM` request whose `TimeoutSeconds` expires before cataloguing finishes, run under `-race`:

```
WARNING: DATA RACE
Write at 0x00c0004a6380 by goroutine 84:
```

# Expected behavior

Nothing is shared between the handler and a goroutine that can outlive the call.

# Actual Behavior

Two unsynchronised variables are shared. Beyond the race itself, the write to `err` can land before the handler reads it in the `switch`, in which case the handler takes `case err == nil` and falls through to `v1beta1.StripSBOM(syftSBOM)` and `syftToDomain(*syftSBOM)`. There is no happens-before edge ordering the two writes, so `syftSBOM` may still be nil there and the sidecar panics.

# Additional context

The existing tests do not catch this because none of them reaches the deadline path: `syft.CreateSBOM` is called directly, so a test cannot make cataloguing outlast the timeout without a genuinely slow image.

The success path is unaffected. `deadline.Run` receives the work function's return value over a channel, which orders the write before the handler's read.

AnshulPatil2005:
## Overview

`deadline.Run` cannot stop its work function, and documents that: on timeout it closes the stopper channel, returns `ErrTimedOut` and leaves the goroutine running. The work function cannot respect the stopper either, since Syft's cataloguers do not observe cancellation, which the note already in `CreateSBOM` records.

So on timeout the Syft goroutine keeps running and eventually executes:

```go
syftSBOM, err = syft.CreateSBOM(context.Background(), src, cfg)
```

assigning two variables the handler reads, after the handler has already read `err` and returned `Incomplete`.

Beyond the race itself, the write to `err` can land before the handler reads it in the `switch`. The handler then takes `case err == nil` and falls through to `v1beta1.StripSBOM(syftSBOM)` and `syftToDomain(*syftSBOM)`. Nothing orders the two writes, so `syftSBOM` may still be nil there and the sidecar panics.

This keeps the work function's result local and publishes it on a buffered channel that the handler receives from only once `dl.Run` has reported success, so nothing is shared with a goroutine that can outlive the call. The receive cannot block: `dl.Run` returns nil only after the work function returned nil, which it does after publishing.

Behaviour on both paths is unchanged. The success path was never racy, because `deadline.Run` passes the work function's return value over a channel, which orders the write before the handler's read.

## Additional Information

`syft.CreateSBOM` moves behind a `createSBOMFn` package variable so the deadline path is reachable from a test, mirroring the `sourceGetter` indirection this file already uses for `syft.GetSource`.

## How to Test

```
go test -race ./pkg/sbomscanner/v1/ -run TestCreateSBOM_TimeoutDoesNotRaceWithAbandonedSyft
```

The test serves a small image from the local mock registry the other tests in this file already use, substitutes a stand-in for Syft that outlives a 1 second deadline and then returns a result exactly as the real cataloguer would, and waits for that late completion before finishing so the write happens inside the window the race lived in.

Against the previous code it reports:

```
WARNING: DATA RACE
Write at 0x00c0004a6380 by goroutine 84:
--- FAIL: TestCreateSBOM_TimeoutDoesNotRaceWithAbandonedSyft
```

and it passes here. `go test -race ./pkg/sbomscanner/v1/` is clean.

## Related issues/PRs:

* Resolved #580

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved SBOM generation timeout handling to prevent late-running operations from altering completed responses.
  - Ensured timed-out requests consistently remain marked as incomplete, even after background processing finishes.

- **Tests**
  - Added regression coverage for timeout behavior and protection against race conditions during delayed SBOM generation.

### DIFF
--- a/pkg/sbomscanner/v1/server.go
+++ b/pkg/sbomscanner/v1/server.go
@@ -77,6 +77,10 @@ func formatResolvedPlatform(os, arch, variant string) string {
 	return strings.Join(parts, "/")
 }
 
+// createSBOMFn is an indirection over syft.CreateSBOM so the deadline handling in
+// CreateSBOM can be unit-tested without cataloguing a real image.
+var createSBOMFn = syft.CreateSBOM
+
 // sourceGetter abstracts the syft.GetSource call so the fallback ordering in
 // resolveSource is testable with a scripted implementation.
 type sourceGetter func(ctx context.Context, ref string, opts *image.RegistryOptions) (source.Source, error)
@@ -295,6 +299,9 @@ func (s *scannerServer) CreateSBOM(ctx context.Context, req *pb.CreateSBOMReques
 		timeout = 5 * time.Minute
 	}
 	dl := deadline.New(timeout)
+	// Buffered so the abandoned goroutine below can always publish and exit, even when
+	// nobody is left to receive.
+	generated := make(chan *sbom.SBOM, 1)
 	err = dl.Run(func(stopper <-chan struct{}) error {
 		defer func(src source.Source) {
 			if err := src.Close(); err != nil {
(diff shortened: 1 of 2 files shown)


matthyx (approved):
LGTM! The fix properly isolates the abandoned Syft goroutine using a 1-buffered channel and local variables, preventing data races on timeout. Verified that `TestCreateSBOM_TimeoutDoesNotRaceWithAbandonedSyft` passes under `-race`.

PR #585, opened 2026-08-10
fix(adapters): clamp the SBOM size limit sent to the sidecar scanner

### LINKED ISSUE #584: Bug: raising maxSBOMSize past 2 GiB makes the sidecar reject every SBOM as too large
# Description

Raising `maxSBOMSize` past 2 GiB stops the sidecar scanner returning any SBOM at all, and reports the opposite of the cause.

`config.MaxSBOMSize` is an `int`, but the scanner protocol carries the limit as an `int32` (`pkg/sbomscanner/v1/proto`, `max_sbom_size`). `SidecarSBOMAdapter.CreateSBOM` converts between them with a plain conversion:

```go
req := sbomscanner.ScanRequest{
	...
	MaxSBOMSize: int32(s.maxSBOMSize),
```

Once `maxSBOMSize` passes `math.MaxInt32` this wraps, and the wrapped value is compared against a real SBOM size in `scannerServer.CreateSBOM`:

```go
if sz > int(req.MaxSbomSize) {
	... Status: helpersv1.TooLarge, StatusReason: domain.ReasonSBOMTooLarge
```

| configured `maxSBOMSize` | value on the wire | effect |
|---|---|---|
| 2 GiB | -2147483648 | every SBOM is TooLarge, nothing is scanned |
| 3 GiB | -1073741824 | every SBOM is TooLarge, nothing is scanned |
| 4 GiB | 0 | every SBOM is TooLarge, nothing is scanned |
| 5 GiB | 1073741824 | a silent 1 GiB limit, far below what was asked for |
| 6 GiB | -2147483648 | every SBOM is TooLarge, nothing is scanned |

# Environment

- OS: N/A, configuration-dependent defect
- Version: `main` (`39f669e`)

# Steps To Reproduce

1. Run with the sidecar scanner enabled.
2. Set `maxSBOMSize` to `2147483648` (2 GiB) or above, for example to effectively remove the limit.
3. Scan any image.
4. Every scan returns `TooLarge` with reason `sbom-too-large`, and no SBOM is stored.

# Expected behavior

Raising the limit relaxes it. A limit larger than the protocol can express is clamped to the largest expressible value, not wrapped.

# Actual Behavior

The limit wraps to a negative or small value, so every SBOM is rejected as too large, and the failure is reported as `SBOM exceeds size limit`, which points an operator at the setting they just raised.

# Additional context

The in-process adapter compares against the `int` directly (`adapters/v1/syft.go`, `if sz > s.maxSBOMSize`) and has no such ceiling, so the same configuration behaves differently depending on which SBOM path is in use. That divergence is the same shape as #505, #512 and #557.

There is no validation on `MaxSBOMSize` in `config`, so nothing rejects or warns about the value at startup.

Separately, `scannerServer.CreateSBOM` treats a non-positive `MaxSbomSize` as a limit rather than as "no limit", unlike `TimeoutSeconds` immediately above it which falls back to a default when non-positive. A client that simply does not set the field therefore gets nothing scanned. That guard is worth adding on the server too, but it needs a seam to stub Syft in a test (see #580), so it is not included here.

AnshulPatil2005:
## Overview

`config.MaxSBOMSize` is an `int`, the scanner protocol carries the limit as an `int32`, and the sidecar adapter converted between them with a plain `int32(...)`. That wraps silently once the configured value passes `math.MaxInt32`, and the wrapped value is then compared against a real SBOM size:

| configured `maxSBOMSize` | on the wire | effect |
|---|---|---|
| 2 GiB | -2147483648 | every SBOM is TooLarge, nothing is scanned |
| 4 GiB | 0 | every SBOM is TooLarge, nothing is scanned |
| 5 GiB | 1073741824 | a silent 1 GiB limit, far below what was asked for |

So raising the limit breaks scanning instead of relaxing it, and the failure reports `SBOM exceeds size limit`, pointing the operator at the setting they just raised. The in-process adapter compares against the `int` directly and has no such ceiling, so the same configuration behaves differently on the two SBOM paths.

Clamp to `math.MaxInt32` and log when it happens, so the limit is at worst the largest value the protocol can express.

## Additional Information

Two related things are deliberately **not** in this PR:

`scannerServer.CreateSBOM` treats a non-positive `MaxSbomSize` as a limit rather than as "no limit", unlike `TimeoutSeconds` immediately above it, which falls back to a default when non-positive. A client that simply does not set the field gets nothing scanned. Adding that guard needs a seam to stub Syft so the size check is actually reached in a test; without one the test passes vacuously, because cataloguing a synthetic image returns `Incomplete` before the size check runs. #581 adds exactly that seam, so this is worth doing as a follow-up on top of it rather than shipping untested here.

There is also no validation on `MaxSBOMSize` in `config`, so nothing warns at startup. The clamp makes the runtime behaviour safe regardless, but a startup warning would surface the mistake sooner.

## How to Test

```
go test ./adapters/v1/ -run TestSBOMSizeLimitForWire
```

Seven cases: the 20 MiB default and zero pass through unchanged, `math.MaxInt32` is the boundary that must not clamp, and 2/4/5 GiB are the wrap cases. Four of them fail against the previous conversion (verified by reverting the clamp). Each positive input is additionally asserted to stay positive on the wire, which is the property the bug violated.

## Related issues/PRs:

* Resolved #584

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented excessively large SBOM size limits from causing invalid values during scanner communication.
  * Preserved zero, default, and valid in-range size limits while safely capping oversized values.

### DIFF
--- a/adapters/v1/sidecar.go
+++ b/adapters/v1/sidecar.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math"
 	"sync"
 	"time"
 
@@ -26,6 +27,27 @@ const (
 	// maxCrashRetries was hit (see #473).
 	retryCountTTL = 1 * time.Hour
 )
+
+// sbomSizeLimitForWire narrows the configured SBOM size limit to the int32 the scanner
+// protocol carries.
+//
+// A plain conversion silently wraps once maxSBOMSize passes math.MaxInt32, and the wrapped
+// value is then compared against a real SBOM size: 2 GiB becomes negative and 4 GiB becomes
+// zero, so every SBOM is classified TooLarge and no image is scanned at all, while 5 GiB
+// becomes 1 GiB and quietly enforces a limit far below the configured one. Raising the limit
+// is exactly when an operator would least expect scans to start failing, and the in-process
+// adapter has no such ceiling, so the same configuration behaves differently on the two paths.
+//
+// Clamp instead, so the limit is at worst the largest the protocol can express.
+func sbomSizeLimitForWire(maxSBOMSize int) int32 {
+	if maxSBOMSize > math.MaxInt32 {
+		logger.L().Warning("configured maxSBOMSize exceeds the scanner protocol limit, clamping",
+			helpers.Int("maxSBOMSize", maxSBOMSize),
+			helpers.Int("clamped", math.MaxInt32))
+		return math.MaxInt32
+	}
+	return int32(maxSBOMSize)
+}
 
 // crashRetry tracks a per-image crash-retry count and when it was last touched, so stale
 // entries (images that crashed once and were never retried) can be evicted.
@@ -108,7 +130,7 @@ func (s *SidecarSBOMAdapter) CreateSBOM(ctx context.Context, name, imageID, imag
 		ImageTag:            pullImageTag,
 		Options:             options,
 		MaxImageSize:        s.maxImageSize,
-		MaxSBOMSize:         int32(s.maxSBOMSize),
+		MaxSBOMSize:         sbomSizeLimitForWire(s.maxSBOMSize),
 		EnableEmbeddedSBOMs: s.scanEmbeddedSBOMs,
 		Timeout:             s.scanTimeout,
 	}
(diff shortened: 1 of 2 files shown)


matthyx (approved):
LGTM! Clamping  to  for wire conversion prevents silent overflow/wrapping when configured size exceeds protocol limits. Unit tests cover boundary and overflow cases cleanly.

PR #587, opened 2026-08-10
fix(sbomscanner): round the scan timeout up when sending it to the sidecar

### LINKED ISSUE #586: Bug: a sub-second scanTimeout makes the sidecar wait five minutes instead
# Description

A sub-second `scanTimeout` makes the sidecar scanner run for five minutes instead, the opposite of the setting.

The scanner protocol carries the scan timeout as whole seconds, and the client truncates to get there (`pkg/sbomscanner/v1/client.go`):

```go
TimeoutSeconds: int64(req.Timeout.Seconds()),
```

The server treats a non-positive value as unset (`pkg/sbomscanner/v1/server.go`):

```go
timeout := time.Duration(req.TimeoutSeconds) * time.Second
if timeout <= 0 {
	timeout = 5 * time.Minute
}
```

So a timeout below one second truncates to zero and is read as "no timeout configured".

| configured `scanTimeout` | on the wire | deadline actually used |
|---|---|---|
| `500ms` | 0 | **5 minutes** |
| `1500ms` | 1 | 1 second |
| `2999ms` | 2 | 2 seconds |
| `90s` | 90 | 90 seconds |

# Environment

- OS: N/A, configuration-dependent defect
- Version: `main` (`39f669e`)

# Steps To Reproduce

1. Run with the sidecar scanner enabled.
2. Set `scanTimeout` below one second, for example `500ms`, to make scans fail fast.
3. Scan an image that takes longer than that.
4. The scan is not aborted at 500ms. It runs until the server's five minute default.

# Expected behavior

A configured timeout is honoured, or at worst rounded up to the nearest second the protocol can express. A positive timeout never reads as unset.

# Actual Behavior

A sub-second timeout becomes a five minute one, and any fractional timeout is silently shortened to the second below.

# Additional context

The in-process adapter passes the same value straight to `deadline.New(s.scanTimeout)` (`adapters/v1/syft.go`) and honours it exactly, including sub-second values, so one config value behaves very differently depending on which SBOM path is in use. That divergence is the same shape as #505, #512, #557 and #584.

The default `scanTimeout` is five minutes, so a default deployment is unaffected. This shows up when the value is lowered, which is exactly when someone is trying to bound scan time.

AnshulPatil2005:
## Overview

The scanner protocol carries the scan timeout as whole seconds, and the client truncated to get there:

```go
TimeoutSeconds: int64(req.Timeout.Seconds()),
```

The server treats a non-positive `TimeoutSeconds` as unset and substitutes a five minute default, so truncation inverts the setting at the low end:

| configured `scanTimeout` | on the wire | deadline actually used |
|---|---|---|
| `500ms` | 0 | **5 minutes** |
| `1500ms` | 1 | 1 second |
| `90s` | 90 | 90 seconds |

A sub-second timeout therefore produces a much longer deadline than the one configured, not a shorter one, and fractional values are quietly shortened.

The in-process adapter passes the same `scanTimeout` straight to `deadline.New` and honours it exactly, so one config value behaved very differently depending on which SBOM path was in use.

This rounds up instead, so the wire value never shortens the configured timeout and a positive timeout never reads as unset. A non-positive timeout still passes through as zero, which is the value that legitimately means "use the server default".

## How to Test

```
go test ./pkg/sbomscanner/v1/ -run TestTimeoutSecondsForWire
```

Eight cases, five of which fail against the previous truncation (verified by reverting). Alongside the expected values, each positive input is asserted to reach the wire as non-zero and to be no shorter than what was configured, which are the two properties the bug violated.

The five minute default and whole-second values are covered too, so the common path is pinned as exact rather than merely non-broken.

## Related issues/PRs:

* Resolved #586

### DIFF
--- a/pkg/sbomscanner/v1/client.go
+++ b/pkg/sbomscanner/v1/client.go
@@ -88,6 +88,29 @@ func NewSBOMScannerClient(ctx context.Context, socketPath string, readinessTimeo
 	return c, nil
 }
 
+// timeoutSecondsForWire converts a scan timeout to the whole seconds the scanner protocol
+// carries, rounding up so a configured timeout is never shortened.
+//
+// Truncating instead inverts the setting at the low end: the server treats a non-positive
+// TimeoutSeconds as unset and falls back to a five minute default, so a sub-second timeout
+// truncated to zero produces a far longer deadline than the one configured, not a shorter
+// one. The in-process adapter passes the same scanTimeout straight to deadline.New and
+// honours it exactly, so the two SBOM paths would otherwise read one config value very
+// differently.
+//
+// A non-positive timeout is passed through as zero, which is the value that legitimately
+// means "use the server default".
+func timeoutSecondsForWire(timeout time.Duration) int64 {
+	if timeout <= 0 {
+		return 0
+	}
+	seconds := timeout / time.Second
+	if timeout%time.Second != 0 {
+		seconds++
+	}
+	return int64(seconds)
+}
+
 func (c *sbomScannerClient) CreateSBOM(ctx context.Context, req ScanRequest) (*ScanResult, error) {
 	// Map domain credentials to proto
 	creds := make([]*pb.RegistryCredentials, len(req.Options.Credentials))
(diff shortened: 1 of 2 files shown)


matthyx (approved):
LGTM! Rounding up fractional scan timeouts prevents sub-second values from truncating to zero (which previously caused the sidecar server to fall back to the 5-minute default) and avoids shortening configured deadlines. Unit tests pass and cover edge cases.

PR #590, opened 2026-08-10
fix(tools): stop SanitizeLabel producing labels that get dropped

### LINKED ISSUE #589: Bug: an image with an uppercase tag silently loses its kubescape.io labels
# Description

An image whose tag contains uppercase loses its `kubescape.io` image labels entirely, so it cannot be found by label filtering.

`SanitizeLabel` (`internal/tools/tools.go`) is documented as returning a valid DNS1123 label, and `LabelsFromImageID` deletes any label that fails validation:

```go
// prune invalid labels
for key, value := range labels {
	if errs := validation.IsDNS1123Label(value); len(errs) != 0 {
		delete(labels, key)
	}
}
```

So anything `SanitizeLabel` leaves invalid is dropped silently rather than stored in a degraded form. Two ordinary inputs produce one:

**Uppercase.** A DNS1123 label must be lowercase, but the reference grammar allows uppercase in a tag (`[a-zA-Z0-9_][a-zA-Z0-9._-]{0,127}`). `SanitizeLabel` never lowercases:

```
"myapp:v1.0-RC1"  ->  "myapp-v1-0-RC1"   invalid, label dropped
```

**Dashes at the ends.** A DNS1123 label must start and end alphanumeric. Adjacent offending characters each become a dash, so `v1_.` becomes `v1--`, and only a single trailing dash is stripped:

```
"registry.io/team/app:v1_."  ->  "registry-io-team-app-v1-"   invalid, label dropped
```

Truncation at 63 characters can land on a dash for the same reason, and a leading offending character produces a leading dash, which is not handled at all.

# Environment

- OS: N/A
- Version: `main` (`39f669e`)

# Steps To Reproduce

1. Scan an image with an uppercase tag, for example `registry.io/team/app:v1.0-RC1`.
2. Inspect the resulting SBOM or vulnerability manifest.
3. The `kubescape.io/image-tag` label is absent, and `kubectl get ... -l kubescape.io/image-tag=...` cannot find the object.

# Expected behavior

`SanitizeLabel` returns a valid DNS1123 label for any input, so the label is stored (lowercased and with offending characters replaced) rather than dropped.

# Actual Behavior

The label is silently absent, with nothing logged.

# Additional context

Release tags carrying uppercase are common (`v1.0-RC1`, `Ubuntu20.04`), so this is not an unusual input.

No code in this repository selects on these labels, so nothing breaks internally. The impact is on operators and external tooling filtering by them, and on the objects being inconsistently labelled depending on how an image happens to be tagged.

AnshulPatil2005:
## Overview

`SanitizeLabel` is documented as returning a valid DNS1123 label, and `LabelsFromImageID` deletes any label that fails validation. So anything `SanitizeLabel` leaves invalid is dropped silently rather than stored in a degraded form, and two ordinary inputs produced one:

```
"myapp:v1.0-RC1"             ->  "myapp-v1-0-RC1"            invalid, label dropped
"registry.io/team/app:v1_."  ->  "registry-io-team-app-v1-"  invalid, label dropped
```

A DNS1123 label must be lowercase, but the reference grammar allows uppercase in a tag, so a tag like `v1.0-RC1` lost its `image-tag` label entirely. A DNS1123 label must also start and end alphanumeric, and adjacent offending characters each become a dash, so `v1_.` became `v1--` and stripping a single trailing dash left one behind. Truncation at 63 characters can land on a dash for the same reason, and a leading offending character produces a leading dash, which was never handled.

Lowercase before substituting, and trim dashes from both ends rather than removing one.

## Additional Information

Lowercasing changes the stored value for inputs that carried uppercase, but the alternative is the current behaviour of storing nothing at all, and DNS1123 leaves no way to preserve case. Two tags differing only in case now map to the same label value; that is acceptable here because nothing selects on these labels for identity, object names carry that.

No code in this repository selects on these labels, so nothing changes internally. The impact is on operators and external tooling filtering by them.

## How to Test

```
go test ./internal/tools/ -run 'TestSanitizeLabel|TestLabelsFromImageID'
```

`TestSanitizeLabel_AlwaysProducesAValidLabel` covers the uppercase and trailing-dash inputs above, plus leading offending characters, interior dashes that must be preserved, and the all-offending and empty inputs. Every non-empty result is additionally asserted to pass `validation.IsDNS1123Label`, which is the contract that was being broken rather than just the expected strings.

`TestSanitizeLabel_TruncationNeverEndsOnADash` covers truncation landing on a separator, and `TestLabelsFromImageID_UppercaseTagIsKept` covers the end-to-end case where the label used to disappear.

Eight of these fail against the previous implementation (verified by reverting). The pre-existing `TestLabelsFromImageID` is unchanged and still passes.

## Related issues/PRs:

* Resolved #589

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved label normalization by converting text to lowercase, replacing invalid characters, enforcing length limits, and removing leading or trailing dashes.
  * Ensured generated labels remain valid DNS-compatible labels, including labels derived from uppercase image tags.

* **Tests**
  * Added coverage for normalization, truncation, empty results, DNS validity, and uppercase image tags.

### DIFF
--- a/internal/tools/tools.go
+++ b/internal/tools/tools.go
@@ -34,13 +34,21 @@ func PackageVersion(name string) string {
 var offendingChars = regexp.MustCompile("[@:/ ._]")
 
 // SanitizeLabel sanitizes a string to be a valid DNS1123 label.
+//
+// Lowercasing and trimming both ends matter because LabelsFromImageID drops any label that
+// fails validation, so anything this leaves invalid disappears silently rather than being
+// stored in a degraded form:
+//
+//   - A DNS1123 label must be lowercase, while an image tag may legally contain uppercase
+//     (the reference grammar allows [a-zA-Z0-9_][a-zA-Z0-9._-]*), so a tag such as v1.0-RC1
+//     produced an invalid label.
+//   - A DNS1123 label must start and end alphanumeric, and adjacent offending characters
+//     collapse into several dashes (v1_. becomes v1--), so stripping a single trailing dash
+//     was not enough. Truncation at 63 can also land on a dash, and a leading offending
+//     character produces a leading one.
 func SanitizeLabel(s string) string {
-	s2 := truncate.Truncate(offendingChars.ReplaceAllString(s, "-"), 63, "", truncate.PositionEnd)
-	// remove trailing dash
-	if len(s2) > 0 && s2[len(s2)-1] == '-' {
-		return s2[:len(s2)-1]
-	}
-	return s2
+	s2 := truncate.Truncate(offendingChars.ReplaceAllString(strings.ToLower(s), "-"), 63, "", truncate.PositionEnd)
+	return strings.Trim(s2, "-")
 }
 
 // LabelsFromImageID returns a map of labels from an image ID.
--- a/internal/tools/tools.go
+++ b/internal/tools/tools.go
@@ -31,7 +31,12 @@ func PackageVersion(name string) string {
 	return "unknown"
 }
 
-var offendingChars = regexp.MustCompile("[@:/ ._]")
+// offendingChars matches everything a DNS1123 label may not contain, rather than an
+// enumerated set. Listing the characters seen in image references ("[@:/ ._]") left anything
+// unlisted to pass through unchanged and fail validation, so the label was dropped: "foo+bar"
+// and non-ASCII input both survived substitution intact. Applied after lowercasing, so no
+// uppercase reaches it.
+var offendingChars = regexp.MustCompile("[^a-z0-9-]")
 
 // SanitizeLabel sanitizes a string to be a valid DNS1123 label.
 //
--- a/internal/tools/tools_test.go
+++ b/internal/tools/tools_test.go
@@ -1,8 +1,10 @@
 package tools
 
 import (
+	"k8s.io/apimachinery/pkg/util/validation"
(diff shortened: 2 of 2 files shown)


matthyx:
`SanitizeLabel` currently uses `var offendingChars = regexp.MustCompile("[@:/ ._]")`, which only replaces `@`, `:`, `/`, ` `, `.`, and `_`.

However, container image tags can contain other characters invalid in DNS-1123 labels, such as `+` (e.g. semver build metadata `v1.0.0+build1`) or `~` (e.g. Debian package versions `1.0.0~deb11`).

For example:
- `SanitizeLabel("v1.0.0+build1")` returns `"v1-0-0+build1"`.
- `validation.IsDNS1123Label("v1-0-0+build1")` fails because of `+`.
- `LabelsFromImageID` deletes the invalid label, so the label is still dropped.

**Suggested Fix:**
Update `offendingChars` to match any non-DNS1123 character (after lowercasing):
```go
var offendingChars = regexp.MustCompile("[^a-z0-9-]")
```
This ensures all non-DNS1123 characters are replaced with hyphens so `SanitizeLabel` consistently returns valid DNS1123 labels.

matthyx (approved):
LGTM!  now properly replaces all non-DNS1123 characters with hyphens.

PR #593, opened 2026-08-11
fix(config): honor env-only trustedVendors/proxyRegistryMap overrides

### LINKED ISSUE #588: Bug: trustedVendors and proxyRegistryMap env-var overrides are silently ignored
## Description

`docs/CONFIGURATION.md` states that "All options can be overridden via environment variables using UPPER_SNAKE_CASE (e.g., `scanTimeout` → `SCANTIMEOUT`)". For `trustedVendors` and `proxyRegistryMap` this isn't true: setting them purely via environment variable (no matching key in `clusterData.json`) is silently ignored, and the resulting behavior differs between the two fields in a way that makes the bug hard to notice.

## Root cause

`LoadConfig` in `config/config.go` relies on `viper.AutomaticEnv()` + `v.Unmarshal(&config)` to populate `Config`. Viper's `Unmarshal` does not pick up env-only values for slice/map struct fields (`[]string`, `map[string]string`) — only `IsSet`/`GetString`/etc. see them. The code already documents this exact pitfall for `cveMatchingMode`:

```go
// Read through viper's getters rather than the unmarshalled struct: with
// AutomaticEnv, a value supplied purely via environment variable is visible
// to IsSet/GetString but is NOT populated by Unmarshal, so relying on the
// struct field here would drop env overrides (and, for the mode, fail
// validation below).
switch {
case v.IsSet("cveMatchingMode"):
    config.CVEMatchingMode = CVEMatchingMode(v.GetString("cveMatchingMode"))
...
```

but the same workaround was never applied to `TrustedVendors` or `ProxyRegistryMap`, which are still read straight off the unmarshalled struct:

```go
TrustedVendors   []string          `mapstructure:"trustedVendors"`
ProxyRegistryMap map[string]string `mapstructure:"proxyRegistryMap"`
...
if len(config.TrustedVendors) == 0 {
    config.TrustedVendors = append([]string{}, defaultTrustedVendors...)
}
```

So:
- `TRUSTEDVENDORS=mycompany-distro` (env-only, no `clusterData.json` key) → `config.TrustedVendors` stays `nil` after `Unmarshal` → the `len(...) == 0` check silently overwrites it with `defaultTrustedVendors`. No error, no log — the operator's override is dropped and the default is used instead.
- `PROXYREGISTRYMAP={"docker.io":"my-mirror.example.com"}` (env-only) → `config.ProxyRegistryMap` stays `nil`/empty, with no fallback logic at all, so registry proxy mirroring silently never activates.

## Reproduction

```go
v := viper.NewWithOptions(viper.KeyDelimiter("::"))
v.AutomaticEnv()
os.Setenv("TRUSTEDVENDORS", "foo,bar")
os.Setenv("PROXYREGISTRYMAP", `{"a":"b"}`)

fmt.Println(v.IsSet("trustedVendors"))       // true
fmt.Println(v.GetStringSlice("trustedVendors")) // ["foo,bar"] (visible)

var cfg Config
v.Unmarshal(&cfg)
fmt.Println(cfg.TrustedVendors)   // nil  <-- lost
fmt.Println(cfg.ProxyRegistryMap) // nil  <-- lost
```

(No `clusterData.json` present, matching a common deployment where these two options are only set via env vars on the container.)

## Expected behavior

Setting `TRUSTEDVENDORS` or `PROXYREGISTRYMAP` as environment variables (as documented) should take effect, consistent with how `CVEMATCHINGMODE` already works.

## Actual behavior

Both env vars are silently dropped:
- `TrustedVendors` quietly reverts to `defaultTrustedVendors` — CPE-matching behavior for adaptive mode is not what the operator configured, with no indication anything went wrong.
- `ProxyRegistryMap` stays empty — registry mirroring never activates, images are pulled from their original registries instead of the configured mirror.

## Environment

- Repo: kubescape/kubevuln, `main` @ `0ee5bd2`
- `config/config.go`, `Config` struct (lines ~74, 84) and `LoadConfig` (lines ~159-162)

## Suggested fix

Extend the same `v.IsSet(...)`/`v.Get...(...)` pattern already used for `cveMatchingMode` to `trustedVendors` and `proxyRegistryMap`, so env-only values are read through viper's getters rather than the unmarshalled struct before falling back to defaults.

bhuvan-somisetty:
## Overview

`viper.AutomaticEnv()` makes an env-only value visible via `IsSet`/`Get`, but `Unmarshal` does not populate slice/map struct fields from it. This gap is already documented and worked around for `cveMatchingMode` in `LoadConfig`, but the same workaround was never applied to `trustedVendors`/`proxyRegistryMap`:

- `TRUSTEDVENDORS` set purely via env var was silently dropped, and `TrustedVendors` quietly reverted to `defaultTrustedVendors`.
- `PROXYREGISTRYMAP` set purely via env var was silently dropped, and `ProxyRegistryMap` stayed empty (no fallback at all) — registry mirroring never activated.

Both are now resolved through viper's getters when `IsSet`, matching the existing `cveMatchingMode` pattern:
- `TRUSTEDVENDORS` is parsed as a comma-separated list (`GetStringSlice` does not split env-var strings, only JSON-file-sourced arrays).
- `PROXYREGISTRYMAP` is parsed as a JSON object string via `GetStringMapString` (which already handles both the JSON-file map shape and a JSON-string env var).

Docs updated with the expected env-var formats for both fields.

## How to Test

- `go build ./...`
- `go vet ./...`
- `golangci-lint run --new-from-rev upstream/main` — 0 issues
- `go test ./...` — all passing (the one pre-existing failure, `TestScanService_NginxTest`, requires a Docker-pulled Grype DB fixture and is unrelated to this change, which only touches `config/`)
- New tests: `TestLoadConfigTrustedVendorsEnv`, `TestLoadConfigProxyRegistryMapEnv` — verify env-only overrides are honored instead of silently dropped

## Related issues/PRs

* Resolved #588

- [x] If it is a core feature, I have added thorough tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Environment-variable configuration for trusted vendors and proxy registry mappings is now loaded correctly.
  * Trusted vendor values support comma-separated entries with whitespace trimming.
  * Default trusted vendors continue to apply when no values are configured.

* **Documentation**
  * Added guidance for configuring trusted vendors and proxy registry mappings through environment variables.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -156,12 +156,44 @@ func LoadConfig(path string) (Config, error) {
 			config.CVEMatchingMode, CVEMatchingOff, CVEMatchingOn, CVEMatchingAdaptive)
 	}
 
+	// Same AutomaticEnv/Unmarshal gap as cveMatchingMode above, but for slice/map
+	// fields: Unmarshal silently drops env-only values for []string and
+	// map[string]string, so an env-only TRUSTEDVENDORS/PROXYREGISTRYMAP would
+	// otherwise be lost. Read through viper's getters when the field was
+	// actually set (via file or env) rather than trusting the unmarshalled
+	// struct field alone.
+	if v.IsSet("trustedVendors") {
+		config.TrustedVendors = trustedVendorsFromViper(v)
+	}
 	if len(config.TrustedVendors) == 0 {
 		// copy to avoid aliasing the package-level default slice
 		config.TrustedVendors = append([]string{}, defaultTrustedVendors...)
 	}
 
+	if v.IsSet("proxyRegistryMap") {
+		config.ProxyRegistryMap = v.GetStringMapString("proxyRegistryMap")
+	}
+
 	return config, nil
+}
+
+// trustedVendorsFromViper resolves trustedVendors regardless of whether it came
+// from the JSON config file (already a []interface{}, handled fine by
+// GetStringSlice) or from a TRUSTEDVENDORS environment variable (a plain string,
+// which GetStringSlice does not split): env values are a comma-separated list,
+// e.g. TRUSTEDVENDORS=echo,chainguard.
+func trustedVendorsFromViper(v *viper.Viper) []string {
+	raw, ok := v.Get("trustedVendors").(string)
+	if !ok {
+		return v.GetStringSlice("trustedVendors")
+	}
+	var vendors []string
+	for _, part := range strings.Split(raw, ",") {
+		if part = strings.TrimSpace(part); part != "" {
+			vendors = append(vendors, part)
+		}
+	}
+	return vendors
 }
 
 type clusterDataBackendServicesConfig struct {
--- a/config/config.go
+++ b/config/config.go
@@ -171,7 +171,11 @@ func LoadConfig(path string) (Config, error) {
 	}
 
 	if v.IsSet("proxyRegistryMap") {
-		config.ProxyRegistryMap = v.GetStringMapString("proxyRegistryMap")
+		proxyMap, err := proxyRegistryMapFromViper(v)
+		if err != nil {
+			return Config{}, err
+		}
+		config.ProxyRegistryMap = proxyMap
 	}
 
 	return config, nil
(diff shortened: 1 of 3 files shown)


matthyx (approved):
LGTM! Verified that env-only overrides for `TRUSTEDVENDORS` (comma-separated list) and `PROXYREGISTRYMAP` (JSON object string with invalid JSON error handling) work as expected and all unit tests pass.

PR #599, opened 2026-08-11
Add diagnostics endpoint for scan mode and backend versions

### LINKED ISSUE #598: Add a diagnostic endpoint reporting active scan mode and backend versions
## Overview
The current `/v1/liveness` and `/v1/readiness` endpoints only report coarse health. They do not say whether kubevuln resolved the sidecar SBOM scanner path or fell back to the in-process Syft path, and they do not surface the active Syft/Grype library versions or the current Grype vulnerability DB version.

## Problem
When scan behavior changes after startup (for example the sidecar readiness check times out and the process falls back to in-process Syft, or the vulnerability DB is refreshed by the background updater), operators currently have to correlate log lines to find out. There is no way to query the running pod directly for this information.

## Solution
Add a dedicated diagnostic endpoint, separate from readiness/liveness, that reports:
- Active SBOM scan mode (sidecar vs in-process)
- Syft/SBOM creator version
- Grype/CVE scanner version
- Grype vulnerability DB version
- Configured scan timeout and scanner readiness timeout
- Whether storage and risk-acceptance (SecurityException CRD integration) are enabled

This should reuse values already computed in `cmd/http/main.go` at startup (sidecar vs Syft selection, adapter construction) and already exposed by existing adapter methods (`Version()`, `DBVersion()` on `ports.SBOMCreator`/`ports.CVEScanner`), without changing scan semantics.

## Alternatives
Rely on startup logs to infer mode and versions. This only works if logs are retained and easy to correlate with a specific pod/restart, and it does not reflect state that changes later (e.g. a DB refresh).

## Additional context
- The CVE DB version is refreshed periodically by a background updater in the Grype adapter, so this should be read live rather than cached once at startup.
- The response must not include credentials or other secret-derived configuration.

Acceptance criteria:
- The endpoint reports sidecar vs in-process mode and scanner versions.
- Sensitive credentials are excluded from the response.
- Tests verify stable JSON output for at least one configuration.

rootp1:
## Overview

Adds a new `GET /v1/diagnostics` endpoint, separate from `/v1/liveness` and `/v1/readiness`, that reports the SBOM scan configuration actually in effect on a running pod: whether the sidecar or in process Syft path was resolved at startup, the Syft and Grype library versions, the current Grype vulnerability DB version, the configured scan and scanner readiness timeouts, and whether storage and risk acceptance (SecurityException CRD integration) are enabled. Today this information is only available by reading startup logs.

The CVE DB version is refreshed periodically in the background by the Grype adapter, so the handler reads it live on every request instead of caching a value from startup. The response never includes credentials or other secret derived configuration.

The values are collected in `cmd/http/main.go` where the SBOM adapter (sidecar vs Syft) and the Grype adapter are already constructed, and passed into `HTTPController` through a small `WithDiagnostics` callback, mirroring the existing `WithMetrics` pattern. The handler itself just serializes the resulting `domain.Diagnostics` struct as JSON.

Run the service locally and curl the new endpoint:

```
CONFIG_DIR=.dev/config go run ./cmd/http
curl http://localhost:8080/v1/diagnostics
```

* Resolved #598

## Summary by CodeRabbit

* **New Features**
  * Added a diagnostics endpoint at `/v1/diagnostics`.
  * Exposes SBOM and CVE scan modes, component versions, timeout settings, storage configuration, and risk-acceptance settings.
  * Diagnostics exclude credentials and other secret-derived values.
  * Returns an empty diagnostics response when no configuration is available.

### DIFF
--- a/cmd/http/main.go
+++ b/cmd/http/main.go
@@ -19,6 +19,7 @@ import (
 	v1 "github.com/kubescape/kubevuln/adapters/v1"
 	"github.com/kubescape/kubevuln/config"
 	"github.com/kubescape/kubevuln/controllers"
+	"github.com/kubescape/kubevuln/core/domain"
 	"github.com/kubescape/kubevuln/core/ports"
 	"github.com/kubescape/kubevuln/core/services"
 	"github.com/kubescape/kubevuln/internal/metrics"
@@ -78,6 +79,7 @@ func main() {
 		}
 	}
 	var sbomAdapter ports.SBOMCreator
+	scanMode := domain.ScanModeInProcess
 	if socketPath := os.Getenv("SBOM_SCANNER_SOCKET"); socketPath != "" {
 		logger.L().Info("connecting to SBOM scanner sidecar", helpers.String("socket", socketPath))
 		scannerClient, err := sbomscanner.NewSBOMScannerClient(ctx, socketPath, c.ScannerReadinessTimeout)
@@ -92,6 +94,7 @@ func main() {
 		} else {
 			memoryLimit := os.Getenv("SCANNER_MEMORY_LIMIT")
 			sbomAdapter = v1.NewSidecarSBOMAdapter(scannerClient, c.ScanTimeout, c.MaxImageSize, c.MaxSBOMSize, c.ScanEmbeddedSboms, memoryLimit, c.ProxyRegistryMap)
+			scanMode = domain.ScanModeSidecar
 		}
 	} else {
 		sbomAdapter = v1.NewSyftAdapter(c.ScanTimeout, c.MaxImageSize, c.MaxSBOMSize, c.ScanEmbeddedSboms, c.ProxyRegistryMap)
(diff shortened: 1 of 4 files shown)


matthyx (approved):
Reviewed the diagnostics endpoint addition — clean implementation, builds and tests pass (`go build ./...`, `go test ./controllers/...`).

One accuracy nit worth fixing (non-blocking, matches CodeRabbit's finding): `RiskAcceptanceEnabled` at `cmd/http/main.go:169` reports `c.RiskAcceptance` directly, but the SecurityException CRD integration is only actually enabled when `storage != nil && c.RiskAcceptance` (see line 104). If `riskAcceptance` config is set but `storage` is disabled/failed, the diagnostics endpoint would misleadingly report risk acceptance as enabled when it's a no-op (per the existing warning log at line ~124). Suggest reporting `c.Storage && c.RiskAcceptance` there for consistency with the actual gating logic.

Everything else — the `WithDiagnostics` callback pattern (mirrors `WithMetrics`), live `DBVersion()` read instead of startup caching, nil-safe handler default, and test coverage — looks solid. Approving; the nit above can be a quick follow-up commit or separate PR.

PR #583, opened 2026-08-10
fix: respect parent context and apply scanTimeout in SyftAdapter.CreateSBOM

NaitikRishu:
**Description:**

Derive image resolution and SBOM generation context from the parent `ctx` and apply `s.scanTimeout` using `context.WithTimeout`. If context deadline is exceeded during image download, mark status as `Incomplete` instead of hanging indefinitely.

## Overview
- **Current behavior:** `SyftAdapter.CreateSBOM()` in `adapters/v1/syft.go` called `context.WithValue(context.Background(), ...)` and `syft.CreateSBOM(context.Background(), ...)`. This completely dropped the parent `ctx` (ignoring parent cancellations like SIGTERM) and failed to apply `s.scanTimeout` during image source resolution. If a registry connection stalled, worker threads would block indefinitely.
- **Future behavior:** Context is derived from the parent `ctx` and respects `s.scanTimeout` via `context.WithTimeout`. If context deadline is exceeded during image download, the SBOM status is set to `helpersv1.Incomplete` cleanly.

## How to Test
1. Run `go test -run Test_syftAdapter_CreateSBOM_CanceledContext ./adapters/v1` to verify immediate cancellation when parent context is canceled.
2. Run `go test -run Test_syftAdapter_CreateSBOM ./adapters/v1` to verify timeout behavior.

## Related issues/PRs:
* Resolved #582

## Checklist before requesting a review

- [x] My code follows the style guidelines of this project
- [x] I have commented on my code, particularly in hard-to-understand areas
- [x] I have performed a self-review of my code
- [x] If it is a core feature, I have added thorough tests.
- [x] New and existing unit tests pass locally with my changes

*(Note: Remember to open the PR against the `dev` branch on `kubescape/kubevuln`!)*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scan cancellation and timeout handling.
  * Incomplete scans caused by image-resolution or generation timeouts are now reported consistently without misleading errors.
  * Scan status, metrics, and logs now better reflect interrupted or timed-out operations.

* **Tests**
  * Added coverage for already-canceled requests and expired scan timeouts.

PR #601, opened 2026-08-11
fix(adapters): stop the abandoned Syft goroutine racing CreateSBOM

### LINKED ISSUE #600: Bug: data race between CreateSBOM and the abandoned Syft goroutine on the in-process path
# Description

`SyftAdapter.CreateSBOM` (`adapters/v1/syft.go`) has the same defect #581 fixed on the sidecar path.

The work function passed to `deadline.Run` assigns two variables the caller reads:

```go
var syftSBOM *sbom.SBOM
...
	...
	syftSBOM, err = syft.CreateSBOM(ctxWithSize, src, cfg)
	if err != nil {
		return fmt.Errorf("failed to generate SBOM: %w", err)
	}
	return nil
})

switch {
case errors.Is(err, deadline.ErrTimedOut) || ...:
	domainSBOM.Status = helpersv1.Incomplete
	return domainSBOM, nil
case err == nil:
	// continue
```

`deadline.Run` cannot stop its work function, and says so:

> It does not (and cannot) simply kill the running function, so if it doesn't respect the stopper channel then it may keep running after the deadline passes.

The work function does not respect the stopper and cannot: Syft's cataloguers do not support cancellation ([anchore/syft#3705](https://github.com/anchore/syft/issues/3705)). Passing a cancellable context, as #583 now does, does not change that.

So on timeout the Syft goroutine keeps running and eventually assigns `syftSBOM` and `err`, after the caller has already read `err` and returned.

# Environment

- OS: N/A, concurrency defect
- Version: `main` (`7b33df6`)

# Steps To Reproduce

Drive `CreateSBOM` with a `scanTimeout` shorter than cataloguing takes, and run under `-race`:

```
WARNING: DATA RACE
Write at 0x00c000880850 by goroutine 71:
```

# Expected behavior

Nothing is shared between the caller and a goroutine that can outlive the call.

# Actual Behavior

Two unsynchronised variables are shared. Beyond the race itself, the write to `err` can land before the caller reads it in the `switch`, in which case the caller takes `case err == nil` and continues to `v1beta1.StripSBOM(syftSBOM)` and `s.syftToDomain(*syftSBOM)`. Nothing orders the two writes, so `syftSBOM` may still be nil there and the scan panics.

# Additional context

`go test -race` does not catch this today because no existing test reaches the deadline path: `syft.CreateSBOM` is called directly, so a test cannot make cataloguing outlast the timeout without a genuinely slow image.

The success path is unaffected. `deadline.Run` receives the work function's return value over a channel, which orders the write before the caller's read.

This is the in-process half of #580.

AnshulPatil2005:
## Overview

The in-process adapter has the same defect #581 fixed on the sidecar path.

`deadline.Run` cannot stop its work function and documents that, and Syft's cataloguers do not observe cancellation, so on timeout the Syft goroutine keeps running and eventually executes:

```go
syftSBOM, err = syft.CreateSBOM(ctxWithSize, src, cfg)
```

assigning two variables the caller reads, after the caller has already read `err` and returned `Incomplete`.

Beyond the race itself, the write to `err` can land before the caller reads it in the `switch`. The caller then takes `case err == nil` and continues to `v1beta1.StripSBOM(syftSBOM)` and `s.syftToDomain(*syftSBOM)`. Nothing orders the two writes, so `syftSBOM` may still be nil there and the scan panics.

Passing a cancellable context to Syft, as #583 now does, does not close this: the cataloguers ignore it, which is why the deadline wrapper exists in the first place.

This keeps the work function's result local and publishes it on a buffered channel the caller only receives from once `dl.Run` reports success, so nothing is shared with a goroutine that can outlive the call. The receive cannot block: `dl.Run` returns nil only after the work function returned nil, which it does after publishing.

Behaviour on both paths is unchanged. The success path was never racy, because `deadline.Run` passes the return value over a channel, which orders the write before the read.

## Additional Information

`syft.CreateSBOM` moves behind a `createSBOMFn` package variable so the deadline path is reachable from a test, mirroring what #581 did for the sidecar.

The test needed source resolution to succeed, which this package had no harness for, so it adds `mockRegistryImage`: a local `httptest` registry serving a minimal single-layer image. That keeps the test hermetic, unlike the existing `CreateSBOM` tests which reference a real Docker Hub image, and it is reusable for any future test that needs to get past `syft.GetSource`.

## How to Test

```
go test -race ./adapters/v1/ -run Test_syftAdapter_CreateSBOM_TimeoutDoesNotRaceWithAbandonedSyft
```

The test substitutes a stand-in for Syft that outlives a 1 second deadline and then returns a result exactly as the real cataloguer would, and waits for that late completion before finishing, so the write happens inside the window the race lived in.

Against the previous code it reports:

```
WARNING: DATA RACE
Write at 0x00c000880850 by goroutine 71:
--- FAIL: Test_syftAdapter_CreateSBOM_TimeoutDoesNotRaceWithAbandonedSyft
```

and it passes here.

## Related issues/PRs:

* Resolved #600
* In-process half of #580, fixed for the sidecar in #581

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SBOM generation timeout handling.
  * Timed-out scans now return an incomplete result without blocking or altering results after the operation ends.
  * Preserved successful scan results when completion occurs before the deadline.

### DIFF
--- a/adapters/v1/syft.go
+++ b/adapters/v1/syft.go
@@ -47,6 +47,10 @@ func formatResolvedPlatform(os, arch, variant string) string {
 	return strings.Join(parts, "/")
 }
 
+// createSBOMFn is an indirection over syft.CreateSBOM so the deadline handling in
+// CreateSBOM can be unit-tested without cataloguing a real image.
+var createSBOMFn = syft.CreateSBOM
+
 // SyftAdapter implements SBOMCreator from ports using Syft's API
 type SyftAdapter struct {
 	maxImageSize      int64
@@ -299,6 +303,9 @@ func (s *SyftAdapter) CreateSBOM(ctx context.Context, name, imageID, imageTag st
 	// generate SBOM
 	// use a deadline to prevent the process from hanging for too long
 	var syftSBOM *sbom.SBOM
+	// Buffered so a goroutine abandoned by the deadline can always publish and exit, even
+	// when nobody is left to receive.
+	generated := make(chan *sbom.SBOM, 1)
 	// ensure no parallel pulls
 	s.pullMutex.Lock()
 	defer s.pullMutex.Unlock()
@@ -328,10 +335,16 @@ func (s *SyftAdapter) CreateSBOM(ctx context.Context, name, imageID, imageTag st
 			// ask Syft to also scan the image for embedded SBOMs
 			cfg.WithCatalogers(pkgcataloging.NewCatalogerReference(sbomcataloger.NewCataloger(), []string{pkgcataloging.ImageTag}))
 		}
-		syftSBOM, err = syft.CreateSBOM(ctxWithSize, src, cfg)
-		if err != nil {
-			return fmt.Errorf("failed to generate SBOM: %w", err)
+		// This closure can outlive the call: deadline.Run cannot stop its work function and
+		// documents as much, and Syft's cataloguers do not observe cancellation either, so on
+		// timeout it keeps running after CreateSBOM has already returned Incomplete. It must
+		// therefore not touch any variable the caller reads. Keep the result local and publish
+		// it on the channel, which is only received from once dl.Run reports success.
+		created, createErr := createSBOMFn(ctxWithSize, src, cfg)
+		if createErr != nil {
+			return fmt.Errorf("failed to generate SBOM: %w", createErr)
 		}
+		generated <- created
 		return nil
 	})
 	switch {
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed the fix and verified it locally.

**Correctness**: The channel-based handoff is sound. `deadline.Run`'s internal goroutine only sends on its own `result` channel (and thus lets `dl.Run` return `nil`) *after* the work function has returned — and the work function now writes `generated <- created` before returning `nil`. That gives a solid happens-before edge, so `syftSBOM = <-generated` on the `err == nil` branch can never observe a stale/nil value or race with a late Syft write. On timeout, the abandoned goroutine still safely publishes into the buffered channel (or drops it if already full) without touching anything the caller reads — no goroutine leak, no race.

**Verification performed**:
- `go build ./...` — passes.
- `go vet ./adapters/v1/...`, `gofmt -l` — clean.
- Reverted `syft.go` to `main` while keeping the new test: confirms `createSBOMFn` doesn't exist there, so the new test genuinely exercises this fix's indirection rather than passing coincidentally.
- `go test -race ./adapters/v1/ -run Test_syftAdapter_CreateSBOM_TimeoutDoesNotRaceWithAbandonedSyft -count=5` — 5/5 pass.
- Full `adapters/v1` suite under `-race` — all pass, no regressions.
- CI is already green across build, lint, tests, CodeQL, DCO, and security scans.

No blockers. The mock registry helper is a nice, reusable addition for future source-resolution tests. Approving.

PR #602, opened 2026-08-11
fix(cmd): give metrics shutdown its own context instead of reusing the expired one

Lakshya77089:
## Summary

In `cmd/http/main.go`'s graceful shutdown sequence, `srvCtx` is created with a 5-second timeout and used for `srv.Shutdown`. `controller.Shutdown(c.ShutdownTimeout)` then runs next, and can take up to `ShutdownTimeout` (20s by default). By the time `m.Shutdown(srvCtx)` runs afterward, `srvCtx` has almost always already expired, so the OTel meter provider shutdown gets a `Done()` context and skips its final metrics flush — while an error is still logged even though nothing actually failed.

This fixes it by giving the metrics shutdown its own fresh 5-second context instead of reusing the spent one.

## Test plan

- [x] `go build ./cmd/http/...`
- [x] `go vet ./cmd/http/...`
- Manual review of the shutdown sequence; no existing test covers this shutdown path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved application shutdown handling to ensure metrics services have sufficient time to stop cleanly.
  * Metrics shutdown errors are now logged reliably, even when the main server shutdown context has expired.

### DIFF
--- a/cmd/http/main.go
+++ b/cmd/http/main.go
@@ -235,8 +235,13 @@ func main() {
 	// doesn't silently run past the pod's terminationGracePeriodSeconds
 	controller.Shutdown(c.ShutdownTimeout)
 
-	if err := m.Shutdown(srvCtx); err != nil {
-		logger.L().Ctx(srvCtx).Error("metrics provider shutdown error", helpers.Error(err))
+	// srvCtx's 5s budget is spent by now (controller.Shutdown alone can take up to
+	// ShutdownTimeout, 20s by default), so reusing it here would hand the meter
+	// provider an already-expired context and skip the final metrics flush.
+	metricsCtx, metricsCancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer metricsCancel()
+	if err := m.Shutdown(metricsCtx); err != nil {
+		logger.L().Ctx(metricsCtx).Error("metrics provider shutdown error", helpers.Error(err))
 	}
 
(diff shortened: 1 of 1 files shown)


matthyx:
Reviewed the diff — the fix itself is correct and well-scoped: giving `m.Shutdown` its own fresh `context.WithTimeout(context.Background(), 5*time.Second)` instead of reusing the already-expired `srvCtx` is the right fix, mirrors the existing pattern used for `srvCtx`, and the comment explaining why (`controller.Shutdown` can consume up to `ShutdownTimeout`, 20s by default) is accurate and helpful. `go build` / `go vet` pass locally on `cmd/http`.

**Blocker:** the DCO check is failing — the commit is missing a `Signed-off-by` line. Please rebase and re-sign:

```
git rebase HEAD~1 --signoff
git push --force-with-lease origin fix/metrics-shutdown-expired-context
```

Once DCO passes, this is good to merge from a code perspective.

AnshulPatil2005:
hi @Lakshya77089, are you still on this one? matthyx said it's good to merge from a code perspective, the only thing left is the missing sign-off:

```
git rebase HEAD~1 --signoff
git push --force-with-lease origin fix/metrics-shutdown-expired-context
```

shout if you'd rather someone else pick it up.

matthyx (approved):
DCO now passes and all checks are green. Approving — good to merge.

PR #603, opened 2026-08-11
docs: note the sidecar's SBOM size ceiling and timeout granularity

AnshulPatil2005:
## Overview

Two behaviours are enforced in the sidecar path but not documented, so an operator only finds them by reading the adapter:

- The scanner protocol carries the SBOM size limit as a 32-bit value, so a configured `maxSBOMSize` above `2147483647` is clamped and a warning is logged (#585).
- It carries the scan timeout as whole seconds, so a sub-second `scanTimeout` is rounded up to `1s` and fractional values are rounded up to the next second (#587).

Neither applies to the in-process `SyftAdapter`, which is the part most worth writing down: the same configuration behaves differently depending on which SBOM path is in use, and nothing in the table said so.

This follows the existing `scannerReadinessTimeout` entry, which documents its own edge-case behaviour (`0` meaning "expire immediately", not "wait forever") in the same place.

## Additional Information

Documentation only, no code change. The behaviour described already landed in #585 and #587.

## Related issues/PRs:

* Documents the behaviour introduced in #585 and #587


matthyx (approved):
Verified against the code: `sbomSizeLimitForWire` in `adapters/v1/sidecar.go` clamps to `math.MaxInt32` with a warning log, and `timeoutSecondsForWire` in `pkg/sbomscanner/v1/client.go` rounds up to whole seconds — both only apply to the sidecar path. `SyftAdapter` (adapters/v1/syft.go) uses `maxSBOMSize` as a plain `int` and `scanTimeout` as a `time.Duration` directly with no such clamping/rounding, confirming the in-process adapter isn't affected. Doc-only change, accurate, all checks green. LGTM.

PR #604, opened 2026-08-11
docs: add the three instruments missing from the API metrics table

AnshulPatil2005:
## Overview

The Instruments table in `docs/API.md` enumerates what `/metrics` exposes, but three counters were added without being listed, so an operator reading it has no way to know they exist:

- `kubevuln_scan_fallbacks_total`
- `kubevuln_scan_source_resolution_total`
- `kubevuln_registry_auth_cache_total`

The first two are what the section's own closing paragraph is already describing, where it says an operator watching only `/metrics` can distinguish a registry auth failure from an oversized image. The table was missing the metrics its surrounding prose depends on.

Label vocabularies are taken from the constants in `internal/metrics` rather than summarised, since each is a closed set and an operator writing a PromQL query needs the exact values.

## Additional Information

Documentation only, no code change.

Checked in both directions: every `kubevuln_*` instrument defined in the code is now in the table, and every metric the table names exists in the code. `kubevuln_worker_pool_queue_depth` is registered in `controllers/http.go` rather than `internal/metrics`, which is why it is defined in a different place from the rest but was already documented correctly.

## Related issues/PRs:

* Documents instruments added in #546, #553 and #571


matthyx (approved):
Verified: metric names, label keys, and label values (`component`, `category`, `strategy`, `outcome`, `result`) exactly match the constants and `meter.Int64Counter` registrations in `internal/metrics/metrics.go`. Docs-only change, consistent table formatting, no code impact. LGTM.

PR #597, opened 2026-08-11
fix(backend): capture non-200 response body snippet and fix dead 429 logging (#596)

bhuvan-somisetty:
## Overview

Fixes event-receiver error response body handling and resolves registry scan command image hash initialization (#596).

### Problem
1. `httpPostWithContext` discarded non-200 HTTP response bodies immediately upon receiving a failure status (closing `resp.Body` without reading it), returning only `received status code: <code >`.
2. `postResults` attempted to inspect `resp.StatusCode` after `httpPostFunc` returned an error (which returns `resp = nil`). As a result, early error returns prevented lines 142–151 from executing, rendering the `429 Too Many Requests` vendor guidance log dead code.
3. `registryScanCommandToScanCommand` left `command.ImageHash` uninitialized (`""`) and passed unnormalized `c.ImageTag` to `names.ImageInfoToSlug`, leading to missing hash attributes in scan failure/telemetry reports.

### Solution
1. **Response Snippet Capture (`adapters/v1/backend.go`)**: Updated `httpPostWithContext` to read up to 512 bytes of response body on non-200 responses and append it to the returned `error` message (e.g. `received status code: 429, body: <snippet>`).
2. **Clean 429 Log Handling (`adapters/v1/backend_utils.go`)**: Refactored `postResults` to inspect error strings for `429` / `Too Many Requests` and log rate-limit guidance without dereferencing a nil `*http.Response` pointer.
3. **Registry Command Normalization (`controllers/http.go`)**: Updated `registryScanCommandToScanCommand` to populate `ImageHash` using `v1.NormalizeImageID("", c.ImageTag)` and pass `tools.NormalizeReference(c.ImageTag)` to `names.ImageInfoToSlug`.
4. **Unit Tests (`adapters/v1/backend_test.go`, `controllers/http_test.go`)**: Added unit tests verifying body snippet inclusion on 429 responses, rate-limit log triggering, and registry command hash/slug normalization.

## How to Test

Run the unit test suite covering `httpPostWithContext`, `postResults`, and `registryScanCommandToScanCommand`:
```bash
go test -v -count=1 ./adapters/v1 -run "TestHttpPostWithContext|TestBackendAdapter_PostResults"
go test -v -count=1 ./controllers -run "Test_registryScanCommandToScanCommand"
```

## Related issues/PRs:

* Fixes #596

## Checklist before requesting a review

- [x] My code follows the style guidelines of this project
- [x] I have commented on my code, particularly in hard-to-understand areas
[...]

PR #607, opened 2026-08-11
fix(controllers): key registry scans on the normalized image reference

### LINKED ISSUE #606: Bug: registry scans key storage on the raw image reference, so equivalent forms never hit the cache
# Description

A registry scan of the same image under a different reference form stores it under a different key, so the cache never hits and the SBOM is regenerated from scratch.

`registryScanCommandToScanCommand` (`controllers/http.go`) derives `ImageSlug` from the raw `ImageTag` while setting `ImageTagNormalized` from the normalized one:

```go
command := domain.ScanCommand{
	ImageTag:           c.ImageTag,
	ImageTagNormalized: tools.NormalizeReference(c.ImageTag),
	...
}
if slug, err := names.ImageInfoToSlug(c.ImageTag, "nohash"); err == nil {
	command.ImageSlug = slug
}
```

`ImageSlug` is the storage key for both the SBOM (`getSBOM`, `CreateSBOM`) and the CVE manifest (`GetCVE`), so equivalent references to one image get separate keys:

| `ImageTag` in the request | resulting `ImageSlug` |
|---|---|
| `nginx` | `nginx-nohash` |
| `nginx:latest` | `nginx-latest-nohash` |
| `docker.io/library/nginx:latest` | `docker.io-library-nginx-latest-nohash` |

All three are the same image, and all three produce `docker.io-library-nginx-latest-nohash` once normalized.

The in-cluster path does not have this problem: `websocketScanCommandToScanCommand` derives the slug from `imageTagNormalized`.

# Environment

- OS: N/A
- Version: `main` (`7b33df6`)

# Steps To Reproduce

1. Trigger a registry scan with `ImageTag: "docker.io/library/nginx:latest"` and let it complete.
2. Trigger a second registry scan for the same image with `ImageTag: "nginx:latest"`.
3. The second scan does not find the stored SBOM or CVE manifest and regenerates both.

# Expected behavior

Equivalent references to one image share a storage key, so the second scan is a cache hit.

# Actual Behavior

Each reference form gets its own key. The second scan is a full cache miss and repeats the image pull and cataloguing, which is the 30s+ cost caching exists to avoid, and the cluster accumulates several SBOM and CVE objects describing one image.

# Additional context

This also puts storage out of step with the other caches keyed off the same command: `rateLimitCacheKey` (`core/services/scan.go`) and the exceptions cache key (`adapters/v1/backend.go`) both already use `ImageTagNormalized`. So the rate-limit backoff and the exception set treat the two requests as the same image while storage does not.

The existing `Test_registryScanCommandToScanCommand` already exercises `nginx:latest` and `docker.io/library/nginx:latest` side by side, but only asserts `ImageTagNormalized`, never `ImageSlug`, which is why this was not caught.

Fixing this orphans objects stored under a raw-derived slug; they are rescanned once under the normalized key.

AnshulPatil2005:
## Overview

`registryScanCommandToScanCommand` derived `ImageSlug` from the raw `ImageTag` while setting `ImageTagNormalized` from the normalized one, so the two disagreed about which image the command referred to.

`ImageSlug` is the storage key for both the SBOM and the CVE manifest, so every equivalent reference to one image got its own key:

| `ImageTag` in the request | `ImageSlug` |
|---|---|
| `nginx` | `nginx-nohash` |
| `nginx:latest` | `nginx-latest-nohash` |
| `docker.io/library/nginx:latest` | `docker.io-library-nginx-latest-nohash` |

Scanning the same image under a different form therefore missed the cache entirely and regenerated the SBOM from scratch, which is the 30s+ pull and catalogue that caching exists to avoid, and left several stored objects describing one image.

It also put storage out of step with the other caches keyed off the same command: `rateLimitCacheKey` and the exceptions cache key both use `ImageTagNormalized` already, so the backoff and the exception set treated two such requests as the same image while storage did not.

This derives the slug from the normalized reference, as `websocketScanCommandToScanCommand` already does for the in-cluster path.

## Additional Information

Objects stored under a raw-derived slug are orphaned by this change and will be rescanned once under the normalized key. That is a one-time cost and self-healing, and it only affects registry scans that used a non-canonical reference, which are the ones already paying for a full rescan on every run.

The in-cluster path is untouched. Note the two paths still produce different slugs for the same image by design, since that one embeds the image digest while registry scans have none and use `"nohash"`.

## How to Test

```
go test ./controllers/ -run Test_registryScanCommandToScanCommand
```

`SlugIsStableAcrossEquivalentReferences` covers the five forms that resolve to the same image (`nginx`, `nginx:latest`, `library/nginx:latest`, `docker.io/library/nginx:latest`, `index.docker.io/library/nginx:latest`) and asserts they share a key. It fails against the previous code.

`SlugDistinguishesImages` is the guard on the other side, so a fix that collapsed everything to one key would not pass: different tags, different repositories and different registries must still differ.

The pre-existing `Test_registryScanCommandToScanCommand` is unchanged and still passes. It already exercised two of these equivalent forms but only asserted `ImageTagNormalized`, never `ImageSlug`, which is why this went unnoticed.

## Related issues/PRs:

* Resolved #606

### DIFF
--- a/controllers/http.go
+++ b/controllers/http.go
@@ -365,16 +365,24 @@ func (h HTTPController) ScanRegistry(c *gin.Context) {
 }
 
 func registryScanCommandToScanCommand(c wssc.RegistryScanCommand) domain.ScanCommand {
+	imageTagNormalized := tools.NormalizeReference(c.ImageTag)
 	command := domain.ScanCommand{
 		CredentialsList:    c.Credentialslist,
 		ImageTag:           c.ImageTag,
-		ImageTagNormalized: tools.NormalizeReference(c.ImageTag),
+		ImageTagNormalized: imageTagNormalized,
 		JobID:              c.JobID,
 		ParentJobID:        c.ParentJobID,
 		Args:               c.Args,
 		Session:            sessionChainToSession(c.Session),
 	}
-	if slug, err := names.ImageInfoToSlug(c.ImageTag, "nohash"); err == nil {
+	// Derive the slug from the normalized reference, as the in-cluster path does. ImageSlug
+	// is the storage key for the SBOM and the CVE manifest, so deriving it from the raw tag
+	// gave the same image a different key per reference form ("nginx:latest" and
+	// "docker.io/library/nginx:latest" are the same image), and every equivalent form missed
+	// the cache and regenerated the SBOM from scratch. The rate-limit and exception caches
+	// already key on the normalized reference, so this also stops the caches disagreeing
+	// about which image they are talking about.
+	if slug, err := names.ImageInfoToSlug(imageTagNormalized, "nohash"); err == nil {
 		command.ImageSlug = slug
 	}
 	return command
--- a/controllers/http_test.go
+++ b/controllers/http_test.go
@@ -756,3 +756,45 @@ func TestHTTPController_MetricsEndpoint_RecordsScanFailureReason(t *testing.T) {
 		})
 	}
 }
+
+// ImageSlug is the storage key for the SBOM and the CVE manifest, so equivalent references to
+// the same image must produce the same slug. Deriving it from the raw tag gave each reference
+// form its own key, so every form missed the cache and regenerated the SBOM from scratch, and
+// the rate-limit and exception caches (which key on the normalized reference) disagreed with
+// storage about which image was in play.
+func Test_registryScanCommandToScanCommand_SlugIsStableAcrossEquivalentReferences(t *testing.T) {
+	equivalent := []string{
+		"nginx",
+		"nginx:latest",
+		"library/nginx:latest",
+		"docker.io/library/nginx:latest",
+		"index.docker.io/library/nginx:latest",
+	}
+
+	var slugs []string
+	for _, ref := range equivalent {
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Reviewed. The fix correctly derives `ImageSlug` from `imageTagNormalized` instead of the raw `ImageTag`, matching the pattern already used in `websocketScanCommandToScanCommand` (in-cluster path), so both paths now agree with `ImageTagNormalized`-keyed caches (rate-limit, exceptions).

Verified:
- `go build ./...` passes.
- `go test ./controllers/... ./internal/tools/...` passes, including the two new tests (`Test_registryScanCommandToScanCommand_SlugIsStableAcrossEquivalentReferences`, `Test_registryScanCommandToScanCommand_SlugDistinguishesImages`), which correctly assert cache-key stability across equivalent reference forms and distinguish different images/registries.
- No regressions to existing `Test_registryScanCommandToScanCommand` or other controller tests.

No blockers found. Approving.

matthyx:
needs rebase

AnshulPatil2005:
@matthyx closing this rather than rebasing, @bhuvan-somisetty's #597 merged in between and already made the same change.

PR #609, opened 2026-08-11
fix(services): route ScanRegistry's SBOM lookup through getSBOM

### LINKED ISSUE #608: Bug: raising maxImageSize never takes effect for registry scans, which keep serving the stale TooLarge SBOM
# Description

Raising `maxImageSize` lets a previously-too-large image be scanned on every path except the registry one, where it stays rejected indefinitely.

`getSBOM` (`core/services/scan.go`) exists to discard a stored SBOM whose verdict no longer matches the configuration: an outdated one, or a `TooLarge` one recorded under a `maxImageSize`, `maxSBOMSize` or scanner memory limit that has since changed. It deletes the stored object and returns empty so the caller regenerates.

`GenerateSBOM`, `ScanCP` and `ScanCVE` all call it:

```go
sbom, err = s.getSBOM(ctx, workload.ImageSlug, s.sbomCreator.Version())   // lines 175, 297, 541
```

`ScanRegistry` reads the repository directly and skips all of it:

```go
sbom, err = s.sbomRepository.GetSBOM(ctx, workload.ImageSlug, s.sbomCreator.Version())   // line 741
```

The retrieved SBOM then goes straight into the status check a few lines below, which turns a cached `TooLarge` into a scan failure without ever reconsidering it:

```go
if sbom.Status == helpersv1.Incomplete || sbom.Status == helpersv1.TooLarge {
	reason := classifySBOMStatusWithAnnotation(sbom.Status, sbom.Annotations)
	_ = s.platform.ReportScanFailure(ctx, scanfailure.ScanFailureSBOMGeneration, reason, nil)
	return &domain.ScanError{Reason: reason, Err: domain.ErrIncompleteSBOM}
}
```

# Environment

- OS: N/A
- Version: `main` (`7b33df6`)

# Steps To Reproduce

1. Registry-scan an image larger than `maxImageSize`. It is stored with status `TooLarge` and a `kubescape.io/max-image-size` annotation recording the limit in force.
2. Raise `maxImageSize` above the image size and restart.
3. Registry-scan the same image again.
4. It fails with `image-too-large` again. The stored SBOM is served as-is and never regenerated, so the new limit has no effect.

Driving `ScanRegistry` against a stored `TooLarge` SBOM whose annotation records a different limit shows it returning `incomplete SBOM, skipping CVE scan` with the stored object left untouched.

The same setup on `ScanCVE` or `ScanCP` deletes the stale SBOM and rescans.

# Expected behavior

The stored verdict is re-evaluated against the current limits on every scan path, so raising a limit takes effect everywhere.

# Actual behavior

Only the registry path keeps the stale verdict. The image stays rejected until the stored SBOM is deleted by hand, and nothing logs or reports why the new limit was ignored.

# Additional context

The divergence is invisible at the call site: both spellings return `(domain.SBOM, error)` and the wrapper is named closely enough to the repository method that reading `ScanRegistry` alone gives no hint anything is missing.

This is the same shape as #484, #501 and #557, where a step present in two of the three scan flows was absent from the third.

AnshulPatil2005:
## Overview

`getSBOM` discards a stored SBOM whose verdict no longer matches the configuration: an outdated one, or a `TooLarge` one recorded under a `maxImageSize`, `maxSBOMSize` or scanner memory limit that has since changed. `GenerateSBOM`, `ScanCP` and `ScanCVE` all go through it. `ScanRegistry` read the repository directly and skipped it.

The retrieved SBOM feeds the status check a few lines below, which turns a cached `TooLarge` straight into a scan failure, so a registry scan kept serving that verdict indefinitely. An operator raising `maxImageSize` to admit a large image saw it take effect on every path except the registry one, where the image stayed rejected until the stored SBOM was deleted by hand, with nothing logged to explain why the new limit was ignored.

This calls `getSBOM` there too, so the limits are re-evaluated on every path.

## Additional Information

The divergence is invisible at the call site: both spellings return `(domain.SBOM, error)`, and the wrapper is named closely enough to the repository method that reading `ScanRegistry` alone gives no hint anything is missing. It is the same shape as #484, #501 and #557, where a step present in two of the three scan flows was absent from the third.

No behaviour change for a registry scan whose stored SBOM is current: `getSBOM` returns it untouched. The only scans affected are ones already stuck, which now regenerate once.

## How to Test

```
go test ./core/services/ -run TestScanService_ScanRegistry_StaleTooLargeSBOMIsRescanned
```

The test stores a `TooLarge` SBOM annotated with an image-size limit that differs from the current one, asserts up front that the two differ so it cannot pass vacuously, then runs `ScanRegistry` and requires the stored SBOM to have been replaced rather than served again. It fails against the previous code, where `ScanRegistry` returns `incomplete SBOM, skipping CVE scan` and leaves the stored object as it was.

## Related issues/PRs:

* Resolved #608

### DIFF
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -738,7 +738,12 @@ func (s *ScanService) ScanRegistry(ctx context.Context) error {
 	sbom := domain.SBOM{}
 	if cve.Content == nil {
 		if s.storage {
-			sbom, err = s.sbomRepository.GetSBOM(ctx, workload.ImageSlug, s.sbomCreator.Version())
+			// Go through getSBOM rather than the repository directly, as GenerateSBOM,
+			// ScanCP and ScanCVE all do. It is what discards an SBOM that was rejected
+			// under limits which have since changed, so without it a registry scan keeps
+			// serving a cached TooLarge verdict even after the operator raised
+			// maxImageSize, maxSBOMSize or the scanner memory limit to admit that image.
+			sbom, err = s.getSBOM(ctx, workload.ImageSlug, s.sbomCreator.Version())
 			if err != nil {
 				logger.L().Ctx(ctx).Warning("getting SBOM", helpers.Error(err),
 					helpers.String("imageSlug", workload.ImageSlug))
--- a/core/services/scan_test.go
+++ b/core/services/scan_test.go
@@ -2431,3 +2431,45 @@ func TestScanService_ScanCVE_CacheHit_DegradedAdditiveSkipsVEX(t *testing.T) {
 
 	assert.Empty(t, repo.vexCalls, "but no VEX document is published from a partial exception set")
 }
+
+// getSBOM discards an SBOM that was rejected under limits which have since changed, so a
+// rescan can admit an image the operator has just made room for. ScanRegistry called the
+// repository directly and skipped that, so it kept serving a cached TooLarge verdict forever:
+// raising maxImageSize had no effect on the registry path while it worked on every other one.
+func TestScanService_ScanRegistry_StaleTooLargeSBOMIsRescanned(t *testing.T) {
+	sbomAdapter := adapters.NewMockSBOMAdapter(false, false, false)
+	repo := repositories.NewMemoryStorage(false, false)
+	s := NewScanService(sbomAdapter, repo, adapters.NewMockCVEAdapter(), repo,
+		&recordingPlatform{}, v1.NewContainerProfileAdapter(repositories.NewMemoryStorage(false, false)),
+		true, false, true, false, false)
+	ctx := context.TODO()
+	s.Ready(ctx)
+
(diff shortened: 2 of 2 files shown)


matthyx (approved):
LGTM. Verified locally: builds cleanly, `go vet` clean, and all `core/services` tests pass (39/39), including the new `TestScanService_ScanRegistry_StaleTooLargeSBOMIsRescanned` regression test which fails against pre-PR code and passes with the fix.

The fix routes `ScanRegistry`'s SBOM lookup through `getSBOM` (matching `GenerateSBOM`/`ScanCP`/`ScanCVE`), so a stale `TooLarge` verdict is correctly discarded and rescanned when limits (`maxImageSize`/`maxSBOMSize`/scanner memory) change, instead of being served indefinitely. No behavior change for non-stale/current SBOMs. Approving.

PR #610, opened 2026-08-11
feat: deduplicate concurrent SBOM generation with singleflight in ScanService (#605)

### LINKED ISSUE #605: feat: deduplicate concurrent SBOM generation with singleflight in ScanService
**Description:**
When `kubevuln` processes multiple concurrent scan jobs for workloads sharing the same container image (e.g. multiple replicas or pods using `nginx:latest`), each worker independently checks storage, finds no existing SBOM, and simultaneously invokes `s.sbomCreator.CreateSBOM()`.

This causes redundant image pulls, duplicate CPU/memory spikes, and frequently triggers `429 Too Many Requests` rate limits from container registries.

### Proposed Solution
1. Add a `golang.org/x/sync/singleflight` Group (`sfGroup`) to `ScanService` (`core/services/scan.go`).
2. Add a `getOrCreateSBOM()` helper method in `ScanService` wrapped with `sfGroup.DoChan()` keyed by `imageDigest + platform`.
3. Ensure the singleflight worker stores the generated SBOM in storage before completing, allowing all waiting goroutines to share the saved result.
4. Use `DoChan` so waiter goroutines can safely enforce their own context deadlines without canceling the active worker goroutine.
5. Add a metric to record singleflight deduplication hits.
6. Add unit tests in `core/services/scan_test.go` verifying concurrent `GenerateSBOM` / `ScanCVE` calls invoke `CreateSBOM` only once.

NaitikRishu:
## Overview
Currently, when multiple pods sharing the same container image (or duplicate scan requests) are scanned simultaneously, `ScanService` triggers multiple redundant `CreateSBOM` calls. This causes unnecessary CPU/memory spikes and exhausts container registry rate limits.

This PR adds a `singleflight.Group` inside `ScanService` (`core/services/scan.go`) to deduplicate concurrent `CreateSBOM` calls for identical images (`imageDigest + platform`) across `GenerateSBOM`, `ScanCVE`, `ScanCP`, and `ScanRegistry` flows.

Key changes:
- Wrapped SBOM creation in `getOrCreateSBOM()` using `sfGroup.DoChan()`.
- Guaranteed singleflight workers store the generated SBOM in storage before completing so waiting goroutines receive a stored result.
- Used `DoChan` so waiter goroutines can safely enforce their own context deadlines without canceling active singleflight workers.
- Added `kubevuln_singleflight_hits_total` Prometheus counter in `internal/metrics`.
- Added unit test `TestScanService_SingleflightSBOMDeduplication` verifying concurrent calls invoke `CreateSBOM` only once.

## Additional Information
Deduplication key uses `rateLimitCacheKey(workload) + opts.Platform` to ensure credentials and target platform are respected without cross-tenant collisions.

## How to Test
Run unit tests locally:
```bash
go test -v -run TestScanService_SingleflightSBOMDeduplication ./core/services
```

## Related issues/PRs:
Resolved #605

- [x] New and existing unit tests pass locally with my changess made in this pull request. e.g. current behavior/future behavior -->

 

## Summary by CodeRabbit

* **Performance**
  * Improved concurrent SBOM generation by preventing duplicate work and storage operations.
  * Added centralized handling for SBOM rate limits and unsupported schemas.

* **Monitoring**
  * Added metrics to track deduplicated scan requests and singleflight activity.

### DIFF
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -35,6 +35,7 @@ import (
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/metric"
 	"go.opentelemetry.io/otel/trace"
+	"golang.org/x/sync/singleflight"
 	"k8s.io/client-go/tools/record"
 )
 
@@ -60,6 +61,7 @@ type ScanService struct {
 	tooManyRequests   *cache.Cache
 	metrics           *metrics.Metrics
 	eventRecorder     record.EventRecorder
+	sfGroup           singleflight.Group
 }
 
 // SetMetrics attaches an optional Metrics instance to the service. It is not
@@ -168,35 +170,9 @@ func (s *ScanService) GenerateSBOM(ctx context.Context) error {
 		return domain.ErrCastingWorkload
 	}
 
-	// check if SBOM is already available
-	sbom := domain.SBOM{}
-	var err error
-	if s.storage {
-		sbom, err = s.getSBOM(ctx, workload.ImageSlug, s.sbomCreator.Version())
-		if err != nil {
-			logger.L().Ctx(ctx).Warning("getting SBOM", helpers.Error(err),
-				helpers.String("imageSlug", workload.ImageSlug))
-		}
-	}
-
-	// if SBOM is not available, create it
-	if sbom.Content == nil {
-		if _, ok := s.tooManyRequests.Get(rateLimitCacheKey(workload)); ok {
-			logger.L().Ctx(ctx).Warning("skipping SBOM creation, image pull previously rate limited",
-				helpers.String("imageSlug", workload.ImageSlug))
-			_ = s.platform.ReportScanFailure(ctx, scanfailure.ScanFailureSBOMGeneration,
-				scanfailure.ReasonSBOMGenerationFailed, domain.ErrTooManyRequests)
-			return &domain.ScanError{Reason: scanfailure.ReasonSBOMGenerationFailed, Err: domain.ErrTooManyRequests}
-		}
-		// create SBOM
-		sbom, err = s.sbomCreator.CreateSBOM(ctx, workload.ImageSlug, workload.ImageHash, workload.ImageTagNormalized, optionsFromWorkload(ctx, workload))
-		s.checkCreateSBOM(err, rateLimitCacheKey(workload))
-		if err != nil {
-			reason := classifySBOMError(err)
-			_ = s.platform.ReportScanFailure(ctx, scanfailure.ScanFailureSBOMGeneration,
-				reason, err)
-			return &domain.ScanError{Reason: reason, Err: err}
-		}
+	sbom, created, err := s.getOrCreateSBOM(ctx, workload)
+	if err != nil {
+		return err
 	}
 
 	// do not treat a timed out or too-large SBOM as a successful scan — mirrors the same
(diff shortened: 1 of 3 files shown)


NaitikRishu:
Hey @matthyx

> This PR adds singleflight deduplication to `ScanService` (#605).
> 
> When multiple pods share the same image, it collapses concurrent scans into 1 single build instead of pulling the image 50 times. This reduces duplicate registry pulls by up to 98% and stops cluster scans from hitting Docker Hub rate limits!

matthyx (changes requested):
Reviewed the singleflight SBOM deduplication change. Design and the `getOrCreateSBOM` refactor look solid overall (rate-limit check, error classification, and stub-storage behavior are preserved for the leader path), but I found a confirmed data race in the code this PR adds.

## Blocker: data race in `getOrCreateSBOM`

`go test -race -run TestScanService_SingleflightSBOMDeduplication ./core/services/` fails (test passes without `-race`, and it's not run with `-race` in CI, which is why this slipped through):

```
WARNING: DATA RACE
Write at 0x00c000b2bd40 by goroutine 54:
  repositories.(*MemoryStore).StoreSBOM()
  ScanService.getOrCreateSBOM.func1()  (scan.go:1351, inside the singleflight worker)

Previous read at 0x00c000b2bd40 by goroutine 52:
  repositories.(*MemoryStore).GetSBOM()
  ScanService.getSBOM()
  ScanService.getOrCreateSBOM()  (scan.go:1299, the "fast path" storage check)
```

Root cause: the "fast path" storage read in `getOrCreateSBOM` (`scan.go:1298-1301`) happens *outside* `sfGroup.DoChan`, so it can run concurrently with another goroutine's singleflight worker that is inside `StoreSBOM` (`scan.go:1349-1353`) for the same key. This is exactly the concurrent-scan scenario the PR is meant to fix, so it will be hit in production traffic, not just the test.

With the real storage backend this may or may not be a runtime crash (depends on that backend's own concurrency safety), but it's still a genuine logic race and it means the new regression test the PR adds does not pass under `-race`, which the project's own CI doesn't catch since `-race` isn't enabled. Worth fixing regardless — e.g. move the storage re-check fully inside the singleflight callback for the "unknown" case, or guard the fast-path read/write path with its own lock/atomic check, then confirm `go test -race ./core/services/...` is clean.

Everything else (metrics counter, context detachment via `context.WithoutCancel`, error handling parity across `GenerateSBOM`/`ScanCVE`/`ScanCP`/`ScanRegistry`) looks correct and I didn't find other issues. Once the race above is addressed this should be good to merge.

NaitikRishu:
Thanks @matthyx 

I've removed the fast-path storage check from outside sfGroup.DoChan so all storage reads and writes for a given image key now execute safely inside singleflight.

Verified locally with go test -race ./core/services/... ./internal/metrics/... and confirmed all tests pass cleanly with zero data races. Ready for re-review!

matthyx (approved):
Re-reviewed at commit d53029f. The data race I flagged earlier is fixed: the storage fast-path read that used to run outside `sfGroup.DoChan` has been removed, and all storage checks now happen inside the singleflight callback.

Verified:
- `go build ./...` — passes.
- `go test -race -run TestScanService_SingleflightSBOMDeduplication -count=3 ./core/services/` — passes consistently (previously failed reliably).
- `go test -race ./core/services/...` (full package) — passes.
- `go vet ./...` — clean.
- CodeRabbit's double-store nitpick is also resolved (`GenerateSBOM` now gates the extra `StoreSBOM` call on the `created` flag returned by `getOrCreateSBOM`).
- CI checks are green (FOSSA jobs still pending/informational only).

Approving.

PR #612, opened 2026-08-11
fix: decode the Auth field when converting registry credentials

### LINKED ISSUE #611: Bug: registryCredentialsFromCredentialsList drops the Auth field, silently discarding credentials supplied without separate username/password
# Description
`registryCredentialsFromCredentialsList` in `core/services/scan.go` (around line 1083) converts each incoming `registry.AuthConfig` entry into a `domain.RegistryCredentials` by reading only `ServerAddress`, `RegistryToken`, `Username`, and `Password`. It never reads `Auth`, the base64 `"username:password"` field that Docker's own config format treats as canonical.

`registry.AuthConfig` (`github.com/docker/docker/api/types/registry`) is exactly the shape of an entry from a `.dockerconfigjson` pull secret, and Kubernetes' own documented way to reuse an existing Docker login ("Registry secret existing credentials", https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#registry-secret-existing-credentials) produces a config where each registry entry has **only** `auth` set — `docker login` does not populate `username`/`password` alongside it. Any `imagePullSecret` built this way (a very common pattern: `kubectl create secret generic regcred --from-file=.dockerconfigjson=~/.docker/config.json`) reaches kubevuln with `Username` and `Password` both empty and `Auth` holding the real credential.

When that happens, `registryCredentialsFromCredentialsList` produces a `domain.RegistryCredentials{}` with no `Username`, `Password`, or `Token` at all (only `Authority`, if `ServerAddress` was set) — the credential is silently dropped. The scan then pulls the image with effectively anonymous access, gets a 401 from the private registry, and fails or falls back to the anonymous/unauthorized path, even though a valid credential was supplied end-to-end.

This affects both SBOM generation paths: `ScanService.optionsFromWorkload` (`core/services/scan.go:1031`) is the single place `workload.CredentialsList` is converted, and its output feeds both the in-process `SyftAdapter` and the `SidecarSBOMAdapter` (via `pkg/sbomscanner/v1/client.go`'s `CreateSBOM`), so neither path can pull an image whose credential arrived only as `Auth`.

The existing unit test for this function (`Test_registryCredentialsFromCredentialsList` in `core/services/scan_test.go`) only covers entries where `Auth` and `Username`/`Password` are both populated together, which is why this gap wasn't caught — every fixture happens to carry the redundant fields.

# Environment
OS: `Any`
Version: `main`

# Steps To Reproduce
1. Create a pull secret the way Kubernetes' docs recommend for reusing an existing Docker login:
   ```
   docker login myregistry.example.com
   kubectl create secret generic regcred \
     --from-file=.dockerconfigjson=$HOME/.docker/config.json \
     --type=kubernetes.io/dockerconfigjson
   ```
   The resulting secret's `auths."myregistry.example.com"` entry has only `auth` (base64 `user:pass`), no `username`/`password` keys.
2. Trigger a scan (`ScanCVE`/`ScanCP`/registry scan) for a private image on `myregistry.example.com`, with that credential forwarded in the request's `credentialsList`.
3. In `core/services/scan.go`, step through `registryCredentialsFromCredentialsList`: for that entry, `cred.Username == "" && cred.Password == ""`, so the `if cred.Username != "" && cred.Password != ""` branch is skipped, and `cred.Auth` is never inspected.
4. The resulting `domain.RegistryCredentials` has no `Username`/`Password`/`Token`.

# Expected behavior
When `Username`/`Password` are empty but `Auth` is set, kubevuln should base64-decode `Auth` (standard `username:password` form) and use the resulting username/password, so a credential supplied only via the canonical Docker `auth` field still authenticates the pull.

# Actual Behavior
The credential is silently discarded. The pull proceeds unauthenticated, fails against a private registry, and there is no log or metric pointing at "credential had `Auth` but no `Username`/`Password`" — it looks identical to "no credential was provided at all".

# Additional context
`IdentityToken` (OAuth2/refresh-token auth, e.g. some registries' token exchange flow) is dropped the same way and has no equivalent field on `domain.RegistryCredentials` at all, but fixing that would need a design decision (map it into `Token`, or add a new field); the `Auth` case is the straightforward, high-impact fix since it maps directly onto the existing `Username`/`Password` fields and the exact same problem exists in `SidecarSBOMAdapter`'s credential path via the shared `optionsFromWorkload` → `registryCredentialsFromCredentialsList` conversion.

bhuvan-somisetty:
## Overview

`registryCredentialsFromCredentialsList` (`core/services/scan.go`) only read `Username`/`Password`/`RegistryToken` off each incoming `registry.AuthConfig` entry, never `Auth` — the base64 `"username:password"` field Docker treats as the canonical credential.

Kubernetes' documented way to reuse an existing Docker login (`kubectl create secret generic regcred --from-file=.dockerconfigjson=~/.docker/config.json`, see https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#registry-secret-existing-credentials) produces a pull secret where each registry entry has only `auth` set, no separate `username`/`password`. Any workload using that pattern reached kubevuln with `Username`/`Password` empty and the real credential sitting unused in `Auth`, so the resulting `domain.RegistryCredentials` came out completely empty and the pull went out anonymous instead of authenticated.

`optionsFromWorkload` is the single place `workload.CredentialsList` gets converted, and its output feeds both SBOM paths (in-process `SyftAdapter` and the `SidecarSBOMAdapter`), so this affected both.

Fix: when `Username`/`Password` are both empty but `Auth` is set, base64-decode it and split on the first `:` to recover username/password. Malformed `Auth` (bad base64, no `:` separator) falls back to empty credentials, same as before.

## How to Test

- `go test ./core/services/... -run 'Test_registryCredentialsFromCredentialsList|Test_credentialsFromAuth'`
- `go build ./...`
- `go test ./...`
- `golangci-lint run --new-from-rev=<merge-base with main>`

## Related issues/PRs:

* Resolved #611

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Docker registry credential handling when credentials are provided only as a Base64-encoded value.
  - Preserved passwords containing colons during credential decoding.
  - Safely ignored empty, invalid, or malformed credential values.

- **Tests**
  - Added coverage for valid credentials, passwords with colons, and invalid input scenarios.

### DIFF
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -3,6 +3,7 @@ package services
 import (
 	"context"
 	"crypto/sha256"
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"maps"
@@ -1090,14 +1091,36 @@ func registryCredentialsFromCredentialsList(credentials []registry.AuthConfig) [
 		if cred.RegistryToken != "" {
 			rc.Token = cred.RegistryToken
 		}
-		if cred.Username != "" && cred.Password != "" {
-			rc.Username = cred.Username
-			rc.Password = cred.Password
+		username, password := cred.Username, cred.Password
+		if username == "" && password == "" && cred.Auth != "" {
+			username, password = credentialsFromAuth(cred.Auth)
+		}
+		if username != "" && password != "" {
+			rc.Username = username
+			rc.Password = password
 		}
 
 		registryCredentials[i] = rc
 	}
 	return registryCredentials
+}
+
+// credentialsFromAuth decodes the standard Docker "auth" field: base64 of "username:password".
+// It's the canonical field in a .dockerconfigjson pull secret; username/password are only
+// supplementary and are often absent. Kubernetes' documented way to reuse an existing docker
+// login (kubectl create secret generic --from-file=.dockerconfigjson=~/.docker/config.json)
+// produces entries where only auth is set, so falling back to it here is required for those
+// credentials to reach the registry pull at all, rather than being silently dropped.
+func credentialsFromAuth(auth string) (username, password string) {
+	decoded, err := base64.StdEncoding.DecodeString(auth)
+	if err != nil {
+		return "", ""
+	}
+	username, password, found := strings.Cut(string(decoded), ":")
+	if !found {
+		return "", ""
+	}
+	return username, password
 }
 
 // parseAuthorityFromServerAddress extracts the authority host:port from a server address.
--- a/core/services/scan_test.go
+++ b/core/services/scan_test.go
@@ -2,6 +2,7 @@ package services
 
 import (
 	"context"
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -1206,9 +1207,16 @@ func Test_registryCredentialsFromCredentialsList(t *testing.T) {
 			Username:      "armosec+testrobot2",
 			Password:      "dummyPassword111",
 		},
+		{
+			// Mirrors a .dockerconfigjson entry produced by reusing an existing
+			// `docker login` (kubectl create secret generic --from-file=.dockerconfigjson=...):
+			// only Auth is set, Username/Password are empty. See #611.
+			ServerAddress: "registry.example.com",
+			Auth:          base64.StdEncoding.EncodeToString([]byte("registryuser:registrypass")),
+		},
 	}
 	registryCredentials := registryCredentialsFromCredentialsList(creds)
-	assert.Equal(t, 3, len(registryCredentials))
+	assert.Equal(t, 4, len(registryCredentials))
 	assert.Equal(t, "quay.io", registryCredentials[0].Authority)
 	assert.Equal(t, "armosec+testrobot1", registryCredentials[0].Username)
 	assert.Equal(t, "dummyPassword", registryCredentials[0].Password)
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Verified this against #611: `registryCredentialsFromCredentialsList` previously required `Username`+`Password` and never looked at `Auth`, so a pull secret created via `kubectl create secret generic ... --from-file=.dockerconfigjson=...` (which only populates `auth`) silently produced empty `domain.RegistryCredentials` and the scan went out unauthenticated. The fix decodes `Auth` (base64 `user:pass`, split on first `:` via `strings.Cut` so passwords containing `:` survive) only when `Username`/`Password` are both empty, and falls back to empty credentials on bad base64 or a missing separator — same anonymous behavior as before, just no longer silently dropping the canonical field.

Checked:
- `go build ./...` — clean
- `go test ./core/services/... -run 'Test_registryCredentialsFromCredentialsList|Test_credentialsFromAuth'` — pass, including the colon-in-password and malformed-input cases
- `go test ./...` — passes outside of two pre-existing `adapters/v1` syft tests (`Test_syftAdapter_CreateSBOM/public_image_with_invalid_credentials_falls_back_to_unauthenticated`, `Test_syftAdapter_transformations`) that are flaky independent of this change — confirmed by reproducing the same intermittent failures on `main` and getting a clean pass re-running the PR branch in isolation
- `golangci-lint run --new-from-rev=<merge-base>` — no new issues in the touched lines
- Imports (`encoding/base64`, `strings`) already present/added correctly; `registry.AuthConfig.Auth` is a real field on `github.com/docker/docker/api/types/registry`

No blockers. Nice, minimal fix with good edge-case test coverage.

PR #614, opened 2026-08-11
feat(adapters): add exponential backoff retry for transient 429 rate limit responses (#613)

### LINKED ISSUE #613: Add exponential backoff retries for transient HTTP 429 rate limit errors
## Description
When scanning container images under heavy load or against rate-limited registries (e.g., Docker Hub), container image resolution and layer fetches may occasionally fail with transient HTTP `429 Too Many Requests` errors. 

Currently, a single transient `429` error immediately causes `kubevuln` to cache a rate-limit failure in `tooManyRequests`, blocking subsequent scan requests for that workload until the cache expires.

## Proposed Solution
Add retry logic with exponential backoff and jitter for transient HTTP `429` rate limit responses during image resolution and layer downloads before marking the image as rate-limited:

1. **Configurable Retries:** Introduce a retry wrapper with configurable max retries (e.g. 3 attempts) and exponential backoff (e.g. 1s, 2s, 4s with jitter).
2. **Respect `Retry-After` Header:** If the registry response includes a `Retry-After` HTTP header, honor the delay specified by the server.
3. **Graceful Fallback:** If all retries fail with `429`, fallback to the existing behavior (caching `ErrTooManyRequests` in `tooManyRequests`).

## Impact
This prevents temporary, transient registry rate-limit hiccups from marking scans as failed, significantly improving scan reliability in high-concurrency environments.

NaitikRishu:
## Overview
When scanning workloads against public or private container registries (such as Docker Hub, Quay.io, or GHCR), temporary network congestion or burst request spikes can cause registries to intermittently respond with HTTP `429 Too Many Requests` status codes.

Previously, a single transient `429` error on image resolution (`syft.GetSource()`) would immediately cause `kubevuln` to cache a rate-limit failure in `tooManyRequests`, blocking subsequent scan attempts for that workload until the cache TTL expired.

This PR implements exponential backoff retries with random jitter for image resolution calls (`syft.GetSource()`) in both the in-process `SyftAdapter` (`adapters/v1/syft.go`) and the sidecar SBOM scanner server (`pkg/sbomscanner/v1/server.go`) before marking an image as rate-limited.

### Detailed Technical Changes:
1. **Generic Backoff & Retry Engine (`internal/tools/retry.go`)**:
   - Implemented `RetryWithBackoff[T](ctx, config, isRetryable, fn)` supporting configurable max attempts (default: 3), initial delay (200ms), max wait cap (2s), and a 2.0x backoff multiplier.
   - Incorporated 0–25% random jitter on each retry delay to prevent thundering herd bursts against registries.
   - Added context cancellation awareness (`<-ctx.Done()`) so canceled or timed-out parent contexts unblock immediately without executing leftover retries.

2. **Rate Limit Detection (`internal/tools/retry.go`)**:
   - Implemented `IsRateLimitError(err)` to recognize rate limits across multiple representations:
     - `domain.ErrTooManyRequests` sentinel.
     - `*transport.Error` with `StatusCode == 429` (from `go-containerregistry`).
     - Error string signatures (`"429 Too Many Requests"`, `"status code: 429"`, `"TOOMANYREQUESTS"`).

3. **Adapter & Sidecar Integration**:
   - Wrapped `syft.GetSource()` in `adapters/v1/syft.go` across all fallback paths (primary image reference, `MANIFEST_UNKNOWN` tag fallback, registry auth provider retries, and anonymous fallback).
   - Wrapped `syft.GetSource()` in `pkg/sbomscanner/v1/server.go` (`resolveSource`) so sidecar daemon scans benefit from identical retry behavior.

4. **Unit Test Coverage**:
   - Added unit tests in `internal/tools/retry_test.go` covering first-try success, retry recovery, retry exhaustion, non-retryable error handling, and context cancellation.
   - Added `Test_syftAdapter_CreateSBOM_Retry429RateLimit` in `adapters/v1/syft_test.go` testing end-to-end `CreateSBOM` recovery against a mock HTTP registry returning a 429 on initial manifest fetch.

## Additional Information
- The retry logic operates before `ScanService.checkCreateSBOM()` is called, ensuring transient 429s are recovered transparently without triggering rate-limit cache entries.
- If all 3 attempts fail with a rate-limit error, `IsRateLimitError()` still returns true and normal rate-limiting error handling takes effect.

## How to Test
Run unit tests locally:
```bash
go test -v -race -run "TestRetryWithBackoff|Test_syftAdapter_CreateSBOM_Retry429RateLimit" ./internal/tools ./adapters/v1 ./pkg/sbomscanner/v1
```

## Related issues/PRs:
Resolves #613

```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SBOM generation reliability when registry services return rate-limit responses.
  * Automatically retries eligible source retrieval and SBOM generation requests with backoff and server-provided retry timing.
  * Applies consistent rate-limit handling across registry source resolution and scanning workflows.
  * Preserves clear “too many requests” errors, including platform details, when retries are exhausted.
  * Stops retrying promptly when operations are canceled or no longer eligible.

### DIFF
--- a/internal/tools/retry.go
+++ b/internal/tools/retry.go
@@ -16,6 +16,7 @@ import (
 	"github.com/kubescape/kubevuln/internal/metrics"
 )
 
+// RetryConfig specifies configuration for retrying operations.
 type RetryConfig struct {
 	MaxAttempts int
 	InitialWait time.Duration
@@ -31,6 +32,9 @@ type RetryConfig struct {
 	MaxRetryAfter time.Duration
 }
 
+// Default429RetryConfig returns standard retry settings for 429 rate limits:
+// 3 max attempts (1 initial attempt + 2 retries), starting at 500ms initial wait up to 2s,
+// with a 30s ceiling on Retry-After.
 func Default429RetryConfig() RetryConfig {
 	return RetryConfig{
 		MaxAttempts:   3,
@@ -38,6 +42,16 @@ func Default429RetryConfig() RetryConfig {
 		MaxWait:       2 * time.Second,
 		Backoff:       2.0,
 		MaxRetryAfter: 30 * time.Second,
+	}
+}
+
+// FastRetryConfig returns fast retry settings for unit testing (1ms initial wait up to 10ms).
+func FastRetryConfig() RetryConfig {
+	return RetryConfig{
+		MaxAttempts: 3,
+		InitialWait: 1 * time.Millisecond,
+		MaxWait:     10 * time.Millisecond,
+		Backoff:     2.0,
 	}
 }
 
@@ -49,6 +63,7 @@ func (c RetryConfig) retryAfterCeiling() time.Duration {
 	return c.MaxWait
 }
 
+// IsRateLimitError reports whether err indicates a 429 Too Many Requests rate-limiting error.
 func IsRateLimitError(err error) bool {
 	if err == nil {
 		return false
@@ -67,30 +82,37 @@ func IsRateLimitError(err error) bool {
 		strings.Contains(errStr, "received status code: 429")
 }
 
+// ParseRetryAfter attempts to extract a Retry-After duration from error text or header strings.
+// Supports both numeric seconds ("120") and HTTP-date formats (RFC1123/RFC1123Z).
 func ParseRetryAfter(err error) (time.Duration, bool) {
 	if err == nil {
 		return 0, false
 	}
 	errStr := err.Error()
-	idx := strings.Index(errStr, "Retry-After:")
-	if idx == -1 {
-		return 0, false
-	}
-	rawVal := strings.TrimSpace(errStr[idx+len("Retry-After:"):])
-	spaceIdx := strings.IndexAny(rawVal, " \t\r\n,")
-	numStr := rawVal
-	if spaceIdx != -1 {
-		numStr = rawVal[:spaceIdx]
-	}
-	if secs, parseErr := strconv.Atoi(numStr); parseErr == nil && secs > 0 {
-		return time.Duration(secs) * time.Second, true
-	}
-	if t, parseErr := time.Parse(time.RFC1123, rawVal); parseErr == nil {
-		d := time.Until(t)
-		if d <= 0 {
-			d = time.Second
+	idx := strings.Index(strings.ToLower(errStr), "retry-after:")
+	if idx != -1 {
+		sub := strings.TrimSpace(errStr[idx+len("retry-after:"):])
+		fields := strings.Fields(sub)
+		if len(fields) > 0 {
+			token := strings.Trim(fields[0], ";,")
+			if seconds, parseErr := strconv.Atoi(token); parseErr == nil && seconds > 0 {
+				return time.Duration(seconds) * time.Second, true
+			}
 		}
-		return d, true
+		dateStr := sub
+		if endIdx := strings.IndexAny(sub, "\r\n"); endIdx != -1 {
+			dateStr = strings.TrimSpace(sub[:endIdx])
+		}
+		if t, parseErr := time.Parse(time.RFC1123, dateStr); parseErr == nil {
+			if d := time.Until(t); d > 0 {
+				return d, true
+			}
+		}
+		if t, parseErr := time.Parse(time.RFC1123Z, dateStr); parseErr == nil {
+			if d := time.Until(t); d > 0 {
+				return d, true
+			}
+		}
 	}
 	return 0, false
 }
@@ -101,6 +123,10 @@ func RetryWithBackoff[T any](ctx context.Context, operation string, config Retry
 	wait := config.InitialWait
 
 	for attempt := 1; ; attempt++ {
+		if ctx.Err() != nil {
+			return zero, ctx.Err()
+		}
+
 		val, err := fn(ctx)
 		if err == nil {
 			if attempt > 1 {
@@ -122,6 +148,7 @@ func RetryWithBackoff[T any](ctx context.Context, operation string, config Retry
 
 		metrics.RecordRetryAttempt(ctx, operation, metrics.RetryOutcomeAttempt)
 
+		// Calculate wait duration with Retry-After header or backoff + jitter
 		delay := wait
 		if retryAfter, ok := ParseRetryAfter(err); ok {
 			delay = retryAfter
@@ -142,6 +169,7 @@ func RetryWithBackoff[T any](ctx context.Context, operation string, config Retry
 		}
 
 		logger.L().Ctx(ctx).Debug("retrying operation after 429 rate limit",
+			helpers.String("operation", operation),
 			helpers.Int("attempt", attempt),
 			helpers.Int("maxAttempts", config.MaxAttempts),
 			helpers.String("delay", delay.String()),
--- a/pkg/sbomscanner/v1/server.go
+++ b/pkg/sbomscanner/v1/server.go
@@ -31,12 +31,6 @@ import (
 	"github.com/kubescape/storage/pkg/apis/softwarecomposition/v1beta1"
 )
 
-// isRegistryRateLimited reports whether err is (or wraps) a registry 429 response.
-// It delegates to tools.IsRateLimitError.
-func isRegistryRateLimited(err error) bool {
-	return tools.IsRateLimitError(err)
-}
-
 // isPlatformMismatch reports whether err is (or wraps) stereoscope's
 // *image.ErrPlatformMismatch, returned once a provider has positively resolved the image but
 // its OS/architecture doesn't match the platform that was requested.
@@ -161,17 +155,16 @@ func (s *scannerServer) CreateSBOM(ctx context.Context, req *pb.CreateSBOMReques
 		}
 	}()
 
-	// Download image from registry, retrying on 429 rate-limit errors with exponential backoff.
+	// Download image from registry
 	logger.L().Debug("downloading image", helpers.String("imageID", imageID))
-	src, err := tools.RetryWithBackoff(ctx, "source_resolution", tools.Default429RetryConfig(), tools.IsRateLimitError, func(rCtx context.Context) (source.Source, error) {
-		return resolveSource(rCtx, func(_ context.Context, ref string, opts *image.RegistryOptions) (source.Source, error) {
-			// Pulls intentionally run on a detached context (see #421); the request ctx is used only for the registry auth provider inside resolveSource.
-			//nolint:staticcheck // stereoscope expects string key image.MaxImageSize
-			ctxWithSize := context.WithValue(context.Background(), image.MaxImageSize, req.MaxImageSize)
+	src, err := resolveSource(ctx, func(_ context.Context, ref string, opts *image.RegistryOptions) (source.Source, error) {
+		// Pulls intentionally run on a detached context (see #421); the request ctx is used only for the registry auth provider inside resolveSource.
+		return tools.RetryWithBackoff(context.Background(), "source_resolution", tools.Default429RetryConfig(), tools.IsRateLimitError, func(retryCtx context.Context) (source.Source, error) {
+			ctxWithSize := context.WithValue(retryCtx, image.MaxImageSize, req.MaxImageSize) //nolint:staticcheck // stereoscope requires string context key
 			return syft.GetSource(ctxWithSize, ref,
 				syft.DefaultGetSourceConfig().WithRegistryOptions(opts).WithPlatform(imgPlatform).WithSources("registry"))
-		}, imageID, imageTag, registryOptions)
-	})
+		})
+	}, imageID, imageTag, registryOptions)
 
 	switch {
 	case err != nil && (errors.Is(err, image.ErrImageTooLarge) || strings.Contains(err.Error(), image.ErrImageTooLarge.Error())):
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: pkg/sbomscanner/v1/server.go [hunk 1]
```diff
@@ -230,11 +231,13 @@ func (s *scannerServer) CreateSBOM(ctx context.Context, req *pb.CreateSBOMReques
...
 	// Download image from registry
 	logger.L().Debug("downloading image", helpers.String("imageID", imageID))
-	src, err := resolveSource(ctx, func(_ context.Context, ref string, opts *image.RegistryOptions) (source.Source, error) {
+	src, err := resolveSource(ctx, func(callCtx context.Context, ref string, opts *image.RegistryOptions) (source.Source, error) {
 		// Pulls intentionally run on a detached context (see #421); the request ctx is used only for the registry auth provider inside resolveSource.
-		ctxWithSize := context.WithValue(context.Background(), image.MaxImageSize, req.MaxImageSize)
-		return syft.GetSource(ctxWithSize, ref,
-			syft.DefaultGetSourceConfig().WithRegistryOptions(opts).WithPlatform(imgPlatform).WithSources("registry"))
+		return tools.RetryWithBackoff(callCtx, tools.Default429RetryConfig(), tools.IsRateLimitError, func(retryCtx context.Context) (source.Source, error) {
+			ctxWithSize := context.WithValue(retryCtx, image.MaxImageSize, req.MaxImageSize) //nolint:staticcheck // stereoscope requires string context key
```
matthyx, line 163: **Blocker: the pull is no longer running on a detached context, contradicting the comment right above it.**

Before this PR, `resolveSource` was called with a callback that ignored its `ctx` parameter (`_ context.Context`) and always used `context.Background()`:

```go
src, err := resolveSource(ctx, func(_ context.Context, ref string, opts *image.RegistryOptions) (source.Source, error) {
    // Pulls intentionally run on a detached context (see #421); ...
    ctxWithSize := context.WithValue(context.Background(), image.MaxImageSize, req.MaxImageSize)
    ...
```

Now the callback captures `callCtx` (which `resolveSource` invokes as the request-scoped `ctx` passed into `CreateSBOM`) and feeds it straight into `tools.RetryWithBackoff` and `context.WithValue`, so the pull is tied to the gRPC request context again — while the unmodified comment still claims it's detached:

```go
src, err := resolveSource(ctx, func(callCtx context.Context, ref string, opts *image.RegistryOptions) (source.Source, error) {
    // Pulls intentionally run on a detached context (see #421); the request ctx is used only for the registry auth provider inside resolveSource.
    return tools.RetryWithBackoff(callCtx, tools.Default429RetryConfig(), tools.IsRateLimitError, func(retryCtx context.Context) (source.Source, error) {
        ctxWithSize := context.WithValue(retryCtx, image.MaxImageSize, req.MaxImageSize) //nolint:staticcheck
        ...
```

Net effect: if the gRPC client disconnects or the caller's deadline is shorter than the retry chain's worst case (up to 3 attempts × up to 2s backoff, plus network time), the whole point of this PR — surviving a transient 429 — gets undermined, because the pull (and its retries) now abort the moment the request context is canceled. That's exactly the regression #421 was meant to prevent.

Suggested fix: keep passing `context.Background()` into `tools.RetryWithBackoff` here (same as the pre-PR behavior), not `callCtx`.

internal/tools/retry.go
```diff
@@ -0,0 +1,151 @@
...
+
+		// Calculate wait duration with Retry-After header or backoff + jitter
+		delay := wait
+		if retryAfter, ok := ParseRetryAfter(err); ok {
+			delay = retryAfter
+		} else if delay > 0 {
+			jitter := time.Duration(rand.Float64() * 0.25 * float64(delay))
+			delay += jitter
+		}
+
```
matthyx: **Server-provided `Retry-After` gets silently clamped to `config.MaxWait` (2s by default).**

```go
delay := wait
    delay = retryAfter
} else if delay > 0 {
    ...
}

if delay > config.MaxWait {
    delay = config.MaxWait
}
```

`ParseRetryAfter` is explicitly built to honor the registry's `Retry-After` header/value (per the PR description: "Calculate wait duration with Retry-After header or backoff + jitter"), but real registries (Docker Hub, GHCR, etc.) commonly send `Retry-After` values well over 2 seconds. This code unconditionally caps even a server-specified `Retry-After` down to `config.MaxWait`, so we end up retrying *sooner* than the registry told us to — which is a mild spec violation (RFC 9110 §10.2.3 treats `Retry-After` as a minimum wait) and risks provoking the very rate limiting this PR is trying to avoid.

This was already flagged by an earlier CodeRabbit pass on this PR and doesn't look addressed yet. Suggest only applying the `MaxWait` cap to the computed backoff+jitter branch, not to a server-supplied `Retry-After`.

matthyx (changes requested):
Solid feature overall — the generic `RetryWithBackoff[T]` engine, jitter, and test coverage are well done. Found two regressions that should block merge, plus one already-flagged issue that looks unaddressed:

1. **Detached-context regression in `pkg/sbomscanner/v1/server.go`** (inline comment) — the sidecar's image pull used to run on `context.Background()` on purpose (per #421); it's now wired to the gRPC request context via `callCtx`, so a client disconnect/short deadline can now abort an otherwise-recoverable retry sequence. The comment above the code still claims the old (detached) behavior, so this reads like an unintentional side effect of adding the retry wrapper.

2. **`Retry-After` gets clamped by `config.MaxWait` in `internal/tools/retry.go`** (inline comment) — a server-specified `Retry-After` (frequently >2s from real registries) is capped down to `MaxWait`, causing retries earlier than the registry asked for. This was already raised by an earlier CodeRabbit pass on this PR and still isn't fixed at HEAD.

3. **Rate-limit classifier mismatch in `pkg/sbomscanner/v1/server.go`** — `RetryWithBackoff` decides whether to retry using `tools.IsRateLimitError` (matches `domain.ErrTooManyRequests`, `transport.Error{429}`, and text like `"TOOMANYREQUESTS"`, `"429 Too Many Requests"`, `"status code: 429"`, `"received status code: 429"`). But after retries are exhausted, the final response is classified with the narrower, pre-existing `isRegistryRateLimited` (only `transport.Error{429}` or `"TOOMANYREQUESTS"`/`"429 Too Many Requests"` text — no `errors.Is(domain.ErrTooManyRequests)` check, no `"status code: 429"` match). So an exhausted-retry error that matched `IsRateLimitError` (e.g. a bare `domain.ErrTooManyRequests`, or an `err.Error()` containing `"status code: 429"` without the exact `"429 Too Many Requests"` string) can fail to set `StatusReason: domain.ReasonTooManyRequests`, silently losing the rate-limit caching this PR's description says it preserves. This was also flagged by the earlier CodeRabbit review ("Update the post-retry rate-limit classification ... to use tools.IsRateLimitError") and doesn't look resolved at HEAD — worth reconciling the two classifiers (or reusing one for both).

Happy to re-review once these are addressed.

NaitikRishu:
Thanks @matthyx! I've updated the PR to address all feedback:

1. **Restored Detached Context (#421)**: Updated `pkg/sbomscanner/v1/server.go` so `tools.RetryWithBackoff` receives `context.Background()` during image resolution, ensuring pulls and retries remain detached from the gRPC request context.
2. **Unclamped `Retry-After` (RFC 9110)**: Modified `internal/tools/retry.go` so server-provided `Retry-After` headers are honored directly without being capped by `config.MaxWait`. `MaxWait` now applies exclusively to calculated exponential backoff delays.
3. **Preserved Rate-Limit Status on Cataloging Failure**: Updated `server.go` so `CreateSBOM` returns `domain.ReasonTooManyRequests` when `createSBOMFn` exhausts 429 retries, preserving rate-limit caching behavior.
4. **Unit Tests**: Added `TestRetryWithBackoff_HonorsServerRetryAfterUnclamped` and `TestCreateSBOM_Exhausted429RateLimitFromCreateSBOMFn`.
5. **Rebased**: Rebased cleanly onto the latest `main` branch.

All unit tests and linter checks pass cleanly. Ready for re-review!

matthyx (changes requested):
Re-reviewed after the latest commit (84185c1e).

**Fixed, resolving those threads:**
- Detached-context regression in `pkg/sbomscanner/v1/server.go` — back to `context.Background()`, matching the pre-PR `#421` behavior. ✅
- `Retry-After` no longer gets clamped by `config.MaxWait` — the cap now only applies to the computed backoff+jitter branch, and there's a new test (`TestRetryWithBackoff_HonorsServerRetryAfterUnclamped`) proving it. ✅

**Still open — rate-limit classifier mismatch, half-fixed:**
The SBOM-*generation* path (`createSBOMFn` retry, around the second `switch` after `dl.Run`) now correctly classifies exhausted-retry errors with `tools.IsRateLimitError` and returns `domain.ReasonTooManyRequests` — nice, and there's a new test for it (`TestCreateSBOM_Exhausted429RateLimitFromCreateSBOMFn`).

But the image-*resolution* path (`resolveSource` / `case err != nil && isRegistryRateLimited(err):`, right after the `resolveSource(...)` call) was **not** touched — it still uses the narrower, pre-existing `isRegistryRateLimited`, which only matches `*transport.Error{429}` or the literal text `"TOOMANYREQUESTS"` / `"429 Too Many Requests"`. `tools.IsRateLimitError` (used to decide whether `RetryWithBackoff` retries the pull) is broader — it also matches `errors.Is(domain.ErrTooManyRequests)` and the text patterns `"status code: 429"` / `"received status code: 429"`.

So the original bug I flagged is still live for this specific path: if the pull's error, after 3 exhausted retries, renders as e.g. `"received status code: 429"` (a real go-containerregistry error format — it's literally one of the test cases for `tools.IsRateLimitError`), `RetryWithBackoff` will have retried it correctly, but the post-exhaustion `isRegistryRateLimited(err)` check here won't match it, so the response falls through to the generic `case err != nil:` branch without `StatusReason: domain.ReasonTooManyRequests` — losing the rate-limit caching this PR's own description says it preserves.

Suggest applying the same fix used for the generation path: swap `isRegistryRateLimited` for `tools.IsRateLimitError` in the `resolveSource`-path switch too (or drop `isRegistryRateLimited` entirely in favor of the one shared classifier), and add a resolution-path equivalent of `TestCreateSBOM_Exhausted429RateLimitFromCreateSBOMFn`.

Once that's aligned, this looks ready to merge.

matthyx (dismissed):
Re-reviewed after commit 4d4682e. The remaining classifier mismatch is fixed properly this time — `isRegistryRateLimited` (server.go) and `isRegistryRateLimitedErr` (core/services/scan.go) were both removed in favor of the single shared `tools.IsRateLimitError`, used consistently for the retry decision, the post-retry `resolveSource` classification, the SBOM-generation classification, and the `ScanService` rate-limit cache. New test `TestCreateSBOM_Exhausted429RateLimitFromResolveSource` covers exactly the `"received status code: 429"` case that was previously falling through the cracks.

All three functional issues I raised are now resolved. Approving on that basis.

Two things worth checking before merging, unrelated to correctness:
- `mergeable_state` is currently `dirty` (conflicts with `main`) — needs a rebase/merge to resolve.
- The full test/lint/build CI suite (`pr-created` workflow) doesn't appear to have run yet for this commit (only GitGuardian + DCO checks are present) — worth confirming it's green before merge.

NaitikRishu:
Hi @matthyx,

I've rebased the branch cleanly onto latest `main`, resolved the conflicts, and verified all unit tests locally with `go test -v -race -cover ./...`. Ready whenever you have time to take a look!

matthyx (dismissed):
The branch was rebased/squashed (now a single commit, 6a603a5), which dismissed my prior approval — re-reviewed the new head and re-approving.

Confirmed all three previously-flagged issues are still fixed after the rebase: the sidecar pull retry uses `context.Background()` (not the request ctx), `Retry-After` is honored unclamped, and both resolution-path and generation-path exhausted-retry errors are classified with the single shared `tools.IsRateLimitError` (the old narrower `isRegistryRateLimited`/`isRegistryRateLimitedErr` helpers are gone). Full CI (lint, build, tests) is now green on this commit too.

Still outstanding, both non-blocking for my review: `mergeable_state` is `dirty` (conflicts with `main`, needs resolving before merge), and there's a fresh CodeRabbit nitpick about checking `ctx.Err()` before the first `fn` call in `RetryWithBackoff` — worth a look but not correctness-critical (it only means retrying once past a client cancellation before context awareness kicks in, not a stuck retry loop).

matthyx (approved):
Re-reviewed after commit 1199816 (single commit now, rebased onto current main).

Good refactor since my last approval: the MANIFEST_UNKNOWN/401 fallback ladder that both `adapters/v1/syft.go` and `pkg/sbomscanner/v1/server.go` duplicated is now a single shared `registryauth.ResolveSource`, and the `Retry-After` ceiling is now its own `MaxRetryAfter` (30s) instead of being capped by `MaxWait` (2s) — a cleaner fix than before. Also picked up the CodeRabbit nitpick from the previous round: `RetryWithBackoff` now checks `ctx.Err()` before each attempt, with tests for cancel-before-first-call and cancel-mid-retry.

All three originally-flagged issues remain fixed, and the rate-limit classification is now more complete than what I approved last time: the sidecar's SBOM-*generation* exhausted-retry path (after `dl.Run`) now also returns `domain.ReasonTooManyRequests` via `tools.IsRateLimitError`, closing a gap that existed even in the version I approved on the 14th (previously that path fell through to the generic Incomplete/error response).

`mergeable_state` is now `MERGEABLE` (conflict resolved) and CI (lint, build, tests, CodeQL, GitGuardian, DCO) is green.

One open, non-blocking nitpick from CodeRabbit's latest pass: `ParseRetryAfter`'s numeric-seconds branch (`time.Duration(seconds) * time.Second`) can overflow `int64` for a very large `Retry-After` value from a malicious/broken registry, which could produce a negative delay. Worth a bounds check, but not correctness-critical given `MaxAttempts` bounds the retry loop regardless.

Approving.

PR #616, opened 2026-08-11
fix(adapters): log the SBOM size limit under its own name

### LINKED ISSUE #615: Bug: the in-process SBOM-too-large warning reports maxSBOMSize under the field name maxImageSize
the in-process `SBOM exceeds size limit` warning prints `maxSBOMSize` but labels it `maxImageSize`:

```go
if sz > s.maxSBOMSize {
	logger.L().Ctx(ctx).Warning("SBOM exceeds size limit",
		helpers.Int("maxImageSize", s.maxSBOMSize),
```

defaults are 512 MiB for maxImageSize and 20 MiB for maxSBOMSize, so on a default setup you get:

```
SBOM exceeds size limit  maxImageSize=20971520
```

when maxImageSize is actually 536870912. reading that, you'd raise maxImageSize and nothing would change.

the sidecar logs the same message with the right name (`pkg/sbomscanner/v1/server.go`), so the two paths disagree and a query on maxSBOMSize misses in-process ones.

version: main (54f68b5)

AnshulPatil2005:
## Overview

the in-process `SBOM exceeds size limit` warning printed `s.maxSBOMSize` but labelled it `maxImageSize`. defaults are 512 MiB and 20 MiB, so on a default setup the line reads `maxImageSize=20971520` while maxImageSize is actually 536870912. you'd raise the wrong option and nothing would change.

the sidecar already logs the same message with the right field name, so the two paths disagreed too.

## Additional Information

no test, nothing here asserts log output and a capture harness would be bigger than the fix. the guard right above is `if sz > s.maxSBOMSize` and the annotation two lines below is already `MaxSBOMSizeAnnotationKey`, so the field name was the only thing naming the wrong option.

swept the rest of the repo for the same mismatch, no others.

## How to Test

`go test ./adapters/v1/` passes unchanged, only the log field name differs:

```
before:  SBOM exceeds size limit  maxImageSize=20971520
after:   SBOM exceeds size limit  maxSBOMSize=20971520
```

## Related issues/PRs:

* Resolved #615

### DIFF
--- a/adapters/v1/syft.go
+++ b/adapters/v1/syft.go
@@ -377,7 +377,7 @@ func (s *SyftAdapter) CreateSBOM(ctx context.Context, name, imageID, imageTag st
 	if sz > s.maxSBOMSize {
 		metrics.RecordScanFallback(ctx, metrics.ComponentInProcess, metrics.FallbackCategorySizeClassification, metrics.FallbackStrategySBOMTooLarge, metrics.FallbackOutcomeClassified)
 		logger.L().Ctx(ctx).Warning("SBOM exceeds size limit",
-			helpers.Int("maxImageSize", s.maxSBOMSize),
+			helpers.Int("maxSBOMSize", s.maxSBOMSize),
 			helpers.Int("size", sz),
 			helpers.String("imageID", imageID))
 		domainSBOM.Status = helpersv1.TooLarge


matthyx (approved):
Verified the fix: `s.maxSBOMSize` was logged under the `maxImageSize` key at adapters/v1/syft.go:380, which is misleading since the adjacent "Image exceeds size limit" branch and the sidecar (`pkg/sbomscanner/v1/server.go`) both correctly use `maxImageSize`/`maxSBOMSize` for their respective fields. This PR aligns the in-process log with the sidecar's naming. Confirmed no other `maxImageSize`/`maxSBOMSize` mismatches exist elsewhere in the repo. One-line change, all CI checks green, no behavioral risk. LGTM.

PR #619, opened 2026-08-11
fix(registryauth): count upstream credential fetches, not missed lookups

### LINKED ISSUE #618: registry auth cache metric counts missed lookups, not upstream fetches
`kubevuln_registry_auth_cache_total` counts lookups that missed, not upstream credential fetches, so it can't show whether the coalescing added in #571 is working.

in `internal/registryauth/cache.go`, every path out of the singleflight records a miss:

```go
select {
case <-ctx.Done():
	metrics.RecordRegistryAuthCache(ctx, c.strategy, metrics.RegistryAuthCacheMiss)
	return nil, ctx.Err()
case res := <-ch:
	metrics.RecordRegistryAuthCache(ctx, c.strategy, metrics.RegistryAuthCacheMiss)
	...
}
```

so 20 concurrent scans against the same registry produce 1 call to `ECRCredsFn` and 20 `result="miss"` increments. that's the exact case #571 exists to fix, and the metric looks identical before and after.

two smaller things on the same path:

- the re-check inside the singleflight closure (`if creds, ok := c.lookup(key); ok`) returns a value served from cache, but the caller still records `miss`.
- the `ctx.Done()` branch records a miss when the caller gave up waiting. no lookup outcome happened at all there.

the doc in `docs/API.md` describes the label as "cache result (hit/miss)", which matches what the code does, so the metric isn't wrong so much as it answers a different question than the one #571 raises.

a few ways to close it, no strong preference:

- record the fetch inside the closure, so there's an exact count of upstream calls
- add a third `result` value for callers that waited on someone else's fetch
- leave `hit`/`miss` as caller-side outcomes and add a separate counter for fetches

version: main (54f68b5), follow-up to #571

AnshulPatil2005:
## Overview

`kubevuln_registry_auth_cache_total{result="miss"}` counted callers whose lookup missed the cache, not upstream credential fetches. Since #571 collapses concurrent lookups for the same key into one fetch, those two numbers stopped being the same thing: 20 concurrent scans against one registry made 1 call to `ECRCredsFn` and 20 `miss` increments, so the metric read identically with and without the coalescing that #571 added.

this picks the second option from the issue, a third `result` value, since it keeps one metric answering one question (what did this lookup cost) and the answer to "how many upstream fetches" falls out of it as `miss`.

after this:

- `hit` served from the cache
- `coalesced` missed, but was served by another caller's in-flight fetch, so no upstream request
- `miss` reached the cloud provider, so summing it gives the number of upstream fetches

the miss is recorded inside the singleflight closure instead of at the select. that is what makes one fetch one miss even when the caller that started it abandons its wait, since the fetch keeps running and still populates the cache. it also fixes the two smaller things from the issue: the re-check inside the closure now records the `hit` it actually was, and a caller that gave up on `ctx.Done()` records nothing, because it never got a lookup outcome either way.

## Additional Information

`ran` is a plain bool, not atomic. each caller passes its own closure to `DoChan` and singleflight runs exactly one of them, so a caller either owns the write or there is no write at all. it is only read in the `<-ch` branch, where the receive happens-after the closure returns. reading it on the `ctx.Done()` path would be a real race, so there is a comment on it. tests run under `-race`.

`hit + miss + coalesced` is now lookups minus abandonments rather than total lookups. that seemed better than keeping a count for callers that walked away, but happy to add a fourth value if you would rather the total stayed whole.

doc row in `docs/API.md` and the counter description updated for the new value.

## How to Test

`go test ./internal/registryauth/ ./internal/metrics/ -race -count=1`

this needed one new test seam. the existing `onMiss` hook fires just before `DoChan`, so a caller can be descheduled in that gap, miss the in-flight execution entirely, and start a fresh one that re-checks straight into a cache hit. that is a real outcome, just not the one the test is about, and it made the first version of this test flaky in CI (18 coalesced, 2 hit). `onJoined` fires once a caller has actually joined the group, so the barrier releases the leader only when all 20 are provably waiting on the same fetch. same shape as `onMiss`, nil-checked, tests only.

two new tests, both scraping the real prometheus handler rather than asserting on a fake:

- `TestCredentialCache_CoalescedCallersAreNotCountedAsUpstreamFetches` uses that barrier so 20 callers provably contend on one key, then asserts `miss` is 1 and `coalesced` is 19, and that the next lookup is a `hit`.
- `TestCredentialCache_AbandonedCallerRecordsNoOutcome` cancels a waiter while the leader is still in `fetch`, and asserts no `coalesced` and that `miss` stays at the one real fetch.

against the current code the first one fails with exactly the number from the issue:

```
kubevuln_registry_auth_cache_total{result="miss",strategy="ecr"} 20
  does not contain
kubevuln_registry_auth_cache_total{result="miss",strategy="ecr"} 1
```

## Related issues/PRs:

* Resolved #618
* Follow-up to #571


AnshulPatil2005:
good catch on the recheck hit. the closure recorded it immediately, so a leader whose ctx was cancelled before the receive still counted one, which is not what the ctx.Done comment claims. moved it out: the closure now only marks what this caller owes, and hit/coalesced are both recorded where the caller is actually served. miss stays in the closure on purpose, one upstream call has to be one miss whether or not the caller sticks around.

added TestCredentialCache_RecheckInsideSingleflightRecordsHit, using onMiss to populate the cache in the window before joining, which is the only way the recheck ever finds anything.

matthyx (dismissed):
Reviewed `cache.go`, `metrics.go`, `cache_test.go`, and the doc update.

The fix is correct: `miss` is now recorded once per singleflight closure execution (unconditionally, before the fetch call), independent of whether the caller that triggered it sticks around — so `sum(miss)` really is upstream fetch count, matching the goal. `hit`/`coalesced` are decided via the per-caller `pending` local, only recorded after receiving from `ch`, never on the `ctx.Done()` path.

Traced the happens-before for the `pending` variable specifically, since it's written in the singleflight leader's spawned goroutine (`go g.doCall(...)` under the hood) and read back in the original caller's goroutine: the closure's return is sequenced-before `doCall` sending on the per-caller channel, and channel send happens-before the corresponding receive, so reading `pending` in the `case res := <-ch` branch is properly synchronized — not a data race, consistent with `-race` passing in CI.

@chatgpt-codex-connector's earlier comment ("record recheck hits only after the caller receives them") is already addressed in the current single commit — the recheck-hit assignment only sets `pending`, the actual `RecordRegistryAuthCache` call moved to after the `<-ch` receive, exactly as your follow-up comment describes. That thread shows as outdated/unresolved but the concern is resolved in the code as it stands.

Tests are a good sign: `TestCredentialCache_CoalescedCallersAreNotCountedAsUpstreamFetches` uses the new `onJoined` barrier so leader/followers are provably contending before the fetch proceeds (no timing-based flakiness), and `TestCredentialCache_AbandonedCallerRecordsNoOutcome` deterministically orders miss-recording before the caller can even observe `started`, since `RecordRegistryAuthCache(Miss)` happens before `fetch()` is invoked. All three scrape the real Prometheus handler rather than a fake, which is the right call for a bug that was specifically about metric label counting.

No blockers from me. Approving.

matthyx:
rebase pls

AnshulPatil2005:
> rebase pls

done

matthyx (approved):
Re-checked after the rebase (base moved to `746a949`, head now `c498ae4`). Diffed against my earlier review: the actual patch content in `docs/API.md`, `internal/metrics/metrics.go`, `internal/registryauth/cache.go`, and `internal/registryauth/cache_test.go` is unchanged — this was a clean rebase onto a newer `main`, not a code change. All CI checks (tests, lint, CodeQL, DCO, GitGuardian) are green on the new commit.

Re-approving since my prior approval was dismissed by the rebase.

PR #621, opened 2026-08-11
chore(repositories): give the timestamp annotation key a constant

### LINKED ISSUE #620: kubescape.io/timestamp is the only metadata key written as a bare string
`repositories/apiserver.go` still has an open TODO on the one metadata key kubevuln writes as a bare string:

```go
enrichedAnnotations["kubescape.io/timestamp"] = strconv.FormatInt(timestamp, 10) // TODO: use a constant
enrichedAnnotations[helpersv1.WlidMetadataKey] = workload.Wlid
enrichedAnnotations[helpersv1.ContainerNameMetadataKey] = workload.ContainerName
```

the two lines under it, and every other annotation and label kubevuln sets on these objects, come from `k8s-interface/instanceidhandler/v1/helpers`. this one is the odd one out.

the reason it never became a constant looks like it is that there is nothing to point it at. `helpers/keys.go` has no key for it. the closest is `ReportTimestampMetadataKey` (`kubescape.io/report-timestamp`), which is a different key that storage reads off container profiles, not this. so the constant has to be a local one rather than a swap to an existing key.

small thing, but the string is duplicated in `apiserver_test.go` too, and it is the sort of key that should not quietly change: it is stamped on every `VulnerabilityManifestSummary` kubevuln has ever written, so a different string would strand the annotation on all of them and stop anything reading it, without failing anything.

version: main (54f68b5)

AnshulPatil2005:
## Overview

closes the `// TODO: use a constant` on `kubescape.io/timestamp` in `enrichSummaryManifestObjectAnnotations`, the only metadata key kubevuln writes as a bare string. the two lines under it and every other key on these objects come from `helpersv1`.

it is a local constant rather than a swap to an existing one because `k8s-interface/instanceidhandler/v1/helpers` has no key for this. `ReportTimestampMetadataKey` (`kubescape.io/report-timestamp`) is the closest and is a different key, written by a different component onto container profiles. the comment on the constant says that, so the next person does not have to go looking for the upstream key that would have replaced it.

no behaviour change, the annotation kubevuln writes is byte for byte the same.

## Additional Information

the refactor costs one thing on its own: `apiserver_test.go` had the key as a literal, so it was accidentally pinning the value, and reading it through the constant instead means the code and the test now agree with each other whatever the string says. `TestTimestampMetadataKeyIsStable` puts that back explicitly.

worth having, because this key is stamped on every `VulnerabilityManifestSummary` already in storage. changing the string would leave the old annotation on all of them and stop anything reading it, and nothing would fail.

## How to Test

`TestAPIServerStore_enrichSummaryManifestObjectAnnotations` covers the annotation itself and is unchanged apart from reading through the constant. `TestTimestampMetadataKeyIsStable` fails if the string changes.

## Related issues/PRs:

* Resolved #620

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -43,6 +43,15 @@ import (
 const (
 	vulnerabilityManifestSummaryKindPlural string = "vulnerabilitymanifests"
 	vulnSummaryContNameFormat              string = "%s-%s-%s" // "<kind>-<name>-<container-name>"
+
+	// timestampMetadataKey is the scan time, in Unix seconds, stamped on every summary
+	// manifest kubevuln writes. It lives here rather than coming from
+	// k8s-interface/instanceidhandler/v1/helpers, like every other metadata key on these
+	// objects, because that package has no key for it: its nearest neighbour,
+	// ReportTimestampMetadataKey ("kubescape.io/report-timestamp"), is a different key
+	// written by a different component. Changing the string would orphan the annotation on
+	// already-stored manifests, so it stays as it is.
+	timestampMetadataKey string = "kubescape.io/timestamp"
 )
 
 // securityExceptionListCacheCleaningInterval/TTL bound how stale the raw SecurityException/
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed the diff, description, and CI (all 8 checks green; CodeRabbit hit its OSS rate limit and didn't produce a review).

The change is a clean, behavior-preserving refactor:
- `timestampMetadataKey` constant replaces the bare `"kubescape.io/timestamp"` string literal and closes the `// TODO: use a constant`.
- The comment on the constant correctly explains why it's local rather than sourced from `helpersv1` (no matching key there; `ReportTimestampMetadataKey` is a different key from a different component) — useful for the next reader.
- `apiserver_test.go` now reads the annotation through the constant instead of pinning the literal, and the new `TestTimestampMetadataKeyIsStable` explicitly locks the string value, which matters since this key is already stamped on stored `VulnerabilityManifestSummary` objects and silently changing it would orphan old annotations.
- No other bare-string metadata keys remain on these objects; scope is tight and matches the stated intent.

No blockers found. Approving.

PR #623, opened 2026-08-11
fix(services): persist TooLarge SBOM markers and stop double-storing

### LINKED ISSUE #622: GenerateSBOM re-stores cached SBOMs and never persists a TooLarge verdict
*rewritten after #610 landed, which moved all four flows onto `getOrCreateSBOM` and changed the shape of this. the original text described the pre-#610 code.*

three things in `getOrCreateSBOM`, all about what gets written to storage.

**a TooLarge or Incomplete verdict is never stored.** the worker gates its store on content:

```go
if s.storage && sbom.Content != nil {
	if storeErr := s.sbomRepository.StoreSBOM(workerCtx, sbom, false); storeErr != nil {
```

but `adapters/v1/syft.go` returns both of those statuses well before `domainSBOM.Content` is assigned:

```go
case err != nil && (errors.Is(err, image.ErrImageTooLarge) || ...):
	domainSBOM.Status = helpersv1.TooLarge
	domainSBOM.Annotations[domain.StatusReasonAnnotationKey] = domain.ReasonImageTooLarge
	domainSBOM.Annotations[domain.MaxImageSizeAnnotationKey] = fmt.Sprintf("%d", s.maxImageSize)
	return domainSBOM, nil
```

so `Content` is nil and nothing is written. `StoreSBOM` handles that case deliberately, storing the object as a status-only marker:

```go
if sbom.Content != nil {
	manifest.Spec.Syft = *sbom.Content
}
...
manifest.Annotations[helpersv1.StatusMetadataKey] = sbom.Status
```

and `getSBOM` re-checks a stored marker against the current `maxImageSize`/`maxSBOMSize` so it stops applying if the limits are raised. that whole path is unreachable right now. the effect is that every scan of an oversized image re-pulls it to reach the answer it already had. before #610 the three scan flows stored the SBOM unconditionally inside the create branch, so the marker did get written.

**GenerateSBOM stores a second time.** it gates on the flag `getOrCreateSBOM` returns, which is `!res.Shared`. that is singleflight's "this result was not shared with other callers", not "this call created something". so a call that is not deduplicated stores the SBOM the worker has already stored.

**and it stores SBOMs it only read.** same flag: on a cache hit the worker returns early without storing, `res.Shared` is false, and the flow writes the SBOM it just read straight back out. `StoreSBOM` is a `Create` that falls back to `Get` plus a full `Update` of the whole spec, so that is a full rewrite of the document.

counted with a `StoreSBOM` counter on `MemoryStore`, against main:

```
TooLarge:      stores=0, and CreateSBOM runs again on the next call
newly created: stores=2
cache hit:     stores=1
```

version: main (cb4aa30)

AnshulPatil2005:
*rebased and rewritten on top of #610, which landed while this was open and moved all four flows onto `getOrCreateSBOM`. the earlier version of this PR described the pre-#610 code.*

## Overview

three fixes in `getOrCreateSBOM`, all about what reaches storage.

**store a TooLarge verdict, which has no content.** the worker gated its store on `sbom.Content != nil`, but the syft adapter returns TooLarge and Incomplete before `Content` is assigned, so the verdict was never written. `StoreSBOM` persists a content-less SBOM as a status-only marker on purpose, and `getSBOM` re-checks that marker against the current `maxImageSize`/`maxSBOMSize` so it stops applying when the limits change. with nothing ever written, that path could not run and every scan of an oversized image re-pulled it to reach the answer it already had. before #610 the scan flows stored unconditionally inside the create branch, so the marker did get written.

Incomplete stays unstored, which is the one place this does not simply restore the pre-#610 behaviour. that status is a timeout rather than a verdict, and `getSBOM` has no staleness rule for it, so a stored one would read back as a cache hit at the same scanner version forever and no retry would ever happen. raised by both bots on the first version of this, and they were right.

**return the store error instead of `!res.Shared`.** `GenerateSBOM` gated its own store on that flag, which is singleflight's "this result was not shared", not "this call created something". so a call that was not deduplicated stored a newly created SBOM twice, and stored a cached one that had just been read back. the worker already stores what it creates, so what the flow actually needs is whether that store failed. it now gets exactly that, and keeps reporting `ReasonSBOMStorageFailed` without repeating the write.

the scan flows are untouched: they take `sbom, _, err` as before, and still ignore a store failure and carry on with an SBOM they could not persist. only `GenerateSBOM`, whose purpose is to store, treats it as a failure.

## Additional Information

the status check stays ahead of the store error, so a too-large image is still reported as `ReasonSBOMTooLarge` rather than a storage failure, which is the precedence it had when the store sat below that check.

`MemoryStore.GetSBOM` returned whatever was put in, so a stored marker read back with nil `Content` and looked absent, while `APIServerStore` always returns `Content: &manifest.Spec.Syft` and reads it back as present. that gap is why the missing marker was invisible in tests, so the double now matches the real store. it also counts `StoreSBOM` calls, same idea as the `summaryStubs` field already there. `MemoryStore` is test-only.

`NewMockSBOMAdapter` always fills `Content`, including for its Incomplete case, so it cannot represent this at all. the new test uses a small creator that mirrors syft's TooLarge path instead.

also replaces the `// FIXME check if we still use this method` above `GenerateSBOM` with the answer, since I had to work it out anyway: `cmd/http/main.go` routes `POST v1/generateSBOM` to it, through `ValidateGenerateSBOM` and the same rejection and scan metrics as the other flows.

## How to Test

`go test ./... -count=1`

- `TestScanService_GenerateSBOM_RemembersTooLargeVerdict` asserts the marker is stored and that a second call reuses it instead of calling `CreateSBOM` again. before this: 0 stores, 2 creations.
- `TestScanService_GenerateSBOM_DoesNotPersistIncompleteVerdict` asserts a timeout is not written and is retried on the next scan.
- `TestScanService_GenerateSBOM_StoresCreatedSBOMOnce` asserts one store for a newly created SBOM and none for one served from storage. before this: 2, then 3.

the existing table test is unchanged, including `phase 2, store SBOM failed`, which still gets `ReasonSBOMStorageFailed` now that the error comes back through `getOrCreateSBOM`.

## Related issues/PRs:

* Resolved #622
* Follow-up to #610

### DIFF
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -181,8 +181,9 @@ func isRegistryRateLimitedErr(err error) bool {
 		strings.Contains(errStr, "429 Too Many Requests")
 }
 
-// GenerateSBOM implements the "Generate SBOM flow"
-// FIXME check if we still use this method
+// GenerateSBOM implements the "Generate SBOM flow". It is live: cmd/http/main.go routes
+// POST v1/generateSBOM here, through ValidateGenerateSBOM and the same rejection and scan
+// metrics as the other flows.
 func (s *ScanService) GenerateSBOM(ctx context.Context) error {
 	ctx, span := otel.Tracer("").Start(ctx, "ScanService.GenerateSBOM")
 	defer span.End()
@@ -195,27 +196,28 @@ func (s *ScanService) GenerateSBOM(ctx context.Context) error {
 		return domain.ErrCastingWorkload
 	}
 
-	sbom, created, err := s.getOrCreateSBOM(ctx, workload)
+	sbom, storeErr, err := s.getOrCreateSBOM(ctx, workload)
 	if err != nil {
 		return err
 	}
 
 	// do not treat a timed out or too-large SBOM as a successful scan — mirrors the same
-	// check in ScanCP/ScanCVE/ScanRegistry, which this flow had lost track of (see #540)
+	// check in ScanCP/ScanCVE/ScanRegistry, which this flow had lost track of (see #540).
+	// Ahead of storeErr so the size verdict stays the reported reason, as it did when the
+	// store lived below this check.
 	if sbom.Status == helpersv1.Incomplete || sbom.Status == helpersv1.TooLarge {
 		reason := classifySBOMStatusWithAnnotation(sbom.Status, sbom.Annotations)
 		_ = s.platform.ReportScanFailure(ctx, scanfailure.ScanFailureSBOMGeneration, reason, nil)
 		return &domain.ScanError{Reason: reason, Err: domain.ErrIncompleteSBOM}
 	}
 
-	// store SBOM if newly created
-	if s.storage && created {
-		err = s.sbomRepository.StoreSBOM(ctx, sbom, false)
-		if err != nil {
-			_ = s.platform.ReportScanFailure(ctx, scanfailure.ScanFailureBackendPost,
-				scanfailure.ReasonSBOMStorageFailed, err)
-			return &domain.ScanError{Reason: scanfailure.ReasonSBOMStorageFailed, Err: err}
-		}
+	// getOrCreateSBOM already stored what it created, so there is nothing to write here.
+	// Storing is the whole point of this endpoint though, so unlike the scan flows, which
+	// carry on with an SBOM they could not persist, a failure there is reported as one.
+	if storeErr != nil {
+		_ = s.platform.ReportScanFailure(ctx, scanfailure.ScanFailureBackendPost,
+			scanfailure.ReasonSBOMStorageFailed, storeErr)
+		return &domain.ScanError{Reason: scanfailure.ReasonSBOMStorageFailed, Err: storeErr}
 	}
 
 	return nil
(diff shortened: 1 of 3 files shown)


AnshulPatil2005:
rebased on #610, which landed while this was open and moved everything onto getOrCreateSBOM. the diff is entirely different now, and it no longer drags in #619 by mistake. also folded in the codex point about asserting the second call doesn't re-store.

AnshulPatil2005:
both AI flagged the same thing so it now stores TooLarge and not Incomplete. TooLarge is a verdict, it carries the limit it was judged against and getSBOM re-checks it when the limits change. Incomplete is a timeout, transient, nothing to re-evaluate. added TestScanService_GenerateSBOM_DoesNotPersistIncompleteVerdict for the retry.

matthyx (dismissed):
Reviewed the diff plus the real `APIServerStore` implementation (not just the `MemoryStore` test double) to check the core claim in the description.

- Verified `APIServerStore.GetSBOM` always sets `Content: &manifest.Spec.Syft` on a hit, including for a status-only marker stored with a nil `Content` — so `getSBOM`'s TooLarge staleness recheck (annotation vs. current `maxImageSize`/`maxSBOMSize`) does run against a stored marker in production, not just in the `MemoryStore` shim added here. That was the load-bearing assumption behind storing TooLarge and it holds.
- Walked every `CreateSBOM` exit path in the syft adapter that returns a nil error: success (`Learning`, `Content` set), real timeout (`Incomplete`), and both size-limit cases (`TooLarge`). The new store condition `sbom.Content != nil || sbom.Status == helpersv1.TooLarge` covers exactly those three and correctly excludes Incomplete, matching the "timeout is transient, no staleness rule for it" reasoning.
- The `GenerateSBOM` / `getOrCreateSBOM` split (store error vs. creation error) is consistent across all return paths; the error-returning branches of the singleflight worker never hit the `sbomCreation` type assertion since `res.Err` short-circuits first.
- New tests (`RemembersTooLargeVerdict`, `DoesNotPersistIncompleteVerdict`, `StoresCreatedSBOMOnce`) target the three fixed behaviors directly and CI is green.

Remaining CodeRabbit comments are all marked trivial/low-value style nitpicks (return-type ergonomics on `getOrCreateSBOM`, doc-comment placement) — non-blocking. Good to merge.

matthyx:
rebase pls

AnshulPatil2005:
> rebase pls

done

matthyx (approved):
Re-reviewed after the rebase (my prior approval on 544b684 was auto-dismissed by the new commit). Diffed the rebased head (1abc6b3) against the version I reviewed: the PR's own changes are unchanged in substance, only line numbers shifted plus one pre-existing, unrelated line inherited from `main` (the `ran`/singleflight-hit tracking, not part of this PR's diff). All checks are green on the new commit. Still good to merge.

PR #625, opened 2026-08-11
fix(adapters): pin platform in SBOM tests for arch-stable fixtures

### LINKED ISSUE #624: go test ./... fails on arm64 due to host-architecture-dependent platform defaults in SBOM test fixtures
# Description
`go test ./...` fails on an arm64 development machine even though the repository builds and publishes `linux/arm64` images. The in-process Syft adapter (`adapters/v1/syft.go`, `adapters/v1/syft_platform.go`) and the sidecar scanner (`pkg/sbomscanner/v1/server.go`) only request a specific image platform when the caller explicitly sets one; when no platform is requested, the underlying registry provider (stereoscope) falls back to the host's own `runtime.GOARCH`. Several test cases in `adapters/v1/syft_test.go` build `domain.RegistryOptions` without an explicit `Platform`, so on an arm64 host they resolve a different image manifest (and therefore a different SBOM) than the golden fixtures under `adapters/v1/testdata/`, which were captured on an amd64 runner. This causes assertion failures unrelated to any real regression and blocks contributors on arm64 machines from trusting `go test ./...`.

One golden fixture, `adapters/v1/testdata/alpine-sbom.format.json`, also has `source.metadata.os` recorded as an empty string, which does not match the `"os": "linux"` value Syft populates once platform resolution is made deterministic.

# Environment
OS: `darwin arm64` (also applies to any non-amd64 contributor machine)
Version: local checkout at `upstream/main`

# Steps To Reproduce
1. On an arm64 machine, run `go test ./adapters/v1/...`.
2. Observe `jsonassert` failures comparing the resolved SBOM against fixtures such as `testdata/alpine-sbom.format.json`, because the image manifest resolved for the test differs from the one the fixture was captured against.
3. Inspect the test cases in `Test_syftAdapter_CreateSBOM` in `adapters/v1/syft_test.go` and note that several do not set `options.Platform`, so platform resolution silently falls back to the host architecture.

# Expected behavior
`go test ./...` should pass deterministically regardless of the host's CPU architecture. Tests that assert against architecture-specific golden fixtures should pin the platform explicitly rather than relying on the host's default architecture.

# Actual Behavior
On an arm64 host, `Test_syftAdapter_CreateSBOM` fails for image fixtures that don't pin a platform, because the resolved manifest (and therefore the generated SBOM) differs from what was captured on amd64. For example:
```
syft_test.go:215: expected string at '$.source.metadata.os' to be '' but was 'linux'
```

# Additional context
Suspected root cause: test cases in `adapters/v1/syft_test.go` build `domain.RegistryOptions` without an explicit `Platform`, so platform selection defers to the registry provider's default, which resolves to the host's `runtime.GOARCH`. The golden fixtures were captured against `linux/amd64`, so any other host architecture produces a different result. Separately, the `alpine-sbom.format.json` fixture recorded `os: ""` instead of `"linux"`.

Proposed direction: pin `Platform: "linux/amd64"` explicitly in the `domain.RegistryOptions` for the fixture-comparison test cases in `Test_syftAdapter_CreateSBOM`, correct the `os` field in `alpine-sbom.format.json` to match, and add a regression test against a local mock multi-arch registry (serving both `linux/amd64` and `linux/arm64` manifests) so platform resolution is verified without depending on the host architecture or a real registry.

Acceptance criteria:
- `go test ./...` passes on an arm64 development machine.
- Golden fixtures compared in `Test_syftAdapter_CreateSBOM` no longer depend on the host's architecture.
- Test coverage exists demonstrating that platform resolution honors the requested platform independent of host architecture.

rootp1:
## Overview
On an arm64 machine, `go test ./adapters/v1/...` fails even though the code itself is architecture correct. Several cases in `Test_syftAdapter_CreateSBOM` build `domain.RegistryOptions` without an explicit platform, so platform resolution falls back to the host's `runtime.GOARCH` instead of the `linux/amd64` manifest the golden fixtures under `testdata/` were captured against. On arm64 this resolves a different image manifest and produces a different SBOM than the fixture expects, which has nothing to do with a real regression.

This PR pins `Platform: "linux/amd64"` explicitly on the fixture-comparison test cases so resolution is deterministic regardless of host architecture, adds a regression test against a local mock multi-arch registry (serving both amd64 and arm64 manifests) to verify platform selection independent of the host, and corrects `testdata/alpine-sbom.format.json`, whose `source.metadata.os` field was recorded as an empty string instead of `linux`.

Related issues/PRs:
* Resolved #624

## How to Test
```
go test ./adapters/v1/...
go test ./...
make verify
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved container image SBOM platform resolution, ensuring explicitly selected architectures are honored consistently.
  * Corrected image metadata to report the Linux operating system accurately.

* **Tests**
  * Added coverage for multi-architecture images and resolved-platform annotations across different host architectures.
  * Expanded registry image fixtures to validate architecture-specific manifests, layers, and configuration data.

### DIFF
--- a/adapters/v1/syft_test.go
+++ b/adapters/v1/syft_test.go
@@ -17,6 +17,7 @@ import (
 	"net/http/httptest"
 	"net/url"
 	"os"
+	"runtime"
 	"strings"
 	"testing"
 	"time"
@@ -49,25 +50,35 @@ func Test_syftAdapter_CreateSBOM(t *testing.T) {
 		wantStatus        string
 	}{
 		{
+			// Platform is pinned explicitly so this assertion is stable across host
+			// architectures: with no platform requested, stereoscope's registry provider
+			// defaults to runtime.GOARCH (see defaultPlatformIfNil in
+			// github.com/anchore/stereoscope/pkg/image/oci), which would otherwise resolve
+			// a different manifest - and produce a different SBOM than this fixture - on
+			// an arm64 runner than on the amd64 runner the fixture was captured on.
 			name:    "empty image produces empty SBOM",
 			imageID: "library/hello-world@sha256:aa0cc8055b82dc2509bed2e19b275c8f463506616377219d9642221ab53cf9fe",
 			format:  "testdata/hello-world-sbom.format.json",
+			options: domain.RegistryOptions{Platform: "linux/amd64"},
 		},
 		{
 			name:    "schema v1 image produces well-formed SBOM",
 			imageID: "quay.io/jitesoft/debian:stretch-slim",
 			format:  "testdata/stretch-slim-sbom.format.json",
+			options: domain.RegistryOptions{Platform: "linux/amd64"},
 		},
 		{
 			name:    "valid image produces well-formed SBOM",
 			imageID: "library/alpine@sha256:e2e16842c9b54d985bf1ef9242a313f36b856181f188de21313820e177002501",
 			format:  "testdata/alpine-sbom.format.json",
+			options: domain.RegistryOptions{Platform: "linux/amd64"},
 		},
 		{
 			name:    "public image with invalid registry credentials falls back to unauthenticated and produces well-formed SBOM",
 			imageID: "library/alpine@sha256:e2e16842c9b54d985bf1ef9242a313f36b856181f188de21313820e177002501",
 			format:  "testdata/alpine-sbom.format.json",
 			options: domain.RegistryOptions{
+				Platform: "linux/amd64",
 				Credentials: []domain.RegistryCredentials{
 					{
 						Authority: "index.docker.io",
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed and verified locally (checked out `fix/arch-stable-sbom-fixtures`):

- `go build ./...`, `go vet ./...`, and `go test ./...` all pass (full suite, not just the touched package).
- Confirmed the root cause claim directly: the repo's actual `stereoscope` dependency (via the `replace github.com/anchore/stereoscope => github.com/matthyx/stereoscope v0.1.22-ks.1` directive) has `defaultPlatformIfNil` fall back to `runtime.GOARCH` in `pkg/image/oci/registry_provider.go` exactly as described — so the arch-instability the PR fixes is real, not speculative.
- Platform pinning (`Platform: "linux/amd64"`) is applied precisely to the cases that assert against a golden fixture (`format` set); error-path/no-fixture cases were correctly left untouched — minimal, well-scoped change.
- New `Test_syftAdapter_CreateSBOM_MultiArchLocalRegistry` is fully hermetic (local `httptest` server, no network/real registry dependency) and directly exercises platform resolution for both amd64 and arm64 against a local multi-arch manifest list — good regression coverage for the underlying bug, not just the fixture symptom.
- `alpine-sbom.format.json` `os` fix (`""` → `"linux"`) matches the actual current SBOM output verified by the passing test.
- All CI checks are green.

No blockers. LGTM.

PR #626, opened 2026-08-11
fix: give ParseImageManifest the same LayerOrder indexing as parseLayersPayload

### LINKED ISSUE #617: Bug: LayerOrder is computed against two different index spaces, so a vulnerability's layer disagrees with the image manifest's layer list
# Description
`adapters/v1/domain_to_armo.go` computes `LayerOrder` for the same physical image layer two different ways in the same report, and they disagree whenever a Dockerfile has any metadata-only instructions (`ENV`, `LABEL`, `WORKDIR`, `USER`, `CMD`, `EXPOSE`, etc.) mixed in with layer-producing ones (`FROM`, `RUN`, `COPY`, `ADD`) — which is true of almost every real image.

`parseLayersPayload` (line 285) builds its layer list by iterating the image config's `History` and only appending entries where `!EmptyLayer`:

```go
listLayers := make([]containerscan.ESLayer, 0)
for i := range jsonConfig.History {
	if !jsonConfig.History[i].EmptyLayer {
		listLayers = append(listLayers, containerscan.ESLayer{LayerInfo: &containerscan.LayerInfo{...}})
	}
}
for i := 0; i < len(listLayers) && i < len(jsonConfig.RootFS.DiffIDs); i++ {
	listLayers[i].LayerHash = jsonConfig.RootFS.DiffIDs[i].String()
	if i > 0 {
		...
		listLayers[i].LayerInfo.LayerOrder = i
	}
	layerMap[listLayers[i].LayerHash] = listLayers[i]
}
```

`i` here is the index into the **compact** slice (empty layers already skipped), so `LayerOrder` counts only real, layer-producing steps. This map feeds `DomainToArmo` (line 198 onward), which copies `LayerOrder` onto every vulnerability's `Layers[j].LayerOrder` (line 216).

`ParseImageManifest` (line 332), used to build the `ImageManifest` attached to the same report, does the opposite — it loops over the **full**, unfiltered `History` and uses the raw loop index for every entry, empty or not:

```go
layerIndex := 0
for i, historyLayer := range config.History {
	layerInfo := containerscan.ESLayer{
		LayerInfo: &containerscan.LayerInfo{
			...
			LayerOrder:  i,
		},
	}
	if !historyLayer.EmptyLayer && layerIndex < len(rawManifest.Layers) {
		layerInfo.LayerHash = rawManifest.Layers[layerIndex].Digest
		...
		layerIndex++
	}
	imageManifest.Layers = append(imageManifest.Layers, layerInfo)
}
```

Both functions run for the same scan report: `SubmitCVE` in `adapters/v1/backend.go` calls `DomainToArmo` (line 461, populating `vulnerabilities[].Layers[].LayerOrder`) and `ParseImageManifest` (line 459), and `Summarize` (`backend_utils.go:197`) attaches the resulting `imageManifest` to `finalReport.Summary` (line 213 in `backend_utils.go`) right alongside those same `vulnerabilities`. So the outgoing report ends up with two inconsistent numbering schemes for the same layer hash within a single payload.

# Environment
OS: `Any`
Version: `main`

# Steps To Reproduce
1. Take any image whose Dockerfile mixes layer-producing and metadata-only instructions, e.g.:
   ```
   FROM alpine        # layer 0 (EmptyLayer=false)
   ENV FOO=bar         # EmptyLayer=true
   LABEL x=y           # EmptyLayer=true
   COPY app /app        # layer 1 (EmptyLayer=false)
   ```
   `config.History` for this image has 4 entries; only entries 0 and 3 are real layers.
2. In `parseLayersPayload`, the `COPY` layer is the 2nd entry in the compact `listLayers`, so it gets `LayerOrder = 1`.
3. In `ParseImageManifest`, the same `COPY` layer is the 4th entry in the raw `History`, so it gets `LayerOrder = 3`.
4. A vulnerability found in a package installed by that `COPY` layer is reported with `Layers[].LayerOrder = 1`, while `Summary.ImageManifest.Layers[3]` (the actual entry for that same layer hash) carries `LayerOrder = 3`.

# Expected behavior
The same physical layer should carry the same `LayerOrder` value everywhere it appears in a single scan report, so a consumer can reliably look up "which build step introduced this vulnerability" by matching a vulnerability's `Layers[].LayerOrder` against `Summary.ImageManifest.Layers[].LayerOrder`.

# Actual Behavior
`LayerOrder` is computed against two different index spaces — one that skips metadata-only history entries, one that doesn't — so the two numbers only happen to agree for images whose Dockerfile has no metadata-only instructions before the layer in question. For any realistic image, matching a vulnerability to "its" manifest layer by `LayerOrder` silently points at the wrong build step (or, once the offset drifts enough, off the end of the layer list).

# Additional context
`parseLayersPayload` and `ParseImageManifest` are two independent re-implementations of essentially the same "walk image history, build the ordered layer list" logic, using different iteration strategies (compact vs. raw index) — the fix should make both use the same indexing scheme rather than patching the numbers after the fact.

bhuvan-somisetty:
## Overview

`adapters/v1/domain_to_armo.go` has two functions that both compute `LayerOrder` for image layers in the same scan report, using two different index spaces:

- `parseLayersPayload` (feeds `DomainToArmo`, which attaches `LayerOrder` to every vulnerability's `Layers[]`) numbers layers by their position among **real, layer-producing** `History` entries only, skipping metadata-only ones (`ENV`, `LABEL`, `CMD`, etc.).
- `ParseImageManifest` (feeds `Summary.ImageManifest.Layers[]`, attached to the same report by `Summarize`) numbered **every** `History` entry, including the metadata-only ones.

Since almost any real Dockerfile mixes both kinds of instructions, the same physical layer got a different `LayerOrder` in the two places it appears in one report, so correlating "which build step introduced this vulnerability" via `LayerOrder` pointed at the wrong layer.

Fix: `ParseImageManifest` now counts only real layers when assigning `LayerOrder`, the same way `parseLayersPayload` does, so both agree on every shared layer hash.

Also updated `testdata/nginx-image-manifest.json` (a real image history with metadata-only entries interleaved between real layers) to the corrected expected values, and added a test that builds a mixed-history fixture and asserts `ParseImageManifest` and `parseLayersPayload` produce the same `LayerOrder` for every layer hash they share.

## How to Test

```
go test ./adapters/v1/... -run 'TestParseImageManifest|Test_parseLayersPayload|Test_layerOrder_consistentBetweenManifestAndVulnerabilities' -v
```

## Related issues/PRs:

* Resolved #617

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected container image layer ordering when metadata-only history entries are present.
  * Improved consistency between manifest layers and vulnerability layer information.
  * Updated image layer metadata to reflect the corrected ordering.

### DIFF
--- a/adapters/v1/domain_to_armo.go
+++ b/adapters/v1/domain_to_armo.go
@@ -352,13 +352,19 @@ func ParseImageManifest(grypeDocument *v1beta1.GrypeDocument) (*containerscan.Im
 		Layers:       []containerscan.ESLayer{},
 	}
 
+	// LayerOrder counts real, layer-producing history entries only, matching
+	// parseLayersPayload's indexing (which builds the layerMap DomainToArmo attaches to
+	// each vulnerability). History also contains metadata-only entries (ENV, LABEL, CMD,
+	// etc.) that don't produce a layer; counting those too, as the raw loop index would,
+	// gives every real layer a different LayerOrder here than in the vulnerability report
+	// for the same layer hash, breaking any lookup that correlates the two by LayerOrder.
 	layerIndex := 0
-	for i, historyLayer := range config.History {
+	for _, historyLayer := range config.History {
 		layerInfo := containerscan.ESLayer{
 			LayerInfo: &containerscan.LayerInfo{
 				CreatedBy:   historyLayer.CreatedBy,
 				CreatedTime: &historyLayer.Created.Time,
-				LayerOrder:  i,
+				LayerOrder:  layerIndex,
 			},
 		}
 		if !historyLayer.EmptyLayer && layerIndex < len(rawManifest.Layers) {
--- a/adapters/v1/domain_to_armo_test.go
+++ b/adapters/v1/domain_to_armo_test.go
@@ -200,6 +200,62 @@ func Test_parseLayersPayload(t *testing.T) {
 	}
 }
 
+// Test_layerOrder_consistentBetweenManifestAndVulnerabilities guards #617: a vulnerability's
+// LayerOrder (via parseLayersPayload, which DomainToArmo attaches to each vulnerability) and
+// the same layer's LayerOrder in ParseImageManifest's output must agree, so a consumer can
+// correlate a vulnerability to its build step. History mixes metadata-only entries (no layer)
+// between real, layer-producing ones, which is what previously made the two disagree.
+func Test_layerOrder_consistentBetweenManifestAndVulnerabilities(t *testing.T) {
+	config := containerRegistryV1.ConfigFile{
+		History: []containerRegistryV1.History{
+			{CreatedBy: "FROM base", EmptyLayer: false},
+			{CreatedBy: "ENV FOO=bar", EmptyLayer: true},
+			{CreatedBy: "LABEL x=y", EmptyLayer: true},
+			{CreatedBy: "COPY app /app", EmptyLayer: false},
+			{CreatedBy: "CMD [\"/app\"]", EmptyLayer: true},
+		},
+		RootFS: containerRegistryV1.RootFS{
+			DiffIDs: []containerRegistryV1.Hash{
+				{Algorithm: "sha256", Hex: "aaaa000000000000000000000000000000000000000000000000000000000000"},
+				{Algorithm: "sha256", Hex: "bbbb000000000000000000000000000000000000000000000000000000000000"},
+			},
+		},
+	}
+	configBytes, err := json.Marshal(config)
+	assert.NoError(t, err)
+
+	imageMetadata := source.ImageMetadata{
+		RawConfig: configBytes,
+		Layers: []source.LayerMetadata{
+			{Digest: "sha256:aaaa000000000000000000000000000000000000000000000000000000000000", Size: 100},
+			{Digest: "sha256:bbbb000000000000000000000000000000000000000000000000000000000000", Size: 200},
+		},
+	}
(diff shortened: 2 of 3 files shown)


matthyx (approved):
Traced the fix by hand against `parseLayersPayload`: `layerIndex` in `ParseImageManifest` now increments only on `!EmptyLayer` entries, same as the `listLayers` build there, so the LayerOrder assigned to a real layer's `containerscan.ESLayer` is read *before* the increment — giving the first real layer order `0`, matching `parseLayersPayload`'s `i`-based indexing (which is also 0 for the first real layer, via the zero-value default).

Cross-checked the updated `testdata/nginx-image-manifest.json`: 7 real layers now get orders 0–6 (first is implicit 0, `RUN`→1 ... `COPY 30-tune`→6), with interleaved `ENV`/`LABEL`/`CMD` metadata entries correctly inheriting the *next* real layer's pending order rather than their own raw History index. That lines up with what `parseLayersPayload` would produce for the same config.

The new `Test_layerOrder_consistentBetweenManifestAndVulnerabilities` test directly asserts `ParseImageManifest` and `parseLayersPayload` agree on `LayerOrder` per shared layer hash using a mixed-history fixture — good regression coverage for #617.

No blockers. CI (build, lint, CodeQL, security/license checks) is green and CodeRabbit found no actionable comments. Nice, minimal fix — approving.

bhuvan-somisetty:
Thanks @matthyx for the detailed review and merge!

PR #629, opened 2026-08-11
fix(services): key the SBOM singleflight on the image, not the tag

### LINKED ISSUE #628: concurrent scans of one tag at different digests are merged into one SBOM
the singleflight key #610 added does not identify the image:

```go
key := fmt.Sprintf("%s|%s|http:%t|skiptls:%t", rateLimitCacheKey(workload), opts.Platform, opts.InsecureUseHTTP, opts.InsecureSkipTLSVerify)
```

`rateLimitCacheKey` is the image tag plus a credentials hash, and it leaves the digest out on purpose:

```go
// ImageHash is avoided because ScanRegistry payloads never populate it.
```

that is the right call for a rate-limit backoff, since a registry rate limits by repository, but it is the wrong identity for "which SBOM am I about to build". the SBOM is stored under `ImageSlug`, which is `sanitizeImage(tag) + "-" + last 12 of the digest`, so two scans of the same tag at two different digests have different slugs and are different work, but land on the same key.

that happens whenever a tag is mutable and moves: a repush, or a rollout where old and new pods coexist and are scanned together by the worker pool.

the loser of the race is handed the leader's SBOM, named for the leader's image:

```
CreateSBOM calls = 1
workload a (slug repo-app-latest-aaaaaaaaaaaa) got SBOM named "repo-app-latest-bbbbbbbbbbbb"
workload b (slug repo-app-latest-bbbbbbbbbbbb) got SBOM named "repo-app-latest-bbbbbbbbbbbb"
```

and it does not stop at the SBOM. `adapters/v1/grype.go` takes the manifest name straight from it:

```go
return domain.CVEManifest{
	Name:               sbom.Name,
```

so the losing workload's CVE results are built from the other image's SBOM and filed under the other image's slug. its own slug gets nothing, and the vulnerabilities reported against it belong to a different digest. nothing errors.

keying on `ImageSlug` instead is enough. all three validators already require it to be non-empty (`ErrMissingImageInfo`), so it is always available by the time `getOrCreateSBOM` runs, and it is exactly the identity of the object being produced. the credentials part of the key is still worth keeping, so a caller with working credentials does not inherit a failure from one without.

version: main (cb4aa30)

AnshulPatil2005:
## Overview

keys the SBOM singleflight on `ImageSlug` instead of `rateLimitCacheKey`.

`rateLimitCacheKey` is the image tag plus a credentials hash, and it leaves the digest out deliberately, which is correct for a rate-limit backoff because a registry rate limits by repository. it is the wrong identity for deciding which callers are doing the same work. the SBOM is stored under `ImageSlug`, which carries the digest, so two scans of one mutable tag at two different digests are different work but shared a key. that happens on a repush, or during a rollout where old and new pods are scanned together by the worker pool.

the loser was handed the leader's SBOM, named for the leader's image, and `adapters/v1/grype.go` takes `CVEManifest.Name` straight from `sbom.Name`, so its findings were filed under the other image's slug while its own slug got nothing. no error anywhere.

`ImageSlug` is the identity of the object being produced, and all three validators already reject an empty one with `ErrMissingImageInfo`, so it is always populated by the time `getOrCreateSBOM` runs.

## Additional Information

the credentials hash stays in the key, so a caller holding working credentials still does not inherit a failure from one without. it moved into a `credentialsFingerprint` helper shared with `rateLimitCacheKey`, whose output is unchanged, so the rate-limit cache and its tests are unaffected.

the new test runs with storage off. `MemoryStore`'s maps are unguarded, and now that two same-tag callers correctly get separate keys their workers store concurrently, which trips the race detector on the double rather than on anything in the code. worth knowing about, since #610 introduced the first concurrency these doubles have had to deal with, but it is a test-only gap and out of scope here.

## How to Test

`go test ./core/services/ -race -count=1`

`TestScanService_getOrCreateSBOM_SameTagDifferentDigestsAreNotShared` runs two callers for one tag at two digests and asserts each gets the SBOM for its own image, and that both images are built.

without this, it fails in 2s with:

```
expected: "repo-app-latest-aaaaaaaaaaaa"
actual  : "repo-app-latest-bbbbbbbbbbbb"
each workload must get the SBOM for its own image
```

the barrier the test uses to force contention is bounded for that reason: merged onto one key only one caller ever reaches `CreateSBOM`, and an unbounded wait would hang rather than fail.

`TestScanService_SingleflightSBOMDeduplication` from #610 is untouched and still passes, so same-image dedup, which is the point of the singleflight, still works.

## Related issues/PRs:

* Resolved #628
* Follow-up to #610

### DIFF
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -105,8 +105,18 @@ func NewScanService(sbomCreator ports.SBOMCreator, sbomRepository ports.SBOMRepo
 // does not cause a cross-tenant denial-of-service for workloads that use valid imagePullSecrets
 // to bypass the rate limit. ImageHash is avoided because ScanRegistry payloads never populate it.
 func rateLimitCacheKey(workload domain.ScanCommand) string {
-	if len(workload.CredentialsList) == 0 {
+	fingerprint := credentialsFingerprint(workload)
+	if fingerprint == "" {
 		return workload.ImageTagNormalized
+	}
+	return fmt.Sprintf("%s|%s", workload.ImageTagNormalized, fingerprint)
+}
+
+// credentialsFingerprint hashes the workload's credentials, so that keys built from it keep
+// callers holding different credentials apart. Empty when the workload carries none.
+func credentialsFingerprint(workload domain.ScanCommand) string {
+	if len(workload.CredentialsList) == 0 {
+		return ""
 	}
 	h := sha256.New()
 	for _, cred := range workload.CredentialsList {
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed the diff and the surrounding code path.

- `ImageSlug` (carries the digest) is the right identity for the SBOM singleflight key, since `rateLimitCacheKey`/tag-based keying was conflating two different digests behind one mutable tag (repush / rollout race) — matches the `adapters/v1/grype.go` `CVEManifest.Name := sbom.Name` path described in the PR body.
- Confirmed `ImageSlug` is always populated by the time `getOrCreateSBOM` runs: all three request-validators (`ValidateGenerateSBOM`, `ValidateScanCVE`, `ValidateScanRegistry`) reject an empty slug before their flow is ever enqueued (`controllers/http.go`), and `ScanCP` builds/validates the slug itself and `continue`s past any image it fails to build one for.
- `credentialsFingerprint` extraction is a pure refactor — `rateLimitCacheKey`'s output is unchanged for both the empty- and populated-credentials cases, so the rate-limit cache semantics are preserved.
- New regression test (`TestScanService_getOrCreateSBOM_SameTagDifferentDigestsAreNotShared`) exercises real contention via a bounded rendezvous barrier rather than sleeps, and correctly runs with storage off to avoid tripping the race detector on `MemoryStore`'s unrelated unguarded maps — that gap is called out as pre-existing and out of scope, which I agree with. `TestScanService_SingleflightSBOMDeduplication` (same-image dedup) is untouched.
- All CI checks (build, lint, tests, CodeQL, GitGuardian, DCO) are green.

No blockers found. LGTM — approving.

PR #631, opened 2026-08-11
fix(services): count only the scans singleflight spared, not the leader

### LINKED ISSUE #630: singleflight hits counts the request that did the work as deduplicated
`kubevuln_singleflight_hits_total` counts the caller that did the work as one of the callers that were spared it.

it is described as "Total number of times a concurrent scan request was deduplicated by singleflight", and recorded on `res.Shared`:

```go
if res.Shared {
	metrics.RecordSingleflightHit(ctx, "sbom_generation")
}
```

`Shared` does not mean "you were deduplicated". from `golang.org/x/sync/singleflight`, it is `c.dups > 0` evaluated once and sent to every waiting channel:

```go
ch <- Result{c.val, c.err, c.dups > 0}
```

so it is true for the leader too. five concurrent scans of one image collapse to one `CreateSBOM` and four spared requests, but report five. for the common case of two racing requests it reports two where one was deduplicated, so the number is double.

with a counter on the creator and the metric scraped, on main:

```
CreateSBOM calls = 1
kubevuln_singleflight_hits_total{target="sbom_generation"} 5
```

the caller that ran the work is the one that can tell the difference, and it is the only one whose closure singleflight actually executes, so a flag set inside the closure separates them cleanly.

also, the metric is not in the table in `docs/API.md`, which lists the other ten.

version: main (cb4aa30), follow-up to #610

AnshulPatil2005:
## Overview

`kubevuln_singleflight_hits_total` was recorded on `res.Shared`, which is not "you were deduplicated". in `golang.org/x/sync/singleflight` it is `c.dups > 0`, evaluated once and sent to every waiting channel, so the leader sees it as well. five concurrent scans of one image collapse to one `CreateSBOM` and four spared requests, and the counter said five. for two racing requests, which is the ordinary case, it reported two where one was deduplicated.

the leader is the only caller whose closure singleflight actually runs, so a flag set inside it separates the two cleanly, and `res.Shared` is not needed at all.

also adds the metric to the table in `docs/API.md`, which lists the other ten.

## Additional Information

`ran` is a plain bool. each caller passes its own closure to `DoChan` and singleflight runs exactly one of them, so a caller either owns the write or there is no write at all, and it is only read in the `res` branch where the receive happens-after the closure returns. same reasoning as `#619`, which fixes the equivalent confusion in the registry auth cache, where every coalesced waiter was recorded as an upstream fetch. the two are independent, no shared code.

## How to Test

`go test ./core/services/ -race -count=1`

`TestScanService_SingleflightHitsCountOnlyDeduplicatedCallers` runs five callers against one image on the blocking creator #610 already uses, then asserts the hits equal the callers that did not run the work.

it asserts `callers - CreateSBOM calls` rather than a fixed four, so it holds even if one caller arrives late enough to start a round of its own, and it runs with storage off so that a late caller reaches `CreateSBOM` instead of being served by the worker's storage re-check, which would break that identity.

against main it fails with the number from the issue:

```
kubevuln_singleflight_hits_total{target="sbom_generation"} 5
  does not contain
kubevuln_singleflight_hits_total{target="sbom_generation"} 4
```

`TestScanService_SingleflightSBOMDeduplication` from #610 is untouched and still passes.

## Related issues/PRs:

* Resolved #630
* Follow-up to #610

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected SBOM-generation metrics so only requests that reuse work already in progress are counted as singleflight hits.
  * Improved metric accuracy for concurrent and late-arriving SBOM requests.

* **Documentation**
  * Documented the `kubevuln_singleflight_hits_total` Prometheus metric, including its target label and purpose for monitoring deduplicated SBOM-generation requests.

### DIFF
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -1304,7 +1304,14 @@ func (s *ScanService) getOrCreateSBOM(ctx context.Context, workload domain.ScanC
 	// Detach workerCtx from caller cancellation so the singleflight worker job completes for all waiters even if the leader cancels early
 	workerCtx := context.WithoutCancel(ctx)
 
+	// ran reports whether this caller's own closure was the one singleflight executed.
+	// singleflight's res.Shared cannot answer that: it means the result was handed to more
+	// than one caller, so it is true for the leader too, and counting on it would report the
+	// caller that did the work as one of the callers that were spared it. Only read in the
+	// res branch below, where receiving happens-after the closure returns.
+	ran := false
 	ch := s.sfGroup.DoChan(key, func() (interface{}, error) {
+		ran = true
 		// Re-check storage inside singleflight worker in case another worker stored it while we waited
 		if s.storage {
 			if sbom, err := s.getSBOM(workerCtx, workload.ImageSlug, s.sbomCreator.Version()); err == nil && sbom.Content != nil {
@@ -1355,7 +1362,7 @@ func (s *ScanService) getOrCreateSBOM(ctx context.Context, workload domain.ScanC
 		if res.Err != nil {
 			return domain.SBOM{}, false, res.Err
 		}
-		if res.Shared {
+		if !ran {
 			metrics.RecordSingleflightHit(ctx, "sbom_generation")
 		}
 		return res.Val.(domain.SBOM), !res.Shared, nil
--- a/core/services/scan_test.go
+++ b/core/services/scan_test.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"net/http/httptest"
 	"os"
 	"sync"
 	"testing"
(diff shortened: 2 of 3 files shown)


matthyx (approved):
Verified the core claim: `DoChan`'s `fn` only runs for the caller that creates the singleflight `call` (the leader); duplicate callers' closures are discarded entirely, so `ran` stays `false` for every caller whose own closure never executed and `true` only for the one that ran the work. The write happens inside the closure, which always completes and sends on the channel before any receiver's `case res := <-ch:` fires (channel send happens-before receive), so reading `ran` there is race-free even though `DoChan` runs the leader's closure on a separate goroutine from the leader's own.

The new test's `callers - CreateSBOM calls` assertion is a nice touch — it stays correct even if a caller arrives late enough to miss the round and starts its own (verified the math: hits summed over any number of rounds always equals `callers - rounds`, and `rounds == CreateSBOM calls`). No blockers. LGTM.

PR #595, opened 2026-08-11
fix: make updateVEX author-aware to prevent overwriting external VEX

h30s:
## Overview
This PR addresses the issue where `updateVEX` unconditionally reset the status and justification of every single statement in an `OpenVulnerabilityExchangeContainer` during a scan update. This previous behavior would corrupt any externally ingested VEX documents (e.g., from Red Hat or Chainguard) on the very next scan since it blindly assumed all statements were created by the local scanner.

**Fix**: Modified `createVEX` and `updateVEX` to be author-aware. Statements generated by the local scanner are now assigned an explicitly namespaced ID (`https://kubescape.io/vex/statement/...`). The loops in `updateVEX` now check for the presence of an external ID before applying resets or modifications, leaving externally injected assertions fully intact.

## Additional Information
Legacy statements that do not have an ID assigned (created prior to this change) continue to be treated as locally-generated statements, ensuring backward compatibility.

## How to Test
1. Run local package tests for the repositories package: `go test -v ./repositories`
2. Create an `OpenVulnerabilityExchangeContainer` and manually inject a statement that does not carry the `https://kubescape.io/` prefix in its ID.
3. Run a new scan targeting the same container.
4. Verify that the manually injected statement is completely unaltered, while other local statements are successfully updated according to the new scanner output.

## Examples/Screenshots
N/A

## Related issues/PRs:
* Resolved #594

## Checklist before requesting a review
- [x] My code follows the style guidelines of this project
- [x] I have commented on my code, particularly in hard-to-understand areas
- [x] I have performed a self-review of my code
- [x] If it is a core feature, I have added thorough tests.
- [x] New and existing unit tests pass locally with my changes

## Summary by CodeRabbit

* **Bug Fixes**
  * VEX statements generated by scanning now use consistent identifiers tied to vulnerability and artifact details.
  * Externally authored VEX statements and their metadata are preserved during status updates, ignores, normalization, and resets.
  * Legacy local statements continue to be reset with the expected status and justification.

PR #633, opened 2026-08-11
fix(repositories): do not let an external VEX statement suppress our own

### LINKED ISSUE #632: an external VEX statement suppresses kubescape's own assessment of that vulnerability
#595 made `updateVEX` leave externally authored statements alone, and the reset loop says so plainly:

```go
// Only reset statements generated by the local scanner.
// External statements with their own ID are left intact.
if !isLocalStatement(vexDoc.Statements[i].ID) {
	continue
}
```

the guard went on the three places that modify a statement, but not on the two that decide whether to append one:

```go
for _, v := range cve.Content.Matches {
	found := false
	for _, s := range vexDoc.Statements {
		if s.Vulnerability.Name != v.Vulnerability.ID {
			continue
		}
		...
		if v.Artifact.PURL == s.Products[0].Subcomponents[0].ID {
			found = true
```

so an external statement counts as "we already have this one". if an external feed has a statement for a vulnerability and package that kubevuln also found, kubevuln does not add its own statement for it.

that is more than not overwriting. every step that maintains an assessment is local-only, so with no local statement to work on:

- `markRelevantVulnerabilitiesAsAffectedInVex` skips it, so it is never marked affected even when relevancy says the package is loaded
- `markIgnoredVulnerabilitiesInVex` skips it, so a SecurityException on it is not reflected
- the reset loop skips it

the vulnerability ends up in the document with only the external author's view and no kubescape assessment at all, which is the opposite of what a consumer merging the two would expect. one external statement silently removes a finding from kubescape's output.

reproducing it on `testdata/nginx-cve.json`: run `StoreVEX`, take one statement kubevuln generated, replace it with an external statement covering the same vulnerability and package, run `StoreVEX` again. the external statement survives, and no local statement for that pair comes back.

guarding the two dedup loops the same way as the other three sites is enough. it also makes the `Products[0].Subcomponents[0]` indexing right rather than lucky: statements kubevuln builds always have exactly one product and one subcomponent, while an external statement can list many and would be compared on its first only.

version: main (cb4aa30), follow-up to #595

AnshulPatil2005:
## Overview

#595 put `isLocalStatement` on the three places in `updateVEX` that modify a statement, but not on the two that decide whether to append one. so an external statement counted as "we already have this", and kubevuln skipped adding its own statement for a vulnerability an external feed happened to cover.

that goes further than not overwriting. every step that maintains an assessment is local-only, so with no local statement to work on, `markRelevantVulnerabilitiesAsAffectedInVex` never marks it affected even when relevancy says the package is loaded, `markIgnoredVulnerabilitiesInVex` never reflects a SecurityException on it, and the reset loop skips it. the vulnerability ends up in the document with the external author's view and no kubescape assessment at all, so one external statement silently drops a finding from kubescape's output.

this puts the same guard on both dedup loops, so kubevuln records its assessment alongside the external one instead of in place of it, and the external statement is still left exactly as it was.

## Additional Information

it also makes the `Products[0].Subcomponents[0]` comparison right rather than lucky. statements kubevuln builds always carry exactly one product and one subcomponent, so index 0 is the whole statement. an external one can list many, and was being compared on its first subcomponent only.

no change when there is no external VEX, which is why the existing tests are untouched.

## How to Test

`TestAPIServerStore_updateVEX_externalStatementDoesNotSuppressLocalOne` runs `StoreVEX` on the nginx testdata, takes a statement kubevuln generated so the vulnerability and package are ones the manifest really contains, swaps it for an external statement covering the same pair, and runs `StoreVEX` again. it asserts both statements are present afterwards and that the external one is unmodified.

before this it fails on the local one:

```
Should be true
kubescape must still record its own assessment alongside the external one
```

`TestAPIServerStore_updateVEX_preservesExternalStatements` from #595 still passes. its external statement is for a vulnerability the manifest does not contain, so it never reaches the dedup.

## Related issues/PRs:

* Resolved #632
* Follow-up to #595

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -1177,6 +1177,14 @@ func (a *APIServerStore) updateVEX(ctx context.Context, cve domain.CVEManifest, 
 		for _, v := range cve.Content.Matches {
 			found := false
 			for _, s := range vexDoc.Statements {
+				// Only our own statements count as already present. An external one is
+				// another author's assessment, and every step below that maintains a
+				// statement (mark-affected, mark-ignored, reset-to-baseline) is local-only,
+				// so treating it as ours would drop kubescape's assessment of this
+				// vulnerability from the document instead of recording it alongside theirs.
+				if !isLocalStatement(s.ID) {
+					continue
+				}
 				if s.Vulnerability.Name != v.Vulnerability.ID {
 					continue
 				}
@@ -1223,6 +1231,14 @@ func (a *APIServerStore) updateVEX(ctx context.Context, cve domain.CVEManifest, 
 		for _, v := range cve.Content.IgnoredMatches {
 			found := false
 			for _, s := range vexDoc.Statements {
+				// Only our own statements count as already present. An external one is
+				// another author's assessment, and every step below that maintains a
+				// statement (mark-affected, mark-ignored, reset-to-baseline) is local-only,
+				// so treating it as ours would drop kubescape's assessment of this
+				// vulnerability from the document instead of recording it alongside theirs.
+				if !isLocalStatement(s.ID) {
+					continue
+				}
 				if s.Vulnerability.Name != v.Vulnerability.ID {
 					continue
 				}
--- a/repositories/apiserver_external_test.go
+++ b/repositories/apiserver_external_test.go
@@ -104,3 +104,77 @@ func TestAPIServerStore_updateVEX_preservesExternalStatements(t *testing.T) {
 	}
 	assert.True(t, foundLegacy, "Legacy statement should remain in the document")
 }
+
+// #595 made updateVEX leave external statements alone, but the dedup that decides whether
+// to append kubevuln's own statement still scans every statement rather than only the local
+// ones. An external statement covering a vulnerability kubevuln also found therefore
+// suppresses kubevuln's statement for it entirely, and since every assessment step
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Reviewed the diff and verified locally (cloned the branch, ran the full `repositories` test suite, and gofmt/go vet).

- The two new `isLocalStatement` guards in `updateVEX` are consistent with the same guard already used in `markRelevantVulnerabilitiesAsAffectedInVex`, `markIgnoredVulnerabilitiesInVex`, and the reset-to-baseline loop — every place that walks `vexDoc.Statements` now agrees on what counts as "already present."
- I reverted just the two guard hunks and reran `TestAPIServerStore_updateVEX_externalStatementDoesNotSuppressLocalOne`: it fails with the exact error shown in the PR description ("kubescape must still record its own assessment alongside the external one"), confirming the new test actually catches the regression. With the fix restored, the full `repositories` package passes (`go test ./repositories/... -count=1`).
- The `Products[0].Subcomponents[0]` comparison is safe for local statements specifically because `createProductStructForImageAndPackage` always builds exactly one product with exactly one subcomponent — so scoping the comparison to local-only statements (as this PR does) makes that indexing correct rather than incidental, matching the PR description.
- `TestAPIServerStore_updateVEX_preservesExternalStatements` (from #595) still passes, and CI is green across build/lint/CodeQL/tests.

No blockers. Good to merge.

PR #634, opened 2026-08-11
fix: bound the sbom-scanner sidecar's graceful shutdown with a timeout

### LINKED ISSUE #627: Bug: sbom-scanner sidecar's GracefulStop is unbounded, so shutdown time is dictated by the in-flight RPC's own timeout instead of a fixed drain deadline
# Description
`cmd/sbom-scanner/main.go` (the `sbom-scanner` sidecar's gRPC server) shuts down with a plain, unbounded `srv.GracefulStop()`:

```go
go func() {
	sig := <-sigCh
	logger.L().Info("received signal, shutting down", helpers.String("signal", sig.String()))
	srv.GracefulStop()
	os.Remove(socketPath)
}()
```

`GracefulStop()` stops accepting new RPCs but blocks until every currently-running RPC handler returns on its own. `CreateSBOM` (`pkg/sbomscanner/v1/server.go:174`) is exactly the kind of handler that can run for a long time: it wraps the actual scan in `deadline.New(timeout)` where `timeout` comes straight from the caller's request (`server.go:297`, `timeout := time.Duration(req.TimeoutSeconds) * time.Second`) and falls back to 5 minutes if unset (`server.go:298-300`). That per-call deadline has no idea the process is shutting down — it isn't wired to `sigCh` or to gRPC's own shutdown machinery, so it keeps running to its own natural completion or its own timeout regardless of SIGTERM.

This is the opposite of how the main `kubevuln` process (a separate binary, `cmd/http/main.go`) handles the same situation: its shutdown sequence deliberately bounds how long it waits for in-flight work via `controller.Shutdown(c.ShutdownTimeout)` (`controllers/http.go:394-412`), abandoning anything still running past that timeout so the process can still exit before Kubernetes sends SIGKILL — a design explicitly called out in that function's comments ("timeout is the backstop for a running task that ignores ctx cancellation... and would otherwise block Stop() indefinitely"). The sidecar has no equivalent bound at all, and no config knob to add one — `cmd/sbom-scanner/main.go` doesn't read `config.LoadConfig` or any environment variable besides `SOCKET_PATH`.

# Environment
OS: `Any`
Version: `main`

# Steps To Reproduce
1. Start the `sbom-scanner` sidecar and issue a `CreateSBOM` RPC for a large/slow image, with a caller-supplied `TimeoutSeconds` of, say, 300 (5 minutes) — or omit it, which also defaults to 5 minutes server-side.
2. While that RPC is still in flight, send SIGTERM to the sidecar process (e.g. `kubectl delete pod`, or a rolling deployment triggering the pod's normal termination).
3. Observe `srv.GracefulStop()` block: it does not return until the in-flight `CreateSBOM` call finishes on its own — up to the full 5 minutes (or whatever `TimeoutSeconds` was requested), not bounded by anything shutdown-related.
4. If the pod's `terminationGracePeriodSeconds` is shorter than that remaining time (very plausible: the sidecar's own default scan timeout is 5 minutes, a common `terminationGracePeriodSeconds` default is 30s), the kubelet SIGKILLs the container mid-`GracefulStop()`, and the graceful path never got a chance to log or clean up (`os.Remove(socketPath)` on line 52 never runs).

# Expected behavior
Sidecar shutdown should be bounded the same way the main process's is: wait for in-flight RPCs up to a fixed timeout, then stop waiting (`srv.Stop()` instead of continuing to block on `GracefulStop()`) so the process can exit promptly and predictably regardless of how long the in-flight scan's own deadline is.

# Actual Behavior
Shutdown time for the sidecar is entirely dictated by whatever `TimeoutSeconds` happened to be requested on the in-flight RPC, with no upper bound and no way to configure one. A caller-controlled per-request value (which callers can set arbitrarily via `req.TimeoutSeconds`, and which server-side defaults to 5 minutes when unset) ends up determining how long a pod takes to terminate.

# Additional context
This is the sidecar-process counterpart to #467 (which bounded the *main* process's own worker-pool drain) — that fix never touched `cmd/sbom-scanner/main.go`, since it's a completely separate binary/process with its own signal handling.

bhuvan-somisetty:
## Overview

`cmd/sbom-scanner/main.go`'s SIGTERM/SIGINT handler called plain `srv.GracefulStop()`, which blocks until every in-flight RPC handler returns on its own. `CreateSBOM` (`pkg/sbomscanner/v1/server.go`) runs under a deadline taken straight from the caller's request (`req.TimeoutSeconds`), defaulting to 5 minutes when unset — that deadline has no idea the process is shutting down, so it keeps running regardless of the signal. Shutdown time ended up dictated entirely by whatever timeout the in-flight scan happened to carry, with no upper bound and no config knob to add one.

This is the sidecar-process counterpart to #467, which bounded the *main* `kubevuln` process's own worker-pool drain — that fix never touched this separate binary.

Fix: added `gracefulStopWithTimeout`, which waits for `GracefulStop` up to a fixed timeout (20s, matching `cmd/http/main.go`'s `shutdownTimeout` default) and falls back to `srv.Stop()` — forcibly closing any still-running RPCs — if the timeout elapses first, so the sidecar exits promptly and predictably instead of hanging on an arbitrary caller-supplied value.

## How to Test

```
go test ./cmd/sbom-scanner/... -v
```

`TestGracefulStopWithTimeout_ForcesStopWhenRPCOutlivesTimeout` starts a real gRPC server with a handler that blocks indefinitely, confirms the call returns near the configured timeout (not once the handler finishes), and confirms the in-flight RPC does return an error afterward (proving `Stop` actually ran). `TestGracefulStopWithTimeout_ReturnsPromptlyWithNoInFlightRPCs` confirms the fast path with nothing in flight doesn't wait for the full timeout.

## Related issues/PRs:

* Resolved #627

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SBOM scanner shutdown handling to allow active operations to finish gracefully.
  * Added a 20-second shutdown limit to prevent the scanner from hanging indefinitely.
  * Ensured prompt shutdown when no operations are in progress.

### DIFF
--- a/cmd/sbom-scanner/main.go
+++ b/cmd/sbom-scanner/main.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"os/signal"
 	"syscall"
+	"time"
 
 	"github.com/kubescape/go-logger"
 	"github.com/kubescape/go-logger/helpers"
@@ -21,6 +22,38 @@ import (
 	// registered". See https://github.com/kubescape/kubevuln/issues/378.
 	_ "modernc.org/sqlite"
 )
+
+// shutdownTimeout bounds how long gracefulStopWithTimeout waits for in-flight CreateSBOM RPCs
+// to finish on their own before forcing the connection closed. Matches cmd/http/main.go's
+// shutdownTimeout default so both processes give a comparable grace period before a
+// terminationGracePeriodSeconds-driven SIGKILL would otherwise cut them off mid-shutdown.
+const shutdownTimeout = 20 * time.Second
+
+// gracefulStopWithTimeout waits up to timeout for srv's in-flight RPCs to finish via
+// GracefulStop, then force-closes the server with Stop if the timeout elapses first.
+//
+// GracefulStop alone blocks until every in-flight RPC handler returns on its own. For
+// CreateSBOM that can take as long as the caller-supplied TimeoutSeconds (5 minutes by
+// default, see pkg/sbomscanner/v1/server.go) -- a per-call deadline with no awareness that
+// the process is shutting down. Without this bound, shutdown time is dictated entirely by
+// whatever timeout the in-flight request happened to carry, not by anything under this
+// process's control. See #627.
+func gracefulStopWithTimeout(srv *grpc.Server, timeout time.Duration) {
+	stopped := make(chan struct{})
+	go func() {
+		srv.GracefulStop()
+		close(stopped)
+	}()
+
+	select {
+	case <-stopped:
+		logger.L().Info("gRPC server stopped gracefully")
+	case <-time.After(timeout):
+		logger.L().Warning("graceful shutdown timeout reached, forcing stop",
+			helpers.String("timeout", timeout.String()))
+		srv.Stop()
+	}
+}
 
 func main() {
 	socketPath := os.Getenv("SOCKET_PATH")
@@ -48,7 +81,7 @@ func main() {
 	go func() {
 		sig := <-sigCh
 		logger.L().Info("received signal, shutting down", helpers.String("signal", sig.String()))
-		srv.GracefulStop()
+		gracefulStopWithTimeout(srv, shutdownTimeout)
 		os.Remove(socketPath)
 	}()
 
--- a/cmd/sbom-scanner/shutdown_test.go
+++ b/cmd/sbom-scanner/shutdown_test.go
@@ -1,0 +1,105 @@
+package main
+
+import (
+	"context"
+	"net"
+	"testing"
+	"time"
+
+	sbomscanner "github.com/kubescape/kubevuln/pkg/sbomscanner/v1"
+	pb "github.com/kubescape/kubevuln/pkg/sbomscanner/v1/proto"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+// blockingScanner is a minimal SBOMScannerServer whose Health call blocks until release is
+// closed, standing in for a CreateSBOM call that outlives shutdown.
+type blockingScanner struct {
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Reviewed the fix and the referenced code paths (`pkg/sbomscanner/v1/server.go`'s `CreateSBOM` deadline handling, `cmd/http/main.go` + `config/config.go`'s `shutdownTimeout` default). No blockers:

- The 5-minute default/caller-controlled `CreateSBOM` deadline claim checks out (`server.go`: `timeout := time.Duration(req.TimeoutSeconds) * time.Second; if timeout <= 0 { timeout = 5 * time.Minute }`), and matches issue #627's repro.
- The 20s constant matches `config.ShutdownTimeout`'s actual default (`config_test.go: TestLoadConfigShutdownTimeoutDefault` asserts `20*time.Second`), so the "parity with the main process" claim is accurate. Hardcoding it here instead of loading config is reasonable — this binary imports nothing from `config` today, and wiring that in would be a bigger change than this bug fix warrants.
- The bounded-then-force-stop pattern (`GracefulStop` in a goroutine, `select` against `time.After`, fall back to `Stop()`) is safe: `Stop()` unblocks the in-progress `GracefulStop()` goroutine rather than racing it, so there's no goroutine leak.
- Tests cover both the force-stop-on-timeout path and the fast/no-RPC path, and CI (build, lint, tests, CodeQL) is green.

One non-blocking nit from CodeRabbit worth a look in a follow-up: the test suite doesn't have a case asserting an in-flight RPC that finishes *within* the timeout is actually allowed to complete via the graceful path (as opposed to just "nothing in flight completes fast" or "still in flight after timeout gets force-stopped"). Not a correctness issue with the shipped code, just a coverage gap.

Approving.

PR #636, opened 2026-08-12
fix: add scan status endpoint

### LINKED ISSUE #635: Add async scan status endpoint keyed by jobID
## Summary
The async scan endpoints return `200 OK` as soon as a request is accepted, but kubevuln does not expose any first class way to look up a submitted `jobID` and determine whether that scan is still queued, actively running, completed successfully, failed, or was abandoned during shutdown.

## Affected Area
`controllers/http.go`
`core/services/scan.go`
HTTP API surface under `cmd/http/main.go`
`docs/API.md`

## Steps to Reproduce
1. Start kubevuln and submit a scan request to `/v1/sbomCreation`, `/v1/scanImage`, `/v1/scanCP`, or `/v1/scanRegistry` with a known `jobID`.
2. Observe that the endpoint returns `200 OK` immediately because the work is queued onto the controller worker pool.
3. Try to determine from the API whether that specific `jobID` is still queued, running, completed, failed, or abandoned during shutdown.

## Expected Behaviour
A caller should be able to query a submitted `jobID` and receive a stable status document that shows the lifecycle state, timestamps, and any machine readable terminal reason.

## Actual Behaviour
There is no read API for per job scan state. The current implementation only logs scan progress and exposes aggregate metrics, so operators cannot distinguish request acceptance from eventual completion for a specific `jobID`.

## Technical Evidence
`docs/API.md` documents that scan endpoints return immediately and are processed asynchronously in a worker pool.

`controllers/http.go` acknowledges accepted work before `workerPool.Submit(...)` finishes and currently has no store or lookup handler for `jobID` state.

`controllers/http.go` shutdown handling explicitly abandons queued but not started scans, but there is no way for a caller to observe that outcome for a specific request.

`core/services/scan.go` logs `jobID` values on scan start and completion paths, which confirms the identifier is already propagated through the scan lifecycle, but those transitions are not exposed through an API.

## Suspected Root Cause
The async execution model has no persistent lifecycle record attached to `jobID`, so acceptance, execution, terminal outcome, and shutdown abandonment are only visible indirectly through logs and metrics.

## Proposed Direction
Add a lightweight status store keyed by `jobID`, update it from the HTTP controller and scan service as scans move through queued, running, succeeded, failed, and abandoned states, and expose a read only endpoint such as `/v1/scanStatus/:jobID`.

## Acceptance Criteria
- A caller can query a submitted `jobID` and receive queued, running, succeeded, or failed state.
- Terminal states include a machine readable reason.
- Status lookup works for at least one completed and one failed scan in tests.

rootp1:
## Overview
This adds a read only scan status API keyed by `jobID` so callers can tell whether an async scan is queued, running, succeeded, failed, or abandoned during shutdown.

The controller now keeps an in memory status record for accepted jobs, updates it through worker pool lifecycle transitions, and exposes `GET /v1/scanStatus/:jobID`. The scan service also publishes coarse phase updates such as `sbom_generation`, `cve_matching`, `result_storage`, and `result_upload` so the status payload shows current execution progress instead of only terminal state.

Focused controller coverage was added for a completed scan, a failed scan with a machine readable reason, and a queued scan abandoned during shutdown. API documentation was updated for the new endpoint.

## Additional Information
This PR is scoped to the async scan status feature only. Unrelated local files in the worktree were left untouched.

## How to Test
- `go test ./controllers -run 'TestHTTPController_ScanStatus_|TestHTTPController_MetricsEndpoint_RecordsScanFailureReason|TestHTTPController_GenerateSBOM|TestHTTPController_ScanCVE|TestHTTPController_Shutdown_BoundedByTimeout'`
- `go test ./cmd/http ./controllers ./core/services`
- `go vet ./...`
- `make lint`
- `make verify`  
  This currently fails in pre existing `adapters/v1` golden tests on this arm64 environment because those tests expect amd64 image metadata and digests. The changed areas in this PR pass independently.

## Related issues/PRs:
* Resolved #635

- [ ] I have commented on my code, particularly in hard-to-understand areas

## Summary by CodeRabbit

- **New Features**
  - Added an endpoint to check vulnerability scan status by job ID.
  - Scan statuses now include queued, running, succeeded, failed, and abandoned states.
  - Added progress phases, timestamps, and failure or shutdown reasons.
  - Scan results now indicate successful completion even when partial results are available.
  - Added diagnostics visibility for scan operations.

- **Documentation**
  - Documented the scan-status endpoint, response fields, lifecycle states, and example response.

### DIFF
--- a/controllers/http.go
+++ b/controllers/http.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"net/http"
 	"strconv"
+	"sync"
 	"time"
 
 	wssc "github.com/armosec/armoapi-go/apis"
@@ -27,11 +28,12 @@ import (
 // HTTPController maps ScanService ports to gin handlers that can be mapped to paths and methods
 // this mapping is usually done in main()
 type HTTPController struct {
-	scanService ports.ScanService
-	workerPool  *workerpool.WorkerPool
-	metrics     *metrics.Metrics
-	diagnostics func(ctx context.Context) domain.Diagnostics
-	statuses    *scanStatusStore
+	scanService  ports.ScanService
+	workerPool   *workerpool.WorkerPool
+	metrics      *metrics.Metrics
+	diagnostics  func(ctx context.Context) domain.Diagnostics
+	statuses     *scanStatusStore
+	statusesOnce sync.Once
 }
 
 // NewHTTPController initializes the HTTPController struct with the injected scanService
@@ -74,10 +76,19 @@ func (h *HTTPController) WithDiagnostics(f func(ctx context.Context) domain.Diag
 }
 
 func (h *HTTPController) ensureStatuses() *scanStatusStore {
-	if h.statuses == nil {
-		h.statuses = newScanStatusStore()
-	}
+	h.statusesOnce.Do(func() {
+		if h.statuses == nil {
+			h.statuses = newScanStatusStore()
+		}
+	})
 	return h.statuses
+}
+
+func (h *HTTPController) claimTrackedJob(jobID string) bool {
+	if jobID == "" {
+		return true
+	}
+	return h.ensureStatuses().markRunning(jobID)
 }
 
 // recordScan records the outcome and duration of a background scan job for the given endpoint.
@@ -167,7 +178,7 @@ func (h *HTTPController) GenerateSBOM(c *gin.Context) {
 		h.ensureStatuses().markPhase(newScan.JobID, phase)
 	}))
 	h.workerPool.Submit(func() {
-		if !h.ensureStatuses().markRunning(newScan.JobID) {
+		if !h.claimTrackedJob(newScan.JobID) {
 			return
 		}
 		start := time.Now()
@@ -263,7 +274,7 @@ func (h *HTTPController) ScanCP(c *gin.Context) {
 		h.ensureStatuses().markPhase(newScan.JobID, phase)
 	}))
 	h.workerPool.Submit(func() {
-		if !h.ensureStatuses().markRunning(newScan.JobID) {
+		if !h.claimTrackedJob(newScan.JobID) {
 			return
 		}
 		start := time.Now()
@@ -325,7 +336,7 @@ func (h *HTTPController) ScanCVE(c *gin.Context) {
 		h.ensureStatuses().markPhase(newScan.JobID, phase)
 	}))
 	h.workerPool.Submit(func() {
-		if !h.ensureStatuses().markRunning(newScan.JobID) {
+		if !h.claimTrackedJob(newScan.JobID) {
 			return
 		}
 		start := time.Now()
@@ -423,7 +434,7 @@ func (h *HTTPController) ScanRegistry(c *gin.Context) {
 		h.ensureStatuses().markPhase(newScan.JobID, phase)
 	}))
 	h.workerPool.Submit(func() {
-		if !h.ensureStatuses().markRunning(newScan.JobID) {
+		if !h.claimTrackedJob(newScan.JobID) {
 			return
 		}
 		start := time.Now()
--- a/controllers/http.go
+++ b/controllers/http.go
@@ -167,7 +167,9 @@ func (h *HTTPController) GenerateSBOM(c *gin.Context) {
 		h.ensureStatuses().markPhase(newScan.JobID, phase)
 	}))
 	h.workerPool.Submit(func() {
-		h.ensureStatuses().markRunning(newScan.JobID)
+		if !h.ensureStatuses().markRunning(newScan.JobID) {
+			return
+		}
 		start := time.Now()
 		err = h.scanService.GenerateSBOM(bgCtx)
 		outcome := "success"
@@ -261,7 +263,9 @@ func (h *HTTPController) ScanCP(c *gin.Context) {
 		h.ensureStatuses().markPhase(newScan.JobID, phase)
 	}))
 	h.workerPool.Submit(func() {
-		h.ensureStatuses().markRunning(newScan.JobID)
+		if !h.ensureStatuses().markRunning(newScan.JobID) {
+			return
+		}
 		start := time.Now()
 		err = h.scanService.ScanCP(bgCtx)
 		if err != nil {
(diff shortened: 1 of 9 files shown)


controllers/scan_status.go
```diff
@@ -0,0 +1,129 @@
...
+import (
+	"sync"
+	"time"
+
+	"github.com/kubescape/kubevuln/core/domain"
+)
+
+type scanStatusStore struct {
+	mu    sync.RWMutex
+	items map[string]domain.ScanStatus
```
matthyx: **Blocker: `items` grows without bound for the life of the process.**

Nothing ever removes an entry from this map — `recordAccepted` only adds, and every `mark*` call updates in place. `kubevuln` is a long-running in-cluster daemon that can process many scans per day; every distinct `jobID` it ever sees (including terminal `succeeded`/`failed`/`abandoned` jobs) stays resident in memory for the pod's entire lifetime. On a busy cluster this is an unbounded memory leak, not just a cosmetic issue.

Please add some form of bounding before merge, e.g.:
- TTL-based eviction (drop entries once `FinishedAt` is older than N minutes), run from a background ticker or lazily on `get`/`recordAccepted`, or
- an LRU/size cap on `items`.

Given the issue's own "lightweight status store" framing, I'd treat this as blocking rather than a follow-up.

controllers/http.go
```diff
@@ -71,6 +73,13 @@ func (h *HTTPController) WithDiagnostics(f func(ctx context.Context) domain.Diag
 	return h
 }
 
+func (h *HTTPController) ensureStatuses() *scanStatusStore {
```
matthyx, line 78: Minor, non-blocking: this lazy nil-check-then-assign isn't synchronized (`h.statuses == nil` read and `h.statuses = ...` write aren't protected by any lock/atomic). It's a non-issue in production since `NewHTTPController` always sets `statuses` up front, but several existing tests in this package build `HTTPController{...}` via struct literal (e.g. `TestHTTPController_ScanCP_MissingArgsDoesNotPanic`), which leaves `statuses` nil until first use. If any test (now or later) ever fires concurrent requests against one of those literally-constructed controllers, this is a data race under `-race`. Simplest fix: drop the nil-guard and always require construction via `NewHTTPController`, or guard the lazy-init with a `sync.Once`.

matthyx (changes requested):
Nice feature — the lifecycle store design (queued/running/succeeded/failed/abandoned), the phase-updater context plumbing, and the shutdown-abandon handling all read cleanly, and the new tests cover the completed/failed/abandoned paths the linked issue asked for. CI is green.

One blocker before merge: `scanStatusStore.items` (`controllers/scan_status.go`) has no eviction — every `jobID` ever seen stays in memory for the process lifetime, which is an unbounded leak on a long-running daemon. Left an inline comment with suggested fixes (TTL eviction or an LRU cap). Also left a minor, non-blocking note on `ensureStatuses()`'s unsynchronized lazy-init.

Happy to re-review once the store has some form of bounding.

controllers/scan_status.go
```diff
@@ -0,0 +1,229 @@
...
+		State:      domain.ScanStateQueued,
+		Phase:      string(domain.ScanStateQueued),
+		AcceptedAt: now,
+		UpdatedAt:  now,
+	}
+	s.evictLocked(now)
+}
+
+func (s *scanStatusStore) markRunning(jobID string) bool {
+	if jobID == "" {
```
matthyx, line 127: **New blocker introduced by this fix, and CI is currently red because of it.**

`markRunning("")` unconditionally returns `false` here for an empty `jobID`. Every call site in `controllers/http.go` (`GenerateSBOM`, `ScanCP`, `ScanCVE`, `ScanRegistry`) now gates the actual scan on that return value:

```go
h.workerPool.Submit(func() {
    if !h.ensureStatuses().markRunning(newScan.JobID) {
        return
    }
    ...
    err = h.scanService.GenerateSBOM(bgCtx)  // never runs when jobID == ""
```

`jobID` is not a required field on the wire (`wssc.WebsocketScanCommand`/`RegistryScanCommand` don't validate it), and several existing tests exercise that exact path (no `jobID` in the payload). The handler still returns `200 OK`, but the worker-pool closure now returns immediately before ever calling into `scanService`, so the scan silently never executes, no metric is recorded, and no error is logged. That's a functional regression, not just a status-tracking bug — it can drop real scans.

This is what's failing CI right now: `TestHTTPController_ContextCancellationIsDetached` ("GenerateSBOM worker was not executed in time") and two subtests of `TestHTTPController_MetricsEndpoint_RecordsScanFailureReason` (`scanRegistryImage, classified reason` and `generateSBOM, unclassified error defaults to unexpected`) all fail on commit 206e4a5 for this reason.

The abandoned-revival fix itself is the right call, but the guard needs to allow the empty-jobID case through (e.g. treat `jobID == ""` as "untracked, always claim" the way `recordAccepted`/`markPhase`/`markTerminal` already do) rather than folding it into the same false/reject path as "job not found" or "already claimed by another worker".

matthyx (changes requested):
Re-reviewed after the latest push (commit 206e4a5). Good progress on the two blockers from the first pass:

- The store retention/eviction gap is fixed — `scanStatusStore` now has a TTL + max-entries eviction policy for terminal records, with tests covering both the mutation-triggered and lookup-triggered sweep.
- `ensureStatuses()` is now guarded with `sync.Once`, and the abandoned-during-shutdown race I hadn't caught (CodeRabbit did) is fixed too — `markRunning` now only claims from `queued`, so a worker can no longer revive an `abandoned` job back to `running`.

But that same `markRunning` change introduced a new regression: it returns `false` unconditionally for an empty `jobID`, and every handler now aborts the actual scan when `markRunning` returns false — so a request with no `jobID` (which is a supported, unvalidated case on the wire) gets a `200 OK` but the scan is silently never run. This is exactly what's failing CI on this commit (`TestHTTPController_ContextCancellationIsDetached` and two `TestHTTPController_MetricsEndpoint_RecordsScanFailureReason` subtests) — left an inline comment with the specific failing tests and a suggested fix.

Re-review once that's addressed and CI is green.

matthyx (approved):
Re-checked after commit d284bd7. The empty-`jobID` regression is fixed via `claimTrackedJob`, which treats an untracked (empty) `jobID` as "always claim" instead of folding it into the reject path — and there's a new regression test (`TestHTTPController_GenerateSBOM_EmptyJobIDStillRuns`) plus the existing suites now assert the scan worker actually starts before asserting on its result. All CI checks are green (lint, build, CodeQL, GitGuardian, DCO, basic tests).

Both original blockers (unbounded status-store growth, unsynchronized lazy-init) and the follow-up regression (dropped scans for untracked jobs) are resolved and covered by tests. LGTM — approving.

PR #637, opened 2026-08-12
feat: honor Retry-After header in HTTP retry logic 

### LINKED ISSUE #592: HTTP retry logic never reads Retry-After, ignoring rate-limit servers' explicit wait instructions
## Description

Nowhere in this codebase does anything read the standard HTTP
`Retry-After` header. `adapters/v1/backend.go`'s retry logic
(`httpPostWithContext`) correctly retries on rate-limit responses, but
always uses its own generic exponential backoff timing, never checking
whether the server explicitly said how long to actually wait.

## Why this matters

The future VEX fetcher (kubevuln#387) will make regular, periodic calls
to external vendor servers (Red Hat, Chainguard) -exactly the traffic
pattern that triggers rate-limiting. Without honoring `Retry-After`,
Kubescape could keep retrying on its own schedule regardless of what the
server actually asked for, risking getting blocked entirely by a vendor
for ignoring their explicit rate-limit response.

## Reproduction (verified)

A server responding `429` with `Retry-After: 3` -the current backoff
logic picks its own wait time (~470ms in a live test), completely
unrelated to the server's actual "wait 3 seconds" instruction.

## Proposed fix

A small, reusable helper that:
- Reads `Retry-After` from a response when present
- Handles both real formats: a plain number of seconds, or a full
  HTTP-date
- Falls back to normal exponential backoff when the header is missing
  or malformed
- Wires into the existing retry loop's per-attempt wait time

Shreya2005-2005:
## Overview

Nowhere in this codebase reads the standard HTTP `Retry-After` header.
`adapters/v1/backend.go`'s retry logic (`httpPostWithContext`) correctly
retries on rate-limit responses, but always uses its own generic
exponential backoff timing, completely ignoring an explicit "wait
exactly N seconds" instruction a server might send.

## Why this matters

The future External VEX Ingestion project (kubevuln#387) will make
regular, periodic calls to the same external vendor servers (Red Hat,
Chainguard) — exactly the traffic pattern that triggers rate-limiting.
Ignoring `Retry-After` risks Kubescape getting genuinely blocked by a
vendor for being a badly-behaved client.

## Before / After

**Before** — a server responding `429 Too Many Requests` with
`Retry-After: 3`:

```
Next retry wait chosen: ~470ms (completely unrelated to the server's actual "wait 3 seconds" instruction)
```

**After** — the same response:

```
Server's actual instruction read: "3" -> waiting exactly 3s, as told
```

Confirmed with a real, live end-to-end test (`TestHttpPostWithContext_HonorsRetryAfter`):
a server returns `429` + `Retry-After: 1` on the first call, `200` on
the second. Actual elapsed time: **1.03s** — matching the server's
request, not the default ~500ms backoff interval.

## Changes

- `parseRetryAfter(resp) (time.Duration, bool)` — reads `Retry-After` in
  both legitimate forms per RFC 9110 §10.2.3: a plain non-negative
  integer of seconds, or a full HTTP-date (via the stdlib's
  `http.ParseTime`). Falls back cleanly (`ok=false`) when the header is
  absent, negative, or malformed. A parsed date already in the past is
  clamped to zero wait, never negative.
- Wired into `httpPostWithContext`'s existing retry path using
  `backoff/v5`'s own native `backoff.RetryAfter(...)` mechanism — the
  exact library already in use here has first-class support for this,
  so no custom retry-timing logic was needed.

## Testing

- `TestParseRetryAfter` — 7 cases: absent header, valid seconds, zero
  seconds, negative (rejected), garbage value (rejected), valid future
  HTTP-date, past HTTP-date (clamped to zero).
- `TestHttpPostWithContext_HonorsRetryAfter` — real end-to-end proof
  using a live `httptest.Server`, described above.

Full repo build, `go vet`, and the full `./adapters/...` suite pass
(147 PASS, 0 FAIL).

## Related issues/PRs:

* Resolved #592

```

## Summary by CodeRabbit

* **Bug Fixes**
  * Retry handling now respects valid `Retry-After` response headers.
  * Requests correctly wait for the specified delay before retrying, including HTTP-date formats.
  * Invalid, missing, negative, or expired retry instructions are handled safely.

### DIFF
--- a/adapters/v1/backend.go
+++ b/adapters/v1/backend.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math"
 	"net/http"
 	"strconv"
 	"strings"
@@ -157,7 +158,13 @@ func httpPostWithContext(ctx context.Context, httpClient httputils.IHttpClient, 
 				return nil, backoff.Permanent(retryErr)
 			}
 			if wait, ok := parseRetryAfter(resp); ok {
-				return nil, backoff.RetryAfter(int(wait.Round(time.Second).Seconds()))
+				if wait > 0 {
+					wait = wait.Round(time.Second)
+					if wait == 0 {
+						wait = time.Second
+					}
+				}
+				return nil, backoff.RetryAfter(int(wait.Seconds()))
 			}
 			return nil, retryErr
 		}
@@ -176,7 +183,8 @@ func parseRetryAfter(resp *http.Response) (time.Duration, bool) {
 		return 0, false
 	}
 	if seconds, err := strconv.ParseInt(v, 10, 64); err == nil {
-		if seconds < 0 {
+		const maxRetryAfterSeconds = math.MaxInt64 / int64(time.Second)
+		if seconds < 0 || seconds > maxRetryAfterSeconds {
 			return 0, false
 		}
 		return time.Duration(seconds) * time.Second, true
--- a/adapters/v1/backend.go
+++ b/adapters/v1/backend.go
@@ -156,10 +156,39 @@ func httpPostWithContext(ctx context.Context, httpClient httputils.IHttpClient, 
 			if !shouldRetryReport(resp) {
 				return nil, backoff.Permanent(retryErr)
 			}
+			if wait, ok := parseRetryAfter(resp); ok {
+				return nil, backoff.RetryAfter(int(wait.Round(time.Second).Seconds()))
+			}
 			return nil, retryErr
 		}
 		return resp, nil
 	}, backoff.WithBackOff(bo), backoff.WithMaxElapsedTime(maxElapsedTime))
+}
+
+// parseRetryAfter reads the standard Retry-After header from resp, in either of its two
+// legitimate forms (RFC 9110 10.2.3): a plain non-negative number of seconds, or an
+// HTTP-date. Returns ok=false if the header is absent, empty, negative, or in neither
+// recognized form. A parsed date that has already passed is treated as "no wait" (zero
+// duration), not a negative one.
+func parseRetryAfter(resp *http.Response) (time.Duration, bool) {
+	v := resp.Header.Get("Retry-After")
+	if v == "" {
+		return 0, false
+	}
+	if seconds, err := strconv.ParseInt(v, 10, 64); err == nil {
+		if seconds < 0 {
+			return 0, false
+		}
+		return time.Duration(seconds) * time.Second, true
+	}
+	if t, err := http.ParseTime(v); err == nil {
+		wait := time.Until(t)
+		if wait < 0 {
+			wait = 0
+		}
+		return wait, true
+	}
+	return 0, false
 }
 
 // shouldRetryReport is derived from the unexported defaultShouldRetry in armosec/utils-go, but
--- a/adapters/v1/backend_test.go
+++ b/adapters/v1/backend_test.go
@@ -28,8 +28,8 @@ import (
 	"github.com/google/uuid"
 	"github.com/kinbiko/jsonassert"
 	beClientV1 "github.com/kubescape/backend/pkg/client/v1"
-	"github.com/kubescape/go-logger"
 	sysreport "github.com/kubescape/backend/pkg/server/v1/systemreports"
+	"github.com/kubescape/go-logger"
 	"github.com/kubescape/kubevuln/core/domain"
 	sev1beta1 "github.com/kubescape/kubevuln/pkg/securityexception/v1beta1"
 	"github.com/kubescape/kubevuln/repositories"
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by Shreya2005-2005** [comment]
Hi @matthyx ,the failing check is on `Test_syftAdapter_CreateSBOM`,
unrelated to this PR (only touched `adapters/v1/backend.go` and
`backend_test.go`). Looks like a network/timeout flake during image
resolution, not a real failure - all relevant tests pass locally
(147 PASS, 0 FAIL) with clean build/vet. Could you re-run the check
when you get a chance?

File: adapters/v1/backend.go [hunk 1]
```diff
@@ -156,12 +156,41 @@ func httpPostWithContext(ctx context.Context, httpClient httputils.IHttpClient,
...
+func parseRetryAfter(resp *http.Response) (time.Duration, bool) {
+	v := resp.Header.Get("Retry-After")
+	if v == "" {
+		return 0, false
+	}
+	if seconds, err := strconv.ParseInt(v, 10, 64); err == nil {
+		if seconds < 0 {
+			return 0, false
+		}
+		return time.Duration(seconds) * time.Second, true
```
matthyx, line 190: Confirmed this independently (and it matches CodeRabbit's finding below): `strconv.ParseInt(v, 10, 64)` accepts any value representable in an `int64`, but `time.Duration(seconds) * time.Second` is raw arithmetic — it silently overflows once `seconds > math.MaxInt64 / time.Second` (≈ 9,223,372,036). A response header like `Retry-After: 9223372037` parses successfully, then the multiplication wraps into a negative `time.Duration`. That negative value flows into `backoff.RetryAfter(int(wait.Round(time.Second).Seconds()))`, and a non-positive duration passed to the backoff timer fires immediately — bypassing backoff entirely for the rest of `maxElapsedTime`.

That's the opposite of this PR's stated goal (being a well-behaved client so vendors like Red Hat/Chainguard don't rate-limit or block us): a single malformed/oversized header value turns into a hot retry loop against the vendor.

Please add an upper-bound check next to the existing negative-seconds guard, e.g.:
```go
const maxRetryAfterSeconds = math.MaxInt64 / int64(time.Second)
if seconds < 0 || seconds > maxRetryAfterSeconds {
    return 0, false
}
```
and add a test case for a value just above that bound (analogous to the existing `"negative seconds rejected"` case).

This is blocking for me — everything else about the approach (using `backoff.RetryAfter`, the HTTP-date handling, the clamp-to-zero-not-negative behavior) looks correct and well tested.

adapters/v1/backend.go
```diff
@@ -156,12 +156,41 @@ func httpPostWithContext(ctx context.Context, httpClient httputils.IHttpClient,
 			if !shouldRetryReport(resp) {
 				return nil, backoff.Permanent(retryErr)
 			}
+			if wait, ok := parseRetryAfter(resp); ok {
+				return nil, backoff.RetryAfter(int(wait.Round(time.Second).Seconds()))
```
matthyx: Non-blocking nit, but agreeing with CodeRabbit's note here too: for an HTTP-date wait under ~500ms, `wait.Round(time.Second)` rounds down to `0`, so `backoff.RetryAfter(0)` retries immediately instead of waiting the sub-second delay the server actually asked for. Rounding positive waits up (rather than to-nearest) would be more faithful to the header, e.g.:
```go
if wait > 0 {
    wait = wait.Round(time.Second)
    if wait == 0 {
        wait = time.Second
    }
}
```
Worth a follow-up, doesn't need to hold up this PR.

matthyx (changes requested):
Reviewed the approach, the `parseRetryAfter` implementation, and both test cases. Overall this is a solid, well-tested change — the use of `backoff/v5`'s native `RetryAfter` mechanism is the right call, the HTTP-date clamp-to-zero-not-negative behavior is correct, and the end-to-end test genuinely proves the header is honored.

Requesting changes for one confirmed bug (inline comment on line 182): the seconds-parsing branch has an unchecked integer overflow — a large-but-valid `Retry-After` integer (>~9.2×10⁹ seconds) wraps `time.Duration(seconds) * time.Second` into a negative duration, which causes an immediate retry instead of backing off. This is the same issue CodeRabbit's automated review already flagged as Major; I independently verified the math and it's real. Given the PR's whole purpose is being a well-behaved client toward rate-limiting vendors, this edge case is worth closing before merge.

Also left a non-blocking nit (line 160) on sub-second HTTP-date rounding, matching CodeRabbit's other comment.

Happy to approve once the overflow guard is added.

Shreya2005-2005:
@matthyx  Fixed both issues  added the overflow check and the rounding fix you
suggested. Tests pass, build is clean. Ready for review again

matthyx (approved):
Verified commit 2575798 fixes both issues:

- **Overflow guard**: `maxRetryAfterSeconds := math.MaxInt64 / int64(time.Second)` with `seconds > maxRetryAfterSeconds` rejected — exactly closes the wraparound I flagged, and there's a new regression test (`"overflowing seconds rejected"`, `"9223372037"`) at the correct boundary.
- **Sub-second rounding**: now only rounds when `wait > 0`, and bumps a result that rounds down to `0` up to `time.Second` instead of retrying immediately.

CI is green across the board (build, lint, CodeQL, GitGuardian, license/security/dependency checks) and CodeRabbit's latest pass found no further actionable comments. Approving.

PR #640, opened 2026-08-12
fix(config): bind env for the keys that have no default

### LINKED ISSUE #639: env-only STORAGE and five other fields are silently ignored
`STORAGE=true` in the environment does not turn storage on unless `clusterData.json` also has a `storage` key. same for five other fields. no error, no log, the config just reads back false.

it is the gap #593 describes, still open for the fields it did not cover. `LoadConfig` calls `v.AutomaticEnv()`, but that only affects viper's getters. `Unmarshal` walks the keys viper knows about, from the config file, a default, or an explicit binding, and it cannot enumerate the environment, so a key in none of those never reaches the struct. the comment above the `cveMatchingMode` resolution already says as much:

```go
// with AutomaticEnv, a value supplied purely via environment variable is visible
// to IsSet/GetString but is NOT populated by Unmarshal
```

so whether an env override works comes down to whether the field happens to have a `SetDefault` above. loading `testdata` (whose `clusterData.json` has no key for any of these) with each variable set on its own:

```
STORAGE              env=true         -> false
KEEPLOCAL            env=true         -> false
RISKACCEPTANCE       env=true         -> false
STOREFILTEREDSBOM    env=true         -> false
NODESBOMGENERATION   env=true         -> false
PARTIALRELEVANCY     env=true         -> false
NAMESPACE            env=custom-ns    -> custom-ns
SCANCONCURRENCY      env=7            -> 7
```

the last two work only because they have defaults set.

`storage` is the one that stings: kubevuln comes up with storage off, so nothing is written to the apiserver, while the deployment looks correctly configured. `riskAcceptance` silently drops the SecurityException path. `accountID` and `clusterName` are in the same position, though those are normally in the file.

`v.BindEnv` on each of the keys that has no default is enough to register them. `cveMatchingMode` and `useDefaultMatchers` should stay out of it, since they are resolved through `IsSet` and already handle the env case.

version: main (cb4aa30), follow-up to #593

AnshulPatil2005:
## Overview

binds the config keys that have no `SetDefault` to the environment, so an env-only value reaches the struct.

`AutomaticEnv` only affects viper's getters. `Unmarshal` walks the keys viper knows about, from the config file, a default, or an explicit binding, and it cannot enumerate the environment, so a key in none of those was silently dropped. whether an env override worked came down to whether the field happened to have a default set above it. the comment on the `cveMatchingMode` resolution already documents the gap, and #593 worked around it for `trustedVendors` and `proxyRegistryMap` by reading through viper's getters.

six fields were still affected. `storage` is the one that stings: `STORAGE=true` on a `clusterData.json` with no `storage` key left kubevuln running with storage off, writing nothing to the apiserver, while the deployment looked correctly configured. `riskAcceptance` silently dropped the SecurityException path. `accountID` and `clusterName` were in the same position and are bound too, though those are normally in the file.

## Additional Information

`BindEnv` rather than `SetDefault`, because a default would make `IsSet` report the key as set even when nothing set it, and that presence detection is load-bearing here. the comment above `cveMatchingMode` calls it out. `BindEnv` leaves `IsSet` false when the variable is absent, so nothing else shifts.

`cveMatchingMode` and `useDefaultMatchers` are deliberately not bound. they are resolved through `IsSet` and already handle env, and binding them would be churn on logic that works.

precedence is unchanged: environment over file over default. there is a test for that, since binding keys is exactly the kind of change that could quietly invert it.

## How to Test

`go test ./config/ -count=1`

- `TestLoadConfigEnvOnlyFieldsWithoutDefaults` sets each variable on its own against the existing `testdata`, whose `clusterData.json` carries none of these keys, and asserts it lands. all six subtests fail before this.
- `TestLoadConfigEnvBindingKeepsFilePrecedence` asserts `accountID` still comes from the file when the environment is silent, and that the environment still wins when it is not.

the existing env tests for `TRUSTEDVENDORS`, `PROXYREGISTRYMAP`, `CVEMATCHINGMODE` and `USEDEFAULTMATCHERS` are untouched and still pass.

## Related issues/PRs:

* Resolved #639
* Follow-up to #593

## Summary by CodeRabbit

- **Bug Fixes**
  - Environment-based configuration values now load correctly for boolean settings.
  - Environment variables override configuration file values when provided, while file values remain unchanged when variables are unset.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -116,6 +116,32 @@ func LoadConfig(path string) (Config, error) {
 
 	v.AutomaticEnv()
 
+	// AutomaticEnv on its own does not make a key reachable by Unmarshal. viper can only
+	// unmarshal keys it knows about, from the config file, a default, or an explicit
+	// binding, and it cannot enumerate the environment, so a key that appears in none of
+	// those is silently dropped. Everything with a SetDefault above is already covered;
+	// these are the remaining fields, which clusterData.json does not always carry. Without
+	// this, STORAGE=true on a config file that has no "storage" key reads back as false and
+	// kubevuln runs with storage off while looking correctly configured. Same gap #593 fixed
+	// for trustedVendors and proxyRegistryMap by reading through viper's getters.
+	//
+	// cveMatchingMode and useDefaultMatchers are deliberately left out: they are resolved
+	// through IsSet below, which already handles the env case.
+	for _, key := range []string{
+		"accountID",
+		"clusterName",
+		"keepLocal",
+		"nodeSbomGeneration",
+		"partialRelevancy",
+		"riskAcceptance",
+		"storage",
+		"storeFilteredSbom",
+	} {
+		if err := v.BindEnv(key); err != nil {
+			return Config{}, fmt.Errorf("binding %s to the environment: %w", key, err)
+		}
+	}
+
 	err := v.ReadInConfig()
 	if err != nil {
 		return Config{}, err
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -291,3 +291,49 @@ func TestLoadConfigProxyRegistryMapEnv_Invalid(t *testing.T) {
 	assert.Error(t, err)
 	assert.Contains(t, err.Error(), "proxyRegistryMap")
 }
+
+// TestLoadConfigEnvOnlyFieldsWithoutDefaults verifies that a field carried only by the
+// environment reaches the struct. AutomaticEnv does not achieve that on its own: viper
+// unmarshals the keys it knows about, and it cannot enumerate the environment, so a key
+// with no default and no entry in clusterData.json was dropped. STORAGE is the one that
+// matters most, since kubevuln would then run with storage off while looking configured
+// for it.
+func TestLoadConfigEnvOnlyFieldsWithoutDefaults(t *testing.T) {
+	tests := []struct {
+		name  string
+		env   string
+		value string
+		got   func(Config) bool
+	}{
+		{"storage", "STORAGE", "true", func(c Config) bool { return c.Storage }},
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Reviewed the fix and traced it against Viper's actual precedence/env semantics — it's correct.

**What I verified**
- The 8 keys bound (`accountID`, `clusterName`, `keepLocal`, `nodeSbomGeneration`, `partialRelevancy`, `riskAcceptance`, `storage`, `storeFilteredSbom`) are exactly the `Config` fields with no `SetDefault` and not already special-cased via `IsSet`+getter (`cveMatchingMode`, `useDefaultMatchers`, `trustedVendors`, `proxyRegistryMap`). No field is missed and none is redundantly bound.
- `BindEnv` makes viper aware of the key so `Unmarshal` can see it, while leaving `find()`'s normal precedence intact: env is checked before the config file, but an *unset* env var falls through to the file value, so file precedence when the env is silent still holds.
- Env var names derived from `BindEnv` (`ACCOUNTID`, `STORAGE`, `KEEPLOCAL`, etc.) match the ones the new tests set — confirmed by running `go test ./config/... -run TestLoadConfig -v -count=1` locally: all pass, including both new tests.
- CI is green (lint, CodeQL, security scans, cross-platform build, `pr-created/test`), and the PR is `MERGEABLE` with no conflicts against `main`.

**Non-blocking nit** (CodeRabbit already flagged this): `TestLoadConfigEnvBindingKeepsFilePrecedence`'s first assertion (file value used when env is silent) doesn't clear `ACCOUNTID` from the test process's environment first. If a CI runner ever had `ACCOUNTID` set, that subtest would spuriously fail. Not a blocker — worth a follow-up `t.Setenv("ACCOUNTID", "")` if you're touching this test again.

No blockers found. Approving.

PR #641, opened 2026-08-12
refactor: extract the duplicated cache-hit exception reconciliation into reconcileCachedCVE

### LINKED ISSUE #638: ScanCP/ScanCVE/ScanRegistry duplicate the exception cache-hit reconciliation block, which is why the same VEX/persistence bug has been fixed three separate times
# Description
`core/services/scan.go`'s three scan entry points — `ScanCP`, `ScanCVE`, `ScanRegistry` — each carry their own near-identical ~40-50 line copy of the "cached CVE manifest was hit, now re-evaluate it against the current SecurityException set" block:

- `ScanCP`: lines 326-364
- `ScanCVE`: lines 526-577
- `ScanRegistry`: lines 690-741

The `ScanCVE` and `ScanRegistry` copies are byte-identical except for a single comment line-wrap (confirmed with `diff`). Each block independently: snapshots which CVEs the stored manifest currently suppresses (`v1.IgnoredMatchKeys`), re-filters the manifest against the live exception set (`v1.RestoreSuppressedMatches` + `applyExceptionsToManifest`), diffs old vs. new suppression sets, hand-computes whether any suppression was *removed* (`hasRemovals`), and gates persistence/VEX-republishing on that. `ScanCP`'s copy is the same block minus the VEX-republish step, since `ScanCP` never calls `storeVEX` on any path (that omission is consistent with its cache-miss behavior, not a bug in itself).

There is no shared function for this sequencing logic — every flow re-derives it by hand, in parallel with the existing shared helpers `applyExceptionsToManifest` (line 794) and `storeVEX` (line 776) that each copy already calls into.

This is not a hypothetical maintainability concern: this exact duplication has already produced three separate, previously-filed and closed bugs against this precise block, each one a case of a fix landing in one flow and being missed in another:
- #469 — deleting a SecurityException never restored suppression on cached scans (the motivation for this block existing).
- #501 — `ScanCP` passed unfiltered manifests to `StoreVEX` and omitted VEX updates on cached exception re-evaluation.
- #557 — `ScanCVE` never generated VEX documents, and `ScanRegistry`'s cache-hit path *still* didn't after a first fix attempt — the same issue had to be re-applied to a second flow.

The adjacent SBOM-creation logic in these same four flows had this exact category of duplication already recognized and fixed: `getOrCreateSBOM` (line 1296) was extracted to unify SBOM creation across all four flows (#605/#610), with #622 (open) chasing a follow-on gap discovered in that consolidated code. This cache-hit reconciliation block is the one piece of orchestration logic from the same subsystem that never received the same treatment.

# Environment
OS: `Any`
Version: `main`

# Steps To Reproduce
1. Open `core/services/scan.go` and locate the three blocks starting with the comment `// A cached manifest was filtered against the exception set at store time.` at lines 327, 526, and 690.
2. Run `diff <(sed -n '526,577p' core/services/scan.go) <(sed -n '690,741p' core/services/scan.go)` — confirms the `ScanCVE` and `ScanRegistry` copies are identical except one comment line-wrap.
3. Look up #469, #501, #557 — each is a fix to this exact block that had to be independently discovered and applied to one flow, then later found missing in another.
4. Note that no shared helper exists for this sequencing (unlike the sibling `getOrCreateSBOM` extraction for SBOM caching), so any future change to this logic (e.g. adjusting the persist-safety rule, or wiring in external VEX ingestion per #387) must again be manually ported across three call sites by hand.

# Expected behavior
This snapshot/diff/persist-safety/republish sequence should exist as a single shared method on `ScanService` (e.g. `reconcileCachedCVE`), called by `ScanCP`, `ScanCVE`, and `ScanRegistry`, with the one real behavioral difference between flows (whether to call `storeVEX`) expressed as an explicit parameter rather than by omitting the whole block. A future change to this logic should require editing one function, not three.

# Actual Behavior
The logic is copy-pasted three times with no shared abstraction. Every change to this behavior requires manually keeping three call sites in sync, and history shows that doesn't reliably happen: #469, #501, and #557 are three separate bug reports that all trace back to this same duplicated block drifting out of sync between flows.

# Additional context
**Affected files for a fix:**
- `core/services/scan.go` — extract the shared helper, replace the three call sites in `ScanCP`, `ScanCVE`, `ScanRegistry`.
- `core/services/scan_test.go` — each flow's cache-hit path is currently tested independently and would need to be re-pointed at the shared helper.
- Possibly `docs/security-exception-design-review.md` if it documents this behavior per-flow.

**Acceptance criteria:**
- A single shared method replaces the three duplicated blocks, preserving existing persist-safety semantics exactly (`exceptionsComplete || !hasRemovals` for `StoreCVE`/`StoreCVESummary`, `exceptionsComplete` alone for `storeVEX`) — this should be a refactor, not a behavior change.
- `ScanCP`'s lack of a `storeVEX` call stays explicit (e.g. a `withVEX bool` parameter) rather than silently reappearing or silently vanishing from the other two flows.
- Existing tests for all three flows' cache-hit paths continue to pass.
- `GenerateSBOM` is unaffected — it doesn't have this cache-hit CVE-reconciliation concern (only SBOM caching) and is out of scope for this extraction.

1. Extract the shared sequence out of `ScanCVE` (the most complete of the three copies) into a private `ScanService` helper.
2. Parameterize the one real difference (whether to call `storeVEX`) explicitly.
3. Replace the three call sites in `ScanCP`, `ScanCVE`, `ScanRegistry` with calls to the new helper.
4. Consolidate the corresponding unit tests in `core/services/scan_test.go` to test the shared helper directly, plus one integration-level assertion per flow that it's wired in correctly.
5. Spot-check `docs/security-exception-design-review.md` for any per-flow description that should now point at the shared helper.

bhuvan-somisetty:
## Overview

`ScanCP`, `ScanCVE`, and `ScanRegistry` (`core/services/scan.go`) each carried their own near-identical ~40-50 line block that re-evaluates a cached CVE manifest against the current SecurityException set: snapshot which CVEs are currently suppressed, restore suppressed matches, re-filter, diff old vs. new suppression to decide what's safe to persist, and — for `ScanCVE`/`ScanRegistry` — republish VEX under a removal-safety rule. The `ScanCVE` and `ScanRegistry` copies were byte-identical except for a single comment line-wrap.

This duplication is why the same class of bug kept resurfacing in one flow after being fixed in another: #469, #501, and #557 are three separate, previously closed issues against this exact block, each a fix applied to one function that had to be independently re-discovered and re-applied to a second flow later.

This PR extracts the sequence into a single `reconcileCachedCVE` method. The one real behavioral difference between callers — whether to republish VEX at that point — is now an explicit `publishVEX` parameter: `ScanCVE` and `ScanRegistry` pass `true` (matching their prior behavior), `ScanCP` passes `false` since it republishes VEX later itself, once relevancy-filtered (`cvep`) results are also available, gated on both exception-completeness flags.

No behavior change — this is a pure refactor, verified against the existing regression suite (see below).

## How to Test

```
go test ./core/services/... -run 'TestScanService_ScanCP$|TestScanService_ScanCVE$|TestScanService_ScanRegistry$|TestScanService_ScanRegistry_StorageAndExceptions|CacheHit|CacheMiss' -v
```

This exercises 54 individual test cases across `scan_test.go`'s existing coverage of cache-hit/cache-miss behavior for all three flows — deleted/added exceptions, `ExpiredOnFix` transitions, degraded exception fetches (partial CRD-list failures), and VEX republish gating — all passing unmodified against the refactored code.

```
go test ./...
```

Full suite passes except three environment-dependent tests requiring a local Docker/testcontainers Grype-DB fixture (`TestScanService_NginxTest`, `Test_grypeAdapter_DBVersion`, `Test_grypeAdapter_ScanSBOM`), unrelated to this change.

## Related issues/PRs:

* Resolved #638

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency when reconciling cached vulnerability results across scan types.
  * Preserved suppressed findings when security exception data is incomplete.
  * Ensured updated exceptions are applied correctly to cached results.
  * Improved VEX publication timing so only relevant, complete results are republished.

### DIFF
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -349,43 +349,9 @@ func (s *ScanService) ScanCP(mainCtx context.Context) error {
 				}
 			}
 		} else {
-			// A cached manifest was filtered against the exception set at store time.
-			// Reconstruct the unfiltered manifest so the CURRENT exception set is
-			// re-evaluated (deleted exceptions and ExpiredOnFix transitions
-			// un-suppress findings) and SubmitCVE/StoreVEX receive the same
-			// unfiltered data a cache miss would produce.
-			prevIgnored := v1.IgnoredMatchKeys(cve.Content)
-			cve.Content = v1.RestoreSuppressedMatches(cve.Content)
-			filteredCve, cveExceptionsComplete = s.applyExceptionsToManifest(ctx, cve)
-
-			if s.storage {
-				curIgnored := v1.IgnoredMatchKeys(filteredCve.Content)
-				if !maps.Equal(prevIgnored, curIgnored) {
-					// Persist additions freely, but never persist removals when the
-					// exception set is incomplete: a transient SecurityException CRD
-					// list failure must not look like a deletion and wipe suppression
-					// from the stored manifest.
-					hasRemovals := false
-					for k := range prevIgnored {
-						if _, ok := curIgnored[k]; !ok {
-							hasRemovals = true
-							break
-						}
-					}
-					if cveExceptionsComplete || !hasRemovals {
-						err = s.cveRepository.StoreCVE(ctx, filteredCve, false)
-						if err != nil {
-							logger.L().Ctx(ctx).Warning("storing CVE with exceptions", helpers.Error(err),
-								helpers.String("imageSlug", slug))
-						}
-						err = s.cveRepository.StoreCVESummary(ctx, filteredCve, domain.CVEManifest{}, false)
-						if err != nil {
-							logger.L().Ctx(ctx).Warning("storing CVE summary with exceptions", helpers.Error(err),
-								helpers.String("imageSlug", slug))
-						}
-					}
-				}
-			}
+			// VEX is not published here: ScanCP republishes it later, once
+			// filteredCvep is also available, gated on both exceptionsComplete flags.
+			cve, filteredCve, cveExceptionsComplete = s.reconcileCachedCVE(ctx, cve, slug, false)
 		}
 
 		// generate SBOM' from SBOM and relevant files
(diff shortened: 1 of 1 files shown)


matthyx (approved):
Reviewed the diff against the pre-refactor code in `ScanCP`, `ScanCVE`, and `ScanRegistry`.

- Traced each call site's return values through to their later use (`ScanCP`'s `filteredCve`/`cveExceptionsComplete` feed the later `storeVEX` gate at line ~407; `ScanCVE`/`ScanRegistry` only reuse `cve`, correctly discarded the rest) — all match prior behavior exactly.
- `reconcileCachedCVE` reproduces the removal-safety persistence rule and VEX-republish gating verbatim, with `publishVEX` correctly wired: `true` for `ScanCVE`/`ScanRegistry`, `false` for `ScanCP` (which republishes later once `cvep` is available).
- Value-receiver `cve domain.CVEManifest` mutate-then-return preserves the original mutate-in-place semantics for the caller's `cve` variable.
- No test or behavioral changes; all CI checks (build, lint, tests, CodeQL, GitGuardian) are green, and CodeRabbit's automated pass found no actionable issues.

No blockers. Nice reduction of the triplicated block that kept causing repeat bugs (#469, #501, #557).

PR #644, opened 2026-08-12
fix(adapters): resolve IntroducedInLayer for packages outside the base layer

### LINKED ISSUE #643: IntroducedInLayer is empty for every package outside the base layer
`IntroducedInLayer` comes out empty for any package that is not in the image's first layer, which is most of them.

`DomainToArmo` resolves it by walking the parent chain:

```go
earlyLayer := ""
for j, layer := range v.Layers {
	if layer.ParentLayerHash == earlyLayer {
		earlyLayer = layer.LayerHash
	}
	...
}
vulnerabilityResults[i].IntroducedInLayer = earlyLayer
```

`parentLayer` gives the image's first layer a parent of `""`, so the walk can only start if the package appears in that layer. for a package added by any later layer nothing matches `earlyLayer == ""`, the loop never advances, and the field is left empty.

on a three layer image, with the package in one layer at a time:

```
package located in sha256:l1  -> IntroducedInLayer="sha256:l1"
package located in sha256:l2  -> IntroducedInLayer=""
package located in sha256:l3  -> IntroducedInLayer=""
```

the base layer holds the distro packages, so those resolve. anything installed by a later build step, which is usually where the interesting findings are, reports nothing. it is sent to the backend on every vulnerability in `CommonContainerVulnerabilityResult`, so a consumer trying to point at the build step that introduced a CVE gets an empty string.

the existing `Test_domainToArmo` case uses a single layer image whose only layer is the placeholder, which is the one shape where the walk does work, so it passes.

the position of a layer in `imageMetadata.Layers` is already known where `parentLayer` is built, so the earliest layer a package appears in can be picked directly instead of walked. the placeholder layer used for a match with no locations is not one of the image's own and needs to keep resolving to itself.

version: main (746a949)

AnshulPatil2005:
## Overview

`IntroducedInLayer` was resolved by walking the parent chain from `""`. the image's first layer is the only one whose parent is `""`, so the walk could only ever start if the package appeared in that layer. for a package added by any later build step nothing matched, the loop never advanced, and the field went out empty.

that is most packages. the base layer holds the distro packages, so those resolved; anything installed later, which is usually where the interesting findings are, reported nothing. the field ships on every vulnerability in `CommonContainerVulnerabilityResult`, so a consumer pointing at the build step that introduced a CVE got an empty string.

the position of each layer in `imageMetadata.Layers` is already known where `parentLayer` is built, so this records it there and picks the earliest layer the package appears in directly. no walking, and it does not care what order `Artifact.Locations` comes back in.

## Additional Information

a match with no locations is given a placeholder layer earlier in the function, and that layer is not one of the image's own, so it has no position. it keeps resolving to itself, which is what the walk did for it before. `Test_domainToArmo`'s existing case is exactly that shape, which is why it passed while the real case was broken, so it is worth keeping as its own path rather than folding it in.

`parentLayer` is still built and used, it feeds `ParentLayerHash` on each layer. only the `IntroducedInLayer` resolution changes.

## How to Test

`go test ./adapters/v1/ -run Test_domainToArmo -count=1`

`Test_domainToArmo_introducedInLayer` puts a package in each layer of a three layer image in turn, then in several layers at once to check the earliest is chosen regardless of the order the locations arrive in, plus a layer the image does not have.

four of the six cases fail before this:

```
middle layer                     expected: "sha256:l2"  actual: ""
top layer                        expected: "sha256:l3"  actual: ""
several layers, earliest wins    expected: "sha256:l2"  actual: ""
several layers, already ordered  expected: "sha256:l2"  actual: ""
```

the two that pass beforehand are the base layer and the unknown layer, which are the shapes the old walk could handle.

## Related issues/PRs:

* Resolved #643

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved vulnerability reporting to identify the earliest image layer where a package was introduced.
  * Added fallback handling when no matching image layer is available, improving consistency for unknown package layers.

* **Tests**
  * Expanded coverage for base, middle, top, multiple, and unknown package-layer scenarios.

### DIFF
--- a/adapters/v1/domain_to_armo.go
+++ b/adapters/v1/domain_to_armo.go
@@ -98,6 +98,9 @@ func DomainToArmo(ctx context.Context, grypeDocument v1beta1.GrypeDocument, vuln
 		parentLayer := map[string]string{
 			dummyLayer: parentLayerHash,
 		}
+		// ...and one of layer to its position in the image, so the earliest layer a package
+		// appears in can be picked directly rather than by walking the chain from the root.
+		layerPosition := map[string]int{}
 
 		target, err := NewTargetFromSource(grypeDocument.Source)
 		if err != nil {
@@ -106,9 +109,10 @@ func DomainToArmo(ctx context.Context, grypeDocument v1beta1.GrypeDocument, vuln
 
 		if target.IsImageTarget() {
 			imageMetadata := target.GetImageMetadata()
-			for _, layer := range imageMetadata.Layers {
+			for i, layer := range imageMetadata.Layers {
 				parentLayer[layer.Digest] = parentLayerHash
 				parentLayerHash = layer.Digest
+				layerPosition[layer.Digest] = i
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Verified the fix: `layerPosition` records each image layer's index once (from `imageMetadata.Layers`), and the earliest-layer lookup now picks the minimum position directly instead of depending on a parent-chain walk that could only ever start from the base layer. Traced through all six new test cases (base/middle/top/multi-layer/ordering/unknown) and the logic holds, including the placeholder-layer fallback for matches with no locations. CI (build, lint, tests) and CodeRabbit are green. LGTM.

PR #646, opened 2026-08-12
fix(services): match relevant-file paths that carry a wildcard placeholder

### LINKED ISSUE #645: relevant-file paths collapsed to a wildcard are never matched, so their packages look irrelevant
`filterSBOM` treats a relevant-file path as dynamic only if it contains `DynamicIdentifier`:

```go
relevantFiles.Each(func(file string) bool {
	if strings.Contains(file, dynamicpathdetector.DynamicIdentifier) {
		segmentCount := strings.Count(file, "/")
		dynamicPathsBySegmentCount[segmentCount] = append(dynamicPathsBySegmentCount[segmentCount], file)
	}
	return false
})
```

there are two placeholders though, not one:

```go
DynamicIdentifier  string = "⋯"  // U+22EF: ⋯ (one segment / one arg)
WildcardIdentifier string = "*"  // zero-or-more path segments (opens only)
```

and the detector produces the second from the first, collapsing runs of `⋯` into a single `*`, so `/a/⋯/⋯/b` is stored in the profile as `/a/*/b`. those paths reach us through `containerProfile.Spec.Opens`, which is exactly where the comment says `*` comes from.

a path like that contains no `⋯`, so it is never collected, never compared with `CompareDynamic`, and only ever gets the exact string equality check, which a pattern cannot pass. the file is dropped from the filtered SBOM, its package goes with it, and the relevancy scan reports a package that is loaded as not relevant.

a path carrying both, `/usr/lib/⋯/*/mod.so`, is collected but then bucketed by its own segment count, and the `*` means the file it matches has a different one, so the bucket never holds it either.

`CompareDynamic` against the two kinds:

```
"/usr/lib/⋯/mod.so"   vs "/usr/lib/python3/mod.so"  match=true   collected=true   segs 4/4
"/usr/lib/⋯/mod.so"   vs "/usr/lib/a/b/mod.so"      match=false  collected=true   segs 4/5
"/usr/lib/*/mod.so"   vs "/usr/lib/a/b/mod.so"      match=true   collected=false  segs 4/5
"/usr/lib/*/mod.so"   vs "/usr/lib/mod.so"          match=true   collected=false  segs 4/3
"/usr/lib/*"          vs "/usr/lib/a/b/c"           match=true   collected=false  segs 3/5
"/usr/lib/⋯/*/mod.so" vs "/usr/lib/a/b/c/mod.so"    match=true   collected=true   segs 5/6
```

the segment-count index from #448 is right for `⋯`, which really is one segment and cannot match across counts, as the second line shows. it is only `*` that does not fit a bucket.

so the split is: keep the index for paths carrying only `⋯`, and compare paths carrying `*` against every file, since no bucket can hold them.

version: main (746a949)

AnshulPatil2005:
## Overview

`filterSBOM` only recognised one of the two placeholders a relevant-file path can carry. it collected paths containing `DynamicIdentifier` (`⋯`, one segment) and ignored `WildcardIdentifier` (`*`, zero or more segments), which the detector produces by collapsing runs of `⋯`, so `/a/⋯/⋯/b` reaches us as `/a/*/b` through `containerProfile.Spec.Opens`.

such a path was never compared with `CompareDynamic` at all. it only got the exact string equality check, which a pattern cannot pass, so the file was dropped from the filtered SBOM, its package went with it, and the relevancy scan reported a package that is loaded as not relevant. a path carrying both, `/usr/lib/⋯/*/mod.so`, was collected but then bucketed by its own segment count, which the `*` makes wrong.

the segment-count index from #448 is correct for `⋯`: it really is one segment and cannot match across counts. so this keeps that index for paths carrying only `⋯` and compares the `*` ones against every file, since no bucket can hold them.

## Additional Information

the two groups are matched separately rather than by concatenating the bucket with the wildcard list, since appending to a slice that came out of the map would write into that bucket's spare capacity and corrupt it for later files.

the `*` list is only the relevant-file paths that contain a `*`, not the whole set, so the extra work is bounded by how many collapsed patterns a profile actually has rather than by the file count.

## How to Test

`go test ./core/services/ -run TestFilterSBOM -count=1`

`TestFilterSBOM_WildcardRelevantPaths` runs a file and the package that owns it through the filter for each placeholder shape.

four of the six fail before this, all of them `*` cases:

```
wildcard spanning several segments   "/usr/lib/*/mod.so"    vs /usr/lib/a/b/mod.so
wildcard spanning none               "/usr/lib/*/mod.so"    vs /usr/lib/mod.so
trailing wildcard                    "/usr/lib/*"           vs /usr/lib/a/b/c
both placeholders together           "/usr/lib/⋯/*/mod.so"  vs /usr/lib/a/b/c/mod.so
```

the other two are the `⋯` cases, one that must match and one that must not, and they pass before and after: the point is to keep #448's index working and to not start over-matching in its place.

## Related issues/PRs:

* Resolved #645
* Follow-up to #448

### DIFF
--- a/adapters/v1/domain_to_armo.go
+++ b/adapters/v1/domain_to_armo.go
@@ -98,6 +98,9 @@ func DomainToArmo(ctx context.Context, grypeDocument v1beta1.GrypeDocument, vuln
 		parentLayer := map[string]string{
 			dummyLayer: parentLayerHash,
 		}
+		// ...and one of layer to its position in the image, so the earliest layer a package
+		// appears in can be picked directly rather than by walking the chain from the root.
+		layerPosition := map[string]int{}
 
 		target, err := NewTargetFromSource(grypeDocument.Source)
 		if err != nil {
@@ -106,9 +109,10 @@ func DomainToArmo(ctx context.Context, grypeDocument v1beta1.GrypeDocument, vuln
 
 		if target.IsImageTarget() {
 			imageMetadata := target.GetImageMetadata()
-			for _, layer := range imageMetadata.Layers {
+			for i, layer := range imageMetadata.Layers {
 				parentLayer[layer.Digest] = parentLayerHash
 				parentLayerHash = layer.Digest
+				layerPosition[layer.Digest] = i
 			}
 		}
 
@@ -202,10 +206,15 @@ func DomainToArmo(ctx context.Context, grypeDocument v1beta1.GrypeDocument, vuln
 
 			// fill extra layer information
 			for i, v := range vulnerabilityResults {
+				// The package is introduced by the earliest layer it appears in. This used to
+				// walk the parent chain from "", which meant it only ever resolved for a
+				// package present in the image's first layer: for anything added by a later
+				// layer no element could start the chain, and the result stayed empty.
 				earlyLayer := ""
+				earlyPosition := 0
 				for j, layer := range v.Layers {
-					if layer.ParentLayerHash == earlyLayer {
-						earlyLayer = layer.LayerHash
+					if position, ok := layerPosition[layer.LayerHash]; ok && (earlyLayer == "" || position < earlyPosition) {
+						earlyLayer, earlyPosition = layer.LayerHash, position
 					}
 					if l, ok := data[layer.LayerHash]; ok {
 						if layer.LayerInfo == nil {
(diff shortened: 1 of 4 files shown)


matthyx (approved):
Reviewed both fixes against the storage `dynamicpathdetector` package semantics (⋯ = exactly one segment, `*` = zero-or-more mid-path / one-or-more trailing) and traced the layer-position logic by hand.

**`core/services/scan.go` (`filterSBOM`)** — correct. Splitting `WildcardIdentifier` paths out of the segment-count buckets and comparing them against every file is the right fix: a `*` pattern can span any number of segments, so no fixed bucket can hold it, and the `⋯`-only paths correctly keep the #448 bucket since that placeholder is always exactly one segment. The new `TestFilterSBOM_WildcardRelevantPaths` cases (spanning several segments, spanning none, trailing wildcard, both placeholders combined) match this semantics and I confirmed they reproduce the described pre-fix failures.

**`adapters/v1/domain_to_armo.go` (`IntroducedInLayer`)** — correct. Replacing the parent-chain walk (which could only ever resolve if a match's `Layers` happened to contain the image's first layer) with a `layerPosition` index and taking the minimum position is the right approach, and it's an improvement even in the untested edge case where a match mixes an image layer with an unrecognized digest — it now deterministically prefers the real image layer regardless of slice order. The fallback for a match with no locations (placeholder layer) also matches the old code's behavior for that case. Test cases (base/middle/top layer, several layers picking the earliest, unrecognized layer) all check out.

No blockers. CI is green (build, lint, tests, CodeQL, GitGuardian, DCO). Nice write-up in the PR description with a clear before/after failure list.

PR #648, opened 2026-08-12
fix(adapters): give the summary a stable severity stats order

### LINKED ISSUE #647: summary severity stats come out in a different order on every scan
`Summarize` collects severity stats in two maps and emits them in iteration order:

```go
for sever := range actualSeveritiesStats {
	summary.SeveritiesStats = append(summary.SeveritiesStats, actualSeveritiesStats[sever])
}
for sever := range exculdedSeveritiesStats {
	summary.ExcludedSeveritiesStats = append(summary.ExcludedSeveritiesStats, exculdedSeveritiesStats[sever])
}
```

go randomises map iteration, so the same findings give a different order every run. that summary is the payload posted to the backend, so two scans of an image that has not changed report their severity stats in a different order each time, and anything diffing consecutive reports sees movement that is not there.

the existing table test already works around it:

```go
got, _ := Summarize(tt.args.report, tt.args.vulnerabilities, ...)
sort.Slice(got.SeveritiesStats, func(i, j int) bool {
	return got.SeveritiesStats[i].Severity < got.SeveritiesStats[j].Severity
})
assert.Equal(t, tt.want, got)
```

which is the test sorting because the code does not. note it only sorts `SeveritiesStats`, so `ExcludedSeveritiesStats` is unordered and unasserted.

sorting both in `Summarize` is enough, and then the test can compare the order directly instead of normalising it away. `parseSeverities` in `repositories/apiserver.go`, which builds the stored CRD summary, walks a slice rather than a map, so it is not affected.

version: main (746a949)

AnshulPatil2005:
## Overview

`Summarize` built `SeveritiesStats` and `ExcludedSeveritiesStats` by ranging over a map, and go randomises map iteration, so the same findings came out in a different order on every run. that summary is the payload posted to the backend, so two scans of an unchanged image reported their severity stats in a different order each time and anything diffing consecutive reports saw movement that was not there.

both are now sorted by severity name before they go on the summary.

## Additional Information

the existing table test was sorting `SeveritiesStats` itself before comparing, which is the test normalising away something the code should not have been doing. that sort is removed, so the assertion pins the order rather than hiding it. it only ever sorted the reported stats, so `ExcludedSeveritiesStats` was neither ordered nor asserted.

sorting by severity name rather than by rank, since that is what the test was already doing and it needs no severity-to-score table to stay stable.

`parseSeverities` in `repositories/apiserver.go`, which builds the stored CRD summary, walks a slice rather than a map, so it was never affected and is untouched.

## How to Test

`go test ./adapters/v1/ -run Summarize -count=1`

`TestSummarize_SeverityStatsOrderIsStable` builds one vulnerability per severity with every other one suppressed, so both lists have several entries, then calls `Summarize` twenty times and asserts the order is sorted and identical each time. repeated because a single call can land in the right order by chance.

before this it fails on the excluded list being unordered, and on later runs disagreeing with the first.

`Test_summarize`'s expectations are unchanged, they just no longer need the sort in front of them.

## Related issues/PRs:

* Resolved #647

## Summary by CodeRabbit

* **Bug Fixes**
  * Severity statistics are now consistently returned in ascending severity-name order.
  * Ordering remains stable for both reported and excluded statistics, including when findings are suppressed.

### DIFF
--- a/adapters/v1/backend_utils.go
+++ b/adapters/v1/backend_utils.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
@@ -319,8 +320,22 @@ func Summarize(report v1.ScanResultReport, vulnerabilities []containerscan.Commo
 	for sever := range exculdedSeveritiesStats {
 		summary.ExcludedSeveritiesStats = append(summary.ExcludedSeveritiesStats, exculdedSeveritiesStats[sever])
 	}
+	// Both were collected in a map, whose iteration order Go randomises, so without this the
+	// same findings produce a differently ordered summary on every scan. That summary is the
+	// payload posted to the backend, so a consumer diffing two reports for the same image
+	// sees the severity stats move around when nothing about the image changed.
+	sortBySeverity(summary.SeveritiesStats)
+	sortBySeverity(summary.ExcludedSeveritiesStats)
 
 	return &summary, vulnerabilities
+}
+
+// sortBySeverity orders severity stats by severity name, so a summary built from the same
+// findings is byte for byte the same each time.
+func sortBySeverity(stats []containerscan.SeverityStats) {
+	sort.Slice(stats, func(i, j int) bool {
+		return stats[i].Severity < stats[j].Severity
+	})
 }
 
 func getCVEExceptionMatchCVENameFromList(srcCVEList []armotypes.VulnerabilityExceptionPolicy, CVEName string, filterFixed bool) []armotypes.VulnerabilityExceptionPolicy {
--- a/adapters/v1/backend_utils_test.go
+++ b/adapters/v1/backend_utils_test.go
@@ -8,7 +8,6 @@ import (
 	"io"
 	"net/http"
 	"os"
-	"sort"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Reviewed the diff and the surrounding `Summarize` code in `adapters/v1/backend_utils.go`.

- The root cause matches: `SeveritiesStats`/`ExcludedSeveritiesStats` were built by ranging over `map[string]containerscan.SeverityStats`, so their order was nondeterministic. Sorting both slices by `Severity` before returning fixes that.
- `sortBySeverity` is a plain, safe `sort.Slice`; works fine on nil/empty slices.
- Removing the sort from `Test_summarize`'s comparison correctly turns that table test into a pin on the new deterministic order rather than a normalizer that hid the bug.
- `TestSummarize_SeverityStatsOrderIsStable` is a solid regression test — it builds both suppressed and non-suppressed findings, runs `Summarize` 20 times, and asserts the order is sorted and stable across runs, which is what actually would have caught the original flake.
- Change is scoped correctly: `parseSeverities` in `repositories/apiserver.go` already walks a slice, so it's correctly left untouched.
- All CI checks (build, lint, CodeQL, tests) are green.

No blockers. One non-blocking note for a future PR: the sort key is severity *name* (alphabetical), not severity *rank*, so `SeveritiesStats` won't come out Critical→...→Negligible — just deterministic. That matches the pre-existing test expectations and is called out in the PR description, so it's fine as-is here.

Approving.

PR #649, opened 2026-08-12
fix: report every singleflight waiter's own SBOM-creation failure

### LINKED ISSUE #642: getOrCreateSBOM's singleflight dedup reports SBOM-creation failures only for the leader, leaving every deduplicated waiter's scan unreported to the backend
# Description
`ScanService.getOrCreateSBOM` (`core/services/scan.go:1278`) deduplicates concurrent SBOM-creation requests for the same image via `singleflight.Group.DoChan`. When SBOM creation fails, `ReportScanFailure` — the structured, reason-classified failure report the backend/dashboard uses to know a specific workload's scan failed — is only ever called once, inside the singleflight closure, using the context of whichever caller happened to become the "leader" (the one caller whose closure singleflight actually executes):

```go
ch := s.sfGroup.DoChan(key, func() (interface{}, error) {
    ran = true
    ...
    sbom, err := s.sbomCreator.CreateSBOM(workerCtx, workload.ImageSlug, workload.ImageHash, workload.ImageTagNormalized, opts)
    ...
    if err != nil {
        reason := classifySBOMError(err)
        _ = s.platform.ReportScanFailure(workerCtx, scanfailure.ScanFailureSBOMGeneration, reason, err)
        ...
        return domain.SBOM{}, &domain.ScanError{Reason: reason, Err: err}
    }
    ...
})

select {
case res := <-ch:
    if res.Err != nil {
        return domain.SBOM{}, false, res.Err
    }
    ...
```

Every other concurrent caller — every "waiter" whose own closure never ran — only receives `res.Err` back at the `case res := <-ch:` branch. Nothing in that branch, and nothing in any of the four call sites (`GenerateSBOM`, `ScanCP`, `ScanCVE`, `ScanRegistry`), re-invokes `ReportScanFailure` (or even a lesser status signal) using the waiter's own context.

`ReportScanFailure` (`adapters/v1/backend.go:323`) reads `workload := ctx.Value(domain.WorkloadKey{})` and builds the report entirely from it — `JobID`, and the `Workloads[0]` entry's `ClusterName`/`Namespace`/`WorkloadKind`/`WorkloadName`/`ContainerName`, all derived from that specific caller's `Wlid`/`JobID`. `workerCtx` inside the closure is `context.WithoutCancel(ctx)` from the leader alone, so the report that goes out to the backend carries only the leader's workload identity — every other affected workload's own identity is never sent.

# Environment
OS: `Any`
Version: `main`

# Steps To Reproduce
1. Scale a Deployment to N ≥ 2 replicas, all running containers that share one image reference. Each pod's scan request gets its own `ScanCVE`/`ScanCP` call with a distinct `JobID`/`Wlid` — confirmed at the call sites: `core/services/scan.go:297` builds a fresh `subWorkload` with `JobID: uuid.NewString()` per container in `ScanCP`; `ScanCVE`/`ScanRegistry` each carry the requesting workload's own `JobID`/`Wlid` from the incoming command.
2. Trigger N concurrent scans of that image close enough together to land in the same singleflight window — realistic under normal operation, since these commonly originate from the same rollout/reconcile event.
3. Make the underlying `CreateSBOM` fail (registry pull error, auth failure, timeout — anything `classifySBOMError` can classify).
4. Trace the code: only the caller whose closure `DoChan` actually executes (`ran = true`) reaches the `s.platform.ReportScanFailure(workerCtx, ...)` call at line 1317. The other N-1 callers each get `res.Err` at line 1344 and return it up through their own flow without any further backend call: `ScanCVE` returns a `*domain.ScanError` with nothing else at `core/services/scan.go:465`; `ScanCP` just logs and `continue`s at `:299`; `ScanRegistry` calls the separate, lesser `ReportError` (a generic `JobFailed` status ping, not the reason-classified `ScanFailureReport`) at `:575`; `GenerateSBOM` just returns the bare error at `:198`.
5. Confirm no other layer fills the gap: `controllers/http.go`, which wraps all four flows, never calls `ReportScanFailure`/`ReportError` itself — it only records local Prometheus metrics and logs (`grep -n "ReportScanFailure\|ReportError\|SendStatus" controllers/http.go` returns only a comment, no call site).

# Expected behavior
When a shared SBOM-creation failure is delivered to N concurrent, distinct callers via singleflight, each of the N callers should result in its own `ReportScanFailure` call, carrying that caller's own workload identity (`JobID`, `Wlid`, `ContainerName`), so the backend/dashboard has a failure record for every affected workload — not just the one that happened to execute the closure.

# Actual Behavior
Only the singleflight "leader" — the one caller whose closure actually ran `CreateSBOM` — gets a `ReportScanFailure` call, using its own workload identity. The other N-1 callers' scans fail locally (visible in the kubevuln pod's own logs) but are never reported to the backend under their own `JobID`/`Wlid`. From the dashboard's perspective those N-1 scans have no completion signal at all: they don't show as failed, they simply never resolve — worse for an operator than a visible failure, since there's no error to explain why the scan never completed.

# Additional context

**Root cause:** `getOrCreateSBOM`'s singleflight deduplication was designed around sharing one result (the SBOM, or one classified error) across callers, but the failure-reporting side effect (`ReportScanFailure`) was left living inside the closure that only the leader executes, instead of being re-emitted once per distinct caller after the shared result is delivered. The dedup mechanics are correct for the SBOM itself; the per-workload observability signal was never adapted to the fact that N distinct workloads can now be sharing one execution.

**Affected components:**
- `core/services/scan.go` — `getOrCreateSBOM` is where the fix belongs: after singleflight delivers `res`, every waiter needs its own `ReportScanFailure` call using its own `ctx`/workload, not just the leader.
- `core/services/scan_test.go` — needs a regression test asserting that N concurrent `getOrCreateSBOM` callers for the same failing image each result in N `ReportScanFailure` calls (one per distinct `JobID`), not one.
- Indirectly `adapters/v1/backend.go`'s `ReportScanFailure`/`ReportError` — no code change expected there, but worth confirming the fix doesn't lean on `ReportError`'s weaker, non-reason-classified signal as a substitute (as `ScanRegistry` currently does for its own non-singleflight-related errors).

**Acceptance criteria:**
- When `getOrCreateSBOM` fails and multiple callers were deduplicated onto one singleflight execution, every caller (leader and every waiter) results in exactly one `ReportScanFailure` call carrying that caller's own `ctx`-derived workload identity.
- The leader's existing `ReportScanFailure` call is not duplicated into an extra report for the leader.
- A concurrency-based regression test (multiple goroutines calling `getOrCreateSBOM` with distinct workloads/contexts for the same image, against a mock platform recording every `ReportScanFailure` call) asserts the count and identity of reports sent equals the number of distinct callers.
- No behavior change to the success path or to single-caller (non-deduplicated) failures.

**High-level solution direction:** move (or duplicate) the `ReportScanFailure` call out of the singleflight closure so it runs once per actual caller rather than once per execution: after `res := <-ch` is received, if `res.Err != nil`, each caller invokes `s.platform.ReportScanFailure` itself using its own `ctx` and the classified reason already carried in `res.Err` (it's already a `*domain.ScanError` with `.Reason` set by `classifySBOMError`, so no reclassification is needed — just re-emit the report with the caller's own context in place of the leader's `workerCtx`). Guard against double-reporting the leader, whose report was already sent inside the closure.

bhuvan-somisetty:
## Overview

`ScanService.getOrCreateSBOM` deduplicates concurrent SBOM-creation requests for the same image via `singleflight.Group.DoChan`. When creation fails, only the caller whose closure singleflight actually executed (the "leader") got a `ReportScanFailure` call, using its own workload identity from inside the closure. Every other caller sharing that same failed result received `res.Err` back and returned it up its own call stack without any further backend call.

Since `ReportScanFailure` builds its report entirely from `ctx.Value(domain.WorkloadKey{})` (JobID, Wlid, ContainerName, etc.), those waiters' scans never got reported to the backend under their own identity. From the dashboard's perspective, N-1 of N concurrently-deduplicated scans simply never resolve — no failure, no completion signal at all.

**Root cause:** the failure-reporting side effect (`ReportScanFailure`) lived inside the singleflight closure, which only the leader executes. The dedup mechanics correctly share one SBOM/error result across callers, but the report was never re-emitted per distinct caller after that shared result is delivered.

**Fix:** after `res := <-ch` is received in `getOrCreateSBOM`, if `res.Err != nil` and this caller wasn't the leader (`!ran`), it now calls a small helper, `reportWaiterSBOMFailure`, which re-emits `ReportScanFailure` using the waiter's own `ctx` and the reason/error already classified on the shared `*domain.ScanError`. The leader's own report (sent inside the closure) is not duplicated. No change to the success path or to single-caller (non-deduplicated) failures.

## How to Test

```
go test ./core/services/... -run TestScanService_SingleflightSBOM -v
```

`TestScanService_SingleflightSBOMFailure_ReportsEveryWaiter` runs 5 concurrent goroutines with distinct `JobID`s calling `getOrCreateSBOM` for the same image against a `CreateSBOM` that blocks then fails; it asserts `CreateSBOM` is called exactly once (dedup preserved) and that all 5 distinct `JobID`s show up in the recorded `ReportScanFailure` calls.

`go build ./...`, `go vet ./core/services/...`, and `go test ./core/services/...` all pass, including the existing `TestScanService_SingleflightSBOMDeduplication`.

## Related issues/PRs:

* Resolved #642

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved failure reporting when concurrent SBOM generation requests fail.
  * Each affected scan now receives an individual failure report with the correct job details.
  * Prevented missing or incomplete error classification for non-primary requests.

### DIFF
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -1387,6 +1387,14 @@ func (s *ScanService) getOrCreateSBOM(ctx context.Context, workload domain.ScanC
 	select {
 	case res := <-ch:
 		if res.Err != nil {
+			// The leader already reported this failure with its own workload identity
+			// from inside the closure. Every other caller sharing the same result would
+			// otherwise go completely unreported to the backend under its own JobID/Wlid
+			// -- singleflight only executes one closure per key, so this is the only
+			// place a waiter's own failure report can still be sent. See #642.
+			if !ran {
+				s.reportWaiterSBOMFailure(workerCtx, res.Err)
+			}
 			return domain.SBOM{}, nil, res.Err
 		}
 		if !ran {
@@ -1397,4 +1405,20 @@ func (s *ScanService) getOrCreateSBOM(ctx context.Context, workload domain.ScanC
 	case <-ctx.Done():
 		return domain.SBOM{}, nil, ctx.Err()
 	}
+}
+
+// reportWaiterSBOMFailure re-emits ReportScanFailure for a singleflight caller whose own
+// closure never ran (a "waiter"), using the same reason/error the leader already reported
+// with its own workload identity. err is always the *domain.ScanError the closure in
+// getOrCreateSBOM returns on failure; the classifySBOMError fallback only guards against
+// that invariant changing.
+func (s *ScanService) reportWaiterSBOMFailure(ctx context.Context, err error) {
+	reason := classifySBOMError(err)
+	reportErr := err
+	var scanErr *domain.ScanError
+	if errors.As(err, &scanErr) {
+		reason = scanErr.Reason
+		reportErr = scanErr.Err
+	}
+	_ = s.platform.ReportScanFailure(ctx, scanfailure.ScanFailureSBOMGeneration, reason, reportErr)
 }
--- a/core/services/scan_test.go
+++ b/core/services/scan_test.go
@@ -2916,6 +2916,15 @@ func (p *failureRecordingPlatform) ReportScanFailure(ctx context.Context, failur
 // each carries a distinct JobID/workload identity the backend/dashboard needs to know that
 // specific scan failed.
 func TestScanService_SingleflightSBOMFailure_ReportsEveryWaiter(t *testing.T) {
+	// Pin the scheduler to one P so the goroutines started below cannot run concurrently with
+	// each other or with this goroutine. Under a single P, a goroutine only yields the
+	// processor by blocking (or explicitly calling runtime.Gosched); none of the code between a
+	// waiter's start and its blocking receive on the singleflight result channel does either, so
+	// each waiter runs start-to-block in one uninterrupted turn. That makes startWg.Wait below a
+	// hard guarantee that every waiter has already joined the shared singleflight.DoChan call by
+	// the time it returns, rather than a timing-based approximation of it.
+	defer runtime.GOMAXPROCS(runtime.GOMAXPROCS(1))
+
 	started := make(chan struct{})
 	release := make(chan struct{})
 	var once sync.Once
@@ -2966,18 +2975,12 @@ func TestScanService_SingleflightSBOMFailure_ReportsEveryWaiter(t *testing.T) {
 		}()
 	}
 
-	startWg.Wait() // ensure all goroutines are running before unblocking the singleflight worker
-	<-started      // wait for the singleflight worker to enter CreateSBOM
-
-	// startWg only proves each goroutine began running, not that it has reached its
-	// singleflight.DoChan call yet -- give the scheduler a generous window to actually get
-	// every waiter registered on the same key before the leader's CreateSBOM is allowed to
-	// return, otherwise a slow-to-schedule waiter can miss the shared call entirely and
-	// trigger its own (spurious) second CreateSBOM execution.
-	for i := 0; i < 1000; i++ {
-		runtime.Gosched()
-	}
-	time.Sleep(50 * time.Millisecond)
+	// With GOMAXPROCS(1) above, this Wait returning already proves every waiter has joined the
+	// shared singleflight.DoChan call: the last one to call startWg.Done can only yield the
+	// processor by blocking on the result channel, so it -- and by the same argument every
+	// waiter before it -- is already parked on <-ch by the time this goroutine runs again.
+	startWg.Wait()
+	<-started // wait for the singleflight worker to enter CreateSBOM
 
 	close(release) // let CreateSBOM return its error
 	wg.Wait()
--- a/core/services/scan_test.go
+++ b/core/services/scan_test.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"os"
+	"runtime"
 	"sync"
 	"sync/atomic"
 	"testing"
@@ -2757,15 +2758,20 @@ type blockingSBOMCreator struct {
 	ports.SBOMCreator
 	calls   int
 	onStart func()
+	failErr error
 	mu      sync.Mutex
 }
 
 func (b *blockingSBOMCreator) CreateSBOM(ctx context.Context, name, imageID, imageTag string, options domain.RegistryOptions) (domain.SBOM, error) {
 	b.mu.Lock()
 	b.calls++
+	failErr := b.failErr
 	b.mu.Unlock()
 	if b.onStart != nil {
 		b.onStart()
+	}
+	if failErr != nil {
+		return domain.SBOM{}, failErr
 	}
 	return b.SBOMCreator.CreateSBOM(ctx, name, imageID, imageTag, options)
 }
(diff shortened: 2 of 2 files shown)


matthyx (changes requested):
## Review

**The fix logic itself looks correct.** `workerCtx` is a local variable derived from each caller's own `ctx` (`context.WithoutCancel(ctx)`), so `reportWaiterSBOMFailure(workerCtx, res.Err)` reports under *that waiter's own* `WorkloadKey` identity, not the leader's — which is exactly what's needed to fix #642. `ran` is only read after `res := <-ch`, which happens-after the singleflight closure returns, so there's no data race, and the leader (`ran == true`) is correctly skipped to avoid double-reporting. The new regression test (`TestScanService_SingleflightSBOMFailure_ReportsEveryWaiter`) exercises this well.

### Blocker: branch is stale and conflicts with current `main`

This PR is 5 commits behind `main` (merge-base is `746a949`, PR HEAD is `9f22f3b`), and GitHub reports `mergeable: CONFLICTING`. More importantly, `main` has since refactored the exact function this PR patches — `getOrCreateSBOM` on `main` now:

- returns `(domain.SBOM, error, error)` instead of `(domain.SBOM, bool, error)` (second value is a store error, not a "created" bool),
- wraps the singleflight result in a new `sbomCreation{sbom, storeErr}` struct,
- adds `TooLarge` status persistence in the store step,
- and its `res.Err != nil` branch is now just `return domain.SBOM{}, nil, res.Err}` — i.e. **the leader-only-report bug this PR fixes is still present, unfixed, on current `main`**.

So this isn't a trivial whitespace conflict — a naive merge would either fail outright or (if force-merged) silently drop the `sbomCreation`/`TooLarge` refactor. Please rebase onto current `main` and reapply `reportWaiterSBOMFailure` inside the new `res.Err != nil` branch (and the corresponding test) against the current shape of `getOrCreateSBOM`.

### Nit (non-blocking)

`TestScanService_SingleflightSBOMFailure_ReportsEveryWaiter` uses `for i := 0; i < 1000; i++ { runtime.Gosched() }` plus a `time.Sleep(50 * time.Millisecond)` to make sure every waiter has registered on the singleflight key before releasing the leader. This is a reasonable belt-and-suspenders approach but is inherently timing-based and a candidate for flakiness under CI load — worth a second look if this test ever flakes, though not a reason to block on its own.

Once rebased onto `main` and the merge conflict is resolved, this looks good to merge.

bhuvan-somisetty:
Rebased onto current main and resolved the conflict — reapplied the waiter-report fix on top of #623's getOrCreateSBOM refactor (the new (domain.SBOM, error, error) signature and sbomCreation struct), so the leader-only-report bug you flagged is fixed on the current shape of the function too. Build, vet and the full test suite (including the new regression test) are all green. Ready for another look whenever you get a chance!

bhuvan-somisetty:
Addressed the coderabbitai nit on the regression test too — swapped the `runtime.Gosched`/`time.Sleep` scheduler-delay hack for `GOMAXPROCS(1)`, so `startWg.Wait` is now a hard guarantee that every waiter has joined the shared singleflight call before `release` closes, not a timing-based approximation. Verified with `go vet` and 30 repeated runs of the test (each now finishing in under a millisecond vs. ~50ms before). Full suite is green.

matthyx (approved):
## Re-review

Both points from my previous "request changes" review are resolved:

1. **Merge conflict** — the branch has been rebased; `mergeable` is now `MERGEABLE` and a local 3-way merge against current `main` auto-resolves cleanly (`Automatic merge went well`). The fix (`reportWaiterSBOMFailure` called from the `res.Err != nil` branch, using each waiter's own `workerCtx`) is correctly reapplied against the current shape of `getOrCreateSBOM`.
2. **Flaky test nit** — the `runtime.Gosched()`/`time.Sleep` hack in `TestScanService_SingleflightSBOMFailure_ReportsEveryWaiter` was replaced with `GOMAXPROCS(1)` + `startWg.Wait()`, which gives a real happens-before guarantee (every waiter must block on `<-ch` before the last `startWg.Done()` caller can hand control back) instead of a timing-based approximation. Good fix.

All CI checks pass. Approving.

PR #651, opened 2026-08-12
fix(adapters): normalize trusted vendor slugs before matching distros

### LINKED ISSUE #650: a trusted vendor slug only works in exactly one spelling, others are silently ignored
a trusted vendor is only trusted if it is written in exactly the form Grype uses. anything else is accepted, stored, and never matches:

```
configured "wolfi"   -> trusted=true
configured "Wolfi"   -> trusted=false
configured "WOLFI"   -> trusted=false
configured " wolfi"  -> trusted=false
configured "wolfi "  -> trusted=false
```

`buildTrustedVendorSet` converts each slug straight to a `distro.Type`:

```go
func buildTrustedVendorSet(vendors []string) map[distro.Type]bool {
	set := make(map[distro.Type]bool, len(vendors))
	for _, v := range vendors {
		set[distro.Type(v)] = true
	}
	return set
}
```

Grype's types are lowercase slugs (`"wolfi"`, `"chainguard"`, `"redhat"`) and `resolveUseDefaultMatchers` looks them up verbatim, so `distro.Type("Wolfi")` is a key nothing ever matches. the effect is quiet: adaptive mode is what turns CPE matching off for a trusted vendor, so the image keeps being matched by CPE and reporting the false positives the setting exists to remove, with nothing logged.

whitespace is the same story. #593 trims the env form of the value, but the config file path goes through `GetStringSlice`, so `"trustedVendors": ["wolfi "]` is untrimmed.

and a slug that is not a Grype distro at all, a typo like `"redhat-enterprise"`, behaves identically: accepted, never matches, nothing said. `distro.All` lists all 24 types, and the four defaults (`echo`, `chainguard`, `wolfi`, `minimos`) are all in it, so an unrecognised one can be called out without the defaults tripping it.

lowercasing and trimming the slug, and logging one that Grype does not know, is enough.

version: main (746a949)

AnshulPatil2005:
## Overview

`buildTrustedVendorSet` converted each configured slug straight to a `distro.Type`. Grype's types are lowercase slugs it compares verbatim, so `"Wolfi"`, `"WOLFI"` or `"wolfi "` went into the set as keys no distro ever matches and the vendor stayed untrusted.

the failure is silent, and it undoes the setting: adaptive mode is what turns CPE matching off for a trusted vendor, so the image keeps being matched by CPE and reporting the false positives the setting exists to remove, with nothing logged to say the slug did not take.

slugs are now lowercased and trimmed, so all of those forms resolve to the same vendor. a slug that is not one of Grype's distros is also logged, since it fails in exactly the same quiet way. it is kept in the set rather than dropped: it can never match anyway, and dropping it would replace one silent behaviour with another.

## Additional Information

whitespace mattered on the config file path in particular. #593 trims the env form of the value, but the file form goes through `GetStringSlice` untouched, so `"trustedVendors": ["wolfi "]` arrived with the space attached.

the warning is safe against the defaults: `distro.All` holds all 24 types and `echo`, `chainguard`, `wolfi` and `minimos` are all among them, so the out-of-the-box configuration logs nothing.

## How to Test

`go test ./adapters/v1/ -run TestBuildTrustedVendorSet -count=1`

- `TestBuildTrustedVendorSet_NormalizesSlugs` resolves each spelling of a vendor against a wolfi image and checks whether it comes out trusted, including a different vendor and a slug that is not a distro, so it pins what must not match as well.
- `TestBuildTrustedVendorSet_SkipsEmpty` covers an empty entry, which would otherwise become a key matching a distro with no type.

four of the seven cases fail before this: capitalised, upper case, leading space and trailing space.

## Related issues/PRs:

* Resolved #650

## Summary by CodeRabbit

* **Bug Fixes**
  * Trusted vendor settings now consistently handle capitalization and surrounding whitespace.
  * Empty vendor entries are ignored.
  * Unrecognized vendor values generate a warning while preserving non-empty entries for compatibility.

### DIFF
--- a/adapters/v1/grype.go
+++ b/adapters/v1/grype.go
@@ -97,10 +97,32 @@ func NewGrypeAdapter(listingURL string, matchingMode config.CVEMatchingMode, tru
 }
 
 // buildTrustedVendorSet maps configured vendor slugs to Grype distro types.
+//
+// Grype's distro types are lowercase slugs that it compares verbatim, so a configured slug
+// is normalised rather than taken as written: "Wolfi", or one with the whitespace a JSON
+// list easily carries, would otherwise sit in the set as a key no distro ever matches, and
+// the vendor would stay untrusted with nothing to say so. A slug that is not one of Grype's
+// types cannot match either, so it is called out for the same reason: adaptive mode is meant
+// to turn CPE matching off for these vendors, and a typo silently leaves it on.
 func buildTrustedVendorSet(vendors []string) map[distro.Type]bool {
+	known := make(map[distro.Type]bool, len(distro.All))
+	for _, t := range distro.All {
+		known[t] = true
+	}
+
 	set := make(map[distro.Type]bool, len(vendors))
 	for _, v := range vendors {
-		set[distro.Type(v)] = true
+		slug := distro.Type(strings.ToLower(strings.TrimSpace(v)))
+		if slug == "" {
+			continue
+		}
+		if !known[slug] {
+			// Kept in the set regardless: it can never match, and dropping it would be a
+			// second silent behaviour rather than a reported one.
+			logger.L().Warning("trusted vendor is not a distro Grype recognizes, it will never match",
+				helpers.String("vendor", v))
+		}
+		set[slug] = true
 	}
 	return set
 }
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed the diff, checked the fix against the vendored grype v0.99.1 distro package (confirmed distro.Type values and distro.All are exactly the 24 lowercase slugs the PR describes, including echo/minimos), and ran the full adapters/v1 suite plus `go build ./...` and `go vet` locally — all green. The normalize-then-warn-but-keep approach in buildTrustedVendorSet is sound: it fixes the case/whitespace mismatch from #650, correctly reports unmatchable slugs without dropping them (avoiding a second silent failure mode), and the new table-driven tests cover the reported repro cases (canonical/capitalised/upper/leading-trailing space) plus the empty-entry and unknown-distro edges. No blockers.

PR #653, opened 2026-08-12
fix(repositories): invalidate SecurityException cache on CRD changes

### LINKED ISSUE #652: Invalidate SecurityException caches when CRDs change
# Description

`GetSecurityExceptions` in `repositories/apiserver.go` currently caches raw `SecurityException` and `ClusterSecurityException` `List()` results by TTL only. That avoids repeated API calls between scans, but it leaves a stale window where scans can continue using outdated suppression data until the cache entry expires.

This is visible in the current shared cache added for `#510`: namespaced exception lists are cached under `se/<namespace>` and cluster-scoped exceptions under `cse`, each with a 30 second TTL. There is no informer or watch path that evicts those cache keys when the backing CRDs change.

# Environment
OS: `Not version-specific`
Version: `main as of 2026-08-12`
 
# Steps To Reproduce
1. Run kubevuln with SecurityException support enabled.
2. Trigger a scan in a namespace so `GetSecurityExceptions` populates the `securityExceptionListCache` entries for that namespace and for the cluster-scoped exception list.
3. Update or delete a `SecurityException` or `ClusterSecurityException` CRD before the 30 second TTL expires.
4. Trigger another scan that reuses the same namespace or cluster-scoped cache entry.

# Expected behavior
The relevant cached exception list should be evicted as soon as the CRD changes so the next scan re-lists fresh data immediately.

# Actual Behavior
The cached list remains valid until TTL expiry, so scans can continue applying stale suppression data for up to 30 seconds after an exception update or deletion.

# Additional context
Relevant code and tests:
- `repositories/apiserver.go`: `listSecurityExceptions`, `listClusterSecurityExceptions`, `securityExceptionListCache`
- `repositories/apiserver_test.go`: `TestAPIServerStore_GetSecurityExceptions_CachesAcrossWorkloads`

Scope of the proposed change:
- Keep the existing shared list cache to collapse repeated scans between change events.
- Add optional watch or informer invalidation for `SecurityException` and `ClusterSecurityException` changes.
- Evict `se/<namespace>` for namespaced events and `cse` for cluster-scoped events on add, update, and delete.

Acceptance criteria:
- Updating or deleting an exception invalidates the relevant cache entry promptly.
- Existing list caching still avoids repeated API calls between change events.
- Regression tests cover namespaced and cluster-scoped invalidation.

rootp1:
## Overview
This wires a best effort informer based invalidation path into `APIServerStore` so the existing shared `SecurityException` list cache is evicted as soon as the backing CRDs change instead of waiting for the 30 second TTL to expire.

The fetch path is unchanged. Repeated scans still reuse the existing cached `List()` results between change events, but namespaced cache keys are now evicted on `SecurityException` add, update, and delete events, and the cluster wide cache key is evicted on `ClusterSecurityException` add, update, and delete events.

## Additional Information
This stays scoped to the exception fetch cache in `repositories/apiserver.go` and adds focused regression coverage in `repositories/apiserver_test.go` for namespaced and cluster scoped invalidation. No unrelated files were changed.

## How to Test
1. `go test ./repositories/...`
2. `make verify`

## Related issues/PRs:
Resolved #652

## Summary by CodeRabbit

* **Bug Fixes**
  * Security exception data now refreshes automatically when relevant Kubernetes resources are created, updated, or deleted.
  * Changes to namespaced resources refresh only the affected namespace, while cluster-wide changes refresh the appropriate cluster-level data.
  * Unexpected resource events now trigger a broader cache refresh to prevent stale results.

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -29,14 +29,17 @@ import (
 	"golang.org/x/mod/semver"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/apimachinery/pkg/util/managedfields"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/dynamic/dynamicinformer"
 	fakedynamic "k8s.io/client-go/dynamic/fake"
 	k8stesting "k8s.io/client-go/testing"
+	k8scache "k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/retry"
 )
 
@@ -77,6 +80,8 @@ type APIServerStore struct {
 	// nil is safe (falls back to always listing, e.g. in tests that construct APIServerStore
 	// literals directly instead of through the constructors below).
 	securityExceptionListCache *cache.Cache
+
+	securityExceptionInformerStop context.CancelFunc
 }
 
 var (
@@ -128,12 +133,14 @@ func NewAPIServerStorage(namespace string) (*APIServerStore, error) {
 	if err != nil {
 		return nil, err
 	}
-	return &APIServerStore{
+	store := &APIServerStore{
 		StorageClient:              clientset.SpdxV1beta1(),
 		DynamicClient:              dynClient,
 		Namespace:                  namespace,
 		securityExceptionListCache: cache.New(securityExceptionListCacheCleaningInterval),
-	}, nil
+	}
+	store.enableSecurityExceptionCacheInvalidation(context.Background())
+	return store, nil
 }
 
 type fakeStorageClientset struct {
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed the diff, ran `go build ./...`, `go vet ./repositories/...`, and `go test ./repositories/... -race` locally — all clean, and CI is green.

The change is well-scoped: it only touches the exception fetch cache path in `repositories/apiserver.go`, keeps the fetch path itself unchanged, and adds solid regression tests for both the namespaced and cluster-scoped eviction paths.

Two non-blocking points CodeRabbit already flagged that are worth a follow-up rather than blocking this PR, since neither regresses behavior below the pre-existing 30s-TTL staleness:

1. **Informer lifecycle**: `enableSecurityExceptionCacheInvalidation` is wired to `context.Background()` in `NewAPIServerStorage`, and `securityExceptionInformerStop` is never called anywhere, so the informer goroutine has no way to stop on shutdown. In practice this is a non-issue today since the store is a single long-lived instance created once in `cmd/http/main.go`, but tying it to an owner context (and adding a `Close`/`Stop` on `APIServerStore`) would make the lifecycle explicit.
2. **TOCTOU on cache population**: `listSecurityExceptions`/`listClusterSecurityExceptions` populate the cache with `Set()` after their `List()` call returns, with no ordering guarantee against a concurrent informer-triggered `Delete()`. If a change event fires while a `List()` for that same key is in flight, the eventual `Set()` can silently repopulate the key with pre-change data for the remainder of the TTL. Given the PR description already frames this as a "best effort" invalidation path, and the worst case is identical to current production behavior, I don't think this needs to hold up the merge — but a per-key generation/version check (bump on invalidate, only `Set()` if the generation hasn't changed since the `List()` started) would close it cleanly if you want to follow up.

Also worth a quick sanity check outside this diff: the informer now needs `watch` (not just `list`/`get`) RBAC on `securityexceptions`/`clustersecurityexceptions` cluster-wide. If that's not already granted in the deployment manifests (not in this repo), the invalidation path will silently no-op and fall back to the existing TTL behavior — worth confirming before/soon after rollout.

None of the above are regressions or block correctness of the fetch path, so approving.

PR #657, opened 2026-08-12
fix: preserve SecurityException VEX semantics

### LINKED ISSUE #656: bug: generated VEX flattens SecurityException outcomes to not_affected
# Description
Generated VEX statements for `SecurityException`-suppressed findings currently flatten every ignored finding into the same local `not_affected` story. The writer overwrites the originating suppression semantics and emits `status: not_affected`, `justification: vulnerable_code_not_present`, and one of two hard-coded impact strings even when the matched `SecurityException` expressed a different outcome.

This is distinct from the earlier `IgnoredMatches` omission bug. The suppressed finding is present in the VEX document now, but its semantic meaning is no longer trustworthy as an audit artifact.

# Environment
OS: `darwin/arm64`
Version: `go1.26.5` workspace run on Wednesday, August 12, 2026`
 
# Steps To Reproduce
1. Confirm the stored ignore-rule schema already preserves suppression provenance, including `Justification` and `ImpactStatement`:
   ```bash
   go doc github.com/kubescape/storage/pkg/apis/softwarecomposition/v1beta1.IgnoreRule
   ```
2. Inspect the local VEX rewrite path in `repositories/apiserver.go`:
   ```bash
   nl -ba repositories/apiserver.go | sed -n '1133,1156p'
   nl -ba repositories/apiserver.go | sed -n '1305,1326p'
   ```
3. Inspect the richer `SecurityException` vulnerability states and the suppression conversion path:
   ```bash
   nl -ba pkg/securityexception/v1beta1/types.go | sed -n '39,117p'
   nl -ba adapters/v1/securityexception.go | sed -n '220,250p'
   ```
4. Observe that ignored local statements are always rewritten to `StatusNotAffected`, `VulnerableCodeNotPresent`, and a hard-coded impact string regardless of the original exception semantics.

# Expected behavior
Generated VEX should preserve the effective suppression semantics that produced the ignored finding, or explicitly encode that kubevuln is publishing a local policy decision instead of silently rewriting it into `not_affected`.

Examples that should remain distinguishable:
- `SecurityException` entries declaring `status: fixed`
- accepted-risk `status: affected` suppressions such as `will_not_fix` or `can_not_fix`
- custom justification and impact text already carried on the stored ignore-rule provenance

# Actual Behavior
The VEX writer rewrites ignored matches into a generic `not_affected` representation during both initial creation and the update reset path.

Relevant code paths today:
- `repositories/apiserver.go:1143-1146`
- `repositories/apiserver.go:1312-1326`

# Additional context
Technical evidence:
- `pkg/securityexception/v1beta1/types.go` supports materially different vulnerability outcomes: `not_affected`, `fixed`, `affected`, and `under_investigation`.
- `adapters/v1/securityexception.go` already preserves suppression provenance such as `justification`, `impactStatement`, `actionStatement`, and `response` while converting matched `SecurityException` entries.
- `repositories/apiserver.go` currently ignores that richer provenance and flattens every suppressed finding to the same `not_affected` output.

Likely root cause:
`StoreVEX` only consults `IgnoredMatches` and then applies hard-coded rewrite rules, instead of reconstructing the effective suppression semantics from the ignore-rule provenance attached to those ignored matches.

Suggested direction:
- Thread the effective suppression semantics from `AppliedIgnoreRules` into VEX generation and update logic.
- Preserve `not_affected`, `fixed`, and accepted-risk `affected` cases instead of treating every suppression as `not_affected`.
- Keep the existing local-only fallback behavior for manifests that do not carry ignore-rule provenance.

Acceptance criteria:
- A `SecurityException` with custom VEX-style rationale is reflected in the stored VEX output.
- Regression tests cover `not_affected`, `fixed`, and accepted-risk `affected` suppressions.
- Existing local-only VEX behavior stays backward compatible for manifests without ignore-rule provenance.

rootp1:
## Overview
This change preserves the suppression semantics that come from `SecurityException` CRDs when kubevuln writes local VEX statements.

Current behavior: any ignored finding is rewritten into the same local `not_affected` statement with a hard coded justification and impact string.
Future behavior: ignored findings keep the effective suppression outcome that caused them to be filtered. `not_affected` keeps its stored justification and impact text, `fixed` is emitted as `fixed`, and accepted risk `affected` suppressions are emitted as `affected` with explicit local policy context.

## Additional Information
The fix is split into two parts:
1. `adapters/v1/securityexception.go` now carries the originating `SecurityException` status into the stored ignore rule provenance.
2. `repositories/apiserver.go` now derives ignored statement semantics from that provenance during create, update, and reset flows, while preserving the previous generic fallback when no exception provenance is available.

Focused regression coverage was added for `not_affected`, `fixed`, and accepted risk `affected` suppressions, and the update path is exercised as well.

## How to Test
1. Run `go test ./adapters/v1/... ./repositories/...`
2. Run `make verify`
3. Optionally inspect a generated `OpenVulnerabilityExchangeContainer` after applying a `SecurityException` with `not_affected`, `fixed`, or accepted risk `affected` semantics and confirm the stored statement preserves that meaning.

## Related issues/PRs:
* Resolved #656

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -1071,6 +1071,95 @@ func statementHasPURL(products []v1beta1.Product, purl string) bool {
 // more specific remediation string.
 const defaultActionStatement = "Upgrade the vulnerable component to a version that is not affected"
 
+const (
+	defaultLocalImpactStatement         = "Vulnerable component is not loaded into the memory"
+	securityExceptionImpactStatement    = "Vulnerability was ignored by a SecurityException"
+	securityExceptionAcceptedRiskAction = "A SecurityException accepted this vulnerability as an affected finding"
+)
+
+type ignoredVEXAssessment struct {
+	status          v1beta1.Status
+	justification   v1beta1.Justification
+	impactStatement string
+	actionStatement string
+	statusNotes     string
+}
+
+func defaultIgnoredVEXAssessment() ignoredVEXAssessment {
+	return ignoredVEXAssessment{
+		status:          v1beta1.Status(vex.StatusNotAffected),
+		justification:   v1beta1.Justification(vex.VulnerableCodeNotPresent),
+		impactStatement: securityExceptionImpactStatement,
+	}
+}
+
+func ignoredMatchAssessment(m v1beta1.IgnoredMatch) ignoredVEXAssessment {
+	assessment := defaultIgnoredVEXAssessment()
+	rule, ok := securityExceptionIgnoreRule(m)
+	if !ok {
+		return assessment
+	}
+
+	switch strings.TrimSpace(rule.FixState) {
+	case string(sev1beta1.VulnerabilityStatusFixed):
+		assessment.status = v1beta1.Status(sev1beta1.VulnerabilityStatusFixed)
+		assessment.justification = ""
+		assessment.impactStatement = ""
+		assessment.statusNotes = ignoredMatchStatusNotes(rule)
+	case string(sev1beta1.VulnerabilityStatusAffected):
+		assessment.status = v1beta1.Status(vex.StatusAffected)
+		assessment.justification = ""
+		assessment.impactStatement = ""
+		assessment.actionStatement = securityExceptionAcceptedRiskAction
+		assessment.statusNotes = ignoredMatchStatusNotes(rule)
+	case "", string(sev1beta1.VulnerabilityStatusNotAffected):
+		if j := strings.TrimSpace(rule.Justification); j != "" {
+			assessment.justification = v1beta1.Justification(j)
+		}
+		if impact := strings.TrimSpace(rule.ImpactStatement); impact != "" {
+			assessment.impactStatement = impact
+		}
+	default:
+		if j := strings.TrimSpace(rule.Justification); j != "" {
+			assessment.justification = v1beta1.Justification(j)
+		}
+		if impact := strings.TrimSpace(rule.ImpactStatement); impact != "" {
+			assessment.impactStatement = impact
+		}
+	}
+
+	return assessment
+}
+
+func securityExceptionIgnoreRule(m v1beta1.IgnoredMatch) (v1beta1.IgnoreRule, bool) {
+	for _, rule := range m.AppliedIgnoreRules {
+		switch rule.SourceKind {
+		case "SecurityException", "ClusterSecurityException":
+			return rule, true
+		}
+	}
+	return v1beta1.IgnoreRule{}, false
+}
+
+func ignoredMatchStatusNotes(rule v1beta1.IgnoreRule) string {
+	parts := make([]string, 0, 2)
+	if j := strings.TrimSpace(rule.Justification); j != "" {
+		parts = append(parts, "justification: "+j)
+	}
+	if impact := strings.TrimSpace(rule.ImpactStatement); impact != "" {
+		parts = append(parts, "impact: "+impact)
+	}
+	return strings.Join(parts, "; ")
+}
+
+func applyIgnoredMatchAssessment(stmt *v1beta1.Statement, assessment ignoredVEXAssessment) {
+	stmt.Status = assessment.status
+	stmt.Justification = assessment.justification
+	stmt.ImpactStatement = assessment.impactStatement
+	stmt.ActionStatement = assessment.actionStatement
+	stmt.StatusNotes = assessment.statusNotes
+}
+
 // buildActionStatement returns a remediation string for an affected VEX statement. When the
 // match reports a fixed state with known fix versions, it names them; otherwise it falls back
 // to defaultActionStatement.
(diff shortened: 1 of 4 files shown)


repositories/apiserver.go
```diff
@@ -1141,9 +1229,7 @@ func markIgnoredVulnerabilitiesInVex(vexDoc *v1beta1.VEX, cve *domain.CVEManifes
 					for _, sc := range p.Subcomponents {
 						if sc.ID == v.Artifact.PURL {
 							vexDoc.Statements[i].Status = v1beta1.Status(vex.StatusNotAffected)
```
matthyx: Nit (non-blocking): this explicit `Status = NotAffected` assignment is now dead code — `applyIgnoredMatchAssessment` on the next line always overwrites `.Status` from `assessment.status`, which can be `fixed` or `affected`, not just `not_affected`. Safe to drop this line.

repositories/apiserver.go
```diff
@@ -957,6 +957,95 @@ func createProductStructForImageAndPackage(imagePullable string, packagePURL str
...
+		assessment.actionStatement = securityExceptionAcceptedRiskAction
+		assessment.statusNotes = ignoredMatchStatusNotes(rule)
+	case "", string(sev1beta1.VulnerabilityStatusNotAffected):
+		if j := strings.TrimSpace(rule.Justification); j != "" {
+			assessment.justification = v1beta1.Justification(j)
+		}
+		if impact := strings.TrimSpace(rule.ImpactStatement); impact != "" {
+			assessment.impactStatement = impact
+		}
+	default:
```
matthyx, line 1115: Nit (non-blocking): `default:` here does exactly the same thing as `case "", string(sev1beta1.VulnerabilityStatusNotAffected):` above — could be merged into one case list for clarity (`case "", string(sev1beta1.VulnerabilityStatusNotAffected), default:` isn't valid Go, but `case "", string(...NotAffected):` followed by removing the separate `default:` block, or vice versa, would read a bit cleaner). Not a bug, since any unrecognized status intentionally falls back to the safe not_affected-shaped assessment.

matthyx (dismissed):
Reviewed the diff, the full `repositories/apiserver.go` and `adapters/v1/securityexception.go` files at the PR head, and the new regression test.

**What it does:** replaces the hardcoded "not_affected / VulnerableCodeNotPresent / ignored by a SecurityException" statement that used to be written for every ignored match with one derived from the `SecurityException`'s actual status (`not_affected` keeps its stored justification/impact, `fixed` emits `fixed`, accepted-risk `affected` emits `affected` with an explicit action statement), via the new `ignoredVEXAssessment`/`ignoredMatchAssessment` helpers applied consistently in `createVEX`, `updateVEX`'s append paths, and its "reset to baseline" pass.

**Correctness check:**
- `isExceptionSourcedIgnore` was correctly updated alongside `buildIgnoreRule` now populating `FixState` — without that change, `RestoreSuppressedMatches` would stop recognizing most exception-sourced ignores as restorable (since `FixState` is no longer reliably empty). Good catch bundling that fix into the same PR.
- The new `TestAPIServerStore_storeVEX_preservesSecurityExceptionSemantics` test exercises all three status branches (not_affected/fixed/affected) through both `createVEX` and `updateVEX` (calling `StoreVEX` twice), which is exactly the path that used to regress the semantics on update. Good coverage.
- `IgnoreRule.FixState` (from `kubescape/storage`) already has an established meaning elsewhere in this repo — Grype's native fix state ("fixed"/"not-fixed", see `adapters/v1/grype_to_domain.go`/`fixstate.go`). This PR reuses the same field to carry the SecurityException status vocabulary instead. It's functionally safe today because every read is gated on `SourceKind == SecurityException/ClusterSecurityException` (native grype ignore rules never set `SourceKind`), but it's worth a short comment somewhere near `ignoredMatchAssessment`/`buildIgnoreRule` noting the field is intentionally repurposed for exception-sourced rules only, so a future reader doesn't assume it's always Grype's vocabulary.
- CI (build, lint, CodeQL, GitGuardian, DCO, tests) is all green.

Left three non-blocking nits inline (dead status assignment in `markIgnoredVulnerabilitiesInVex`, a mergeable duplicate switch branch in `ignoredMatchAssessment`, and a stale doc comment on `isExceptionSourcedIgnore` that a diff-scoped review comment couldn't attach to — it's the comment directly above the function, describing the old "only the vulnerability ID is set" signature that no longer matches the updated logic).

No blockers — approving.

matthyx:
needs rebase

rootp1:
@matthyx PTAL

matthyx (approved):
Re-checked after the new commit (`5c2436b` "fix: address review nits on SecurityException VEX semantics"). All three non-blocking nits from my earlier review are addressed:

- Dead `Status = NotAffected` assignment in `markIgnoredVulnerabilitiesInVex` removed.
- The duplicate `fixed`/`affected`/not-affected-fallback switch branches in `ignoredMatchAssessment` are merged into a single `default` case with a clarifying comment.
- The stale doc comment on `isExceptionSourcedIgnore` is updated to match the actual logic.
- Added a clarifying comment in `buildIgnoreRule` on why `FixState` is repurposed to carry the SecurityException status vocabulary for exception-sourced rules only.

No new issues introduced. CI (build, lint, CodeQL, GitGuardian, DCO, tests) is all green. Re-approving.

PR #659, opened 2026-08-12
fix(adapters): key relevancy on the package version too, not just its name

### LINKED ISSUE #658: one loaded version of a library marks every version of it relevant
`markRelevantVulnerabilities` decides relevancy on the CVE plus the package name:

```go
cvepIndices := make(map[string]struct{}, len(relevantVulnerabilities))
for _, v := range relevantVulnerabilities {
	cvepIndices[v.Name+"+"+v.RelatedPackageName] = struct{}{}
}
for i, v := range vulnerabilities {
	_, isRelevant := cvepIndices[v.Name+"+"+v.RelatedPackageName]
	vulnerabilities[i].IsRelevant = &isRelevant
}
```

a package name is not unique within an image though. two versions of one library sitting side by side is ordinary, particularly in Java and Node images, and `DomainToArmo` emits a record per (vulnerability, artifact) pair, so both versions produce a record carrying the same `RelatedPackageName`.

when that happens, one loaded version marks every version relevant:

```
vulnerabilities:  CVE-2024-0001 commons-collections 3.2.1
                  CVE-2024-0001 commons-collections 4.4
relevant (CVEp):  CVE-2024-0001 commons-collections 4.4

result:           both marked IsRelevant=true
```

which is the opposite of what relevancy is for. the point of the CVEp scan is to say which findings are on code that actually runs, and here a finding on a version that was never loaded is promoted because a different version of it was. it also feeds `RelevantCount` in the summary, so the count is inflated too.

the comment on the function says the pair "is the record identity" and cites the uniqueness assertion in `backend_test.go`, but that assertion checks the nginx fixture has no duplicate pairs, which is a property of that fixture rather than of the data in general.

the version is on the record already as `PackageVersion`, and both manifests are built from the same artifacts (the filtered one from a subset of the same SBOM), so the two sides carry the same strings and adding it to the key is enough.

version: main (746a949)

AnshulPatil2005:
## Overview

`markRelevantVulnerabilities` keyed relevancy on the CVE and the package name. a package name is not unique within an image: two versions of one library sitting side by side is ordinary, particularly in Java and Node images, and `DomainToArmo` emits a record per (vulnerability, artifact) pair, so both versions produce a record under the same name.

one loaded version then marked every version relevant, which is the opposite of what the CVEp scan is for. a finding on a version that was never loaded got promoted because a different version of it was, and `RelevantCount` in the summary was inflated to match.

the version is already on the record as `PackageVersion`, and both manifests come from the same artifacts (the filtered one from a subset of the same SBOM), so the two sides carry the same strings.

## Additional Information

the key is a small struct rather than a longer concatenated string, so no separator has to be chosen that cannot occur in a package name. npm scoped names contain `@`, which is the obvious separator to reach for.

the function's comment cited the uniqueness assertion in `backend_test.go` as grounds for the old key. that assertion checks the nginx fixture contains no duplicate (CVE, package name) pairs, which holds for that fixture rather than in general, so the comment now says what actually makes two records the same finding.

## How to Test

`go test ./adapters/v1/ -run TestMarkRelevantVulnerabilities -count=1`

`TestMarkRelevantVulnerabilities_DistinguishesPackageVersions` puts two versions of one library in the full manifest, marks only one of them loaded, and asserts the other stays irrelevant.

before this it fails on the unloaded copy:

```
IsRelevant mismatch for CVE-2024-0001+commons-collections@3.2.1
expected: false
actual  : true
```

`TestMarkRelevantVulnerabilities` is unchanged and still passes: its cases use one version per package, which is the case that already worked.

## Related issues/PRs:

* Resolved #658

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved vulnerability relevance matching by distinguishing package versions.
  * Prevented vulnerabilities affecting different versions of the same package from being incorrectly treated as identical.

* **Tests**
  * Added regression coverage for package-version-specific relevance matching.

### DIFF
--- a/adapters/v1/backend.go
+++ b/adapters/v1/backend.go
@@ -583,18 +583,35 @@ func httpPostDebug(httpClient httputils.IHttpClient, fullURL string, headers map
 	return httputils.HttpPostWithContext(context.Background(), httpClient, fullURL, headers, body, -1, func(resp *http.Response) bool { return true })
 }
 
+// relevancyIdentity is what makes two vulnerability records the same finding: the CVE, and
+// the package it was found on. The package needs its version as well as its name, because a
+// name is not unique within an image. Two versions of one library can sit side by side, which
+// is ordinary in Java and Node images, and DomainToArmo emits a record per (vulnerability,
+// artifact) pair, so both versions produce a record carrying the same name.
+type relevancyIdentity struct {
+	cve            string
+	packageName    string
+	packageVersion string
+}
+
+func identifyForRelevancy(v cs.CommonContainerVulnerabilityResult) relevancyIdentity {
+	return relevancyIdentity{cve: v.Name, packageName: v.RelatedPackageName, packageVersion: v.PackageVersion}
+}
+
 // markRelevantVulnerabilities annotates each vulnerability with IsRelevant=true iff the same
-// (CVE, package) pair also appeared in the relevancy (CVEp) scan. Keying by CVE id alone would
-// mark a CVE relevant on every package it affects even when only one of those packages was
-// executed; the pair is the record identity (see RelatedPackageName in DomainToArmo and the
-// uniqueness assertion in backend_test.go).
+// finding also appeared in the relevancy (CVEp) scan. Keying by CVE id alone would mark a CVE
+// relevant on every package it affects even when only one of those packages was executed, and
+// keying by CVE and package name alone does the same thing one level down, marking an
+// unloaded version of a library relevant because another version of it was loaded. Both
+// manifests are built from the same artifacts, the filtered one from a subset of the same
+// SBOM, so the versions on either side are the same strings.
 func markRelevantVulnerabilities(vulnerabilities, relevantVulnerabilities []cs.CommonContainerVulnerabilityResult) {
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed the diff and the surrounding code.

- `relevancyIdentity{cve, packageName, packageVersion}` correctly matches how `DomainToArmo` populates `RelatedPackageName`/`PackageVersion` per (vulnerability, artifact) pair, so two versions of the same package no longer collide under one key. Confirmed `IsRelevant` is only consumed downstream in `backend_utils.go` for the summary's `RelevantCount`, so no other call site needs updating.
- The new regression test (`TestMarkRelevantVulnerabilities_DistinguishesPackageVersions`) reproduces the bug (asserts the unloaded 3.2.1 copy stays irrelevant while only the loaded 4.4 copy is marked relevant) and the existing `TestMarkRelevantVulnerabilities` still covers the one-version-per-package case.
- Struct-keyed map avoids the separator-collision risk called out in the description (e.g. `@` in npm scoped names) that a concatenated string key would have.
- CI is green (tests, lint, CodeQL, build) and CodeRabbit found no actionable comments.

No blockers. Approving.

PR #661, opened 2026-08-12
fix: preserve VEX metadata on no-op rescans

### LINKED ISSUE #660: bug: identical VEX rescans always bump version and ID
# Description
Identical rescans of an existing VEX document always bump `Metadata.Version`, rewrite `Metadata.LastUpdated`, and recalculate `Metadata.ID` even when the effective statements are unchanged. This creates avoidable object churn and makes the VEX ID unstable across no-op rescans.

## Summary
No-op rescans should preserve the existing VEX version and ID, or skip the update entirely, when the semantic document content has not changed.

## Affected Area
`repositories/apiserver.go`, specifically `StoreVEX` and `updateVEX`

## Technical Evidence
The update path in `updateVEX` currently unconditionally executes metadata mutations before recalculating the canonical hash:
- `vexDoc.Metadata.LastUpdated = time.Now().Format(time.RFC3339)`
- `vexDoc.Metadata.Version += 1`
- `calculateVexCanonicalHash(vexDoc)`

Running `go test ./repositories -count=1` passes on the current tree, and inspecting the current code confirms there is no unchanged-document branch before those mutations.

# Environment
OS: `darwin/arm64`
Version: `go1.26.5 workspace run on Wednesday, August 12, 2026`
 
# Steps To Reproduce
1. Run `go test ./repositories -count=1`.
2. Inspect `repositories/apiserver.go` around the `updateVEX` metadata update path.
3. Confirm that identical update inputs still set `LastUpdated`, increment `Metadata.Version`, and recalculate `Metadata.ID`.
4. Observe that there is no semantic diff short-circuit for an unchanged document.

# Expected behavior
Repeating the same `StoreVEX` input should leave `Metadata.Version` unchanged and preserve `Metadata.ID` when the effective VEX document content is unchanged.

# Actual Behavior
Every update path increments VEX version and emits a fresh ID, even for semantically identical rescans.

# Additional context
## Suspected Root Cause
`StoreVEX` currently treats every revisit of an existing VEX object as a state-changing update.

## Proposed Direction
Compute a semantic hash or compare a metadata-neutral copy of the document before mutating update metadata, and short-circuit when statements and metadata-relevant fields are unchanged.

## Acceptance Criteria
- Repeating the same `StoreVEX` input leaves `Metadata.Version` unchanged.
- `Metadata.ID` remains stable across identical rescans.
- A regression test covers create, unchanged update, and changed update cases.

rootp1:
## Overview
This change short-circuits `updateVEX` when the recomputed VEX document and merged object metadata are unchanged, so identical rescans keep the same VEX version, canonical ID, and `LastUpdated` value. It also adds a regression test that covers create, unchanged update, and changed update behavior, and updates the retry path tests so they still exercise real updates now that no-op rescans skip storage writes.

## How to Test
`go test ./repositories -count=1`

`make verify`

`go test ./...`

## Related issues/PRs:
Resolved #660

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented unchanged vulnerability scan results from creating unnecessary updates.
  * Preserved existing document metadata when rescans contain identical content.
  * Ensured changed scan results receive updated version and identification metadata.

* **Tests**
  * Added coverage for no-op rescans, metadata preservation, conflict handling, and update behavior.

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -5,7 +5,9 @@ import (
 	"crypto/sha256"
 	stderrors "errors"
 	"fmt"
+	"maps"
 	"net/url"
+	"reflect"
 	"sort"
 	"strconv"
 	"strings"
@@ -1167,7 +1169,8 @@ func (a *APIServerStore) updateVEX(ctx context.Context, cve domain.CVEManifest, 
 	imagePullable := cve.Annotations[helpersv1.ImageIDMetadataKey]
 
 	// Extend the VEX document with vulnerability data from full vulnerability manifest
-	vexDoc := vexContainer.Spec
+	originalVEX := *vexContainer.Spec.DeepCopy()
+	vexDoc := *vexContainer.Spec.DeepCopy()
 
 	// Statements written before the ID/Name mapping was corrected to match createVEX
 	// carry the CVE identifier in ID and the data source URL in Name. Normalize them in
@@ -1336,6 +1339,14 @@ func (a *APIServerStore) updateVEX(ctx context.Context, cve domain.CVEManifest, 
 	_ = markIgnoredVulnerabilitiesInVex(&vexDoc, &cve)
 	_ = markIgnoredVulnerabilitiesInVex(&vexDoc, &cvep)
 
+	mergedAnnotations := mergeMaps(maps.Clone(vexContainer.Annotations), cvep.Annotations)
+	mergedLabels := mergeMaps(maps.Clone(vexContainer.Labels), cvep.Labels)
+	if vexDocumentsEqualIgnoringUpdateMetadata(originalVEX, vexDoc) &&
+		maps.Equal(vexContainer.Annotations, mergedAnnotations) &&
+		maps.Equal(vexContainer.Labels, mergedLabels) {
+		return nil
+	}
+
 	// Update the VEX document metadata
 	vexDoc.Metadata.LastUpdated = time.Now().Format(time.RFC3339)
 	vexDoc.Metadata.Version += 1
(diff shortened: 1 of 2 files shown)


matthyx (dismissed):
Reviewed the fix and it holds up well.

**Correctness**
- `updateVEX` now deep-copies `vexContainer.Spec` into both `originalVEX` and `vexDoc` before mutating, avoiding slice/map aliasing between the two — needed for the equality check to be trustworthy.
- The no-op comparison happens after all statement mutation (dedup normalization, add/reset/mark-affected/mark-ignored) but before `LastUpdated`/`Version`/`ID` are bumped, so it correctly captures "did the effective content change" rather than "did we touch the struct."
- `vexDocumentsEqualIgnoringUpdateMetadata` zeroing `existing.LastUpdated/Version/ID` works because `v1beta1.VEX` embeds `Metadata` anonymously, so those are the same promoted fields `vexDoc.Metadata.*` mutates later — confirmed against the `kubescape/storage` type def.
- Good catch cloning `vexContainer.Annotations`/`Labels` before merging: `mergeMaps` mutates its first argument in place, so comparing against the original after merging without the clone would have made the annotations/labels equality check trivially true.
- The retry-path test updates (adding a differing annotation to `cve`) are necessary, not incidental — without them those tests would now hit the new no-op short-circuit and never exercise the conflict/retry behavior they're meant to guard.

**Verified locally**
- `go build ./...`, `go vet ./repositories/...`, `gofmt -l` clean.
- `go test ./...` and `go test ./repositories/... -race` (including the new `TestAPIServerStore_storeVEX_noopUpdatePreservesMetadata`) pass.
- Branch is ~9 commits behind `main`, including #666 which touches the same function (`updateVEX`, `Products[0].Subcomponents[0]` → `statementHasPURL`). Merged locally to double check: clean 3-way merge, both fixes coexist correctly, build+tests pass on the merged tree. GitHub also reports `mergeable: MERGEABLE` (the `BLOCKED` state looks like a review-required gate, not a conflict).

**Non-blocking nit**
- `TestAPIServerStore_StoreVEX_concurrentCreateRace`'s final assertion was weakened from `Version > 0` to just checking the object exists by name, since the create-then-fallback-update in that test now legitimately no-ops under the new logic. That's a correct consequence of the fix, but the assertion no longer proves the fallback went through `updateVEX` rather than being skipped — not worth blocking on, just flagging.

Approving — the fix matches #660's acceptance criteria (version/ID stable across no-op rescans, still bumps on real changes) and is well covered by tests.

matthyx (approved):
Re-checked after the new commit (5b55ced).

The only change is test-only: `TestAPIServerStore_storeVEX_noopUpdatePreservesMetadata` now installs an `update` reactor and asserts `updateCalls == 0` on a no-op rescan. That closes a real gap in the original test — before this, the test only checked that the resulting Version/ID/LastUpdated were unchanged, which would have also passed if `updateVEX` issued a storage Update that happened to write back identical values. Now it directly proves the write is skipped, matching the actual intent of the fix.

No production code changed since my last review. Re-verified:
- `go build ./...`, `go vet ./repositories/...`, `gofmt -l` clean.
- `go test ./...` passes; new test passes under `-race`.
- Still merges cleanly against current `main` (re-checked, no conflicts).

My prior approval was auto-dismissed by the new push — re-approving.

PR #662, opened 2026-08-12
fix: report actual SecurityException integration state in Diagnostics, not raw config flag

AHS0003:
## Summary
`Diagnostics.RiskAcceptanceEnabled` reported the raw `c.RiskAcceptance`
config flag instead of whether SecurityException integration was actually
active. SecurityException integration is only active when both
`RiskAcceptance` is enabled and `storage` is available.
When `storage == nil`, `seRepo` silently falls back to
`NoOpSecurityExceptionRepository`, which never suppresses anything.
So with `RiskAcceptance=true` and `storage == nil`, `/diagnostics` reported
`riskAcceptanceEnabled: true` while no suppression was actually happening.
This matters because, per its own doc comment, this endpoint exists
specifically so operators can check the running configuration "instead of
inferring it from logs"  it is intended to provide the effective runtime
state.
#562 identified the same misconfiguration (storage set,
`RiskAcceptance` unset), and #564 fixed it by adding a warning log. That
fix covered the logs; this PR covers the same gap on the diagnostics
endpoint, which was still reporting the wrong state.
## Changes
- Extracted the activation condition into a small, independently testable
  function: `isRiskAcceptanceActive(storage, riskAcceptance)`
- Reused the same function for both the `seRepo` branch (behavior
  unchanged) and `Diagnostics.RiskAcceptanceEnabled` (the fix), providing
  a single source of truth
## Testing
- Added `TestIsRiskAcceptanceActive`, covering all four combinations of
  storage/flag, including the regression case:
  `RiskAcceptance=true`, `storage=nil` → `false`
- Ran `go build ./...`
- Ran `go test ./cmd/http/... -run TestIsRiskAcceptanceActive -v` 
  all 4 subtests pass
- Ran `go test ./cmd/http/... -v`  existing `TestScan` suite still passes

## Summary by CodeRabbit

- **Bug Fixes**
  - SecurityException integration is now considered active only when both storage and risk-acceptance settings are enabled.
  - Diagnostics now accurately reflect the effective integration status instead of reporting the configured flag alone.

- **Tests**
  - Added coverage for enabled and disabled risk-acceptance settings with configured and unconfigured storage.

### DIFF
--- a/cmd/http/main.go
+++ b/cmd/http/main.go
@@ -104,7 +104,8 @@ func main() {
 	// SecurityException CRD integration requires storage and riskAcceptance RBAC
 	var seRepo ports.SecurityExceptionRepository
 	var eventRecorder record.EventRecorder
-	if storage != nil && c.RiskAcceptance {
+	riskAcceptanceActive := isRiskAcceptanceActive(storage, c.RiskAcceptance)
+	if riskAcceptanceActive {
 		seRepo = storage
 		logger.L().Info("SecurityException CRD integration enabled")
 
@@ -166,7 +167,7 @@ func main() {
 			ScanTimeout:             c.ScanTimeout.String(),
 			ScannerReadinessTimeout: c.ScannerReadinessTimeout.String(),
 			StorageEnabled:          c.Storage,
-			RiskAcceptanceEnabled:   c.RiskAcceptance,
+			RiskAcceptanceEnabled:   riskAcceptanceActive,
 		}
 	})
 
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed. The fix is correct and minimal:

- `isRiskAcceptanceActive(storage, riskAcceptance)` = `storage != nil && riskAcceptance` matches the pre-existing `seRepo` activation condition exactly, so that branch's behavior is unchanged (verified against `cmd/http/main.go`).
- `Diagnostics.RiskAcceptanceEnabled` now uses the same computed `riskAcceptanceActive` value instead of the raw `c.RiskAcceptance` flag, which is the actual bug fix — closes the gap left by #564 (which only fixed the log-side case from #562) for the diagnostics endpoint.
- `TestIsRiskAcceptanceActive` covers all 4 storage/flag combinations, including the regression case (`RiskAcceptance=true`, `storage=nil` → `false`).
- All CI checks green (build, lint, CodeQL, GitGuardian, DCO), no CodeRabbit actionable comments.

No blockers. Approving.

PR #664, opened 2026-08-13
fix: preserve external VEX statements without IDs

### LINKED ISSUE #663: Bug: isLocalStatement incorrectly identifies external VEX statements with empty IDs as local
# Description
When `kubevuln` updates a VEX document, it tries to keep its own local scan statements separate from external statements (like those from Red Hat or Chainguard) so it doesn't accidentally overwrite them. It does this using the `isLocalStatement` function, which assumes a statement is local if its ID starts with a kubescape URL or if its ID is completely empty (`id == ""`).

However, the OpenVEX specification says that statement IDs are completely optional. If an external VEX feed provides statements without IDs, `kubevuln` will incorrectly think they are local statements. When this happens, `updateVEX` will run its reset loop and forcefully overwrite the external statement's status and justification, completely erasing the vendor's data.

# Environment
OS: `Linux`
Version: `main branch`
 
# Steps To Reproduce
1. Fetch an external VEX document that has statements with no `@id` field (which is totally valid according to OpenVEX).
2. Save this document into an `OpenVulnerabilityExchangeContainer`.
3. Run a local scan that triggers `updateVEX` on that same container.
4. Check the VEX document again. The external statements will have their status overwritten by kubescape's own local assessment.

# Expected behavior
Kubevuln should not assume that an empty ID automatically means the statement belongs to kubescape. External statements without IDs should be preserved and ignored during the `updateVEX` reset loop, exactly like external statements that do have IDs.

# Actual Behavior
External VEX statements without an ID are treated as local kubescape statements. They get caught in the `updateVEX` reset loop, where their status and impact statements are wiped and replaced with kubescape's local scan data. This destroys the original external assessment.

# Additional context
This bug will directly break the upcoming LFX project that aims to consume external VEX feeds. If any vendor feed decides to omit the optional ID field, all of their data will be silently corrupted the moment a local scan runs.

A potential fix could be dropping the `id == ""` check in `isLocalStatement`. If older versions of kubescape generated statements without IDs, it's safer to treat them as external (which just leaves them alone and safely appends a new local statement) rather than risking the destruction of actual external feeds.

h30s:
## Overview
Fixes isLocalStatement misclassifying external VEX statements with no `@id`
as local kubescape statements. Per the OpenVEX spec, IDs are optional — any
external feed that omits them was having its statements silently overwritten
by the updateVEX reset loop.

## Fix
Removed the `id == ""` branch from isLocalStatement. A statement is now only
"local" if its ID has the kubescape URL prefix. No-ID statements are now
treated as external and left alone.

## Additional Information
Tradeoff: if any pre-existing kubescape-generated statements exist without
the URL-prefixed ID (e.g. from an older format), they'll now be treated as
external — left in place and duplicated by a new local statement, rather
than reset in place. This is the safer failure mode.

## How to Test
go test ./repositories/... -run TestAPIServerStore_updateVEX_preservesExternalStatements

Resolved #663

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected VEX statement handling so statements without IDs are preserved as external statements.
  * Local Kubescape statements continue to receive the appropriate baseline assessment while remaining available for regeneration.
  * Improved preservation of externally supplied vulnerability assessments during manifest updates.
* **Tests**
  * Updated coverage to verify correct handling of both ID-less external statements and local statements.

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -968,7 +968,7 @@ func buildActionStatement(v v1beta1.Match) string {
 }
 
 func isLocalStatement(id string) bool {
-	return id == "" || strings.HasPrefix(id, "https://kubescape.io/vex/statement/")
+	return strings.HasPrefix(id, "https://kubescape.io/vex/statement/")
 }
 
 func markRelevantVulnerabilitiesAsAffectedInVex(vexDoc *v1beta1.VEX, cvep *domain.CVEManifest) error {
--- a/repositories/apiserver_external_test.go
+++ b/repositories/apiserver_external_test.go
@@ -34,6 +34,15 @@ func TestAPIServerStore_updateVEX_preservesExternalStatements(t *testing.T) {
 	vexContainer, err := a.StorageClient.OpenVulnerabilityExchangeContainers(a.Namespace).Get(context.Background(), cveManifestFull.Name, metav1.GetOptions{})
 	require.NoError(t, err)
 
+	var target v1beta1.Statement
+	for _, s := range vexContainer.Spec.Statements {
+		if isLocalStatement(s.ID) && len(s.Products) > 0 && len(s.Products[0].Subcomponents) > 0 {
+			target = s
+			break
+		}
+	}
+	require.NotEmpty(t, target.Vulnerability.Name)
+
 	// Inject a purely external statement
 	externalStmt := v1beta1.Statement{
 		ID: "https://chainguard.dev/vex/statement/CVE-EXTERNAL-1234",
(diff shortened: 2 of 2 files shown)


matthyx (changes requested):
The fix itself looks correct and directly addresses #663: `isLocalStatement` no longer treats an empty ID as local, so ID-less external VEX statements are left alone by the `updateVEX` reset loop instead of being silently overwritten.

The updated test (`TestAPIServerStore_updateVEX_preservesExternalStatements`) is solid — it now exercises the ID-less-external case with a real vulnerability/product from the manifest (addressing CodeRabbit's earlier comment about that), asserts the statement is left byte-for-byte untouched, confirms a new local statement gets generated alongside it, and separately verifies a true kubescape-prefixed local statement still gets reset. The companion test for the dedup/suppression issue (#595) also still passes with this change.

Only blocker: the **DCO check is failing** — the commit (`7772370`) has no `Signed-off-by` trailer. Please amend the commit with `git commit --amend -s` and force-push so DCO passes.

Once DCO is green this is good to merge from my side.

matthyx:
@h30s the DCO check is failing on this PR — the commit doesn't have a `Signed-off-by` trailer. Could you amend it and force-push?

```
git commit --amend -s
git push --force-with-lease
```

Everything else looks good to merge once that's green.

h30s:
@matthyx Done! PTAL

matthyx (approved):
DCO is green now and the diff is unchanged apart from the sign-off. LGTM — approving.

PR #666, opened 2026-08-13
fix: match VEX statements across all products/subcomponents, not just the first 

### LINKED ISSUE #665: VEX matching only checks the first product/package, ignoring the rest
### What's happening

In `repositories/apiserver.go`, there are 3 places where the code checks
if a package is already recorded in a VEX statement. But instead of
checking all the packages listed, it only ever checks the very first one:

    if v.Artifact.PURL == s.Products[0].Subcomponents[0].ID {

(Lines 1203, 1257, and 1318)

### Why this is a problem

A VEX statement can list more than one product/package. But this code
only looks at the first one and ignores the rest. So if the package
we're looking for is the 2nd, 3rd, etc. in the list, the code will
wrongly say "not found" — even though it's really there.

### Why it hasn't caused issues yet

Right now, Kubescape's own code (`createProductStructForImageAndPackage`)
always creates statements with exactly one product/package, so this bug
never gets triggered  by accident, not by design.

### Simple proof

A small test confirms this:

    BEFORE FIX -> found: false   (package was really there, just not first)
    AFTER FIX  -> found: true    (checking all products/packages, not just the first)

### Suggested fix

Change the code to loop through all products and subcomponents instead
of only checking the first one. Add a test case with a statement that
has more than one product/package.

Shreya2005-2005:
## Overview

`repositories/apiserver.go` has three sites that match a VEX statement's package by indexing only the first product and first subcomponent:

    if v.Artifact.PURL == s.Products[0].Subcomponents[0].ID {

This is dormant today: `createProductStructForImageAndPackage` always builds exactly one product with exactly one subcomponent, so every local statement is single-product by construction. But the External VEX Ingestion project (kubevuln#387) will introduce statements from Red Hat CSAF and Chainguard OpenVEX, which can legitimately list multiple products/subcomponents per statement. Once external statement matching is wired up, the `[0][0]` indexing would silently miss real matches, with no error.

## Changes

Add `anyPURLMatches`, a shared traversal over every product/subcomponent in a statement, used by:

- `statementHasPURL`  the dedup check when adding new vulnerabilities (2 call sites)
- the ignored-vulnerability lookup in the statement-reset loop (1 call site)

All three sites previously ran their own independent `[0][0]` index checks; they now share one tested traversal.

## Before / After

Before (a plain, unprotected client):

    BEFORE FIX -> found: false

A package genuinely present in a statement, just not at `Products[0].Subcomponents[0]`, was reported as not found.

After (`anyPURLMatches`):

    AFTER FIX -> found: true

The same package is now found correctly regardless of its position.

## Testing

3 tests, covering the fix at both the unit and integration level:

- `TestStatementHasPURL_MatchesAcrossMultipleProducts` - unit-level: a match in a non-first position, a match in the first position, a genuine non-match, and nil input.
- `TestAPIServerStore_storeVEX_ignoredMatches_multiProductStatement`  integration-level: injects a real multi-product statement, runs it through the actual `StoreVEX`/`updateVEX` path, and confirms the ignore transition still fires correctly.
- `TestAPIServerStore_storeVEX_dedup_multiProductStatement`  integration-level: confirms the dedup check recognizes an existing multi-product statement through the real `StoreVEX` path and does not append a duplicate.

Full `repositories` package build and `go test ./...` pass with no failures (85 PASS, 0 FAIL).

## Related issues/PRs:

* Resolved [VEX matching only checks the first product/package, ignoring the rest](https://github.com/kubescape/kubevuln/issues/665) #665

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved vulnerability statement matching across all listed products and subcomponents.
  - Prevented duplicate vulnerability entries when a matching product is not listed first.
  - Correctly preserves and resets ignored status for statements containing multiple products or subcomponents.

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -953,6 +953,29 @@ func createProductStructForImageAndPackage(imagePullable string, packagePURL str
 	return &product, nil
 }
 
+// anyPURLMatches reports whether fn returns true for any subcomponent PURL across any
+// product in products. This is the shared traversal used by both statementHasPURL and
+// the ignored-vulnerability lookup in updateVEX, so both stay correct together if the
+// traversal logic ever needs to change.
+func anyPURLMatches(products []v1beta1.Product, fn func(purl string) bool) bool {
+	for _, p := range products {
+		for _, sc := range p.Subcomponents {
+			if fn(sc.ID) {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// statementHasPURL reports whether any subcomponent across any product in products
+// matches purl. External VEX statements (Red Hat CSAF, Chainguard OpenVEX) can list
+// multiple products/subcomponents per statement, so callers must not assume the match
+// is always at Products[0].Subcomponents[0].
+func statementHasPURL(products []v1beta1.Product, purl string) bool {
+	return anyPURLMatches(products, func(p string) bool { return p == purl })
+}
+
 // defaultActionStatement is used when a match does not carry enough fix data to build a
 // more specific remediation string.
 const defaultActionStatement = "Upgrade the vulnerable component to a version that is not affected"
@@ -1197,10 +1220,7 @@ func (a *APIServerStore) updateVEX(ctx context.Context, cve domain.CVEManifest, 
 				if s.Vulnerability.Name != v.Vulnerability.ID {
 					continue
 				}
-				if len(s.Products) == 0 || len(s.Products[0].Subcomponents) == 0 {
-					continue
-				}
-				if v.Artifact.PURL == s.Products[0].Subcomponents[0].ID {
+				if statementHasPURL(s.Products, v.Artifact.PURL) {
 					found = true
 					break
 				}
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed the fix, tests, and CI.

- `anyPURLMatches`/`statementHasPURL` correctly replace all three `Products[0].Subcomponents[0]` sites (dedup checks in `updateVEX` for both matches and ignored matches, plus the ignored-lookup in the statement-reset loop). Confirmed via grep that no `[0][0]` indexing on `Products`/`Subcomponents` remains in `apiserver.go`.
- Traversal logic is correct and matches regardless of product/subcomponent position; nil-products input is handled safely.
- Pulled the branch locally: `go build ./...` and `go vet ./repositories/...` are clean, and `go test ./repositories/...` passes (including the 3 new regression tests, one of which round-trips through the real `StoreVEX`/`updateVEX` path rather than just the isolated helper).
- No behavior change for today's local statements (still single-product/single-subcomponent by construction) — this only matters once external VEX ingestion (#387) lands multi-product statements, which is exactly the intent per the linked issue #665.
- All CI checks green, CodeRabbit found no actionable comments.

No blockers. Approving.

PR #668, opened 2026-08-13
test(domain): add unit tests for ScanError

AHS0003:
## Summary

`core/domain/scan.go` has no test file, yet it exports `ScanError`   a
type whose own doc comment makes a specific contract promise: wrapping a
sentinel error in `ScanError` must keep existing `errors.Is` checks
working unchanged, and callers up the stack must be able to recover the
`Reason` field via `errors.As` without re-parsing the error string. That
contract was only exercised indirectly, through one case in
`controllers/http_test.go`.

## Changes

- `TestScanError_Error` / `TestScanError_Unwrap`  : basic interface behavior
- `TestScanError_ErrorsIs`  :confirms wrapping a sentinel error (e.g.
  `ErrTooManyRequests`) doesn't break `errors.Is` checks against it
- `TestScanError_ErrorsAs`  :  confirms the `Reason` field is recoverable via
  `errors.As`
- `TestScanError_WrappedInFmtErrorf`  : confirms both properties still hold
  when `ScanError` is wrapped one level further (e.g. via `errors.Join`)

## Testing

- Ran `go build ./...`
- Ran `go test ./core/domain/... -run TestScanError -v`   all 5 tests pass

## Summary by CodeRabbit

- **Tests**
  - Added coverage for scan error messages and unwrapping behavior.
  - Verified compatibility with standard error inspection, including wrapped and combined errors.
  - Confirmed scan errors can be identified and retrieved correctly through contextual error chains.

### DIFF
--- a/core/domain/scan_test.go
+++ b/core/domain/scan_test.go
@@ -1,0 +1,55 @@
+package domain
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestScanError_Error(t *testing.T) {
+	inner := errors.New("boom")
+	scanErr := &ScanError{Reason: "some-reason", Err: inner}
+
+	assert.Equal(t, "boom", scanErr.Error())
+}
+
+func TestScanError_Unwrap(t *testing.T) {
+	inner := errors.New("boom")
+	scanErr := &ScanError{Reason: "some-reason", Err: inner}
+
+	assert.Equal(t, inner, scanErr.Unwrap())
+}
+
+func TestScanError_ErrorsIs(t *testing.T) {
+	// ScanError's own doc comment promises that wrapping a sentinel error keeps
+	// existing errors.Is-based checks working exactly as before wrapping.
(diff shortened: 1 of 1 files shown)


matthyx (approved):
Reviewed the new `core/domain/scan_test.go`. All five cases check out against `ScanError`'s actual behavior:

- `Error()`/`Unwrap()` delegate straight to the wrapped `Err`, matching the assertions.
- `errors.Is` against `ErrTooManyRequests` succeeds and against an unrelated sentinel (`ErrMissingSBOM`) correctly fails, since `Unwrap` exposes the original error unchanged.
- `errors.As` recovers `Reason` as documented.
- The `errors.Join` case confirms both `errors.As` and `errors.Is` still work one level further down the chain.

Test-only change, no production code touched, all CI checks green. Non-blocking nit already flagged by CodeRabbit: `TestScanError_WrappedInFmtErrorf` (line 45) actually wraps via `errors.Join`, not `fmt.Errorf` — the name/comment could be updated for accuracy, but it doesn't affect correctness. Good to merge.

PR #669, opened 2026-08-13
fix: sweep stale stereoscope temp dirs periodically, not once at startup

### LINKED ISSUE #667: Scan temp dirs from stereoscope leak for the life of long-running pods; sidecar has no sweep at all, and existing dead cleanup code masks it
# Description

`kubevuln` leaks stereoscope-managed image temp directories on disk over the lifetime of a long-running pod, with no in-process recovery — only a one-time sweep at startup. Two separate gaps combine to make this a real accumulation risk rather than a cosmetic one:

**1. The startup sweep only ever runs once, and only in one of the two processes.** PR #491 added `tools.CleanupStaleTempDirs(os.TempDir(), "stereoscope-", c.ScanTimeout)`, called exactly once in `cmd/http/main.go` before the HTTP server starts (`cmd/http/main.go:190`). Its own comment frames it as reclaiming dirs "orphaned by previous process invocations killed by SIGKILL... before their defer-based cleanup could run" — i.e. it was designed for the crash-restart case, not for leaks that occur during ordinary, un-crashed operation of a single long-running process. `kubevuln` pods are not restarted per scan; once started, this sweep never runs again for the rest of that pod's uptime, however many days or weeks that is. Worse, `cmd/sbom-scanner/main.go` — the sidecar process that actually performs the pod-less registry pulls described below — never calls `tools.CleanupStaleTempDirs` at all (confirmed: no reference to it anywhere in `cmd/sbom-scanner/`); it has no sweep, startup or otherwise.

**2. Leaks happen during ordinary operation, not just crashes.** The vendored fork (`github.com/matthyx/stereoscope@v0.1.22-ks.1`, replacing `github.com/anchore/stereoscope` in `go.mod`) creates the per-scan temp directory *before* the point where cleanup becomes reachable:

```go
// pkg/image/oci/registry_provider.go:53-95 (registryImageProvider.Provide)
imageTempDir, err := p.tmpDirGen.NewDirectory("oci-registry-image")   // dir created here
...
descriptor, err := remote.Get(ref, options...)                        // return nil, err — dir leaked
...
img, err := descriptor.Image()                                        // return nil, err — dir leaked
...
c, err := img.ConfigFile()                                            // return nil, err — dir leaked
...
if err := validatePlatform(platform, c.OS, c.Architecture, c.Variant); err != nil {
    return nil, err                                                    // return nil, err — dir leaked
}
...
out := image.New(img, p.tmpDirGen, imageTempDir, metadata...)          // only past this point is out.Cleanup() reachable
```

Every `return nil, err` between directory creation and `image.New(...)` leaks `imageTempDir` permanently — there is no owner left holding a reference to clean it up. Critically, `validatePlatform` failing (platform mismatch) is not an edge case here: `kubevuln` explicitly treats platform mismatch as a routine, expected outcome on pod-less scan paths — registry rescans and periodic CRD-based rescans — and even has a dedicated `FallbackStrategyPlatformMismatch` metric for it (`pkg/sbomscanner/v1/server.go:190-192`). Registry auth failures and transient registry errors hit the same leak path. None of these are crashes; the process keeps running and keeps leaking.

**3. `kubevuln`'s own code makes the situation look handled when it isn't.** Both call sites that talk to stereoscope construct a *second*, unrelated `TempDirGenerator` and defer its `Cleanup()`, appearing to provide exactly the missing safety net:

```go
// adapters/v1/syft.go:183-190
t := file.NewTempDirGenerator("stereoscope")
defer func(t *file.TempDirGenerator) {
    err := t.Cleanup()
    ...
}(t)
```
```go
// pkg/sbomscanner/v1/server.go:223-229 — identical pattern, comment says
// "Prepare temp dir for stereoscope" / "own temp dir (via its own file.NewTempDirGenerator)"
```

`file.NewTempDirGenerator(name)` only sets `rootPrefix`; `rootLocation` stays `""` until `NewDirectory()`/`NewGenerator()` is called on it. Neither call site ever calls either method on `t` — it is used *exclusively* to invoke `Cleanup()`. Per `TempDirGenerator.Cleanup()`'s own implementation, `t.rootLocation == ""` and `t.children == nil`, so the loop and `os.RemoveAll` both no-op: **this `defer t.Cleanup()` deletes nothing, every single call.** The real temp dirs are created by stereoscope's own package-level `rootTempDirGenerator` (`client.go:19`, reached via `providers.go:30`'s `rootTempDirGenerator.NewGenerator()`), a completely different object `t` has no relationship to. `kubevuln` never calls `stereoscope.Cleanup()` (the function that would flush `rootTempDirGenerator`) anywhere either.

# Environment

OS: any
Version: `main` (Aug 2026)

# Steps To Reproduce

1. In `adapters/v1/syft.go`, read `SyftAdapter.CreateSBOM` lines 183-190: `t` is constructed, `Cleanup()` is deferred, and `t.NewDirectory()`/`t.NewGenerator()` is never called anywhere in the function.
2. Confirm the same pattern in `pkg/sbomscanner/v1/server.go:223-229`.
3. In the vendored fork (`$(go env GOMODCACHE)/github.com/matthyx/stereoscope@v0.1.22-ks.1/pkg/file/temp_dir_generator.go`), confirm `Cleanup()` is a no-op when `rootLocation == ""`.
4. In the same module, `pkg/image/oci/registry_provider.go`, confirm `imageTempDir` is created before `remote.Get`, `descriptor.Image()`, `img.ConfigFile()`, and `validatePlatform`, and that every one of those failure branches returns without any reference to `imageTempDir` being retained.
5. In `cmd/http/main.go:188-196`, confirm `tools.CleanupStaleTempDirs` is called exactly once, before `router := gin.New()`, with no periodic re-invocation anywhere in the file (no ticker, no cron).

# Expected behavior

A registry-scan failure that occurs after the image temp directory is created (auth failure, transient registry error, platform mismatch) should not leave an orphaned directory on disk for the remaining lifetime of the pod. Since `kubevuln` pods run for extended periods without restarting, temp-dir reclamation needs to happen periodically during normal operation, not just once at process startup.

# Actual Behavior

- Every registry-provider failure between `NewDirectory` and `image.New(...)` leaks a temp directory that nothing in-process ever cleans up.
- The `defer t.Cleanup()` in both `adapters/v1/syft.go` and `pkg/sbomscanner/v1/server.go` gives the false impression that these leaks are handled — it is a guaranteed no-op.
- `tools.CleanupStaleTempDirs` only runs once, at process startup, so none of the above leaks are reclaimed until the pod restarts — which, for a long-running scanner pod, may be days or weeks, during which every failed registry pull adds one more orphaned directory (`oci-registry-image-*` under a `stereoscope-*` root) to the filesystem. On a node with many scanned images or noisy/flaky registries, this is a real path to disk pressure and eventual pod eviction or scan failures from `ENOSPC`.

# Additional context

## Proposed Solution

1. **Run the sweep periodically in both processes, not just once at `cmd/http` startup.** Replace the one-shot call in `cmd/http/main.go` with a `time.Ticker`-driven loop, started alongside the existing shutdown machinery and stopped on `ctx.Done()`, and add the same periodic sweep to `cmd/sbom-scanner/main.go`, which currently has none at all. Keep the existing `ScanTimeout`-based staleness threshold — a dir idle longer than `ScanTimeout` can only belong to a dead or already-timed-out scan, whether or not the process crashed. A reasonable sweep interval should be config-driven, defaulting to something on the order of `ScanTimeout` itself.
2. **Remove or fix the misleading dead code.** Delete the no-op `t := file.NewTempDirGenerator("stereoscope")` / `defer t.Cleanup()` blocks in `adapters/v1/syft.go` and `pkg/sbomscanner/v1/server.go` (and their comments claiming they prepare/clean stereoscope's temp dir) — they do nothing today and give reviewers a false sense that this is handled. If in-process cleanup is wanted in addition to the periodic sweep, it would need to be wired to the object stereoscope actually creates the directory through, which the current call sites don't have access to.
3. **Add a metric for sweep effectiveness**, consistent with the existing `metrics.RecordScanFallback` pattern already used for other fallback/reclamation paths, so operators can see how many stale dirs are being reclaimed per sweep — a non-zero, growing count in steady state is itself a signal that leaks are outpacing the interval.
4. **Regression tests**: a table-driven test for the ticker-based sweep (dirs older than the threshold get removed on tick, fresh in-flight dirs don't); a test asserting the dead `t.Cleanup()` removal doesn't regress the (already broken) coverage; and ideally a test against `oci.NewRegistryProvider` confirming which failure paths leak (documenting the upstream gap even if not fixed at the source, since that lives in the `matthyx/stereoscope` fork kubevuln vendors via a `replace` directive — a maintainer-controlled fork, so the source-level fix is a realistic follow-up but out of scope for this issue).

## Acceptance Criteria

- [ ] Temp-dir reclamation runs periodically for the life of the process in both `cmd/http` (currently once at startup) and `cmd/sbom-scanner` (currently never).
- [ ] The dead-code `TempDirGenerator`/`Cleanup()` call sites in `adapters/v1/syft.go` and `pkg/sbomscanner/v1/server.go` are removed or actually wired to real cleanup, and their misleading comments are corrected.
- [ ] Sweep activity (dirs removed per run) is observable via metrics.
- [ ] Regression tests cover the periodic sweep and document the registry-provider leak paths.

bhuvan-somisetty:
## Problem

`kubevuln` leaks stereoscope-managed image temp directories on disk over the lifetime of a long-running pod, with no in-process recovery.

## Root cause

Two gaps combine:

1. **The reclaim sweep only ever ran once.** `tools.CleanupStaleTempDirs` (added in #491) was called exactly once, at `cmd/http` startup, before the HTTP server starts listening. Its own comment frames it around dirs orphaned by a crashed *previous* process invocation — it does nothing for leaks that occur during ordinary operation of the *current*, still-running process. `cmd/sbom-scanner` never called it at all, despite being the process that performs pod-less registry pulls (registry rescans, periodic CRD-based rescans).
2. **Leaks happen during ordinary operation, not just crashes.** The vendored fork (`github.com/matthyx/stereoscope@v0.1.22-ks.1`) creates the per-scan temp directory in `registryImageProvider.Provide` *before* the point where cleanup becomes reachable — `remote.Get`, `descriptor.Image()`, `img.ConfigFile()`, and `validatePlatform` can all return an error with the temp dir already on disk and nothing left holding a reference to clean it up. `validatePlatform` failing (platform mismatch) is not an edge case here: `kubevuln` explicitly treats it as a routine, metric-tracked outcome on pod-less scan paths (`FallbackStrategyPlatformMismatch`).

Making it worse, both call sites (`adapters/v1/syft.go`, `pkg/sbomscanner/v1/server.go`) constructed a second, unrelated `file.NewTempDirGenerator("stereoscope")` purely to `defer` its `Cleanup()`, which looked like the missing safety net. It wasn't: `NewDirectory`/`NewGenerator` was never called on that generator, so its `rootLocation` stayed `""` and `Cleanup()` was a guaranteed no-op on every call — dead code that actively misled anyone reading it into thinking this was handled.

Since `kubevuln` pods run for days/weeks without restarting rather than being recycled per scan, every leak from (2) accumulated for the pod's entire uptime with zero reclamation until the next restart.

## Fix

- Added `tools.StartPeriodicTempDirSweep`: runs the existing `CleanupStaleTempDirs` immediately (covering the original startup case) and then again on every tick of a configurable interval, until the process shuts down.
- `cmd/http/main.go` now runs the sweep on a `ScanTimeout`-driven ticker instead of once at startup — same staleness threshold as before, so a dir idle longer than `ScanTimeout` still can only belong to a dead or already-timed-out scan.
- `cmd/sbom-scanner/main.go` gets a periodic sweep for the first time (it had none, startup or otherwise), on a 5-minute interval matching the same default `CreateSBOM` already falls back to when a request carries no `TimeoutSeconds`.
- Removed the dead, no-op `TempDirGenerator`/`Cleanup()` blocks and their misleading comments in `adapters/v1/syft.go` and `pkg/sbomscanner/v1/server.go`.
- Added `kubevuln_temp_dir_sweep_removed_total`, a counter (by `component`) so sweep activity is observable — a steady, non-zero rate is itself a signal that leaks are occurring at least as fast as the sweep interval.

## Testing

- New regression tests in `internal/tools/tools_test.go`: immediate sweep runs synchronously before `StartPeriodicTempDirSweep` returns, a dir leaked *after* startup is reclaimed by a later tick (the scenario a one-shot sweep can't cover), the sweep goroutine stops ticking once `stop` is closed, and a nil `onSweep` doesn't panic.
- New tests in `internal/metrics/metrics_test.go`: the new counter is exposed correctly, a zero/negative `removed` is a no-op (no zero-valued series emitted on every steady-state tick), and the exported `RecordTempDirSweep` doesn't panic before `metrics.New()` has run (the sidecar's actual production condition today).
- `go build ./...`, `go vet ./...`, and the full test suite for every touched package pass.
- `golangci-lint run --new-from-rev=<merge-base>` (matching this repo's diff-scoped `make lint`) reports 0 issues.

## Impact

Bounds worst-case leaked-disk accumulation to roughly one sweep interval's worth of failures, in both processes, for the life of the pod — instead of unbounded accumulation between restarts. No behavior change for callers; `CleanupStaleTempDirs` itself is untouched.

Fixes #667

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup of stale temporary files during and after scanning.
  * Prevented leaked temporary directories from accumulating between scans or after unexpected shutdowns.
  * Prevented cleanup from interfering with active scans.
  * Invalid non-positive scan timeout values are now rejected.

* **Monitoring**
  * Added metrics and logging for temporary-directory cleanup activity, errors, and removals.

* **Reliability**
  * Cleanup now runs periodically in supported services and stops cleanly during shutdown.
  * Improved protection against cleanup conflicts during active SBOM generation.

### DIFF
--- a/cmd/http/main.go
+++ b/cmd/http/main.go
@@ -181,20 +181,30 @@ func main() {
 		logger.L().Ctx(ctx).Fatal("metrics registration error", helpers.Error(err))
 	}
 
-	// Best-effort startup sweep: reclaim stereoscope temp dirs orphaned by previous process
-	// invocations killed by SIGKILL (OOM kill, or terminationGracePeriodSeconds exceeded)
-	// before their defer-based cleanup could run. The threshold is ScanTimeout: the main
-	// process and the SBOM-scanner sidecar run in the same Pod on the same machine (same
-	// kernel clock), so no clock-skew margin is needed, and a dir idle longer than
-	// ScanTimeout can only belong to a dead or already-timed-out scan.
-	if removed, err := tools.CleanupStaleTempDirs(os.TempDir(), "stereoscope-", c.ScanTimeout); err != nil {
-		logger.L().Ctx(ctx).Warning("startup temp dir sweep error",
-			helpers.Error(err),
-			helpers.Int("removed", removed))
-	} else if removed > 0 {
-		logger.L().Ctx(ctx).Info("startup temp dir sweep complete",
-			helpers.Int("removed", removed))
-	}
+	// Best-effort periodic sweep: reclaim stereoscope temp dirs left behind either by a
+	// previous process invocation killed by SIGKILL (OOM kill, or terminationGracePeriodSeconds
+	// exceeded) before its defer-based cleanup could run, or by an ordinary (non-crash) registry
+	// pull failure in *this* process — stereoscope creates the image temp dir before it has
+	// anything to Cleanup() it, so auth failures, transient registry errors, and platform
+	// mismatches can all leak one. This process runs for the life of a long-lived pod rather
+	// than restarting per scan, so a single startup-only sweep would never reclaim leaks from
+	// its own uptime; run it on a recurring interval instead. The threshold is ScanTimeout: the
+	// main process and the SBOM-scanner sidecar run in the same Pod on the same machine (same
+	// kernel clock), so no clock-skew margin is needed, and a dir idle longer than ScanTimeout
+	// can only belong to a dead or already-timed-out scan. ScanTimeout also doubles as the sweep
+	// interval, keeping worst-case accumulation bounded to roughly one ScanTimeout's worth of
+	// leaks at a time.
+	tools.StartPeriodicTempDirSweep(ctx.Done(), os.TempDir(), "stereoscope-", c.ScanTimeout, c.ScanTimeout, func(removed int, err error) {
+		if err != nil {
+			logger.L().Ctx(ctx).Warning("temp dir sweep error",
+				helpers.Error(err),
+				helpers.Int("removed", removed))
+		} else if removed > 0 {
+			logger.L().Ctx(ctx).Info("temp dir sweep complete",
+				helpers.Int("removed", removed))
+		}
+		metrics.RecordTempDirSweep(ctx, metrics.ComponentInProcess, removed)
+	})
 
 	gin.SetMode(gin.ReleaseMode)
 	router := gin.New()
--- a/cmd/sbom-scanner/main.go
+++ b/cmd/sbom-scanner/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"net"
 	"os"
 	"os/signal"
@@ -10,6 +11,8 @@ import (
 
 	"github.com/kubescape/go-logger"
 	"github.com/kubescape/go-logger/helpers"
+	"github.com/kubescape/kubevuln/internal/metrics"
+	"github.com/kubescape/kubevuln/internal/tools"
 	sbomscanner "github.com/kubescape/kubevuln/pkg/sbomscanner/v1"
 	pb "github.com/kubescape/kubevuln/pkg/sbomscanner/v1/proto"
 	"google.golang.org/grpc"
@@ -29,6 +32,14 @@ import (
 // shutdownTimeout default so both processes give a comparable grace period before a
 // terminationGracePeriodSeconds-driven SIGKILL would otherwise cut them off mid-shutdown.
 const shutdownTimeout = 20 * time.Second
+
+// tempDirSweepInterval bounds how long a stereoscope temp dir leaked by this process (e.g. a
+// registry pull that fails after the image temp dir is created but before there's an
+// image.Image to Cleanup() it -- see internal/tools.StartPeriodicTempDirSweep) can sit on disk
+// before being reclaimed. This process has no config file of its own (unlike cmd/http, which
+// derives the same cadence from ScanTimeout), so it uses the same 5-minute value CreateSBOM
+// already falls back to when a request carries no TimeoutSeconds (pkg/sbomscanner/v1/server.go).
+const tempDirSweepInterval = 5 * time.Minute
 
 // gracefulStopWithTimeout waits up to timeout for srv's in-flight RPCs to finish via
 // GracefulStop, then force-closes the server with Stop if the timeout elapses first.
@@ -79,12 +90,26 @@ func main() {
 	)
 	pb.RegisterSBOMScannerServer(srv, sbomscanner.NewScannerServer())
 
+	// See tempDirSweepInterval: this process (unlike cmd/http) previously had no temp-dir
+	// sweep at all, startup or periodic, despite being the one that performs pod-less
+	// registry pulls (registry rescans, periodic CRD-based rescans).
+	stopSweep := make(chan struct{})
+	tools.StartPeriodicTempDirSweep(stopSweep, os.TempDir(), "stereoscope-", tempDirSweepInterval, tempDirSweepInterval, func(removed int, err error) {
+		if err != nil {
+			logger.L().Warning("temp dir sweep error", helpers.Error(err), helpers.Int("removed", removed))
+		} else if removed > 0 {
+			logger.L().Info("temp dir sweep complete", helpers.Int("removed", removed))
+		}
+		metrics.RecordTempDirSweep(context.Background(), metrics.ComponentSidecar, removed)
+	})
+
 	sigCh := make(chan os.Signal, 1)
 	signal.Notify(sigCh, syscall.SIGTERM, syscall.SIGINT)
 
 	go func() {
 		sig := <-sigCh
 		logger.L().Info("received signal, shutting down", helpers.String("signal", sig.String()))
+		close(stopSweep)
 		gracefulStopWithTimeout(srv, shutdownTimeout)
 		if err := os.Remove(socketPath); err != nil && !os.IsNotExist(err) { // #nosec G703 -- SOCKET_PATH is operator-controlled deployment config; path is cleaned above
 			logger.L().Warning("failed to remove socket file on shutdown", helpers.Error(err), helpers.String("path", socketPath))
--- a/cmd/sbom-scanner/main.go
+++ b/cmd/sbom-scanner/main.go
@@ -44,8 +44,12 @@ const tempDirSweepInterval = 5 * time.Minute
 
 // defaultMetricsAddr is the listen address for the Prometheus /metrics endpoint, used when
 // METRICS_ADDR is unset. Overridable per-deployment via METRICS_ADDR, mirroring SOCKET_PATH
-// below.
-const defaultMetricsAddr = ":8080"
+// below. Deliberately not :8080: this process runs as a sidecar container in the same Pod as
+// cmd/http, which binds :8080 for its own server (also serving /metrics there). Containers in
+// a Pod share one network namespace, so both processes defaulting to :8080 would collide --
+// this one would fail to bind and its component="sidecar" metrics, including
+// kubevuln_temp_dir_sweep_removed_total, would silently never be served.
+const defaultMetricsAddr = ":8081"
 
 // gracefulStopWithTimeout waits up to timeout for srv's in-flight RPCs to finish via
 // GracefulStop, then force-closes the server with Stop if the timeout elapses first.
--- a/cmd/sbom-scanner/main.go
+++ b/cmd/sbom-scanner/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"net"
+	"net/http"
 	"os"
 	"os/signal"
 	"path/filepath"
@@ -40,6 +41,11 @@ const shutdownTimeout = 20 * time.Second
 // derives the same cadence from ScanTimeout), so it uses the same 5-minute value CreateSBOM
 // already falls back to when a request carries no TimeoutSeconds (pkg/sbomscanner/v1/server.go).
 const tempDirSweepInterval = 5 * time.Minute
+
+// defaultMetricsAddr is the listen address for the Prometheus /metrics endpoint, used when
+// METRICS_ADDR is unset. Overridable per-deployment via METRICS_ADDR, mirroring SOCKET_PATH
+// below.
+const defaultMetricsAddr = ":8080"
 
 // gracefulStopWithTimeout waits up to timeout for srv's in-flight RPCs to finish via
 // GracefulStop, then force-closes the server with Stop if the timeout elapses first.
@@ -84,6 +90,24 @@ func main() {
 		logger.L().Fatal("failed to listen on socket", helpers.Error(err), helpers.String("path", socketPath))
 	}
 
+	// Without this, RecordTempDirSweep below (and any other metrics.Record* called from the
+	// sidecar's own code paths, e.g. adapters that run in-process here) reads a nil counter and
+	// is a no-op: the component="sidecar" series would never be emitted, silently.
+	m, err := metrics.New()
+	if err != nil {
+		logger.L().Fatal("metrics initialization error", helpers.Error(err))
+	}
+	metricsAddr := os.Getenv("METRICS_ADDR")
+	if metricsAddr == "" {
+		metricsAddr = defaultMetricsAddr
+	}
+	metricsServer := &http.Server{Addr: metricsAddr, Handler: m.Handler()}
+	go func() {
+		if err := metricsServer.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+			logger.L().Warning("metrics server stopped unexpectedly", helpers.Error(err))
+		}
+	}()
+
 	srv := grpc.NewServer(
 		grpc.MaxRecvMsgSize(sbomscanner.MaxgRPCMessageSize),
 		grpc.MaxSendMsgSize(sbomscanner.MaxgRPCMessageSize),
@@ -111,6 +135,11 @@ func main() {
 		logger.L().Info("received signal, shutting down", helpers.String("signal", sig.String()))
 		close(stopSweep)
 		gracefulStopWithTimeout(srv, shutdownTimeout)
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), shutdownTimeout)
+		defer cancel()
+		if err := metricsServer.Shutdown(shutdownCtx); err != nil {
+			logger.L().Warning("metrics server shutdown error", helpers.Error(err))
+		}
 		if err := os.Remove(socketPath); err != nil && !os.IsNotExist(err) { // #nosec G703 -- SOCKET_PATH is operator-controlled deployment config; path is cleaned above
 			logger.L().Warning("failed to remove socket file on shutdown", helpers.Error(err), helpers.String("path", socketPath))
 		}
--- a/cmd/sbom-scanner/main.go
+++ b/cmd/sbom-scanner/main.go
@@ -105,7 +105,14 @@ func main() {
 	if metricsAddr == "" {
 		metricsAddr = defaultMetricsAddr
 	}
-	metricsServer := &http.Server{Addr: metricsAddr, Handler: m.Handler()}
+	metricsServer := &http.Server{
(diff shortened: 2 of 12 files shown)


cmd/sbom-scanner/main.go
```diff
@@ -29,6 +33,19 @@ import (
...
+// image.Image to Cleanup() it -- see internal/tools.StartPeriodicTempDirSweep) can sit on disk
+// before being reclaimed. This process has no config file of its own (unlike cmd/http, which
+// derives the same cadence from ScanTimeout), so it uses the same 5-minute value CreateSBOM
+// already falls back to when a request carries no TimeoutSeconds (pkg/sbomscanner/v1/server.go).
+const tempDirSweepInterval = 5 * time.Minute
+
+// defaultMetricsAddr is the listen address for the Prometheus /metrics endpoint, used when
+// METRICS_ADDR is unset. Overridable per-deployment via METRICS_ADDR, mirroring SOCKET_PATH
+// below.
+const defaultMetricsAddr = ":8080"
```
matthyx: Potential port collision: `cmd/http/main.go` already binds `:8080` for its main HTTP server, which also serves `/metrics` (`router.GET("/metrics", gin.WrapH(m.Handler()))`, `cmd/http/main.go:216,229`). This PR's comment in `cmd/http/main.go` (and the PR description) states the HTTP process and this sidecar "run in the same Pod on the same machine" — in Kubernetes, containers in the same Pod share one network namespace/IP, so two containers can't both bind TCP `:8080`.

If `METRICS_ADDR` isn't set to something else for this sidecar container in the deployment manifests, `metricsServer.ListenAndServe()` will fail with "address already in use" here. That failure is only logged as a `Warning` (not `Fatal`), so the process keeps running but silently never serves `component="sidecar"` metrics — including the new `kubevuln_temp_dir_sweep_removed_total` counter this PR adds specifically to make sweep activity on this process observable, which would defeat that goal in production.

Could you confirm the deployment manifests (helm chart / k8s YAML, which aren't in this repo) already set a distinct `METRICS_ADDR` for the sidecar container? If not, consider defaulting to a different port here (e.g. `:8081`) to avoid the footgun.

internal/tools/tools.go
```diff
@@ -156,6 +181,46 @@ func CleanupStaleTempDirs(dir, prefix string, olderThan time.Duration) (int, err
...
+// process, e.g. a registry pull that fails after stereoscope creates its image temp dir but
+// before an image.Image exists to Cleanup() it (auth failures, transient registry errors,
+// platform mismatches - none of which are crashes). Since kubevuln processes stay up for the
+// life of a long-running pod rather than restarting per scan, those leaks would otherwise
+// accumulate for the pod's entire uptime. onSweep, if non-nil, is invoked after every sweep
+// (including the immediate one) with its outcome, so callers can log/record metrics without
+// this function taking a dependency on either.
+func StartPeriodicTempDirSweep(stop <-chan struct{}, dir, prefix string, olderThan, interval time.Duration, onSweep func(removed int, err error)) {
+	sweep := func() {
+		if activeTempDirUsers.Load() > 0 {
```
matthyx, line 256: Non-blocking note: `activeTempDirUsers` is process-wide, not per-directory, so the sweep is skipped in full whenever *any* scan is active. `pkg/sbomscanner/v1/server.go`'s `CreateSBOM` explicitly avoids a mutex/semaphore to allow concurrent RPCs (see the comment above that handler), so under sustained concurrent traffic on the sidecar — a new request starting before the previous one's cataloguing finishes — this counter may rarely or never reach zero, and the periodic sweep could be skipped indefinitely.

Not a regression (the sidecar had zero sweep of any kind before this PR) and probably fine in practice given real request rates, but worth flagging that the "bounded to roughly one sweep interval" guarantee described in the PR only holds when the process has idle gaps between scans, not under continuous concurrent load.

matthyx (commented):
Solid fix overall — I traced the root cause claims against the actual vendored stereoscope fork (`github.com/matthyx/stereoscope@v0.1.22-ks.1`) and they check out: `registryImageProvider.Provide` does create the temp dir via the process-singleton `rootTempDirGenerator` (prefix `"stereoscope"`) before any error path, and the removed `file.NewTempDirGenerator("stereoscope")`/`Cleanup()` blocks were indeed guaranteed no-ops (`rootLocation` never set without a `NewDirectory`/`NewGenerator` call). The periodic sweep, active-use guard, config validation, and new metric are all reasonable and well tested.

One likely deployment blocker and one non-blocking design note below — holding off on approval until the port question is resolved.

cmd/sbom-scanner/main.go
bhuvan-somisetty: Good catch — confirmed both processes default to :8080 in the same Pod, that would have collided in production. Changed the sidecar's default to :8081 in 62b2e24 (also documented why, so it's not accidentally reverted back to :8080 later). Let me know if you'd prefer a different port or an env-only default with no fallback.

internal/tools/tools.go
bhuvan-somisetty: Agreed, that's the correct tradeoff given CreateSBOM's no-mutex design — not changing it to per-directory tracking here. Added a doc-comment note on StartPeriodicTempDirSweep spelling out the concurrent-load caveat so it's explicit for the next reader.

cmd/sbom-scanner/main.go
```diff
@@ -69,19 +90,56 @@ func main() {
...
+	// is a no-op: the component="sidecar" series would never be emitted, silently.
+	m, err := metrics.New()
+	if err != nil {
+		logger.L().Fatal("metrics initialization error", helpers.Error(err))
+	}
+	metricsAddr := os.Getenv("METRICS_ADDR")
+	if metricsAddr == "" {
+		metricsAddr = defaultMetricsAddr
+	}
+	metricsServer := &http.Server{Addr: metricsAddr, Handler: m.Handler()}
```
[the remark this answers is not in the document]
bhuvan-somisetty: Fair point, these were unbounded. Added ReadHeaderTimeout/ReadTimeout/WriteTimeout/IdleTimeout in c92a105.

bhuvan-somisetty:
Pushed two more commits addressing everything above — moved the sidecar's metrics port off :8080 to avoid the collision with cmd/http in the same Pod, added bounded timeouts to the sidecar's metrics server per the latest CodeRabbit pass, and documented the concurrent-load caveat on the sweep guard. Build, vet, and the full test suite are all green. Resolved the threads above accordingly — let me know if anything still needs another look.

matthyx (dismissed):
Port collision resolved — sidecar now defaults to :8081, with a doc comment explaining why so it doesn't drift back to :8080. The concurrent-load sweep-skip caveat is now documented on `StartPeriodicTempDirSweep` as I'd flagged, and the sidecar metrics server timeouts are bounded per CodeRabbit's follow-up. CI is green across the board.

Approving.

matthyx:
rebase pls

bhuvan-somisetty:
Rebased onto latest \main\. All tests passing.

matthyx (dismissed):
Re-approving after the branch was rebased onto current `main` — verified via patch-id that the PR's own commits are unchanged content-wise (3/4 identical patch-ids, the 4th differs only in line-number context from unrelated commits that landed on `main` in the interim). CI is green.

matthyx:
;_; sorry again rebase

bhuvan-somisetty:
No worries! Rebased onto main and resolved the two conflicts (both in the pullMutex/temp-dir-liveness area where #690 landed in the meantime - kept both the pullMutex unlock and the BeginActiveTempDirUse liveness release in the deadline closure, and merged the metrics-server-shutdown + safer-os.Remove socket cleanup from cmd/sbom-scanner). Build/tests for everything this PR touches are green.

Heads up separately: main itself doesn't currently build - adapters/v1/domain_to_syft.go:73 calls a deduplicateErrors() that was never defined, from #703 (638069e^..085f60b). Confirmed it in a clean upstream/main checkout, unrelated to this PR. Figured you'd want to know since it'll break CI on anything rebasing through it.

cmd/http/main.go
```diff
@@ -181,20 +181,30 @@ func main() {
...
+	// anything to Cleanup() it, so auth failures, transient registry errors, and platform
+	// mismatches can all leak one. This process runs for the life of a long-lived pod rather
+	// than restarting per scan, so a single startup-only sweep would never reclaim leaks from
+	// its own uptime; run it on a recurring interval instead. The threshold is ScanTimeout: the
+	// main process and the SBOM-scanner sidecar run in the same Pod on the same machine (same
+	// kernel clock), so no clock-skew margin is needed, and a dir idle longer than ScanTimeout
+	// can only belong to a dead or already-timed-out scan. ScanTimeout also doubles as the sweep
+	// interval, keeping worst-case accumulation bounded to roughly one ScanTimeout's worth of
+	// leaks at a time.
+	tools.StartPeriodicTempDirSweep(ctx.Done(), os.TempDir(), "stereoscope-", c.ScanTimeout, c.ScanTimeout, func(removed int, err error) {
```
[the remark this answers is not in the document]
bhuvan-somisetty: Confirmed - both containers do mount the same tmp-dir volume at /tmp per the Helm chart, so this was a real cross-container race. Gave BeginActiveTempDirUse/StartPeriodicTempDirSweep a shared cross-process lease (flock on a lease file under the same dir, via gofrs/flock so it stays cross-platform) alongside the existing in-process counter - a crashed process's hold releases automatically via the kernel, so no orphaned-lock cleanup needed. Added regression tests that simulate the second process via an independent lock handle on the same file. Pushed in 67cf771.

matthyx (approved):
Checked the two new commits since my last approval — this round has real substance, not just a rebase.

**67cf771** adds a cross-process flock lease so `cmd/http` and `cmd/sbom-scanner` (separate containers sharing one `/tmp` volume per the Helm chart) don't step on each other's temp dirs — `activeTempDirUsers` alone can't see across the process boundary. **85a6976** then fixes a real race CodeRabbit caught in that same commit: the first version acquired the exclusive lock, checked it, and released it *before* calling `CleanupStaleTempDirs`, leaving a window where another process could grab the lease and start reading a directory the sweep was about to delete. The fix now holds the lock for the whole cleanup pass via a returned release func — correct.

Verified:
- `go test -race ./internal/tools/...` passes, including the new cross-process lease tests (they simulate the other process via an independent `*flock.Flock` handle on the same lease file, since flock semantics are per open-file-description, not per-process — a legitimate way to test this without spawning a real second process).
- All four call sites (`cmd/http`, `cmd/sbom-scanner`, `syft.go`, `server.go`) consistently pass `os.TempDir()`, so the lease file path actually matches across processes.
- Building the exact PR head in isolation currently fails (`adapters/v1/domain_to_syft.go: undefined: deduplicateErrors`) — not something this PR touches. It's rebased onto a point in `main`'s history between #704 (which dropped that helper) and #739 (which restored it), a transient breakage in `main` itself. Confirmed it's a non-issue for merging: building `main`'s current tip merged with this PR's head passes build/vet/full test suite cleanly, and that's what CI actually validates (GitHub tests the PR-into-base merge ref, which is why CI is green). Worth a rebase for hygiene before merge, but not a blocker.

CodeRabbit's one remaining nitpick (add a test asserting the sweep doesn't delete the lease file itself) is a reasonable but non-blocking suggestion — `CleanupStaleTempDirs` already filters on `IsDir()` so the lease file survives today regardless.

Approved.

PR #671, opened 2026-08-13
Fix/external vex masking

### LINKED ISSUE #670: Hardcoded SecurityException assertions in VEX generation will mask external VEX data
# Description
When scanning an image, Kubevuln takes any vulnerabilities that were ignored and writes them into the generated VEX document. However, the code blindly assumes that every ignored vulnerability was suppressed by a Kubescape `SecurityException` rule. 

This means if Grype suppresses a finding natively (for example, when we pass an external VEX feed to it using the `--vex` flag in the upcoming LFX project), the generated VEX document will falsely claim that a `SecurityException` was the reason. It completely masks the true origin of the suppression and creates a duplicate local statement.

# Environment
OS: `linux/amd64` (or any standard environment)
Version: `main` branch (latest commit)
 
# Steps To Reproduce
1. Run a scan where a vulnerability is ignored by a native Grype rule (or simulate it in a Go test). This produces a `CVEManifest` where the `IgnoredMatches` array contains a match, but the `AppliedIgnoreRules` inside it does not belong to a `SecurityException`.
2. Let `repositories/apiserver.go` generate the VEX document via `StoreVEX`.
3. Inspect the generated VEX statements in the cluster.
4. You will see that the `ImpactStatement` is hardcoded to `"Vulnerability was ignored by a SecurityException"`, even though no SecurityException was involved.

Alternatively, you can run this Go test to reproduce the issue locally:
```go
func TestStoreVEX_ExternalVEXMasking(t *testing.T) {
	ctx := context.Background()
	namespace := "default"
	store := NewFakeAPIServerStorage(namespace)

	// Simulate a CVE manifest where Grype suppressed a finding natively
	cveManifest := domain.CVEManifest{
		Name: "deployment-my-app",
		Content: &v1beta1.GrypeDocument{
				{
					Match: v1beta1.Match{
						Vulnerability: v1beta1.Vulnerability{
							VulnerabilityMetadata: v1beta1.VulnerabilityMetadata{ID: "CVE-2023-1234"},
						},
						Artifact: v1beta1.GrypePackage{PURL: "pkg:apk/alpine/curl@8.1.2-r0"},
					},
					AppliedIgnoreRules: []v1beta1.IgnoreRule{{Vulnerability: "CVE-2023-1234"}},
				},
			},
		},
		Annotations: map[string]string{"kubescape.io/image-id": "docker://alpine@sha256:abcd"},
	}

	store.StoreVEX(ctx, cveManifest, cveManifest, false)
	vexContainer, _ := store.StorageClient.OpenVulnerabilityExchangeContainers(namespace).Get(ctx, cveManifest.Name, metav1.GetOptions{})
	
	if vexContainer.Spec.Statements[0].ImpactStatement == "Vulnerability was ignored by a SecurityException" {
		t.Fatalf("BUG: Suppression was falsely attributed to a SecurityException")
	}
}
```

# Expected behavior
The generated VEX statement should reflect the true source of the suppression. If the vulnerability was ignored by Grype natively or by an external VEX feed, it should not falsely claim that a Kubescape `SecurityException` was responsible. 

# Actual Behavior
The code hardcodes the impact statement to `"Vulnerability was ignored by a SecurityException"` for every item in the `IgnoredMatches` list. 

# Additional context
This issue will directly affect the upcoming LFX Mentorship project for consuming external VEX feeds. If we pass external VEX feeds to Grype using the `--vex` flag, Grype will correctly suppress the findings, but Kubevuln will rewrite the history and claim a `SecurityException` did it.

To fix this, we can export and use the `isExceptionSourcedIgnore(im)` helper function (which currently lives in `adapters/v1/securityexception.go`) inside `apiserver.go` to safely check if the suppression actually came from a CRD before hardcoding the string.

h30s:
## Overview
This PR fixes a bug where `ignoredMatchAssessment()` blindly hardcodes `ImpactStatement: "Vulnerability was ignored by a SecurityException"` for any ignored vulnerability. This behavior falsely attributes native Grype suppressions (such as those from external VEX feeds) to CRD-driven SecurityExceptions, rewriting history and masking external data.

The fix updates `ignoredMatchAssessment()` to explicitly check `rule.SourceKind` to determine the suppression's true provenance:
* If the suppression originated from a CRD (`SourceKind == SecurityException`), it retains the original message.
* If the suppression did not originate from a CRD (e.g., native scanner config or external VEX), it emits `"Vulnerability was ignored by an external VEX document or scanner configuration"`.

## Related issues/PRs:
* Resolved #670

## How to Test
1. Observe the new unit test `TestStoreVEX_ExternalVEXMasking` in `repositories/apiserver_test.go`.
2. This regression test validates both code paths side-by-side, proving that an external VEX suppression receives the new statement while a genuine `SecurityException` retains the original statement.
3. Run `go test ./...` — all tests pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected impact statements for ignored matches based on their suppression source.
  * Matches from external VEX or scanner suppressions now display the appropriate impact statement.
  * SecurityException-based suppressions retain their distinct impact statement.
  * Improved handling of mixed ignored matches for more accurate results.


matthyx (approved):
Reviewed the fix and it correctly resolves #670.

- `ignoredMatchAssessment()` now defers the impact statement decision to `securityExceptionIgnoreRule()`, which inspects `AppliedIgnoreRules[].SourceKind`. I traced `SourceKind` back to `adapters/v1/securityexception.go` (`buildIgnoreRule`), where it's only ever set to `"SecurityException"` or `"ClusterSecurityException"` when a CRD actually produced the suppression — native Grype-produced `IgnoreRule`s (`grypeToDomainIgnoredMatchesAppliedIgnoreRules`) never populate it. So the `ok` check is a reliable signal of true provenance, and the new "external VEX document or scanner configuration" message only fires for non-CRD suppressions, as intended.
- The `FixState` repurposing for SecurityException status vocabulary is correctly gated behind `ok`, so it can't collide with Grype's native fix-state vocabulary on the external path.
- New regression test `TestStoreVEX_ExternalVEXMasking` covers both code paths side by side, and existing `ignoredMatches` tests were updated consistently with the new default message.
- Verified locally: `go build ./...` succeeds, `go test ./...` passes (all packages), and all PR CI checks are green.

No blockers. LGTM, approving.

PR #673, opened 2026-08-13
fix(repositories): index VEX statements instead of rescanning

### LINKED ISSUE #672: VEX update rescans every statement instead of using an indexed lookup
# Description
VEX generation and update in `repositories/apiserver.go` repeatedly rescan the full statement slice instead of using a lookup keyed by vulnerability and package. `markRelevantVulnerabilitiesAsAffectedInVex` and `markIgnoredVulnerabilitiesInVex` each loop over every incoming match and, for each one, scan every statement (and every product/subcomponent within it) looking for a match. `updateVEX` does the same thing again for its dedup checks before appending new statements. For an image with N tracked vulnerabilities and M incoming matches, this is effectively O(N*M) work on every scan, on the exact code path (VEX persistence) that already has to handle the largest manifests kubevuln produces.

A benchmark against a synthetic manifest with 5000 matches (`go test ./repositories/... -run '^$' -bench BenchmarkAPIServerStore_StoreVEX_LargeManifest -benchmem`) shows `StoreVEX`'s two-call create+update path dropping from ~938ms to ~560ms/op (allocation count unchanged) once the nested scans are replaced with a single indexed lookup, confirming the nested-loop cost is measurable at realistic scale.

# Environment
OS: `darwin/arm64`
Version: not applicable, this is a build-from-source code path issue rather than a shipped binary version issue

# Steps To Reproduce
1. Check out `repositories/apiserver.go` and look at `markRelevantVulnerabilitiesAsAffectedInVex`, `markIgnoredVulnerabilitiesInVex`, and the dedup loops inside `updateVEX`.
2. Note that each one scans `vexDoc.Statements` (and nested products/subcomponents) for every entry in `cve.Content.Matches` / `cve.Content.IgnoredMatches`.
3. Run a benchmark with a large synthetic `CVEManifest` (a few thousand matches) through `APIServerStore.StoreVEX` and observe CPU time scaling with the product of statement count and match count.

# Expected behavior
VEX update cost should be close to linear in the number of statements/findings, using a single pass to build a lookup that dedup, reset, and status-marking can all reuse.

# Actual Behavior
The code performs repeated full traversals of the statement slice (and nested products/subcomponents) for dedup, mark-affected, and mark-ignored, multiplying cost by the number of incoming matches.

# Additional context
Likely root cause: the VEX statement logic grew incrementally around append-and-rescan patterns rather than a keyed index built once per update.

Suggested direction: build a `vulnerability name + package PURL` index once per `createVEX`/`updateVEX` call (covering every product/subcomponent of a local statement, since a statement is not guaranteed to have exactly one) and reuse that index for dedup, mark-affected, and mark-ignored instead of rescanning. A benchmark exercising `StoreVEX` on a large synthetic manifest can guard against regressions.

rootp1:
## Overview
VEX generation and update in `repositories/apiserver.go` repeatedly rescanned every statement (and every product/subcomponent inside it) once per incoming match, in `markRelevantVulnerabilitiesAsAffectedInVex`, `markIgnoredVulnerabilitiesInVex`, and the dedup checks inside `updateVEX`. That made VEX persistence roughly quadratic in the number of tracked vulnerabilities, on the exact path that already has to handle the largest images kubevuln scans.

This PR builds a single `vulnerability name + package PURL` index over the local statements once per `createVEX`/`updateVEX` call and reuses it for dedup, mark-affected, and mark-ignored, instead of rescanning the statement slice for each match. The index covers every product/subcomponent of a local statement, not just the first, since a statement isn't guaranteed to carry exactly one.

A new benchmark, `BenchmarkAPIServerStore_StoreVEX_LargeManifest`, exercises `StoreVEX` against a synthetic 5000-match manifest through both the create and update paths. It goes from about 938ms/op to about 560ms/op with this change, with the same allocation count, confirming the win comes from removing the repeated scans rather than incidental changes.

## How to Test
```
go test ./repositories/...
go test ./repositories/... -run '^$' -bench BenchmarkAPIServerStore_StoreVEX_LargeManifest -benchmem
```

## Related issues/PRs:
Resolved #672

## Summary by CodeRabbit

* **Performance**
  * Improved VEX statement matching and deduplication for faster processing of large vulnerability manifests.
  * Optimized initial VEX creation and subsequent updates, including repeated status-assignment operations.
  * Improved handling of product subcomponents and newly added statements during processing.

* **Bug Fixes**
  * Corrected processing so duplicate local VEX statements are consistently updated when they match the same vulnerability and package.
  * Prevented external statements from being included in local matching results.


repositories/apiserver.go
```diff
@@ -1171,38 +1171,54 @@ func isLocalStatement(id string) bool {
...
+// treat that as a valid shape to match against.
+func buildLocalVexStatementIndex(statements []v1beta1.Statement) map[vexStatementKey]int {
+	idx := make(map[vexStatementKey]int, len(statements))
+	for i, s := range statements {
+		if !isLocalStatement(s.ID) {
+			continue
+		}
+		for _, p := range s.Products {
+			for _, sc := range p.Subcomponents {
+				idx[vexStatementKey{name: s.Vulnerability.Name, purl: sc.ID}] = i
```
matthyx: **Blocker: the index collapses duplicate local statements, dropping updates the pre-PR code applied to all of them.**

`idx[key] = i` overwrites on collision, so when two local statements share the same `(Vulnerability.Name, PURL)` key, only the *last* one in `statements` survives in the index. `markRelevantVulnerabilitiesAsAffectedInVex` / `markIgnoredVulnerabilitiesInVex` then only update that one statement.

Pre-PR, `markRelevantVulnerabilitiesAsAffectedInVex`'s inner loop had no `break` on the outer `for i, s := range vexDoc.Statements` — it updated *every* matching local statement, not just the first/last. So documents that already contain duplicate local statements for the same vulnerability+package get updated inconsistently after this change: one copy is promoted to `Affected`/ignored, the sibling is left stuck at whatever the "reset to baseline" pass wrote (`NotAffected`).

This isn't hypothetical for this codebase — the ID/Name-swap migration code a few lines above this hunk (`// Statements written before the ID/Name mapping was corrected...`) and `TestAPIServerStore_updateVEX_normalizesLegacyStatements` both exist specifically because production VEX documents have accumulated duplicate local statements for the same (vulnerability, PURL) in the past. That migration only prevents *new* duplicates going forward — it doesn't merge/remove ones that already exist in already-persisted documents. Any such existing document will now silently stop getting one of its duplicate statements promoted.

I verified this concretely: I checked out this PR branch and the PR's base commit (`770090a`) locally, and added a test that (1) creates a VEX doc, (2) duplicates one local statement in place (simulating a pre-existing doc with a collision), (3) reruns `StoreVEX` with that vulnerability marked relevant/ignored.
- On base `main` (770090a): both duplicate statements end up `Affected` (2/2) — matches the no-break behavior described above.
- On this PR branch: only 1/2 gets promoted to `Affected`; the other stays `not_affected` with `vulnerable_code_not_present`.

Suggested fix: make the index tolerant of duplicates, e.g. `map[vexStatementKey][]int`, and have both mark-functions loop over all indices for a key instead of just one — that restores the pre-PR "update every matching local statement" semantics while keeping the O(1) lookup win. A regression test using two local statements sharing a key (rather than assuming the key is always unique) would also catch this going forward.

matthyx (changes requested):
Nice performance win overall — I built the branch, ran the full `repositories` test suite (all green) and the new `BenchmarkAPIServerStore_StoreVEX_LargeManifest` locally, and the quadratic→linear improvement is real and clearly demonstrated by the benchmark.

However, I found one correctness blocker in how the new index handles duplicate keys — see the inline comment on `buildLocalVexStatementIndex`. In short: `map[vexStatementKey]int` silently drops all but one statement when two local statements share a `(vulnerability, PURL)` key, whereas the pre-PR scan updated every matching statement. I reproduced this locally against both this branch and the PR's base commit to confirm it's a real regression, not a pre-existing issue (details in the inline comment).

Given the ID/Name-swap migration code already in this file exists specifically because such duplicates have shown up in real persisted VEX documents, I'd like to see this addressed (or explicitly ruled out with more certainty) before merging.

matthyx (approved):
Blocker is fixed. `buildLocalVexStatementIndex` now returns `map[vexStatementKey][]int` and both mark-functions loop over all indices for a key, so duplicate local statements sharing a (vulnerability, PURL) key are all updated again, matching pre-PR behavior. The new `TestAPIServerStore_storeVEX_duplicateLocalStatementsBothUpdated` covers exactly this.

I re-verified independently: pulled commit `b712b64`, built, ran the full `repositories` suite (pass), and reran my own local reproduction from the previous review (duplicate-statement collision) against this commit — both duplicates are now promoted to `Affected` (2/2), same as base `main`.

The remaining open CodeRabbit nit (product-level `Component.ID` PURL test coverage) is about a gap that predates this PR — `anyPURLMatches`/`statementHasPURL` only ever checked `Subcomponents`, same as before — so it's not a regression introduced here and doesn't need to block merge.

Approving.

PR #675, opened 2026-08-13
perf(sbom): index the Syft metadata reattachment instead of scanning

### LINKED ISSUE #674: Syft metadata reattachment is quadratic in the package count, and duplicated across both SBOM paths
converting a Syft SBOM into the storage types pairs each stored package with its Syft counterpart by scanning the whole document for a matching ID:

```go
for i := range syftSBOM.Artifacts {
	for j := range doc.Artifacts {
		if syftSBOM.Artifacts[i].ID == doc.Artifacts[j].ID {
			syftSBOM.Artifacts[i].MetadataType = doc.Artifacts[j].MetadataType
			...
			break
		}
	}
}
```

that is quadratic in the package count, and it runs on every SBOM that gets created. benchmarking the loop against an indexed version on identical input:

```
packages   scan        indexed
500        0.55 ms     0.19 ms
2000       4.50 ms     0.49 ms
8000     281.40 ms    12.90 ms
```

500 packages is a small image and the cost is nothing. a Java or Node image in the low thousands is ordinary, and at 8000 the scan is 281 ms of pure comparison per SBOM, against 12.9 ms for the same work indexed. the remaining 12.9 ms is the `json.Marshal` of each package's metadata, which is inherent.

the loop exists because `v1beta1.SyftPackage` holds `Metadata` as a `json.RawMessage`, so the metadata does not survive the JSON round-trip into the storage types and has to be put back afterwards. that part is necessary; searching the whole document for each package is not.

there are two copies of it, one per SBOM path:

- `adapters/v1/syft_to_domain.go`, the in-process `SyftAdapter`
- `pkg/sbomscanner/v1/server.go`, the sidecar scanner's gRPC server

they are identical apart from the sidecar's dropping the warning when a package's metadata will not marshal. that pair of near-duplicates has produced three bugs already through drifting apart (#585, #587, #601), and `internal/` is importable from both, so one copy would do.

version: main (ea2fcfc)

AnshulPatil2005:
## Overview

putting each package's metadata back after the JSON round-trip into the storage types paired the two lists by scanning the whole Syft document for every stored package. that is quadratic in the package count and runs on every SBOM created.

against the same input, indexed by ID instead:

```
packages   scan        indexed
500        0.55 ms     0.19 ms
2000       4.50 ms     0.49 ms
8000     281.40 ms    12.90 ms
```

500 packages costs nothing either way. a Java or Node image in the low thousands is ordinary, and at 8000 the pairing alone was 281 ms per SBOM. the 12.9 ms left is the `json.Marshal` of each package's metadata, which is the actual work.

the loop itself is needed: `v1beta1.SyftPackage` holds `Metadata` as a `json.RawMessage`, so it does not survive being decoded and has to be reattached. only the search is replaced.

## Additional Information

there were two copies, one per SBOM path (`adapters/v1/syft_to_domain.go` for the in-process adapter, `pkg/sbomscanner/v1/server.go` for the sidecar), identical apart from the sidecar's dropping the warning when metadata will not marshal. this puts one copy in `internal/syftmeta` and has both call it. `adapters/v1` imports `pkg/sbomscanner/v1`, so the shared code cannot live in either of them, but `internal/` is already imported by both.

that de-duplication is the reason for a new package rather than the same fix written twice: those two paths drifting apart has caused three bugs on its own (#585, #587, #601).

pairing stays by ID rather than by position. the two lists are aligned in practice, since one is marshalled from the other, but nothing in the round-trip promises an order and a positional pairing would attach the wrong metadata if that ever stopped holding. on a duplicate ID the first Syft package wins, which is what the search did.

behaviour is otherwise unchanged, including leaving a stored package alone when the document has no counterpart, and keeping its existing metadata when marshalling fails. the sidecar gains the warning it was missing on that last case.

## How to Test

`go test ./internal/syftmeta/ ./adapters/v1/ ./pkg/sbomscanner/... -count=1`

six tests in the new package cover the pairing, ID-not-position pairing, unmatched packages, duplicate IDs, empty inputs, and unmarshalable metadata.

for the numbers above:

```
go test ./internal/syftmeta/ -run XXX -bench BenchmarkReattach -benchtime 30x
```

the existing adapter and sidecar tests are untouched and still pass.

## Related issues/PRs:

* Resolved #674

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved restoration of package metadata when converting scan results.
  - Ensured metadata remains correctly associated by package ID, even when package ordering differs or duplicate IDs exist.
  - Preserved existing metadata for unmatched packages or invalid metadata values.

- **Tests**
  - Added coverage for matching, duplicates, empty inputs, unmatched packages, and invalid metadata.
  - Added benchmarks for processing scans containing hundreds to thousands of packages.

### DIFF
--- a/adapters/v1/syft_to_domain.go
+++ b/adapters/v1/syft_to_domain.go
@@ -5,8 +5,7 @@ import (
 
 	"github.com/anchore/syft/syft/format/syftjson"
 	"github.com/anchore/syft/syft/sbom"
-	"github.com/kubescape/go-logger"
-	"github.com/kubescape/go-logger/helpers"
+	"github.com/kubescape/kubevuln/internal/syftmeta"
 	"github.com/kubescape/storage/pkg/apis/softwarecomposition/v1beta1"
 )
 
@@ -25,19 +24,7 @@ func (s *SyftAdapter) syftToDomain(sbomSBOM sbom.SBOM) (*v1beta1.SyftDocument, e
 	if err := json.Unmarshal(b, &syftSBOM); err != nil {
 		return nil, err
 	}
-	for i := range syftSBOM.Artifacts {
-		for j := range doc.Artifacts {
-			if syftSBOM.Artifacts[i].ID == doc.Artifacts[j].ID {
-				syftSBOM.Artifacts[i].MetadataType = doc.Artifacts[j].MetadataType
-				if b, err := json.Marshal(doc.Artifacts[j].Metadata); err == nil {
-					syftSBOM.Artifacts[i].Metadata = b
-				} else {
-					logger.L().Warning("failed to marshal Artifacts[j].Metadata", helpers.Error(err))
-				}
-				break
-			}
-		}
-	}
+	syftmeta.Reattach(syftSBOM.Artifacts, doc.Artifacts)
 
 	return syftSBOM, nil
 }
--- a/internal/syftmeta/reattach.go
+++ b/internal/syftmeta/reattach.go
@@ -1,0 +1,52 @@
+// Package syftmeta carries the one step that has to happen after a Syft SBOM is round-tripped
+// through JSON into the storage types: putting each package's metadata back.
+package syftmeta
+
+import (
+	"encoding/json"
+
+	"github.com/anchore/syft/syft/format/syftjson/model"
+	"github.com/kubescape/go-logger"
+	"github.com/kubescape/go-logger/helpers"
+	"github.com/kubescape/storage/pkg/apis/softwarecomposition/v1beta1"
+)
+
+// Reattach copies MetadataType and Metadata from the Syft document onto the matching stored
+// packages, which the JSON round-trip that produced them drops: v1beta1.SyftPackage holds
+// Metadata as a json.RawMessage, so it does not survive being decoded into a typed field.
+//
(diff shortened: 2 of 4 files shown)


matthyx (approved):
Reviewed the diff (`adapters/v1/syft_to_domain.go`, `internal/syftmeta/reattach.go`+test, `pkg/sbomscanner/v1/server.go`) and confirmed:

- `Reattach` preserves the original semantics exactly: `MetadataType` is always overwritten on a match, `Metadata` only on a successful marshal (so a marshal failure keeps the old metadata but the new type — same as the code it replaces), and on duplicate IDs the first document package wins, matching the old scan-with-`break`.
- The ID index turns the pairing from O(n·m) into O(n), which is the actual perf win claimed; edge cases (empty inputs, unmatched stored packages, duplicate IDs) are all covered by tests.
- Both call sites (`adapters/v1` in-process and `pkg/sbomscanner/v1` sidecar) now share one implementation — good fix for the drift that caused #585/#587/#601. The sidecar picks up the missing marshal-failure warning as an intentional, documented behavior change.
- All CI checks are green (build, lint, tests, CodeQL, security/license scans), and `gh pr view` confirms the PR is actually mergeable (no real conflict — the earlier CodeRabbit conflict note was against a prior commit before the branch was updated).

No blockers. LGTM.

PR #679, opened 2026-08-13
fix(tools): bound a Retry-After so a registry cannot park a scan

### LINKED ISSUE #677: a registry's Retry-After is honoured unbounded, so one 429 can stall scanning
`RetryWithBackoff` caps the backoff it calculates, but not a delay it takes from a `Retry-After`:

```go
delay := wait
	delay = retryAfter
} else {
	if delay > 0 {
		jitter := time.Duration(rand.Float64() * 0.25 * float64(delay))
		delay += jitter
	}
	if delay > config.MaxWait {
		delay = config.MaxWait
	}
}
```

the `MaxWait` clamp is inside the `else`. a `Retry-After` goes to `time.After(delay)` as it stands, and that number comes from the registry's 429 response, so it is not ours to choose.

with `Default429RetryConfig` (`MaxAttempts: 3`, `MaxWait: 2s`) and an error carrying `Retry-After: 3`:

```
config MaxWait=2s MaxAttempts=3
attempts=3 elapsed=6.0008281s
```

six seconds against a two second ceiling, from a header asking for three. `Retry-After: 86400` gets 48 hours the same way.

nothing else on the path stops it:

- the scan context is `context.WithoutCancel(...)` in `controllers/http.go`, and go's `WithoutCancel` returns a context with no deadline
- `adapters/v1/sidecar.go` calls the RPC on that context with no `WithTimeout`, and `pkg/sbomscanner/v1/client.go` hands it to gRPC unchanged
- the sidecar's own `scanTimeout` is `deadline.New(timeout)` further down `CreateSBOM`, wrapping SBOM generation, while this retry is source resolution and runs before it

so the `case <-ctx.Done()` arm of the select never fires and the sleep runs to completion, holding the gRPC handler and the kubevuln worker-pool slot behind it. `scanConcurrency` defaults to 1, so one registry answering 429 with a large `Retry-After` stops scanning for as long as it likes.

`ParseRetryAfter` accepts any positive integer, and the RFC1123 branch takes `time.Until` of any future date, so there is no bound on either form.

a ceiling on the value is enough. capping at `MaxWait` would work but leaves the header nearly useless at 2s, so something like a separate bound in `RetryConfig`, comfortably under the scan timeout, keeps it meaningful.

version: main (770090a), follow-up to #655

AnshulPatil2005:
## Overview

`RetryWithBackoff` clamped the backoff it calculates to `MaxWait`, but the clamp sat inside the `else`, so a delay taken from a `Retry-After` went to `time.After` verbatim. that number comes from the registry's 429 response.

nothing else on the path bounds it. the scan context is `context.WithoutCancel(...)`, which in go carries no deadline; `adapters/v1/sidecar.go` calls the RPC on it with no `WithTimeout` and the client hands it to gRPC unchanged; and the sidecar's `scanTimeout` wraps SBOM generation further down `CreateSBOM`, while this retry is source resolution and runs before it. so `case <-ctx.Done()` never fires, the sleep runs to completion, and the worker-pool slot behind it is held for the duration. `scanConcurrency` defaults to 1.

with the default config, `MaxWait: 2s` and 3 attempts, an error carrying `Retry-After: 3` took 6s. `Retry-After: 86400` takes 48 hours the same way.

this adds `MaxRetryAfter` to `RetryConfig`, defaulted to 30s, and caps the header at it.

## Additional Information

capping at `MaxWait` instead would have been a smaller change, but at 2s it makes reading the header close to pointless, and honouring a registry's backoff is the reason #655 read it in the first place. 30s keeps it useful, and two waits at that ceiling is a small fraction of the 5 minute default scan timeout. it is one constant to change if you want it longer or shorter.

a config that never sets `MaxRetryAfter` falls back to `MaxWait` rather than to no bound, so a zero value cannot reintroduce the same thing, and there is a test for that.

the cap is logged at debug when it bites, so a registry asking for much longer than we are willing to wait is visible rather than silent.

## How to Test

`go test ./internal/tools/ -count=1`

- `TestRetryWithBackoff_CapsRetryAfter` asks for `Retry-After: 2` against a 20ms ceiling and asserts the whole run stays under 500ms. before this it takes 4s.
- `TestRetryWithBackoff_UnsetCeilingFallsBackToMaxWait` does the same with `MaxRetryAfter` unset. before this it takes 2s.
- `TestRetryWithBackoff_HonoursRetryAfterUnderTheCeiling` asks for `Retry-After: 1` under a 1 minute ceiling and asserts it is still waited out, so the cap does not quietly disable the header.

the existing retry tests, including the two metric ones from #655, are untouched and still pass.

## Related issues/PRs:

* Resolved #677
* Follow-up to #655

## Summary by CodeRabbit

* **Improvements**
  * Added a configurable maximum wait time for server-provided retry delays.
  * Retry delays exceeding the configured limit are now capped, preventing unexpectedly long waits.
  * Existing behavior remains unchanged when the new limit is not configured.

* **Tests**
  * Added coverage for capped, honored, and default retry-delay scenarios.

### DIFF
--- a/internal/tools/retry.go
+++ b/internal/tools/retry.go
@@ -21,15 +21,32 @@ type RetryConfig struct {
 	InitialWait time.Duration
 	MaxWait     time.Duration
 	Backoff     float64
+	// MaxRetryAfter bounds a delay taken from a registry's Retry-After. That number is
+	// chosen by the registry rather than by us, and nothing else on this path stops the
+	// clock: the scan context is built with context.WithoutCancel, so it carries no
+	// deadline, and the sidecar's own scanTimeout only wraps SBOM generation, which comes
+	// after this. A ceiling well under the scan timeout keeps the header useful without
+	// letting one response decide how long a worker is held. Non-positive falls back to
+	// MaxWait.
+	MaxRetryAfter time.Duration
 }
 
 func Default429RetryConfig() RetryConfig {
 	return RetryConfig{
-		MaxAttempts: 3,
-		InitialWait: 500 * time.Millisecond,
-		MaxWait:     2 * time.Second,
-		Backoff:     2.0,
+		MaxAttempts:   3,
+		InitialWait:   500 * time.Millisecond,
+		MaxWait:       2 * time.Second,
+		Backoff:       2.0,
+		MaxRetryAfter: 30 * time.Second,
 	}
+}
+
+// retryAfterCeiling is the largest delay a Retry-After is allowed to ask for.
+func (c RetryConfig) retryAfterCeiling() time.Duration {
+	if c.MaxRetryAfter > 0 {
+		return c.MaxRetryAfter
+	}
+	return c.MaxWait
 }
 
 func IsRateLimitError(err error) bool {
@@ -108,6 +125,12 @@ func RetryWithBackoff[T any](ctx context.Context, operation string, config Retry
 		delay := wait
 		if retryAfter, ok := ParseRetryAfter(err); ok {
 			delay = retryAfter
+			if ceiling := config.retryAfterCeiling(); delay > ceiling {
+				logger.L().Ctx(ctx).Debug("capping Retry-After to the configured ceiling",
+					helpers.String("requested", retryAfter.String()),
+					helpers.String("ceiling", ceiling.String()))
+				delay = ceiling
+			}
 		} else {
 			if delay > 0 {
 				jitter := time.Duration(rand.Float64() * 0.25 * float64(delay))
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed the diff, the full resulting `retry.go`, both call sites, and CI (all green, CodeRabbit found nothing actionable).

The fix is correctly scoped:
- `retryAfterCeiling()` falls back to `MaxWait` when `MaxRetryAfter` is unset/non-positive, so no config can silently reintroduce the unbounded wait — matches `TestRetryWithBackoff_UnsetCeilingFallsBackToMaxWait`.
- The cap sits inside the shared `if retryAfter, ok := ParseRetryAfter(err); ok` block, so it applies to both the numeric-seconds and RFC1123-date parse paths, not just the common case.
- Both real call sites (`core/services/scan.go` for `sbom_generation`, `pkg/sbomscanner/v1/server.go` for `source_resolution`) go through `Default429RetryConfig()`, so both automatically pick up the new 30s ceiling — no caller needed a separate change.
- Tests cover capped, honored-under-ceiling, and fallback-to-MaxWait cases, and the numbers in each test back up the assertions (e.g. 20ms ceiling vs. a 2s requested delay, asserting well under 500ms elapsed).

One non-blocking edge case for awareness, not a request for changes: if a caller ever constructs a `RetryConfig` with both `MaxWait` and `MaxRetryAfter` left at zero, `retryAfterCeiling()` returns 0 and any `Retry-After` gets capped straight to a no-wait retry. Doesn't affect `Default429RetryConfig()` or either current caller, just worth knowing if a new zero-value config shows up later.

LGTM — approving.

PR #682, opened 2026-08-13
fix(repositories): adopt the VEX statements we wrote before #595

### LINKED ISSUE #681: the VEX statements kubevuln wrote before #595 are now treated as external, so they freeze and get duplicated
#664 changed `isLocalStatement` to stop treating an empty ID as ours:

```go
-	return id == "" || strings.HasPrefix(id, "https://kubescape.io/vex/statement/")
+	return strings.HasPrefix(id, "https://kubescape.io/vex/statement/")
```

that is right for what it set out to fix: OpenVEX makes statement IDs optional, so a feed can ship statements without one and those must not be overwritten.

the trouble is that statements kubevuln itself wrote before #595 also have no ID. #595 is the change that started setting one, `createVEX` had no `ID` field before it:

```
$ git show dc070b6^:repositories/apiserver.go
1031:			vexDoc.Statements = append(vexDoc.Statements, v1beta1.Statement{
1032-				Vulnerability: v1beta1.VexVulnerability{
```

that is why #595's `isLocalStatement` had the `id == ""` clause, and why it came with a test for a "local legacy statement (empty ID)". #664 removed the clause and replaced that test case with an external one.

so on any cluster that ran kubevuln before #595, every statement already in storage is now read as another author's data:

- the reset loop skips it, so it keeps whatever status it had when it was written
- `markRelevantVulnerabilitiesAsAffectedInVex` and `markIgnoredVulnerabilitiesInVex` skip it
- the ID/Name normalization skips it, that being `isLocalStatement`-guarded too
- the dedup skips it (#633), so kubevuln appends a fresh statement for the same finding

taking one statement from a stored document, clearing its ID the way a pre-#595 one would be, and running `StoreVEX` again leaves two statements for the same vulnerability and package:

```
{                                    ... CVE-2005-2541 ... tar@1.29b-1.1 ... affected      ...}
{https://kubescape.io/vex/statement/CVE-2005-2541/... tar@1.29b-1.1 ... not_affected  ...}
```

the first is the orphaned one, still carrying the status it was frozen at; the second is the one kubevuln just added. the document grows a stale copy of every finding it had, and the stale copies say whatever they said at the time.

the two cases are only separable by what is on the statement. a pre-#595 one carries an impact statement kubevuln wrote itself, either `"Vulnerable component is not loaded into the memory"` or `"Vulnerability was ignored by a SecurityException"`, and those strings are still the constants used today. adopting an ID-less statement that carries one of them, and leaving every other ID-less statement alone, covers the old data without weakening what #664 added.

version: main (770090a), follow-up to #664

AnshulPatil2005:
## Overview

#664 stopped `isLocalStatement` reading an empty ID as ours, so that a feed's ID-less statement is not overwritten. that is right, and this keeps it.

statements kubevuln wrote before #595 also have no ID though, #595 being the change that started setting one. so on any cluster that ran kubevuln before then, everything already in storage now reads as another author's data: skipped by the reset loop, by both marking steps, by the ID/Name normalization, and by the dedup, which appends a fresh statement beside it for the same finding. the document ends up with a stale copy of every finding it had, each frozen at whatever status it was written with.

the two cases are separable by what is on the statement. one of ours carries an impact statement we wrote, `"Vulnerable component is not loaded into the memory"` or `"Vulnerability was ignored by a SecurityException"`, and both are still the constants used today. this stamps the ID we would write now onto ID-less statements carrying one of them, before the local/external split, so they come back under management. every other ID-less statement is left exactly as #664 leaves it.

## Additional Information

adoption is deliberately narrow. it needs an empty ID, one of our two impact statements, and a product with a subcomponent to build the ID from. a feed would have to reproduce our own wording word for word to be picked up, and #664's test case, whose impact statement is `"External feed assessed this one (no ID)"`, is untouched and still passes.

the ID format was written out at four call sites, so it is now `localStatementID`, which the adoption also uses. that keeps the stamped ID identical to the one `createVEX` writes rather than a second copy of the format string that could drift from it.

this is only reachable through `updateVEX`, so a fresh document is unaffected, and adoption is a no-op once a statement has an ID.

## How to Test

`TestAPIServerStore_updateVEX_adoptsPre595StatementsWithoutIDs` takes a statement from a stored document, reshapes it the way it would have been stored before #595 (no ID, our impact statement, a stale status), and runs `StoreVEX` again. it asserts one statement for that vulnerability and package rather than two, and that it has been reset to baseline.

before this it fails with the duplicate:

```
{                                    ... CVE-2005-2541 ... tar@1.29b-1.1 ... affected     ...}
{https://kubescape.io/vex/statement/CVE-2005-2541/... tar@1.29b-1.1 ... not_affected ...}
should have 1 item(s), but has 2
```

`TestAPIServerStore_updateVEX_preservesExternalStatements` from #664 and `TestAPIServerStore_updateVEX_externalStatementDoesNotSuppressLocalOne` from #633 are both untouched and still pass.

## Related issues/PRs:

* Resolved #681
* Follow-up to #664

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of previously created VEX statements that do not have IDs.
  - Existing locally authored vulnerability statements are now recognized and reused instead of duplicated.
  - Statement identifiers are assigned consistently during creation and updates.
  - Legacy vulnerability field mappings are recognized and normalized automatically.
  - Adopted legacy statements are reset to the current baseline status during subsequent updates, ensuring displayed vulnerability information remains accurate.

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -1198,9 +1198,68 @@ func applyIgnoredMatchAssessment(stmt *v1beta1.Statement, assessment ignoredVEXA
 // to defaultActionStatement.
 func buildActionStatement(v v1beta1.Match) string {
 	if v.Vulnerability.Fix.State == "fixed" && len(v.Vulnerability.Fix.Versions) > 0 {
-		return fmt.Sprintf("Upgrade %s to version %s", v.Artifact.PURL, strings.Join(v.Vulnerability.Fix.Versions, " or "))
+		return upgradeActionStatementPrefix(v.Artifact.PURL) + strings.Join(v.Vulnerability.Fix.Versions, " or ")
 	}
 	return defaultActionStatement
+}
+
+// hasSwappedVulnerabilityFields reports whether a statement carries the CVE in
+// Vulnerability.ID and the data source URL in Name, the mapping used before it was corrected
+// to match createVEX. Both the adoption below and the normalization that follows it key off
+// this, so they cannot disagree about which statements are written the old way round.
+func hasSwappedVulnerabilityFields(v v1beta1.VexVulnerability) bool {
+	return !strings.Contains(v.ID, "://") && (v.Name == "" || strings.Contains(v.Name, "://"))
+}
+
+// vulnerabilityCVEName returns the CVE identifier a statement carries, from whichever field
+// holds it. Statements written before #595 can be affected by both legacy shapes at once, no
+// ID and the swapped mapping, and stamping an ID needs the CVE rather than whatever happens
+// to be in Name.
+func vulnerabilityCVEName(v v1beta1.VexVulnerability) string {
+	if hasSwappedVulnerabilityFields(v) {
+		return v.ID
+	}
+	return v.Name
+}
+
+// upgradeActionStatementPrefix is the fixed part of the action statement
+// buildActionStatement writes when the fix versions are known. It is shared with
+// isOwnActionStatement so the two cannot drift on what our own wording looks like.
+func upgradeActionStatementPrefix(purl string) string {
+	return fmt.Sprintf("Upgrade %s to version ", purl)
+}
+
+// isOwnActionStatement reports whether an action statement is one kubevuln writes itself.
+//
+// markRelevantVulnerabilitiesAsAffectedInVex blanks the impact statement on a statement it
+// marks affected and fills this in instead, so on those this is the only wording of ours
+// left to recognise. Action statements arrived in #404, ten days before the IDs in #595, so
+// a statement written in between has one of these and no ID.
+//
+// The parameterised form is matched against the statement's own subcomponent rather than as
+// a loose prefix, so a feed's text that happens to open the same way does not qualify.
+func isOwnActionStatement(actionStatement, purl string) bool {
+	if actionStatement == defaultActionStatement {
+		return true
+	}
+	return purl != "" && strings.HasPrefix(actionStatement, upgradeActionStatementPrefix(purl))
+}
+
+// localStatementID is the ID kubevuln stamps on statements it authors.
+func localStatementID(cveName, purl string) string {
+	return fmt.Sprintf("https://kubescape.io/vex/statement/%s/%s", url.PathEscape(cveName), url.PathEscape(purl))
+}
+
+// isOwnImpactStatement reports whether an impact statement is one kubevuln writes itself.
+// It is the only thing left on a statement we stored before #595 that marks it as ours,
+// those having been written without an ID, and the wording is our own rather than anything
+// a feed would produce.
+func isOwnImpactStatement(impactStatement string) bool {
+	switch impactStatement {
+	case defaultLocalImpactStatement, securityExceptionImpactStatement:
+		return true
+	}
+	return false
 }
 
 func isLocalStatement(id string) bool {
@@ -1359,6 +1418,31 @@ func (a *APIServerStore) updateVEX(ctx context.Context, cve domain.CVEManifest, 
 	originalVEX := *vexContainer.Spec.DeepCopy()
 	vexDoc := *vexContainer.Spec.DeepCopy()
 
+	// Statements we stored before #595 have no ID, that being the change which started
+	// setting one. #664 stopped reading an empty ID as ours so that a feed's ID-less
+	// statement is not overwritten, which is right, but it also left every statement we
+	// wrote before #595 looking like another author's: passed over by the reset loop and by
+	// every marking step, and passed over by the dedup too, so a second statement gets
+	// appended beside it for the same finding. Stamp the ID we would write today onto the
+	// ones carrying wording of ours, which brings them back under management and leaves
+	// anything else alone. Both fields are checked: a statement last written while affected
+	// had its impact statement blanked and an action statement put in its place, so that is
+	// the only wording left on it.
+	for i := range vexDoc.Statements {
+		s := &vexDoc.Statements[i]
+		if s.ID != "" {
+			continue
+		}
+		if len(s.Products) == 0 || len(s.Products[0].Subcomponents) == 0 {
+			continue
+		}
+		purl := s.Products[0].Subcomponents[0].ID
+		if !isOwnImpactStatement(s.ImpactStatement) && !isOwnActionStatement(s.ActionStatement, purl) {
+			continue
+		}
+		s.ID = localStatementID(vulnerabilityCVEName(s.Vulnerability), purl)
+	}
+
 	// Statements written before the ID/Name mapping was corrected to match createVEX
 	// carry the CVE identifier in ID and the data source URL in Name. Normalize them in
 	// place so the dedup below (which now keys on Name) also finds these older entries,
@@ -1367,7 +1451,7 @@ func (a *APIServerStore) updateVEX(ctx context.Context, cve domain.CVEManifest, 
 		if !isLocalStatement(s.ID) {
 			continue
 		}
-		if !strings.Contains(s.Vulnerability.ID, "://") && (s.Vulnerability.Name == "" || strings.Contains(s.Vulnerability.Name, "://")) {
+		if hasSwappedVulnerabilityFields(s.Vulnerability) {
 			vexDoc.Statements[i].Vulnerability.ID, vexDoc.Statements[i].Vulnerability.Name = s.Vulnerability.Name, s.Vulnerability.ID
 		}
 	}
--- a/repositories/apiserver_external_test.go
+++ b/repositories/apiserver_external_test.go
@@ -260,3 +260,246 @@ func TestNewLocalStatement(t *testing.T) {
 	assert.Empty(t, stmt.ActionStatement)
 	assert.Empty(t, stmt.StatusNotes)
 }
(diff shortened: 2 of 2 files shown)


repositories/apiserver.go
```diff
@@ -1366,6 +1383,25 @@ func (a *APIServerStore) updateVEX(ctx context.Context, cve domain.CVEManifest,
...
+	// leaves anything else alone.
+	for i := range vexDoc.Statements {
+		s := &vexDoc.Statements[i]
+		if s.ID != "" || !isOwnImpactStatement(s.ImpactStatement) {
+			continue
+		}
+		if len(s.Products) == 0 || len(s.Products[0].Subcomponents) == 0 {
+			continue
+		}
+		s.ID = localStatementID(s.Vulnerability.Name, s.Products[0].Subcomponents[0].ID)
```
matthyx: This loop runs *before* the ID/Name normalization loop right below it (the one added for statements written before the ID/Name mapping was corrected to match `createVEX`, which swaps `Vulnerability.ID`/`Vulnerability.Name` back when `Vulnerability.Name` still holds a data-source URL).

That normalization loop is gated on `isLocalStatement(s.ID)`, so before this PR it never touched pre-#595 statements (they had no `s.ID` at all). Now that this loop stamps an ID first, a statement affected by *both* legacy bugs — no top-level ID **and** the old swapped `Vulnerability.ID`/`Vulnerability.Name` — gets its ID computed from `s.Vulnerability.Name` while that field still holds the data-source URL, not the CVE. Concretely:

1. Adoption: `s.ID = localStatementID(s.Vulnerability.Name /* = data-source URL here */, purl)` → malformed ID.
2. Now `isLocalStatement(s.ID)` is true, so the normalization loop fires and swaps `Vulnerability.ID`/`Vulnerability.Name` back to the correct values.
3. End state: `Vulnerability.Name` is now correct, but the statement's `ID` was already stamped from the wrong value and is never recomputed (adoption is a no-op once `ID != ""`), so it permanently diverges from what `createVEX` would generate for the same CVE/package — the exact invariant this PR's description calls out as intentional ("keeps the stamped ID identical to the one `createVEX` writes").

Since nothing else in this file compares `s.ID` for equality (only `isLocalStatement`'s prefix check), this doesn't break kubevuln's own reset/dedup logic, but it does mean the emitted OpenVEX document can carry a statement `id` that doesn't encode the CVE it describes for this specific double-legacy case, and the new test (which builds `legacy` directly from a statement whose `Vulnerability.ID`/`Name` were never swapped) doesn't exercise it.

Given the whole point of this PR is reaching statements old enough to predate #595, and the ID/Name swap is an older bug from around the same era, this combination seems plausible on a long-lived cluster. Could you either run the ID/Name normalization pass before stamping IDs here (and broaden its guard so it also covers `ID == ""` statements that are about to be adopted), or derive the ID from whichever of `Vulnerability.ID`/`Vulnerability.Name` doesn't look like a URL? Want to make sure this isn't a real gap before it merges.

matthyx (changes requested):
Nice fix, and the write-up makes the two historical bugs and their interaction easy to follow. The narrow-adoption gate (empty ID + our exact impact-statement wording + a product to build the ID from) looks right, and `TestAPIServerStore_updateVEX_adoptsPre595StatementsWithoutIDs` demonstrates the duplicate is gone for the straightforward case.

One correctness concern before merge, left inline: the new adoption loop runs before the existing ID/Name-normalization loop and reads `s.Vulnerability.Name` to build the stamped ID before that field is corrected, so a statement affected by *both* legacy bugs (no ID **and** the old swapped Vulnerability.ID/Name) gets a permanently wrong `id`. Given this PR is specifically about reaching very old pre-#595 data, that combination seems worth ruling out (or handling) rather than assuming away.

CodeRabbit already flagged the `Products[0].Subcomponents[0]` assumption — I think that one's fine as-is since every statement that can pass `isOwnImpactStatement` was built by kubevuln's own `createProductStructForImageAndPackage`, which always writes exactly one product with one subcomponent, so it's not an independent blocker for me, just flagging that I saw it and read it as already covered.

AnshulPatil2005:
good catch, that combination is real and the stamped id was wrong for it. a statement can carry both legacy shapes at once, no id and the swapped Vulnerability.ID/Name, and adoption reads Name to build the id before normalization has repaired it.

adoption has to stay ahead of the normalization loop, since that one only touches statements already recognised as ours and an ID-less statement is not, so instead it now reads the CVE from whichever field holds it. both share one predicate for recognising the old mapping rather than spelling it out twice.

TestAPIServerStore_updateVEX_adoptsStatementsWithBothLegacyShapes covers it. without the fix it stamps

    https://kubescape.io/vex/statement/https:%2F%2Fsecurity-tracker.debian.org%2Ftracker%2FCVE-2005-2541/...

instead of the CVE, and it also asserts normalization still repairs the mapping afterwards.

on the Products[0].Subcomponents[0] point, agreed, and same reading: anything that passes isOwnImpactStatement was built by createProductStructForImageAndPackage, which writes exactly one product with one subcomponent. left as is.

also rebased on main, #689 landed in the same region.

matthyx:
Checked the latest push (`fb65c66`). The double-legacy-shape fix looks right: `vulnerabilityCVEName`/`hasSwappedVulnerabilityFields` give adoption and the normalization loop one shared predicate instead of two that could drift apart, and `TestAPIServerStore_updateVEX_adoptsStatementsWithBothLegacyShapes` reproduces exactly the scenario I raised (asserts the stamped ID is built from the CVE via `localStatementID(cve, pkg)`, not from `Name` pre-normalization). Resolved my thread.

One thing still open that I think should be resolved before merge: [Codex's P1 comment](https://github.com/kubescape/kubevuln/pull/682#discussion_r3778289682) about affected-status legacy statements. `markRelevantVulnerabilitiesAsAffectedInVex` blanks `ImpactStatement` to `""` for any statement it marks affected, so a pre-#595 statement that was last stored while affected has no impact statement left, and `isOwnImpactStatement` — the only signal adoption has for "is this ours" — rejects it. That statement stays unmanaged and keeps getting duplicated/frozen on every rescan, same failure mode this PR fixes, just for the affected case instead of not-affected. It's still showing as unresolved on the latest commit and none of the new tests cover it (all three exercise `defaultLocalImpactStatement`/not-affected statements).

I get that this one's harder — an affected legacy statement with a blank impact statement doesn't obviously carry a marker that couldn't also belong to a genuine external "affected" assessment, so the same wording-match trick may not directly apply. But given the PR's goal is reaching *all* pre-#595 local data, not just the not-affected subset, I'd like to see this either handled here or explicitly scoped out (e.g. a follow-up issue) before merging, so it doesn't get lost. Happy to take another look once that's addressed.

AnshulPatil2005:
you're right that it's the same failure mode, and it is reachable: `markRelevantVulnerabilitiesAsAffectedInVex` blanks the impact statement, so an affected statement has nothing for `isOwnImpactStatement` to match on.

it does still carry one of our strings though, just in a different field. the same block that blanks the impact statement fills in the action statement:

```go
vexDoc.Statements[i].ImpactStatement = ""
vexDoc.Statements[i].ActionStatement = buildActionStatement(v)
```

and `buildActionStatement` never returns empty, it is either `defaultActionStatement` or `"Upgrade <purl> to version <versions>"`.

the dates work out too. action statements came in with #404 on 2026-08-01 and the ids with #595 on 2026-08-11, so anything written in that window, which is what a cluster that has scanned at all recently would have, has an action statement and no id. so adoption can key on it the same way: the constant, or the parameterised form checked against the statement's own subcomponent purl rather than as a loose prefix, which is not something a feed would land on by accident.

that leaves statements last written before #404, which have no id, no impact statement and no action statement either. there is nothing on those to tell them apart from a bare external "affected" statement, so I would not guess at them, and a cluster still holding one has not rescanned that workload in over two weeks. I will scope those out explicitly rather than let the gate get vague.

pushed, with tests for both the fallback wording and the named-versions form, plus one on the recogniser itself that an action statement naming a different package is not ours.

matthyx (approved):
Checked the latest push (`cbe69b2`). The fix for the affected-statement gap is well-targeted: `isOwnActionStatement` recognizes either the fallback wording or the parameterized "Upgrade `<purl>` to version …" form validated against the statement's *own* subcomponent PURL rather than as a loose prefix, so an external statement whose action text happens to open the same way but names a different package is correctly rejected (covered by `TestIsOwnActionStatement`'s negative case). `TestAPIServerStore_updateVEX_adoptsAffectedStatementsByActionStatement` exercises both wording forms against the exact shape `markRelevantVulnerabilitiesAsAffectedInVex` leaves behind (blank impact statement, populated action statement) and confirms adoption + reset-to-baseline.

The remaining gap — statements last written before #404 (no impact statement and no action statement either) — is explicitly and reasonably scoped out rather than silently dropped: there's genuinely nothing left on those to distinguish them from a bare external "affected" assessment, and the timeline argument (10-day window between #404 and #595, so only a cluster that hasn't rescanned in 2+ weeks would still hold one) makes that an acceptable, well-justified boundary rather than an oversight.

Both concerns I had are resolved and verified against the actual diff and tests, CI is green, and the scope is narrow and well-reasoned throughout. Approving.

PR #684, opened 2026-08-13
fix(services): count a singleflight waiter even when the creation failed

### LINKED ISSUE #683: singleflight hits are not counted when the shared SBOM creation fails
`kubevuln_singleflight_hits_total` counts nothing when the shared SBOM creation fails, because the error is handled before the hit is recorded:

```go
select {
case res := <-ch:
	if res.Err != nil {
		if !ran {
			s.reportWaiterSBOMFailure(workerCtx, res.Err)
		}
		return domain.SBOM{}, nil, res.Err
	}
	if !ran {
		metrics.RecordSingleflightHit(ctx, "sbom_generation")
	}
```

a caller that waited on that creation was spared it either way. it did not pull the image and it did not build an SBOM; it was handed someone else's result. that is what the counter is for, and #642/#649 treat those same callers as real enough to each get their own `ReportScanFailure` two lines above.

five concurrent scans of an image whose creation fails, with the metric scraped afterwards:

```
# HELP target_info Target metadata
# TYPE target_info gauge
target_info{service_name="kubevuln"} 1
```

no `kubevuln_singleflight_hits_total` at all, where four are expected.

the effect is not evenly spread. a burst of concurrent scans for one image is most likely when that image keeps being retried, which is what happens while it is failing: a registry outage, an auth problem, an image too large. so the deduplication is invisible in exactly the window where there is most of it, and the counter reads as though singleflight were doing nothing.

the early return came in with #610 and #631 left it alone: that change was about `res.Shared` counting the leader as deduplicated, and only swapped the condition on the line without moving it.

recording the hit before the error is handled is enough.

version: main (770090a)

AnshulPatil2005:
## Overview

`kubevuln_singleflight_hits_total` recorded nothing when the shared SBOM creation failed, because the error was handled and returned before the hit was recorded.

a caller that waited on that creation was spared it either way: it pulled no image and built no SBOM, it was handed someone else's result. #642 and #649 already treat those same callers as real enough to each get their own `ReportScanFailure`, two lines above where the counter was being skipped.

the gap is not evenly spread. a burst of concurrent scans for one image is most likely while that image keeps failing, a registry outage or an auth problem or an image over the size limit, so the deduplication went uncounted in the window that has most of it and the counter read as though singleflight were doing nothing.

the hit is now recorded before the error is handled.

## Additional Information

the early return came in with #610. #631 left it where it was: that change was about `res.Shared` counting the leader as one of the callers it spared, and only swapped the condition on the line without moving it. I wrote #631 and did not notice the ordering at the time.

no change to what counts as a hit, only to when it is recorded, so the leader is still excluded via `ran`.

## How to Test

`go test ./core/services/ -run Singleflight -count=1`

`TestScanService_SingleflightHitsCountedWhenCreationFails` runs five callers against one image whose `CreateSBOM` fails, on the blocking creator #610 already uses, and asserts the hits equal the callers that did not run the work. it asserts `callers - CreateSBOM calls` rather than a fixed four, so a caller arriving late enough to start its own round does not make it flap.

before this the counter is absent from the scrape entirely:

```
# HELP target_info Target metadata
# TYPE target_info gauge
target_info{service_name="kubevuln"} 1
does not contain kubevuln_singleflight_hits_total{target="sbom_generation"} 4
```

`TestScanService_SingleflightHitsCountOnlyDeduplicatedCallers` from #631 and `TestScanService_SingleflightSBOMDeduplication` from #610 are untouched and still pass.

## Related issues/PRs:

* Resolved #683

### DIFF
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -1382,6 +1382,13 @@ func (s *ScanService) getOrCreateSBOM(ctx context.Context, workload domain.ScanC
 
 	select {
 	case res := <-ch:
+		if !ran {
+			// Counted before the error is handled: a caller that waited on a shared creation
+			// was spared the work whether or not that creation succeeded, and a failing
+			// image is the one a burst of scans keeps arriving for, so leaving those out
+			// hides the deduplication where there is most of it.
+			metrics.RecordSingleflightHit(ctx, "sbom_generation")
+		}
 		if res.Err != nil {
 			// The leader already reported this failure with its own workload identity
 			// from inside the closure. Every other caller sharing the same result would
@@ -1392,9 +1399,6 @@ func (s *ScanService) getOrCreateSBOM(ctx context.Context, workload domain.ScanC
 				s.reportWaiterSBOMFailure(workerCtx, res.Err)
 			}
 			return domain.SBOM{}, nil, res.Err
-		}
-		if !ran {
-			metrics.RecordSingleflightHit(ctx, "sbom_generation")
 		}
 		created := res.Val.(sbomCreation)
 		return created.sbom, created.storeErr, nil
--- a/core/services/scan_test.go
+++ b/core/services/scan_test.go
@@ -2836,6 +2836,62 @@ func TestScanService_SingleflightHitsCountOnlyDeduplicatedCallers(t *testing.T) 
 		"the caller that did the work is not one of the callers it spared")
 }
 
+// A caller that waited on a shared SBOM creation was spared the work whether or not that
+// creation succeeded, so it is deduplicated either way. The error path returned before the
+// hit was recorded, so every waiter on a failing image went uncounted, and a failing image
+// is exactly the one a burst of scans keeps retrying.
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Verified the fix against the full `getOrCreateSBOM` closure: `ran` correctly distinguishes the leader (executes the closure) from waiters, and is only read after the closure has returned, so moving `RecordSingleflightHit` above the `res.Err != nil` check records every waiter — success or failure — without double-counting the leader. This matches the existing waiter-failure-reporting pattern (`reportWaiterSBOMFailure`, #642) that already treats waiters on a failed creation as real callers.

The new test derives its expected count as `callers - ranTheWork` rather than a fixed number, avoiding a flaky assertion if a caller starts its own round late. Reused test helpers (`blockingSBOMCreator`, `NewMockSBOMAdapter`) match existing usage elsewhere in the suite. All CI checks (build, lint, DCO, CodeQL, security) are green.

LGTM, approving.

PR #686, opened 2026-08-13
refactor(registryauth): share the registry auth fallback ladder

### LINKED ISSUE #685: the registry auth fallback ladder is written out twice, and the copies have started to drift
the registry auth fallback ladder, the one that retries MANIFEST_UNKNOWN against the tag and then walks 401 through the provider and on to anonymous, exists twice:

- `adapters/v1/syft.go`, in the in-process `SyftAdapter`
- `pkg/sbomscanner/v1/server.go`, in the sidecar scanner's `resolveSource`

about fifty lines each, and the same fifty. the fallback order, the metric calls, the "err/src retain the last attempt" reasoning, the wrapping of the anonymous failure back into the original 401, all of it is written out twice.

the two have already drifted in a small way. the sidecar walks the provider list by hand:

```go
for _, provider := range registryauth.Providers {
	if !provider.Matches(pullRef) {
		continue
	}
	...
	break
}
```

which is `registryauth.For` reimplemented:

```go
func For(imageID string) (Provider, bool) {
	for _, p := range Providers {
		if p.Matches(imageID) {
			return p, true
		}
	}
	return nil, false
}
```

the in-process copy calls `For`. the sidecar imports the same package and open-codes it.

this pair drifting apart is where #585, #587, #601 and #675 came from, so a fifty line copy of the credential path between them is worth removing rather than watching.

there is a second reason. `resolveSource` on the sidecar has five tests covering the ladder, the fallback metrics, a custom provider, nil credentials and the pullRef used for matching. the in-process copy of the same logic has none: nothing in `adapters/v1` exercises a 401 at all. one implementation is one set of tests, covering both.

`internal/` is importable from both, `adapters/v1` already imports `pkg/sbomscanner/v1` so the shared code cannot live in either of them, and both already depend on `internal/registryauth` and `internal/metrics`.

version: main (770090a)

AnshulPatil2005:
## Overview

the registry auth fallback ladder existed twice, about fifty lines each and the same fifty: retry MANIFEST_UNKNOWN against the tag, then walk a 401 through the matching provider and on to anonymous, wrapping the anonymous failure back into the original 401. one copy in the in-process `SyftAdapter`, one in the sidecar scanner's `resolveSource`.

this puts it in `internal/registryauth.ResolveSource` and has both call it. that removes 51 lines from `adapters/v1/syft.go` and 51 from `pkg/sbomscanner/v1/server.go`.

that pair drifting apart is where #585, #587, #601 and #675 came from, and this copy had already started: the sidecar walked `registryauth.Providers` by hand rather than calling `registryauth.For`, which is that loop exactly, and which the in-process copy does call.

## Additional Information

`internal/` is where it has to go. `adapters/v1` imports `pkg/sbomscanner/v1`, so the shared code cannot live in either, and both already depend on `internal/registryauth` and `internal/metrics`.

`ResolveSource` is generic over what a pull returns, so `internal/registryauth` does not have to import Syft. The ladder is about which reference and which credentials to try next, not about what comes back, and the sidecar's `sourceGetter` had already made the pull itself a parameter, so the shape was mostly there.

behaviour is unchanged on both sides. the pull keeps its own context on both, the in-process one carrying the size limit and the sidecar detaching for #421; the context handed to `ResolveSource` is the one used for the credential lookup and the metrics, which is `ctxWithTimeout` in-process and the request context on the sidecar, exactly as before. the metrics component label is a parameter.

## How to Test

`go test ./internal/registryauth/ ./adapters/... ./pkg/... -count=1`

the sidecar's five existing tests (`TestResolveSource`, `..._RecordsFallbackMetrics`, `..._CustomProvider`, `..._ProviderNilCredentialsFallsBackAnonymous`, `..._ProviderMatchesUsesPullRef`) are untouched and pass against the shared implementation, which is what says the extraction did not change it.

four new tests in `internal/registryauth` cover the ladder directly: the MANIFEST_UNKNOWN retry against the tag, the anonymous retry dropping refused credentials, an unrelated error going straight back after one attempt, and the caller's `RegistryOptions` surviving unmodified, since the ladder swaps credentials as it retries and a caller reusing its options for a later pull would otherwise inherit that.

the in-process ladder had no tests at all before this, nothing in `adapters/v1` exercises a 401, so it is covered now by the same tests as the sidecar rather than by nothing.

## Related issues/PRs:

* Resolved #685

### DIFF
--- a/adapters/v1/syft.go
+++ b/adapters/v1/syft.go
@@ -201,57 +201,16 @@ func (s *SyftAdapter) CreateSBOM(ctx context.Context, name, imageID, imageTag st
 
 	//nolint:staticcheck // stereoscope expects string key image.MaxImageSize
 	ctxWithSize := context.WithValue(ctxWithTimeout, image.MaxImageSize, s.maxImageSize)
-	pullRef := rewriteImageRef(imageID, s.proxyRegistryMap)
-	usedFallback := false
-	src, err := syft.GetSource(ctxWithSize, pullRef, syftGetSourceConfig(&registryOptions, imgPlatform))
-
-	if err != nil && strings.Contains(err.Error(), "MANIFEST_UNKNOWN") {
-		usedFallback = true
-		logger.L().Debug("got MANIFEST_UNKNOWN, retrying with imageTag",
-			helpers.String("imageTag", imageTag),
-			helpers.String("imageID", imageID))
-		pullRef = rewriteImageRef(imageTag, s.proxyRegistryMap)
-		src, err = syft.GetSource(ctxWithSize, pullRef, syftGetSourceConfig(&registryOptions, imgPlatform))
-	}
-
-	if err != nil && strings.Contains(err.Error(), "401 Unauthorized") {
-		usedFallback = true
-		unauthorizedErr := err
-		if provider, ok := registryauth.For(pullRef); ok {
-			if creds, credErr := provider.Credentials(ctxWithTimeout, pullRef); credErr != nil || creds == nil {
-				metrics.RecordScanFallback(ctx, metrics.ComponentInProcess, metrics.FallbackCategoryRegistryAuth, provider.Strategy(), metrics.FallbackOutcomeFailed)
-				logger.L().Debug("registry auth provider credentials unavailable, falling back to anonymous",
-					helpers.Error(credErr),
-					helpers.String("imageID", imageID))
-			} else {
-				registryOptions.Credentials = []image.RegistryCredentials{*creds}
-				src, err = syft.GetSource(ctxWithSize, pullRef, syftGetSourceConfig(&registryOptions, imgPlatform))
-				outcome := metrics.FallbackOutcomeFailed
-				if err == nil {
-					outcome = metrics.FallbackOutcomeSucceeded
-				}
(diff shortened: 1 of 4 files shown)


matthyx (approved):
Reviewed the diff line-by-line against both call sites' prior behavior.

- `internal/registryauth.ResolveSource` reproduces the MANIFEST_UNKNOWN → 401/provider → anonymous ladder from both `adapters/v1/syft.go` and `pkg/sbomscanner/v1/server.go` exactly, including the "don't mutate caller's opts" and provider-match ordering (`For` walks `Providers` in the same `[GCP, ECR]` order the sidecar's manual loop did).
- `rewriteImageRef` is applied before the imageID/imageTag are handed to `ResolveSource`, matching the old inline `pullRef` reassignment.
- Neither caller reads `registryOptions`/`opts` after the call, so passing it by value instead of mutating in place changes nothing observable.
- One very minor note, not a blocker: in the in-process path, metrics recording now goes through `ctxWithTimeout` instead of the plain `ctx` used before (both were already used for credential lookup). Since `context.WithTimeout` preserves parent values and the OTel counter's `Add` doesn't gate on `ctx.Err()`, this has no observable effect.
- New tests in `internal/registryauth/resolve_test.go` cover the tag retry, anonymous fallback, options-immutability, and non-matching-error passthrough. CI (build, lint, tests, CodeQL) is green.

Good to merge.

PR #689, opened 2026-08-13
refactor(repositories): build the local VEX statement in one place

### LINKED ISSUE #688: the local VEX statement is built in four places, and the four are the same
the statement kubevuln writes into a VEX document is built in four places, and the four are the same:

```
repositories/apiserver.go:1248  createVEX, over cve.Content.Matches
repositories/apiserver.go:1279  createVEX, over cve.Content.IgnoredMatches
repositories/apiserver.go:1414  updateVEX, over cve.Content.Matches
repositories/apiserver.go:1464  updateVEX, over cve.Content.IgnoredMatches
```

1248 and 1414 are byte for byte identical, and so are 1279 and 1464. each of the four also repeats the two steps in front of it, collecting the aliases out of `RelatedVulnerabilities` and calling `createProductStructForImageAndPackage`, so it is about 27 lines four times over.

the pairs differ only in what happens after the literal: the ignored ones hand the statement to `applyIgnoredMatchAssessment`, which overwrites all five of the fields it touches, so the baseline shape they start from is the same one the match loops keep.

this has cost something already. the ID was `fmt.Sprintf("https://kubescape.io/vex/statement/%s/%s", ...)` written out at all four, so #682, adopting statements stored before #595, had to stamp that same format a fifth time to produce an ID the dedup would recognise. a field added to the statement has four places to reach and three is enough to look right in a test.

one builder taking the match and the image reference covers all four. the `Match` embedded in an `IgnoredMatch` is the same type, so the ignored loops can pass `v.Match` and keep their assessment call afterwards.

version: main (770090a)

AnshulPatil2005:
## Overview

the statement kubevuln writes into a VEX document was built in four places: `createVEX` and `updateVEX` each build one over `Matches` and one over `IgnoredMatches`. the two match sites were byte for byte identical, as were the two ignored ones, and each of the four also repeated collecting the aliases and calling `createProductStructForImageAndPackage` in front of it.

`newLocalStatement` is now the one place that shape is defined. 90 lines out, 39 in.

the ignored loops pass `v.Match`, the `Match` embedded in an `IgnoredMatch` being the same type, and keep their `applyIgnoredMatchAssessment` call afterwards. that call sets all five of the fields it touches, so the baseline the builder returns is fully replaced for a suppression and nothing about those statements changes.

## Additional Information

this overlaps #682 by a few lines. that one adopts statements stored before #595 and had to stamp the same `https://kubescape.io/vex/statement/%s/%s` format a fifth time to produce an ID the dedup recognises, which is what having four copies costs. whichever of the two lands first, the other's rebase is a small manual merge in the same region, and the end state is the format written once either way.

`newLocalStatement` returns an error because `createProductStructForImageAndPackage` does. worth knowing that it never actually returns one, its only return is `nil`, so the `if err != nil` at each site was already dead. I left the plumbing as it was rather than fold a signature change into a de-duplication, but dropping the error from both would take the four call sites down to a line each.

## How to Test

the existing VEX tests are untouched and cover this end to end: `Test_StoreVEX`, the external-statement tests from #595 and #664, and #633's. they exercise all four sites through `StoreVEX`, so them passing is what says the four literals really were the same.

`TestNewLocalStatement` pins the shape directly: that the ID is one `isLocalStatement` accepts and carries both escaped components, that the CVE goes in `Name` and the data source in `ID` (the mapping `updateVEX`'s normalization exists to repair on older documents), the aliases, the single product and subcomponent, and the baseline status, justification and impact statement with no action statement or status notes.

## Related issues/PRs:

* Resolved #688

## Summary by CodeRabbit

* **Refactor**
  * Streamlined generation of local vulnerability assessment statements.
  * Preserved existing assessment handling and error behavior for regular and ignored matches.

* **Tests**
  * Added coverage verifying statement identifiers, vulnerability metadata, package associations, baseline status, and optional fields.

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -1149,6 +1149,39 @@ func ignoredMatchStatusNotes(rule v1beta1.IgnoreRule) string {
 	return strings.Join(parts, "; ")
 }
 
+// newLocalStatement builds the VEX statement kubevuln writes for a match, in the baseline
+// not_affected shape. createVEX and updateVEX each build one for a match and one for an
+// ignored match, and all four spelled it out; this is the one place that shape is defined,
+// so a field added to it cannot reach three sites and miss the fourth.
+//
+// A caller recording a suppression overwrites the baseline with applyIgnoredMatchAssessment,
+// which sets every one of the five fields it touches.
+func newLocalStatement(m v1beta1.Match, imagePullable string) (v1beta1.Statement, error) {
+	product, err := createProductStructForImageAndPackage(imagePullable, m.Artifact.PURL)
+	if err != nil {
+		return v1beta1.Statement{}, err
+	}
+
+	var aliases []string
+	for _, alias := range m.RelatedVulnerabilities {
+		aliases = append(aliases, alias.ID)
+	}
+
+	return v1beta1.Statement{
+		ID: fmt.Sprintf("https://kubescape.io/vex/statement/%s/%s", url.PathEscape(m.Vulnerability.ID), url.PathEscape(m.Artifact.PURL)),
+		Vulnerability: v1beta1.VexVulnerability{
+			ID:          m.Vulnerability.DataSource,
+			Name:        m.Vulnerability.ID,
+			Description: m.Vulnerability.Description,
+			Aliases:     aliases,
+		},
+		Products:        []v1beta1.Product{*product},
+		Status:          v1beta1.Status(vex.StatusNotAffected),
+		Justification:   v1beta1.Justification(vex.VulnerableCodeNotPresent),
+		ImpactStatement: defaultLocalImpactStatement,
+	}, nil
+}
+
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed the diff against the pre-refactor code: `newLocalStatement` reproduces the four inlined statement literals field-for-field (same ID format, same `Vulnerability`/`Products`/`Status`/`Justification`/`ImpactStatement`), the ignored-match call sites still pass `v.Match` and follow up with `applyIgnoredMatchAssessment` unchanged, and the dead `err != nil` check is left in place as the description notes. `TestNewLocalStatement` pins the shape directly and the existing VEX tests (`Test_StoreVEX`, `TestAPIServerStore_updateVEX_externalStatementDoesNotSuppressLocalOne`, etc.) exercise all four call sites through `StoreVEX`. CI is green (lint, tests, CodeQL, GitGuardian, DCO). CodeRabbit's only comment is a trivial nitpick about asserting the full statement ID in the new test rather than fragments — non-blocking.

No blockers. Approving.

PR #690, opened 2026-08-13
fix: release pullMutex only when abandoned Syft goroutine finishes

### LINKED ISSUE #687: pullMutex in SyftAdapter.CreateSBOM doesn't serialize actual pull/cataloging work; releases on timeout while the abandoned goroutine keeps running
## Problem

`SyftAdapter.CreateSBOM` (the in-process scan path used whenever the sbom-scanner sidecar is disabled or not yet ready — `cmd/http/main.go:91-100`) holds `s.pullMutex` for the express purpose of preventing concurrent image pulls/cataloging from exhausting local disk (`adapters/v1/syft.go:137-139`, `298-300`). On a scan timeout, that guarantee is violated: the mutex is released while the timed-out work is still running against disk, so a new scan can start pulling and cataloging fully concurrently with the "abandoned" one — exactly the condition the mutex exists to prevent.

## Reproduction / Evidence

`adapters/v1/syft.go:292-338`:

```go
generated := make(chan *sbom.SBOM, 1)
// ensure no parallel pulls
s.pullMutex.Lock()
defer s.pullMutex.Unlock()
dl := deadline.New(s.scanTimeout)
    defer func(src source.Source) {
        if err := src.Close(); err != nil { ... }
    }(src)
    ...
    created, createErr := createSBOMFn(ctxWithSize, src, cfg)
    if createErr != nil {
        return fmt.Errorf("failed to generate SBOM: %w", createErr)
    }
    generated <- created
    return nil
})
```

`dl.Run` is `github.com/eapache/go-resiliency/deadline` v1.7.0. Its own doc comment states the contract plainly (`deadline.go:24-29`):

```go
// Run runs the given function, passing it a stopper channel. If the deadline passes before
// the function finishes executing, Run returns ErrTimeOut to the caller and closes the stopper
// channel so that the work function can attempt to exit gracefully. It does not (and cannot)
// kill the running function's goroutine, so if the function doesn't respect the stopper channel,
// then it may keep running after the deadline passes.
```

and its implementation (`deadline.go:30-47`) launches the work closure in a goroutine and returns `ErrTimedOut` on timeout without waiting for that goroutine:

```go
func (d *Deadline) Run(work func(<-chan struct{}) error) error {
    result := make(chan error, 1)
    stopper := make(chan struct{})
    go func() {
        result <- work(stopper)
    }()
    timer := time.NewTimer(d.timeout)
    select {
    case ret := <-result:
        timer.Stop()
        return ret
    case <-timer.C:
        close(stopper)
        return ErrTimedOut
    }
}
```

The work closure in `syft.go` receives `stopper` but never reads from it anywhere in its body — there is no cancellation check. The code even documents this at `syft.go:327-331`:

```go
// This closure can outlive the call: deadline.Run cannot stop its work function and
// documents as much, and Syft's cataloguers do not observe cancellation either, so on
// timeout it keeps running after CreateSBOM has already returned Incomplete.
```

So on timeout:
1. `dl.Run` returns `ErrTimedOut` immediately; the goroutine keeps pulling layers / running Syft's cataloger against disk.
2. `CreateSBOM` takes the `deadline.ErrTimedOut` branch at `syft.go:340-345`, sets `Incomplete`, and returns.
3. The `defer s.pullMutex.Unlock()` registered at line 300 fires on that return — **while the abandoned goroutine is still doing disk I/O**.
4. A concurrent `CreateSBOM` call for a different image can now acquire `pullMutex` and start its own pull, running fully concurrently with the still-alive abandoned goroutine.

`ScanTimeout` defaults to 5 minutes (`config/config.go:100`) and is user-configurable; it is intentionally set aggressively in some deployments to bound worst-case scan latency. Under repeated timeouts (slow registries, large images, transient network issues), abandoned goroutines accumulate — each still pulling/cataloging — while `pullMutex` keeps admitting new scans as if nothing were in flight.

## Root Cause

`pullMutex` only serializes the *synchronous* call to `dl.Run`, not the actual pull/cataloging work, because that work escapes into an unmanaged goroutine that outlives the mutex's critical section. The mutex's `Unlock()` is tied to `CreateSBOM` returning, not to the work actually finishing — but `dl.Run`'s documented semantics guarantee those two events can diverge.

## Impact

- Defeats the disk-space-protection invariant the mutex is documented to provide (`syft.go:137-139`), under exactly the condition (timeouts) most likely to correlate with resource pressure in the first place.
- Unbounded accumulation of abandoned goroutines/`source.Source` instances under sustained timeout pressure — each holds pulled layers on disk until its own (unbounded-duration) `Close()` eventually runs.
- Affects the in-process `SyftAdapter` scan path, which is live production code (used as the primary path when the sidecar is disabled, and as a fallback when the sidecar isn't ready yet — `cmd/http/main.go:84-100`).

## Affected Components

- `adapters/v1/syft.go` (`SyftAdapter.CreateSBOM`)
- Transitively: any deployment relying on the in-process scanner (sidecar disabled) or hitting the sidecar-not-ready fallback.

## Acceptance Criteria

- `pullMutex` (or its replacement) is held for the actual duration of pull + cataloging work, not just for the synchronous portion of the call — i.e., on timeout, a subsequent `CreateSBOM` call must not start a new pull/cataloging pass while a prior one is still running.
- A regression test demonstrates that two `CreateSBOM` calls where the first times out cannot execute their disk-touching work (pull/cataloging) concurrently.
- No change to the documented behavior that a timed-out call itself returns `Incomplete` promptly to its caller — only the *serialization* guarantee needs fixing, not the caller-facing latency contract.

## Suggested Direction

Decouple "the caller stops waiting" from "the mutex is released": release `pullMutex` only when the background goroutine actually finishes (success, error, or eventual completion after timeout), not when `dl.Run` returns to the caller. Concretely, this likely means moving the unlock into the goroutine passed to `dl.Run` (or a wrapper around it) rather than relying on `defer` in `CreateSBOM`, while still letting `CreateSBOM` itself return `Incomplete` immediately on timeout as it does today. Care is needed to preserve the "return promptly on timeout" behavior for the calling goroutine while making the *next* scan attempt actually wait for the previous disk work to finish.

bhuvan-somisetty:
Fixes #687

## Problem

`SyftAdapter.CreateSBOM` holds `pullMutex` specifically to prevent concurrent image pulls/cataloging from exhausting local disk. On a scan timeout, that guarantee didn't hold: the mutex was released as soon as `CreateSBOM` returned, while the timed-out cataloging work kept running in the background — so a subsequent scan could start pulling/cataloging fully concurrently with the "abandoned" one.

## Root Cause

`dl.Run` (`github.com/eapache/go-resiliency/deadline`) spawns the work closure in its own goroutine and returns `ErrTimedOut` on timeout without waiting for that goroutine — its own doc comment says as much. The closure never reads from the `stopper` channel it's given, and Syft's cataloguers don't support cancellation, so on timeout the closure keeps running well past `CreateSBOM`'s return.

`pullMutex` was unlocked via `defer` at the top of `CreateSBOM`, tying its release to *the call returning* rather than to *the work finishing*. Those two events are exactly the ones `dl.Run`'s documented semantics guarantee can diverge — so the mutex only ever serialized the synchronous portion of the call, not the actual disk-touching work it exists to protect.

## Fix

Move `pullMutex.Unlock()` from `CreateSBOM`'s own `defer` into the closure passed to `dl.Run`, so it fires only once that closure actually finishes — promptly on normal success/failure, or late if it was abandoned after a timeout. `CreateSBOM`'s own caller-facing behavior is unchanged: it still returns `Incomplete` immediately on timeout. Only the mutex's release point moves, so a subsequent call now genuinely waits for the prior pull/cataloging work to finish before starting its own.

This is a minimal, targeted change (one file, ~10 lines) — no new abstractions, no behavior change outside the serialization guarantee itself.

## Testing

Added `Test_syftAdapter_CreateSBOM_SerializesWithAbandonedGoroutineAfterTimeout`, which:
1. Starts a `CreateSBOM` call whose cataloging step is held open past the adapter's scan timeout (via a stand-in `createSBOMFn` blocked on a channel), and confirms `CreateSBOM` returns `Incomplete` promptly.
2. Starts a second `CreateSBOM` call concurrently and asserts it does **not** reach its own cataloging step while the first call's abandoned goroutine is still running.
3. Releases the first call's abandoned goroutine and confirms the second call only then proceeds and completes.

This directly reproduces the bug on the pre-fix code (fails with the old release-on-return behavior) and passes on the fix.

Ran locally, all green:
- `go build ./...`
- `go vet ./adapters/v1/...`
- `go test ./adapters/v1/...` (full package, including the existing timeout/cancellation/multi-arch suite), run several times to confirm no new flakiness
- The new test specifically, run repeatedly (`-count=3`) to confirm it's deterministic

Note: I couldn't run `go test -race` in my local environment (no `gcc`/cgo available), so I'd appreciate CI confirming the race detector is clean on this change as well.

## Impact

- Restores the disk-space-protection invariant `pullMutex` is documented to provide, specifically on the timeout path most likely to correlate with disk/resource pressure in the first place.
- No change to scan latency or caller-facing timeout behavior — a timed-out call still returns `Incomplete` immediately.
- Scoped to the in-process `SyftAdapter` path (used when the sbom-scanner sidecar is disabled or not yet ready).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SBOM generation safety when an operation times out.
  * Prevented concurrent cataloging operations from starting while a previous timed-out operation is still finishing.
  * Ensured disk-affecting work remains properly serialized, reducing the risk of conflicts or inconsistent results.
  * Improved handling of interrupted operations and unusually long waits, helping subsequent scans proceed safely and predictably.
  * Reduced the risk of inconsistent results when multiple scans run close together.

### DIFF
--- a/adapters/v1/syft.go
+++ b/adapters/v1/syft.go
@@ -269,9 +269,17 @@ func (s *SyftAdapter) CreateSBOM(ctx context.Context, name, imageID, imageTag st
 	generated := make(chan *sbom.SBOM, 1)
 	// ensure no parallel pulls
 	s.pullMutex.Lock()
-	defer s.pullMutex.Unlock()
 	dl := deadline.New(s.scanTimeout)
 	err = dl.Run(func(stopper <-chan struct{}) error {
+		// Unlock here, not via a defer at the top of CreateSBOM: this closure can outlive
+		// CreateSBOM's own return (see the comment below), so unlocking when CreateSBOM
+		// returns would let a subsequent CreateSBOM call start pulling/cataloging while this
+		// one is still running against disk, defeating the "ensure no parallel pulls" purpose
+		// of pullMutex on exactly the timeout path most likely to correlate with disk
+		// pressure. Unlocking only once this closure actually finishes - promptly on success
+		// or failure, late if the deadline fired first - is what makes pullMutex serialize the
+		// disk-touching work itself, not just the synchronous portion of the call.
+		defer s.pullMutex.Unlock()
 		// make sure we clean the temp dir
 		defer func(src source.Source) {
 			if err := src.Close(); err != nil {
--- a/adapters/v1/syft.go
+++ b/adapters/v1/syft.go
@@ -201,6 +201,27 @@ func (s *SyftAdapter) CreateSBOM(ctx context.Context, name, imageID, imageTag st
 
 	//nolint:staticcheck // stereoscope expects string key image.MaxImageSize
 	ctxWithSize := context.WithValue(ctxWithTimeout, image.MaxImageSize, s.maxImageSize)
+	// ensure no parallel pulls: syft.GetSource (inside registryauth.ResolveSource) is what
+	// actually downloads image layers to local disk (via stereoscope), so pullMutex must
+	// guard it too, not just cataloging - otherwise a second CreateSBOM call can download a
+	// full image concurrently with a still-running "abandoned" download left behind by a first
+	// call that already timed out (#687).
+	// Ownership of the unlock transfers to the dl.Run closure below once source resolution
+	// succeeds; every early-return branch between here and there must unlock for itself.
+	pullMutexUnlockOnce := sync.Once{}
+	unlockPullMutex := func() { pullMutexUnlockOnce.Do(s.pullMutex.Unlock) }
+
+	lockWaitStart := time.Now()
+	// Arm a warning timer before acquiring Lock so the log line fires while still blocked if another call holds pullMutex longer than scanTimeout.
+	if s.scanTimeout > 0 {
+		timer := time.AfterFunc(s.scanTimeout, func() {
+			logger.L().Ctx(ctx).Warning("waited unusually long to acquire pullMutex; a previous scan may be stuck",
+				helpers.String("waited", time.Since(lockWaitStart).String()), helpers.String("imageID", imageID))
+		})
+		defer timer.Stop()
+	}
+	s.pullMutex.Lock()
+
 	// The MANIFEST_UNKNOWN and 401 fallback ladder lives in internal/registryauth, shared
 	// with the sidecar scanner, which ran a copy of it. The pull keeps its own context,
 	// carrying the size limit; ctxWithTimeout is what bounds the credential lookup.
@@ -222,6 +243,7 @@ func (s *SyftAdapter) CreateSBOM(ctx context.Context, name, imageID, imageTag st
 		logger.L().Ctx(ctx).Warning("Syft timed out during image resolution",
 			helpers.String("imageID", imageID))
 		domainSBOM.Status = helpersv1.Incomplete
+		unlockPullMutex()
 		return domainSBOM, nil
 	case err != nil && (errors.Is(err, image.ErrImageTooLarge) || strings.Contains(err.Error(), image.ErrImageTooLarge.Error())):
 		metrics.RecordScanFallback(ctx, metrics.ComponentInProcess, metrics.FallbackCategorySizeClassification, metrics.FallbackStrategyImageTooLarge, metrics.FallbackOutcomeClassified)
@@ -231,9 +253,11 @@ func (s *SyftAdapter) CreateSBOM(ctx context.Context, name, imageID, imageTag st
 		domainSBOM.Status = helpersv1.TooLarge
 		domainSBOM.Annotations[domain.StatusReasonAnnotationKey] = domain.ReasonImageTooLarge
 		domainSBOM.Annotations[domain.MaxImageSizeAnnotationKey] = fmt.Sprintf("%d", s.maxImageSize)
+		unlockPullMutex()
 		return domainSBOM, nil
 	case err != nil && strings.Contains(err.Error(), "401 Unauthorized"):
 		domainSBOM.Status = helpersv1.Unauthorize
+		unlockPullMutex()
 		return domainSBOM, err
 	case err != nil:
 		// Requested-but-unavailable platforms surface here as *image.ErrPlatformMismatch
@@ -250,6 +274,7 @@ func (s *SyftAdapter) CreateSBOM(ctx context.Context, name, imageID, imageTag st
 		// match options.Platform). Propagated as-is: classifySBOMError in core/services
 		// recognizes it via errors.As and reports a distinct "platform not found" reason
 		// instead of the generic SBOM-generation-failed fallback (see #512).
+		unlockPullMutex()
 		return domainSBOM, err
 	}
 
@@ -267,19 +292,18 @@ func (s *SyftAdapter) CreateSBOM(ctx context.Context, name, imageID, imageTag st
 	// Buffered so a goroutine abandoned by the deadline can always publish and exit, even
 	// when nobody is left to receive.
 	generated := make(chan *sbom.SBOM, 1)
-	// ensure no parallel pulls
-	s.pullMutex.Lock()
+	// pullMutex is already held (see the Lock() before syft.GetSource above), covering source
+	// resolution too. Unlock here, not via a defer at the top of CreateSBOM: this closure can
+	// outlive CreateSBOM's own return (see the comment below), so unlocking when CreateSBOM
+	// returns would let a subsequent CreateSBOM call start pulling/cataloging while this
+	// one is still running against disk, defeating the "ensure no parallel pulls" purpose
+	// of pullMutex on exactly the timeout path most likely to correlate with disk
+	// pressure. Unlocking only once this closure actually finishes - promptly on success
+	// or failure, late if the deadline fired first - is what makes pullMutex serialize the
+	// disk-touching work itself, not just the synchronous portion of the call.
 	dl := deadline.New(s.scanTimeout)
 	err = dl.Run(func(stopper <-chan struct{}) error {
-		// Unlock here, not via a defer at the top of CreateSBOM: this closure can outlive
-		// CreateSBOM's own return (see the comment below), so unlocking when CreateSBOM
-		// returns would let a subsequent CreateSBOM call start pulling/cataloging while this
-		// one is still running against disk, defeating the "ensure no parallel pulls" purpose
-		// of pullMutex on exactly the timeout path most likely to correlate with disk
-		// pressure. Unlocking only once this closure actually finishes - promptly on success
-		// or failure, late if the deadline fired first - is what makes pullMutex serialize the
-		// disk-touching work itself, not just the synchronous portion of the call.
-		defer s.pullMutex.Unlock()
+		defer unlockPullMutex()
 		// make sure we clean the temp dir
 		defer func(src source.Source) {
 			if err := src.Close(); err != nil {
--- a/adapters/v1/syft.go
+++ b/adapters/v1/syft.go
@@ -208,19 +208,29 @@ func (s *SyftAdapter) CreateSBOM(ctx context.Context, name, imageID, imageTag st
 	// call that already timed out (#687).
 	// Ownership of the unlock transfers to the dl.Run closure below once source resolution
 	// succeeds; every early-return branch between here and there must unlock for itself.
-	pullMutexUnlockOnce := sync.Once{}
-	unlockPullMutex := func() { pullMutexUnlockOnce.Do(s.pullMutex.Unlock) }
-
-	lockWaitStart := time.Now()
-	// Arm a warning timer before acquiring Lock so the log line fires while still blocked if another call holds pullMutex longer than scanTimeout.
+	// The warning below is armed *before* Lock() and fired by its own timer, not checked after
+	// Lock() returns: sync.Mutex.Lock never returns until the prior holder releases it, so a
+	// post-acquisition check can only ever report a wait that already ended - it would stay
+	// silent for the exact case it exists to catch, a previous cataloguing goroutine hung
+	// indefinitely (createSBOMFn/Syft's cataloguers do not observe cancellation - see the
+	// dl.Run comment below). That's an accepted tradeoff, not a bug: there is no watchdog that
+	// force-releases the mutex, so a sufficiently pathological image (corrupt archive,
+	// cataloguer bug, stalled local FS) can keep it held indefinitely. This log line is the
+	// operator-visible signal for that case, since neither the liveness nor readiness probe
+	// currently reflects it.
+	var longWaitWarning *time.Timer
 	if s.scanTimeout > 0 {
-		timer := time.AfterFunc(s.scanTimeout, func() {
-			logger.L().Ctx(ctx).Warning("waited unusually long to acquire pullMutex; a previous scan may be stuck",
-				helpers.String("waited", time.Since(lockWaitStart).String()), helpers.String("imageID", imageID))
+		longWaitWarning = time.AfterFunc(s.scanTimeout, func() {
+			logger.L().Ctx(ctx).Warning("waiting unusually long to acquire pullMutex; a previous scan may be stuck",
+				helpers.String("imageID", imageID))
 		})
-		defer timer.Stop()
 	}
 	s.pullMutex.Lock()
+	if longWaitWarning != nil {
+		longWaitWarning.Stop()
+	}
+	pullMutexUnlockOnce := sync.Once{}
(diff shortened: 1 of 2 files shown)


adapters/v1/syft.go
```diff
@@ -308,9 +308,17 @@ func (s *SyftAdapter) CreateSBOM(ctx context.Context, name, imageID, imageTag st
 	generated := make(chan *sbom.SBOM, 1)
 	// ensure no parallel pulls
 	s.pullMutex.Lock()
-	defer s.pullMutex.Unlock()
 	dl := deadline.New(s.scanTimeout)
```
matthyx, line 314: `pullMutex` is only acquired here, **after** `syft.GetSource` (line 206) has already run unlocked. `GetSource` is the actual "pull" — it downloads image layers into the stereoscope temp dir on local disk — so it stays fully unserialized both before and after this change: two concurrent `CreateSBOM` calls can still download full image layers to disk at the same time, including a still-running "abandoned" download racing a fresh one after a timeout.

This gap predates this PR, but it does mean the PR description ("Restores the disk-space-protection invariant `pullMutex` is documented to provide") overstates what actually changes here, and the fix doesn't fully satisfy #687's own acceptance criterion: *"`pullMutex` ... is held for the actual duration of pull + cataloging work, not just for the synchronous portion of the call."* Right now it's held for cataloging only — CodeRabbit's review flagged the same gap.

Given the mutex's stated purpose is disk-space protection and "pull" is literally in its name, I think this needs either: (a) `Lock()` moved before `GetSource`, with unlock handled on every early-return path there and handed off to the closure only once the pull succeeds, or (b) the PR description scoped down to acknowledge only the cataloging phase is protected, with a follow-up tracked for the pull phase.

adapters/v1/syft.go
```diff
@@ -308,9 +308,17 @@ func (s *SyftAdapter) CreateSBOM(ctx context.Context, name, imageID, imageTag st
...
 	err = dl.Run(func(stopper <-chan struct{}) error {
+		// Unlock here, not via a defer at the top of CreateSBOM: this closure can outlive
+		// CreateSBOM's own return (see the comment below), so unlocking when CreateSBOM
+		// returns would let a subsequent CreateSBOM call start pulling/cataloging while this
+		// one is still running against disk, defeating the "ensure no parallel pulls" purpose
+		// of pullMutex on exactly the timeout path most likely to correlate with disk
+		// pressure. Unlocking only once this closure actually finishes - promptly on success
+		// or failure, late if the deadline fired first - is what makes pullMutex serialize the
+		// disk-touching work itself, not just the synchronous portion of the call.
+		defer s.pullMutex.Unlock()
```
matthyx: Worth flagging as a new failure mode this change introduces (not present before): Syft's cataloguers don't observe `stopper`/ctx cancellation, so a pathological hang inside `createSBOMFn` — corrupt archive, cataloguer bug, stalled local FS — now holds `pullMutex` forever instead of just leaking one abandoned goroutine as before.

I checked how the pod behaves in that scenario: the Helm chart's liveness probe (`/v1/liveness`) returns an unconditional 200 regardless of scan-worker state, and the readiness probe only checks the Grype DB freshness — neither would notice `CreateSBOM` calls piling up. `CreateSBOM` runs off a bounded worker pool with no per-task timeout beyond this internal deadline, so one degenerate image could eventually exhaust the whole pool and silently stop all future SBOM generation, with no automated recovery.

Not a reason to block #687's fix on its own — it's a real improvement over the status quo (see general review comment, I verified `-race` is clean). But since there's no watchdog on the cataloging phase, I'd want this called out as an accepted tradeoff, or mitigated with some hard upper bound after which the pod considers itself unhealthy, before merging.

matthyx (changes requested):
Verified independently: cloned this branch and ran `go build ./...`, `go vet ./adapters/v1/...`, and `go test ./adapters/v1/... -race -count=5 -run 'Test_syftAdapter_CreateSBOM_SerializesWithAbandonedGoroutineAfterTimeout|Test_syftAdapter_CreateSBOM_TimeoutDoesNotRaceWithAbandonedSyft'` — all clean across 5 consecutive runs, no data races, confirming the fix's core claim (mutex release now waits for the abandoned goroutine) and answering the "please confirm race detector clean" note in the PR description. Full `./adapters/v1/...` suite with `-race` also passes.

The mechanics of the fix itself (moving `Unlock()` into the `dl.Run` closure via `defer`) are correct for what they cover: the closure always runs exactly once regardless of timeout, so the mutex is released exactly when cataloging actually finishes, not when `CreateSBOM` returns.

Two things worth resolving before merge — left as inline comments:
1. **Scope gap vs #687's acceptance criteria**: `pullMutex` is still acquired only after `syft.GetSource` (the actual image pull) already ran unlocked, so concurrent image downloads to disk remain unserialized. This predates the PR but means the "restores the disk-space-protection invariant" claim overstates the fix.
2. **New failure mode**: since Syft's cataloguers ignore cancellation, a truly-hung cataloging call now holds `pullMutex` forever (not just leaks one goroutine), and neither the liveness nor readiness probe would detect a worker pool wedged on it.

Requesting changes to get these addressed or explicitly scoped/acknowledged — happy to re-review once updated.

adapters/v1/syft_test.go
```diff
@@ -514,6 +515,91 @@ func Test_syftAdapter_CreateSBOM_TimeoutDoesNotRaceWithAbandonedSyft(t *testing.
...
+		t.Fatal("first CreateSBOM call never returned once its deadline elapsed")
+	}
+	// CreateSBOM has returned Incomplete, but its abandoned goroutine is still blocked in
+	// createSBOMFn, holding pullMutex.
+
+	secondReturned := make(chan struct{})
+	go func() {
+		defer close(secondReturned)
+		_, _ = adapter.CreateSBOM(context.Background(), "test", "", host+"/test-image:latest", domain.RegistryOptions{InsecureUseHTTP: true})
+	}()
```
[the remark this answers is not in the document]
bhuvan-somisetty: Good catch. Both calls now send their result through a channel and get asserted from the test goroutine, so nothing runs inside the worker closures anymore. Pushed in 4743795.

adapters/v1/syft.go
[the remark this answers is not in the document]
bhuvan-somisetty: You're right, GetSource was still unserialized. Moved pullMutex.Lock() to right before it, with every early-return branch in between unlocking for itself and ownership handed to the dl.Run closure only once source resolution succeeds. Pushed in 4743795.

bhuvan-somisetty: Fair, you're right that the PR description overstated it and #687's own criterion wasn't fully met. Went with (a): Lock() now happens before GetSource, with every early-return between there and the deadline closure unlocking for itself, and ownership only transferring to the closure once source resolution actually succeeds. So pullMutex now covers pull + cataloging end to end, not just cataloging.

adapters/v1/syft.go
bhuvan-somisetty: Agreed this is a real gap and I don't think it's worth blocking on a full watchdog for it right now - going with calling it out as an accepted tradeoff rather than building liveness/readiness plumbing for it, since that'd mean threading a health signal through the SBOMCreator interface across both the in-process and sidecar paths, which is a lot more surface than this fix should carry. Added a warning log when acquiring pullMutex takes longer than scanTimeout, so a stuck previous scan is at least visible to an operator grepping logs, even though nothing automatically recovers from it yet. Happy to open a follow-up issue for the watchdog/health-check piece if you'd rather track it separately.

bhuvan-somisetty:
Hey matthyx, thanks for digging into this so carefully - you were right on both counts. Pushed a fix: pullMutex.Lock() now happens before GetSource instead of after, so the actual pull is serialized too, not just cataloguing, with every early-return path unlocking for itself and ownership handing off to the deadline closure only once source resolution succeeds. For the stuck-mutex scenario, I didn't want to bolt on liveness/readiness plumbing for this fix specifically, so I went with logging a warning whenever acquiring pullMutex takes longer than scanTimeout - at least gives an operator something to grep for. Also fixed the test issue CodeRabbit flagged (both calls now assert from the test goroutine via channels instead of one asserting inside its own worker). Build/vet/tests all green, DCO signed off. Let me know if the watchdog piece is worth its own issue or if the log line is enough for now.

bhuvan-somisetty:
@coderabbitai review

adapters/v1/syft.go
```diff
@@ -201,6 +201,28 @@ func (s *SyftAdapter) CreateSBOM(ctx context.Context, name, imageID, imageTag st
...
+	if waited := time.Since(lockWaitStart); s.scanTimeout > 0 && waited > s.scanTimeout {
+		// A wait this long means a previous call's cataloguing goroutine is still running well
+		// past its own deadline (createSBOMFn/Syft's cataloguers do not observe cancellation -
+		// see the dl.Run comment below) and is holding pullMutex for the whole time. That's an
+		// accepted tradeoff, not a bug: there is no watchdog that force-releases the mutex, so a
+		// sufficiently pathological image (corrupt archive, cataloguer bug, stalled local FS)
+		// can keep it held indefinitely. This log line is the operator-visible signal for that
+		// case, since neither the liveness nor readiness probe currently reflects it.
+		logger.L().Ctx(ctx).Warning("waited unusually long to acquire pullMutex; a previous scan may be stuck",
+			helpers.String("waited", waited.String()), helpers.String("imageID", imageID))
```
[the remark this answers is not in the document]
bhuvan-somisetty: Good catch, that was actually broken - Lock() blocking means the post-acquisition check could never fire while a scan was genuinely stuck. Switched to arming a timer before Lock() and firing the warning from it while still blocked, stopping it once acquired. Pushed in 22c4532.

bhuvan-somisetty:
Fixed the pullMutex warning bug CodeRabbit caught - it was checking elapsed time after Lock() returned, which meant it could never actually fire while stuck (Lock is blocking, so by the time you're past it, the wait is already over). Now it's a timer armed before Lock() that fires the warning while still blocked and gets stopped once acquired, so it actually does what the comment above it claims. Build/vet/tests green, pushed in 22c4532.

matthyx (dismissed):
Both requested-changes items are resolved:

1. **Pull-phase serialization**: `pullMutex.Lock()` now happens before `syft.GetSource`, with every early-return branch between there and the deadline closure unlocking for itself via `unlockPullMutex` (backed by `sync.Once`). Confirmed by reading the current `CreateSBOM` — `pullMutex` now covers GetSource through cataloging end-to-end, matching #687's acceptance criterion.
2. **Hung-cataloging failure mode**: explicitly documented as an accepted tradeoff (not solved) rather than scope creep into liveness/readiness plumbing — reasonable given the surface a full watchdog would add. The operator-visible warning is now armed via a timer *before* `Lock()` and fires while still blocked, fixing the earlier version where the post-acquisition check could never fire during a genuine hang.

Re-verified independently rather than trusting the diff: cloned `pr690`, ran `go build ./...`, `go vet ./adapters/v1/...`, and `go test ./adapters/v1/... -race -count=5` targeting both the original and new regression tests — clean across 5 runs, no races — plus the full `./adapters/v1/...` suite with `-race`, also clean. All CI checks are green and my prior review threads are resolved.

Approving.

matthyx:
rebase pls

bhuvan-somisetty:
Rebased onto latest \main\. All tests passing.

matthyx (approved):
Re-approving — my previous approval was auto-dismissed when this branch got rebased onto latest `main` (new commit SHAs, old merge-conflict commit replaced by a clean rebase).

Checked that the rebase didn't change this PR's own content: diffed the current PR diff against the version I already reviewed and race-tested — the `pullMutex` logic is unchanged (only a comment reword). The `tools.RetryWithBackoff` wrapping now visible around `syft.GetSource`/`createSBOMFn` came in from `main` (already merged there), not from this PR — confirmed it's outside this PR's diff. Also confirmed `pullMutex.Lock()` still wraps the entire `registryauth.ResolveSource` call, so 429-retry attempts still happen under the mutex.

Re-ran `go build`, `go vet`, and `go test ./adapters/v1/... -race -count=5` on the rebased branch for both existing regression tests plus the new (main-derived) `Test_syftAdapter_CreateSBOM_Retry429RateLimit` — all clean, no races, 5/5. Full `./adapters/v1/...` suite with `-race` also clean. All CI checks green, all 5 review threads resolved.

PR #694, opened 2026-08-14
perf(adapters): index CVE exceptions once per scan instead of per match

### LINKED ISSUE #692: Exception lookup rescans the full exception list per CVE match, causing O(matches x exceptions) work every scan
# Description
`getCVEExceptionMatchCVENameFromList` in `adapters/v1/backend_utils.go` re-scans the entire exception list, and every `VulnerabilityPolicies` entry within it, for every single vulnerability match in a scan. It is called once per match from both `adapters/v1/domain_to_armo.go` (inside `DomainToArmo`'s match loop) and `adapters/v1/securityexception.go` (inside `ApplySecurityExceptions`'s match loop). For a scan with N matches and an exception list of M policies, this does O(N*M) work even though the exception list never changes across matches within the same scan.

For a large image scan (thousands of CVE matches is common for base-image-heavy workloads) combined with a moderately sized exception list (tens to hundreds of `SecurityException`/`ClusterSecurityException` policies, expected in a multi-tenant cluster), this adds real CPU cost and latency to every scan and rescan, which matters for a service whose core metric is scan duration.

# Environment
OS: Not applicable (server-side scanning service, not OS specific)
Version: main branch, as of commit 770090a

# Steps To Reproduce
1. Build an exception list with M `armotypes.VulnerabilityExceptionPolicy` entries, each with several `VulnerabilityPolicies`.
2. Build a Grype match document with N vulnerability matches.
3. Run the match loop in `ApplySecurityExceptions` (or `DomainToArmo`), which calls `getCVEExceptionMatchCVENameFromList(exceptions, m.Vulnerability.ID, isFixed)` once per match.
4. Observe that each call performs a full linear scan of the exception list and a nested scan of each exception's policies, so total work scales with N*M rather than N+M.

I wrote a benchmark (`BenchmarkGetCVEExceptionMatch` in `adapters/v1/backend_utils_test.go`) comparing the existing per-match linear scan against a per-scan index built once and reused for every match, using 200 exceptions (5 policies each) and 5000 matches:

```
BenchmarkGetCVEExceptionMatch/LinearScanPerMatch-10         3   55640361 ns/op
BenchmarkGetCVEExceptionMatch/IndexBuiltOncePerScan-10      3     965389 ns/op
```

# Expected behavior
Exception lookup by CVE name/alias should be close to O(1) amortized per match, for example by building a `map[string][]int` (or similar) index over the exception list once per scan, before the match loop starts, and reusing it for every match.

# Actual Behavior
`getCVEExceptionMatchCVENameFromList` walks the full exception list, and the full `VulnerabilityPolicies` slice of every exception, on every single call, so every scan repeats this O(M) work N times.

# Additional context
Proposed direction: build an index keyed by lower-cased CVE id/alias once per scan (for example a `cveExceptionIndex` wrapping a `map[string][]int` into the original exception slice), and have both `DomainToArmo` and `ApplySecurityExceptions` look up matches through that index instead of calling the linear-scan helper directly inside their per-match loops. The existing `getCVEExceptionMatchCVENameFromList` behavior (case-insensitive name match, `ExpiredOnFix` filtering, at-most-once-per-exception dedup) needs to be preserved exactly, since it is covered by existing tests in `backend_utils_test.go`.

rootp1:
## Overview
getCVEExceptionMatchCVENameFromList rescanned the full exception list, and every VulnerabilityPolicies entry within it, once per vulnerability match. Both DomainToArmo and ApplySecurityExceptions call it inside their match loops, so a scan with N matches against an exception list of M entries did O(N*M) work even though the exception list does not change between matches within a scan. This PR builds a name indexed lookup once per scan and reuses it for every match instead.

## Related issues/PRs:

* Resolved #692

## What changed
- Added `cveExceptionIndex`/`buildCVEExceptionIndex`/`lookup` in `adapters/v1/backend_utils.go`, which builds a lower cased CVE name to exception index map in one pass and returns exactly what the old linear scan returned for a given CVE name and `filterFixed` value.
- `DomainToArmo` (`adapters/v1/domain_to_armo.go`) and `ApplySecurityExceptions` (`adapters/v1/securityexception.go`) now build the index once before their match loops and call `lookup` per match instead of `getCVEExceptionMatchCVENameFromList`.
- Kept `getCVEExceptionMatchCVENameFromList` in place since it is still directly covered by existing tests, and added `TestCVEExceptionIndex_MatchesLinearScan` plus an assertion in the existing table test to confirm the index produces identical results to the linear scan for every case.
- Added `BenchmarkGetCVEExceptionMatch` in `adapters/v1/backend_utils_test.go` comparing the two approaches with 200 exceptions (5 policies each) against 5000 matches. On my machine this showed the indexed version at roughly 1/57th the time of the linear scan (55.6ms vs 0.97ms per iteration).

## Verification
- `go build ./...` passed
- `go test ./...` passed (all packages)
- `make lint` passed with 0 new issues
- `go vet ./...` shows 4 pre existing findings in controllers/http.go unrelated to this change, confirmed present on main before this branch

## Summary by CodeRabbit

* **Performance**
  * Improved vulnerability exception matching efficiency, especially for scans with many vulnerabilities or exception policies.
  * Reused indexed exception lookups while preserving existing CVE, alias, case-insensitive, expiration, and subcomponent matching behavior.

* **Bug Fixes**
  * Prevented duplicate exception policy results during matching.
  * Ensured exceptions expired by a vulnerability fix continue to be filtered correctly.

### DIFF
--- a/adapters/v1/backend_utils.go
+++ b/adapters/v1/backend_utils.go
@@ -363,3 +363,62 @@ func getCVEExceptionMatchCVENameFromList(srcCVEList []armotypes.VulnerabilityExc
 	}
 	return nil
 }
+
+// cveExceptionIndex maps a lower-cased CVE/alias name to the indices, into the exception
+// list it was built from, of every exception that declares a VulnerabilityPolicy with that
+// name. It lets a scan with many matches look up candidate exceptions in roughly constant
+// time per match instead of re-walking every exception (and every policy within it) for
+// each one, which is what getCVEExceptionMatchCVENameFromList does on its own.
+type cveExceptionIndex struct {
+	srcCVEList []armotypes.VulnerabilityExceptionPolicy
+	byName     map[string][]int
+}
+
+// buildCVEExceptionIndex builds a cveExceptionIndex over srcCVEList in a single pass. Build
+// it once per scan (the exception list does not change across matches within a scan) and
+// reuse it via lookup for every match.
+func buildCVEExceptionIndex(srcCVEList []armotypes.VulnerabilityExceptionPolicy) *cveExceptionIndex {
+	idx := &cveExceptionIndex{
+		srcCVEList: srcCVEList,
+		byName:     make(map[string][]int, len(srcCVEList)),
+	}
+	for i := range srcCVEList {
+		seen := make(map[string]struct{})
+		for j := range srcCVEList[i].VulnerabilityPolicies {
+			name := strings.ToLower(srcCVEList[i].VulnerabilityPolicies[j].Name)
+			if _, ok := seen[name]; ok {
+				continue
+			}
+			seen[name] = struct{}{}
+			idx.byName[name] = append(idx.byName[name], i)
+		}
+	}
+	return idx
+}
+
+// lookup returns the same result getCVEExceptionMatchCVENameFromList(idx.srcCVEList, CVEName,
+// filterFixed) would return, but only inspects the exceptions that actually declare a policy
+// named CVEName rather than the full exception list.
+func (idx *cveExceptionIndex) lookup(CVEName string, filterFixed bool) []armotypes.VulnerabilityExceptionPolicy {
+	if idx == nil {
+		return nil
+	}
+	indices := idx.byName[strings.ToLower(CVEName)]
+	if len(indices) == 0 {
+		return nil
+	}
+
+	var l []armotypes.VulnerabilityExceptionPolicy
+	for _, i := range indices {
+		exc := idx.srcCVEList[i]
+		if filterFixed && exc.ExpiredOnFix != nil && *exc.ExpiredOnFix {
+			continue
+		}
+		l = append(l, exc)
+	}
+
+	if len(l) > 0 {
+		return l
+	}
+	return nil
+}
--- a/adapters/v1/backend_utils_test.go
+++ b/adapters/v1/backend_utils_test.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	_ "embed"
 	"encoding/json"
+	"fmt"
 	"io"
 	"net/http"
 	"os"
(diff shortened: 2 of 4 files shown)


adapters/v1/backend_utils.go
```diff
@@ -363,3 +363,62 @@ func getCVEExceptionMatchCVENameFromList(srcCVEList []armotypes.VulnerabilityExc
...
+}
+
+// lookup returns the same result getCVEExceptionMatchCVENameFromList(idx.srcCVEList, CVEName,
+// filterFixed) would return, but only inspects the exceptions that actually declare a policy
+// named CVEName rather than the full exception list.
+func (idx *cveExceptionIndex) lookup(CVEName string, filterFixed bool) []armotypes.VulnerabilityExceptionPolicy {
+	if idx == nil {
+		return nil
+	}
+	indices := idx.byName[strings.ToLower(CVEName)]
```
matthyx, line 406: Nitpick, non-blocking: `strings.ToLower` + map equality isn't strictly identical to `strings.EqualFold` (used by `getCVEExceptionMatchCVENameFromList`) for a handful of Unicode edge cases — e.g. the Greek sigma forms `Σ`/`σ`/`ς` all fold together under `EqualFold` but `ToLower` only maps `Σ`→`σ`, leaving `ς` distinct. `TestCVEExceptionIndex_MatchesLinearScan` only exercises ASCII CVE-style names, so this divergence isn't covered.

In practice this is very unlikely to matter — CVE IDs and GHSA/ALAS/etc. aliases are always ASCII — so I don't consider this a blocker, just leaving it for the record (CodeRabbit's automated review flagged the same thing).

matthyx (approved):
Reviewed the diff and traced both call sites (`DomainToArmo`, `ApplySecurityExceptions`).

The new `cveExceptionIndex`/`buildCVEExceptionIndex`/`lookup` is a faithful indexed replacement for `getCVEExceptionMatchCVENameFromList`:
- Index is built once per scan, outside the match loop, in both call sites.
- Per-name dedup within an exception (`seen` map) matches the original's `break`-after-first-match dedup, including the case-insensitive alias-collision case.
- Indices for a given name are appended in ascending source-list order, so lookup results preserve the same ordering as the original linear scan.
- `filterFixed`/`ExpiredOnFix` filtering is applied identically at lookup time.
- Backed by `TestCVEExceptionIndex_MatchesLinearScan` plus an added assertion in the existing table test, so parity is actually verified, not just asserted in the PR description.

CI is green (build, lint, tests, CodeQL, GitGuardian, DCO). Left one non-blocking nitpick inline about a narrow Unicode case-folding edge case (`ToLower` vs `EqualFold`) that doesn't affect real-world CVE/GHSA-style identifiers.

Good to merge.

PR #696, opened 2026-08-14
fix: key ID-less matches in IgnoredMatchKeys so cache-diff republish detects them

### LINKED ISSUE #691: IgnoredMatchKeys silently drops matches with an empty Vulnerability.ID, breaking the cache-diff republish check
# Description
`IgnoredMatchKeys` (`adapters/v1/securityexception.go:542-554`) is documented as keying suppressed matches by vulnerability ID + artifact name + version, but any match with an empty `Vulnerability.ID` is silently skipped and contributes no key at all.

`reconcileCachedCVE` (`core/services/scan.go`, around lines 694-699) relies on `maps.Equal(prevIgnored, curIgnored)` between the previous and current key sets to decide whether the set of suppressed findings changed since the cached scan. That comparison is what decides whether to re-persist the manifest and republish VEX. The comment at `core/services/scan.go:676-679` documents that this exact class of bug has already been fixed three separate times (`#469`, `#501`, `#557`).

If a suppression state change affects only matches with no known CVE ID, the current and previous key sets are unaffected (both omit that entry), so `maps.Equal` reports no change and the republish is skipped. VEX output can silently go stale for exactly the matches this cache-diff mechanism is supposed to protect.

# Environment
OS: not applicable, this is a logic bug independent of OS
Version: current `main`

# Steps To Reproduce
1. Inspect `IgnoredMatchKeys` at `adapters/v1/securityexception.go:542-554`: the loop only inserts a key when `m.Vulnerability.ID != ""`; when it is empty, the match is skipped with no key emitted.
2. Inspect the existing tests `TestIgnoredMatchKeys` (`adapters/v1/securityexception_test.go:780`) and `TestIgnoredMatchKeys_DistinctKeysPerPackage` (`:797`) — both only construct matches with a populated `Vulnerability.ID`, so this path is untested.
3. Construct (or imagine) two `GrypeDocument`s representing a "before" and "after" cached scan, where a match with `Vulnerability.ID == ""` and populated `Artifact.Name`/`Version` changes suppression state between them (e.g. present in `IgnoredMatches` in one document but not the other).
4. Expected: `IgnoredMatchKeys` on the two documents produces different key sets, so `reconcileCachedCVE`'s `maps.Equal` check detects the change and triggers a republish.
5. Actual: because the ID-less match never contributes a key in either document, the two key sets are identical for that entry, so the change is invisible to the diff.

# Expected behavior
A suppression-state change for any match, including ones with no known CVE ID, should be detectable by the cache-diff and trigger a manifest/VEX republish.

# Actual Behavior
Matches with an empty `Vulnerability.ID` are excluded from the key set entirely, so their suppression state changes cannot affect the diff outcome, and VEX can go stale without any signal.

# Additional context
Likely root cause: `IgnoredMatchKeys` assumes `Vulnerability.ID` is always populated and uses it directly as part of the key, with no fallback for the empty case.

Suggested direction: when `Vulnerability.ID` is empty, fall back to a key derived from `Artifact.Name` + `Artifact.Version` (plus a marker distinguishing "no CVE ID" from a real ID, to avoid an empty ID colliding with itself across artifacts), so the entry still participates in the diff instead of being silently dropped. Add a regression test mirroring the existing `#469`/`#501`/`#557` pattern, extending `TestIgnoredMatchKeys*` to cover the empty-ID case, and confirm the fix does not change key output for matches with a populated ID.

rootp1:
## Overview
IgnoredMatchKeys keys suppressed matches by vulnerability ID plus artifact name and version, but it silently skipped any match whose Vulnerability.ID was empty. reconcileCachedCVE relies on comparing this key set between cache hits to decide whether to republish VEX and repersist the manifest, so a suppression change that only affects an ID less match was invisible to that diff. This PR gives ID less matches their own key, prefixed so they can never collide with the ID keyed namespace, so those suppression changes are detected again.

## Related issues/PRs:

Resolved #691

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tracking of ignored security findings that do not include a vulnerability ID.
  * Changes to the suppression status of these findings are now detected reliably.
  * Findings are distinguished correctly when they involve different package names or versions.
* **Tests**
  * Expanded coverage for ignored findings and suppression-state changes.

### DIFF
--- a/adapters/v1/securityexception.go
+++ b/adapters/v1/securityexception.go
@@ -539,6 +539,10 @@ func isExceptionSourcedIgnore(im v1beta1.IgnoredMatch) bool {
 // fix (hasKnownFix, via getCVEExceptionMatchCVENameFromList), so two matches of the same CVE can
 // have different suppression states. The manifest content is fixed across a cache hit, so these
 // keys are stable and detect both ID- and fix-state-driven changes to the ignored set.
+//
+// Matches with no known CVE ID still get a key, prefixed with \x01 to keep them out of the
+// ID-keyed namespace, so that a suppression-state change limited to ID-less matches is still
+// visible to callers (like reconcileCachedCVE) diffing this set across cache hits.
 func IgnoredMatchKeys(doc *v1beta1.GrypeDocument) map[string]struct{} {
 	keys := map[string]struct{}{}
 	if doc == nil {
@@ -548,6 +552,8 @@ func IgnoredMatchKeys(doc *v1beta1.GrypeDocument) map[string]struct{} {
 		m := im.Match
 		if m.Vulnerability.ID != "" {
 			keys[m.Vulnerability.ID+"\x00"+m.Artifact.Name+"\x00"+m.Artifact.Version] = struct{}{}
+		} else {
+			keys["\x01"+m.Artifact.Name+"\x00"+m.Artifact.Version] = struct{}{}
 		}
 	}
 	return keys
--- a/adapters/v1/securityexception_test.go
+++ b/adapters/v1/securityexception_test.go
@@ -786,11 +786,52 @@ func TestIgnoredMatchKeys(t *testing.T) {
 			},
 		}
 
-		assert.Equal(t, map[string]struct{}{"CVE-A\x00\x00": {}}, IgnoredMatchKeys(doc))
+		assert.Equal(t, map[string]struct{}{
+			"CVE-A\x00\x00": {},
+			"\x01\x00":      {},
+		}, IgnoredMatchKeys(doc))
 	})
 
 	t.Run("nil document returns an empty set", func(t *testing.T) {
 		assert.Empty(t, IgnoredMatchKeys(nil))
+	})
+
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Verified the fix against the actual caller: `reconcileCachedCVE` (core/services/scan.go) diffs `IgnoredMatchKeys` sets via `maps.Equal` across cache hits to decide whether to republish VEX/repersist the manifest. Before this change, ID-less matches were silently dropped from the key set, so a suppression-state change limited to those matches was invisible to that diff — this PR fixes exactly that gap.

The `\x01` prefix can't collide with the existing `ID\x00Name\x00Version` namespace since no real vulnerability ID starts with `\x01`, and the fix is covered by a direct regression test (`before`/`after` docs with an ID-less match producing different key sets) plus a distinct-packages test.

One non-blocking nit already flagged by CodeRabbit: the "distinct packages" test could also cover same-name-different-version to explicitly assert the version component discriminates the key too — worth adding but not a blocker.

CI is green (lint, CodeQL, security/license checks, GitGuardian, DCO); the Go test job was still finishing as of this review, worth a final glance once it completes.

Approving.

PR #701, opened 2026-08-14
fix: prevent RestoreSuppressedMatches from undoing VEX suppressions

### LINKED ISSUE #699: RestoreSuppressedMatches misidentifies external-VEX suppressions as SecurityExceptions, undoing them on cache hits
# Description
While looking into the codebase for the LFX Mentorship project (issue #387), I found a bug that will break the new external VEX feature. External VEX suppressions get accidentally undone whenever there is a cache hit. 

The problem happens in two steps:
1. When Grype uses a VEX file to ignore a vulnerability, it adds helpful tags to the rule (like `Namespace: "vex"`). But when Kubevuln converts this data in `adapters/v1/grype_to_domain.go`, it accidentally drops these tags. It only saves the vulnerability ID.
2. Later, in `adapters/v1/securityexception.go`, a function called `isExceptionSourcedIgnore` tries to check if a CVE was ignored by a `SecurityException` CRD. It uses a very loose check: `return r.FixState == ""`. Because the VEX rule lost its tags in step 1, it matches this loose check. The system mistakenly thinks the VEX rule is a `SecurityException`.

Because of this mix-up, the `RestoreSuppressedMatches` function takes the VEX-ignored CVE and puts it back into the active `Matches` list.

# Environment
OS: `N/A` (This is a logic bug in the code)
Version: `main` branch (Checked against `grype v0.99.1` and `kubescape/storage v0.0.302`)
 
# Steps To Reproduce
Since the VEX feature is still being built, the easiest way to see this bug in action is by running a quick unit test.

1. Add this test to `adapters/v1/securityexception_test.go`:

```go
func TestRestoreSuppressedMatches_ExternalVEXFalsePositive(t *testing.T) {
	// This simulates what happens when Grype processes a VEX file and 
	// Kubevuln drops the tags during conversion. We are left with a single 
	// rule that only has the Vulnerability ID.
	doc := &v1beta1.GrypeDocument{
		Matches: []v1beta1.Match{
			{Vulnerability: v1beta1.Vulnerability{VulnerabilityMetadata: v1beta1.VulnerabilityMetadata{ID: "CVE-KEEP"}}},
		},
			{
				Match:              v1beta1.Match{Vulnerability: v1beta1.Vulnerability{VulnerabilityMetadata: v1beta1.VulnerabilityMetadata{ID: "CVE-VEX-SUPPRESSED"}}},
				AppliedIgnoreRules: []v1beta1.IgnoreRule{{Vulnerability: "CVE-VEX-SUPPRESSED"}},
			},
		},
	}

	restored := RestoreSuppressedMatches(doc)

	require.NotNil(t, restored)
	assert.Len(t, restored.Matches, 1, "VEX-suppressed match should NOT be restored")
	assert.Len(t, restored.IgnoredMatches, 1, "VEX-suppressed match should stay ignored")
}
```
2. Run the test using your terminal: `go test -run TestRestoreSuppressedMatches_ExternalVEXFalsePositive ./adapters/v1/`
3. See error: The test fails. It shows that `restored.Matches` has 2 items instead of 1. The VEX-ignored CVE was incorrectly put back into the active list.

# Expected behavior
When an external VEX file says a CVE should be ignored, it should stay ignored even during a cache hit. The `RestoreSuppressedMatches` function should recognize that it's not a `SecurityException` CRD and leave it alone.

# Actual Behavior
Because `isExceptionSourcedIgnore` gets confused by the dropped tags, it thinks the VEX rule is a CRD exception. It moves the VEX-ignored CVE back to the active list. When the system tries to re-apply CRD exceptions later, it doesn't know how to read VEX files, so the CVE stays active. Basically, the VEX suppression gets completely undone on the very first cache hit.

# Additional context
* **Why this matters:** This directly breaks the main goal of the LFX Mentorship project (issue #387). If we don't fix this, external VEX files won't work properly across multiple scans.
* **Code Comment:** There is a comment in `securityexception.go:496` that says the code is safe from Grype's `VexProcessor`, but it actually isn't safe right now because of the dropped tags.
* **How we could fix it:** 
  1. Fix the conversion: Update the code in `grype_to_domain.go` (and maybe the storage package) so it doesn't drop the VEX tags.
  2. Fix the check: In `isExceptionSourcedIgnore`, remove the `r.FixState == ""` fallback. Make it require a real tag (like `r.SourceKind != ""`) to be considered a CRD exception.

h30s:
## Overview

Right now, if an external VEX document ignores a vulnerability, it accidentally gets un-ignored the next time you scan the same image (when the cache is used). This happens because the code gets confused and thinks the blank VEX rule is a SecurityException, so it puts the vulnerability back into the active list.

This PR fixes that. It makes the check much stricter. Now, the code only thinks something is a SecurityException if it actually has the right tags (like `SourceKind`). This stops the VEX ignores from being undone.

## Additional Information

> Any additional information that may be useful for reviewers to know 

The real root cause is that VEX tags get dropped when converting data in `grype_to_domain.go`. We can't fully fix that here because we need to add fields to `v1beta1.IgnoreRule` in the `kubescape/storage` repo first. This PR is a necessary, quick fix to stop the bug for now. The other fix will come in a future PR.

I also fixed a wrong comment above the `RestoreSuppressedMatches` function.

## How to Test

> Please provide instructions on how to test the changes made in this pull request

I added two new unit tests:
- `TestRestoreSuppressedMatches_ExternalVEXFalsePositive`: Proves that a VEX rule with lost tags stays ignored.
- `TestRestoreSuppressedMatches_GenuineCRDException`: Proves that real CRD exceptions (which always have a `SourceKind` tag) still get restored properly.

**To test manually:**
1. Scan an image that has a VEX-ignored vulnerability.
2. Scan the exact same image again (this triggers the cache).
3. Check that the vulnerability stays ignored instead of coming back.

## Examples/Screenshots

> Here you add related screenshots 

*(No screenshots needed for this logic fix)*

## Related issues/PRs:

* Related to #387
* Resolved #699

**Please open the PR against the `dev` branch (Unless the PR contains only documentation changes)**

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved restoration of suppressed vulnerability matches by recognizing their CRD provenance.
  * Prevented externally sourced VEX suppressions from being incorrectly restored as security exceptions.
  * Correctly identifies genuine security exception suppressions even when fix-state information is absent.
  * Consolidated repeated conversion warnings into clearer messages with occurrence counts.

### DIFF
--- a/adapters/v1/domain_to_syft.go
+++ b/adapters/v1/domain_to_syft.go
@@ -77,6 +77,18 @@ func warnConversionErrors[T any](converted []T, errors []error) []T {
 	return converted
 }
 
+func deduplicateErrors(errors []error) []string {
+	errorCounts := make(map[string]int)
+	var errorMessages []string
+	for _, e := range errors {
+		errorCounts[e.Error()] = errorCounts[e.Error()] + 1
+	}
+	for msg, count := range errorCounts {
+		errorMessages = append(errorMessages, fmt.Sprintf("%q occurred %d time(s)", msg, count))
+	}
+	return errorMessages
+}
+
 func toSyftFiles(files []model.File) sbom.Artifacts {
 	ret := sbom.Artifacts{
 		FileMetadata: make(map[file.Coordinates]file.Metadata),
--- a/adapters/v1/securityexception.go
+++ b/adapters/v1/securityexception.go
@@ -500,10 +500,10 @@ func emitSuppressionEvent(recorder record.EventRecorder, p armotypes.Vulnerabili
 // so the shared stored object is never modified.
 //
 // Only ignored matches carrying the signature ApplySecurityExceptions writes (a single ignore
-// rule with only the vulnerability ID set) are restored. Any other entry (e.g. one that would be
-// produced if GrypeAdapter ever wired up IgnoreRules/VexProcessor) is preserved, so the helper is
-// self-guarding instead of relying on an assumption. Reconstruction is lossless because
-// IgnoredMatch embeds the full Match.
+// rule with CRD provenance tags like SourceKind or Justification set) are restored. Any other
+// entry (e.g. one produced by GrypeAdapter wiring up IgnoreRules/VexProcessor, which drops CRD
+// tags) is preserved, so the helper is self-guarding instead of relying on an assumption.
+// Reconstruction is lossless because IgnoredMatch embeds the full Match.
 func RestoreSuppressedMatches(doc *v1beta1.GrypeDocument) *v1beta1.GrypeDocument {
 	if doc == nil {
 		return nil
@@ -526,8 +526,7 @@ func RestoreSuppressedMatches(doc *v1beta1.GrypeDocument) *v1beta1.GrypeDocument
 
 // isExceptionSourcedIgnore reports whether an ignored match carries the AppliedIgnoreRules
 // signature ApplySecurityExceptions writes: non-empty AppliedIgnoreRules where every rule has no
-// package set and its SourceKind is a known exception source ("SecurityException" or
-// "ClusterSecurityException"), or matches the empty legacy-rule fallback.
+// package set, and exception provenance (source/justification/impact).
 func isExceptionSourcedIgnore(im v1beta1.IgnoredMatch) bool {
 	if len(im.AppliedIgnoreRules) == 0 {
 		return false
@@ -537,9 +536,6 @@ func isExceptionSourcedIgnore(im v1beta1.IgnoredMatch) bool {
 			return false
 		}
 		if r.SourceKind == "SecurityException" || r.SourceKind == "ClusterSecurityException" {
-			continue
-		}
-		if r.SourceKind == "" && r.SourceName == "" && r.SourceNamespace == "" && r.Justification == "" && r.ImpactStatement == "" && r.FixState == "" {
 			continue
 		}
 		return false
--- a/adapters/v1/securityexception_test.go
+++ b/adapters/v1/securityexception_test.go
@@ -790,11 +790,11 @@ func TestRestoreSuppressedMatches(t *testing.T) {
 			IgnoredMatches: []v1beta1.IgnoredMatch{
 				{
 					Match:              v1beta1.Match{Vulnerability: v1beta1.Vulnerability{VulnerabilityMetadata: v1beta1.VulnerabilityMetadata{ID: "CVE-A"}}},
-					AppliedIgnoreRules: []v1beta1.IgnoreRule{{Vulnerability: "CVE-A"}},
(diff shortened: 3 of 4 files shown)


matthyx (dismissed):
Verified the core fix: `buildIgnoreRule` (unchanged by this PR) always sets `SourceKind` on ignore rules produced by `ApplySecurityExceptions`, since `buildSuppressionAttributes` unconditionally populates `attrs["sourceKind"]`. So tightening `isExceptionSourcedIgnore` to require actual provenance tags (and dropping the `FixState == ""` fallback) is safe for genuine CRD-sourced exceptions and correctly stops externally-tagged-VEX ignores (which have none of those fields) from being misclassified and un-suppressed. New tests cover both directions and CI is green.

One non-blocking nit: the doc comment directly above `isExceptionSourcedIgnore` (adapters/v1/securityexception.go) still says the signature is "either exception provenance ... or an empty FixState" — that's now stale since the `FixState == ""` fallback was removed. Worth a follow-up tweak so the comment matches the tightened logic, but it doesn't affect correctness. Approving.

matthyx:
rebase pls

h30s:
@matthyx PTAL

matthyx (approved):
Re-reviewed after the new push (my prior approval was auto-dismissed by the commit).

`main` moved since my last look — `isExceptionSourcedIgnore` is now multi-rule aware (iterates `AppliedIgnoreRules` and requires each rule's `SourceKind` to equal `"SecurityException"`/`"ClusterSecurityException"`), which also correctly handles the case where two policies suppress the same finding and `buildIgnoreRules` emits more than one rule. This PR's actual change on top of that is removing the "all fields blank" legacy fallback branch — exactly the bug: an external VEX-derived `IgnoreRule` with only `Vulnerability` set was previously falling into that fallback and getting incorrectly restored/un-suppressed. `buildIgnoreRules` still always sets `SourceKind` from `buildSuppressionAttributes`'s unconditional `attrs["sourceKind"]`, so genuine CRD exceptions are unaffected. Tests cover both cases and CI is green.

Two non-blocking notes:
- The updated doc comment on `isExceptionSourcedIgnore` now says "exception provenance (source/justification/impact)", but the code only checks `SourceKind` — `SourceName`/`Justification`/`ImpactStatement` aren't inspected. Comment doesn't quite match the implementation; worth a follow-up wording fix.
- `deduplicateErrors` (adapters/v1/domain_to_syft.go) is unrelated to the VEX-suppression fix this PR is about — a nice cleanup (consolidating repeated SBOM conversion warnings), but scope creep for a "quick fix" PR, and it has no unit test (CodeRabbit flagged the same). Not a blocker, just flagging.

No blockers found. Approving.

PR #704, opened 2026-08-14
security: remediate gosec SAST findings in kubevuln

sagarkhandagre998:
Resolve 16 gosec findings across the kubevuln codebase:
- G101: suppress false-positive metric label (not a credential)
- G104: handle previously ignored errors (io.Copy drain, file close, gRPC/HTTP connection close, socket file removal)
- G112: set ReadHeaderTimeout on the HTTP server to mitigate Slowloris
- G115: bounds-check int->uint32 / int->int32 conversions before casting
- G304: filepath.Clean file paths before os.ReadFile
- G404: suppress non-security RNG use (retry backoff jitter)
- G703: clean and validate operator-controlled SOCKET_PATH used in os.Remove (path traversal taint finding)

Re-scan with gosec reports 0 findings (excluding cmd/http taint rules, which deadlock this gosec build's analyzer on cmd/http's large dependency graph).

### DIFF
--- a/adapters/v1/grype.go
+++ b/adapters/v1/grype.go
@@ -591,8 +591,12 @@ func checkDBDirWritable(dir string) error {
 	if err != nil {
 		return err
 	}
-	f.Close()
-	os.Remove(f.Name())
+	if err := f.Close(); err != nil {
+		return err
+	}
+	if err := os.Remove(f.Name()); err != nil {
+		return err
+	}
 	return nil
 }
 
--- a/adapters/v1/backend.go
+++ b/adapters/v1/backend.go
@@ -411,7 +411,9 @@ func (a *BackendAdapter) ReportScanFailure(ctx context.Context, failureCase scan
 		return err
 	}
 	defer resp.Body.Close()
-	io.Copy(io.Discard, resp.Body)
+	if _, err := io.Copy(io.Discard, resp.Body); err != nil {
+		logger.L().Ctx(ctx).Debug("failed to drain response body", helpers.Error(err))
+	}
 	if resp.StatusCode >= http.StatusBadRequest {
 		return fmt.Errorf("scan failure report returned HTTP %d", resp.StatusCode)
 	}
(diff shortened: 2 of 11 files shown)


sagarkhandagre998:
@matthyx Please take look.

adapters/v1/grype.go
```diff
@@ -591,8 +591,12 @@ func checkDBDirWritable(dir string) error {
 	if err != nil {
 		return err
 	}
-	f.Close()
-	os.Remove(f.Name())
+	if err := f.Close(); err != nil {
+		return err
+	}
+	if err := os.Remove(f.Name()); err != nil {
```
matthyx, line 597: Non-blocking nit: previously `f.Close()` and `os.Remove(f.Name())` both ran unconditionally, so the probe temp file was always cleaned up. Now, if `f.Close()` errors, the function returns early and `os.Remove` is skipped, leaving the temp file behind in `DBRootDir`. Since this only feeds a warning log in `updateDBBackground`, worth considering `defer os.Remove(f.Name())` (or removing regardless of the close error) so cleanup isn't skipped on that path.

matthyx (approved):
Reviewed the gosec remediation changes. Checked out the branch locally and verified:
- `go build ./...` succeeds
- `go vet ./...` findings are pre-existing on `main` (unrelated to this diff)
- Full test suite for touched packages (`adapters/...`, `internal/...`, `config/...`, `pkg/sbomscanner/...`) passes
- Re-ran `gosec ./...`: the 16 findings described in the PR are resolved; the only remaining findings (G103, unsafe pointer use) are in generated protobuf code, pre-existing and out of scope
- Each `#nosec` suppression is preceded by an actual guard (bounds check, path cleaning, or a genuine non-security use), not a blanket silence

Left one non-blocking nit on `checkDBDirWritable`'s cleanup ordering. No blockers — approving.

Note: the `PR Lint` GitHub Actions workflow hasn't run yet (only DCO/GitGuardian/CodeRabbit/FOSSA checks are present) — likely pending workflow approval since this is a first-time external contributor. Worth a maintainer approving the workflow run before merge.

PR #706, opened 2026-08-14
refactor: remove dead isGlobal parameter from incrementCounter

### LINKED ISSUE #705: bug: incrementCounter's isGlobal parameter is dead flexibility
# Description
The `incrementCounter` function in `adapters/v1/backend_utils.go` currently accepts an `isGlobal` boolean parameter, which it uses to evaluate `if isGlobal && isIgnored { return }`. However, across all of its ~5 call sites within the `Summarize()` function, `isGlobal` is unconditionally passed as `true`. This results in dead flexibility and unnecessary parameter passing. We should remove the `isGlobal` parameter and simplify the condition to `if isIgnored { return }`.

# Steps To Reproduce
1. Open `adapters/v1/backend_utils.go`
2. Look at the usages of `incrementCounter` inside `Summarize()`.
3. Notice that all of them pass `true` as the second argument.

# Expected behavior
The `isGlobal` parameter should be removed and the `if` condition simplified to improve readability.

# Actual Behavior
The code retains a dead branch of flexibility.

# Additional context
This is a minor refactoring/simplification opportunity, not a functional bug.

doraem-on:
## Overview
This PR removes the dead `isGlobal` parameter from `incrementCounter` in `adapters/v1/backend_utils.go`. Across all call sites in `Summarize()`, this parameter was unconditionally passed as `true`, meaning the branching logic `if isGlobal && isIgnored` was effectively just `if isIgnored`. This change simplifies the function signature without altering any functional behavior.

## How to Test
 Run unit tests in `adapters/v1` to verify logic remains intact.
```bash
go test ./adapters/v1
```

## Related issues/PRs:
Closes #705

- [ ] I have commented on my code, particularly in hard-to-understand areas

### DIFF
--- a/adapters/v1/backend_utils.go
+++ b/adapters/v1/backend_utils.go
@@ -186,8 +186,8 @@ func (a *BackendAdapter) sendVulnerabilities(ctx context.Context, chunksChan <-c
 	}
 }
 
-func incrementCounter(counter *int64, isGlobal, isIgnored bool) {
-	if isGlobal && isIgnored {
+func incrementCounter(counter *int64, isIgnored bool) {
+	if isIgnored {
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


matthyx (approved):
Verified: `isGlobal` was passed as `true` at all 5 call sites in `Summarize()` prior to this change, so `if isGlobal && isIgnored` was equivalent to `if isIgnored` — dropping the parameter is a no-op behaviorally. Confirmed `incrementCounter` is only defined/used in this one file (no other call sites left on the old signature), and all CI checks (build, lint, CodeQL, DCO, GitGuardian) are green. No blockers — good to merge.

PR #710, opened 2026-08-14
fix(adapters): document and test isExpired's now-equality boundary

### LINKED ISSUE #709: isExpired boundary semantics (expiresAt == now) are undocumented and untested for VEX exception expiry
# Description
`isExpired` in `adapters/v1/securityexception.go` decides whether a SecurityException/ClusterSecurityException vulnerability suppression has expired by checking `expiresAt.Before(now)`. This means an exception whose `expiresAt` is exactly equal to `now` (to the nanosecond) is treated as **not expired**, so its suppression stays active at that exact instant. That boundary choice is never documented on the function and isn't exercised by any existing test, so it's unclear whether "expires at exactly now" is meant to be inclusive (suppression should end at that instant) or exclusive (suppression should still apply through that instant), and a future refactor of the comparison could silently flip the behavior with nothing to catch it.

# Environment
OS: N/A (logic bug, reproducible via unit test on any platform)
Version: main branch, `adapters/v1/securityexception.go`

# Steps To Reproduce
1. `grep -n "func Test.*Expired" adapters/v1/securityexception_test.go` shows `TestConvertExpiredOnFix`, `TestConvertSkipsExpired`, and `TestConvertPerEntryExpiresAt*`.
2. Reading each test body shows they all use timestamps clearly in the past or future relative to execution time; none sets `expiresAt` equal to the injected/captured `now`.
3. Run `go test ./adapters/v1/... -run 'TestConvert.*Expired' -v` — all pass, confirming the existing suite never exercises the `expiresAt == now` boundary.
4. Look at `isExpired`:
   ```go
   func isExpired(expiresAt *metav1.Time, now time.Time) bool {
       return expiresAt != nil && expiresAt.Time.Before(now)
   }
   ```
   There is no doc comment stating whether equality should count as expired.

# Expected behavior
The exact boundary condition (`expiresAt == now`) should be explicitly documented on `isExpired` (stating the intended inclusive/exclusive semantics) and covered by a dedicated test case, so the behavior is intentional and protected against regression.

# Actual Behavior
`isExpired`'s comparison relies on `time.Before`'s natural semantics with no comment explaining the boundary, and no test in `adapters/v1/securityexception_test.go` sets `expiresAt` equal to the mocked/injected `now` to verify which way the boundary resolves.

# Additional context
Relevant area: `adapters/v1/securityexception.go` (see `isExpired`, `effectiveExpiresAt`, and the `ConvertToVulnerabilityExceptionPolicies` loop where `now := time.Now()` is captured once per call and reused for every exception in the batch).

Suggested fix direction: add a doc comment to `isExpired` clarifying the intended inclusive/exclusive boundary behavior, and add a table-driven test case with `expiresAt` set to exactly the injected `now` to lock in that behavior.

Acceptance criteria:
- A test explicitly covers `expiresAt == now`.
- `isExpired`'s doc comment states the intended inclusive/exclusive behavior.

rootp1:
## Overview
`isExpired` in `adapters/v1/securityexception.go` decides whether a SecurityException/ClusterSecurityException vulnerability suppression has expired, but it had no doc comment stating which way the boundary resolves when `expiresAt` is exactly equal to `now`, and no test exercised that exact instant. This PR documents the intended behavior (the boundary is exclusive, so an exception still suppresses through the instant it expires) and adds a test that locks it in.

## How to Test
Run `go test ./adapters/v1/... -run TestIsExpiredBoundary -v` to see the new boundary test pass, or `go test ./...` for the full suite.

## Related issues/PRs:

Resolved #709

put an [x] in the box to get it checked 

## Summary by CodeRabbit

* **Documentation**
  * Clarified that an expiration time equal to the current time remains active.

* **Tests**
  * Added coverage for expiration boundary conditions, including equal, earlier, and unspecified expiration times.

### DIFF
--- a/adapters/v1/securityexception.go
+++ b/adapters/v1/securityexception.go
@@ -102,6 +102,9 @@ func ConvertToVulnerabilityExceptionPolicies(exceptions []sev1beta1.SecurityExce
 	return policies, stats
 }
 
+// isExpired reports whether expiresAt has passed relative to now. The boundary is
+// exclusive: an expiresAt exactly equal to now is not yet expired, so the suppression
+// it governs remains active through that exact instant and only expires strictly after it.
 func isExpired(expiresAt *metav1.Time, now time.Time) bool {
 	return expiresAt != nil && expiresAt.Time.Before(now)
 }
--- a/adapters/v1/securityexception_test.go
+++ b/adapters/v1/securityexception_test.go
@@ -216,6 +216,16 @@ func TestConvertSkipsExpired(t *testing.T) {
 		"one expired namespaced and one expired cluster-scoped exception should each be counted once")
 }
 
+func TestIsExpiredBoundary(t *testing.T) {
+	now := time.Now()
+	exact := metav1.NewTime(now)
+
+	assert.False(t, isExpired(&exact, now), "expiresAt equal to now must not be treated as expired")
+	assert.True(t, isExpired(&metav1.Time{Time: now.Add(time.Nanosecond)}, now.Add(2*time.Nanosecond)),
+		"expiresAt strictly before now must be treated as expired")
+	assert.False(t, isExpired(nil, now), "a nil expiresAt must never be treated as expired")
+}
+
 func TestConvertMatchResources(t *testing.T) {
 	exceptions := []sev1beta1.SecurityException{
 		{


matthyx (approved):
Small, well-scoped change: adds a doc comment to `isExpired` clarifying the now-equality boundary and a test (`TestIsExpiredBoundary`) locking it in. No production logic changed — verified the doc comment accurately describes the existing `expiresAt.Time.Before(now)` behavior (exclusive boundary: equal-to-now is not expired). New test assertions are correct (equal→not expired, strictly-before→expired, nil→not expired). All CI checks pass and no other reviewer comments raised concerns. LGTM.

PR #703, opened 2026-08-14
fix: restore silenced Syft conversion warnings

doraem-on:
**Overview**
In `adapters/v1/domain_to_syft.go`, the SBOM parsing logic uses hand-copied Syft internals (originally from `v0.98.0`). During this copy, the logging inside `warnConversionErrors()` and adjacent conversion functions was completely commented out.

This led to a silent failure mode where any conversion failures—such as `unknown-relationship-type` errors and invalid CPE format errors during SBOM loading—were entirely swallowed. They were neither returned to the caller nor logged anywhere in the application. This silent dropping occurred on the critical path that feeds every cached-SBOM scan into Grype, potentially leading to incomplete vulnerability scans without any operational visibility.

This PR restores these warnings using the standard `kubescape/go-logger`, ensuring that invalid attributes, unknown relationship types, and file mode parsing failures are appropriately surfaced as warnings instead of being silently dropped.

**Which issue(s) this PR fixes:**
Fixes #702

**output logs**
A new test `Test_domainJSONToSyft_DropsErrors` has been added to verify that invalid CPEs and relationships are correctly parsed without panicking, and the new log output confirms that these omissions are now logged.

```text
=== RUN   Test_domainJSONToSyft_DropsErrors
[warning] excluding invalid Attributes. cpe: ; error: unable to parse Attributes string: failed to parse Attributes="": wfn: unsupported format ""
[warning] SBOM conversion error. error: "unknown relationship type: unknown-completely-invalid-relationship" occurred 1 time(s)
--- PASS: Test_domainJSONToSyft_DropsErrors (0.00s)
PASS
ok  	github.com/kubescape/kubevuln/adapters/v1	1.371s
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SBOM conversion resilience by handling malformed CPEs, invalid file modes, unknown relationship types, and other conversion issues without preventing successful parsing.
  * Added warnings to make recoverable conversion issues more visible.
* **Tests**
  * Added coverage confirming invalid entries are discarded while valid SBOM packages remain available.

PR #712, opened 2026-08-14
test(adapters): point the exception lookup tests at the live implementation

### LINKED ISSUE #711: the exception-suppression tests cover a function no scan reaches any more
#694 moved both callers of `getCVEExceptionMatchCVENameFromList` onto `cveExceptionIndex.lookup`:

```go
-	matched := getCVEExceptionMatchCVENameFromList(exceptions, m.Vulnerability.ID, isFixed)
+	matched := exceptionIndex.lookup(m.Vulnerability.ID, isFixed)
```

that was the right call and the index looks equivalent. but the old function was left behind, and it has no callers outside tests now:

```
$ grep -rn getCVEExceptionMatchCVENameFromList --include='*.go' . | grep -v _test
adapters/v1/backend_utils.go:341:func getCVEExceptionMatchCVENameFromList(...)
(and three comments)
```

seven test call sites still point at it. so the tests for how a `SecurityException` decides to suppress a finding, the at-most-once behaviour from #579, the `ExpiredOnFix` filtering, the alias matching and the case handling, all now exercise a function that nothing in a scan reaches. they pass whatever `lookup` does.

that is worth more than the dead lines: if the two ever disagree the tests keep saying it is fine, and the thing they cover is which vulnerabilities get hidden.

pointing the seven at `buildCVEExceptionIndex(list).lookup(...)` and deleting the old function is enough. all seven pass against the index unchanged, which is also the evidence that the two really are equivalent today, so there is nothing to preserve by keeping it.

version: main (be94585), follow-up to #694

AnshulPatil2005:
## Overview

#694 moved both callers of `getCVEExceptionMatchCVENameFromList` onto `cveExceptionIndex.lookup` and left the old function behind. it has no callers outside tests now, but seven test call sites still point at it, so the tests for how a `SecurityException` decides to suppress a finding cover a function no scan reaches: the at-most-once behaviour from #579, the `ExpiredOnFix` filtering, the alias matching and the case handling.

they pass whatever `lookup` does, which matters more than the dead lines. if the two ever disagree the tests keep saying it is fine, and what they cover is which vulnerabilities get hidden.

this points the seven at `buildCVEExceptionIndex(list).lookup(...)` and deletes the old function.

## Additional Information

all seven pass against the index unchanged, which is the evidence that the two are equivalent today and that nothing is lost by dropping the linear one. I checked the two against each other before touching anything:

- ordering: the index appends exception positions in ascending order while walking the list once, and `lookup` walks that ascending, so both return matches in list order.
- at most once per exception: the linear one `break`s out of the policy loop, the index keeps a per-exception `seen` set while building, so an exception listing an alias equal to its id contributes once either way.
- `ExpiredOnFix`: skipped before the policies in one, at lookup in the other, same set either way.
- case: `strings.EqualFold` against `strings.ToLower` on both sides. these part company on some non-ASCII input, `EqualFold` doing full case folding, but a CVE id or alias is ASCII.

the comments that named the old function are updated rather than left pointing at something that is gone, including the one on `lookup` itself, which described its result as "the same result getCVEExceptionMatchCVENameFromList would return" and now states the behaviour directly.

## How to Test

`go test ./adapters/v1/ -count=1`

no test is added or removed, seven change which function they call. `TestBuildCVEExceptionIndex_MatchesLinearLookup` from #694, `Test_getCVEExceptionMatchCVENameFromList`, the #579 at-most-once case and the `ExpiredOnFix` and alias cases all pass unchanged against the index.

## Related issues/PRs:

* Resolved #711
* Follow-up to #694

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved vulnerability exception matching for consistent suppression across multiple policies and actions.
  * Added case-insensitive matching, prevented duplicate policy matches, and excluded exceptions that expired after a fix.
  * Preserved exception ordering and existing suppression behavior.
* **Performance**
  * Accelerated vulnerability exception lookups, particularly for larger policy sets.
* **Documentation**
  * Clarified terminology in security exception documentation.

### DIFF
--- a/adapters/v1/backend_utils_test.go
+++ b/adapters/v1/backend_utils_test.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"strings"
 	"sync"
 	"testing"
 	"time"
@@ -25,6 +26,30 @@ import (
 	"github.com/stretchr/testify/require"
 	"k8s.io/utils/pointer"
 )
+
+// linearExceptionLookup is the straightforward scan the index replaced, kept here so the
+// equivalence test below has something independent to compare against. Production has one
+// implementation now, and an index checked against itself asserts nothing.
+func linearExceptionLookup(srcCVEList []armotypes.VulnerabilityExceptionPolicy, CVEName string, filterFixed bool) []armotypes.VulnerabilityExceptionPolicy {
+	var l []armotypes.VulnerabilityExceptionPolicy
+	for i := range srcCVEList {
+		if filterFixed && srcCVEList[i].ExpiredOnFix != nil && *srcCVEList[i].ExpiredOnFix {
+			continue
+		}
+		for j := range srcCVEList[i].VulnerabilityPolicies {
+			if strings.EqualFold(srcCVEList[i].VulnerabilityPolicies[j].Name, CVEName) {
+				// an exception contributes at most once however many of its policies carry
+				// the name, matching what the index does with its per-exception seen set
+				l = append(l, srcCVEList[i])
+				break
+			}
+		}
+	}
+	if len(l) > 0 {
+		return l
+	}
+	return nil
+}
 
 func TestGetCVEExceptionMatchCVENameFromList(t *testing.T) {
 	testCases := []struct {
@@ -206,16 +231,13 @@ func TestGetCVEExceptionMatchCVENameFromList(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			actual := getCVEExceptionMatchCVENameFromList(tc.srcCVEList, tc.CVEName, tc.isFixed)
+			actual := buildCVEExceptionIndex(tc.srcCVEList).lookup(tc.CVEName, tc.isFixed)
 			assert.Equal(t, tc.expected, actual)
-
-			indexed := buildCVEExceptionIndex(tc.srcCVEList).lookup(tc.CVEName, tc.isFixed)
-			assert.Equal(t, tc.expected, indexed)
 		})
 	}
 }
 
-// buildCVEExceptionIndex+lookup must return exactly what getCVEExceptionMatchCVENameFromList
+// buildCVEExceptionIndex+lookup must return exactly what a linear scan of the exception list
 // returns for the same inputs, since callers switched from the linear scan to the index to
 // avoid re-walking the full exception list per match.
 func TestCVEExceptionIndex_MatchesLinearScan(t *testing.T) {
@@ -239,7 +261,7 @@ func TestCVEExceptionIndex_MatchesLinearScan(t *testing.T) {
 
 	for _, cve := range []string{"CVE-2021-1234", "cve-2021-1234", "CVE-2021-5678", "CVE-9999-0000"} {
 		for _, filterFixed := range []bool{false, true} {
-			want := getCVEExceptionMatchCVENameFromList(srcCVEList, cve, filterFixed)
+			want := linearExceptionLookup(srcCVEList, cve, filterFixed)
 			got := index.lookup(cve, filterFixed)
 			assert.Equal(t, want, got, "cve=%s filterFixed=%v", cve, filterFixed)
 		}
@@ -268,7 +290,7 @@ func BenchmarkGetCVEExceptionMatch(b *testing.B) {
 	b.Run("LinearScanPerMatch", func(b *testing.B) {
 		for i := 0; i < b.N; i++ {
 			for _, cve := range cveNames {
-				getCVEExceptionMatchCVENameFromList(srcCVEList, cve, false)
+				linearExceptionLookup(srcCVEList, cve, false)
 			}
 		}
 	})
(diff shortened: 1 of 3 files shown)


adapters/v1/backend_utils_test.go
```diff
@@ -239,7 +239,7 @@ func TestCVEExceptionIndex_MatchesLinearScan(t *testing.T) {
 
 	for _, cve := range []string{"CVE-2021-1234", "cve-2021-1234", "CVE-2021-5678", "CVE-9999-0000"} {
 		for _, filterFixed := range []bool{false, true} {
-			want := getCVEExceptionMatchCVENameFromList(srcCVEList, cve, filterFixed)
+			want := buildCVEExceptionIndex(srcCVEList).lookup(cve, filterFixed)
```
matthyx: Blocker: this test no longer verifies what its name/doc claim.

`want` is now built the same way `got` is — both call `buildCVEExceptionIndex(srcCVEList).lookup(...)`:

```go
index := buildCVEExceptionIndex(srcCVEList)
...
want := buildCVEExceptionIndex(srcCVEList).lookup(cve, filterFixed)
got := index.lookup(cve, filterFixed)
assert.Equal(t, want, got, ...)
```

Since `getCVEExceptionMatchCVENameFromList` was deleted, `TestCVEExceptionIndex_MatchesLinearScan` is now comparing the index to itself. It's deterministic, so `want == got` by construction regardless of whether `buildCVEExceptionIndex`/`lookup` is actually correct — it can never fail, even if a real bug is introduced in the index logic. That directly undercuts the PR's stated goal of proving today's equivalence and catching future divergence.

CodeRabbit flagged the same gap in its review (line ~218 area: "add an independent test-only linear scan of the exception list, and compute `want` with that oracle"). That hasn't been addressed yet.

Suggest either:
- keep a small test-local linear-scan helper (unexported, `_test.go`-only) as the oracle for `want`, so this test still catches a regression in the index, or
- if that's no longer valuable now that there's only one implementation, delete this test rather than leave a tautological one with a misleading name/doc.

Separately, non-blocking: in `TestGetCVEExceptionMatchCVENameFromList` (~line 206-213), the `actual`/`indexed` pair now both call `buildCVEExceptionIndex(...).lookup(...)` with identical args and both assert against `tc.expected` — the second block is redundant and can be dropped.

Also agree with CodeRabbit's nitpick on `BenchmarkGetCVEExceptionMatch`: `b.Run("LinearScanPerMatch", ...)` (line ~268) now builds the index per match rather than doing a linear scan — the label is stale and should be renamed (e.g. `BuildIndexPerMatch`).

matthyx (changes requested):
Reviewed the change (deleting `getCVEExceptionMatchCVENameFromList` and pointing the seven callers at `buildCVEExceptionIndex(...).lookup(...)`). CI is green and the switch itself looks correct, but one test regression should be fixed before merge — see the inline comment on `backend_utils_test.go`.

`TestCVEExceptionIndex_MatchesLinearScan` lost its independent oracle: with the linear function gone, `want` and `got` are both computed via `buildCVEExceptionIndex(...).lookup(...)`, so the test now compares the index to itself and can't detect an actual regression in the index logic — despite its name and doc comment claiming to. CodeRabbit raised the same point earlier and it hasn't been addressed. Requesting changes so this test either gets a real (test-local) oracle back or is removed if it's no longer meaningful.

Also picked up two non-blocking nitpicks in the same file (redundant duplicate-assertion block in `TestGetCVEExceptionMatchCVENameFromList`, and CodeRabbit's stale `LinearScanPerMatch` benchmark label) — worth a quick cleanup in the same pass but not blockers on their own.

Happy to re-review once the oracle issue is addressed.

AnshulPatil2005:
you and coderabbit are right, and it was worse than just that test. my repoint was a blanket replace across the file, so it also took the two places that were deliberately not the index:

- `TestCVEExceptionIndex_MatchesLinearScan` ended up comparing the index with itself.
- the `LinearScanPerMatch` benchmark stopped scanning linearly and started rebuilding the index per match, so the comparison it exists for was index-rebuilt-per-match against index-built-once.
- and `TestGetCVEExceptionMatchCVENameFromList` asserted the same expectation twice, both calls having become the same one.

the linear scan is back as a test-local `linearExceptionLookup`, which the equivalence test and the benchmark baseline both use. production still has the one implementation.

it detects a real regression now: dropping the ExpiredOnFix skip from `lookup` fails it on the filterFixed cases, where before it passed whatever the index did.

test names still say getCVEExceptionMatchCVENameFromList. happy to rename them in this pass if you want, left alone for now to keep the diff to what you flagged.

matthyx (approved):
The oracle issue is fixed: `linearExceptionLookup` is now a test-local independent implementation, and `TestCVEExceptionIndex_MatchesLinearScan` compares against it instead of against itself. The redundant duplicate-assertion block in `TestGetCVEExceptionMatchCVENameFromList` was also removed, and the benchmark now genuinely does a linear scan again, so the `LinearScanPerMatch` label is accurate. Approving the code change.

Separate note, not a reason to hold this PR: the failing checks (lint, cross-platform build, build-and-push-image) are unrelated to this diff. They fail on `adapters/v1/domain_to_syft.go:73:19: undefined: deduplicateErrors` — that function was deleted as "dead code" by #717 (merged 16:48 UTC) and then called again when #703 (merged 17:22 UTC) restored the warning it fed, and this repo doesn't run build/lint on direct pushes to `main`, so the break slipped through. `main` is currently broken independently of this PR; merging will need that fixed first (or this branch rebased once it is), but it's not something to ask this PR's author to fix.

AnshulPatil2005:
stale, that was against 188c754. the benchmark calls linearExceptionLookup now instead of rebuilding the index per lookup, so LinearScanPerMatch is the accurate name as it stands.

matthyx:
rebase once https://github.com/kubescape/kubevuln/pull/739 merges

AnshulPatil2005:
> rebase once #739 merges

done

PR #714, opened 2026-08-14
fix(repositories): stop calling cloud exception suppressions external VEX

### LINKED ISSUE #713: suppressions from cloud exception policies are described as external VEX in the generated document
#671 stopped attributing every suppression to a `SecurityException`, which was right, but the branch it added claims something that is wrong for most of them:

```go
rule, ok := securityExceptionIgnoreRule(m)
if !ok {
	assessment.impactStatement = "Vulnerability was ignored by an external VEX document or scanner configuration"
	return assessment
}
```

`securityExceptionIgnoreRule` needs `SourceKind` to be `SecurityException` or `ClusterSecurityException`. only CRD-sourced policies ever have one: `buildIgnoreRule` takes it from `p.Attributes["sourceKind"]`, which `buildSuppressionAttributes` sets, and that runs from `buildPolicy`, which only the CRD path calls. a policy delivered by the backend goes through `ApplySecurityExceptions` like any other and comes out with a rule holding the vulnerability id and nothing else.

so a finding suppressed by a cloud exception ends up in the VEX document saying this:

```
cloud-sourced exception -> "Vulnerability was ignored by an external VEX document or scanner configuration"
CRD SecurityException   -> "Vulnerability was ignored by a SecurityException"
```

it was not ignored by an external VEX document or by scanner configuration. it was ignored by an exception policy, kubescape's own risk acceptance, delivered over the backend rather than as a CRD. cloud exceptions are the older of the two mechanisms, so on a cluster using them that wording is on every suppressed finding in the document, and the document is meant to be the audit artifact.

the case the branch was written for cannot happen yet. `ApplySecurityExceptions` is the only thing that writes `IgnoredMatches`, and Grype is handed no VEX documents and no ignore rules, so it contributes none of its own. consuming external VEX feeds (#387) is what would make it reachable.

the shapes are separable. Grype states an ignore rule against a package, `buildIgnoreRule` never sets one, so a rule carrying a package did not come from us. the fixture in `TestStoreVEX_ExternalVEXMasking` is commented as a Grype-native suppression but is `{{Vulnerability: "CVE-2023-1234"}}`, which is the shape our own code writes for a cloud policy.

version: main (be94585), follow-up to #671

AnshulPatil2005:
## Overview

#671 was right that a suppression without CRD provenance must not be attributed to a `SecurityException`. this keeps that. what it added in its place says something else that is wrong for most suppressions:

> Vulnerability was ignored by an external VEX document or scanner configuration

only CRD-sourced policies ever carry a `SourceKind`, since `buildIgnoreRule` reads it from the attributes `buildSuppressionAttributes` sets and that runs from `buildPolicy`, which only the CRD path calls. a policy delivered by the backend suppresses through the same code and comes out with a rule holding the vulnerability id and nothing else, so it takes the `!ok` branch and the document says it came from an external VEX document. it did not, it came from an exception policy, and cloud exceptions being the older of the two mechanisms that is every suppressed finding on a cluster using them.

so the branch now distinguishes the two, and the external wording is kept for a rule that really did not come from us.

## Additional Information

the shapes are separable: Grype states an ignore rule against a package and `buildIgnoreRule` never sets one, so a rule carrying a package did not come from us. `isOwnIgnoreRule` is that check.

nothing produces the external case today. `ApplySecurityExceptions` is the only writer of `IgnoredMatches`, and Grype is given no VEX documents and no ignore rules, so it contributes none. the wording is kept rather than dropped because #387 is what makes it reachable, and it is a named constant now like the other two instead of a literal in the branch.

`TestStoreVEX_ExternalVEXMasking` from #671 changes here. its fixture is commented as a Grype-native suppression but is `{{Vulnerability: "CVE-2023-1234"}}`, which is the shape our own code writes for a cloud policy, so it was asserting the misattribution. it now expects the exception-policy wording and still asserts the thing that test exists for, that the statement is not attributed to a `SecurityException`.

## How to Test

`TestIgnoredMatchAssessment_AttributesBySource` covers the four shapes: a namespaced CRD, a cluster-scoped CRD, a backend-delivered policy, and a rule carrying a package. before this the backend-delivered case fails with

```
expected: "Vulnerability was ignored by an exception policy"
actual  : "Vulnerability was ignored by an external VEX document or scanner configuration"
```

and all four assert the status is untouched, since every one of them is a suppression.

## Related issues/PRs:

* Resolved #713
* Follow-up to #671

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -1079,8 +1079,18 @@ func statementHasPURL(products []v1beta1.Product, purl string) bool {
 const defaultActionStatement = "Upgrade the vulnerable component to a version that is not affected"
 
 const (
-	defaultLocalImpactStatement         = "Vulnerable component is not loaded into the memory"
-	securityExceptionImpactStatement    = "Vulnerability was ignored by a SecurityException"
+	defaultLocalImpactStatement      = "Vulnerable component is not loaded into the memory"
+	securityExceptionImpactStatement = "Vulnerability was ignored by a SecurityException"
+	// cloudExceptionImpactStatement covers a suppression from an exception policy that did
+	// not come from a CRD, which today means one delivered by the backend. Those reach
+	// ApplySecurityExceptions like any other policy but never go through buildPolicy, so
+	// they carry no sourceKind and their ignore rule carries no SourceKind either.
+	cloudExceptionImpactStatement = "Vulnerability was ignored by an exception policy"
+	// externalIgnoreImpactStatement covers an ignore that did not come from our exception
+	// machinery at all. Nothing produces one today, ApplySecurityExceptions being the only
+	// writer of IgnoredMatches and Grype being given no VEX documents or ignore rules, but
+	// consuming external VEX feeds (#387) is what would.
+	externalIgnoreImpactStatement       = "Vulnerability was ignored by an external VEX document or scanner configuration"
 	securityExceptionAcceptedRiskAction = "A SecurityException accepted this vulnerability as an affected finding"
 )
 
@@ -1100,7 +1110,14 @@ func ignoredMatchAssessment(m v1beta1.IgnoredMatch) ignoredVEXAssessment {
 
 	rule, ok := securityExceptionIgnoreRule(m)
 	if !ok {
-		assessment.impactStatement = "Vulnerability was ignored by an external VEX document or scanner configuration"
+		// No CRD provenance does not make it someone else's. A backend-delivered exception
+		// policy suppresses through the same path and leaves a rule with only the
+		// vulnerability id on it, so calling that an external VEX document would describe
+		// most suppressions wrongly on any cluster using cloud exceptions.
+		assessment.impactStatement = externalIgnoreImpactStatement
+		if isOwnIgnoreRule(m) {
+			assessment.impactStatement = cloudExceptionImpactStatement
+		}
 		return assessment
 	}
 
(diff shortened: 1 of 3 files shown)


repositories/apiserver.go
```diff
@@ -1131,6 +1148,17 @@ func ignoredMatchAssessment(m v1beta1.IgnoredMatch) ignoredVEXAssessment {
 // This is distinct from the adapter's isExceptionSourcedIgnore because it strictly
 // matches the SourceKind rather than inferring provenance from rule shape, ensuring
 // that only explicit CRD-driven rules receive the SecurityException impact statement.
+// isOwnIgnoreRule reports whether an ignored match was suppressed by our own exception
+// machinery. buildIgnoreRule writes exactly one rule per suppression and never sets Package,
+// while Grype expresses its own ignore rules in terms of a package, so one carrying a package
+// did not come from us.
```
matthyx, line 1159: Nit: this new doc comment got inserted in the middle of the existing `securityExceptionIgnoreRule` doc block, so it now reads as documenting `isOwnIgnoreRule` while `securityExceptionIgnoreRule` below is left with no doc comment of its own. Worth splitting into two separate comment blocks, one directly above each function.

matthyx (dismissed):
Traced the claims against the actual code paths: `buildPolicy` (and thus `buildSuppressionAttributes`, which sets `sourceKind`) is only reached from the CRD conversion path (`securityexception.go:67,93`), while `BackendAdapter.GetCVEExceptions` seeds `vulnExceptionList` straight from the backend API response before CRD policies are appended — so a cloud/backend-delivered policy genuinely has no `sourceKind` attribute, and `buildIgnoreRule` never sets `Package`, matching the `isOwnIgnoreRule` check. `go build ./...`, `go vet ./repositories/...`, and `go test ./repositories/...` all pass, including the new `TestIgnoredMatchAssessment_AttributesBySource` cases and the updated `TestStoreVEX_ExternalVEXMasking`.

One nit left inline (doc-comment misplacement) — not a blocker. Good to merge.

matthyx:
rebase pls

repositories/apiserver.go
```diff
@@ -1136,6 +1153,17 @@ func ignoredMatchAssessment(m v1beta1.IgnoredMatch) ignoredVEXAssessment {
...
```
matthyx, line 1159: Nit (still applies after the rebase): this doc comment is inserted in the middle of the existing `securityExceptionIgnoreRule` doc block, so it now reads as documenting `isOwnIgnoreRule` while `securityExceptionIgnoreRule` below is left with no doc comment of its own. Worth splitting into two separate comment blocks, one directly above each function.

matthyx (approved):
Re-reviewed after the rebase (my previous approval got dismissed by the force-push to `8d4bccd`). The diff is content-identical to what I reviewed before — same logic, same tests — so the earlier verification still holds. Rebuilt and re-ran the suite against the new commit: `go build ./repositories/...`, `go vet ./repositories/...`, and `go test ./repositories/...` all pass, including `TestIgnoredMatchAssessment_AttributesBySource` and the updated `TestStoreVEX_ExternalVEXMasking`.

Unrelated FYI, not a blocker on this PR: `go build ./...` currently fails on `main` at `adapters/v1/domain_to_syft.go:73: undefined: deduplicateErrors` — traced it to #717 ("remove dead code and ineffectual assignment"), which deleted a function that's still called. Pre-existing on `main`, nothing to do with this branch's changes, but worth a separate fix.

Same non-blocking nit as before left inline. Approving — good to merge.

matthyx:
rebase once https://github.com/kubescape/kubevuln/pull/739 merged

PR #723, opened 2026-08-15
refactor(services): store the filtered CVE in one place

### LINKED ISSUE #722: ScanCVE and ScanRegistry carry the same twenty-two line store sequence
ScanCVE and ScanRegistry both finish matching by doing the same twenty-two lines: apply the current exception set, store the filtered manifest, store its summary, and publish the VEX if that exception set was complete.

```
core/services/scan.go:482-503   ScanCVE
core/services/scan.go:602-623   ScanRegistry
```

the two are identical apart from the wording of one comment, which says "matching how ScanCP gates its own call" in one and "matching how ScanCVE and ScanCP gate their own calls" in the other. that comment is itself the tell: the gating rule is stated three times, in three flows, and each copy has to name the other two.

these flows drifting apart is where #469, #501, #557, #608, #609 and #623 came from, which is why #638 pulled the cached-CVE reconciliation out into `reconcileCachedCVE` and said so in its comment:

> ScanCP, ScanCVE, and ScanRegistry each used to carry their own copy of this sequence, and that duplication is exactly why the same class of bug kept resurfacing in one flow after being fixed in another

the store sequence is the same shape and has not been pulled out yet. ScanCP cannot share it as it stands, its version carrying the relevancy manifest alongside the full one, but ScanCVE and ScanRegistry can.

one thing to keep if it moves: `applyExceptionsToManifest` is called outside the `if s.storage` guard, and it is also what reports the degraded, expired and active exception counts, so it has to keep running when storage is off.

version: main (be94585)

AnshulPatil2005:
## Overview

ScanCVE and ScanRegistry both finished matching with the same twenty-two lines: apply the current exception set, store the filtered manifest, store its summary, and publish the VEX if that exception set was complete. the two were identical apart from one comment's wording.

`storeFilteredCVE` is now that sequence, once. ScanCP keeps its own, which carries the relevancy manifest alongside the full one and is a different call in each of the three places it stores.

these flows drifting apart is where #469, #501, #557, #608, #609 and #623 came from, and #638 pulled `reconcileCachedCVE` out for exactly that reason. this is the same shape one step further along the same function.

## Additional Information

`applyExceptionsToManifest` stays outside the storage guard. It is also what reports the degraded, expired and active exception counts for the scan, so moving it inside would stop those being recorded on a deployment running without storage. The helper applies first and returns early after.

the store calls assigned to the enclosing `err` before, and nothing read it: both flows assign to `err` again at their next use, `SubmitCVE` in one and `SendStatus` in the other. so scoping those errors to the helper changes nothing and takes away a variable that looked like it carried a result out of the block and did not.

the gating comment used to name the other two flows from inside each copy, one saying "matching how ScanCP gates its own call" and the other "matching how ScanCVE and ScanCP gate their own calls". there is one copy of it now.

## How to Test

`go test ./core/... ./repositories/ -count=1`

no test changes. the existing ScanCVE and ScanRegistry tests cover both call sites, including `Test_ScanCVE`, `Test_ScanRegistry` and the exception-completeness cases that assert VEX is only published for a complete exception set, so them passing unchanged is what says the sequence is the same one.

## Related issues/PRs:

* Resolved #722

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved CVE scanning so configured exceptions are consistently applied across scans.
  - Preserved the original scan manifest when submitting results.
  - Ensured summaries and VEX output reflect filtered findings when applicable.
  - Prevented CVE storage issues from interrupting scan workflows.

### DIFF
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -479,29 +479,7 @@ func (s *ScanService) ScanCVE(ctx context.Context) error {
 			return &domain.ScanError{Reason: scanfailure.ReasonCVEMatchingFailed, Err: fmt.Errorf("scanning SBOM: %w", err)}
 		}
 
-		// apply security exceptions for storage (copy — original stays intact for SubmitCVE)
-		filteredCve, exceptionsComplete := s.applyExceptionsToManifest(ctx, cve)
-
-		// store filtered CVE
-		if s.storage {
-			domain.UpdateScanPhase(ctx, "result_storage")
-			err = s.cveRepository.StoreCVE(ctx, filteredCve, false)
-			if err != nil {
-				logger.L().Ctx(ctx).Warning("storing CVE", helpers.Error(err),
-					helpers.String("imageSlug", workload.ImageSlug))
-			}
-			err = s.cveRepository.StoreCVESummary(ctx, filteredCve, domain.CVEManifest{}, false)
-			if err != nil {
-				logger.L().Ctx(ctx).Warning("storing CVE summary", helpers.Error(err),
-					helpers.String("imageSlug", workload.ImageSlug))
-			}
-			// Only publish VEX built from a known-complete exception set, matching how
-			// ScanCP gates its own call: a degraded fetch would otherwise emit a document
-			// asserting fewer suppressions than the user actually configured.
-			if exceptionsComplete {
-				s.storeVEX(ctx, filteredCve, filteredCve, false, workload.ImageSlug)
-			}
-		}
+		s.storeFilteredCVE(ctx, cve, workload.ImageSlug)
 	} else {
 		cve, _, _ = s.reconcileCachedCVE(ctx, cve, workload.ImageSlug, true)
 	}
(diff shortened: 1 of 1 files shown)


matthyx (approved):
Verified the refactor is behavior-preserving:

- `storeFilteredCVE` reproduces the original `ScanCVE`/`ScanRegistry` sequence exactly — `applyExceptionsToManifest` still runs unconditionally before the `s.storage` check (so exception stats are still recorded when storage is off), and the VEX gate on `exceptionsComplete` is unchanged.
- Confirmed the claim that dropping the outer `err` assignment is safe: in both `ScanCVE` (line ~490, `SubmitCVE`) and `ScanRegistry` (line ~587, `SendStatus`), `err` is reassigned before it's next read, so scoping it to the new helper changes nothing observable.
- `ScanCP` is correctly left alone since its storage sequence differs (relevancy manifest).
- Ran `go build ./...`, `go vet ./...` (pre-existing unrelated vet warnings in `controllers/http.go`, untouched by this diff), and `go test ./core/... ./repositories/ -count=1` — all green.
- All CI checks (lint, tests, CodeQL, DCO, GitGuardian) are passing.

No blockers. Good to merge.

PR #725, opened 2026-08-15
refactor(repositories): share the create-or-update sequence in StoreSBOM

### LINKED ISSUE #724: StoreSBOM writes the same create-or-update sequence twice, and four other stores repeat it
every write to storage in `repositories/apiserver.go` is the same shape: `Create`, and on `AlreadyExists` a `RetryOnConflict` that reads the object back, merges annotations and labels onto it, sets the spec, and updates. it appears six times:

```
577   StoreCVE
820   StoreCVESummary
908   StoreCVESummaryStub
974   StoreVEX
1711  StoreSBOM   (filtered)
1748  StoreSBOM   (unfiltered)
```

the last two are the same thirty-seven lines twice inside one function, differing only in which client they call, which object they pass, and whether the word "filtered" appears in the log line. `SBOMSyftFiltered` and `SBOMSyft` have the same shape, `convertToFilteredSBOM` copies all four fields across, and the generated clients differ only in the object type, so there is nothing to tell the two branches apart beyond the names.

the other four are the same skeleton with real differences in the middle: `StoreCVESummaryStub` reads the full object rather than metadata only, because it needs the spec to check `summaryHasVulnerabilityData` before deciding not to overwrite a real summary with a stub, and `StoreCVE` and `StoreCVESummary` carry an extra field in their log lines.

the pair inside `StoreSBOM` can be collapsed as it stands. the other four want a little more from the helper first, and are worth doing separately rather than bending the shape around all six at once.

version: main (be94585)

AnshulPatil2005:
## Overview

`StoreSBOM` carried the create-then-read-modify-write sequence twice, thirty-seven lines each, once per branch of its own `if isFiltered`. The two differ only in which client they call, which object they pass, and whether the log line says "filtered SBOM" or "SBOM".

`createOrUpdate` is that sequence once. The two branches are eight lines each now.

## Additional Information

the generated clients are nominally distinct per resource but identical in shape, so `objectStore[T]` is satisfied by each for its own object type and the helper does not need to know which resource it is writing. that is the whole reason the two branches could be the same thirty-seven lines in the first place.

the merge is a callback because that is the part that genuinely differs: it works on the concrete type, and it runs inside the retry rather than once, since a conflict means someone wrote in between and the merge has to happen again on top of what they wrote. that was already true of both branches and is unchanged.

log messages and wrapped errors are the same strings as before, `kind` supplying the "filtered SBOM"/"SBOM" that used to be written into each copy.

the same skeleton appears four more times, in `StoreCVE`, `StoreCVESummary`, `StoreCVESummaryStub` and `StoreVEX`, and this does not touch them. they are not quite this shape: `StoreCVESummaryStub` reads the full object rather than metadata only, needing the spec to check `summaryHasVulnerabilityData` before it declines to overwrite a real summary with a stub, and two of the others carry an extra field in their log lines. those want a couple more parameters, and adding them speculatively here would shape the helper around callers it does not have yet.

## How to Test

no test changes. the tests for this come in mirrored pairs, one per branch, which is the duplication showing up a second time:

```
TestAPIServerStore_StoreSBOM_transientError                    / ...StoreSBOMFiltered_transientError
TestAPIServerStore_StoreSBOM_updateGetFailure_transientError   / ...StoreSBOMFiltered_updateGetFailure_transientError
TestAPIServerStore_StoreSBOM_retryExhausted_transientError     / ...StoreSBOMFiltered_retryExhausted_transientError
TestAPIServerStore_StoreSBOM_ctxPropagated                     / ...StoreSBOMFiltered_ctxPropagated
```

they cover the create, the read-modify-write behind `AlreadyExists`, a failing read inside the retry, the retry giving up, and the context reaching each call. all eight pass unchanged, and both halves of each pair now run the same code, so what used to be two sets of assertions about two copies is two sets about one.

## Related issues/PRs:

* Resolved #724

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -1706,81 +1706,19 @@ func (a *APIServerStore) StoreSBOM(ctx context.Context, sbom domain.SBOM, isFilt
 	manifest.Annotations[helpersv1.StatusMetadataKey] = sbom.Status // for the moment stored as an annotation
 	if isFiltered {
 		filtered := convertToFilteredSBOM(&manifest)
-		_, err := a.StorageClient.SBOMSyftFiltereds(a.Namespace).Create(ctx, filtered, metav1.CreateOptions{})
-		switch {
-		case errors.IsAlreadyExists(err):
-			retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-				if err := ctx.Err(); err != nil {
-					return err
-				}
-				// retrieve the latest version before attempting update
-				// RetryOnConflict uses exponential backoff to avoid exhausting the apiserver
-				result, getErr := a.StorageClient.SBOMSyftFiltereds(a.Namespace).Get(ctx, sbom.Name, metav1.GetOptions{ResourceVersion: "metadata"})
-				if getErr != nil {
-					return getErr
-				}
-				// update the filtered sbom
-				result.Annotations = mergeMaps(result.Annotations, filtered.Annotations)
-				result.Labels = mergeMaps(result.Labels, filtered.Labels)
-				result.Spec = filtered.Spec
-				// try to send the updated filtered sbom
-				_, updateErr := a.StorageClient.SBOMSyftFiltereds(a.Namespace).Update(ctx, result, metav1.UpdateOptions{})
-				return updateErr
+		return createOrUpdate(ctx, a.StorageClient.SBOMSyftFiltereds(a.Namespace), "filtered SBOM", sbom.Name, filtered,
+			func(existing *v1beta1.SBOMSyftFiltered) {
+				existing.Annotations = mergeMaps(existing.Annotations, filtered.Annotations)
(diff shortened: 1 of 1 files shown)


matthyx (approved):
Traced both branches of the new `StoreSBOM` against the old code line by line, and checked the generic `createOrUpdate`/`objectStore[T]` helper: the create → `AlreadyExists` → retry-on-conflict → get/merge/update sequence, the log lines, and the wrapped error strings are all preserved exactly, just parameterized by `kind`/`store`/`merge`. The `if isFiltered { return ... }` / fallthrough `return ...` structure correctly mirrors the original if/else. No behavior change.

CI (build, lint, CodeQL, GitGuardian, DCO, tests) is green, and the description's claim that `StoreCVE`/`StoreCVESummary`/`StoreCVESummaryStub`/`StoreVEX` are intentionally left untouched checks out — this PR only touches `apiserver.go` in the `StoreSBOM` region.

No blockers. Good to merge.

Issue #566, opened 2026-08-10
Track diff-based lint verification separately from full-tree lint cleanup


rootp1:
## Overview
The repository currently mixes several different linting goals into one bucket: keeping contributor verification usable today, wiring the same behavior into PR CI, and cleaning up the existing full-tree lint backlog over time.

## Problem
Right now those concerns are easy to conflate. Contributors need a branch-scoped lint workflow that tells them whether their changes introduced new issues. Maintainers also need PR CI to enforce the same behavior automatically, and separately need a plan for paying down the existing repository-wide lint findings. Treating all of that as one task creates confusing expectations and makes the workflow harder to reason about.

## Solution
Track this as three coordinated parts:

1. Establish and document the supported diff-based lint workflow for contributor verification.
2. Wire the same branch-scoped lint behavior into PR CI so local and remote checks match.
3. Track the exact current full-tree lint findings and clean them up in smaller follow-up passes until enforcing a full-tree green baseline becomes practical.

Success for part 1 means contributors can run the documented lint workflow without being blocked by unrelated baseline issues.

Success for part 2 means pull requests are checked in CI with the same diff-based lint model used locally.

Success for part 3 means the current backlog is explicitly tracked and reduced in reviewable batches instead of one oversized PR.

## Proposed PR Breakdown For Part 3
The full-tree cleanup should be done as **6 focused PRs**. The goal is to keep each review small, avoid mixing mechanical and behavior-sensitive changes, and make the backlog reduction visible.

| PR | Scope | Goal | Notes |
| --- | --- | --- | --- |
| 1 | Baseline capture | Record the current `make lint-all` backlog grouped by package and linter | No functional code changes. This creates the reference point for the follow-up cleanup PRs. |
| 2 | Test-only cleanup | Fix lint findings in test files across lower-risk areas first | Prefer `*_test.go` updates only. Keep production behavior out of this PR. |
| 3 | Internal and utility packages | Fix non-test lint findings in `internal/`, `config/`, and small utility packages | Restrict to low-blast-radius packages and mostly mechanical changes. |
| 4 | Adapter and controller cleanup | Fix lint findings in `adapters/` and `controllers/` production code | Review separately because these packages are closer to request/response and orchestration behavior. |
| 5 | Core and repository cleanup | Fix lint findings in `core/` and `repositories/` production code | Keep this isolated because service and repository changes are more behavior-sensitive. |
[...]

PR #727, opened 2026-08-15
fix(controllers): stop copying HTTPController's sync.Once by value

### LINKED ISSUE #726: go vet fails on main: HTTPController's sync.Once is copied by four value receivers
`go vet` does not pass on main:

```
$ go vet ./controllers/
controllers\http.go:99:9:  recordScan passes lock by value: controllers.HTTPController contains sync.Once contains sync.noCopy
controllers\http.go:133:9: recordRejection passes lock by value: ...
controllers\http.go:206:9: Alive passes lock by value: ...
controllers\http.go:211:9: Ready passes lock by value: ...
```

`HTTPController` picked up a `sync.Once` when #636 added the scan status store:

```go
func (h *HTTPController) ensureStatuses() *scanStatusStore {
	h.statusesOnce.Do(func() { ... })
	return h.statuses
}
```

four of its methods still take a value receiver, so each call copies the struct and the `Once` inside it.

nothing is broken today. none of those four reach `ensureStatuses`, so no second store gets built. it is the next line of code that is the problem: a copy's `Once` starts unfired, so a value-receiver method that called `ensureStatuses` would allocate its own `scanStatusStore` and record status into an object no request ever reads, and `/v1/scanStatus/:jobID` would keep answering 404 for a job it had accepted. `Alive` and `Ready` are gin handlers, and adding a status or queue-depth line to one of them is an ordinary thing to do.

CI does not catch it. lint runs against the diff rather than the whole tree (#566/#567), so these four have sat on main since the `Once` arrived without turning anything red.

changing the four receivers to pointers is enough. production only ever holds a `*HTTPController`, `NewHTTPController` returning one, and the tests build addressable locals, so nothing needs to change at the call sites.

version: main (be94585)

AnshulPatil2005:
## Overview

`go vet ./controllers/` fails on main with four copylocks reports. `HTTPController` picked up a `sync.Once` when #636 added the scan status store, and `recordScan`, `recordRejection`, `Alive` and `Ready` still take a value receiver, so each call copies the struct and the `Once` in it.

this makes them pointer receivers. `go vet ./...` is clean afterwards.

## Additional Information

nothing is broken today: none of those four reach `ensureStatuses`, so no second store is built. it is what comes next that this is about. a copy's `Once` starts unfired, so a value-receiver method that called `ensureStatuses` would allocate its own `scanStatusStore` and record into an object no request reads, leaving `/v1/scanStatus/:jobID` answering 404 for a job it had just accepted. `Alive` and `Ready` are gin handlers, and putting a status or queue-depth line in one is an ordinary thing to want.

nothing changes at the call sites. production only ever holds a `*HTTPController`, `NewHTTPController` returning one, and the tests build addressable locals, so the method calls resolve the same way.

lint runs against the diff rather than the whole tree (#566/#567), which is why four of these sat on main from the commit that added the `Once` without turning anything red.

## How to Test

`go vet ./... && go test ./controllers/ -race -count=1`

`TestHTTPController_NoValueReceiverMethods` asks the value type for its method set, which is empty once every method takes a pointer, and names any that do not. putting the receiver back on `Alive` alone fails it with

```
Should be empty, but was [Alive]
these take a value receiver and so copy the sync.Once: [Alive]
```

reflect lists exported methods only, which is where the risk sits: those are the gin handlers. `recordScan` and `recordRejection` are unexported and stay covered by `go vet` rather than by this.

## Related issues/PRs:

* Resolved #726

### DIFF
--- a/controllers/http.go
+++ b/controllers/http.go
@@ -96,7 +96,7 @@ func (h *HTTPController) claimTrackedJob(jobID string) bool {
 // domain.ErrPartialContainerProfile result is a warning-level expected outcome, not a failure.
 // err is the error returned by the scan (nil for success/partial) and is only consulted to
 // resolve the bounded reason label on a failed outcome; see scanFailureReason.
-func (h HTTPController) recordScan(ctx context.Context, endpoint string, start time.Time, outcome string, err error) {
+func (h *HTTPController) recordScan(ctx context.Context, endpoint string, start time.Time, outcome string, err error) {
 	if h.metrics == nil {
 		return
 	}
@@ -130,7 +130,7 @@ func scanFailureReason(outcome string, err error) string {
 // "too_many_requests" for registry back-pressure (ErrTooManyRequests) and "invalid_request"
 // for every other validation error, keeping the rejection-rate signal distinct from
 // malformed-payload noise while staying low cardinality.
-func (h HTTPController) recordRejection(ctx context.Context, endpoint string, err error) {
+func (h *HTTPController) recordRejection(ctx context.Context, endpoint string, err error) {
 	if h.metrics == nil {
 		return
 	}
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed. The fix is correct: `recordScan`, `recordRejection`, `Alive`, and `Ready` were the last four `HTTPController` methods still on a value receiver, so each call copied the struct's `sync.Once` — consistent with every other method on this type already using `*HTTPController`. `NewHTTPController` returns a `*HTTPController` and both `cmd/http/main.go` and `main_test.go` bind the handlers off a pointer-typed `controller`, so this changes no call sites. All CI checks (build, lint, tests) are green, and `TestHTTPController_NoValueReceiverMethods` is a solid regression guard against this recurring — good catch tying it to the actual `go vet` failure mode. LGTM, approving.

PR #717, opened 2026-08-14
fix: remove dead code and ineffectual assignment

rootp1:
## Overview
Removes dead code flagged by golangci-lint: an unused helper function, two unused test fixture variables, and an ineffectual assignment.

## Additional Information
`golangci-lint run ./...` was reporting these four findings:
- `adapters/v1/domain_to_syft.go:78` - `deduplicateErrors` had no callers (its only call site was commented out)
- `adapters/v1/backend_test.go:51,56` - `securityExceptionGVR` and `clusterSecurityExceptionGVR` were declared but never referenced anywhere. Checked git history and they were never wired into any test since they were added, so removing them does not drop any coverage.
- `core/services/scan.go:301` - `filteredCve := domain.CVEManifest{}` was always overwritten before being read, so it's now declared with `var filteredCve domain.CVEManifest` instead.

## How to Test
- `go build ./...`
- `go vet ./adapters/... ./core/...`
- `golangci-lint run --enable-only unused,ineffassign ./...` (0 issues, confirms all four findings are resolved)
- `go test ./adapters/... ./core/services/...`

## Related issues/PRs:
Resolved #707

## Checklist before requesting a review
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] New and existing unit tests pass locally with my changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scan processing so cached vulnerability information remains available for subsequent VEX handling.
* **Refactor**
  * Removed unused test declarations, imports, and obsolete internal error-handling code.

PR #730, opened 2026-08-15
test(adapters): cover deduplicateErrors so it does not read as dead code again

### LINKED ISSUE #729: main does not build: undefined: deduplicateErrors
main does not build:

```
$ go build ./...
# github.com/kubescape/kubevuln/adapters/v1
adapters/v1/domain_to_syft.go:73:19: undefined: deduplicateErrors
```

every open PR is red on build-and-push-image, lint and the cross-platform build, and the failure is in a file most of them do not touch.

two PRs did this between them, neither wrong on its own:

`deduplicateErrors` had been dead for a long time. its only reference was a commented-out line inside `warnConversionErrors`, left over from when the syft logging was stubbed out:

```go
func warnConversionErrors[T any](converted []T, errors []error) []T {
	//errorMessages := deduplicateErrors(errors)
	//for _, msg := range errorMessages {
	//	log.Warn(msg)
	//}
	return converted
}
```

#717 (a951c9d) removed the function, correctly, as dead code.

#703 (085f60b) uncommented the call, which was the whole point of that PR:

```go
errorMessages := deduplicateErrors(errors)
for _, msg := range errorMessages {
	logger.L().Warning("SBOM conversion error", helpers.String("error", msg))
}
```

one deletes lines 78 to 89, the other edits lines 71 to 73. git merges that clean and the result does not compile. both were green when they were reviewed because each was branched before the other landed.

the fix is to put the function back, since #703 wanted those warnings and they are the reason the call exists now. re-commenting the call would undo #703 instead.

worth noting the same thing can happen again: a helper whose only caller is commented out looks exactly like dead code to the next person reading it, and nothing in CI distinguishes the two.

version: main (e36e17f)

AnshulPatil2005:
## Overview

#739 restored `deduplicateErrors`, so this is now just the tests for it. rebased onto that, the function half of this PR dropped out as already applied and what is left is `adapters/v1/domain_to_syft_test.go`.

## Additional Information

the function was deleted once already. #717 removed it as dead code, correctly at the time, because its only reference was a commented-out line inside `warnConversionErrors`. #703 then uncommented that line and the merge stopped building.

what made it look dead is that nothing except one commented-out caller mentioned it. a test is what keeps that from reading the same way next time.

## How to Test

`go test ./adapters/v1/ -run 'Test_deduplicateErrors|Test_warnConversionErrors' -count=1`

`Test_deduplicateErrors` covers nothing, one error, the same error three times, and two distinct errors keeping separate counts. it compares as a set, since the order comes from a map range.

it holds the counting rather than just calling the function: dropping the accumulation to `errorCounts[e.Error()] = 1` fails the third and fourth cases.

`Test_warnConversionErrors_PassesConvertedThrough` pins the part that must not change, that `warnConversionErrors` returns its input untouched whatever the errors say.

## Related issues/PRs:

* Resolved #729

### DIFF
--- a/adapters/v1/domain_to_syft_test.go
+++ b/adapters/v1/domain_to_syft_test.go
@@ -1,6 +1,7 @@
 package v1
 
 import (
+	"errors"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


AnshulPatil2005:
rebased this onto #739, the function dropped out as already applied so what is left is just the tests for it. keeping them because the reason it got deleted in #717 was that nothing but a commented-out line referenced it.

matthyx (approved):
Verified: `deduplicateErrors`/`warnConversionErrors` in `adapters/v1/domain_to_syft.go` on `main` are already byte-identical to this branch's version (restored by #739), so this PR is test-only as described — no production code changes to review.

Checked the test logic against the actual implementation:
- `Test_deduplicateErrors`: all four cases (nil, single, repeated, distinct) match `deduplicateErrors`'s counting behavior, and `assert.ElementsMatch` correctly handles the map-range ordering nondeterminism.
- `Test_warnConversionErrors_PassesConvertedThrough`: correctly pins that the generic passthrough returns `converted` untouched regardless of errors.

All CI checks (build, lint, tests, CodeQL, GitGuardian, DCO) are green. No blockers found — approving.

PR #734, opened 2026-08-16
fix: close TOCTOU race between SecurityException cache invalidation and in-flight List()

### LINKED ISSUE #733: SecurityException cache: in-flight List() can win a race against informer invalidation, re-populating stale exception data
# Description

`enableSecurityExceptionCacheInvalidation` (`repositories/apiserver.go:216-246`, added in #653 to close #652) wires a dynamic informer that evicts `securityExceptionListCache` entries as soon as the backing `SecurityException`/`ClusterSecurityException` CRDs change, instead of waiting for the 30s TTL. But the evict-then-repopulate sequence has no synchronization between the informer's `Delete` and an in-flight `List()`'s `Set`, so a `List()` started *before* a CRD change can still write its (now stale) result back into the cache *after* the informer has already invalidated it — silently undoing the invalidation and reintroducing the exact staleness window #653 was written to eliminate.

`listSecurityExceptions`/`listClusterSecurityExceptions` (`repositories/apiserver.go:340-368`, `370-400`) both follow the same pattern:

```go
if cached, ok := a.securityExceptionListCache.Get(cacheKey); ok {
    return cached...., nil
}
seList, err := a.DynamicClient.Resource(...).List(listCtx, metav1.ListOptions{})
...
a.securityExceptionListCache.Set(cacheKey, exceptions, securityExceptionListCacheTTL) // unconditional
```

The underlying cache, `github.com/akyoto/cache` (`v1.0.6`, `cache.go`), backs `Set`/`Delete`/`Get` with a plain `sync.Map` — `Set` is an unconditional `Store` with no compare-and-swap, version, or generation check available to callers. There is no way, using this library as currently used, to make the post-`List()` `Set` conditional on "no invalidation happened since I started."

# Reproduction / Evidence

Interleaving that reproduces the bug (single namespace, single cache key `se/<namespace>`):

1. Goroutine A (servicing a scan request) calls `GetSecurityExceptions`, cache-misses on `se/ns1`, and starts `DynamicClient....List(listCtx, ...)` against the apiserver.
2. Before A's `List()` returns, a user deletes or narrows a `SecurityException` in `ns1` (e.g. during an incident response, revoking a suppression that was masking a newly-critical CVE).
3. The informer's `DeleteFunc`/`UpdateFunc` fires almost immediately and calls `invalidateSecurityExceptionCacheForObject`, which does `a.securityExceptionListCache.Delete("se/ns1")`.
4. A's in-flight `List()` (started in step 1, before the edit) returns the **pre-change** result and calls `a.securityExceptionListCache.Set("se/ns1", exceptions, 30*time.Second)` — silently re-populating the cache with stale data, *after* the invalidation already ran.
5. Nothing invalidates the cache again until the fixed 30s TTL from step 4 expires. Every scan in `ns1` during that window (`GetCVEExceptions` → `core/services/scan.go:748` → exception-application in the manifest) is served the stale, already-revoked exception, exactly the guarantee #653 exists to provide.

This is a genuine TOCTOU race, not a hypothetical: `List()` against a real apiserver is not instantaneous (network RTT + apiserver processing), and the informer's watch event for a change typically arrives with comparable or lower latency than a fresh `List()` call for the same resource, so step 2-3 landing inside step 1's `List()` window is a realistic, not contrived, timing.

Existing tests (`repositories/apiserver_test.go:2225-2225, 2267, 2296, 2316`) only cover single-shot invalidation (`Set` then `Delete` then `Get` misses) — none exercise the interleaving where a `List()` in flight at invalidation time completes afterward. The race is directly reproducible in a test using a fake dynamic client whose `List()` implementation is held open on a channel until after `invalidateSecurityExceptionCacheForObject` is called.

# Root Cause

`Set()` after `List()` in `listSecurityExceptions`/`listClusterSecurityExceptions` is unconditional — it has no way to know whether an invalidation happened for that key while the `List()` call was in flight, so a `List()` that raced a CRD change wins the race against the informer and clobbers the correctly-invalidated cache state.

# Impact

`GetSecurityExceptions` backs CVE suppression policy (`core/services/scan.go:748`, `adapters/v1/backend.go:284`). A revoked or narrowed `SecurityException`/`ClusterSecurityException` can continue suppressing a CVE in scan results for up to the full cache TTL (30s) after the CRD change is confirmed applied by the Kubernetes API — this is a security/compliance-relevant staleness gap (e.g. during incident response, an operator narrowing an exception to stop masking a critical CVE has no guarantee the scanner reflects that promptly), and it is silent: nothing logs or signals that a `Set` raced an invalidation.

# Affected Components

- `repositories/apiserver.go`: `listSecurityExceptions`, `listClusterSecurityExceptions`, `enableSecurityExceptionCacheInvalidation`, `invalidateSecurityExceptionCacheForObject`, `invalidateClusterSecurityExceptionCache`, `invalidateAllSecurityExceptionCaches`
- Consumers of `GetSecurityExceptions`: `adapters/v1/backend.go` (`GetCVEExceptions`), `core/services/scan.go` (exception application during scans)

# Acceptance Criteria

- A `List()` result that was in flight when an invalidating CRD event occurred for its key must never be written to the cache (i.e. the cache converges to "evicted" / fresh-on-next-read, not back to the stale value).
- No change to `GetSecurityExceptions`'s external behavior on the non-racing path (still cache-hit within TTL, still evict-then-refetch on CRD change with no interleaving).
- Regression test(s) that deterministically reproduce the race (e.g. a fake `List()` blocked on a channel, invalidation fired while blocked, then unblocked) and assert the cache does **not** end up holding the stale value.
- `go test ./repositories/...` and `make verify` pass.

# Suggested Solution Direction

Stamp each cache entry (or track per-key) with a monotonically increasing generation/version counter that `invalidateSecurityExceptionCacheForObject`/`invalidateClusterSecurityExceptionCache`/`invalidateAllSecurityExceptionCaches` bump on every invalidation. `listSecurityExceptions`/`listClusterSecurityExceptions` capture the generation before calling `List()`, and only call `Set()` if the generation is unchanged when `List()` returns (compare-and-swap semantics layered on top of `akyoto/cache`, e.g. via a small wrapper struct `{data, seenGeneration}` stored as the cached value, checked against a `sync/atomic` counter per key or per store). This keeps the fix scoped to `repositories/apiserver.go` without needing to replace the underlying cache library.

bhuvan-somisetty:
## Problem

`GetSecurityExceptions` caches `SecurityException`/`ClusterSecurityException` `List()` results in `securityExceptionListCache`, invalidated by a dynamic informer as soon as the backing CRDs change (#653). But the evict-then-repopulate sequence has no synchronization between the informer's invalidation and an in-flight `List()`'s cache write: a `List()` started *before* a CRD change can still write its now-stale result back to the cache *after* the informer's invalidation for that key already ran, silently undoing the invalidation.

## Root Cause

`listSecurityExceptions`/`listClusterSecurityExceptions` called `securityExceptionListCache.Set()` unconditionally after `List()` returned, with no way to know whether an invalidation had happened for that key while `List()` was in flight. `github.com/akyoto/cache` backs `Set`/`Delete` with a plain `sync.Map` — no compare-and-swap or versioning available to build that check on top of directly.

## Fix

Each cache key now gets a small `securityExceptionCacheEntry{mu sync.Mutex, generation uint64}`, created lazily. Invalidation bumps the generation under the key's mutex; a list call snapshots the generation before calling `List()` and only writes the result back — under that same mutex — if the generation is unchanged. Whichever of an invalidation and a racing write acquires the mutex last determines the outcome, and the result is correct either way:

- If invalidation runs last: cache ends up evicted (correct — the change is real and later).
- If the write runs last but the generation moved: the write is skipped, leaving the cache in the invalidated state.

The `List()` call itself stays outside the lock: holding it across a network call would let a slow `List()` delay an unrelated invalidation for the same key, and `client-go` informers deliver events to a handler serially, so that could stall processing of *other* CRD events too.

`invalidateAllSecurityExceptionCaches` (the fallback for CRD events `unstructuredFromEvent` can't resolve to a specific namespace) now ranges over the generation-entry map instead of the value cache, so it also covers a key whose `List()` is in flight but has no cached value yet.

## Testing

- Two new regression tests reproduce the exact race from the issue using a hook that blocks the fake dynamic client's `List()` call mid-flight, fires the real invalidation handler, then unblocks it — one for the namespaced path, one for cluster-scoped. **Verified both fail against the pre-fix code** (reverted the fix locally, confirmed both tests fail with the stale value present in the cache, then restored the fix).
- Direct unit tests of the generation compare-and-swap (`trySetSecurityExceptionCache`) for both the racing and non-racing paths, and of `invalidateAllSecurityExceptionCaches` covering an in-flight key.
- All pre-existing `repositories` tests still pass unchanged, confirming no behavior change on the non-racing path.
- `go build ./repositories/...`, `go vet ./repositories/...`, `go test ./repositories/...` all pass.
- `go test ./...`: every package that builds passes, including `repositories`. `adapters/v1` and its dependents (`cmd/http`, `controllers`, `core/services`) currently fail to build on `main` independent of this change — `adapters/v1/domain_to_syft.go:73` calls an undefined `deduplicateErrors`, introduced by #703. This PR does not touch that path; confirmed by checking out a clean `main` and reproducing the same build failure there.
- Could not run `go test -race` locally (no cgo/gcc in this environment); the new tests are goroutine-based and channel-synchronized (no sleeps), so they should be race-detector-clean, but I'd appreciate CI confirming it.

## Impact

Closes the residual staleness window from #733: a revoked/narrowed `SecurityException` or `ClusterSecurityException` now reliably takes effect as soon as the informer processes the change, instead of possibly being silently re-cached by a racing `List()` for up to the full cache TTL (30s).

Fixes #733

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented outdated SecurityException list results from repopulating the cache after invalidation.
  * Improved cache invalidation for both namespaced and cluster-scoped results, including requests already in progress.

* **Tests**
  * Added coverage for cache invalidation races, generation mismatches, successful updates, and invalidate-all behavior.

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -11,6 +11,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/akyoto/cache"
@@ -83,7 +84,76 @@ type APIServerStore struct {
 	// literals directly instead of through the constructors below).
 	securityExceptionListCache *cache.Cache
 
+	// securityExceptionCacheEntries backs the compare-and-swap that keeps a List() in flight
+	// when a CRD change invalidates its cache key from silently re-populating the cache with
+	// its now-stale result afterward — see securityExceptionCacheEntry's doc comment and #733.
+	// Zero value (empty sync.Map) is ready to use, so this needs no constructor initialization.
+	securityExceptionCacheEntries sync.Map // cacheKey (string) -> *securityExceptionCacheEntry
+
 	securityExceptionInformerStop context.CancelFunc
+}
+
+// securityExceptionCacheEntry pairs a cache key's invalidation generation with the mutex that
+// makes reading/bumping it, and conditionally writing securityExceptionListCache for that key,
+// atomic with respect to each other. See securityExceptionCacheEntry (the method) for why this
+// exists: a plain generation counter alone still leaves a check-then-act gap between reading it
+// and calling Set().
+type securityExceptionCacheEntry struct {
+	mu         sync.Mutex
+	generation uint64
+}
+
+// securityExceptionCacheEntry returns the entry for cacheKey, creating it on first use. Entries
+// are never removed: the number of distinct keys is bounded by the number of namespaces that
+// have ever had SecurityExceptions listed, plus one for the cluster-scoped key, which is small
+// and stable for the life of the process.
+func (a *APIServerStore) securityExceptionCacheEntry(cacheKey string) *securityExceptionCacheEntry {
+	v, _ := a.securityExceptionCacheEntries.LoadOrStore(cacheKey, &securityExceptionCacheEntry{})
+	return v.(*securityExceptionCacheEntry)
+}
+
+// invalidateSecurityExceptionCacheKey evicts cacheKey from securityExceptionListCache and bumps
+// its generation, atomically with respect to beginSecurityExceptionCacheRefresh/
+// trySetSecurityExceptionCache below: whichever of an invalidation and a racing List()'s
+// conditional Set() acquires the entry's mutex last determines the outcome, and in either order
+// the result is correct (see trySetSecurityExceptionCache's doc comment for why).
+func (a *APIServerStore) invalidateSecurityExceptionCacheKey(cacheKey string) {
+	entry := a.securityExceptionCacheEntry(cacheKey)
+	entry.mu.Lock()
+	defer entry.mu.Unlock()
+	entry.generation++
+	a.securityExceptionListCache.Delete(cacheKey)
+}
+
+// beginSecurityExceptionCacheRefresh snapshots cacheKey's current generation before a caller
+// starts a List() call for it, to later pass to trySetSecurityExceptionCache.
+func (a *APIServerStore) beginSecurityExceptionCacheRefresh(cacheKey string) (seenGeneration uint64) {
+	entry := a.securityExceptionCacheEntry(cacheKey)
+	entry.mu.Lock()
+	defer entry.mu.Unlock()
+	return entry.generation
+}
+
+// trySetSecurityExceptionCache writes value to securityExceptionListCache under cacheKey only if
+// cacheKey's generation is still seenGeneration, i.e. no invalidation has happened for it since
+// the caller's beginSecurityExceptionCacheRefresh snapshot. This closes the race #733 describes:
+// without it, a List() started before a CRD change could still write its now-stale result to the
+// cache after the informer's invalidation for that change already ran, silently undoing it. The
+// check and the write happen under the same per-key mutex invalidateSecurityExceptionCacheKey
+// uses, so there is no gap between "check the generation" and "write the cache" for a concurrent
+// invalidation to land in — unlike a bare atomic counter compared just before an unguarded Set().
+// The List() call itself is deliberately not covered by this lock: holding it across a network
+// call would let a slow List() delay an unrelated invalidation for the same key, and informers
+// typically deliver events to a handler serially, so that delay could stall processing of other,
+// unrelated CRD events too.
+func (a *APIServerStore) trySetSecurityExceptionCache(cacheKey string, seenGeneration uint64, value interface{}) {
+	entry := a.securityExceptionCacheEntry(cacheKey)
+	entry.mu.Lock()
+	defer entry.mu.Unlock()
+	if entry.generation != seenGeneration {
+		return
+	}
+	a.securityExceptionListCache.Set(cacheKey, value, securityExceptionListCacheTTL)
 }
 
 var (
@@ -256,7 +326,7 @@ func (a *APIServerStore) invalidateSecurityExceptionCacheForObject(obj interface
 		return
 	}
 	if namespace := u.GetNamespace(); namespace != "" {
-		a.securityExceptionListCache.Delete("se/" + namespace)
+		a.invalidateSecurityExceptionCacheKey("se/" + namespace)
 		return
 	}
 	a.invalidateAllSecurityExceptionCaches()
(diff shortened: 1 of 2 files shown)


bhuvan-somisetty:
@coderabbitai Addressed all 4 nitpicks in 422d4e2:

- Moved the generation snapshot inside the `securityExceptionListCache != nil` guard in both `listSecurityExceptions` and `listClusterSecurityExceptions`, so a nil-cache config no longer creates unused `securityExceptionCacheEntries` bookkeeping.
- Extracted the `"se/"` prefix into a `securityExceptionListCacheKeyPrefix` constant and updated all three call sites, including the load-bearing filter in `invalidateAllSecurityExceptionCaches`.
- Corrected the `securityExceptionCacheEntry` doc comment to state that entries are retained indefinitely and grow with every distinct namespace ever scanned, rather than claiming a stable bound.
- In both race regression tests, the goroutine now captures `GetSecurityExceptions`' error into a variable and the assertion runs on the main goroutine after `done`, instead of calling `assert.NoError` from the goroutine (which could panic on the timeout path).

`go build`/`go vet`/`go test ./repositories/...` all pass.

matthyx (approved):
Reviewed `repositories/apiserver.go` and `repositories/apiserver_test.go` in full.

The generation/mutex compare-and-swap correctly closes the TOCTOU race: `beginSecurityExceptionCacheRefresh` always snapshots the generation before the `List()` call starts, and `trySetSecurityExceptionCache` checks+writes under the same per-key mutex `invalidateSecurityExceptionCacheKey` uses, so there's no gap for a racing invalidation to land in either order. Verified `invalidateAllSecurityExceptionCaches`' `"se/"`-prefix fallback is only wired to the namespaced-kind informer handler, never the cluster-scoped one, so it can't wrongly evict the `"cse"` key.

All 4 CodeRabbit nitpicks from the earlier review are fully applied (generation snapshot moved inside the nil-cache guard in both list functions, `securityExceptionListCacheKeyPrefix` constant extracted and used consistently at every site, the entry-retention doc comment now correctly states unbounded growth, and both new regression tests assert the captured error on the main goroutine instead of inside the spawned one). The two new race regression tests use the existing `ctxCapturingResource`/`hook` test infra correctly to deterministically reproduce the pre-fix interleaving. All CI checks (build, vet, tests, lint, CodeQL) are green.

LGTM.

PR #736, opened 2026-08-16
chore(tools): drop RemoveContainerFromSlug, which nothing calls

### LINKED ISSUE #735: RemoveContainerFromSlug is unused and panics when the slug starts with the container name
`internal/tools.RemoveContainerFromSlug` has no callers:

```
$ grep -rn "RemoveContainerFromSlug" --include=*.go .
./internal/tools/tools.go:202:func RemoveContainerFromSlug(slug, container string) string {
./internal/tools/tools_test.go:160:func TestRemoveContainerFromSlug(t *testing.T) {
./internal/tools/tools_test.go:188:  assert.Equal(t, tt.want, RemoveContainerFromSlug(tt.slug, tt.container))
```

it arrived in d109fe7 ("use application profile instead of sbomp", 2024-06-04) together with its test, and that commit is the only one `git log -S RemoveContainerFromSlug --all` reports, so the count has never changed since. it was never wired to anything.

that would be fine on its own. what makes it worth deleting rather than leaving is that it is wrong in two ways, and it looks like a helper someone would reasonably reach for:

```go
func RemoveContainerFromSlug(slug, container string) string {
	i := strings.LastIndex(slug, container)
	if i == -1 {
		return slug
	}
	return slug[:i-1]
}
```

`slug[:i-1]` assumes a separator sits before the match. when the slug starts with the container name, `i` is 0 and it panics:

```
RemoveContainerFromSlug("nginx", "nginx")
panic: runtime error: slice bounds out of range [:-1]
```

and `strings.LastIndex(slug, "")` returns `len(slug)`, not -1, so an empty container falls past the guard and quietly drops the last character:

```
RemoveContainerFromSlug("nginx-abc", "")  =>  "nginx-ab"
```

the existing test has three cases, and all three put the container somewhere in the middle of the slug with a separator in front, so neither shape is covered.

deleting it and its test is the cheap fix. if a caller ever wants this, it wants the separator handled and the empty case rejected, and that is easier to write fresh than to notice is missing.

version: main (e2b0dd3)

AnshulPatil2005:
## Overview

`internal/tools.RemoveContainerFromSlug` has no callers. it arrived in d109fe7 (2024-06-04) with its test and nothing else, and that is still the only commit `git log -S RemoveContainerFromSlug --all` reports, so it has never been wired to anything.

this deletes it and its test.

## Additional Information

being unused is not on its own a reason to delete something. the reason here is that it is wrong in two ways and reads like a helper someone would reach for:

```go
return slug[:i-1]
```

assumes a separator sits before the match, so a slug that starts with the container name gives `i == 0` and panics with `slice bounds out of range [:-1]`.

`strings.LastIndex(slug, "")` returns `len(slug)` rather than -1, so an empty container walks past the `i == -1` guard and the function quietly returns the slug minus its last character.

the three existing test cases all put the container in the middle of the slug with a separator in front, so neither shape is covered and both would survive a refactor.

nothing else in `tools.go` used it, and `strings` is still needed by the rest of the file.

## How to Test

`go build ./... && go vet ./... && go test ./internal/tools/ -count=1`

`grep -rn "RemoveContainerFromSlug" --include=*.go .` returns nothing afterwards.

the behaviour claimed above is two lines to reproduce against the deleted function:

```
RemoveContainerFromSlug("nginx", "nginx")     panic: slice bounds out of range [:-1]
RemoveContainerFromSlug("nginx-abc", "")      "nginx-ab"
```

note the tree does not build on main right now for an unrelated reason (#729), so the build check above needs #730 or a base that has it.

## Related issues/PRs:

* Resolved #735

## Summary by CodeRabbit

* **Refactor**
  * Removed an obsolete slug-processing utility that is no longer used.
  * Deleted its associated automated tests.

### DIFF
--- a/internal/tools/tools.go
+++ b/internal/tools/tools.go
@@ -198,11 +198,3 @@ func appendUnique(forms []string, form string) []string {
 	}
 	return append(forms, form)
 }
-
-func RemoveContainerFromSlug(slug, container string) string {
-	i := strings.LastIndex(slug, container)
-	if i == -1 {
-		return slug
-	}
-	return slug[:i-1]
-}
--- a/internal/tools/tools_test.go
+++ b/internal/tools/tools_test.go
@@ -157,39 +157,6 @@ func TestNormalizeReference(t *testing.T) {
 	}
 }
 
-func TestRemoveContainerFromSlug(t *testing.T) {
-	tests := []struct {
-		name      string
-		slug      string
-		container string
-		want      string
-	}{
-		{
-			name:      "naked pod",
-			slug:      "pod-nn-nn-7c1c-a197",
-			container: "nn",
-			want:      "pod-nn",
-		},
-		{
-			name:      "replicaset",
-			slug:      "replicaset-nginx-bf5d5cf98-nginx-b532-f893",
-			container: "nginx",
-			want:      "replicaset-nginx-bf5d5cf98",
-		},
-		{
-			name:      "tricky",
-			slug:      "replicaset-local-path-provisioner-988d74bc-local-path-provisioner-4b6b-20c0",
-			container: "local-path-provisioner",
-			want:      "replicaset-local-path-provisioner-988d74bc",
-		},
-	}
-	for _, tt := range tests {
-		t.Run(tt.name, func(t *testing.T) {
-			assert.Equal(t, tt.want, RemoveContainerFromSlug(tt.slug, tt.container))
-		})
-	}
-}
-
 func TestReferenceMatchForms(t *testing.T) {
 	const digest = "sha256:aaaabbbbccccddddeeeeffff0000111122223333444455556666777788889999"
 	tests := []struct {


matthyx (approved):
Verified the claims in the description:

- `RemoveContainerFromSlug` has no callers anywhere in the repo (confirmed via code search on `main`).
- The two bugs are real: a slug that starts with the container name gives `i == 0` and `slug[:i-1]` panics; `strings.LastIndex(slug, "")` returns `len(slug)`, so an empty container silently drops the last character instead of hitting the `i == -1` guard.
- `strings` is still used elsewhere in `tools.go` (`SanitizeLabel`, `CleanupStaleTempDirs`), so the import stays valid after the deletion.
- Diff is a clean removal of the function and its test only, no other changes in scope.
- All CI checks (build, lint, tests, CodeQL, GitGuardian, DCO) are green, and CodeRabbit found no actionable comments.

No blockers. LGTM, approving.

PR #738, opened 2026-08-16
refactor(controllers): share the tracked-scan submit sequence

### LINKED ISSUE #737: the tracked-scan submit sequence is written out three times in http.go
`GenerateSBOM`, `ScanCVE` and `ScanRegistry` in `controllers/http.go` each submit the same twenty-two lines to the worker pool:

```go
h.workerPool.Submit(func() {
	if !h.claimTrackedJob(newScan.JobID) {
		return
	}
	start := time.Now()
	err = h.scanService.ScanCVE(bgCtx)
	outcome := "success"
	if err != nil {
		outcome = "error"
	}
	h.recordScan(bgCtx, "scanCVE", start, outcome, err)
	if err != nil {
		h.ensureStatuses().markFailed(newScan.JobID, scanFailureReason(outcome, err))
	} else {
		h.ensureStatuses().markSucceeded(newScan.JobID)
	}
	if err != nil {
		logger.L().Ctx(bgCtx).Error("service error - ScanCVE", helpers.Error(err), ...)
	}
})
```

claim the job, time the call, derive the outcome, record the metric, mark the status, log on error. the three copies differ only in the endpoint label, the service method, and which fields go on the error log line.

this is the same shape as #722 and #724: a sequence that is one thing conceptually, written out once per flow, where the flows then drift. that drift is what #469, #501, #557, #608, #609 and #623 were.

`ScanCP` is not part of it. `ErrPartialContainerProfile` is a third outcome there, recorded as `partial` for the metric while the job is still marked succeeded, and nothing equivalent exists in the other three, so folding it in would mean a flag that only one caller ever sets.

one thing worth fixing while collapsing them: all three write the scan result to the handler's own `err` with `err =`, not `err :=`. the worker goroutine does that after the handler has already returned. nothing reads `err` after `Submit` today so there is no race as it stands, but it only holds as long as nobody adds a line below the submit. scoping the error to the closure removes the question.

version: main (e2b0dd3)

AnshulPatil2005:
## Overview

`GenerateSBOM`, `ScanCVE` and `ScanRegistry` each submitted the same twenty-two lines to the worker pool: claim the job, time the call, derive the outcome, record the metric, mark the status, log on error. The three differed only in the endpoint label, which service method they called, and the fields on the error log line.

`runTrackedScan` is that sequence once. The three call sites are five lines each now.

## Additional Information

`ScanCP` keeps its own copy. `ErrPartialContainerProfile` is a third outcome there, recorded as `partial` for the metric while the job is still marked succeeded, and there is nothing equivalent in the other three, so folding it in would mean a parameter only one caller ever sets.

the error is scoped to the closure now. all three used `err =` against the handler's own `err`, which the worker goroutine writes after the handler has already returned. nothing reads it after `Submit` today, so this is not a live race, only one that would appear the moment a line is added below the submit.

the log call goes from a fixed argument list to `append([]helpers.IDetails{helpers.Error(err)}, errDetails...)`, so the error stays the first field and the per-endpoint fields follow in the order they did before. same message strings, same endpoint labels, same fields.

`markSucceeded` moved into an early-return shape rather than an `else`, which is the only structural change to the sequence itself.

## How to Test

`go build ./... && go vet ./controllers/... && go test ./controllers/ -race -count=1`

no test changes. the paths this touches are already covered end to end:

```
TestHTTPController_GenerateSBOM / ScanCVE / ScanRegistry
TestHTTPController_ScanStatus_Succeeded / _Failed / _AbandonedOnShutdown
TestHTTPController_GenerateSBOM_EmptyJobIDStillRuns
TestHTTPController_MetricsEndpoint_RecordsScanFailureReason
```

they pass unchanged, and all three endpoints now run the same code, so what was three sets of assertions about three copies is three sets about one.

that they actually constrain the helper rather than just compiling against it: replacing the recorded outcome inside `runTrackedScan` with a hardcoded success fails `TestHTTPController_MetricsEndpoint_RecordsScanFailureReason` on all three endpoints, generateSBOM, scanCVE and scanRegistryImage, plus its unclassified-error case.

`TestHTTPController_GenerateSBOM_EmptyJobIDStillRuns` is the one that pins `claimTrackedJob("")` returning true, so an untracked job still runs.

note main does not build right now for an unrelated reason (#729), so the commands above need #730 or a base that has it.

## Related issues/PRs:

* Resolved #737

### DIFF
--- a/controllers/http.go
+++ b/controllers/http.go
@@ -177,28 +177,39 @@ func (h *HTTPController) GenerateSBOM(c *gin.Context) {
 	bgCtx := context.WithoutCancel(domain.WithScanPhaseUpdater(ctx, func(phase string) {
 		h.ensureStatuses().markPhase(newScan.JobID, phase)
 	}))
+	h.runTrackedScan(bgCtx, newScan.JobID, "generateSBOM", h.scanService.GenerateSBOM,
+		"service error - GenerateSBOM",
+		helpers.String("imageSlug", newScan.ImageSlug),
+		helpers.String("imageTagNormalized", newScan.ImageTagNormalized),
+		helpers.String("imageHash", newScan.ImageHash))
+}
+
+// runTrackedScan submits scan to the worker pool and reports its result to both the scan
+// metrics and the status store. GenerateSBOM, ScanCVE and ScanRegistry all report the same
+// way, differing only in the endpoint label, the service call and the fields on the error
+// log line, which is what errMsg and errDetails carry.
+//
+// ScanCP keeps its own copy: ErrPartialContainerProfile is a third outcome there, recorded
+// as "partial" for the metric while still marking the job succeeded, and it has no
+// equivalent in the other three.
+func (h *HTTPController) runTrackedScan(bgCtx context.Context, jobID, endpoint string, scan func(context.Context) error, errMsg string, errDetails ...helpers.IDetails) {
 	h.workerPool.Submit(func() {
-		if !h.claimTrackedJob(newScan.JobID) {
+		if !h.claimTrackedJob(jobID) {
 			return
 		}
 		start := time.Now()
-		err = h.scanService.GenerateSBOM(bgCtx)
+		err := scan(bgCtx)
 		outcome := "success"
 		if err != nil {
 			outcome = "error"
 		}
-		h.recordScan(bgCtx, "generateSBOM", start, outcome, err)
+		h.recordScan(bgCtx, endpoint, start, outcome, err)
 		if err != nil {
-			h.ensureStatuses().markFailed(newScan.JobID, scanFailureReason(outcome, err))
-		} else {
-			h.ensureStatuses().markSucceeded(newScan.JobID)
+			h.ensureStatuses().markFailed(jobID, scanFailureReason(outcome, err))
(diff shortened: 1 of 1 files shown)


matthyx (approved):
Reviewed the diff and pulled the branch locally to verify.

- `runTrackedScan` is a faithful extraction: signature (`func(context.Context) error`) matches `GenerateSBOM`/`ScanCVE`/`ScanRegistry` on `ports.ScanService`, and each of the three call sites passes the correct endpoint label, service method, log message, and per-endpoint log fields in the original order.
- The `err` scoping change (closure-local `err := scan(bgCtx)` instead of writing the handler's outer `err`) is behavior-preserving — nothing reads the outer `err` after `Submit()` in any of the three handlers, both before and after this change.
- `ScanCP` is correctly left untouched, since its `partial` outcome (via `ErrPartialContainerProfile`) has no equivalent in the other three and folding it in would add a parameter only one caller needs.
- Ran locally on `bedb31b`: `go build ./...`, `go vet ./controllers/...`, and `go test ./controllers/ -race -count=1` all pass.
- Also confirmed the PR description's mutation-testing claim: hardcoding `outcome` to always be `"success"` inside `runTrackedScan` fails `TestHTTPController_MetricsEndpoint_RecordsScanFailureReason` across all three endpoints, so the shared test suite genuinely constrains the new helper rather than just compiling against it.

No blockers. LGTM.

PR #741, opened 2026-08-16
refactor(adapters): convert a grype match in one place

### LINKED ISSUE #740: the grype match conversion is written out twice and the ignored copy is untested
`grypeToDomainMatches` and `grypeToDomainIgnoredMatches` in `adapters/v1/grype_to_domain.go` each build the same `v1beta1.Match`, thirty-two lines, written out twice:

```
adapters/v1/grype_to_domain.go:47-78    inside grypeToDomainMatches
adapters/v1/grype_to_domain.go:167-199  inside grypeToDomainIgnoredMatches
```

they are identical today. diffing the two blocks gives back only the wrapper lines, `result = append(result, v1beta1.Match{` against `Match: v1beta1.Match{`.

they are converting the same thing. `models.IgnoredMatch` embeds `models.Match`:

```go
type IgnoredMatch struct {
	Match
	AppliedIgnoreRules []IgnoreRule `json:"appliedIgnoreRules"`
}
```

so every field added to the match conversion has to be added in both places by hand, and nothing says so.

the second copy is also the one that is not covered. the existing `Test_grypeToDomain` builds its ignored match as

```go
IgnoredMatches: []models.IgnoredMatch{{
	AppliedIgnoreRules: []models.IgnoreRule{{}},
}},
```

with the embedded `Match` left zero, so the thirty-two lines inside `grypeToDomainIgnoredMatches` convert nothing in that test and a difference between the two paths fails nothing. dropping a field from just the ignored copy passes the suite as it stands.

that matters more for ignored matches than it looks: they are what `ApplySecurityExceptions` writes, and what the VEX statements and the exception attribution in `repositories/apiserver.go` are built from, so a field quietly missing there shows up as wrong suppression data rather than as a missing field.

extracting the conversion into one function is enough, and `models.IgnoredMatch` embedding `models.Match` means the ignored path can call it with `m.Match` directly.

for the record, three fields grype has are dropped by this conversion, `models.Package.ID`, `models.MatchDetails.Fix` and `models.Fix.Available`. all three are dropped because `v1beta1.GrypePackage`, `v1beta1.MatchDetails` and `v1beta1.Fix` have nowhere to put them, so they are not part of this.

version: main (909b6ec)

AnshulPatil2005:
## Overview

`grypeToDomainMatches` and `grypeToDomainIgnoredMatches` each built the same `v1beta1.Match`, thirty-two identical lines twice. `models.IgnoredMatch` embeds `models.Match`, so the two are converting the same thing and a field added to one had to be added to the other by hand.

`grypeToDomainMatch` is that conversion once. The ignored path calls it with `m.Match`.

## Additional Information

the ignored copy was the untested one. the existing case sets only `AppliedIgnoreRules` and leaves the embedded `Match` zero, so those thirty-two lines converted nothing under test and a difference between the two paths failed nothing.

that copy is the one feeding suppression data. ignored matches are what `ApplySecurityExceptions` writes, and what the VEX statements and the exception attribution in `repositories/apiserver.go` read, so a field going missing there reads as wrong suppression rather than as an obviously absent field.

three grype fields are dropped by this conversion and stay dropped: `models.Package.ID`, `models.MatchDetails.Fix` and `models.Fix.Available`. `v1beta1.GrypePackage`, `v1beta1.MatchDetails` and `v1beta1.Fix` have no field for any of them, so that is a storage-side limit and not something this can fix.

no behaviour change otherwise. same fields, same order, same helpers.

## How to Test

`go build ./... && go vet ./adapters/... && go test ./adapters/... -count=1`

`Test_grypeToDomain` passes unchanged.

`Test_grypeToDomain_ignoredMatchConvertsLikeAMatch` is new. it converts one populated match through both paths in a single document and asserts `got.Matches[0]` equals `got.IgnoredMatches[0].Match`, then checks the ID, package name, fix versions and cvss are actually set, so the equality is not two zero values agreeing.

it catches what it is for. making the ignored path drop `Fix.Versions` while the regular path keeps it fails the new test and leaves `Test_grypeToDomain` passing, which is the gap this closes.

## Related issues/PRs:

* Resolved #740

### DIFF
--- a/adapters/v1/grype_to_domain.go
+++ b/adapters/v1/grype_to_domain.go
@@ -44,41 +44,48 @@ func toRawMessage(v interface{}) json.RawMessage {
 func grypeToDomainMatches(matches []models.Match) []v1beta1.Match {
 	var result []v1beta1.Match
 	for _, m := range matches {
-		result = append(result, v1beta1.Match{
-			Vulnerability: v1beta1.Vulnerability{
-				VulnerabilityMetadata: v1beta1.VulnerabilityMetadata{
-					ID:          m.Vulnerability.VulnerabilityMetadata.ID,
-					DataSource:  m.Vulnerability.VulnerabilityMetadata.DataSource,
-					Namespace:   m.Vulnerability.VulnerabilityMetadata.Namespace,
-					Severity:    m.Vulnerability.VulnerabilityMetadata.Severity,
-					URLs:        m.Vulnerability.VulnerabilityMetadata.URLs,
-					Description: m.Vulnerability.VulnerabilityMetadata.Description,
-					Cvss:        grypeToDomainMatchesCvss(m.Vulnerability.VulnerabilityMetadata.Cvss),
-				},
-				Fix: v1beta1.Fix{
-					Versions: m.Vulnerability.Fix.Versions,
-					State:    m.Vulnerability.Fix.State,
-				},
-				Advisories: grypeToDomainMatchesAdvisories(m.Vulnerability.Advisories),
-			},
-			RelatedVulnerabilities: grypeToDomainMatchesRelatedVulnerabilities(m.RelatedVulnerabilities),
-			MatchDetails:           grypeToDomainMatchesMatchDetails(m.MatchDetails),
-			Artifact: v1beta1.GrypePackage{
-				Name:         m.Artifact.Name,
-				Version:      m.Artifact.Version,
-				Type:         v1beta1.SyftType(m.Artifact.Type),
-				Locations:    grypeToDomainMatchesLocations(m.Artifact.Locations),
-				Language:     v1beta1.SyftLanguage(m.Artifact.Language),
-				Licenses:     m.Artifact.Licenses,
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed the diff directly (bot reviewers — Codex and CodeRabbit — both hit their rate limits and left no actual review).

This is a clean extract-method refactor: `grypeToDomainMatch` pulls the field-by-field `models.Match` → `v1beta1.Match` conversion out of `grypeToDomainMatches` and `grypeToDomainIgnoredMatches`. Compared the before/after field lists line by line — every field, helper call, and order is identical in both call sites, so there's no behavior change.

The new test (`Test_grypeToDomain_ignoredMatchConvertsLikeAMatch`) is a real improvement: it closes the gap where the ignored-match path's embedded `Match` conversion was previously exercised only with zero values, so a divergence between the two paths wouldn't have failed anything.

CI is green across the board (build, vet, test, lint, CodeQL, DCO). No blockers — approving.

PR #743, opened 2026-08-16
fix: bound the sbom-scanner sidecar's image pull by the request timeout

### LINKED ISSUE #742: Bug: sbom-scanner sidecar's image-pull phase has no timeout anywhere in the call chain, so a stalling registry permanently wedges a scanConcurrency worker slot
# Bug: sbom-scanner sidecar's image-pull phase has no timeout anywhere in the call chain, so a stalling registry permanently wedges a scanConcurrency worker slot

## Description

`pkg/sbomscanner/v1/server.go`'s `CreateSBOM` gRPC handler resolves the image source (registry pull) with **no deadline enforced at any layer** — client, gRPC transport, or server. A registry that accepts a TCP/TLS connection but then stalls (no error, just no timely response — a slow/half-open connection, not an outright refusal) hangs the pull indefinitely, and with it the RPC and the caller's worker slot.

This is the sidecar-only counterpart of a gap that was already found and fixed for the in-process adapter in #582/`adapters/v1/syft.go`: `CreateSBOM`'s own doc comment claims `resolveSource` "mirrors the retry chain in `adapters/v1/syft.go` so the sidecar and in-process adapters behave the same" (`pkg/sbomscanner/v1/server.go:71-72`) — but on this specific dimension they don't.

## Evidence (file:line, `upstream/main` @ `909b6ec`)

**Sidecar path — no timeout on the pull:**
- `pkg/sbomscanner/v1/server.go:160-165`:
  ```go
  src, err := resolveSource(ctx, func(_ context.Context, ref string, opts *image.RegistryOptions) (source.Source, error) {
      // Pulls intentionally run on a detached context (see #421); the request ctx is used only for the registry auth provider inside resolveSource.
      return tools.RetryWithBackoff(context.Background(), "source_resolution", tools.Default429RetryConfig(), tools.IsRateLimitError, func(retryCtx context.Context) (source.Source, error) {
          ctxWithSize := context.WithValue(retryCtx, image.MaxImageSize, req.MaxImageSize)
          return syft.GetSource(ctxWithSize, ref, ...)
      })
  }, imageID, imageTag, registryOptions)
  ```
  The actual pull (`syft.GetSource`) runs on `context.Background()` — fully detached, no deadline, not even the request's own `ctx`.
- `pkg/sbomscanner/v1/server.go:226-230`: `timeout := time.Duration(req.TimeoutSeconds) * time.Second; dl := deadline.New(timeout)` is the *only* deadline `CreateSBOM` establishes, and it wraps only the SBOM-generation step (`dl.Run(...)`), which starts strictly *after* `resolveSource` (line 160) has already returned. It provides zero protection for the pull phase — a hung pull never reaches this code at all.

**No backstop upstream of the server either:**
- `core/services/scan.go:1298`: `workerCtx := context.WithoutCancel(ctx)` — `context.WithoutCancel` also strips any deadline the parent had (`Deadline()` returns `ok == false` on the result), so even if an upstream HTTP/scan context had a deadline, it's gone by the time `sbomCreator.CreateSBOM(workerCtx, ...)` is called at line 1325.
- `adapters/v1/sidecar.go:141`: `s.client.CreateSBOM(ctx, req)` passes that same deadline-free `ctx` straight through.
- `pkg/sbomscanner/v1/client.go:141`: `c.client.CreateSBOM(ctx, pbReq)` — the gRPC client call itself adds no `context.WithTimeout`/`grpc.WithTimeout`. `pkg/sbomscanner/v1/client.go:104-115`'s `codes.DeadlineExceeded` handling in the error switch is dead code for this path: nothing ever arms a deadline that could produce it, for the pull phase.
- `NewSBOMScannerClient` (`pkg/sbomscanner/v1/client.go`) sets `MaxCallRecvMsgSize`/`MaxCallSendMsgSize` as default call options but no keepalive/timeout dial options that would bound a stalled unary call either.

**Confirmed asymmetry with the sibling in-process adapter (`adapters/v1/syft.go`):**
```go
// adapters/v1/syft.go:195-238
ctxWithTimeout := ctx
if s.scanTimeout > 0 {
    ctxWithTimeout, cancel = context.WithTimeout(ctx, s.scanTimeout)
    defer cancel()
}
ctxWithSize := context.WithValue(ctxWithTimeout, image.MaxImageSize, s.maxImageSize)
...
src, err := registryauth.ResolveSource(ctxWithTimeout, metrics.ComponentInProcess,
    func(_ context.Context, ref string, opts *image.RegistryOptions) (source.Source, error) {
        return tools.RetryWithBackoff(ctxWithSize, "source_resolution", ..., func(retryCtx context.Context) (source.Source, error) {
            return syft.GetSource(retryCtx, ref, syftGetSourceConfig(opts, imgPlatform))
        })
    }, ...)
```
Here the pull genuinely is bounded by `s.scanTimeout` via `ctxWithTimeout`. The sidecar server never received the equivalent of this.

## Root cause, and why this isn't the already-accepted tradeoff

`CreateSBOM`'s doc comment (`pkg/sbomscanner/v1/server.go:93-98`) already documents an *accepted* tradeoff: Syft's cataloguers don't support cancellation, so a cataloguing timeout still leaves an abandoned goroutine running in the background — but the RPC itself still returns `Incomplete` to the caller once `dl.Run` times out, freeing the worker slot. **The pull phase has no such protection at all**: it isn't wrapped in anything like `dl.Run`, so a stalled pull doesn't time out, doesn't return `Incomplete`, and doesn't free the caller's slot — the RPC simply never returns.

The `// Pulls intentionally run on a detached context (see #421)` comment is stale/misleading: `#421` ("every `APIServerStore` CRUD call passes `context.Background()`...") was fixed by PR #422, which touched only `repositories/apiserver.go` and `repositories/apiserver_test.go` (verified via `gh pr view 422 --json files`) — it never touched `pkg/sbomscanner/v1/server.go`. The comment was added later, in PR #424, whose own description explicitly says *"The surrounding `context.Background()` calls in this file are out of scope here — tracked separately in #421."* So the sidecar's pull-phase detachment was flagged as tracked-for-later, that tracking issue closed without covering this file, and the file's comment now reads as if it's a settled design decision. It isn't: the sibling adapter's fix (bounding the pull by `scanTimeout`) is direct proof this class of gap is fixable and was already judged worth fixing here — just never ported to the sidecar.

## Impact

Any deployment running the sidecar scanner (the mode this session's own recent commits — metrics port collision fix, graceful-shutdown timeout, periodic temp-dir sweep — show is actively being hardened) is exposed: a single slow or partially-unresponsive registry (network partition mid-handshake, an overloaded/misconfigured mirror, a proxy that accepts connections but never completes a response) can wedge one `scanConcurrency` slot per affected image indefinitely, with no timeout, no error, and no automatic recovery — only process-level signals like `kubevuln_worker_pool_queue_depth` climbing are visible. Enough stuck pulls from one registry can starve scanning for the whole cluster, including workloads pulling from unrelated, healthy registries.

## Affected components

- `pkg/sbomscanner/v1/server.go` (`CreateSBOM`, `resolveSource`)
- `pkg/sbomscanner/v1/client.go` (`CreateSBOM` — needs the timeout to actually reach the server as an enforceable deadline)
- `core/services/scan.go` (`getOrCreateSBOM` — `workerCtx` strips any inherited deadline; worth confirming the fix doesn't need to route around this too, since the server-side fix is self-sufficient regardless)

## Acceptance criteria

- The pull phase (`resolveSource`'s call chain in `CreateSBOM`) is bounded by `req.TimeoutSeconds` (falling back to the same 5-minute default used for generation), consistent with how `adapters/v1/syft.go` bounds its pull with `scanTimeout`.
- A stalled/never-returning `sourceGetter` causes `CreateSBOM` to return (with a status that distinguishes "pull timed out" similarly to how `Incomplete` is used for a cataloguing timeout) instead of hanging the RPC indefinitely.
- A new regression test in `pkg/sbomscanner/v1/server_test.go` — analogous to `TestCreateSBOM_TimeoutDoesNotRaceWithAbandonedSyft` but for the pull phase — uses a `sourceGetter` that blocks past the configured timeout and asserts the RPC returns rather than hanging. (`TestCreateSBOM_ContextCancelled` doesn't cover this: it passes an already-canceled context and never exercises a slow-but-not-canceled pull.)
- Existing tests (`TestResolveSource*`, `TestCreateSBOM_*`) continue to pass unchanged.
- `go build ./pkg/sbomscanner/...`, `go vet ./pkg/sbomscanner/...`, `go test ./pkg/sbomscanner/...` pass.

## Suggested fix direction

Port the `adapters/v1/syft.go:195-198` pattern into `pkg/sbomscanner/v1/server.go`: derive a `ctxWithTimeout` from the incoming `ctx`, bounded by `time.Duration(req.TimeoutSeconds) * time.Second` (defaulting the same way the existing `timeout`/`dl` variables already do), and thread it into `resolveSource`'s closure and its `tools.RetryWithBackoff` call in place of the bare `context.Background()`. Update the stale `// Pulls intentionally run on a detached context (see #421)` comment to reflect the new behavior and drop the misleading reference. Keeping the pull on a context derived from the *request* timeout rather than the (still-detached) caller `ctx` preserves the existing "caller cancellation doesn't abort an in-flight scan" behavior while finally giving the pull phase the ceiling the generation phase already has.

bhuvan-somisetty:
## Problem

`pkg/sbomscanner/v1/server.go`'s `CreateSBOM` gRPC handler resolves the image source (registry pull) on `context.Background()` - fully detached, with no deadline at any layer (client, gRPC transport, or server). A registry that accepts a connection but stalls on the manifest/blob response (not an outright error - a slow or half-open connection) hangs the pull, and the RPC, indefinitely. The only timeout the handler establishes (`deadline.New(timeout)`) wraps SBOM *generation*, which starts strictly after the pull already returned, so it gives the pull no protection at all.

## Root cause

`adapters/v1/syft.go` (the in-process adapter) already bounds its own pull by `scanTimeout` via a `context.WithTimeout(ctx, s.scanTimeout)`. `CreateSBOM`'s doc comment claims `resolveSource` "mirrors the retry chain in `adapters/v1/syft.go` so the sidecar and in-process adapters behave the same" - but on this dimension it didn't.

The `// Pulls intentionally run on a detached context (see #421)` comment is stale: it traces to PR #424, whose description says the file's `context.Background()` calls were "tracked separately in #421" - #421 was fixed by PR #422, which touched only `repositories/apiserver.go` and never this file. The comment reads like a settled design decision; it isn't one that was ever acted on here.

## Fix

Compute the request's timeout once (same fallback-to-5-minutes logic as before), derive a `context.WithTimeout(ctx, timeout)` from it, and thread that through `resolveSource` and its `RetryWithBackoff` call in place of `context.Background()`. This mirrors `adapters/v1/syft.go`'s pattern of independent timeout-bounded windows for the pull and generation phases (not one budget split across both).

A pull that times out now surfaces as `Status: Incomplete` - the same status a cataloguing timeout already uses - instead of hanging the RPC. The `errors.Is(err, context.DeadlineExceeded) || strings.Contains(err.Error(), "context deadline exceeded")` check mirrors the identical one in `adapters/v1/syft.go`, needed because stereoscope/oci-registry wraps deadline errors in a type that doesn't implement `Unwrap()`.

## Testing

- New regression test `TestCreateSBOM_PullTimeout_ReturnsIncompleteInsteadOfHanging`: a local mock registry accepts the connection but never responds to the manifest request (the exact "stalling registry" scenario), and the test asserts `CreateSBOM` returns `Status: Incomplete` within the configured timeout rather than hanging.
- **Verified to fail against the pre-fix code**: reverted the production change locally, reran the new test, and confirmed it fails after 10s with "the pull is not bounded by any deadline"; restored the fix and reran to confirm it passes in ~1s.
- `go build ./...` - clean.
- `go vet ./pkg/sbomscanner/...` - clean.
- `go test ./pkg/sbomscanner/... -count=1` - all tests pass, including the existing `TestResolveSource*`, `TestCreateSBOM_*`, and `TestHealth` suites, unchanged.
- Could not run `go test -race` locally (no cgo/gcc in this environment, same limitation noted in prior PRs in this repo).

## Impact

Any deployment running the sidecar scanner is exposed today: a single slow or partially-unresponsive registry can wedge one `scanConcurrency` slot per affected image indefinitely, with no timeout, no error, and no automatic recovery. Enough stuck pulls from one registry can starve scanning cluster-wide, including workloads pulling from unrelated, healthy registries. This fix gives the pull phase the same timeout ceiling the generation phase already has.

Fixes #742

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scan timeout handling during image retrieval and SBOM generation.
  * Scans that exceed the configured timeout now complete promptly with an **Incomplete** status instead of hanging or returning an error.
  * Registry retries and source resolution now respect the scan timeout.

### DIFF
--- a/pkg/sbomscanner/v1/server.go
+++ b/pkg/sbomscanner/v1/server.go
@@ -146,11 +146,24 @@ func (s *scannerServer) CreateSBOM(ctx context.Context, req *pb.CreateSBOMReques
 		Credentials:           credentials,
 	}
 
-	// Download image from registry
+	// timeout bounds both the pull below and SBOM generation further down, as two independent
+	// windows (not one budget split across both) - matching adapters/v1/syft.go's scanTimeout,
+	// which bounds its own pull and cataloging phases the same way.
+	timeout := time.Duration(req.TimeoutSeconds) * time.Second
+	if timeout <= 0 {
+		timeout = 5 * time.Minute
+	}
+
+	// Download image from registry, bounded by timeout: a registry that accepts a connection
+	// but then stalls (no error, just no timely response) used to hang this call - and the
+	// caller's scanConcurrency slot with it - indefinitely, since neither the request ctx nor
+	// any deadline reached this pull; the only deadline in this method wrapped generation,
+	// which starts after the pull already returned. See #742.
 	logger.L().Debug("downloading image", helpers.String("imageID", imageID))
-	src, err := resolveSource(ctx, func(_ context.Context, ref string, opts *image.RegistryOptions) (source.Source, error) {
-		// Pulls intentionally run on a detached context (see #421); the request ctx is used only for the registry auth provider inside resolveSource.
-		return tools.RetryWithBackoff(context.Background(), "source_resolution", tools.Default429RetryConfig(), tools.IsRateLimitError, func(retryCtx context.Context) (source.Source, error) {
+	ctxWithTimeout, cancelPull := context.WithTimeout(ctx, timeout)
+	defer cancelPull()
+	src, err := resolveSource(ctxWithTimeout, func(_ context.Context, ref string, opts *image.RegistryOptions) (source.Source, error) {
+		return tools.RetryWithBackoff(ctxWithTimeout, "source_resolution", tools.Default429RetryConfig(), tools.IsRateLimitError, func(retryCtx context.Context) (source.Source, error) {
 			ctxWithSize := context.WithValue(retryCtx, image.MaxImageSize, req.MaxImageSize) //nolint:staticcheck // stereoscope requires string context key
 			return syft.GetSource(ctxWithSize, ref,
 				syft.DefaultGetSourceConfig().WithRegistryOptions(opts).WithPlatform(imgPlatform).WithSources("registry"))
@@ -158,6 +171,15 @@ func (s *scannerServer) CreateSBOM(ctx context.Context, req *pb.CreateSBOMReques
 	}, imageID, imageTag, registryOptions)
 
 	switch {
+	// stereoscope/oci-registry wraps deadline errors in a custom error type that does not
+	// implement Unwrap(), so the strings.Contains check is required alongside errors.Is (same
+	// as adapters/v1/syft.go's identical check on its own pull timeout).
+	case err != nil && (errors.Is(err, context.DeadlineExceeded) || strings.Contains(err.Error(), "context deadline exceeded")):
+		metrics.RecordScanFallback(ctx, metrics.ComponentSidecar, metrics.FallbackCategorySizeClassification, metrics.FallbackStrategyIncomplete, metrics.FallbackOutcomeClassified)
+		logger.L().Warning("image pull timed out", helpers.String("imageID", imageID))
+		return &pb.CreateSBOMResponse{
+			Status: helpersv1.Incomplete,
+		}, nil
 	case err != nil && (errors.Is(err, image.ErrImageTooLarge) || strings.Contains(err.Error(), image.ErrImageTooLarge.Error())):
 		metrics.RecordScanFallback(ctx, metrics.ComponentSidecar, metrics.FallbackCategorySizeClassification, metrics.FallbackStrategyImageTooLarge, metrics.FallbackOutcomeClassified)
 		logger.L().Warning("Image exceeds size limit",
@@ -212,12 +234,9 @@ func (s *scannerServer) CreateSBOM(ctx context.Context, req *pb.CreateSBOMReques
 		resolvedPlatform = formatResolvedPlatform(meta.OS, meta.Architecture, meta.Variant)
 	}
 
-	// Generate SBOM with timeout
+	// Generate SBOM with timeout (reuses the same timeout duration computed above for the
+	// pull, as its own independent window).
 	var syftSBOM *sbom.SBOM
-	timeout := time.Duration(req.TimeoutSeconds) * time.Second
-	if timeout <= 0 {
-		timeout = 5 * time.Minute
-	}
 	dl := deadline.New(timeout)
 	// Buffered so the abandoned goroutine below can always publish and exit, even when
 	// nobody is left to receive.
--- a/pkg/sbomscanner/v1/server_test.go
+++ b/pkg/sbomscanner/v1/server_test.go
@@ -693,6 +693,64 @@ func TestCreateSBOM_PlatformMismatch_LocalRegistry(t *testing.T) {
 	assert.Empty(t, resp.Status)
 }
 
+// TestCreateSBOM_PullTimeout_ReturnsIncompleteInsteadOfHanging is a regression test for #742: a
+// registry that accepts the connection but never responds to the manifest request used to hang
+// the pull - and the RPC with it - indefinitely, since neither the request context nor any
+// deadline bounded resolveSource's call chain. The mock registry below blocks on the manifest
+// request until the test explicitly releases it, simulating exactly that stalling registry.
+// CreateSBOM must still return, bounded by TimeoutSeconds, well before that release happens.
+func TestCreateSBOM_PullTimeout_ReturnsIncompleteInsteadOfHanging(t *testing.T) {
+	release := make(chan struct{})
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.URL.Path == "/v2/" {
+			w.WriteHeader(http.StatusOK)
+			return
+		}
+		// Never respond to the manifest request until the test releases it - a connection
+		// that accepts the request but stalls, not one that errors or refuses outright.
+		<-release
+		w.WriteHeader(http.StatusNotFound)
+	}))
+	// close(release) must run before server.Close(), which blocks until the handler above
+	// returns - deferred in this order so it does, since defers run LIFO.
+	defer server.Close()
+	defer close(release)
+
+	u, err := url.Parse(server.URL)
+	require.NoError(t, err)
+
+	client, cleanup := startTestServer(t)
+	defer cleanup()
+
+	type result struct {
+		resp *pb.CreateSBOMResponse
+		err  error
+	}
+	done := make(chan result, 1)
+	go func() {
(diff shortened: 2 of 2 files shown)


bhuvan-somisetty:
@coderabbitai Fixed in 5b0c266: the test now records the start time and asserts the response returns within 4s of the 1s configured `TimeoutSeconds`, keeping the original 10s branch only as a deadlock fallback rather than the actual bound.

`go test ./pkg/sbomscanner/... -count=1` passes (new test at 1.01s, well under the 4s bound).

matthyx (approved):
Reviewed the diff against `adapters/v1/syft.go`'s existing `CreateSBOM` (the in-process adapter this sidecar mirrors) and I don't see any blockers.

What checks out:
- The root cause is accurate: the stale `// Pulls intentionally run on a detached context (see #421)` comment traced to #424, whose "#421" pointer was fixed by #422 in a different file (`repositories/apiserver.go`) — it never covered this pull path.
- `context.WithTimeout(ctx, timeout)` + `defer cancelPull()` for the pull, and a separate `deadline.New(timeout)` for generation, is exactly the same independent-window pattern `adapters/v1/syft.go` already uses (`ctxWithTimeout`/`s.scanTimeout` there vs. `timeout` here) — not a regression risk, a consistency fix.
- The new `errors.Is(err, context.DeadlineExceeded) || strings.Contains(err.Error(), "context deadline exceeded")` check is copied verbatim from the in-process adapter's identical workaround for stereoscope/oci-registry's non-`Unwrap()`-able deadline error type, so it'll actually catch what it's meant to catch.
- `cancelPull` is deferred (no context leak), and it only bounds the pull — generation still runs on `context.Background()` per the existing, deliberate tradeoff documented on `CreateSBOM` (Syft's catalogers don't observe cancellation).
- The regression test reproduces the real failure mode (a registry that accepts the connection but stalls on the manifest response, not an outright error) and is asserted to fail pre-fix (10s) and pass post-fix (~1s). The one actionable CodeRabbit comment (loosen/clarify the timing assertion) was already addressed in 5b0c266 with an explicit start-time-based 4s bound plus a 10s deadlock-only fallback.
- `go build`/`go vet`/`go test` all reported clean; CI (Dependency Quality, Security Analysis, License Compliance) is green.

Nice catch, and a clean, well-scoped fix. Approving.

PR #745, opened 2026-08-16
fix(repositories): report the summary's own name when storing it fails

### LINKED ISSUE #744: StoreCVESummary logs the CVE manifest's name, not the summary it stored
`StoreCVESummary` stores a `VulnerabilityManifestSummary` under a name derived from the workload, not under the name of the CVE manifest it was built from:

```go
summaryK8sResourceName, err := GetCVESummaryK8sResourceNameWithCVEName(ctx, cve.Name)
...
manifest := v1beta1.VulnerabilityManifestSummary{
		Name: summaryK8sResourceName,
```

and `GetCVESummaryK8sResourceNameWithCVEName` returns `kind-name-container` built from the wlid, falling back to `cveName` only when kind, name and container name are all empty:

```go
if kind == "" && name == "" && contName == "" && cveName != "" {
	return cveName, nil
}
```

so on any scan that has a workload, `cve.Name` and the stored object's name are two different strings.

four log lines report a name in that function. three of them report the wrong one:

```
repositories/apiserver.go:841   failed to update CVE summary manifest in storage   cve.Name
repositories/apiserver.go:846   updated CVE summary manifest in storage            cve.Name
repositories/apiserver.go:850   failed to store CVE summary manifest in storage    cve.Name
repositories/apiserver.go:855   stored CVE summary manifest in storage             manifest.Name
```

the `Get` inside the retry already uses `manifest.Name`, so storage is doing the right thing. it is only the reporting that is wrong, and only on the three lines that matter most: when a summary fails to store, the name in the log identifies nothing that exists, so looking for that object finds nothing. the one line that names it correctly is the success line, where you least need it.

that the same switch does it both ways is the tell. there is one name here and four places that each had to remember which variable held it.

#725 pulled this sequence out into `createOrUpdate` for `StoreSBOM` and left the other four call sites for a follow-up, saying they wanted a couple more parameters first. `StoreCVE` and `StoreCVESummary` want exactly one, an extra field for their log lines, both carrying `relevant=`. folding those two in passes the name once and removes the chance of them disagreeing.

`StoreCVESummaryStub` and `StoreVEX` still do not fit and are not part of this: the stub reads the full object rather than metadata only so it can check `summaryHasVulnerabilityData` and decline the update, and `StoreVEX` reads before it creates and re-reads with full options inside its retry.

version: main (6450ca8)

AnshulPatil2005:
## Overview

`StoreCVESummary` writes a `VulnerabilityManifestSummary` named for the workload, `kind-name-container`, not for the CVE manifest it was built from. Those are different strings on any scan that has a workload. Three of its four log lines reported `cve.Name`, which names nothing in storage, and only the success line reported the object that was actually written.

This folds `StoreCVE` and `StoreCVESummary` into `createOrUpdate`, the follow-up #725 left open. The name is passed once there, and is what both the `Get` inside the retry and every log line use, so the two cannot drift apart again.

## Additional Information

storage was never wrong. the `Get` inside the retry already used `manifest.Name`, so the right object was read and written the whole time. what was wrong was the reporting, on the three lines where it matters: a failed summary store logged a name that resolves to nothing, and the only line that named the object correctly was the success one.

`createOrUpdate` gains a variadic `extra ...helpers.IDetails` appended to each of its four log lines. that is the "couple more parameters" #725 said these callers needed, and it turned out to be one: both carry `relevant=`.

`StoreCVESummaryStub` and `StoreVEX` are still not folded in and this does not touch them. the stub reads the full object rather than metadata only, because it needs the spec to check `summaryHasVulnerabilityData` before declining to overwrite a real summary with a stub, and `StoreVEX` reads before creating and re-reads with full options inside its own retry. both want more than an extra log field.

## How to Test

`go build ./... && go vet ./... && go test ./repositories/ -count=1`

the existing tests cover both folded callers and pass unchanged, including the create path, the read-modify-write behind `AlreadyExists`, a failing read inside the retry, the retry giving up, and context propagation:

```
TestAPIServerStore_StoreCVESummary_transientError
TestAPIServerStore_StoreCVESummary_updateGetFailure_transientError
TestAPIServerStore_StoreCVESummary_retryExhausted_transientError
TestAPIServerStore_StoreCVESummary_ctxPropagated
TestAPIServerStore_StoreCVESummary_RetryConflict_ctxPropagated
```

they hold the name too, rather than just the sequence: passing `cve.Name` to `createOrUpdate` in place of `manifest.Name` fails `..._retryExhausted_transientError`, `..._RetryConflict_ctxPropagated` and `..._NilContentDoesNotPanic`.

`TestAPIServerStore_StoreCVESummary_reportsTheSummaryName` is new and states the premise directly: for a real wlid it asserts `cve.Name` and the derived summary name differ, that the stored object exists under the summary name, and that nothing was written under the CVE manifest name.

the log text itself is not asserted, since that would mean capturing the logger's output. the change that keeps those lines right is structural, one name reaching all four of them.

## Related issues/PRs:

* Resolved #744

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected CVE summary storage to use the workload-derived summary name.
  * Prevented duplicate or incorrectly named summary resources from being created.

* **Tests**
  * Added regression coverage to verify correct summary naming and storage behavior.

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -572,47 +572,14 @@ func (a *APIServerStore) StoreCVE(ctx context.Context, cve domain.CVEManifest, w
 	if cve.Content != nil {
 		manifest.Spec.Payload = *cve.Content
 	}
-	_, err := a.StorageClient.VulnerabilityManifests(a.Namespace).Create(ctx, &manifest, metav1.CreateOptions{})
-	switch {
-	case errors.IsAlreadyExists(err):
-		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-			if err := ctx.Err(); err != nil {
-				return err
-			}
-			// retrieve the latest version before attempting update
-			// RetryOnConflict uses exponential backoff to avoid exhausting the apiserver
-			result, getErr := a.StorageClient.VulnerabilityManifests(a.Namespace).Get(ctx, cve.Name, metav1.GetOptions{ResourceVersion: "metadata"})
-			if getErr != nil {
-				return getErr
-			}
-			// update the vulnerability manifest
-			result.Annotations = mergeMaps(result.Annotations, manifest.Annotations)
-			result.Labels = mergeMaps(result.Labels, manifest.Labels)
-			result.Spec = manifest.Spec
-			// try to send the updated vulnerability manifest
-			_, updateErr := a.StorageClient.VulnerabilityManifests(a.Namespace).Update(ctx, result, metav1.UpdateOptions{})
-			return updateErr
-		})
-		if retryErr != nil {
-			logger.L().Debug("failed to update CVE manifest in storage", helpers.Error(retryErr),
-				helpers.String("name", cve.Name),
-				helpers.String("relevant", strconv.FormatBool(withRelevancy)))
-			return fmt.Errorf("failed to update CVE manifest in storage: %w", retryErr)
-		}
-		logger.L().Debug("updated CVE manifest in storage",
-			helpers.String("name", cve.Name),
-			helpers.String("relevant", strconv.FormatBool(withRelevancy)))
-	case err != nil:
-		logger.L().Debug("failed to store CVE manifest in storage", helpers.Error(err),
-			helpers.String("name", cve.Name),
-			helpers.String("relevant", strconv.FormatBool(withRelevancy)))
-		return fmt.Errorf("failed to store CVE manifest in storage: %w", err)
-	default:
-		logger.L().Debug("stored CVE manifest in storage",
-			helpers.String("name", cve.Name),
-			helpers.String("relevant", strconv.FormatBool(withRelevancy)))
-	}
-	return nil
+	return createOrUpdate(ctx, a.StorageClient.VulnerabilityManifests(a.Namespace),
+		"CVE manifest", cve.Name, &manifest,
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed the diff and checked out `f74133`.

- `createOrUpdate` now takes the object's own `name` (already used by the retry's `Get`) plus a variadic `extra ...helpers.IDetails`, and both `StoreCVE`/`StoreCVESummary` route through it passing `relevant=`. Confirmed no behavioral difference for `StoreCVE` (its own name was already `cve.Name`) and a real fix for `StoreCVESummary`, whose four log lines previously reported `cve.Name` instead of the actual stored `manifest.Name` on 3 of 4 paths.
- `go build ./...`, `go vet ./...`, `gofmt -l`, and `go test ./repositories/... -count=1` all pass locally.
- Mutation-checked the regression claim: swapping `manifest.Name` back to `cve.Name` in `StoreCVESummary`'s call reproduces the failures the PR description lists (`..._retryExhausted_transientError`, `..._RetryConflict_ctxPropagated`, `..._NilContentDoesNotPanic`), so the new test suite does pin the fix.
- CodeRabbit's nitpick (assert the logged `name` detail directly rather than only storage side-effects) is a reasonable follow-up but not a blocker — the storage-based assertions already prove the two names can't drift, which was the actual bug.

No other issues found. LGTM.

PR #747, opened 2026-08-16
fix(adapters): restore cloud-sourced suppressions again

### LINKED ISSUE #746: cloud-sourced suppressions can never be lifted on the cached path
a finding suppressed by a backend-delivered exception policy can never be un-suppressed on the cached path. once the policy has hidden it, deleting the policy leaves it hidden.

#701 narrowed `isExceptionSourcedIgnore` so a VEX-suppressed match Grype produced would stop being restored, which was right. what it now accepts is a known CRD sourceKind and nothing else:

```go
for _, r := range im.AppliedIgnoreRules {
	if r.Package != nil {
		return false
	}
	if r.SourceKind == "SecurityException" || r.SourceKind == "ClusterSecurityException" {
		continue
	}
	return false
}
```

backend-delivered policies do not have one. they reach `ApplySecurityExceptions` like any other policy but never go through `buildPolicy`, so they carry no `sourceKind` attribute and `buildIgnoreRules` leaves `SourceKind` empty. both sides of the tree already say so:

```
adapters/v1/securityexception.go:324   A policy with no sourceKind attribute (e.g. a
                                       cloud-sourced exception, not a CRD)
repositories/apiserver.go:1087         they carry no sourceKind and their ignore rule
                                       carries no SourceKind either
```

so `RestoreSuppressedMatches` now preserves those instead of restoring them, and that is where it bites:

```go
prevIgnored := v1.IgnoredMatchKeys(cve.Content)
cve.Content = v1.RestoreSuppressedMatches(cve.Content)
filteredCve, exceptionsComplete = s.applyExceptionsToManifest(ctx, cve)
```

`reconcileCachedCVE` restores first precisely so the current exception set decides the outcome. a match that is not restored is not in `Matches` for the re-apply to look at, so it stays ignored no matter what the exception set now says.

to see it, apply a policy with no `sourceKind` and then restore:

```
stored rule: {Vulnerability:CVE-2021-44228 FixState: Package:nil SourceKind:
             SourceName:some-cloud-rule-id SourceNamespace:
             Justification:vulnerable_code_not_present ImpactStatement:}
```

`SourceKind` is empty, so it is dropped, even though `SourceName` and `Justification` say plainly where it came from.

that is the distinction worth keying on. `buildIgnoreRules` stamps whatever provenance the policy carried onto the rule, and `grypeToDomainIgnoredMatchesAppliedIgnoreRules` copies only `Vulnerability`, `FixState` and `Package` off a Grype rule:

```go
result = append(result, v1beta1.IgnoreRule{
	Vulnerability: r.Vulnerability,
	FixState:      r.FixState,
	Package:       (*v1beta1.IgnoreRulePackage)(r.Package),
})
```

so `SourceName`, `SourceNamespace`, `Justification` and `ImpactStatement` cannot come from Grype, and a rule holding any of them is ours. `FixState` is not usable for this, since Grype does set it in its own fix-state vocabulary.

that keeps #701 intact: a bare rule, which is the shape a VEX-suppressed match arrives in, still has none of the four and is still preserved.

version: main (6450ca8)

AnshulPatil2005:
## Overview

A finding suppressed by a backend-delivered exception policy could never be un-suppressed on the cached path: once the policy had hidden it, deleting the policy left it hidden.

#701 narrowed `isExceptionSourcedIgnore` to a known CRD `sourceKind`, correctly, so a VEX-suppressed match Grype produced would stop being restored. Backend-delivered policies went with it. They never go through `buildPolicy`, so they carry no `sourceKind` and `buildIgnoreRules` leaves `SourceKind` empty on the rule it writes.

A rule with no `SourceKind` but some other exception provenance is ours again.

## Additional Information

the cached path is where it mattered. `reconcileCachedCVE` restores before it re-applies the current exception set, so the set decides the outcome. a match that is not restored is not in `Matches` for the re-apply to look at, and stays ignored whatever the set now says.

the four fields keyed on are `SourceName`, `SourceNamespace`, `Justification` and `ImpactStatement`. that is safe rather than arbitrary: `grypeToDomainIgnoredMatchesAppliedIgnoreRules` copies only `Vulnerability`, `FixState` and `Package` off a Grype rule, so nothing Grype produces can set any of the four, while `buildIgnoreRules` stamps whichever the policy carried.

`FixState` is deliberately not one of them. Grype does set it, in its own fix-state vocabulary, so a rule carrying only that is not evidence of us. It stays preserved.

#701's case is untouched. a bare rule, which is the shape a VEX-suppressed match arrives in, carries none of the four and is still preserved.

a policy whose only attribute is `status` would produce a rule with just `Vulnerability` and `FixState`, which this cannot tell apart from a native Grype rule. it is preserved, i.e. it errs the way #701 wanted.

## How to Test

`go build ./... && go vet ./... && go test ./adapters/... ./core/... -count=1`

`TestRestoreSuppressedMatches_ExternalVEXFalsePositive` and `TestRestoreSuppressedMatches_GenuineCRDException`, both from #701, pass unchanged. That is the main thing: this widens what counts as ours without giving back the VEX false positive.

`TestRestoreSuppressedMatches_RestoresCloudSourcedSuppressions` is new and goes through the real path rather than hand-building a rule: it applies a policy with no `sourceKind`, asserts the stored rule's `SourceKind` is empty so the premise is on the record, then restores and expects the match back in `Matches`. Reverting the fix fails it.

`TestRestoreSuppressedMatches_LeavesGrypeSourcedIgnoresAlone` holds the other side, for both shapes Grype can produce: a bare rule, and one carrying only `FixState`.

## Related issues/PRs:

* Resolved #746

## Summary by CodeRabbit

* **Bug Fixes**
  * Cloud-sourced security suppressions are now correctly restored even when their source type is unavailable.
  * Grype and VEX-sourced ignores continue to remain suppressed as expected.

* **Tests**
  * Added regression coverage for restoring cloud suppressions and preserving Grype/VEX ignores.

### DIFF
--- a/adapters/v1/securityexception.go
+++ b/adapters/v1/securityexception.go
@@ -500,7 +500,7 @@ func emitSuppressionEvent(recorder record.EventRecorder, p armotypes.Vulnerabili
 // so the shared stored object is never modified.
 //
 // Only ignored matches carrying the signature ApplySecurityExceptions writes (a single ignore
-// rule with CRD provenance tags like SourceKind or Justification set) are restored. Any other
+// rule carrying CRD or backend exception provenance) are restored. Any other
 // entry (e.g. one produced by GrypeAdapter wiring up IgnoreRules/VexProcessor, which drops CRD
 // tags) is preserved, so the helper is self-guarding instead of relying on an assumption.
 // Reconstruction is lossless because IgnoredMatch embeds the full Match.
@@ -526,7 +526,8 @@ func RestoreSuppressedMatches(doc *v1beta1.GrypeDocument) *v1beta1.GrypeDocument
 
 // isExceptionSourcedIgnore reports whether an ignored match carries the AppliedIgnoreRules
 // signature ApplySecurityExceptions writes: non-empty AppliedIgnoreRules where every rule has no
-// package set, and exception provenance (source/justification/impact).
+// package set, and either a known CRD sourceKind or some other exception provenance, which is
+// what a backend-delivered policy leaves behind.
 func isExceptionSourcedIgnore(im v1beta1.IgnoredMatch) bool {
 	if len(im.AppliedIgnoreRules) == 0 {
 		return false
@@ -536,6 +537,17 @@ func isExceptionSourcedIgnore(im v1beta1.IgnoredMatch) bool {
 			return false
 		}
 		if r.SourceKind == "SecurityException" || r.SourceKind == "ClusterSecurityException" {
+			continue
+		}
+		// A policy that never went through buildPolicy carries no sourceKind, which today means
+		// one delivered by the backend rather than a CRD. buildIgnoreRules still stamps whatever
+		// provenance that policy did carry onto the rule, so a rule holding any of it is ours.
+		//
+		// These four are safe to key on because grypeToDomainIgnoredMatchesAppliedIgnoreRules
+		// copies only Vulnerability, FixState and Package off a Grype rule, so nothing Grype
+		// produces can set them. FixState is deliberately not among them: Grype does set it, in
+		// its own fix-state vocabulary, so a rule carrying only that is not evidence of us.
+		if r.SourceName != "" || r.SourceNamespace != "" || r.Justification != "" || r.ImpactStatement != "" {
 			continue
 		}
 		return false
--- a/adapters/v1/securityexception_test.go
+++ b/adapters/v1/securityexception_test.go
@@ -1251,8 +1251,8 @@ func TestConvertAffectedRecordsProvenance(t *testing.T) {
 }
 
 func TestRestoreSuppressedMatches_ExternalVEXFalsePositive(t *testing.T) {
-	// This simulates what happens when Grype processes a VEX file and 
-	// Kubevuln drops the tags during conversion. We are left with a single 
+	// This simulates what happens when Grype processes a VEX file and
+	// Kubevuln drops the tags during conversion. We are left with a single
 	// rule that only has the Vulnerability ID.
 	doc := &v1beta1.GrypeDocument{
 		Matches: []v1beta1.Match{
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Verified the core claim end to end:

- `grypeToDomainIgnoredMatchesAppliedIgnoreRules` (adapters/v1/grype_to_domain.go) only ever copies `Vulnerability`, `FixState`, `Package` off a Grype-native rule — it's the only other constructor of `v1beta1.IgnoreRule` in this repo, and it never touches `SourceKind`/`SourceName`/`SourceNamespace`/`Justification`/`ImpactStatement`. So the widened check in `isExceptionSourcedIgnore` can't accidentally treat a Grype/VEX-native ignore as ours.
- `reconcileCachedCVE` (core/services/scan.go) does restore-then-reapply exactly as described: `RestoreSuppressedMatches` → `applyExceptionsToManifest` → `ApplySecurityExceptions`. A match this helper fails to restore never reaches the re-apply, so under-restoring silently pins a stale suppression — confirms why widening this predicate is the right fix rather than a cosmetic one.
- Checked for any other producer of a non-empty, non-CRD `SourceKind` (CodeRabbit's nitpick about the new fallback not gating on `r.SourceKind == ""`) — found none. `buildSuppressionAttributes` is the only writer of the `sourceKind` attribute and it's hardcoded to `"SecurityException"`/`"ClusterSecurityException"`; nothing today (no `CustomIgnore` or similar) can produce a third value. So that's a reasonable future-hardening suggestion, not a live bug — not blocking.
- CodeRabbit's other suggestion (isolate `SourceName`-only / `SourceNamespace`-only cases in the new test) is a fair coverage improvement, also non-blocking — the existing `TestRestoreSuppressedMatches_RestoresCloudSourcedSuppressions` already exercises a realistic backend-policy shape (ruleId + justification, no sourceKind) and correctly proves the fix.

`#701`'s regression tests are untouched and the new tests correctly cover both directions (cloud suppression restored, Grype/VEX-native ignores left alone). CI (build/vet/tests/lint/CodeQL/GitGuardian/DCO) is green. No blockers — approving.

Issue #499, opened 2026-08-07
feature: bound worker-pool queue depth with backpressure (maintainer decision: reject vs. block)


AdeshDeshmukh:
## Description

Every scan handler (`GenerateSBOM`, `ScanCP`, `ScanCVE`, `ScanRegistry` in `controllers/http.go`) validates the request, returns `200 OK`, then submits the actual work to a `gammazero/workerpool.WorkerPool` via `h.workerPool.Submit(...)` with a `context.WithoutCancel` context — a deliberate fire-and-forget design (`docs/ARCHITECTURE.md`: "Scan requests return immediately; work is queued").

---

## Problem

`workerpool.New(concurrency)` bounds concurrent *workers*, not the *queue*. Per the library (`gammazero/workerpool@v1.1.3`): `Submit` never blocks regardless of how many tasks are already queued. Under sustained load past `scanConcurrency` (default `1`), jobs pile into an uncapped backlog:

- Each queued job is a closure capturing the full scan command (image, credentials, args) — an unbounded backlog is unbounded memory growth.
- With `scanTimeout` at 5m, a job that waits 10m in the queue is already stale before it starts.
- `kubevuln_worker_pool_queue_depth` lets us *observe* the backlog growing, but nothing stops it from growing.

#467's acceptance criteria named this explicitly:

> "Either the worker pool gets bounded queue depth (real backpressure on `Submit`, e.g. reject or block past a configured limit) so an unbounded backlog can't accumulate in the first place, or the shutdown path explicitly caps how many queued-but-not-started jobs it will wait for."

#468 implemented only the second half (bounding shutdown's wait, via `Stop()` + `shutdownTimeout`), with @matthyx noting on that PR:

> "agreed those need a maintainer decision on client-facing behavior, not a guess inside this PR."

This issue is that decision.

---

## Environment

- **OS:** N/A — architectural/queue management design
- **Version:** `main` — `controllers/http.go`

---

## Solution Options

| Option | Behavior at capacity | Client sees | Tradeoff |
|---|---|---|---|
| **Reject (recommended)** | Depth ≥ limit → handler returns `503` instead of `200` | Caller learns the scan was *not* queued | Breaks "200 = accepted" for the first time, but callers can retry — same shape as the existing `ErrTooManyRequests` → `429` path |
| **Block** | Handler blocks on a semaphore until room frees up | `200` contract preserved; latency grows under load | Inverts the async design — handlers stop being fast |
[...]

PR #749, opened 2026-08-17
fix: bound the HTTP controller's scan queue with admission control

### LINKED ISSUE #748: HTTP scan queue has no bound or backpressure — Submit() accepts unlimited work regardless of scanConcurrency (#499 closed without a fix)
## Problem

`HTTPController`'s scan queue has no bound. Every scan endpoint (`GenerateSBOM`, `ScanCP`, `ScanCVE`, `ScanRegistry` in `controllers/http.go`) validates the request, writes `200 OK`, and hands the actual work to a `gammazero/workerpool.WorkerPool` via `h.workerPool.Submit(...)` — a fire-and-forget design (`docs/ARCHITECTURE.md:74`: "Scan requests return immediately; work is queued"). `workerpool.New(concurrency)` bounds concurrent *workers*, not the *queue* behind them, and nothing else in the request path enforces a limit. Under sustained load past `scanConcurrency` (which defaults to `1` — `config/config.go:99`), accepted-but-not-started jobs accumulate without limit.

This is exactly the gap **#499** ("feature: bound worker-pool queue depth with backpressure (maintainer decision: reject vs. block)") described. That issue has zero comments, no linked commit or PR, and was closed as `COMPLETED` on 2026-08-07 — but the fix it called for was never merged. I'm re-raising it against current `main` (`a86d391`) with concrete evidence, since the tracker state doesn't match the code.

## Reproduction / Evidence

The unbounded-queue claim is directly confirmed by the vendored library itself, not inferred:

`github.com/gammazero/workerpool@v1.1.3/workerpool.go:94-97,107-111`:
```go
// Submit will not block regardless of the number of tasks submitted. Each task
// is immediately given to an available worker or to a newly started worker. If
// there are no available workers, and the maximum number of workers are
// already created, then the task is put onto a waiting queue.
...
func (p *WorkerPool) Submit(task func()) {
	if task != nil {
		p.taskQueue <- task
	}
}
```
The backing structure is an unbounded `deque.Deque[func()]` (`waitingQueue`, `workerpool.go:50`) — `PushBack` has no capacity check anywhere in the dispatcher (`workerpool.go:178-203`).

On the kubevuln side, every accept path calls `Submit` unconditionally, with no queue-depth check beforehand:
- `controllers/http.go:196` (`runTrackedScan`, shared by `GenerateSBOM`/`ScanCVE`/`ScanRegistry`)
- `controllers/http.go:287` (`ScanCP`'s own inline `Submit`)

The one place queue depth is even looked at is observability, not control: `controllers/http.go:54-61` exposes `kubevuln_worker_pool_queue_depth` via `h.workerPool.WaitingQueueSize()`, and `Shutdown` (`controllers/http.go:463-465`) logs it once on the way out. Neither reads it before accepting new work.

Concretely: with the documented default (`scanConcurrency=1`, `docs/CONFIGURATION.md:171`), if a kubescape operator dispatches a batch of scan requests for many workloads at once (a common trigger — an initial cluster scan, a rollout, or a periodic full rescan), every request past the first is queued behind a single active worker, each one:
- returns `200 OK` immediately (the caller has no way to learn it was merely queued, not accepted for timely execution), and
- carries a closure over the full `domain.ScanCommand`, including `CredentialsList` — registry pull credentials for that image, held in memory for as long as the job waits.

There is no request path back to the caller once `200` is written (`ValidateGenerateSBOM`/`ValidateScanCP`/etc. already reject with `429` via `domain.ErrTooManyRequests` for registry-side rate limiting — `controllers/http.go:361-362` — but that check happens before enqueueing and has nothing to do with local queue depth).

## Root Cause

`workerpool.New(concurrency)` in `NewHTTPController` (`controllers/http.go:40-46`) only caps concurrent execution. Admission control — whether to accept a job before it's queued — was never implemented, because PR #468 (which fixed the *shutdown*-time unbounded wait, #467) explicitly deferred it:

> "Not touched (intentionally, per the issue's own scoping): backpressure on `Submit`/bounded queue depth ... flagged in #467 as needing a maintainer decision on desired client-facing behavior, not something to guess at silently in this PR."

#499 was opened to be that decision and PR. It never got either.

## Impact

- **Reliability / availability**: a sustained burst above `scanConcurrency` grows the backlog without limit. Each queued closure retains a `domain.ScanCommand` (image refs, args, and registry credentials) until it runs, so a large-enough backlog is unbounded memory growth, up to OOM-killing the pod under load the design already anticipated it would face (this is the same class of problem #467/#468 fixed for the *shutdown* path — the backlog those PRs bound *the wait for* can still grow unboundedly *before* shutdown).
- **Correctness of the client contract**: `200 OK` is the same response whether a job starts in milliseconds or sits queued for longer than `scanTimeout` (5m default) before it even begins, at which point the queued-but-not-yet-started job is already effectively stale. Callers have no signal to distinguish "accepted and running soon" from "queued behind an arbitrarily long backlog."
- **Inconsistent backpressure story**: kubevuln already has a real 429/`ErrTooManyRequests` mechanism for registry-side rate limiting, but nothing equivalent for its own local admission control — the one place it would be cheapest to apply backpressure (before doing any work at all) is the one place it's missing.

## Affected Components

- `controllers/http.go` (`HTTPController`, `NewHTTPController`, `runTrackedScan`, `ScanCP`, `Shutdown`, `WithMetrics`)
- `config/config.go` (`Config`, `LoadConfig` — no `maxQueueDepth`/equivalent field exists today)
- `docs/CONFIGURATION.md` / `docs/ARCHITECTURE.md` (would need updating alongside the fix)

## Acceptance Criteria

- A configurable bound on total pending scans (running + waiting, or waiting-only — pick one and document it explicitly, since it will diverge from the existing waiting-only `kubevuln_worker_pool_queue_depth` gauge if "total" is chosen).
- Default preserves current behavior (unbounded) so this ships without changing anyone's runtime behavior until they opt in — e.g. `maxQueueDepth: 0` = unbounded.
- At capacity, new scan requests are rejected synchronously in the handler, before `Submit` is called, reusing the existing rejection machinery (`recordRejection`, `RejectCounter`, `validationStatusCode`) the same way `ErrTooManyRequests` already maps to `429` — consistent with #499's own recommendation, which I'd endorse over the blocking alternative (blocking would turn the fast, async handlers this design deliberately built into slow, synchronous ones under load).
- The bound is enforced without overshoot under concurrent submitters (e.g. an atomic compare-and-swap admission check, not a check-then-act race against `WaitingQueueSize()`).
- Unit test(s) proving: (a) requests are accepted up to the configured limit, (b) the next request past the limit is rejected with the same rejection path/metric as an `ErrTooManyRequests` validation failure, and (c) default configuration (`0`) behaves identically to today (unbounded, always `200`).
- `docs/CONFIGURATION.md` documents the new field and its interaction with `scanConcurrency`, `scanTimeout`, and the existing queue-depth gauge.

## Solution Direction

Add a `maxQueueDepth` config field (default `0` = unbounded). In each handler, before calling `Submit`, atomically check-and-increment a counter guarding admission (or use a buffered channel/semaphore sized to `maxQueueDepth` as the CAS primitive) and reject with the existing `429`/`recordRejection` path if it's already at the limit; decrement when the job's `runTrackedScan`/`ScanCP` closure returns. This keeps the change scoped to `controllers/http.go` and `config/config.go`, mirrors the existing `ErrTooManyRequests` client-facing shape so callers already handling `429` need no new logic, and requires no change to the vendored `workerpool` library itself.

bhuvan-somisetty:
## Overview

**Problem:** every scan handler (`GenerateSBOM`/`ScanCP`/`ScanCVE`/`ScanRegistry`) validates a request, writes `200 OK`, and hands the work to `HTTPController.workerPool` via `Submit(...)`. `gammazero/workerpool`'s `Submit` never blocks and its waiting queue is unbounded regardless of `scanConcurrency` (default `1`) — confirmed directly from the vendored library source, not inferred. Under a sustained burst of requests, accepted-but-not-started jobs pile up without limit, each one holding a closure over the full `domain.ScanCommand`, including registry pull credentials, for as long as it waits.

**Root cause:** `workerpool.New(concurrency)` bounds concurrent *workers*, not the *queue* behind them, and nothing in the request path enforced any limit before this change — admission control (whether to accept a job at all) had never been implemented.

**Solution:** an opt-in `maxQueueDepth` config option (default `0` = unbounded, so this ships with no behavior change for anyone who doesn't set it). Once the controller is holding `maxQueueDepth` accepted-but-unfinished scans (queued + running), a new request is rejected with `503`/`ErrQueueFull` *before* it is queued, instead of joining an unbounded backlog. The rejection reuses the same machinery `ErrTooManyRequests` already goes through (`recordRejection`, `RejectCounter`, `validationStatusCode`), so it shows up in the existing `kubevuln_scan_rejections_total` metric under a new `queue_full` reason, distinct from `too_many_requests`/`invalid_request`.

Admission is tracked with a small CAS-guarded counter (`tryAdmit`/`release`) that is independent of the worker pool's own internal queue, so the same check applies uniformly across all four scan endpoints — including `ScanCP`, which submits to the pool via its own closure rather than the shared `runTrackedScan` helper the other three use.

This resolves the follow-up decision #499 asked for (deferred out of #468/#467's scope at the time) and that #748 re-raised with current evidence after #499 was closed without an implementing change.

## How to Test

```
go build ./...
go vet ./...
golangci-lint run --timeout=10m --new-from-rev="$(git merge-base HEAD upstream/main)"
go test ./...
```

New regression coverage added in `controllers/http_test.go`:
- `TestHTTPController_TryAdmitRelease` — unbounded default, bounded reject-at-capacity, release freeing a slot, and non-positive depth meaning unbounded.
- `TestHTTPController_TryAdmitNoOvershootUnderConcurrency` — 200 goroutines racing a depth-10 budget admit exactly 10, proving the CAS loop doesn't overshoot or starve.
- `TestHTTPController_GenerateSBOM_QueueFull` / `TestHTTPController_ScanCP_QueueFull` — HTTP-level: a full queue returns `503` for both submission paths, and a freed slot is usable by the next request.
- `TestHTTPController_MaxQueueDepth_ReleasesSlotAfterScanCompletes` — end-to-end: a real tracked scan's `runTrackedScan` closure actually calls the deferred `release()` once it finishes, not just that `tryAdmit`/`release` are correct in isolation.
- `TestHTTPController_MetricsEndpoint_RecordsQueueFullRejection` — the `queue_full` reason is recorded distinctly from the other two rejection reasons.
- `TestValidationStatusCode` extended for `ErrQueueFull` → `503`.
- `config.TestLoadConfigMaxQueueDepthDefault` — default is `0`.

All of the above pass locally. The full `go test ./...` run also shows two failures in `adapters/v1` (`Test_syftAdapter_CreateSBOM_TimeoutDoesNotRaceWithAbandonedSyft`, `Test_syftAdapter_CreateSBOM_SerializesWithAbandonedGoroutineAfterTimeout`); this branch does not touch `adapters/` at all, and both pass individually when re-run in isolation — they're pre-existing, timing-sensitive tests that flake under full-suite load on this machine (Windows file-locking noise visible in their output), not a regression from this change.

## Related issues/PRs:

* Resolved #748
* Follow-up to #499, #467, #468

## Summary by CodeRabbit

* **New Features**
  * Added configurable limits for queued and running scans.
  * Requests exceeding the configured queue capacity now receive HTTP 503 responses.
  * Added queue-full metrics to distinguish capacity-based rejections.

* **Documentation**
  * Documented the `maxQueueDepth` setting, including its unbounded default and configuration schema.

### DIFF
--- a/cmd/http/main.go
+++ b/cmd/http/main.go
@@ -158,6 +158,7 @@ func main() {
 		service.SetEventRecorder(eventRecorder)
 	}
 	controller := controllers.NewHTTPController(service, c.ScanConcurrency)
+	controller = controller.WithMaxQueueDepth(c.MaxQueueDepth)
 	controller = controller.WithDiagnostics(func(diagCtx context.Context) domain.Diagnostics {
 		return domain.Diagnostics{
 			ScanMode:                scanMode,
--- a/config/config.go
+++ b/config/config.go
@@ -66,6 +66,7 @@ type Config struct {
 	KeepLocal               bool              `mapstructure:"keepLocal"`
 	ListingURL              string            `mapstructure:"listingURL"`
 	MaxImageSize            int64             `mapstructure:"maxImageSize"`
+	MaxQueueDepth           int               `mapstructure:"maxQueueDepth"`
 	MaxSBOMSize             int               `mapstructure:"maxSBOMSize"`
 	Namespace               string            `mapstructure:"namespace"`
 	NodeSbomGeneration      bool              `mapstructure:"nodeSbomGeneration"`
@@ -97,6 +98,12 @@ func LoadConfig(path string) (Config, error) {
 	v.SetDefault("maxImageSize", 512*1024*1024)
 	v.SetDefault("maxSBOMSize", 20*1024*1024)
 	v.SetDefault("scanConcurrency", 1)
+	// maxQueueDepth bounds how many scans HTTPController will accept but not yet finish
+	// (queued or running) before rejecting new requests instead of queuing them (see #748:
+	// the underlying workerpool.Submit never blocks and its queue is otherwise unbounded).
+	// 0 preserves the pre-existing unbounded behavior, so this ships without changing
+	// anyone's runtime behavior until they opt in.
+	v.SetDefault("maxQueueDepth", 0)
 	v.SetDefault("scanTimeout", 5*time.Minute)
 	v.SetDefault("scannerReadinessTimeout", 60*time.Second)
 	// cmd/http/main.go spends up to 5s on the HTTP server's own Shutdown before this
--- a/config/config.go
+++ b/config/config.go
@@ -101,8 +101,8 @@ func LoadConfig(path string) (Config, error) {
 	// maxQueueDepth bounds how many scans HTTPController will accept but not yet finish
 	// (queued or running) before rejecting new requests instead of queuing them (see #748:
 	// the underlying workerpool.Submit never blocks and its queue is otherwise unbounded).
-	// 0 preserves the pre-existing unbounded behavior, so this ships without changing
-	// anyone's runtime behavior until they opt in.
+	// 0 (or less) preserves the pre-existing unbounded behavior, so this ships without
+	// changing anyone's runtime behavior until they opt in.
 	v.SetDefault("maxQueueDepth", 0)
 	v.SetDefault("scanTimeout", 5*time.Minute)
 	v.SetDefault("scannerReadinessTimeout", 60*time.Second)
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -29,6 +29,16 @@ func TestLoadConfigShutdownTimeoutDefault(t *testing.T) {
 	assert.Equal(t, 20*time.Second, c.ShutdownTimeout)
 }
 
+// TestLoadConfigMaxQueueDepthDefault is a regression test for #748: maxQueueDepth must
+// default to 0 (unbounded), preserving the pre-existing behavior for anyone who doesn't
+// explicitly opt in to bounding the HTTP controller's admission queue.
+func TestLoadConfigMaxQueueDepthDefault(t *testing.T) {
+	viper.Reset()
+	c, err := LoadConfig("testdata")
+	assert.NoError(t, err)
+	assert.Equal(t, 0, c.MaxQueueDepth)
+}
+
 func TestLoadConfigNotFound(t *testing.T) {
 	viper.Reset()
 	_, err := LoadConfig("testdataInvalid")
--- a/controllers/http.go
+++ b/controllers/http.go
@@ -37,12 +37,14 @@ type HTTPController struct {
 	statusesOnce sync.Once
 
 	// maxQueueDepth bounds the number of scans tryAdmit will let through; see
-	// WithMaxQueueDepth. Zero (the default) means unbounded.
+	// WithMaxQueueDepth. Zero (the default) or a negative value means unbounded.
 	maxQueueDepth int
 	// pending counts scans currently accepted but not yet finished (queued or running),
 	// guarded against maxQueueDepth by tryAdmit/release below. Unused when maxQueueDepth
-	// is zero.
-	pending atomic.Int32
+	// is zero. int64, not int32: comparing against int64(h.maxQueueDepth) instead of
+	// narrowing maxQueueDepth into an int32 avoids the wraparound a large configured
+	// value would otherwise cause on a 64-bit build (int is 64 bits there).
+	pending atomic.Int64
 }
 
 // NewHTTPController initializes the HTTPController struct with the injected scanService
@@ -109,9 +111,10 @@ func (h *HTTPController) tryAdmit() bool {
 	if h.maxQueueDepth <= 0 {
 		return true
 	}
+	limit := int64(h.maxQueueDepth)
 	for {
 		current := h.pending.Load()
-		if current >= int32(h.maxQueueDepth) { // #nosec G115 -- maxQueueDepth is operator-configured, not attacker-controlled
+		if current >= limit {
 			return false
 		}
 		if h.pending.CompareAndSwap(current, current+1) {
--- a/controllers/http.go
+++ b/controllers/http.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"strconv"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	wssc "github.com/armosec/armoapi-go/apis"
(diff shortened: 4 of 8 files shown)


bhuvan-somisetty:
Good catches, thanks — fixed both.

The int32 truncation was the real one: `tryAdmit` was narrowing `maxQueueDepth` down to int32 for the comparison, so a configured depth above ~2.1B would wrap negative and reject every single request. Switched `pending` to `atomic.Int64` and compare against `int64(maxQueueDepth)` instead, so there's no narrowing conversion left to overflow. Added a test that pins a depth above `math.MaxInt32` and asserts it still admits.

Also aligned the docs/schema with what the code actually does — runtime already treated any `n <= 0` as unbounded (matches `WithMaxQueueDepth`'s own doc comment), but the config docs and JSON schema only said `0`, and the schema's `minimum: 0` outright disallowed the negative case. Updated the prose to say "0 or a negative value" and dropped the now-contradictory minimum.

Pushed as a follow-up commit. `go build`, `go vet`, `golangci-lint run --new-from-rev`, and `go test ./...` all clean on top of it.

matthyx (approved):
Reviewed the diff and independently verified the fix.

The design is sound: an opt-in, CAS-guarded admission counter (`tryAdmit`/`release`) bounds accepted-but-unfinished scans ahead of the unbounded `workerpool.Submit` queue, defaults to today's unbounded behavior (`maxQueueDepth <= 0`), and applies uniformly across all four scan endpoints including `ScanCP`'s separate submit closure. Slot release is correctly deferred inside each submitted closure so it fires on every exit path, including the `claimTrackedJob` early return.

CodeRabbit flagged two real issues on the first commit, both now fixed in `2ff0547` with regression tests:
- `tryAdmit` narrowed `maxQueueDepth` to `int32` before comparing; a configured depth above `math.MaxInt32` wrapped negative, so every admission read `pending` as already over budget and rejected 100% of requests. Now compares against `int64(maxQueueDepth)` with `pending` as `atomic.Int64`, covered by `TestHTTPController_TryAdmitRelease/depth_above_math.MaxInt32...`.
- Docs/JSON-schema said only `0` means unbounded while the code treated any `n <= 0` as unbounded. Aligned: docs now say "`0` or a negative value", and the schema's contradictory `minimum: 0` was dropped.

Independently verified on the head commit:
- `go build ./...` — clean
- `go vet ./controllers/... ./config/... ./core/domain/...` — clean
- `go test ./...` (full suite) — all packages pass, including the previously-flaky `adapters/v1` tests the PR description called out
- `go test -race -run TestHTTPController_TryAdmit ./controllers/...` — clean

New test coverage (concurrency-safety, HTTP-level 503 + slot recovery, end-to-end release-on-completion, metrics labeling, config default) is thorough and exercises the real failure modes. Good to merge.

bhuvan-somisetty:
Thanks for digging into this one, Matthias - appreciate the independent verification, not just a rubber stamp. Good catches from CodeRabbit too, glad the fixes and regression tests hold up on your end.

PR #752, opened 2026-08-17
fix(adapters): stop metadata-only history entries claiming a layer's order

### LINKED ISSUE #751: LayerOrder does not name one layer: metadata-only history entries take the next layer's order
`LayerOrder` does not name one layer. in `ParseImageManifest` a history entry that produces no layer is given the order of the next real one, so the two come out with the same number.

the loop stamps every entry with `layerIndex`, which only advances on real layers:

```go
layerIndex := 0
for _, historyLayer := range config.History {
	layerInfo := containerscan.ESLayer{
		LayerInfo: &containerscan.LayerInfo{
			CreatedBy:   historyLayer.CreatedBy,
			CreatedTime: &historyLayer.Created.Time,
			LayerOrder:  layerIndex,
		},
	}
	if !historyLayer.EmptyLayer && layerIndex < len(rawManifest.Layers) {
		...
		layerIndex++
	}
```

so `ENV`, `LABEL`, `CMD` and the like carry whatever `layerIndex` has reached, which is the index the next real layer is about to take.

the fixture already in this repo shows it. `testdata/nginx-image-manifest.json`, grouped by order:

```
order=0   real       ADD file:e23ba17afc7850bcc...
order=1   metadata   CMD ["bash"]
order=1   metadata   LABEL maintainer=NGINX Docker Maintainers
order=1   metadata   ENV NGINX_VERSION=1.27.0
order=1   metadata   ENV NJS_VERSION=0.8.4
order=1   metadata   ENV NJS_RELEASE=2~bookworm
order=1   metadata   ENV PKG_RELEASE=2~bookworm
order=1   real       RUN /bin/sh -c set -x && groupadd --system ...
order=2-6 real       COPY ...
order=7   metadata   ENTRYPOINT / EXPOSE / STOPSIGNAL / CMD   (four entries)
```

seven entries answer to order 1, one of which is a layer. four answer to order 7, none of which is.

that is the field's whole job. the comment directly above the loop says so:

> LayerOrder counts real, layer-producing history entries only, matching parseLayersPayload's indexing (which builds the layerMap DomainToArmo attaches to each vulnerability). ... breaking any lookup that correlates the two by LayerOrder.

the numbering of real layers is right, 0 through 6, and matches `parseLayersPayload`. it is the entries that are not layers that also hold those numbers, so going from a vulnerability's order back to the layer that produced it does not land on one thing.

giving a non-layer entry a negative order is enough. real layers are numbered from zero, so there is no in-range value that can mean "not a layer" without colliding with one.

correction to an earlier version of this issue: i had written here that the two hashes never intersect, `parseLayersPayload` keying on `RootFS.DiffIDs` while `ParseImageManifest` uses `rawManifest.Layers[i].Digest`. that is wrong. stereoscope sets that field from `layer.DiffID()`, and says so on the field itself, "Digest is the sha256 digest of the layer contents (the docker \"diff id\")", so both are diff ids and they do agree. i had checked it with made-up values rather than real ones. the order problem above is unaffected: it was read off the fixture in this repo, not from that.

version: main (0beb5a2)

AnshulPatil2005:
## Overview

`ParseImageManifest` stamped every history entry with `layerIndex`, the count of real layers seen so far. For an entry that produces no layer that is the order of the *next* real layer, so the entry and that layer come out holding the same one.

On the nginx fixture in this repo, six `ENV`/`LABEL`/`CMD` entries and the `RUN` layer following them are all order 1, and four more entries share order 7. `LayerOrder` is what correlates a vulnerability, carrying the order `parseLayersPayload` gave it, back to the layer that produced it, and the comment above the loop says exactly that, so looking up order 1 there returns seven entries.

Real layers are unchanged. An entry with no layer now gets `metadataOnlyLayerOrder`.

## Additional Information

the order is negative because there is no in-range value that can mean "not a layer". real layers are numbered from zero, so any non-negative sentinel collides with one of them.

every history entry is still reported, with its `CreatedBy` and `CreatedTime` as before. this changes only the number on the ones that are not layers, so nothing is dropped from the manifest.

the numbering of real layers was already right, 0 through 6 on nginx, matching `parseLayersPayload`. this does not touch it.

the golden file moves in exactly that shape: ten metadata rows go from a real layer's order to `-1`, and the seven real layers are untouched.

correction to an earlier version of this description: i had written that `parseLayersPayload` and `ParseImageManifest` key their hashes off different digests and so never intersect. that is wrong, and nothing in this change rests on it. stereoscope fills `LayerMetadata.Digest` from `layer.DiffID()` and documents it as "the sha256 digest of the layer contents (the docker \"diff id\")", so both sides are diff ids and they match. i had tested that claim with invented values instead of real ones. the duplicate orders were read off the fixture in this repo and are unaffected.

## How to Test

`go build ./... && go vet ./... && go test ./adapters/... ./core/... -count=1`

`TestParseImageManifest` still compares against `testdata/nginx-image-manifest.json` and passes with the updated golden file.

`TestParseImageManifest_LayerOrderNamesOneLayer` is new. it builds a config with two real layers and two metadata-only entries between them, then asserts all four entries are still reported, that each entry with no hash carries `metadataOnlyLayerOrder`, and that the real layers hold orders 0 and 1 once each.

putting the old `order := layerIndex` back fails both of them.

## Related issues/PRs:

* Resolved #751

### DIFF
--- a/adapters/v1/domain_to_armo.go
+++ b/adapters/v1/domain_to_armo.go
@@ -347,6 +347,11 @@ func syftCoordinatesToCoordinates(c []v1beta1.SyftCoordinates) []containerscan.C
 
 }
 
+// metadataOnlyLayerOrder marks a history entry that produced no layer (ENV, LABEL, CMD and
+// the like). Real layers are numbered from zero, so there is no in-range value that means
+// "not a layer"; a negative one says so without colliding with any of them.
+const metadataOnlyLayerOrder = -1
+
 func ParseImageManifest(grypeDocument *v1beta1.GrypeDocument) (*containerscan.ImageManifest, error) {
 	if grypeDocument == nil || grypeDocument.Source == nil {
 		return nil, fmt.Errorf("empty grype document")
@@ -376,13 +381,22 @@ func ParseImageManifest(grypeDocument *v1beta1.GrypeDocument) (*containerscan.Im
 	// etc.) that don't produce a layer; counting those too, as the raw loop index would,
 	// gives every real layer a different LayerOrder here than in the vulnerability report
 	// for the same layer hash, breaking any lookup that correlates the two by LayerOrder.
+	// A metadata-only entry has no layer, so it gets metadataOnlyLayerOrder rather than a
+	// real one. Stamping it with layerIndex, as this used to, handed it the order of the
+	// next real layer, which is the one thing LayerOrder must not do: on the nginx image
+	// six ENV/LABEL/CMD entries and the RUN layer they precede all came out as order 1, so
+	// the lookup this exists to serve returned seven entries for one layer.
 	layerIndex := 0
 	for _, historyLayer := range config.History {
+		order := metadataOnlyLayerOrder
+		if !historyLayer.EmptyLayer {
+			order = layerIndex
+		}
 		layerInfo := containerscan.ESLayer{
 			LayerInfo: &containerscan.LayerInfo{
 				CreatedBy:   historyLayer.CreatedBy,
 				CreatedTime: &historyLayer.Created.Time,
-				LayerOrder:  layerIndex,
+				LayerOrder:  order,
 			},
 		}
 		if !historyLayer.EmptyLayer && layerIndex < len(rawManifest.Layers) {
--- a/adapters/v1/domain_to_armo_test.go
+++ b/adapters/v1/domain_to_armo_test.go
@@ -420,3 +420,46 @@ func Test_domainToArmo_introducedInLayer(t *testing.T) {
 		})
 	}
 }
+
+// LayerOrder is what correlates a vulnerability, which carries the order parseLayersPayload
+// assigned, with the layer in the manifest that produced it. That only works if one order
+// names one layer. Metadata-only history entries used to be stamped with layerIndex, which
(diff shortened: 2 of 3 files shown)


matthyx (approved):
Verified the fix by hand-tracing `TestParseImageManifest_LayerOrderNamesOneLayer`: with two real layers and two metadata-only entries interleaved, `ParseImageManifest` now assigns `0`/`1` to the real layers only, and `metadataOnlyLayerOrder` (-1) to the ENV/CMD entries — no collision. This lines up with `parseLayersPayload`, which already only indexes real (non-`EmptyLayer`) history entries starting at 0, so the two sides now agree on `LayerOrder` for the same layer hash. The nginx golden-file diff is consistent with that (only the metadata rows move to `-1`; the 7 real layers are untouched). `go build`/`vet`/tests and all CI checks (lint, DCO, security) are green.

One non-blocking question, also raised by CodeRabbit's automated pass: `LayerInfo.LayerOrder` is a plain `int` in `armoapi-go`, so `-1` serializes fine, but is there any downstream consumer (ES mapping, dashboard queries) that assumes `layerOrder` is always non-negative? Worth a quick sanity check before this reaches production, but it's not a reason to hold the merge.

Approving — the fix is correct and matches the description.

PR #753, opened 2026-08-17
fix(registryauth): bound the credential cache with a size-capped LRU

### LINKED ISSUE #750: registryauth: credential cache never evicts entries, growing without bound for the process lifetime
registryauth: credential cache never evicts entries, growing without bound for the process lifetime

# Description
`credentialCache` (`internal/registryauth/cache.go`) never removes an entry once it's written. `lookup` treats an expired entry as a cache miss but leaves it sitting in the map (`cache.go:141-149`); the only write path is `get`'s singleflight closure, which only ever adds (`cache.go:114-118`); and the only code that clears the map (`reset`, `cache.go:154-158`) is test-only. There is no TTL sweep, no size cap, no LRU — nothing bounds `entries` for the life of the process.

Both provider caches built on top of it have effectively unbounded key spaces, since they're keyed by registry host, not by a fixed, operator-configured set of registries:
- `gcpCache` is keyed by `host(imageID)` (`registryauth.go:88`) — any host matching `gcr.io` / `*.gcr.io` / `*-docker.pkg.dev` (`registryauth.go:92-95`).
- `ecrCache` is keyed by the full ECR hostname (`registryauth.go:146`), which embeds a 12-digit account ID and region (`ecrHost`, `registryauth.go:122`) — up to 10^12 × ~20 regions of distinct keys.

The cache key comes directly from a workload's image reference (`ResolveSource`, `resolve.go:35-89`, only reached after a real pull returns `401 Unauthorized` from the registry). The account digits in an ECR hostname aren't validated against the caller's actual AWS account before the cache entry is created — `ecrCredentials` (`registryauth.go:162-172`) fetches a token using whatever the ambient AWS credential chain resolves to, independent of the account ID embedded in the hostname that was used as the cache key. So any principal able to set a Pod's image field (any tenant with pod-create in a shared cluster kubevuln scans) can mint a fresh, permanent cache entry per crafted, syntactically-valid-but-nonexistent ECR/GCR host, and organic fleet growth across many accounts/regions over kubevuln's lifetime does the same thing without any adversarial intent.

# Environment
Not version-specific — architectural gap present on `main`, in the credential-caching layer added by #571 / #569.

# Steps To Reproduce
1. In a cluster kubevuln scans, run/observe pods whose image references point at private ECR/GCR hosts that trigger a `401 Unauthorized` on pull (e.g. `<12-digit-account>.dkr.ecr.<region>.amazonaws.com/...`).
2. Each distinct host that reaches `ResolveSource`'s 401 fallback and successfully fetches ambient credentials adds one entry to `ecrCache`/`gcpCache` (`registryauth.go:85`, `registryauth.go:142`) via `credentialCache.get` (`cache.go:80-139`).
3. Wait past that entry's reported expiry (an ECR token is valid ~12h) and confirm via a debugger/added instrumentation that the entry is still present in `credentialCache.entries` — `lookup` (`cache.go:141-149`) returns "not found" for it, but never calls `delete`.
4. Repeat with new distinct hostnames (new fabricated account IDs, or organic fleet churn over time): `entries` grows monotonically, with no code path that ever shrinks it.

# Expected behavior
Expired entries are actually removed from the cache, and the cache enforces a maximum size so a burst of distinct keys — adversarial or organic — cannot grow it without limit for a long-running, privileged, cluster-wide process.

# Actual Behavior
`credentialCache.entries` only grows. Expired entries are skipped on lookup but never deleted, and there is no capacity bound at all, so the map's size is unbounded by anything except how many distinct registry hosts have ever been resolved over the process's lifetime.

# Impact
- Unbounded memory growth in a long-running, privileged, cluster-wide security-scanning process.
- The growth rate is attacker-influenced, not just organic: the cache key is derived from a workload's own image reference, which any pod-creating tenant controls.
- Each new key also means a real upstream STS/ADC token fetch (`ecrCredentials`/`gcpCredentials`), so this amplifies calls to AWS/GCP auth APIs too — a secondary resource-exhaustion / cost vector shared with other workloads on the same node/IAM role, not just an in-process memory leak.

# Affected components
- `internal/registryauth/cache.go` (`credentialCache`, `lookup`, `get`)
- `internal/registryauth/registryauth.go` (`gcpCache`, `ecrCache`)
- Both SBOM paths that go through `ResolveSource` and reach these caches on a 401: the in-process Syft adapter and the sidecar scanner.

# Acceptance criteria
- An expired entry is actually removed from `credentialCache.entries`, not just skipped on lookup (lazily on access and/or via a periodic sweep).
- The cache enforces a maximum size (e.g. bounded LRU) so a burst of distinct keys cannot grow it past a configured cap.
- Regression tests proving: (a) an expired entry is purged, not merely shadowed, and (b) the cache stops growing past its cap and evicts the least-recently-used entry instead of an arbitrary one.
- No change to the existing singleflight-dedup / cache-hit-metrics semantics documented in `cache.go`'s `get`.

# Additional context
Not a duplicate: #618 fixed a metrics-counting bug in this same cache (unrelated to eviction/growth); #526/#528 concern a different cache (`tooManyRequests`) and cross-replica sharing, not local eviction. No maintainer discussion on #569/#571 (where this cache was introduced and reviewed) raised eviction or unbounded growth — the review there focused on cross-account key collisions and context-cancellation propagation, both since fixed.

Suggested direction: replace the raw map with a bounded LRU (e.g. `hashicorp/golang-lru/v2`), sized via a config knob with a sane default, and evict proactively on expiry — either inline in `lookup` when `now().After(entry.expiry)`, or via a periodic sweep goroutine mirroring the `cleaningInterval` pattern already used by the `tooManyRequests` cache in `core/services/scan.go`. Needs care around the existing singleflight/locking invariants documented in `cache.go`.

bhuvan-somisetty:
Fixes #750

## Problem
`credentialCache` (`internal/registryauth/cache.go`) never removed an entry once it was written. `lookup` treated an expired entry as a cache miss but left it sitting in the map, and the only write path (inside `get`'s singleflight closure) only ever added. Nothing ever shrank it.

Both provider caches built on it key by registry host, which is not a small, operator-controlled set: an ECR hostname embeds a 12-digit account ID and region, and a GCR/Artifact Registry hostname is only pattern-matched. Both are read straight out of a scanned workload's image reference, so the key space — and therefore the cache — could grow without bound for the life of a long-running, privileged, cluster-wide process.

## Root cause
- `lookup` (`cache.go`) checked `now().Before(entry.expiry)` and returned "not found" for an expired entry, but never called `delete` on it.
- The only code that ever cleared the map (`reset`) was test-only.
- No size cap existed at all.

## Fix
Replaced the raw `map[string]cacheEntry` with a size-capped LRU (`hashicorp/golang-lru/v2`, already an indirect dependency of this repo — now promoted to direct):
- `lookup` now actively removes an entry it finds expired, instead of leaving it for the LRU to age out on its own schedule.
- The LRU caps `entries` at 1024 and evicts the least-recently-used entry once full, so a burst of distinct keys — adversarial or organic — can't grow the cache past its configured capacity.
- `lru.Cache` is internally thread-safe, so the `sync.Mutex` that only ever guarded `entries` was removed along with it.
- No change to the existing singleflight-dedup / cache-hit-metrics behavior.

## Testing
- `go build ./...` — clean.
- `go vet ./...` — clean (the four `controllers/http.go` diagnostics are unrelated pre-existing findings, confirmed present on `main` before this branch).
- `go test ./internal/registryauth/...` — all 14 tests pass, including two new regression tests:
  - `TestCredentialCache_ExpiredEntryIsPurgedNotJustShadowed` — proves an expired entry is actually removed from the cache, not merely hidden by the expiry check.
  - `TestCredentialCache_BoundedSizeEvictsLeastRecentlyUsed` — fills the cache to its cap, adds one more distinct key, and confirms the cache stays bounded and the least-recently-used key was evicted (forcing a refetch) rather than the cache growing past its limit.
- `go test ./adapters/v1/... ./pkg/sbomscanner/v1/...` (the two callers of this package) — pass.
- `golangci-lint run --new-from-rev=main ./internal/registryauth/...` — 0 issues (note: local golangci-lint is v2.12.2; CI pins v1.64.8, so CI's lint run is the authoritative check).
- Race detector not runnable locally (no C compiler for `-race`/cgo in this environment); the concurrency-sensitive tests in this package (singleflight coalescing, coalesced-metrics) already existed and are unmodified by this change other than the internal storage swap.

## Impact
Bounds memory growth of `internal/registryauth`'s credential caches for the life of the process, and removes the amplification path where each new key also drives a real upstream STS/ADC token fetch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved credential caching with bounded storage to prevent unbounded growth.
  * Automatically evicts the least recently used entries when the cache reaches capacity.
  * Removes expired credentials from the cache more reliably.
  * Preserves existing credential lookup and refresh behavior.
  * Improved reliability when credentials expire and refresh concurrently.

### DIFF
--- a/internal/registryauth/cache.go
+++ b/internal/registryauth/cache.go
@@ -2,6 +2,7 @@ package registryauth
 
 import (
 	"context"
+	"sync"
 	"time"
 
 	"github.com/anchore/stereoscope/pkg/image"
@@ -47,8 +48,15 @@ type credentialFetch func(ctx context.Context) (*image.RegistryCredentials, time
 // HTTP controller's worker pool runs many scans in parallel) each fired their own upstream
 // request, risking cloud provider rate limiting. See #569.
 type credentialCache struct {
-	// entries is a size-bounded LRU, not a plain map: it is already safe for concurrent
-	// use, so entries no longer needs a mutex of its own the way a map would.
+	// mu guards the compound operations below, not individual entries calls: lru.Cache is
+	// already safe for a single Get/Add/Remove call on its own, but lookup's
+	// read-expiry-then-remove and get's re-check-then-add are each a check-then-act
+	// sequence across two separate calls into it, and lru.Cache exposes no compare-and-
+	// delete or compare-and-add to make either one atomic by itself. Without mu, one
+	// goroutine's expiry-driven Remove can race a concurrent refresh's Add for the same
+	// key and delete the fresh entry the refresh just cached instead of the stale one it
+	// observed. See #753 review.
+	mu       sync.Mutex
 	entries  *lru.Cache[string, cacheEntry]
 	group    singleflight.Group
 	strategy string
@@ -129,7 +137,9 @@ func (c *credentialCache) get(ctx context.Context, key string, fetch credentialF
 			return nil, ferr
 		}
 		if !expiry.IsZero() {
+			c.mu.Lock()
 			c.entries.Add(key, cacheEntry{creds: creds, expiry: expiry.Add(-expiryLeeway)})
+			c.mu.Unlock()
 		}
 		return creds, nil
 	})
@@ -154,6 +164,8 @@ func (c *credentialCache) get(ctx context.Context, key string, fetch credentialF
 }
 
 func (c *credentialCache) lookup(key string) (*image.RegistryCredentials, bool) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
 	entry, ok := c.entries.Get(key)
 	if !ok {
 		return nil, false
@@ -161,7 +173,11 @@ func (c *credentialCache) lookup(key string) (*image.RegistryCredentials, bool) 
 	if !c.now().Before(entry.expiry) {
 		// Remove it now instead of leaving it for the LRU to evict on its own schedule: a
 		// key that's looked up once and never again would otherwise sit in the cache,
-		// still occupying a slot, until capacity eventually forced it out.
+		// still occupying a slot, until capacity eventually forced it out. Removing it
+		// under the same mu critical section as the Get/expiry-check above -- rather than
+		// as a separate, later call -- is what keeps this atomic: a concurrent refresh's
+		// Add for this key can only happen entirely before this section starts or entirely
+		// after it ends, never in the middle of it.
 		c.entries.Remove(key)
 		return nil, false
 	}
@@ -172,5 +188,7 @@ func (c *credentialCache) lookup(key string) (*image.RegistryCredentials, bool) 
 // need the next Credentials() call to actually reach the override instead of a value
 // cached by an earlier test.
 func (c *credentialCache) reset() {
+	c.mu.Lock()
+	defer c.mu.Unlock()
 	c.entries.Purge()
 }
--- a/internal/registryauth/cache.go
+++ b/internal/registryauth/cache.go
@@ -2,10 +2,10 @@ package registryauth
 
 import (
 	"context"
-	"sync"
 	"time"
 
 	"github.com/anchore/stereoscope/pkg/image"
+	lru "github.com/hashicorp/golang-lru/v2"
 	"github.com/kubescape/kubevuln/internal/metrics"
 	"golang.org/x/sync/singleflight"
 )
@@ -22,6 +22,17 @@ const expiryLeeway = 1 * time.Minute
 // the result -- for every other caller still waiting on it.
 const fetchTimeout = 30 * time.Second
 
+// maxCacheEntries bounds a credentialCache's size. The cache is keyed by registry host
+// (registryauth.go), and that key space is not a small, operator-controlled set: an ECR
+// hostname embeds a 12-digit account ID and region, and a GCR/Artifact Registry hostname is
+// only pattern-matched, not looked up against a known list. Both are read straight out of a
+// scanned workload's image reference, so without a cap, a fleet with many distinct
+// registries -- or a workload able to set its own image reference -- can grow the cache
+// without bound for the life of the process. Bounding it with an LRU also means an entry
+// that's never looked up again after it expires doesn't sit in the map forever: it ages out
+// under normal turnover instead of only being removed by the lazy check in lookup. See #750.
+const maxCacheEntries = 1024
+
 // credentialFetch is a provider's underlying fetch: it returns credentials plus the time
 // they stop being valid. A zero expiry means "no reported expiry" -- the credential is
 // still returned, but not cached, since there's nothing to bound the cache entry's
@@ -36,8 +47,9 @@ type credentialFetch func(ctx context.Context) (*image.RegistryCredentials, time
 // HTTP controller's worker pool runs many scans in parallel) each fired their own upstream
 // request, risking cloud provider rate limiting. See #569.
 type credentialCache struct {
-	mu       sync.Mutex
-	entries  map[string]cacheEntry
+	// entries is a size-bounded LRU, not a plain map: it is already safe for concurrent
+	// use, so entries no longer needs a mutex of its own the way a map would.
+	entries  *lru.Cache[string, cacheEntry]
 	group    singleflight.Group
 	strategy string
 	// now is overridable in tests so expiry can be exercised without a real time.Sleep.
@@ -61,7 +73,12 @@ type cacheEntry struct {
 }
 
 func newCredentialCache(strategy string) *credentialCache {
-	return &credentialCache{entries: map[string]cacheEntry{}, strategy: strategy, now: time.Now}
+	entries, err := lru.New[string, cacheEntry](maxCacheEntries)
+	if err != nil {
+		// lru.New only errors when size <= 0, and maxCacheEntries is a positive constant.
+		panic(err)
+	}
+	return &credentialCache{entries: entries, strategy: strategy, now: time.Now}
 }
 
 // get returns a cached, still-valid credential for key if one exists, otherwise calls
@@ -112,9 +129,7 @@ func (c *credentialCache) get(ctx context.Context, key string, fetch credentialF
 			return nil, ferr
 		}
 		if !expiry.IsZero() {
-			c.mu.Lock()
-			c.entries[key] = cacheEntry{creds: creds, expiry: expiry.Add(-expiryLeeway)}
-			c.mu.Unlock()
+			c.entries.Add(key, cacheEntry{creds: creds, expiry: expiry.Add(-expiryLeeway)})
 		}
 		return creds, nil
 	})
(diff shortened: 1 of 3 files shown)


bhuvan-somisetty:
@matthyx would appreciate a look here when you have a chance. Happy to make any changes you'd like.

internal/registryauth/cache.go
```diff
@@ -139,10 +154,15 @@ func (c *credentialCache) get(ctx context.Context, key string, fetch credentialF
...
-	if !ok || !c.now().Before(entry.expiry) {
+	entry, ok := c.entries.Get(key)
+	if !ok {
+		return nil, false
+	}
+	if !c.now().Before(entry.expiry) {
+		// Remove it now instead of leaving it for the LRU to evict on its own schedule: a
+		// key that's looked up once and never again would otherwise sit in the cache,
+		// still occupying a slot, until capacity eventually forced it out.
+		c.entries.Remove(key)
```
matthyx, line 181: **Blocker: TOCTOU race can delete a freshly-refreshed entry instead of the stale one it meant to evict.**

`lookup()` used to run under `c.mu`, so the read-expiry-check-delete sequence was atomic. Now `entries.Get` (line 157) and `entries.Remove` (line 165) are two independent, separately-locked LRU calls with no lock spanning both. `lru.Cache` being internally thread-safe only guarantees each *individual* call is safe — it doesn't make this compound "check then act" sequence atomic across goroutines.

Concrete interleaving where this bites:

1. Key `K` holds an expired entry `E1`.
2. Goroutine **A** calls `lookup(K)`: `entries.Get(K)` returns `E1`, expiry check fails → A decides to remove.
3. Before A's `Remove` executes, goroutine **B** (a concurrent `get(K)` call that also missed on `E1`, became singleflight leader, fetched fresh credentials) runs `entries.Add(K, E2)` inside the closure at line 137 — a valid, non-expired entry.
4. A now calls `entries.Remove(K)` (line 165) — this deletes **B's freshly-cached `E2`**, not the stale `E1` A actually observed.

Net effect: a valid, just-fetched credential can be silently evicted moments after being cached, forcing an avoidable refetch. Under the worker pool's concurrent-scan pattern this file's own doc comments describe, multiple callers hitting the same registry key around its expiry boundary is a realistic pattern, not just a theoretical race — and it partially undermines the exact upstream-rate-limiting goal #569/this PR are trying to protect.

`lru.Cache` doesn't expose a compare-and-delete/conditional-remove primitive, so a per-call lock alone can't restore atomicity here. Suggest either:
- Re-introducing a mutex that spans the whole check-then-remove (and the `Add` in `get`'s closure), i.e. keep the LRU for eviction/bounding but keep coordinating compound read-modify-write access to it explicitly, or
- Passing the expected stale entry/expiry into a guarded removal so a concurrent refresh isn't stomped.

(CodeRabbit's automated summary on this PR flags the same race independently as "Moderate" merge risk — worth resolving before merge given this cache guards credential fetch behavior for a cluster-wide, privileged process.)

matthyx (changes requested):
Nice fix for the unbounded-growth issue (#750) — the LRU cap and the active removal-on-lookup are both correct improvements, and the new regression tests (`ExpiredEntryIsPurgedNotJustShadowed`, `BoundedSizeEvictsLeastRecentlyUsed`) are solid single-threaded proofs of that behavior. `go.mod` promotion to a direct dependency is correct.

One blocker before merge: dropping `sync.Mutex` in favor of relying on `lru.Cache`'s internal thread-safety introduces a TOCTOU race in `lookup()` — see inline comment on `cache.go`. The check-then-remove sequence (`entries.Get` then `entries.Remove`) is no longer atomic across goroutines, so a concurrent cache refresh for the same key can be undone by a stale expiry check, deleting a just-fetched valid credential. This is a real, not just theoretical, race given the worker pool's concurrent-scan pattern, and CodeRabbit's automated review flagged the same issue independently.

Holding off on approval until that's addressed (or explicitly accepted with reasoning) — happy to take another look once it's updated.

bhuvan-somisetty:
Good catch, thanks for the detailed interleaving walkthrough — you're right, entries being internally thread-safe per-call doesn't cover the compound check-then-act sequences. Fixed in 020ec2e: reintroduced mu spanning lookup's Get+expiry-check+Remove and get's Add, so a concurrent refresh and an expiry-driven removal for the same key can no longer interleave, only run fully before or after each other. Added a concurrent stress test (TestCredentialCache_ExpiryRemovalDoesNotRaceConcurrentRefresh, 300 iterations) that reproduces the exact scenario you described. Let me know if this looks right.

bhuvan-somisetty:
Nice catch — key "0" was both oldest and LRU there, so a FIFO cache would've passed just as well. Fixed in 2dfc145: re-access "0" before the overflow insert so "1" becomes the true LRU victim, and assert eviction lands there instead. Verified locally, all 12 tests pass.

matthyx (dismissed):
Re-reviewed after the latest pushes. The blocker is resolved correctly, not just papered over: `mu` is now re-scoped around the *entire* compound `Get`+expiry-check+`Remove` sequence in `lookup()`, and around `Add` in `get()`'s closure. That's the actual fix — since a Remove decision is now made against a freshly-read entry taken under the same lock it's acted on, a `Remove` can never fire against an entry that was replaced by a concurrent `Add` after the stale read; the two critical sections simply can't interleave anymore. Verified by walking the interleavings by hand, not just reading the diff.

The `TestCredentialCache_ExpiryRemovalDoesNotRaceConcurrentRefresh` stress test (300 iterations) demonstrates this. One nitpick, already flagged by CodeRabbit: the test's two goroutines share a single `start` barrier but nothing forces the specific interleaving that used to trigger the bug (stale `Get` completing, then a full concurrent fetch+`Add`, then the stale `Remove`) — goroutine 1's path is a few quick lock ops while goroutine 2's involves a full singleflight+fetch round trip, so in the old buggy code this test may not have reliably caught the race across many runs. That's a test-rigor gap worth tightening with an explicit synchronization seam (e.g. an `onMiss`/`onJoined`-style hook to force the interleaving deterministically, matching the pattern already used elsewhere in this file) — but it doesn't undermine the fix itself, which is structurally correct regardless of scheduling. Non-blocking; a good follow-up.

The LRU-vs-FIFO test fix (re-touching key "0" before the overflow insert) is also correct — the original test couldn't have distinguished the two eviction policies.

All CI checks green. Approving.

matthyx:
rebase pls

bhuvan-somisetty:
Done, rebased onto main. Had to resolve one conflict in cache.go — #767 landed the same expired-entry eviction this PR already does via the LRU, so I kept the LRU version and folded that test into the existing eviction test. Ready for another look.

matthyx (approved):
Re-checked after the rebase onto main. This is the same fix I already reviewed and approved (183e529 → 2dfc145) — the rebase just replays it onto a newer base after resolving the conflict with #767, which had independently landed an overlapping-but-weaker expired-entry eviction. The diff confirms the resolution kept this PR's LRU-backed approach (with the mutex correctly re-scoped around the full check-then-act sequences in `lookup()`/`get()`) and dropped #767's now-redundant `TestCredentialCache_ExpiredEntryIsEvicted` in favor of this PR's stronger `TestCredentialCache_ExpiredEntryIsPurgedNotJustShadowed` and the concurrency stress test. No regressions introduced by the rebase.

All CI checks green. Re-approving (my prior approval was auto-dismissed by the new commits, not by any finding).

PR #755, opened 2026-08-17
fix(adapters): stop a zero Retry-After turning a retry into a hot loop

### LINKED ISSUE #754: a zero Retry-After turns the report retry into a hot loop
a `Retry-After` of zero turns the report retry into a hot loop. measured against a local server that always answers 429:

```
Retry-After: 0                         25216 requests in 2s
Retry-After: <date already passed>     14782 requests in 2s
no Retry-After (ordinary backoff)          3 requests in 1.09s
```

`httpPostWithContext` reads the header and hands it to backoff:

```go
if wait, ok := parseRetryAfter(resp); ok {
	if wait > 0 {
		wait = wait.Round(time.Second)
		if wait == 0 {
			wait = time.Second
		}
	}
	return nil, backoff.RetryAfter(int(wait.Seconds()))
}
```

the guard covers a wait that rounds *down* to zero. it does not cover one that is already zero, because `wait > 0` is false and the whole block is skipped.

zero then does more than wait zero. backoff resets its interval whenever it sees a `RetryAfterError`:

```go
var retryAfter *RetryAfterError
if errors.As(err, &retryAfter) {
	next = retryAfter.Duration
	args.BackOff.Reset()
}
```

so each attempt throws away the exponential interval it had built up and sleeps for nothing. there is no backoff left at all, for the rest of the budget.

zero arrives two ways, and `TestParseRetryAfter` already records both as valid parses:

```
{"zero seconds", "0", true, 0},
{"past HTTP-date clamped to zero, not negative", <past date>, true, 0},
```

the second is the one that does not need a misbehaving server. a date-form `Retry-After` that has already passed is what a few seconds of clock skew between kubevuln and the backend produces on its own.

what makes it worse is which statuses reach here. `shouldRetryReport` deliberately retries 429:

```go
		resp.StatusCode != http.StatusNotFound
}
```

so the request being throttled is the one that spins hardest, against a backend that just asked it to slow down. and it is not a corner of the code: `httpPostFunc` defaults to `httpPostWithContext`, `ReportScanFailure` runs it with a 30 second budget and the scan-result submission in `backend_utils.go` with 60, which at the measured rate is on the order of hundreds of thousands of requests from one scan.

rounding and flooring unconditionally is enough, which is what the existing guard was already reaching for.

version: main (0beb5a2)

AnshulPatil2005:
## Overview

A `Retry-After` of zero was honoured literally, which turned the report retry into a hot loop: around 25,000 requests in two seconds against a server answering 429, versus about three with ordinary backoff.

The guard that was there covered a wait that rounds *down* to zero, not one that is already zero. Rounding and flooring unconditionally covers both.

## Additional Information

zero does more than wait zero. backoff resets its interval whenever it sees a `RetryAfterError`, so every attempt threw away the exponential interval it had built and then slept for nothing, leaving no backoff at all for the rest of the budget.

it arrives two ways, and `TestParseRetryAfter` already recorded both as valid parses: an explicit `Retry-After: 0`, and a date-form header that has already passed. the second needs no misbehaving server, only a few seconds of clock skew between kubevuln and the backend.

`shouldRetryReport` deliberately retries 429, so the request being throttled was the one that spun hardest against a backend already asking it to slow down. `httpPostFunc` defaults to this function, with a 30 second budget from `ReportScanFailure` and 60 from the scan-result submission in `backend_utils.go`.

nothing else about the header handling changes. a positive `Retry-After` is still honoured as before, and `parseRetryAfter` is untouched, so absent, negative, overflowing and garbage values still fall through to ordinary backoff.

## How to Test

`go build ./... && go vet ./... && go test ./adapters/... -count=1`

`TestHttpPostWithContext_HonorsRetryAfter` and `TestParseRetryAfter` pass unchanged, which is what says a positive wait and the parsing are the same as before.

`TestHttpPostWithContext_ZeroRetryAfterDoesNotSpin` is new and covers both routes to zero, the explicit `0` and the date already passed. it spends a three second budget against a server that always answers 429 and asserts the attempt count stays in single figures. the bound is loose on purpose: what it rules out is thousands, not a precise number.

putting the old `if wait > 0` back fails it with 38350 attempts on the explicit case and 6490 on the date case.

## Related issues/PRs:

* Resolved #754

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved retry behavior when servers return zero or past `Retry-After` values.
  - Ensures retries wait at least one second, preventing rapid retry loops after rate limiting.

- **Tests**
  - Added regression coverage for zero-valued and expired `Retry-After` responses.

### DIFF
--- a/adapters/v1/backend.go
+++ b/adapters/v1/backend.go
@@ -158,11 +158,17 @@ func httpPostWithContext(ctx context.Context, httpClient httputils.IHttpClient, 
 				return nil, backoff.Permanent(retryErr)
 			}
 			if wait, ok := parseRetryAfter(resp); ok {
-				if wait > 0 {
-					wait = wait.Round(time.Second)
-					if wait == 0 {
-						wait = time.Second
-					}
+				// Rounded because backoff.RetryAfter takes whole seconds, and floored at one
+				// because a zero wait is worse than no header at all: backoff resets its
+				// interval whenever it sees a RetryAfterError, so a zero one retries
+				// immediately with no backoff left, for the whole remaining budget. Zero
+				// arrives from Retry-After: 0 and from a date-form header that has already
+				// passed, which a few seconds of clock skew is enough to produce, and this
+				// path retries 429, so the request being throttled is the one that would
+				// spin against a backend already asking it to slow down.
+				wait = wait.Round(time.Second)
+				if wait <= 0 {
+					wait = time.Second
 				}
 				return nil, backoff.RetryAfter(int(wait.Seconds()))
 			}
--- a/adapters/v1/backend_test.go
+++ b/adapters/v1/backend_test.go
@@ -1524,3 +1524,42 @@ func TestHttpPostWithContext_HonorsRetryAfter(t *testing.T) {
 		"should have waited close to the requested 1s, not the default ~500ms backoff interval")
 	assert.Less(t, elapsed, 5*time.Second, "should not have waited far longer than requested")
 }
+
+// A Retry-After of zero used to be honoured literally. backoff resets its interval whenever
+// it sees a RetryAfterError, so a zero one left no backoff at all and the retry spun for the
+// rest of the budget: measured at roughly 25,000 requests in two seconds against about three
+// with ordinary backoff. Zero arrives from an explicit "0" and from a date-form header that
+// has already passed, which a few seconds of clock skew produces on its own, and this path
+// retries 429, so the request being throttled was the one that would spin.
+func TestHttpPostWithContext_ZeroRetryAfterDoesNotSpin(t *testing.T) {
+	tests := []struct {
+		name   string
+		header func() string
+	}{
+		{"explicit zero", func() string { return "0" }},
+		{"date already passed", func() string {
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Verified the logic: `parseRetryAfter` only ever returns non-negative durations when `ok`, so the old `if wait > 0` guard let an already-zero wait (explicit `Retry-After: 0`, or an expired date header) skip the floor entirely and hit `backoff.RetryAfter(0)` — the hot-loop bug. Unconditionally rounding then flooring at `<= 0` closes both routes to zero while leaving positive waits, and absent/negative/garbage headers (which `parseRetryAfter` already filters out via `ok=false`), unaffected.

New test covers both zero routes (explicit `"0"` and an already-passed date) against a server that always returns 429, and bounds attempts loosely (≤10) over a 3s budget — enough to catch a regression back to thousands of attempts without being a flaky exact-count assertion.

CI is green (build, vet, lint, tests, CodeQL, security scans) and CodeRabbit's automated pass found no actionable issues. No blockers from me — approving.

PR #757, opened 2026-08-17
refactor(controllers): bind the scan command in one place

### LINKED ISSUE #756: the scan handlers repeat the same bind-and-convert preamble four times
the four scan handlers in `controllers/http.go` each open with the same ten lines:

```go
ctx := c.Request.Context()

var websocketScanCommand wssc.WebsocketScanCommand
err := c.ShouldBindJSON(&websocketScanCommand)
if err != nil {
	logger.L().Ctx(ctx).Error("handler error", helpers.Error(err))
	_, _ = problem.Of(http.StatusBadRequest).WriteTo(c.Writer)
	return
}

newScan := websocketScanCommandToScanCommand(websocketScanCommand)
```

`GenerateSBOM`, `ScanCP` and `ScanCVE` are byte identical here, the three of them hashing the same. `ScanRegistry` differs only in binding `RegistryScanCommand` and calling `registryScanCommandToScanCommand`.

that is the last block of its size left in the package. a 6-line duplicate scan over non-test, non-generated code puts the tree at 2.4%, down from 8.7% earlier, and this is the largest remaining group at three occurrences.

folding it needs the wire type to be a parameter, since `ScanRegistry` binds a different one, which a generic helper handles without either caller changing shape.

two smaller things while in here.

`markPhase` in `controllers/scan_status.go` writes the terminal-state check out by hand:

```go
if status.State == domain.ScanStateSucceeded || status.State == domain.ScanStateFailed || status.State == domain.ScanStateAbandoned {
	return
}
```

`isTerminal` is two functions above it and is exactly that expression, and `markTerminal` calls it. so a new terminal state needs both updated and only one of them says so.

and `ScanCP` is the one handler with no test for a body that will not bind. `GenerateSBOM`, `ScanCVE` and `ScanRegistry` all have an "invalid request" case asserting the 400 and its body; `ScanCP` has `_MissingArgsDoesNotPanic` and `_QueueFull` but nothing for the bind. it is also the handler whose worker closure is still its own copy rather than `runTrackedScan`, so it is the one least covered by the others.

version: main (0beb5a2)

AnshulPatil2005:
## Overview

The four scan handlers each opened with the same ten lines: read the request context, bind the body, log and answer 400 if it will not bind, convert to a `domain.ScanCommand`. Three were byte identical; `ScanRegistry` differed only in which wire type it binds and which converter it calls.

`bindScanCommand` is that once, generic over the wire type so `ScanRegistry` uses it too.

## Additional Information

no behaviour change. same 400 status and body, same log line and message, same order of operations, and the converters are unchanged.

the generic parameter is the wire command rather than the result, because the result is a `domain.ScanCommand` in all four cases and only the input type differs.

`err` in each handler now comes from the validation call rather than from the bind, so the assignments below became declarations. nothing else read it between the two.

two small things came with it, both in the same package:

`markPhase` spelled out the terminal-state check by hand while `isTerminal` sits two functions above it and `markTerminal` calls it, so adding a terminal state needed two edits and only one of them said so.

`ScanCP` had no test for a body that will not bind. the other three each have an "invalid request" case; `ScanCP` had `_MissingArgsDoesNotPanic` and `_QueueFull` but nothing on that path, and it is the handler whose worker closure is still its own copy rather than `runTrackedScan`, so it is the one the others cover least.

for scale, a 6-line duplicate scan over non-test, non-generated code goes from 2.4% to 2.1% of lines with this.

## How to Test

`go build ./... && go vet ./... && go test ./controllers/ -race -count=1`

no changes to the existing tests. the "invalid request" cases in `TestHTTPController_GenerateSBOM`, `TestHTTPController_ScanCVE` and `TestHTTPController_ScanRegistry` assert the exact 400 body, `{"status":400,"title":"Bad Request"}`, so them passing unchanged is what says the response is the same one.

`TestHTTPController_ScanCP_InvalidRequest` is new and gives `ScanCP` the same case.

they constrain the helper rather than just compiling against it: making `bindScanCommand` report success on a body that will not bind fails all four, `GenerateSBOM`, `ScanCVE`, `ScanRegistry` and the new `ScanCP` one. before adding the last one it failed three.

## Related issues/PRs:

* Resolved #756

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of invalid scan requests with consistent `400 Bad Request` responses.
  - Prevented updates to scans that have already reached a terminal state.

- **Tests**
  - Added regression coverage for invalid application profile requests.

### DIFF
--- a/controllers/http.go
+++ b/controllers/http.go
@@ -224,23 +224,33 @@ func (h *HTTPController) recordRejection(ctx context.Context, endpoint string, e
 	))
 }
 
+// bindScanCommand decodes the request body into T and converts it to a domain.ScanCommand.
+// A body that will not bind is answered here with the same 400 each handler used to write
+// for itself, and reported as ok=false so the caller returns without scanning. T is the
+// wire command the endpoint accepts, which is WebsocketScanCommand for three of the four
+// and RegistryScanCommand for ScanRegistry.
+func bindScanCommand[T any](c *gin.Context, ctx context.Context, convert func(T) domain.ScanCommand) (domain.ScanCommand, bool) {
+	var cmd T
+	if err := c.ShouldBindJSON(&cmd); err != nil {
+		logger.L().Ctx(ctx).Error("handler error", helpers.Error(err))
+		_, _ = problem.Of(http.StatusBadRequest).WriteTo(c.Writer)
+		return domain.ScanCommand{}, false
+	}
+	return convert(cmd), true
+}
+
 // GenerateSBOM unmarshalls the payload and calls scanService.GenerateSBOM
 func (h *HTTPController) GenerateSBOM(c *gin.Context) {
 	ctx := c.Request.Context()
 
-	var websocketScanCommand wssc.WebsocketScanCommand
-	err := c.ShouldBindJSON(&websocketScanCommand)
-	if err != nil {
-		logger.L().Ctx(ctx).Error("handler error", helpers.Error(err))
-		_, _ = problem.Of(http.StatusBadRequest).WriteTo(c.Writer)
+	newScan, ok := bindScanCommand(c, ctx, websocketScanCommandToScanCommand)
+	if !ok {
 		return
 	}
 
-	newScan := websocketScanCommandToScanCommand(websocketScanCommand)
-
 	details := problem.Detailf("ImageHash=%s", newScan.ImageHash)
 
-	ctx, err = h.scanService.ValidateGenerateSBOM(ctx, newScan)
+	ctx, err := h.scanService.ValidateGenerateSBOM(ctx, newScan)
 	if err != nil {
 		logger.L().Ctx(ctx).Error("validation error", helpers.Error(err),
 			helpers.String("imageSlug", newScan.ImageSlug),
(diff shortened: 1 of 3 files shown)


matthyx (approved):
Reviewed the diff against the pre-refactor `controllers/http.go`/`controllers/scan_status.go` line by line.

- `bindScanCommand[T any]` is behaviorally identical to the four inlined binds it replaces: same `ShouldBindJSON`, same error log line, same 400 body via `problem.Of(http.StatusBadRequest)`. Verified for all four call sites (`GenerateSBOM`, `ScanCP`, `ScanCVE`, `ScanRegistry`) that the right converter (`websocketScanCommandToScanCommand` vs `registryScanCommandToScanCommand`) is wired in, and generic type inference from the passed converter works fine under Go 1.26 (go.mod).
- The `err` declaration change (`ctx, err = ...` → `ctx, err := ...`) is a mechanical consequence of `err` no longer being declared by the removed inline bind, not a behavior change — `ctx` was already declared, `err` is new, so `:=` is correct.
- `ScanCP`'s own worker-closure copy (not using `runTrackedScan`) is untouched by this refactor and still closes over the right `err`.
- `markPhase`'s `isTerminal(status.State)` matches the inlined `Succeeded||Failed||Abandoned` check exactly (confirmed against `isTerminal`'s definition), and it's the same helper `markTerminal` already used.
- New `TestHTTPController_ScanCP_InvalidRequest` fills the real gap called out in the description — `ScanCP` was the one handler with no invalid-body coverage. The follow-up commit (`27b6c69`) already addressed CodeRabbit's resource-cleanup comment (`defer c.Shutdown`, `defer file.Close()`) from the first commit.
- All CI is green (build, vet, race tests, lint, CodeQL, license/security/dependency checks) and the description's claim that the new test fails without the helper (and that `ScanCP`'s case was the last one needed) is consistent with the diff.

No behavior change, no correctness issues found. Approving.

PR #759, opened 2026-08-18
fix(controllers): close the race between a scan request's Submit and Shutdown

### LINKED ISSUE #758: controllers/http: graceful shutdown can panic an in-flight scan request with "send on closed channel"
controllers/http: graceful shutdown can panic an in-flight scan request with "send on closed channel"

# Description
`HTTPController.Shutdown` (`controllers/http.go:570-589`) stops the scan worker pool by calling `h.workerPool.Stop()`. Per `gammazero/workerpool`'s own implementation (`stop()`, `workerpool.go:245-262`), that closes the pool's `taskQueue` channel. Nothing in kubevuln synchronizes this against a request handler that is concurrently calling `Submit` (`workerpool.go:107-111`, an unconditional `p.taskQueue <- task`), which the library's own doc comment on `Stop` explicitly warns about: *"Tasks must not be submitted to the worker pool after calling stop."* Sending on a channel that another goroutine closes concurrently — or has already closed — is a guaranteed Go runtime panic.

`cmd/http/main.go` gives `srv.Shutdown` a 5-second budget to let in-flight HTTP requests finish before calling `controller.Shutdown`, but its own comment shows the authors already know that budget can be missed:

```go
srvCtx, srvCancel := context.WithTimeout(context.Background(), 5*time.Second)
defer srvCancel()
if err := srv.Shutdown(srvCtx); err != nil {
    // Log and continue rather than exit: a failed/timed-out HTTP shutdown must not
    // skip the worker-pool drain below, since that's where actual queued/running
    // scan work would otherwise be silently lost (#467).
    logger.L().Ctx(srvCtx).Error("server shutdown error", helpers.Error(err))
}
controller.Shutdown(c.ShutdownTimeout)
```

`net/http`'s `Server.Shutdown` does not kill handler goroutines that are still active when its context expires — it just returns the timeout error while those goroutines keep running. So a request handler can still be between `admitQueueSlot` (`controllers/http.go:142`, which succeeds) and `runTrackedScan`'s `h.workerPool.Submit(...)` (`controllers/http.go:280-286`, and `ScanCP`'s own copy at `controllers/http.go:370-381`) at the exact moment — or after — `controller.Shutdown` calls `workerPool.Stop()`.

# Environment
Not version-specific — architectural gap present on `main`, in the shutdown sequencing added around #467/#748/#749.

# Steps To Reproduce
1. Configure a scan concurrency such that the worker pool has active/queued work, and send a scan request (`GenerateSBOM`/`ScanCVE`/`ScanRegistry`/`ScanCP`) that passes validation and `admitQueueSlot`.
2. Before that request's handler goroutine reaches `h.workerPool.Submit(...)`, trigger the shutdown sequence (SIGTERM) and let `srv.Shutdown`'s 5s deadline elapse with that handler still running (e.g. a slow validation call, GC pause, or scheduler contention — any delay past 5s on that one goroutine is enough).
3. `controller.Shutdown` calls `h.workerPool.Stop()`, closing `taskQueue`.
4. The still-running handler's `Submit` call reaches `p.taskQueue <- task` on a closed channel and panics.

This is not timing-dependent to demonstrate directly: calling `workerpool.New(n).Stop()` and then `Submit` on the same pool reliably panics with "send on closed channel" — the only question in production is whether a request handler goroutine can still be mid-flight when `Stop()` runs, and `srv.Shutdown`'s own timeout-and-continue path (the `#467` comment above) confirms it can.

# Expected behavior
A scan request that reaches `Submit` after (or concurrently with) `controller.Shutdown` starting must not panic. It should be rejected the same way an at-capacity queue already is (503/`domain.ErrQueueFull`), and any admission slot it reserved via `tryAdmit` must be released.

# Actual Behavior
The request handler's goroutine panics inside `workerPool.Submit`. `gin.Recovery()` (`cmd/http/main.go:212`) catches it, so the process doesn't crash, but:
- The client gets an opaque 500 instead of a clean, retryable rejection, during the exact window (a rolling deploy) where that distinction matters most.
- `h.ensureStatuses().recordAccepted(...)` already ran before the panic, so that job's status is permanently stuck at "accepted" — the panic happens before the submitted closure (which would otherwise report a result) is ever handed to the pool.
- If `WithMaxQueueDepth` is configured, the admission slot reserved by `tryAdmit` is never released: `defer h.release()` lives inside the closure passed to `Submit` (`controllers/http.go:285`), and that closure is never created/run when the channel send itself panics.

# Impact
- A recurring, not rare, event: any Kubernetes rolling deployment or restart of kubevuln under load hits this shutdown path.
- Silent-ish failure: caught by `gin.Recovery`, so nothing crashes, but a client sees a 500 instead of a signal it can sensibly retry on, and (if `WithMaxQueueDepth` is set) the leaked admission slot compounds across repeated deploys/restarts.

# Affected components
- `controllers/http.go` (`runTrackedScan`, `ScanCP`'s inline copy, `Shutdown`)
- `cmd/http/main.go` (shutdown sequencing)

# Acceptance criteria
- A scan request that reaches `Submit` after (or concurrently with) `Shutdown` starting is rejected cleanly (no panic), via the existing 503/`ErrQueueFull` path or an equivalent.
- Any admission slot reserved by `tryAdmit` for such a request is released.
- `Shutdown` does not close the pool out from under a request that is still between `admitQueueSlot` and `Submit`.
- A deterministic (non-timing-based) regression test reproduces the race — e.g. block a goroutine right before its `Submit` call, call `Shutdown`, then release it, and assert no panic and a clean rejection.
- No change to `Shutdown`'s existing timeout-bounded-drain behavior (#467) for work already submitted before shutdown begins.

# Additional context
Not a duplicate: #467 (closed) is about the drain having no time bound, already fixed by the 5s/`ShutdownTimeout` budgets this very race lives inside of. #748/#749 (closed/merged) added the `maxQueueDepth`/`tryAdmit` admission control this issue's fix needs to cooperate with, but neither touched shutdown synchronization — matthyx's review and approval on #749 focused on the `int32` narrowing bug and metrics, and never raised this. No existing test (`TestHTTPController_Shutdown_BoundedByTimeout`) covers a race between an in-flight `Submit` and `Shutdown`; that test only covers a task already running when `Shutdown` is called.

Suggested direction: give `HTTPController` a "stopping" gate (e.g. an `atomic.Bool` set at the start of `Shutdown`, before `workerPool.Stop()`) that `runTrackedScan`/`ScanCP`'s closure check immediately before calling `Submit`, falling back to the same rejection path `admitQueueSlot` already uses on a lost race, with `release()` called for any slot that request had reserved. The remaining, much narrower race (checking the gate and then calling `Submit` are two separate steps) can be closed with a `sync.WaitGroup`-style in-flight counter that `Shutdown` waits to drain (bounded by its own timeout) before calling `workerPool.Stop()`.

bhuvan-somisetty:
Fixes #758

## Problem
A request handler (`GenerateSBOM`/`ScanCVE`/`ScanRegistry`/`ScanCP`) can still be between `admitQueueSlot` and `workerPool.Submit` when `HTTPController.Shutdown` calls `workerPool.Stop()`. Per `gammazero/workerpool`'s own implementation, `Stop()` closes the pool's `taskQueue` channel, and `Submit` is an unconditional `p.taskQueue <- task`. Sending on a channel that's been (or is being) closed concurrently is a guaranteed Go runtime panic — the library's own doc comment on `Stop` warns against exactly this: *"Tasks must not be submitted to the worker pool after calling stop."*

`cmd/http/main.go` gives `srv.Shutdown` a 5-second budget to let in-flight HTTP requests finish first, but `net/http`'s `Server.Shutdown` does not kill handler goroutines still active when that budget expires — it just returns the timeout error while those goroutines keep running. So a slow handler can still reach `Submit` after `controller.Shutdown` has already called `workerPool.Stop()`.

## Root cause
Nothing in `controllers/http.go` synchronized a handler's `workerPool.Submit` call against `Shutdown`'s `workerPool.Stop()`. `tryAdmit`/`admitQueueSlot` only ever checked capacity, never "has shutdown started."

## Impact
`gin.Recovery()` catches the panic, so the process doesn't crash, but:
- The client gets an opaque 500 instead of a clean, retryable signal, during exactly the window (a rolling deploy) where that distinction matters most.
- The job's status is stuck at "accepted" forever: the submitted closure that would otherwise report success/failure never gets created, since the channel send itself panics before that.
- If `WithMaxQueueDepth` is configured, the admission slot reserved by `tryAdmit` is never released (`defer h.release()` lives inside the closure that never gets created).

## Fix
Added a `submitGate sync.RWMutex` + `shuttingDown bool` to `HTTPController`, and a new `submit(task)` helper that all four scan handlers now go through instead of calling `workerPool.Submit` directly:
- `submit` holds `submitGate` for **read** while it checks `shuttingDown` and (if clear) calls `workerPool.Submit`.
- `Shutdown` takes `submitGate` for **write** to set `shuttingDown`, before calling `workerPool.Stop()`. That write-lock can only succeed once every in-flight `submit()` call has released its read lock, so by the time `Shutdown` moves on to `Stop()`, every `submit()` call either already finished handing its task to the (still open) pool, or will see `shuttingDown` and never touch the pool at all. There's no interleaving where a `submit()` call reaches the pool after `Stop()` closes it.
- When `submit()` loses the race, the caller (`runTrackedScan` and `ScanCP`'s own copy) releases the admission slot itself and marks the job abandoned via a new `scanStatusStore.markAbandoned` (mirroring the existing `markAbandonedQueued`), instead of leaving it stuck at "accepted." The 200 OK response was already written before `Submit` is ever called, so this can't change the HTTP response — it can only make sure the job's final state reflects what actually happened.

## Testing
- `go build ./...` — clean.
- `go vet ./...` — clean.
- `go test ./controllers/...` — all 33 tests pass, including three new regression tests:
  - `TestHTTPController_Submit_RejectsAfterShutdown` — a `submit()` call after `Shutdown` has completed is refused and the task never runs.
  - `TestHTTPController_SubmitRacingShutdown_DoesNotPanic` — deterministically reproduces the actual race (no sleeps): a `submit()` call is pinned, via a test-only hook, in the exact window between its shutdown check and handing the task to the pool, while `Shutdown` runs concurrently. Proves `Shutdown` can't proceed past that window and the in-flight call still succeeds without panicking. Before this fix, a version of this race could crash the test process with "send on closed channel".
  - `TestHTTPController_RunTrackedScan_AbandonsJobWhenShuttingDown` — when `submit()` loses the race, the admission slot is released and the job's status becomes `Abandoned`/`shutdown_abandoned` instead of staying stuck at "accepted."
- `golangci-lint run --new-from-rev=main ./controllers/...` — reported 0 issues on the first pass; a longer re-run to get a clean confirmation without an unrelated linter-internal timeout message was still in progress when this PR was opened (note: local golangci-lint is v2.12.2; CI pins v1.64.8, so CI's lint run is the authoritative check).
- Race detector not runnable locally (no C compiler for `-race`/cgo in this environment). `TestHTTPController_SubmitRacingShutdown_DoesNotPanic` is deliberately structured to make the race deterministic via channel synchronization rather than relying on `-race` to catch it probabilistically.

## Impact of the fix
Removes a panic that a real Kubernetes rolling deployment or restart under load can trigger on kubevuln's HTTP scan endpoints, and ensures a scan request caught by that shutdown window ends up with an honest terminal status instead of one stuck at "accepted" forever.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shutdown handling to prevent task-submission races and worker-pool panics.
  * New submissions are safely rejected once shutdown begins.
  * In-progress jobs accepted during shutdown are correctly marked as abandoned.
  * Admission capacity is released reliably for abandoned jobs.
  * Improved scan status consistency for jobs already completed or no longer available.

### DIFF
--- a/controllers/http.go
+++ b/controllers/http.go
@@ -36,6 +36,24 @@ type HTTPController struct {
 	statuses     *scanStatusStore
 	statusesOnce sync.Once
 
+	// submitGate and shuttingDown guard workerPool.Submit against Shutdown's
+	// workerPool.Stop(), which closes the pool's task channel: submitting to a closed
+	// channel panics, and nothing about admitQueueSlot/tryAdmit above prevents a handler
+	// from still being between admission and Submit when shutdown begins (#758). submit()
+	// holds submitGate for read while it checks shuttingDown and calls Submit; Shutdown
+	// takes submitGate for write to set shuttingDown, which can only complete once every
+	// in-flight submit() call has released its read lock -- so by the time Shutdown moves
+	// on to workerPool.Stop(), either a given submit() call already finished handing its
+	// task to the (still open) pool, or it will see shuttingDown and never touch the pool
+	// at all. There is no interleaving where a submit() reaches the pool after Stop().
+	submitGate   sync.RWMutex
+	shuttingDown bool
+	// submitBeforeHook, if set, is called synchronously by submit() once it has confirmed
+	// shutdown hasn't started but before it hands the task to the pool. Tests use it as a
+	// deterministic barrier to prove Shutdown really does wait for this window to close
+	// before calling workerPool.Stop(), instead of a timing-based sleep.
+	submitBeforeHook func()
+
 	// maxQueueDepth bounds the number of scans tryAdmit will let through; see
 	// WithMaxQueueDepth. Zero (the default) or a negative value means unbounded.
 	maxQueueDepth int
@@ -149,6 +167,24 @@ func (h *HTTPController) admitQueueSlot(c *gin.Context, ctx context.Context, end
 	h.recordRejection(ctx, endpoint, domain.ErrQueueFull)
 	_, _ = problem.Of(validationStatusCode(domain.ErrQueueFull)).Append(details).WriteTo(c.Writer)
 	return false
+}
+
+// submit hands task to the worker pool, unless Shutdown has already begun, in which case
+// it returns false without touching the pool at all. Callers must treat false the same as
+// a lost race with shutdown: release the admission slot task's defer h.release() will now
+// never run to release, and mark the job abandoned instead of leaving it stuck "accepted"
+// forever, since its 200 OK response was already written before this call (see #758).
+func (h *HTTPController) submit(task func()) bool {
+	h.submitGate.RLock()
+	defer h.submitGate.RUnlock()
+	if h.shuttingDown {
+		return false
+	}
+	if h.submitBeforeHook != nil {
+		h.submitBeforeHook()
+	}
+	h.workerPool.Submit(task)
+	return true
 }
 
 func (h *HTTPController) ensureStatuses() *scanStatusStore {
@@ -277,11 +313,13 @@ func (h *HTTPController) GenerateSBOM(c *gin.Context) {
 // as "partial" for the metric while still marking the job succeeded, and it has no
 // equivalent in the other three.
 //
-// Called only once the caller has already reserved a slot via admitQueueSlot, so the
-// submitted closure owns exactly one matching release() call; deferred first so it still
-// fires on the claimTrackedJob early return below.
+// Called only once the caller has already reserved a slot via admitQueueSlot, so exactly
+// one of the submitted closure's release() call or the shutdown fallback below owns
+// returning that slot. The submitted closure's is deferred first so it still fires on the
+// claimTrackedJob early return; the fallback only runs when submit() never handed the
+// closure to the pool in the first place, so its own defer never gets the chance to.
 func (h *HTTPController) runTrackedScan(bgCtx context.Context, jobID, endpoint string, scan func(context.Context) error, errMsg string, errDetails ...helpers.IDetails) {
-	h.workerPool.Submit(func() {
+	task := func() {
 		defer h.release()
 		if !h.claimTrackedJob(jobID) {
 			return
@@ -299,7 +337,11 @@ func (h *HTTPController) runTrackedScan(bgCtx context.Context, jobID, endpoint s
 			return
 		}
 		h.ensureStatuses().markSucceeded(jobID)
-	})
+	}
+	if !h.submit(task) {
+		h.release()
+		h.ensureStatuses().markAbandoned(jobID, domain.ScanReasonShutdownAbandoned)
+	}
 }
 
 // Alive returns 200 OK
@@ -377,7 +419,7 @@ func (h *HTTPController) ScanCP(c *gin.Context) {
 	bgCtx := context.WithoutCancel(domain.WithScanPhaseUpdater(ctx, func(phase string) {
 		h.ensureStatuses().markPhase(newScan.JobID, phase)
 	}))
-	h.workerPool.Submit(func() {
+	task := func() {
 		defer h.release()
 		if !h.claimTrackedJob(newScan.JobID) {
 			return
@@ -404,7 +446,11 @@ func (h *HTTPController) ScanCP(c *gin.Context) {
 			h.recordScan(bgCtx, "scanCP", start, "success", nil)
 			h.ensureStatuses().markSucceeded(newScan.JobID)
 		}
-	})
+	}
+	if !h.submit(task) {
+		h.release()
+		h.ensureStatuses().markAbandoned(newScan.JobID, domain.ScanReasonShutdownAbandoned)
+	}
 }
 
 // ScanCVE unmarshalls the payload and calls scanService.ScanCVE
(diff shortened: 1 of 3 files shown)


bhuvan-somisetty:
Good catch, fixed in 2126ead — markAbandoned was falling through to the generic "completed" Phase instead of matching markAbandonedQueued's convention for the same state. Added an assertion on Phase to the regression test so it can't silently drift again.

matthyx (approved):
Reviewed and verified locally against the PR branch (`fix/http-shutdown-submit-race`):

- `go build ./...`, `go vet ./controllers/...` — clean
- `go test ./controllers/...` — all tests pass, including the three new regression tests
- `go test -race ./controllers/...` — clean (race detector was runnable here, unlike the author's environment; confirms `TestHTTPController_SubmitRacingShutdown_DoesNotPanic` really does exercise the race deterministically)
- `golangci-lint run ./controllers/...` — 0 issues
- `gofmt -l controllers/` — clean

Design check: `submit()` is now the single call-site for `workerPool.Submit` (verified via grep — no other caller bypasses it), so all four handlers (`GenerateSBOM`/`ScanCVE`/`ScanRegistry` via `runTrackedScan`, plus `ScanCP`'s own copy) are covered. The `submitGate` RWMutex correctly serializes `Shutdown`'s write-lock against every in-flight `submit()`'s read-lock, so there's no interleaving where `Submit` reaches a closed `taskQueue`. `runTrackedScan`/`ScanCP` release the admission slot and mark the job abandoned when `submit()` loses the race, matching the issue's acceptance criteria — the 200 OK is already written by then so it can't be turned into a 503, which the PR description explains clearly.

One pre-existing (not introduced by this PR) edge case worth noting for awareness, not a blocker: a task can be successfully handed to the pool by `submit()` just before `Shutdown`'s `markAbandonedQueued` sweep runs, and since the sweep only looks at store state (not pool state), it can still flip that same job to `Abandoned` before the task executes — `claimTrackedJob` then bails and the scan silently never runs, though the slot is still released correctly. This same race already existed before this change (`Shutdown` had no synchronization at all previously), so it's unrelated to the panic being fixed here.

No blockers. Nice, well-scoped fix with a genuinely deterministic race regression test. Approving.

PR #762, opened 2026-08-18
fix(safefetch): block IPv4 addresses carried inside an IPv6 one

### LINKED ISSUE #761: safefetch allows the cloud metadata endpoint when it is written as an IPv6 address
`safefetch` blocks the cloud metadata endpoint when it is written as `169.254.169.254`, and allows it when it is written as `64:ff9b::a9fe:a9fe`.

an IPv6 address can carry an IPv4 one inside it, and a packet sent to it is delivered to that IPv4 destination. `net.IP.To4` unwraps only the IPv4-mapped form, `::ffff:a.b.c.d`, which is why the mapped ones are covered. for the rest, `checkIPAllowed` sees an ordinary global-unicast IPv6 address, matches none of its checks, and returns nil:

```
ALLOWED   64:ff9b::a9fe:a9fe      NAT64 well-known prefix -> 169.254.169.254
ALLOWED   64:ff9b::a00:1          NAT64 well-known prefix -> 10.0.0.1
ALLOWED   64:ff9b::7f00:1         NAT64 well-known prefix -> 127.0.0.1
ALLOWED   64:ff9b:1::a9fe:a9fe    NAT64 local-use prefix (RFC 8215)
ALLOWED   2002:a9fe:a9fe::        6to4 -> 169.254.169.254
ALLOWED   ::a9fe:a9fe             IPv4-compatible -> 169.254.169.254
ALLOWED   ::ffff:0:a9fe:a9fe      IPv4-translated -> 169.254.169.254
blocked   ::ffff:169.254.169.254  IPv4-mapped, covered by To4
```

NAT64 is the one that matters. an IPv6-only cluster running DNS64/NAT64 is an ordinary Kubernetes setup, and on one, `64:ff9b::a9fe:a9fe` reaches the metadata endpoint that this package's own doc comment names as what it blocks:

> rejecting private, loopback, and link-local destinations (including cloud metadata endpoints such as 169.254.169.254)

and it is reachable through DNS, not only as a literal. the dial-time check exists precisely because a hostname's resolution is not trusted, so a name resolving to a AAAA record in `64:ff9b::/96` goes through the same gap.

this is the same class of hole the file already handles twice. `carrierGradeNAT` and `thisHostOnThisNetwork` are both there because `IsPrivate()` and `IsLoopback()` miss a range that still reaches somewhere internal, each with a comment saying so. this is a third.

refusing the prefixes outright would be wrong: on an IPv6-only cluster NAT64 is also how a legitimate public IPv4 host is reached, so blocking `64:ff9b::/96` would block `64:ff9b::808:808`, which is 8.8.8.8. extracting the embedded address and running it through the same checks keeps that working and closes the hole.

the RFC 8215 local-use prefix `64:ff9b:1::/48` is the exception. RFC 6052 splits the embedded address around the u octet for prefixes shorter than /96, and a network-specific translation prefix is not somewhere a public feed is published, so refusing it outright is the simpler answer there.

nothing calls this package yet. it exists as the primitive for VEXSource ingestion, #387, which is exactly the code path that will be handed user-supplied URLs.

version: main (0beb5a2)

AnshulPatil2005:
## Overview

`safefetch` blocked the cloud metadata endpoint written as `169.254.169.254` and allowed it written as `64:ff9b::a9fe:a9fe`.

An IPv6 address can carry an IPv4 one, and the packet is delivered to that IPv4 destination. `net.IP.To4` unwraps only the IPv4-mapped form, so for NAT64, 6to4, IPv4-compatible and IPv4-translated addresses `checkIPAllowed` saw an ordinary global-unicast IPv6 address, matched none of its checks, and allowed it.

## Additional Information

NAT64 is the one that matters. an IPv6-only cluster running DNS64/NAT64 is an ordinary Kubernetes setup, and on one `64:ff9b::a9fe:a9fe` reaches the endpoint this package's doc comment names as what it blocks. `64:ff9b::a00:1` and `64:ff9b::7f00:1` likewise reach `10.0.0.1` and `127.0.0.1`. it is reachable through DNS as well as a literal, which is what the dial-time check exists for.

this is the same shape as the two ranges already handled here. `carrierGradeNAT` and `thisHostOnThisNetwork` are both present because a stdlib check misses a range that still reaches somewhere internal, each with a comment saying so; this is a third.

the embedded address is extracted and run through the same checks rather than the prefixes being refused outright, because on such a cluster NAT64 is also how a legitimate public IPv4 host is reached: refusing `64:ff9b::/96` would refuse `64:ff9b::808:808`, which is 8.8.8.8.

`64:ff9b:1::/48` is refused outright instead. RFC 6052 splits the embedded address around the u octet below /96, so reassembling it is fiddly, and a network-specific translation prefix is not somewhere a public feed is published.

the recursion is one level deep by construction: `embeddedIPv4` returns an IPv4 address, and it returns nil for anything `To4` already unwraps, so the inner call cannot produce another.

no caller changes. nothing calls this package yet; it is the primitive for VEXSource ingestion, #387.

## How to Test

`go build ./... && go vet ./... && go test ./internal/safefetch/ -count=1`

`TestCheckIPAllowed` passes unchanged, so the ranges already covered still are.

`TestCheckIPAllowed_EmbeddedIPv4` is new. it covers each transition mechanism pointed at metadata, RFC1918 and loopback, and the three that must stay allowed: NAT64 and 6to4 to a public IPv4, and an ordinary public IPv6 address.

removing the embedded check fails it on six of its cases, the NAT64, 6to4, IPv4-compatible and IPv4-translated ones, and leaves `TestCheckIPAllowed` passing, which is the gap this closes.

## Related issues/PRs:

* Resolved #761

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved network request protection by detecting IPv4 addresses embedded in IPv6 transition formats.
  * Blocked embedded private, loopback, metadata, and local-use destinations.
  * Rejected restricted NAT64 local-use addresses while continuing to allow valid public destinations.

* **Tests**
  * Added coverage for NAT64, 6to4, IPv4-compatible, IPv4-translated, and IPv4-mapped IPv6 addresses.

### DIFF
--- a/internal/safefetch/safefetch.go
+++ b/internal/safefetch/safefetch.go
@@ -31,6 +31,55 @@ var carrierGradeNAT = mustParseCIDR("100.64.0.0/10")
 // working bypass of the loopback check if left unhandled.
 var thisHostOnThisNetwork = mustParseCIDR("0.0.0.0/8")
 
+// The IPv6 transition mechanisms below each carry an IPv4 address inside an IPv6 one,
+// and a packet sent to them is delivered to that IPv4 destination. net.IP.To4 unwraps
+// only the IPv4-mapped form (::ffff:a.b.c.d), so for these the checks in checkIPAllowed
+// see an ordinary global-unicast IPv6 address, match none of them, and allow it: on an
+// IPv6-only cluster running DNS64/NAT64, which is an ordinary Kubernetes setup,
+// 64:ff9b::a9fe:a9fe reaches 169.254.169.254.
+//
+// The embedded address is extracted and checked rather than the prefix being rejected
+// outright, because on such a cluster NAT64 is also how a legitimate public IPv4 host is
+// reached, and blocking the prefix would block those too.
+var (
+	// nat64WellKnown is RFC 6052's well-known prefix, which carries the IPv4 address
+	// in its low 32 bits.
+	nat64WellKnown = mustParseCIDR("64:ff9b::/96")
+	// nat64LocalUse is RFC 8215's local-use prefix. RFC 6052 splits the IPv4 address
+	// around the u octet for prefixes shorter than /96, so rather than reassemble it
+	// this is refused outright: it is a network-specific translation prefix, not
+	// somewhere a public feed is published.
+	nat64LocalUse = mustParseCIDR("64:ff9b:1::/48")
+	// sixToFour is RFC 3056's 6to4 prefix, which carries the IPv4 address in the two
+	// groups after the prefix rather than in the low 32 bits.
+	sixToFour = mustParseCIDR("2002::/16")
+	// ipv4Compatible is the ::a.b.c.d form deprecated by RFC 4291, and ipv4Translated
+	// the ::ffff:0:a.b.c.d form from RFC 2765. Both carry the address in the low 32
+	// bits, and neither is unwrapped by To4.
+	ipv4Compatible = mustParseCIDR("::/96")
+	ipv4Translated = mustParseCIDR("::ffff:0:0:0/96")
+)
+
+// embeddedIPv4 returns the IPv4 address an IPv6 transition mechanism carries inside ip,
+// or nil when ip carries none. It reports nothing for addresses To4 already unwraps,
+// since checkIPAllowed's own checks cover those.
+func embeddedIPv4(ip net.IP) net.IP {
+	if ip.To4() != nil {
+		return nil
+	}
+	ip16 := ip.To16()
+	if ip16 == nil {
+		return nil
+	}
+	switch {
+	case sixToFour.Contains(ip16):
+		return net.IPv4(ip16[2], ip16[3], ip16[4], ip16[5])
+	case nat64WellKnown.Contains(ip16), ipv4Compatible.Contains(ip16), ipv4Translated.Contains(ip16):
+		return net.IPv4(ip16[12], ip16[13], ip16[14], ip16[15])
+	}
+	return nil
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Verified the byte-index math for each transition format against RFC 3056 (6to4, bytes 2-5), RFC 6052 (NAT64 well-known /96, bytes 12-15), RFC 4291 (IPv4-compatible `::/96`) and RFC 2765 (IPv4-translated `::ffff:0:0:0/96`, both bytes 12-15) — all correct, and the prefixes don't overlap so the switch is unambiguous.

`embeddedIPv4`'s early `ip.To4() != nil` guard correctly excludes IPv4-mapped addresses (already handled by `IsLoopback`/`IsPrivate` via their own internal `To4()` unwrap) and terminates the recursion at one level, since `net.IPv4()` returns a mapped-form IP that `To4()` unwraps on the recursive call.

`nat64LocalUse` (`64:ff9b:1::/48`) is correctly rejected outright before the embedded check, avoiding the RFC 6052 u-octet reassembly problem the PR description calls out.

Test coverage in `TestCheckIPAllowed_EmbeddedIPv4` exercises all four transition mechanisms against metadata/RFC1918/loopback, plus the three cases that must stay allowed (NAT64 and 6to4 to a public IP, ordinary public IPv6). CI (build, vet, tests, lint, CodeQL, GitGuardian) is green and CodeRabbit found no actionable comments. No blockers — good to merge.

PR #764, opened 2026-08-18
feat: add a safe temp-file writer for VEX documents

### LINKED ISSUE #760: grype's VEX matching engine exists but is never used in the real scan path
### What's happening

Grype (the scanner Kubescape already uses every scan) has its own built-in
tool for reading vendor VEX notes (like Red Hat's or Chainguard's) and using
them to quiet down matching findings  it's called `vex.Processor`.

I checked, and this tool is never actually called anywhere in the real scan
path (`GrypeAdapter.ScanSBOM` in `adapters/v1/grype.go`). It's just sitting
there unused, confirmed by searching the whole repo for `ApplyVEX` /
`vex.NewProcessor` / `vex.Processor`  nothing comes up outside grype's own
library code.

### Why this matters

If Kubescape ever wants to read a vendor's VEX notes and use them to stay
quiet about already-resolved bugs, this is the exact tool it would need to
call. Right now it's a free, ready-to-use capability that nothing in this
repo touches.

### The real blocker, once you try to actually use it

I checked grype's own source code directly. `vex.Processor` only accepts
**real file paths on disk** — it never accepts a VEX document already sitting
in memory:

    func (ovm *Processor) ReadVexDocuments(docs []string) (interface{}, error) {
        vexdata, err := openvex.MergeFiles(docs)
        ...
    }

So before anyone can wire this tool into a real scan, they first need a
**safe way to write a fetched VEX document to a temporary file** — safely
named, cleaned up afterward, and protected against being overwritten or
abused.

Shreya2005-2005:
## Overview

grype's own `vex.Processor` (`grype/vex`) has real VEX-matching support, but
nothing in this repo calls it  confirmed by a repo-wide search for
`ApplyVEX`/`vex.NewProcessor`/`vex.Processor` turning up nothing outside
grype's own library code.

The processor's `ReadVexDocuments` only accepts real file paths
(`openvex.MergeFiles(docs []string)`, `csaf.LoadAdvisory(fname string)`) - it
has no way to accept an in-memory document. So before anyone can wire VEX
matching into a real scan, they first need a safe way to write a fetched VEX
document to disk.

## Changes

Adds `internal/vexdoc.WriteToTempFile`, which writes a document to a
uniquely-named file inside the OS temp directory, with restrictive (`0o600`)
permissions, and returns a cleanup function - matching the standard
`os.CreateTemp` pattern. This does **not** wire VEX matching into the scan
path itself; it only closes the concrete blocker anyone doing that would hit
first.

## Before / After

A live demo confirmed the real danger this prevents. A naive writer using a
predictable filename (e.g. keyed on CVE ID) silently loses data the moment
two different vendors comment on the same CVE:

```
=== BEFORE: naive, predictable-filename writer ===
Red Hat's document written to:    /tmp/vex-CVE-2024-0001.json
Chainguard's document written to: /tmp/vex-CVE-2024-0001.json
What's actually on disk now:      {"vendor": "Chainguard says: affected!"}
-> Red Hat's note is GONE. Chainguard's silently overwrote it.
```

`WriteToTempFile`'s random filenames make this collision structurally
impossible:

```
=== AFTER: internal/vexdoc.WriteToTempFile ===
Red Hat's document written to:    /tmp/kubevuln-vexdoc-fc107fff....json
Chainguard's document written to: /tmp/kubevuln-vexdoc-1b5c42e5....json
Red Hat's file still contains:    {"vendor": "Red Hat says: not affected"}
Chainguard's file still contains: {"vendor": "Chainguard says: affected!"}
-> Both documents survive, safely, on two different real paths.
```

## Testing

9 tests, all passing:
- Real content is written correctly
- Empty documents are rejected
- Two documents never collide on the same path (both pairwise and at
  1000-call scale)
- Cleanup actually removes the file
- Cleanup is safe to call on the error path
- File permissions are correctly restrictive
- The file is always written inside the OS temp directory, never a
  caller-chosen path
- `O_EXCL` genuinely rejects a write to an already-existing path rather than
  silently overwriting it

Full build, `go vet`, and `go test ./internal/vexdoc/...` pass with no
failures.

## Related issues/PRs

* Resolved #760

```

## Summary by CodeRabbit

* **New Features**
  * Added secure temporary document creation with unique, unpredictable filenames.
  * Temporary files are stored with owner-only permissions and safely cleaned up after use.
  * Empty documents are rejected with a clear error.

* **Bug Fixes**
  * Prevented accidental overwrites and ensured partial files are removed when an operation fails.

* **Tests**
  * Added comprehensive coverage for content, cleanup, security, collision prevention, and repeated usage.

### DIFF
--- a/internal/vexdoc/vexdoc.go
+++ b/internal/vexdoc/vexdoc.go
@@ -1,0 +1,101 @@
+// Package vexdoc provides a safe way to write a VEX document to a temporary
+// file on disk. grype's own vex.Processor (grype/vex) only accepts real file
+// paths - it has no way to accept an in-memory document - so anything that
+// wants to hand a fetched VEX document to grype's matcher must first write
+// it to disk safely. This package is that safe primitive: nothing in this
+// repo currently calls grype's VEX matching (confirmed by a repo-wide search
+// for ApplyVEX/vex.NewProcessor/vex.Processor turning up nothing outside
+// grype's own library code), so wiring VEX matching in is separate, later
+// work - this package only closes the concrete blocker anyone doing that
+// would hit first.
+package vexdoc
+
+import (
+	"crypto/rand"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+// Errors returned by WriteToTempFile.
+var (
+	// ErrEmptyDocument is returned when data is empty. An empty file is
+	// never useful to a VEX processor and is more likely a caller bug
+	// (e.g. an unchecked failed fetch) than a genuine empty document.
+	ErrEmptyDocument = errors.New("vexdoc: document is empty")
+)
+
+// WriteToTempFile safely writes data (a fetched VEX document) to a new,
+// uniquely-named file inside the OS's temp directory, and returns its path
+// along with a cleanup function that removes it.
+//
+// Safety properties:
+//   - The filename is derived from random bytes, not from any
+//     caller-supplied or document-derived value, so two different documents
+//     can never collide on the same path and one write can never overwrite
+//     another.
+//   - The file is created with 0o600 permissions (owner read/write only),
+//     not the more permissive default, since a VEX document's contents
+//     (which images/packages a vendor has commented on) are not meant to be
+//     world-readable.
+//   - The file is always created inside os.TempDir() - never a
+//     caller-supplied directory - so this function cannot be used to write
+//     to an arbitrary, attacker-chosen location.
+//   - The returned cleanup function is always safe to call, even if writing
+//     failed partway through; callers should defer it immediately after
+//     checking err, matching the standard os.CreateTemp pattern.
+//
+// This function does not itself enforce a maximum document size. It is not
+// responsible for fetching data over the network - a caller pulling a VEX
+// document from an external source should bound its size before it ever
+// reaches this function (see internal/safefetch.Fetcher.MaxBytes for the
+// existing pattern this repo already uses for that).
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed `internal/vexdoc.WriteToTempFile`. This is a small, self-contained, additive package (no existing call sites touched), and the implementation is sound:

- Filename derived from `crypto/rand` (16 bytes), never from caller/document-derived data → no collision risk between concurrent writers.
- `os.O_WRONLY|os.O_CREATE|os.O_EXCL` with `0o600` → atomic create, no TOCTOU overwrite, owner-only perms.
- Always writes inside `os.TempDir()`, never a caller-supplied path.
- Error paths (empty doc, open failure, write failure, close failure) all correctly clean up and return non-empty errors; `cleanup()` is a safe no-op to call multiple times or after a failed write.
- All CI checks green (build, lint, CodeQL, GitGuardian, DCO, tests).

No blockers. One non-blocking nit, already flagged by CodeRabbit inline on `vexdoc_test.go` around line 105-127: `TestWriteToTempFile_NeverOverwritesAnExistingFile` proves the `O_EXCL` collision guarantee via a raw `os.OpenFile` call rather than through `WriteToTempFile` itself, so the test wouldn't catch a regression that dropped `O_EXCL` from the real function. Worth tightening in a follow-up (e.g. a package-private filename seam), but it doesn't undermine confidence in the shipped code — the production path plainly uses `O_EXCL`, and the other 8 tests do exercise it.

Approving.

PR #766, opened 2026-08-18
fix(services): stop dropping registry credentials that carry only half a pair

### LINKED ISSUE #765: registry credentials are silently dropped when a pull secret carries only half a username/password pair
two ways a pull secret's credentials silently never reach the registry. both end the same, an anonymous pull and a 401, with nothing saying the credentials were discarded on the way.

the first is in the fallback to the auth field:

```go
username, password := cred.Username, cred.Password
if username == "" && password == "" && cred.Auth != "" {
    username, password = credentialsFromAuth(cred.Auth)
}
if username != "" && password != "" {
    rc.Username = username
    rc.Password = password
}
```

`credentialsFromAuth`'s own comment says why the fallback is there: auth is the canonical field, "username/password are only supplementary and are often absent", and falling back is "required for those credentials to reach the registry pull at all, rather than being silently dropped".

an entry can carry one half without the other, and then neither branch takes it. the fallback wants both missing, so it is skipped; the check below wants both present, so it fails; the credentials are dropped:

```
auth only (canonical docker login)             -> credentials present
username+password+auth (kubectl create secret) -> credentials present
username set, password empty, auth present     -> DROPPED
password set, username empty, auth present     -> DROPPED
```

the second is in the authority:

```go
// server address has no scheme
if !strings.HasPrefix(serverAddress, "http") {
    res, _, _ := strings.Cut(serverAddress, "/")
    return res
}
parsedURL, err := url.Parse(serverAddress)
if err != nil || parsedURL.Host == "" {
    return serverAddress
}
```

an address is not a URL because it starts with those four letters. a registry called `http-registry.internal` takes the URL branch, `url.Parse` reports an empty Host for a string with no scheme, and the whole address comes back with its path still on it:

```
reg.example.com/v2/              -> "reg.example.com"
https://reg.example.com/v2/      -> "reg.example.com"
http-registry.internal/v2/       -> "http-registry.internal/v2/"
httpsregistry.example.com/v2/    -> "httpsregistry.example.com/v2/"
```

`Authority` is what stereoscope matches a credential against the registry it is pulling from, so one carrying a path matches nothing and the credential goes unused.

the first is the one worth fixing regardless of hostname: consulting auth whenever the pair is incomplete, rather than only when both halves are gone, is what the comment already describes. taking only a complete decoded pair keeps a malformed auth from overwriting a username that was set.

version: main (ff11524)

AnshulPatil2005:
## Overview

Two ways a pull secret's credentials silently never reached the registry. Both ended the same, an anonymous pull and a 401, with nothing saying they had been discarded.

The auth field is the canonical one and username/password are supplementary, so an entry can carry one without the other. The fallback to auth required both halves to be missing, so an entry with a username and no password fell through it, then failed the both-non-empty check below it, and was dropped.

`parseAuthorityFromServerAddress` treated any address starting with the four letters `http` as a URL, so `http-registry.internal/v2/` came back as the authority with its path still attached and matched no registry.

## Additional Information

`credentialsFromAuth`'s comment already describes the intent: auth is canonical, username and password "are only supplementary and are often absent", and the fallback is "required for those credentials to reach the registry pull at all, rather than being silently dropped". Requiring both halves to be empty is what left the incomplete case outside it.

only a complete decoded pair is taken, so a malformed auth cannot overwrite a username that was set: that entry ends up with no credentials, as it did before, rather than with half a pair from a failed decode.

on the authority, `url.Parse` reports an empty Host for a string with no scheme, which is why the whole address came back. `Authority` is what stereoscope matches a credential against the registry being pulled from, so one carrying a path matches nothing.

the hostname half needs a registry literally named `http...` to bite, which is narrow; the credential half does not depend on the hostname at all.

## How to Test

`go build ./... && go vet ./... && go test ./core/... -count=1`

`Test_registryCredentialsFromCredentialsList` passes unchanged, including its #611 case where only auth is set, so the canonical shapes still convert the way they did.

`Test_registryCredentialsFromCredentialsList_incompletePair` is new: username without password, password without username, a complete pair being kept over auth, and a malformed auth not clobbering a username that was set.

`Test_parseAuthorityFromServerAddress` gains `http-registry.internal/v2/` and `httpsregistry.example.com/v2/`, plus a host:port case.

putting either condition back fails them: the old fallback fails both incomplete-pair cases, and the old prefix check fails the authority test.

## Related issues/PRs:

* Resolved #765

### DIFF
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -898,8 +898,15 @@ func registryCredentialsFromCredentialsList(credentials []registry.AuthConfig) [
 			rc.Token = cred.RegistryToken
 		}
 		username, password := cred.Username, cred.Password
-		if username == "" && password == "" && cred.Auth != "" {
-			username, password = credentialsFromAuth(cred.Auth)
+		// Consulted whenever the pair is incomplete, not only when both halves are
+		// missing: an entry carrying a username but no password is still an entry whose
+		// credentials live in auth, and requiring both to be empty dropped it entirely
+		// rather than falling back. Only a complete decoded pair is taken, so a
+		// malformed auth cannot overwrite a username that was set.
+		if (username == "" || password == "") && cred.Auth != "" {
+			if u, p := credentialsFromAuth(cred.Auth); u != "" && p != "" {
+				username, password = u, p
+			}
 		}
 		if username != "" && password != "" {
 			rc.Username = username
@@ -935,8 +942,11 @@ func parseAuthorityFromServerAddress(serverAddress string) string {
 		return ""
 	}
 
-	// server address has no scheme
-	if !strings.HasPrefix(serverAddress, "http") {
+	// A scheme is "http://" or "https://", not a host that merely starts with those
+	// letters. Treating http-registry.internal as a URL sends it to url.Parse, which
+	// returns an empty Host for a string with no scheme, so the whole thing came back
+	// as the authority with its path still attached and matched no registry.
+	if !strings.HasPrefix(serverAddress, "http://") && !strings.HasPrefix(serverAddress, "https://") {
 		res, _, _ := strings.Cut(serverAddress, "/")
 		return res
 	}
--- a/core/services/scan_test.go
+++ b/core/services/scan_test.go
@@ -1442,6 +1442,61 @@ func Test_parseAuthorityFromServerAddress(t *testing.T) {
 	assert.Equal(t, "quay.io", parseAuthorityFromServerAddress("quay.io"))
 	assert.Equal(t, "x.quay.io", parseAuthorityFromServerAddress("https://x.quay.io"))
 	assert.Equal(t, "europe-docker.pkg.dev", parseAuthorityFromServerAddress("europe-docker.pkg.dev/xxx/xxx"))
+	assert.Equal(t, "quay.io:5000", parseAuthorityFromServerAddress("quay.io:5000"))
+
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Reviewed both changes in `core/services/scan.go` against the full diff and the surrounding functions (`credentialsFromAuth`, `registryCredentialsFromCredentialsList`, `parseAuthorityFromServerAddress`).

**Credential fallback fix**: correct. Falling back to `Auth` whenever either half of username/password is missing (instead of only when both are missing) closes the drop case for `.dockerconfigjson` entries with a partial pair. Only taking the decoded pair when both halves are non-empty means a malformed `Auth` can't clobber a good half — verified this against the new `Test_registryCredentialsFromCredentialsList_incompletePair` cases (partial pair via each side, complete pair taking precedence over `Auth`, and malformed `Auth` leaving the entry fully empty rather than partially populated). The existing `#611` regression case in `Test_registryCredentialsFromCredentialsList` still passes unchanged.

**Authority parsing fix**: correct. Restricting the scheme check to literal `http://`/`https://` prefixes instead of any string starting with `http` fixes the `http-registry.internal` false positive, and the added `httpsregistry.example.com` / `quay.io:5000` cases confirm no regression for legitimate scheme and host:port handling.

No behavioral or test-coverage gaps found. CI (build, lint, tests, CodeQL, GitGuardian, DCO) is green. Approving.

PR #767, opened 2026-08-18
fix(registryauth): evict expired entries from the credential cache

magic-peach:
## Overview

`credentialCache.lookup()` correctly treats an expired entry as a cache miss, but never removes it from the map. The only place that ever deletes from `entries` is `reset()`, which is test-only. So every distinct registry host a cluster ever pulled a private image from stays in the map forever, even long after its credential expired.

This adds the delete on the expired-entry path in `lookup()`.

It doesn't address the other half of #750 (unbounded key cardinality from ECR account/region combinations, before anything even expires) - that needs an actual size cap or LRU, which felt like a separate, bigger decision than this fix. Happy to follow up on that separately if it's wanted.

## How to Test

New test `TestCredentialCache_ExpiredEntryIsEvicted` - fails on current `main` (map still has the entry after it's expired), passes with this change.

## Related issues/PRs:

* Related to #750 (partial fix - see note above)

## Summary by CodeRabbit

* **Bug Fixes**
  * Expired credential cache entries are now removed during lookup, preventing stale data from being retained.

* **Tests**
  * Added coverage to verify that expired entries are properly evicted.

### DIFF
--- a/internal/registryauth/cache.go
+++ b/internal/registryauth/cache.go
@@ -142,7 +142,14 @@ func (c *credentialCache) lookup(key string) (*image.RegistryCredentials, bool) 
 	c.mu.Lock()
 	defer c.mu.Unlock()
 	entry, ok := c.entries[key]
-	if !ok || !c.now().Before(entry.expiry) {
+	if !ok {
+		return nil, false
+	}
+	if !c.now().Before(entry.expiry) {
+		// expired entries are never overwritten by get() until something asks for this
+		// key again, so without deleting here they'd sit in the map forever -- across
+		// however many distinct ECR/GCR hosts a cluster ever pulls from.
+		delete(c.entries, key)
 		return nil, false
 	}
 	return entry.creds, true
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed the diff and the full `cache.go`/`cache_test.go` files.

- The fix is minimal and correct: `lookup()` now deletes an entry once it's found to be expired, closing the leak described in the PR body (every distinct registry host stayed in the map forever). The delete happens under the existing `c.mu` lock, so it's race-safe with the rest of the cache (`get()`'s singleflight re-check, `reset()`).
- No behavior change for callers: a lookup on an expired key still returns `(nil, false)` either way; the only difference is the stale entry no longer lingers in the map.
- New test `TestCredentialCache_ExpiredEntryIsEvicted` directly exercises the fix (fails on `main`, passes here) and follows the existing pattern of overriding `now` for deterministic expiry.
- Scoping out the ECR key-cardinality half of #750 as a separate follow-up (needs a real size cap/LRU, a bigger design decision) is the right call for this PR.

PR #770, opened 2026-08-18
Fix/vex canonical hash determinism

### LINKED ISSUE #769: calculateVexCanonicalHash produces non-deterministic VEX IDs due to unsorted map iteration and missing alias separator
# Description
I was looking at `calculateVexCanonicalHash` in `repositories/apiserver.go` while reading up on the VEX ingestion proposal (#387) and noticed a few bugs in how the canonical hash is calculated:

1. **Map iteration in `cstringFromComponent` isn't sorted:**
   It iterates `c.Hashes` and `c.Identifiers` with `range`. Because Go map iteration is randomized at runtime, calling `calculateVexCanonicalHash` on the same document gives different hashes across runs whenever a component has multiple hashes (like sha256 + sha1) or identifiers.

2. **Missing colon before aliases in `cstringFromVulnerability`:**
   `cString` ends right after `v.Name`. When aliases exist, `strings.Join(list, ":")` is appended without a leading `:`, so the first alias gets glued directly to the CVE name. For example, `Name: "CVE-2023-1"` with alias `"000"` becomes `":<ID>:CVE-2023-1000"`, colliding with `Name: "CVE-2023-1000"` with no aliases.

3. **`sortVexStatements` doesn't break ties:**
   It only compares `(Vulnerability.Name, Timestamp)`. When a single CVE affects multiple packages in an image, both statements have identical CVE names and empty timestamps (defaulted to the doc timestamp). Because there is no secondary sort key, the resulting hash depends on whatever order the statements were in the slice.

4. **In-place slice mutation:**
   `sortVexStatements` is called directly on `vexDoc.Statements`, which modifies the order of the caller's slice backing array during a hash calculation.

# Environment
OS: Linux
Version: main (ff11524)

# Steps To Reproduce

Running this test in `repositories/apiserver_test.go` shows the hash changing on the same document:

```go
func TestCalculateVexCanonicalHash_Nondeterministic(t *testing.T) {
	doc := v1beta1.VEX{
		Metadata: v1beta1.Metadata{
			Context:   "https://openvex.dev/ns/v0.2.0",
			Author:    "kubescape.io",
			Timestamp: "2026-01-01T00:00:00Z",
			Version:   1,
		},
		Statements: []v1beta1.Statement{
			{
				Vulnerability: v1beta1.VexVulnerability{ID: "https://nvd.nist.gov", Name: "CVE-2023-1234"},
				Status:        v1beta1.Status("not_affected"),
				Products: []v1beta1.Product{
					{
						Component: v1beta1.Component{
							ID: "pkg:deb/debian/libssl3@3.0.11",
							Hashes: map[v1beta1.Algorithm]v1beta1.Hash{
								"sha256": "abcdef1234567890abcdef1234567890abcdef1234567890abcdef1234567890",
								"sha1":   "1234567890abcdef1234567890abcdef12345678",
							},
						},
					},
				},
			},
		},
	}

	firstHash, _ := calculateVexCanonicalHash(doc)
	for i := 0; i < 20; i++ {
		h, _ := calculateVexCanonicalHash(doc)
		if h != firstHash {
			t.Fatalf("run %d gave different hash: %s != %s", i, h, firstHash)
		}
	}
}
```

# Expected behavior
- `calculateVexCanonicalHash` should return the exact same hash every time for the same document.
- Vulnerability name and aliases shouldn't collide.
- Statement order in the input slice shouldn't change the document hash.
- Calculating the hash shouldn't mutate the slice order of the caller's struct.

# Actual Behavior
- `calculateVexCanonicalHash` gives different hashes across iterations due to Go map randomization.
- `CVE-2023-1` with alias `000` and `CVE-2023-1000` with no aliases produce the exact same hash `282bc98e206f39d0d42de30a8a100b78d9b7222d914525a21e96b322bc2c0c54`.
- Swapping statement order in the slice changes the resulting hash.

# Additional context
This matters because `OpenVulnerabilityExchangeContainer` CRDs use this hash for `Metadata.ID`, and feed ingestion (#387) needs reliable content hashes so it doesn't trigger useless apiserver updates when upstream VEX data hasn't changed.

A fix would be:
1. Sort map keys before looping in `cstringFromComponent`.
2. Add `":"` before appending aliases in `cstringFromVulnerability`.
3. Add secondary tie-breakers (like package PURL and status) in `sortVexStatements`.
4. Clone the statements slice before sorting it.
5. Add unit tests for `calculateVexCanonicalHash`.

h30s:
## Overview

This PR fixes a few bugs in `calculateVexCanonicalHash` that caused non-deterministic VEX document hashes and potential ID collisions.

### Current Behavior:
- **Random map iteration:** Iterating over `Component.Hashes` and `Component.Identifiers` directly with Go's `range` produced different hash outputs on the same document across runs whenever multiple hashes or identifiers were present.
- **Missing alias delimiter:** `cstringFromVulnerability` glued aliases directly onto the CVE name without a `:` separator, causing collisions (for example, `CVE-2023-1` with alias `000` produced the same string as `CVE-2023-1000`).
- **Statement order dependence:** `sortVexStatements` only checked `Vulnerability.Name` and `Timestamp`. When a single CVE affected multiple packages, both had identical CVE names and defaulted timestamps, leaving the sort order tied and dependent on slice order.
- **In-place slice mutation:** `sortVexStatements` sorted the caller's slice directly, modifying the caller's slice backing array.

### Future Behavior:
- Map keys for hashes and identifiers are sorted before building the canonical string.
- A `:` separator is placed before appending aliases.
- `sortVexStatements` has deterministic tie-breakers (including vulnerability ID/aliases, status, justification, products/subcomponents, and statement ID).
- Product subcomponents are sorted alphabetically to prevent ordering mismatches within multi-package products.
- `calculateVexCanonicalHash` clones the statements slice before sorting it.

## Additional Information

- Kept a fast-path comparison on `Vulnerability.Name` at the start of statement sorting so 99% of comparisons between different CVEs return immediately without formatting full canonical strings.
- Gracefully handles invalid/empty statement timestamps by falling back to the document timestamp.

## How to Test

Run the new and existing unit tests in the `repositories` package:

```bash
go test ./repositories -v -run "TestCalculateVex|TestSortVex|TestCstring"
```

Or run all tests in the repository:

```bash
go test ./...
```

## Related issues/PRs:

* Resolved #769

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved VEX canonical hash consistency across equivalent document arrangements.
  * Prevented hashing and metadata enrichment from modifying original document data.
  * Added reliable timestamp fallback for missing or malformed statement timestamps.
  * Ensured deterministic ordering for statements, products, subcomponents, identifiers, aliases, and component data.

* **Tests**
  * Added comprehensive coverage for hash stability, ordering independence, collision handling, timestamp fallback, nil maps, and data immutability.

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -1670,11 +1670,8 @@ func calculateVexCanonicalHash(vexDoc v1beta1.VEX) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	cString := fmt.Sprintf("%d", ts.Unix())
-
-	cString += fmt.Sprintf(":%d", vexDoc.Version)
-
-	cString += fmt.Sprintf(":%s", vexDoc.Author)
+	docTsStr := ts.UTC().Format(time.RFC3339Nano)
+	cString := fmt.Sprintf(":%d:%s:%d:%d:%s", len(docTsStr), docTsStr, vexDoc.Version, len(vexDoc.Author), vexDoc.Author)
 
 	stmts := make([]v1beta1.Statement, len(vexDoc.Statements))
 	copy(stmts, vexDoc.Statements)
@@ -1685,15 +1682,16 @@ func calculateVexCanonicalHash(vexDoc v1beta1.VEX) (string, error) {
 		// 5a. Vulnerability
 		cString += cstringFromVulnerability(s.Vulnerability)
 		// 5b. Status + Justification
-		cString += fmt.Sprintf(":%s:%s", s.Status, s.Justification)
-		// 5c. Statement time, in unixtime. If it exists, if not the doc's (malformed timestamps fallback to doc time)
+		cString += fmt.Sprintf(":%d:%s:%d:%s", len(s.Status), s.Status, len(s.Justification), s.Justification)
+		// 5c. Statement time, in RFC3339Nano. If it exists, if not the doc's (malformed timestamps fallback to doc time)
 		stmtTs, err := time.Parse(time.RFC3339, s.Timestamp)
 		if err != nil {
 			stmtTs = ts
 		}
-		cString += fmt.Sprintf(":%d", stmtTs.Unix())
+		stmtTsStr := stmtTs.UTC().Format(time.RFC3339Nano)
+		cString += fmt.Sprintf(":%d:%s", len(stmtTsStr), stmtTsStr)
 		// 5d. Sorted product strings
-		cString += fmt.Sprintf(":%s", cstringFromProducts(s.Products))
+		cString += cstringFromProducts(s.Products)
 	}
 
 	h := sha256.New()
@@ -1714,13 +1712,17 @@ func cstringFromProducts(products []v1beta1.Product) string {
 			}
 			sort.Strings(subprods)
 			for _, scStr := range subprods {
-				prodString += scStr
+				prodString += fmt.Sprintf(":sub:%d:%s", len(scStr), scStr)
 			}
 		}
 		prods = append(prods, prodString)
 	}
 	sort.Strings(prods)
-	return strings.Join(prods, ":")
+	var result string
+	for _, pStr := range prods {
+		result += fmt.Sprintf(":prod:%d:%s", len(pStr), pStr)
+	}
+	return result
 }
 
 // sortVexStatements sorts a slice of VEX statements deterministically in place.
@@ -1798,26 +1800,29 @@ func sortVexStatements(stmts []v1beta1.Statement, documentTimestamp time.Time) {
 }
 
 func cstringFromVulnerability(v v1beta1.VexVulnerability) string {
-	cString := fmt.Sprintf(":%s:%s", v.ID, v.Name)
+	cString := fmt.Sprintf(":%d:%s:%d:%s", len(v.ID), v.ID, len(v.Name), v.Name)
 	if len(v.Aliases) > 0 {
 		list := make([]string, len(v.Aliases))
 		copy(list, v.Aliases)
 		sort.Strings(list)
-		cString += fmt.Sprintf(":%s", strings.Join(list, ":"))
+		for _, alias := range list {
+			cString += fmt.Sprintf(":%d:%s", len(alias), alias)
+		}
 	}
 	return cString
 }
 
 // cstringFromComponent renders one component into the canonicalization string. Both maps are
-// walked in sorted key order because Go randomizes map iteration, and this feeds the document's
-// canonical hash, which becomes its Metadata.ID: the same content has to render the same string
-// every time or the document changes identity between runs for no reason.
+// walked in sorted key order because Go randomizes map iteration, and fields are length-prefixed
+// and tagged to prevent delimiter collisions. This feeds the document's canonical hash, which
+// becomes its Metadata.ID: the same content has to render the same string every time or the
+// document changes identity between runs for no reason.
 //
 // Statements kubevuln writes itself carry only an ID (see createProductStructForImageAndPackage),
 // so both maps are empty today and the order never showed. An ingested OpenVEX component
 // routinely carries several identifiers and hashes, which is where it would.
 func cstringFromComponent(c v1beta1.Component) string {
-	s := fmt.Sprintf(":%s", c.ID)
+	s := fmt.Sprintf(":%d:%s", len(c.ID), c.ID)
 
 	algos := make([]string, 0, len(c.Hashes))
 	for algo := range c.Hashes {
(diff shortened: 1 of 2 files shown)


matthyx (dismissed):
Reviewed the diff and verified independently (checked out `11b0a4d`, ran `go build ./...`, `go vet ./repositories/...`, and the full `repositories` test suite — all green).

Findings:
- The clone-before-sort in `calculateVexCanonicalHash` correctly stops the caller's slice from being mutated (`TestCalculateVexCanonicalHash_DoesNotMutateCallerSlice` confirms).
- Sorted iteration over `Component.Hashes`/`Identifiers` and `Vulnerability.Aliases` removes the map-iteration non-determinism.
- The `:` separator before aliases in `cstringFromVulnerability` fixes the `CVE-2023-1`+`000` vs `CVE-2023-1000` collision.
- `sortVexStatements` now has full deterministic tie-breakers (vuln cstring, timestamp, status, justification, products, statement ID, remaining fields), so same-CVE/multi-package statements no longer depend on input order.
- Confirmed the earlier CodeRabbit finding about zero RFC3339 timestamps being wrongly treated as "missing" (via `IsZero()`) was already fixed in the last commit (`11b0a4d`), which switched to checking the parse error instead — `TestSortVexStatements_ZeroRFC3339Timestamp` and `TestCalculateVexCanonicalHash_ZeroRFC3339StatementTimestamp` cover this correctly.
- No behavior change to the document-level timestamp parsing/error path, and no other callers of `sortVexStatements`/`cstringFrom*` outside this hash computation.

No blockers. CI is green across the board. Approving.

matthyx:
rebase pls

h30s:
@matthyx ptal rebase done!

matthyx (dismissed):
Re-reviewed after the rebase + new commit `b747911` ("length-prefix canonical fields to prevent delimiter collisions"), since my prior approval on `11b0a4d` was dismissed by the push.

Verified independently: checked out `b747911`, ran `go build ./...`, `go vet ./repositories/...`, and the full `repositories` test suite — all green. All GitHub Actions checks (lint, CodeQL, build, DCO, GitGuardian) are green on this head too.

This commit specifically addresses the CodeRabbit finding from the prior revision (`a363ce4`) about ambiguous field concatenation letting distinct documents collide on the same `Metadata.ID`:
- `cstringFromVulnerability`, `cstringFromComponent`, and `cstringFromProducts` now length-prefix every dynamic field (`:<len>:<content>`) instead of relying on bare `:`/`@` separators, so content containing those characters (e.g. an alias `"A:B"`) can no longer be confused with two separate fields (`"A"`, `"B"`).
- Hashes vs. Identifiers are now tagged (`h:`/`i:`) so a hash algorithm and an identifier type with the same key/value no longer produce identical component strings.
- New tests (`TestCalculateVexCanonicalHash_AliasCollision`'s colon-alias case, `TestCstringFromVulnerability`'s colon-containing-alias case, `TestCalculateVexCanonicalHash_ComponentFieldCollision`, `TestCstringFromComponent_FieldCollision`) directly cover these cases and pass.
- The zero-RFC3339-timestamp fix from the previous revision is unchanged and still correctly falls back to the document timestamp only on parse error, not on `IsZero()`.

This length-prefixed encoding is injective by construction (each segment is unambiguously length-delimited before content), so it holds up under further review, not just the cases enumerated by the tests.

No blockers. Approving.

matthyx (approved):
Re-reviewed again after another push — head is now `8e6f814`, which also fixes the fractional-timestamp gap CodeRabbit flagged on the intermediate commit `f7f1e55`: timestamps were previously rendered via `.Unix()` (second precision), so two statements differing only in sub-second timestamps would canonicalize identically. Both the document and statement timestamps are now formatted with `.UTC().Format(time.RFC3339Nano)` and length-prefixed, with new tests (`TestCalculateVexCanonicalHash_FractionalTimestamp`, `TestSortVexStatements_FractionalTimestamp`) covering distinct fractional timestamps hashing differently and equivalent instants in different zone offsets hashing the same.

Verified independently against `8e6f814`: `go build ./...`, `go vet ./repositories/...`, and the full `repositories` test suite all pass. All GitHub Actions checks (Basic-Test, cross-platform build, lint, CodeQL, build-and-push-image, GitGuardian, DCO) are completed and green, plus FOSSA license/dependency/security checks.

Note on the CodeRabbit walkthrough's "Out of Scope Changes" warning about summary-annotation immutability: I diffed this head against the PR's merge base directly and confirmed `enrichSummaryManifestObject`/annotation-copy logic isn't touched by this PR — only `calculateVexCanonicalHash` and its helpers changed in `apiserver.go`. That warning appears to be describing code already on `main`, not a change in this diff, so it's not a real out-of-scope concern here.

No blockers. Approving.

PR #771, opened 2026-08-18
fix(repositories): make GetSBOM's version-freshness check symmetric

### LINKED ISSUE #768: GetSBOM's version-freshness check only catches an older manifest, not a newer one, and mislabels the version it returns
`GetSBOM`'s freshness check only discards a stored SBOM when it is **older** than the version the caller asked for. A manifest written by a **newer** scanner passes straight through as "fresh" — and gets handed back mislabeled as the caller's own (older) version.

```go
// repositories/apiserver.go
func (a *APIServerStore) GetSBOM(ctx context.Context, name, SBOMCreatorVersion string) (domain.SBOM, error) {
	...
	// discard the manifest if it was created by an older version of the scanner
	if semver.Compare(manifest.Spec.Metadata.Tool.Version, SBOMCreatorVersion) == -1 {
		...
		return result, domain.ErrOutdatedSBOM
	}
	result := domain.SBOM{
		Name:               name,
		Annotations:        manifest.Annotations,
		Labels:             manifest.Labels,
		SBOMCreatorVersion: SBOMCreatorVersion, // <- the caller's requested version, not manifest.Spec.Metadata.Tool.Version
		Content:            &manifest.Spec.Syft,
	}
	...
	return result, nil
}
```

`semver.Compare(a, b) == -1` means "`a` < `b`", i.e. the stored manifest's tool version is older than what the caller wants. That's the only branch that discards. When the stored manifest is *newer* than the caller's version (`semver.Compare == 1`), or equal, the function falls through to the success path — and stamps the returned `SBOMCreatorVersion` field with the caller's requested value, not `manifest.Spec.Metadata.Tool.Version`, the version the content was actually produced by.

### Contrast with `GetCVE`

`GetCVE`, right above it in the same file, does the equivalent check with exact equality across all three of its version fields, discarding on a mismatch in *either* direction:

```go
// repositories/apiserver.go:483
if manifest.Annotations[helpersv1.ToolVersionMetadataKey] != SBOMCreatorVersion ||
	manifest.Spec.Metadata.Tool.Version != CVEScannerVersion ||
	manifest.Spec.Metadata.Tool.DatabaseVersion != CVEDBVersion {
	// discard as outdated, either direction
```

`GetSBOM` is the only one of the two freshness gates in this file that only checks one direction.

### Why this is reachable, not hypothetical

`SBOMSyft` objects are shared cluster state, keyed by image slug (not by scanner version — `StoreSBOM` names the object `sbom.Name`, the slug), and read by whichever kubevuln replica handles the next request for that image. During a normal rolling deployment (or a rollback), replicas at version N and N+1 run concurrently for a window and both can serve the same image slug:

1. Replica N+1 scans an image first, and `StoreSBOM` writes an `SBOMSyft` object with `Spec.Metadata.Tool.Version = N+1`.
2. Replica N (still draining, or a rolled-back deployment sitting behind it) requests the same slug's SBOM. `semver.Compare(N+1, N) == 1`, not `-1`, so replica N's `GetSBOM` treats the object as fresh and returns its `Content` — produced by a scanner build replica N was never validated against — under `SBOMCreatorVersion: N` (its own, wrong, older value).

The one-directional check exists specifically to stop a binary from consuming SBOM content it isn't known-compatible with (that's the whole point of the version gate, matching `GetCVE`'s symmetric one) — but as written it only defends against the backward case, leaving the forward case, which a rolling deploy or rollback produces routinely, unguarded.

### Downstream effect of the mislabeling

Independent of the compatibility question, the field itself is wrong data. `filterSBOM` (`core/services/scan.go:986`) copies `sbom.SBOMCreatorVersion` straight into the filtered SBOM (SBOM') it builds, which `StoreSBOM` then persists as *that* object's own `Spec.Metadata.Tool.Version` — so the corruption doesn't stay in-memory, it gets written back into another cluster object under a caller-supplied value that has nothing to do with what actually produced it. `GrypeAdapter.ScanSBOM` (`adapters/v1/grype.go:517`) does the same into the returned `CVEManifest.SBOMCreatorVersion`, which is part of what gets submitted to the platform backend (`SubmitCVE`) — so the reported provenance of a scan result can be wrong too.

### Root cause

`GetSBOM`'s freshness check was written as a one-directional "not older than" test instead of an equality test, unlike its `GetCVE` sibling, and the success path unconditionally returns the caller's requested version instead of what it actually read.

### Affected components

- `repositories/apiserver.go`: `GetSBOM`
- `core/services/scan.go`: `filterSBOM` (propagates the mislabeled field into a second persisted object)
- `adapters/v1/grype.go`: `ScanSBOM` (propagates it into the submitted `CVEManifest`)

### Evidence of the gap

`repositories/apiserver_test.go`'s `TestAPIServerStore_GetSBOM` has one version-mismatch case ("SBOMCreatorVersion mismatch"), and it only exercises the caller-wants-newer / manifest-is-older direction (requested `v1.1.0` vs. stored `v1.0.0`). There is no case for the reverse (stored newer than requested), which is exactly the branch that's wrong.

### Acceptance criteria

- `GetSBOM` discards (returns `domain.ErrOutdatedSBOM`, matching the existing older-manifest branch's contract) when the stored manifest's tool version differs from the requested version in *either* direction, not just when it's older.
- On the success path, the returned `domain.SBOM.SBOMCreatorVersion` reflects `manifest.Spec.Metadata.Tool.Version` (what was actually read), never the caller's requested value.
- New test case: manifest newer than the requested version is discarded as outdated (currently unmarked and, per the code above, wrongly accepted).
- New test/assertion: the returned `SBOMCreatorVersion` on a successful `GetSBOM` call matches the stored manifest's actual tool version, not the input parameter.

### Suggested direction

Mirror `GetCVE`'s pattern: replace the `semver.Compare(...) == -1` check with an exact-match discard (`semver.Compare(...) != 0`, or a direct string comparison if exact string equality is intended rather than semantic version equality — worth deciding explicitly, since `GetCVE` uses plain `!=` on strings rather than `semver.Compare`, so there may already be an implicit "these two checks should agree on comparison semantics" expectation worth reconciling as part of the fix). Stop overwriting `SBOMCreatorVersion` with the input parameter in the success branch; return the manifest's own value instead, the same as the existing outdated-branch code already does.

version: main (ff11524)

bhuvan-somisetty:
## Problem

`GetSBOM`'s version-freshness check only discarded a stored SBOM when it was **older** than the version the caller asked for. A manifest written by a **newer** scanner version passed straight through as "fresh" and was handed back mislabeled with the caller's own (older, wrong) requested version instead of the manifest's actual one.

## Root cause

```go
if semver.Compare(manifest.Spec.Metadata.Tool.Version, SBOMCreatorVersion) == -1 {
    // discard as outdated
}
result := domain.SBOM{
    ...
    SBOMCreatorVersion: SBOMCreatorVersion, // the caller's requested version, not the manifest's actual one
    ...
}
```

`semver.Compare(a, b) == -1` only catches "manifest older than requested." `GetCVE`, right above it in the same file, does the equivalent check with exact equality across all its version fields, discarding a mismatch in either direction — `GetSBOM` was the only one-directional gate of the two.

`SBOMSyft` objects are shared cluster state, keyed by image slug rather than by scanner version, and read by whichever kubevuln replica handles the next request for that slug. During a rolling deploy/rollback, replicas at different versions run concurrently and can serve the same slug, so an older replica could read back a newer replica's SBOM as if it were its own — the exact case the check exists to prevent, just missed in one direction.

The mislabeled version wasn't just cosmetic either: `filterSBOM` (`core/services/scan.go`) copies it into the filtered SBOM it builds, which then gets persisted as *that* object's own `Tool.Version`, and `GrypeAdapter.ScanSBOM` copies it into the `CVEManifest.SBOMCreatorVersion` that's submitted to the platform backend.

## Fix

- Changed the comparison to `semver.Compare(...) != 0`, so a mismatch in either direction is discarded, matching `GetCVE`'s pattern.
- The success path now always returns `manifest.Spec.Metadata.Tool.Version` (what was actually read) instead of echoing back the caller's requested value — `semver.Compare` treats build-metadata differences as equal, so the two strings can still legitimately differ even when the check passes.

Both changes are confined to `GetSBOM`; the existing outdated-manifest branch already returned the manifest's own version, so this brings the success branch in line with it.

## Testing

Extended `TestAPIServerStore_GetSBOM` in `repositories/apiserver_test.go`:
- New case: manifest **newer** than the requested version is now discarded as outdated (previously silently accepted — this is the regression test for the bug).
- New case: matching versions return the SBOM as before.
- Every version-comparison case now also asserts the returned `SBOMCreatorVersion` reflects the manifest's own recorded value rather than the caller's input, covering the mislabeling half of the bug independently of the discard decision.

Existing callers (`core/services/scan_test.go`) round-trip `GetSBOM` using the same version on write and read, so they're unaffected by the now-symmetric check.

Fixes #768

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -1780,9 +1780,15 @@ func (a *APIServerStore) GetSBOM(ctx context.Context, name, SBOMCreatorVersion s
 	case err != nil:
 		return domain.SBOM{}, fmt.Errorf("failed to get SBOM from apiserver: %w", err)
 	}
-	// discard the manifest if it was created by an older version of the scanner
-	if semver.Compare(manifest.Spec.Metadata.Tool.Version, SBOMCreatorVersion) == -1 {
-		logger.L().Debug("discarding SBOM with outdated scanner version",
+	// Discard the manifest if it was created by a different version of the scanner, in
+	// either direction. A manifest older than SBOMCreatorVersion may predate a schema or
+	// content change the caller now relies on; a manifest newer than SBOMCreatorVersion may
+	// carry one the caller isn't known to handle yet (e.g. one replica of a rolling
+	// deployment/rollback writes it, and a different-versioned replica serving the same
+	// image slug reads it back next). Comparing only for "older" let the newer case through
+	// as if it were fresh (see #768).
+	if semver.Compare(manifest.Spec.Metadata.Tool.Version, SBOMCreatorVersion) != 0 {
+		logger.L().Debug("discarding SBOM with mismatched scanner version",
 			helpers.String("name", name),
 			helpers.String("manifest scanner version", manifest.Spec.Metadata.Tool.Version),
 			helpers.String("wanted scanner version", SBOMCreatorVersion))
@@ -1802,7 +1808,10 @@ func (a *APIServerStore) GetSBOM(ctx context.Context, name, SBOMCreatorVersion s
 		Name:               name,
 		Annotations:        manifest.Annotations,
 		Labels:             manifest.Labels,
-		SBOMCreatorVersion: SBOMCreatorVersion,
+		// The manifest's own recorded version, not the caller's requested SBOMCreatorVersion:
+		// semver.Compare treats build-metadata differences as equal, so the two can still
+		// differ as strings even when the check above passes.
+		SBOMCreatorVersion: manifest.Spec.Metadata.Tool.Version,
 		Content:            &manifest.Spec.Syft,
 	}
 	if status, ok := manifest.Annotations[helpersv1.StatusMetadataKey]; ok {
--- a/repositories/apiserver_test.go
+++ b/repositories/apiserver_test.go
@@ -164,11 +164,13 @@ func TestAPIServerStore_GetSBOM(t *testing.T) {
 		SBOMCreatorVersion string
 	}
 	tests := []struct {
-		name          string
-		args          args
-		sbom          domain.SBOM
-		wantErr       error
-		wantEmptySBOM bool
+		name                    string
+		args                    args
+		sbom                    domain.SBOM
+		wantErr                 error
+		wantEmptySBOM           bool
+		wantSBOMCreatorVersion  string
+		checkSBOMCreatorVersion bool
 	}{
 		{
 			"valid SBOM is retrieved",
@@ -181,6 +183,8 @@ func TestAPIServerStore_GetSBOM(t *testing.T) {
 				Content: &v1beta1.SyftDocument{},
 			},
 			nil,
+			false,
+			"",
 			false,
 		},
 		{
@@ -195,9 +199,13 @@ func TestAPIServerStore_GetSBOM(t *testing.T) {
 			},
 			nil,
 			false,
+			"",
+			false,
 		},
 		{
-			"SBOMCreatorVersion mismatch",
+			// The stored manifest predates the version the caller wants: content it was
+			// never validated against would otherwise be served as if it were current.
+			"SBOMCreatorVersion mismatch, manifest older than requested",
 			args{
 				ctx:                context.TODO(),
 				name:               name,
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Reviewed the logic and tests in detail.

- The fix makes `GetSBOM`'s freshness check symmetric (`!= 0` instead of `== -1`), matching `GetCVE`'s existing exact-match pattern right above it in the same file — good consistency.
- Verified `StoreSBOM` maps `domain.SBOM.SBOMCreatorVersion` → `manifest.Spec.Metadata.Tool.Version`, so the new test cases (`newer than requested`, `match`) correctly exercise the changed comparison and the corrected success-path return value.
- Traced the only production caller, `ScanService.getSBOM` (`core/services/scan.go`), and confirmed it already handles `domain.ErrOutdatedSBOM` correctly (including the `TooLarge`+delete path), so the newly-discarded "newer manifest" case is handled safely.
- Confirmed `filterSBOM` copies `sbom.SBOMCreatorVersion` through to the filtered SBOM, so the mislabeling fix (returning the manifest's actual version instead of echoing the caller's request) correctly propagates downstream, as described in the PR body.
- `core/services/scan_test.go` uses a `mockSBOMRepository`, bypassing the real `apiserver.go` comparison logic entirely, so it's correctly unaffected by this change as claimed.
- All CI checks (build, lint, CodeQL, security) are green.

No blockers. LGTM.

PR #773, opened 2026-08-18
fix(repositories): treat a dropped SecurityException as an incomplete set

### LINKED ISSUE #772: a SecurityException that fails to convert is reported as a complete exception set
a SecurityException that fails to convert is dropped and the list still comes back with a nil error, so the caller is told the exception set is complete when it is one entry short. the missing suppression then reads as a deletion.

```go
for i := range seList.Items {
	var se sev1beta1.SecurityException
	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(seList.Items[i].Object, &se); err != nil {
		logger.L().Ctx(ctx).Warning("failed to convert SecurityException", helpers.Error(err))
		continue
	}
	exceptions = append(exceptions, se)
}

if a.securityExceptionListCache != nil {
	a.trySetSecurityExceptionCache(cacheKey, seenGeneration, exceptions)
}
return exceptions, nil
```

that nil is what the whole degraded path keys on. `GetCVEExceptions` only degrades when the list call itself errored:

```go
seList, cseList, crdErr := a.securityExceptionRepo.GetSecurityExceptions(ctx, namespace)
if crdErr != nil {
	cacheable = false
	degraded = true
}
```

`applyExceptionsToManifest` then returns `!degraded`, i.e. complete, and `reconcileCachedCVE` acts on it:

```go
if exceptionsComplete || !hasRemovals {
	... StoreCVE(filteredCve) / StoreCVESummary(...) ...
	if publishVEX && exceptionsComplete {
		s.storeVEX(ctx, filteredCve, filteredCvep, false, imageSlug)
	}
}
```

so a dropped exception can unsuppress a CVE in the stored manifest and publish a VEX document asserting it. that is the outcome the degraded mechanism exists to prevent, and its own comment says so: "a transient SecurityException CRD list failure must not look like a deletion and wipe suppression from the stored manifest."

the partial result is also cached, which `GetSecurityExceptions` documents as something that does not happen:

> A List() failure is never cached, so a transient apiserver hiccup self-heals on the next call instead of being pinned for the TTL.

a conversion failure is cached, so the short set is served for the TTL rather than being retried.

`listClusterSecurityExceptions` has the same loop with the same behaviour.

what makes this cheap to fix is that the shape already exists. `GetSecurityExceptions` joins its errors and returns the lists alongside them, and `GetCVEExceptions` uses those lists even when the error is set, so returning the successfully-converted entries together with an error degrades the set without losing the suppressions that do work.

how often conversion actually fails is the open part: it takes an object whose field types do not match the typed struct, so a CRD schema change or a version skew rather than an everyday event. the consequence is what makes it worth reporting, not the frequency.

version: main (3f711ab)

AnshulPatil2005:
## Overview

A SecurityException that fails to convert is skipped with a warning and the list comes back with a nil error, so the caller is told the exception set is complete when it is one entry short. The missing suppression then reads as a deletion.

That flag is what decides whether removals are persisted: `reconcileCachedCVE` writes them and republishes VEX only when the set is complete, so a conversion failure could quietly unsuppress a CVE in the stored manifest and publish a VEX document asserting it. The partial result was cached too, so the short set was served for the TTL.

## Additional Information

this is the outcome the degraded mechanism exists to prevent, in its own words: "a transient SecurityException CRD list failure must not look like a deletion and wipe suppression from the stored manifest." a failed List() is handled; a failed conversion produces the same incomplete set and was not.

the caching half contradicts a property `GetSecurityExceptions` documents for itself, that "a List() failure is never cached, so a transient apiserver hiccup self-heals on the next call instead of being pinned for the TTL".

the current behaviour is deliberate, and `GetSecurityExceptions`' own comment says so: a conversion failure is "only logged and skipped, since that's a malformed single object rather than a listing-wide failure, and skipping it doesn't risk masking a systemic API problem the way swallowing a List() error would".

that reasoning is about whether to abandon the whole list, and this does not abandon it. the caller still receives and applies every entry that converted, so nothing is swallowed and no good suppression is lost. what changes is only whether the set may be called complete, and a skipped exception leaves it short for the same reason a failed List() does. the comment is updated in a second commit so it no longer describes behaviour the code no longer has.

the fix reuses the shape that already exists rather than adding one. `GetSecurityExceptions` joins its errors and returns the lists alongside them, and `GetCVEExceptions` applies those lists even when the error is set, so returning the entries that did convert together with an error marks the set degraded while keeping the suppressions that work. losing them as well would drop good suppressions to report a bad one.

the error names the object, namespace and name, so which exception was dropped is in the message rather than only in the warning above it.

`listClusterSecurityExceptions` had the same loop and gets the same treatment.

how often conversion fails is the open part: it takes an object whose field types do not match the typed struct, so a CRD schema change or version skew rather than an everyday event. the consequence is what makes it worth fixing, not the frequency.

## How to Test

`go build ./... && go vet ./... && go test ./repositories/ ./core/... -count=1`

the existing SecurityException tests pass unchanged, including `_PropagatesListErrors`, `_NoErrorOnSuccess`, `_CachesAcrossWorkloads`, `_DoesNotCacheListFailures` and `_HonorsCallerCancellation`, so the list-failure and caching behaviour either side of this is the same.

`TestAPIServerStore_GetSecurityExceptions_ConversionFailureDegrades` is new. it lists one convertible exception and one whose `spec.vulnerabilities` is a string where the typed struct wants a list, then asserts three things: an error comes back, the one that converted is still returned, and a second call goes to the apiserver again rather than being served a cached short set.

dropping the new early return fails it on the first of those.

## Related issues/PRs:

* Resolved #772

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of security exception conversion failures by preserving valid results, reporting errors clearly, and retrying retrieval when results are incomplete.
  * Prevented incomplete security data from being cached.
  * Improved consistency when enriching summaries and labels.
  * Made VEX output hashing deterministic for more reliable comparisons.
  * SBOM retrieval now clearly rejects incompatible scanner versions and reports the stored manifest version.

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -379,10 +379,16 @@ func unstructuredFromEvent(obj interface{}) *unstructured.Unstructured {
 // GetSecurityExceptions lists both namespaced SecurityExceptions and cluster-scoped
 // ClusterSecurityExceptions. A List() failure is returned as an error rather than only
 // logged: the caller (BackendAdapter.GetCVEExceptions) relies on a non-nil error here to
-// avoid caching a degraded, incomplete exception set for exceptionsCacheTTL — see #477. A
-// conversion failure on an individual item is still only logged and skipped, since that's a
-// malformed single object rather than a listing-wide failure, and skipping it doesn't risk
-// masking a systemic API problem the way swallowing a List() error would.
+// avoid caching a degraded, incomplete exception set for exceptionsCacheTTL — see #477.
+//
+// A conversion failure on an individual item is reported the same way, while the items that
+// did convert are still returned. It used to be only logged and skipped, on the grounds that
+// a malformed single object is not a listing-wide failure. That holds for whether to abandon
+// the list, and this does not abandon it: the caller still receives and applies everything
+// that converted. What it cannot do is call the set complete, because the flag derived from
+// this error is what decides whether a suppression that is now missing counts as a deletion
+// (see reconcileCachedCVE) — and a skipped exception is missing for exactly the same reason
+// a failed List() leaves the set short.
 //
 // Both lists are served from securityExceptionListCache when available and fresh: the raw
 // List() results are the same for every workload in a given namespace (SecurityException) or
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -437,13 +437,25 @@ func (a *APIServerStore) listSecurityExceptions(ctx, listCtx context.Context, na
 	}
 
 	var exceptions []sev1beta1.SecurityException
+	var convErrs []error
 	for i := range seList.Items {
 		var se sev1beta1.SecurityException
 		if err := runtime.DefaultUnstructuredConverter.FromUnstructured(seList.Items[i].Object, &se); err != nil {
-			logger.L().Ctx(ctx).Warning("failed to convert SecurityException", helpers.Error(err))
+			logger.L().Ctx(ctx).Warning("failed to convert SecurityException", helpers.Error(err),
+				helpers.String("namespace", namespace), helpers.String("name", seList.Items[i].GetName()))
+			convErrs = append(convErrs, fmt.Errorf("converting SecurityException %s/%s: %w", namespace, seList.Items[i].GetName(), err))
 			continue
 		}
 		exceptions = append(exceptions, se)
+	}
+
+	// A dropped exception leaves the set incomplete just as a failed List() does, so it is
+	// reported the same way and not cached. Returning it as complete would let the caller
+	// persist the missing suppressions as removals and republish VEX from a set that is
+	// quietly short an entry; caching it would pin that for the TTL rather than letting the
+	// next call self-heal.
+	if len(convErrs) > 0 {
+		return exceptions, stderrors.Join(convErrs...)
 	}
 
 	if a.securityExceptionListCache != nil {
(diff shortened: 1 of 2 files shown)


matthyx (dismissed):
Verified the fix against the surrounding call chain, not just the diff:

- `BackendAdapter.GetCVEExceptions` (adapters/v1/backend.go) does apply `seList`/`cseList` even when `crdErr != nil` (the `len(seList) > 0 || len(cseList) > 0` block runs unconditionally before the `degraded`/`cacheable=false` check), so returning the converted subset alongside a joined error is handled correctly by the caller rather than silently dropped.
- The early return on `convErrs` skips `trySetSecurityExceptionCache` the same way the existing List()-failure path does, and `beginSecurityExceptionCacheRefresh` is just a generation snapshot with no matching "end" call to worry about leaking — so no cache/generation state is left dangling.
- The stale "malformed single object... only logged and skipped" doc comment was the only copy of that claim in the repo (checked `SecurityExceptionRepository` interface and callers too), so the comment update in the second commit doesn't leave a contradicting description anywhere else.
- New test (`TestAPIServerStore_GetSecurityExceptions_ConversionFailureDegrades`) exercises the actual regression: error returned, partial result preserved, and a second call re-lists instead of serving a cached short set.
- CI is green across build/vet/tests/lint/CodeQL/GitGuardian.

Only a non-blocking nitpick from CodeRabbit to extend the new test to also cover the cluster-scoped conversion-failure path — worth doing but doesn't affect correctness of what's here. No blockers found.

matthyx:
rebase pls

AnshulPatil2005:
> rebase pls

done. #787 and #793 too, since they were behind as well.

matthyx (dismissed):
Re-reviewed after the rebase (matthyx asked for it, Anshul did it in commit 1483e471). Diff content is unchanged from what I approved before — same two commits, same logic, only the diff context shifted because the base moved forward. Re-approving since the rebase dismissed my prior review.

One thing to flag, not a code issue: the `lint` check on the latest commit (1483e471) is currently failing, but the log shows it's a `golangci-lint` timeout during tool download/install ("Timeout exceeded: try increasing it by passing --timeout option"), not an actual lint violation — no findings were reported before it timed out. Worth re-running that job before merge; doesn't reflect a problem with this PR's code.

matthyx:
rebase pls

matthyx (approved):
Re-reviewed after the second rebase (head now 16f812c8, onto a newer main). Pulled the PR-level diff directly (base→head, not a commit-to-commit diff) and it's byte-for-byte the same two-commit change I approved twice already: the `convErrs`/early-return logic in `listSecurityExceptions`/`listClusterSecurityExceptions` plus the updated doc comment, only the line numbers moved (main advanced in between). Re-approving since the force-push dismissed the prior approval again.

Note: CodeRabbit's latest walkthrough on this commit lists unrelated-looking changes (map-copying on enrichment, deterministic VEX hashing, SBOM scanner-version checks) and an "Out of Scope Changes" warning. That's an artifact of it diffing between two specific commit SHAs from different force-pushes rather than the PR's actual base→head diff — those items are pre-existing code that landed on `main` from other merged PRs between rebases, now present on this branch after rebasing onto the newer main, not new content this PR introduces. The PR's own base→head diff (`get_diff`) confirms only `repositories/apiserver.go` and `repositories/apiserver_test.go` changed, matching what's described above.

All 8 checks are green now, including `lint` (the earlier timeout was a one-off). `mergeable_state` shows "unknown" right after the push — that's GitHub still computing it post-force-push, should settle on its own. No blockers.

PR #775, opened 2026-08-18
fix(repositories): canonicalize a VEX component's maps in sorted order

### LINKED ISSUE #774: calculateVexCanonicalHash is not canonical: component hashes and identifiers are walked in map order
`calculateVexCanonicalHash` is not canonical for a component that carries more than one hash or identifier. the same document hashes differently between runs, and that hash is the document's `Metadata.ID`.

`cstringFromComponent` walks both maps with a bare range:

```go
func cstringFromComponent(c v1beta1.Component) string {
	s := fmt.Sprintf(":%s", c.ID)

	for algo, val := range c.Hashes {
		s += fmt.Sprintf(":%s@%s", algo, val)
	}

	for t, id := range c.Identifiers {
		s += fmt.Sprintf(":%s@%s", t, id)
	}

	return s
}
```

go randomizes map iteration, so the canonicalization string is assembled in a different order each time. hashing one unchanged document two hundred times:

```
component with 3 identifiers and 3 hashes  ->  9 distinct hashes
component carrying only an ID              ->  1 distinct hash
```

both call sites assign the result to the document's identity:

```go
calculatedId, err := calculateVexCanonicalHash(vexDoc)
if err != nil {
	return err
}
vexDoc.Metadata.ID = calculatedId
```

so a document whose content did not change can be written back under a different ID.

it does not bite today, and that is the whole reason it is easy to miss. `createProductStructForImageAndPackage` builds products with an ID and nothing else:

```go
product := v1beta1.Product{
	Component: v1beta1.Component{
		ID: imageField,
	},
}
```

both maps are empty for everything kubevuln writes, one entry orders the same as none, and the second row above is the shape currently in storage.

where it starts mattering is external VEX. an ingested OpenVEX component routinely carries several identifiers (purl, cpe22, cpe23) and several hashes, and `updateVEX` recalculates the hash over every statement in the container, not only the ones kubevuln authored. so the first foreign statement with two identifiers makes the container's ID unstable.

sorting the keys is enough, and it leaves the empty case rendering the exact string it renders now, so nothing already stored rehashes.

for the record I looked at this earlier in a different context and set it aside as unreachable. that was right about the data in storage and wrong about it not being worth closing: the fix costs nothing today and the reachable case is the feature #387 is about.

version: main (3f711ab)

AnshulPatil2005:
## Overview

`calculateVexCanonicalHash` becomes the document's `Metadata.ID`, so the same content has to render the same hash every time. `cstringFromComponent` walked `Component.Hashes` and `Identifiers` with a bare range, and Go randomizes map iteration, so a component carrying more than one of either hashed differently between runs and the document changed identity for no reason.

Hashing one unchanged document two hundred times: a component with three identifiers and three hashes produced **nine** distinct hashes. A component carrying only an ID produced one.

## Additional Information

it does not bite today. `createProductStructForImageAndPackage` builds products with an ID and nothing else, so both maps are empty for everything kubevuln writes, and one entry orders the same as none. that is the shape currently in storage.

where it starts mattering is external VEX. an ingested OpenVEX component routinely carries several identifiers (purl, cpe22, cpe23) and several hashes, and `updateVEX` recalculates the hash over every statement in the container rather than only the ones kubevuln authored, so one foreign statement with two identifiers is enough to make the container's ID unstable.

sorting the keys leaves the empty case rendering the exact string it rendered before, so nothing already stored rehashes and no existing document changes ID on the next update. that is what the second subtest pins.

the keys are converted to `string` for sorting because `Algorithm` and `IdentifierType` are distinct named string types; the values are read back through the original key types so nothing about the rendered pairs changes.

`sort` was already imported here, so no new dependency.

## How to Test

`go build ./... && go vet ./... && go test ./repositories/ -count=1`

every existing test passes unchanged, including the VEX ones that assert on stored documents, which is what says the shape already in storage hashes as it did.

`TestCalculateVexCanonicalHash_StableAcrossMapOrder` is new and has two subtests. The first hashes a component with three identifiers and three hashes two hundred times and requires exactly one distinct result. The second does the same for a component carrying only an ID, pinning that the shape kubevuln writes today is unaffected.

putting the bare ranges back fails the first subtest and leaves the second passing, which is the gap.

## Related issues/PRs:

* Resolved #774

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -1750,15 +1750,33 @@ func cstringFromVulnerability(v v1beta1.VexVulnerability) string {
 	return cString
 }
 
+// cstringFromComponent renders one component into the canonicalization string. Both maps are
+// walked in sorted key order because Go randomizes map iteration, and this feeds the document's
+// canonical hash, which becomes its Metadata.ID: the same content has to render the same string
+// every time or the document changes identity between runs for no reason.
+//
+// Statements kubevuln writes itself carry only an ID (see createProductStructForImageAndPackage),
+// so both maps are empty today and the order never showed. An ingested OpenVEX component
+// routinely carries several identifiers and hashes, which is where it would.
 func cstringFromComponent(c v1beta1.Component) string {
 	s := fmt.Sprintf(":%s", c.ID)
 
-	for algo, val := range c.Hashes {
-		s += fmt.Sprintf(":%s@%s", algo, val)
+	algos := make([]string, 0, len(c.Hashes))
+	for algo := range c.Hashes {
+		algos = append(algos, string(algo))
+	}
+	sort.Strings(algos)
+	for _, algo := range algos {
+		s += fmt.Sprintf(":%s@%s", algo, c.Hashes[v1beta1.Algorithm(algo)])
 	}
 
-	for t, id := range c.Identifiers {
-		s += fmt.Sprintf(":%s@%s", t, id)
+	types := make([]string, 0, len(c.Identifiers))
+	for t := range c.Identifiers {
+		types = append(types, string(t))
+	}
+	sort.Strings(types)
+	for _, t := range types {
+		s += fmt.Sprintf(":%s@%s", t, c.Identifiers[v1beta1.IdentifierType(t)])
 	}
 
 	return s
--- a/repositories/apiserver_test.go
+++ b/repositories/apiserver_test.go
@@ -2488,9 +2488,6 @@ func TestAPIServerStore_ListClusterSecurityExceptions_DoesNotRepopulateAfterRaci
 	assert.False(t, ok, "a List() that raced an invalidation must not repopulate the cache with its now-stale result")
 }
 
-
-
-
 // TestAPIServerStore_TrySetSecurityExceptionCache_GenerationMismatchSkipsWrite unit-tests the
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Reviewed the diff and the surrounding `calculateVexCanonicalHash` / `cstringFromComponent` code.

- The fix is correct: `c.Hashes` and `c.Identifiers` are both maps, Go randomizes map iteration order, and the rendered order feeds directly into the canonical string that becomes `Metadata.ID` via `sha256`. Sorting the keys before appending makes the output deterministic regardless of insertion order.
- The empty-map case (what `createProductStructForImageAndPackage` actually produces today, since it only sets `ID`) is unaffected — sorting zero or one entries renders the same string as before, so nothing already in storage rehashes.
- `sort` was already imported in this file, no new dependency.
- New test `TestCalculateVexCanonicalHash_StableAcrossMapOrder` hashes the same document 200 times for both a multi-key component and an ID-only component, asserting exactly one distinct hash in each case. That's a solid regression guard for map-order flakiness, which is otherwise hard to catch (~1/9 runs would differ before this fix given 3 hash + 3 identifier entries).
- CI is green (build, vet, lint, tests, CodeQL, security scans all pass), and `go.mod`/deps are untouched.

No blockers. Approving.

Issue #763, opened 2026-08-18
ErrScannerCrashed has no producer, so the sidecar crash-retry and OOM reporting never run


AnshulPatil2005:
`ErrScannerCrashed` has no producer. nothing outside tests ever returns it, so everything keyed on it is unreachable, including the crash-retry path and the OOM reporting the sidecar architecture exists for.

the only code that turns a gRPC failure into one of these sentinels is the client:

```go
st, ok := status.FromError(err)
if ok && st.Code() == codes.Unavailable {
	return nil, fmt.Errorf("%w: %v", ErrScannerUnavailable, err)
}
```

a sidecar OOM-killed mid-scan drops the connection, which is `codes.Unavailable`, so what comes back is `ErrScannerUnavailable`. `ErrScannerCrashed` is produced only by the mock in `adapters/v1/sidecar_test.go`, and `pkg/sbomscanner/v1/client_test.go` asserts five times over that the real client does not produce it:

```go
assert.NotErrorIs(t, err, ErrScannerCrashed)
```

so this branch never runs:

```go
result, err := s.client.CreateSBOM(ctx, req)
if err != nil {
	if errors.Is(err, sbomscanner.ErrScannerCrashed) {
		return s.handleCrash(ctx, name, imageID, imageTag, options, domainSBOM, err)
	}
	return domainSBOM, err
}
```

and with it, everything `handleCrash` reaches: `maxCrashRetries`, the `retryCount` map, `crashRetry`, `evictStaleRetriesLocked`, `retryCountTTL`, and the `TooLarge` plus `domain.ReasonScannerOOM` plus memory-limit annotation it writes on exhaustion. `sidecar.go:214` is the only place that sets `ReasonScannerOOM`, so the readers of it are dead too: `scan.go:1243`, and `classifySBOMStatusWithAnnotation`'s `ReasonScannerOOMKilled` cases. So is `classifySBOMError`'s first branch.

what actually happens when the sidecar is OOM-killed during a scan: `ErrScannerUnavailable` reaches `classifySBOMError`, which returns `ReasonSBOMGenerationFailed`. no retry, and the report says the SBOM generation failed rather than that the scanner was killed for memory, which is the one thing an operator needs to know to raise the sidecar's limit.

worth saying this was maintained, not abandoned: #473 fixed a leak in `retryCount` by adding `retryCountTTL`, on a map nothing writes to.

two ways out and this is a call for whoever owns the sidecar design, so I have not guessed at it:

[...]

PR #777, opened 2026-08-18
docs(architecture): document the sidecar SBOM scanner and the missing packages

### LINKED ISSUE #776: ARCHITECTURE.md does not mention the sidecar SBOM scanner at all
`docs/ARCHITECTURE.md` does not mention the sidecar SBOM scanner anywhere. searching its 622 lines for `sidecar` or `grpc` returns nothing, and the only hits for `scanner` are `mockcve.go`'s comment and the `CVEScanner` port name.

so a reader comes away with one binary and one SBOM path, when there are two: `SyftAdapter` in-process, and `SidecarSBOMAdapter` talking to `cmd/sbom-scanner` over gRPC on a unix socket when `SBOM_SCANNER_SOCKET` is set. that is not a detail, it is which container Syft's memory is charged to and which one the kernel kills when an image is too big.

it also costs contributors time in a specific way: the two paths are parallel implementations of the same behaviour, so a change to one usually needs the same change to the other. that has produced a steady run of bugs where only one side got updated (#585, #587, #601, #675, #686 among them), and nothing in the architecture doc tells a newcomer the second side is there to update.

the directory structure listing is out of date too:

```
├── internal/
│   └── tools/                  # Internal utilities
```

`internal/` also holds `metrics`, `registryauth`, `syftmeta`, `safefetch` and `vexdoc`. there is no `pkg/` in the listing at all, so `pkg/sbomscanner/v1` and `pkg/securityexception/v1beta1` are absent, and `cmd/` shows only `http` and `cli`, not `sbom-scanner`. `adapters/v1` lists `syft.go` without `sidecar.go` beside it.

worth documenting alongside the mode itself:

- startup is fail-soft. the client health-checks the sidecar, bounded by `scannerReadinessTimeout`, and falls back to in-process rather than refusing to start, so a deployment can be running in a different mode than its config suggests. `/v1/diagnostics` reports which, as `scanMode`.
- concurrency differs on purpose. the in-process path serialises pulls behind `pullMutex`; the sidecar deliberately does not, and says so in a comment citing #473. someone who assumes the two match will get one of them wrong.

version: main (3f711ab)

AnshulPatil2005:
## Overview

`ARCHITECTURE.md` does not mention the sidecar anywhere: searching its 622 lines for `sidecar` or `grpc` returns nothing, and the only hits for `scanner` are a mock file's comment and the `CVEScanner` port name. A reader comes away with one binary and one SBOM path when there are two.

Adds a **SBOM Generation Modes** section, a **Sidecar SBOM Adapter** entry alongside the Syft one, and brings the directory structure back in line with the tree.

## Additional Information

the new section covers which mode runs and when, why the split exists (Syft's peak memory scales with image size, and in the sidecar the kernel kills that container instead of kubevuln), and two things that catch people reading the code:

- startup is fail-soft. the client health-checks the sidecar, bounded by `scannerReadinessTimeout`, and falls back to in-process rather than refusing to start, so a deployment can be in a different mode than its config suggests. `/v1/diagnostics` reports which, as `scanMode`.
- concurrency differs deliberately. the in-process path serialises pulls behind `pullMutex`; the sidecar does not, and bounds in-flight work by the caller's `scanConcurrency` instead.

the second point matters beyond documentation: the two paths are parallel implementations, so a change to one usually needs the same change to the other, and the doc now says so. that is the shape behind #585, #587, #601, #675 and #686.

the directory structure listed `internal/tools` as the only thing under `internal/`, had no `pkg/` tree at all, and no `cmd/sbom-scanner`. Adds `metrics`, `registryauth`, `syftmeta`, `safefetch`, `vexdoc`, `pkg/sbomscanner/v1`, `pkg/securityexception/v1beta1`, and `sidecar.go` next to `syft.go`.

every claim is checked against the code rather than inferred: `SBOM_SCANNER_SOCKET` selecting the mode (`cmd/http/main.go:83`), the fallback log and adapter swap on readiness failure, `var _ ports.SBOMCreator = (*SidecarSBOMAdapter)(nil)`, `ScanModeSidecar` reaching diagnostics, and the "do not add a mutex/semaphore around this method" comment in `pkg/sbomscanner/v1/server.go` for the concurrency difference.

one thing deliberately not claimed: `SCANNER_MEMORY_LIMIT` is described as read at startup and passed to the adapter, not as something recorded when the sidecar is OOM-killed. the path that would record it is `handleCrash`, which nothing currently reaches (#763), and documenting it as live would be wrong until that is settled.

## How to Test

Docs only, no code changes: `go build ./...` and the suite are unaffected.

Read the new **SBOM Generation Modes** section under Component Details, and check the directory structure against `find pkg internal cmd -maxdepth 2 -type d`.

The table of contents is unchanged because it lists `##` headings only and both additions are `###`, consistent with the other subsections.

## Related issues/PRs:

* Resolved #776

## Summary by CodeRabbit

- **Documentation**
  - Updated architecture guidance for in-process and sidecar SBOM generation.
  - Documented startup configuration, Unix-socket communication, and automatic fallback behavior.
  - Added an overview of registry authentication, metrics, metadata handling, safe fetching, and VEX support.
  - Clarified diagnostics, memory isolation, concurrency differences, and related environment variables.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -182,9 +182,11 @@ Kubevuln follows the hexagonal architecture pattern to achieve:
 kubevuln/
 ├── adapters/                    # External service implementations
 │   ├── v1/
-│   │   ├── syft.go             # SBOM generation using Syft
+│   │   ├── syft.go             # SBOM generation using Syft, in-process
+│   │   ├── sidecar.go          # SBOM generation via the sbom-scanner sidecar
 │   │   ├── grype.go            # CVE scanning using Grype
 │   │   ├── backend.go          # Kubescape backend communication
+│   │   ├── securityexception.go # SecurityException CRDs -> exception policies
 │   │   └── container_profile.go # Relevancy provider
 │   ├── mockcve.go              # Mock CVE scanner for testing
 │   ├── mockplatform.go         # Mock platform for testing
@@ -193,6 +195,8 @@ kubevuln/
 ├── cmd/
 │   ├── http/
 │   │   └── main.go             # HTTP server entry point
+│   ├── sbom-scanner/
+│   │   └── main.go             # Sidecar SBOM scanner entry point (gRPC over a Unix socket)
 │   └── cli/
 │       └── main.go             # CLI entry point (not implemented)
 │
@@ -218,11 +222,20 @@ kubevuln/
 │       └── scan.go             # Main scanning business logic
 │
 ├── internal/
-│   └── tools/                  # Internal utilities
+│   ├── tools/                  # Internal utilities
+│   ├── metrics/                # OpenTelemetry instruments and recorders
+│   ├── registryauth/           # Registry credential resolution, shared by both SBOM paths
+│   ├── syftmeta/               # Syft metadata reattachment
+│   ├── safefetch/              # SSRF-guarded HTTPS fetcher for user-supplied URLs
+│   └── vexdoc/                 # Safe temp-file writer for VEX documents
+│
+├── pkg/                        # Importable by other projects
+│   ├── sbomscanner/v1/         # Sidecar scanner protocol: gRPC client, server, proto
+│   └── securityexception/v1beta1/ # SecurityException / ClusterSecurityException CRD types
 │
 └── repositories/
     ├── apiserver.go            # Kubernetes API server storage
-    ├── memory.go               # In-memory storage
+    ├── memory.go               # In-memory storage (testing)
     └── broken.go               # Always-failing storage (testing)
 ```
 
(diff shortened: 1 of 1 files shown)


matthyx (approved):
Reviewed docs-only change against the actual code (`cmd/http/main.go`, `cmd/sbom-scanner/main.go`, `adapters/v1/syft.go`, `adapters/v1/sidecar.go`, and the `internal/`/`pkg/` trees). No blockers.

Checked the four CodeRabbit findings from the earlier run against source:
- "selected" vs "attempted" on `SBOM_SCANNER_SOCKET` — already fixed in 8f8a429.
- `SBOM_SCANNER_SOCKET` vs `SOCKET_PATH` confusion — not actually present; the doc already separates them correctly ("`SBOM_SCANNER_SOCKET` selects the mode on the kubevuln side; `SOCKET_PATH` ... configure[s] the sidecar"), matching `cmd/http/main.go` (reads `SBOM_SCANNER_SOCKET`) vs `cmd/sbom-scanner/main.go` (reads `SOCKET_PATH`).
- Missing language tag on the new ASCII diagram fence — every other code fence in this file is also untagged (no markdownlint config/CI gate in this repo), so this would be inconsistent with existing style rather than a fix.
- Removing the "mutex prevents disk exhaustion" claim — this is actually correct, not overstated: `adapters/v1/syft.go`'s own comment says `pullMutex` guards `syft.GetSource` specifically because concurrent pulls download image layers to local disk, and the regression test in `syft_test.go` explicitly frames it as "prevent concurrent pulls/cataloging from exhausting disk" (#687).

Also spot-checked the new directory-structure entries (`internal/{metrics,registryauth,syftmeta,safefetch,vexdoc}`, `pkg/{sbomscanner/v1,securityexception/v1beta1}`, `cmd/sbom-scanner`, `adapters/v1/{sidecar.go,securityexception.go}`) and the `memory.go` "(testing)" annotation — all match the current tree/usage.

Good to merge.

PR #782, opened 2026-08-19
refactor(controllers): drop the sort before deleting every evicted record

### LINKED ISSUE #779: evictLocked sorts the records it is about to delete, all of them
`evictLocked` sorts the candidates it is about to delete, then deletes all of them:

```go
slices.SortFunc([]candidate(selected), func(a, b candidate) int {
	return a.finishedAt.Compare(b.finishedAt)
})
for _, cand := range selected {
	delete(s.items, cand.jobID)
}
```

every element of `selected` is deleted regardless of order, so the sort cannot change which records survive. the heap above it is what picks the oldest terminal entries; by the time the sort runs, the selection is already made.

it is O(k log k) on every eviction pass that goes over the cap, done on the way to discarding its own result. `slices` is imported for this and nothing else.

not a correctness bug, just work with no effect, and a reader has to check the loop below to satisfy themselves the ordering is not load-bearing.

version: main (3f711ab)

AnshulPatil2005:
## Overview

`evictLocked` sorted the candidates it was about to delete and then deleted all of them, so the ordering could not change which records survived. The heap above already picks the oldest terminal entries; the sort was work done on the way to discarding its own result.

## Additional Information

no behaviour change: every element of `selected` is deleted either way, and the heap that chooses them is untouched.

`slices` was imported for this and nothing else here, so the import goes with it.

a short comment replaces it saying why nothing sorts, since "these are deleted in no particular order" is the kind of thing the next reader would otherwise re-derive from the loop below.

## How to Test

`go build ./... && go vet ./controllers/... && go test ./controllers/ -race -count=1`

The eviction tests pass unchanged, which is the point: they assert which records survive, and that is decided by the heap rather than by the order they are removed in.

```
TestScanStatusStore_EvictsExpiredTerminalEntries
TestScanStatusStore_GetEvictsExpiredTerminalEntries
TestScanStatusStore_EvictsOldestTerminalEntriesOverCap
```

## Related issues/PRs:

* Resolved #779

### DIFF
--- a/controllers/scan_status.go
+++ b/controllers/scan_status.go
@@ -2,7 +2,6 @@ package controllers
 
 import (
 	"container/heap"
-	"slices"
 	"sync"
 	"time"
 
@@ -71,9 +70,8 @@ func (s *scanStatusStore) evictLocked(now time.Time) {
 			heap.Fix(&selected, 0)
 		}
 	}
-	slices.SortFunc([]candidate(selected), func(a, b candidate) int {
-		return a.finishedAt.Compare(b.finishedAt)
-	})
+	// Not sorted first: every selected entry is deleted, so their order changes nothing
+	// about which records survive. The heap above is what picks them.
 	for _, cand := range selected {
 		delete(s.items, cand.jobID)
 	}


matthyx (approved):
Verified the claim in the description: `selected` (built by the max-heap over `overflow` size) already contains exactly the oldest terminal entries before the removed sort ran, and every element of `selected` is unconditionally deleted afterward — so sorting first never changed which records survive or influenced anything downstream. No other code path reads `selected`'s order.

Checked out the PR branch locally and ran:
- `go build ./...` — passes
- `go vet ./controllers/...` — passes
- `go test ./controllers/... -race -count=1` — all pass, including `TestScanStatusStore_EvictsOldestTerminalEntriesOverCap`

Single-file, minimal, no behavior change. LGTM, approving.

PR #783, opened 2026-08-19
docs(development): fix debug-logging instructions that do nothing

### LINKED ISSUE #780: DEVELOPMENT.md's two ways to enable debug logging both do nothing
`docs/DEVELOPMENT.md` gives two ways to turn on debug logging and neither does anything.

```bash
# Run with debug logging
./kubevuln -alsologtostderr -v=4

# Or set environment variables
export GOLOG_LOG_LEVEL=debug
./kubevuln
```

`cmd/http/main.go` parses no command-line flags at all, so `-alsologtostderr -v=4` are ignored and the server starts normally. and `GOLOG_LOG_LEVEL` appears nowhere in the codebase; the variable `kubescape/go-logger` actually reads is `KS_LOGGER_LEVEL`:

```go
EnvLoggerLevel = "KS_LOGGER_LEVEL"
EnvLoggerName  = "KS_LOGGER_NAME"
```

so someone follows the guide, gets the default level, and has nothing telling them the setting was ignored. `KS_LOGGER_LEVEL` is not documented anywhere in the repo.

two smaller things in the same file follow from the no-flags point:

- build verification says to run `./kubevuln --help`. that starts the server rather than printing help.
- the VS Code launch snippet passes `"args": ["-alsologtostderr", "-v=4"]`, the same flags that do not exist.

version: main (3f711ab)

AnshulPatil2005:
## Overview

`DEVELOPMENT.md` gives two ways to turn on debug logging and neither does anything. `cmd/http/main.go` parses no command-line flags, so `-alsologtostderr -v=4` are ignored, and `GOLOG_LOG_LEVEL` appears nowhere in the codebase. The variable `kubescape/go-logger` actually reads is `KS_LOGGER_LEVEL`, which was not documented anywhere.

## Additional Information

the failure mode is silent: the server starts, logs at the default level, and nothing indicates the setting was ignored.

two smaller corrections follow from the same fact that the binary takes no flags:

- build verification said to run `./kubevuln --help`, which starts the server rather than printing help. now says so.
- the VS Code launch snippet passed `"args": ["-alsologtostderr", "-v=4"]`. those are replaced with `KS_LOGGER_LEVEL` in its `env` block, and the snippet still parses as valid JSON.

`KS_LOGGER_NAME` is mentioned alongside it since it is the other variable the same logger reads, and the accepted levels are listed so the value is not guesswork.

## How to Test

Docs only. The claims are checkable directly:

```
grep -rn "flag\." cmd/http/main.go          # no flag parsing
grep -rn "GOLOG_LOG_LEVEL" --include=*.go . # no hits
```

and `KS_LOGGER_LEVEL` / `KS_LOGGER_NAME` are the constants `EnvLoggerLevel` and `EnvLoggerName` in `kubescape/go-logger`.

Setting `KS_LOGGER_LEVEL=debug` before starting kubevuln produces debug output; the previous instructions produce none.

## Related issues/PRs:

* Resolved #780

### DIFF
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -149,8 +149,8 @@ docker buildx build --platform linux/amd64,linux/arm64 \
 ### Build Verification
 
 ```bash
-# Check binary
-./kubevuln --help
+# Check the binary runs (it takes no flags, so this starts the server; Ctrl-C to stop)
+./kubevuln
 
 # Check size
 ls -lh kubevuln
@@ -292,13 +292,16 @@ func TestScanService_GenerateSBOM_Error(t *testing.T) {
 
 ### Local Debugging
 
+kubevuln parses no command-line flags: everything is configured through the config file
+and environment. Log level is one of `debug`, `info`, `warning`, `error` or `fatal`.
+
 ```bash
 # Run with debug logging
-./kubevuln -alsologtostderr -v=4
-
-# Or set environment variables
-export GOLOG_LOG_LEVEL=debug
+export KS_LOGGER_LEVEL=debug
 ./kubevuln
+
+# Optionally name the logger in the output
+export KS_LOGGER_NAME=pretty
 ```
 
 ### Using Delve Debugger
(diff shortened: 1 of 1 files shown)


matthyx (approved):
Verified the claims directly against the code:

- `cmd/http/main.go` parses no flags — `-alsologtostderr -v=4` were always silently ignored, confirmed.
- `GOLOG_LOG_LEVEL` has zero references anywhere in this repo, confirmed.
- `KS_LOGGER_LEVEL` / `KS_LOGGER_NAME` are exactly `EnvLoggerLevel` / `EnvLoggerName` in `kubescape/go-logger`'s `methods.go`, confirmed.
- The updated `launch.json` snippet is still valid JSON after moving the value into `env` and dropping `args`.

Nit (non-blocking): `helpers.Level` also has a `success` level (`ToLevel`/`SetLevel` accept it alongside debug/info/warning/error/fatal). Not required for this fix, just flagging in case you want the list fully exhaustive.

Good catch on a genuinely misleading set of instructions — approving.

PR #787, opened 2026-08-19
fix(repositories): copy the caller's maps when enriching a summary object

### LINKED ISSUE #784: enrichSummaryManifestObject* write into the caller's annotation and label maps
`enrichSummaryManifestObjectAnnotations` and `enrichSummaryManifestObjectLabels` take the caller's own manifest maps and alias them, so everything they add for the summary object is written into the manifest that was passed in.

```go
if annotations == nil {
	annotations = make(map[string]string)
}
enrichedAnnotations := annotations   // same map
...
enrichedAnnotations[timestampMetadataKey] = ...
```

`StoreCVESummary` calls both with `cve.Annotations` and `cve.Labels`, so a manifest goes in with its own entries and comes back carrying the summary's:

```
before: annotations=1 labels=1
after:  annotations=4 labels=8

caller's labels now hold:
  mine=keep
  kubescape.io/context=filtered
  kubescape.io/workload-api-group=apps
  kubescape.io/workload-api-version=v1
  kubescape.io/workload-kind=deployment
  kubescape.io/workload-name=c
  kubescape.io/workload-namespace=b
  kubescape.io/workload-container-name=app
```

`applyExceptionsToManifest` already clones both maps before touching them, for this exact reason:

```go
filtered := cve
if cve.Labels != nil {
	filtered.Labels = maps.Clone(cve.Labels)
}
if cve.Annotations != nil {
	filtered.Annotations = maps.Clone(cve.Annotations)
}
```

so the hazard is recognised elsewhere in the same flow; these two are the pair that does not.

one detail makes the labels function slightly worse than the annotations one: it writes the context label onto the incoming map before aliasing, so that entry lands on the caller even on the paths that return an error immediately afterwards.

I have not traced an ordering today where a mutated manifest is then stored or submitted with the wrong labels, so this is a latent aliasing bug rather than an observed one. it is cheap to close and the alternative is that the next reordering of the store calls decides it.

version: main (3f711ab)

AnshulPatil2005:
## Overview

`enrichSummaryManifestObjectAnnotations` and `enrichSummaryManifestObjectLabels` aliased the caller's manifest maps instead of copying them, so everything they add for the summary object was written into the manifest passed in. A manifest handed to `StoreCVESummary` came back with eight labels where it passed one.

## Additional Information

`applyExceptionsToManifest` already clones both maps before touching them for exactly this reason, so the hazard is recognised elsewhere in the same flow; these two were the pair that did not.

the labels function was slightly worse than the annotations one: it set the context label on the incoming map *before* aliasing, so that entry landed on the caller even on the paths that return an error immediately after.

both now build their own map and copy the caller's entries in, so the returned map is unchanged in content and the input is untouched. the nil handling comes along for free: copying from a nil map is a no-op and the result is still non-nil, which is what the old nil checks provided.

I have not traced an ordering today where a mutated manifest is then stored or submitted with the wrong labels, so this is latent rather than observed. it is cheap to close and the alternative is that the next reordering of the store calls decides it.

## How to Test

`go build ./... && go vet ./repositories/... && go test ./repositories/ -count=1`

Existing tests pass unchanged, including the `StoreCVESummary` ones, which is what says the returned maps still carry what they did.

`TestEnrichSummaryManifestObject_DoesNotMutateCaller` is new: it passes a one-entry map to each, asserts the caller's maps are still exactly that one entry afterwards, and that the returned maps carry both the caller's entry and the added ones.

`TestEnrichSummaryManifestObject_HandlesNilMaps` pins that nil in still gives a usable map out.

restoring the alias fails the first of those.

## Related issues/PRs:

* Resolved #784

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -748,10 +748,12 @@ func parseSeverities(cve domain.CVEManifest, cvep domain.CVEManifest, withReleva
 }
 
 func enrichSummaryManifestObjectAnnotations(ctx context.Context, annotations map[string]string) (map[string]string, error) {
-	if annotations == nil {
-		annotations = make(map[string]string)
-	}
-	enrichedAnnotations := annotations
+	// Copied, not aliased: the caller passes its own manifest's map, and the entries added
+	// below belong to the summary object being built rather than to that manifest. Writing
+	// through left a CVE manifest carrying summary annotations it never had, which is the
+	// same hazard applyExceptionsToManifest already clones to avoid.
+	enrichedAnnotations := make(map[string]string, len(annotations)+3)
+	maps.Copy(enrichedAnnotations, annotations)
 
 	workload, ok := ctx.Value(domain.WorkloadKey{}).(domain.ScanCommand)
 	if !ok {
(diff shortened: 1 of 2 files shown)


matthyx (dismissed):
Confirmed the bug: `cve.Annotations`/`cve.Labels` (the caller's manifest maps) were being aliased and written into directly in both `enrichSummaryManifestObjectAnnotations` and `enrichSummaryManifestObjectLabels`, even though the result is only ever used to populate an unrelated `VulnerabilityManifestSummary` object in `StoreCVESummary`/`StoreCVESummaryStub`. The fix copies into a fresh map in both functions (consistent with how `applyExceptionsToManifest` already clones), preserves nil-in/non-nil-out behavior, and adds regression tests (`TestEnrichSummaryManifestObject_DoesNotMutateCaller`, `TestEnrichSummaryManifestObject_HandlesNilMaps`) that would fail against the old aliasing code. `maps` is already imported in this file. Diff is minimal and scoped, all CI checks are green. Approving.

matthyx:
rebase pls

AnshulPatil2005:
> rebase pls

done

matthyx (approved):
Re-reviewed after the rebase (my prior approval was auto-dismissed by the rebase). Diff content is unchanged from what I reviewed before: `enrichSummaryManifestObjectAnnotations`/`enrichSummaryManifestObjectLabels` now copy into a fresh map instead of aliasing/writing into the caller's `cve.Annotations`/`cve.Labels`, nil-in/non-nil-out is preserved, and the two new regression tests still cover it. All CI checks are green on the rebased commit. Re-approving.

PR #788, opened 2026-08-19
fix(tools,adapters): join filesystem paths with filepath, not path

### LINKED ISSUE #785: filesystem paths are built with path.Join in four places instead of filepath.Join
Four places build filesystem paths with `path.Join`, which always separates with a forward slash, rather than `filepath.Join`:

```
internal/tools/tools.go:122   DeleteContents          os.RemoveAll(path.Join([]string{dir, c.Name()}...))
internal/tools/tools.go:225   CleanupStaleTempDirs    os.RemoveAll(path.Join(dir, e.Name()))
adapters/v1/grype.go:90       NewGrypeAdapter         DBRootDir: path.Join(xdg.CacheHome, "grype", "db")
adapters/v1/grype.go:223      NewGrypeAdapterFixedDB  DBRootDir: path.Join(xdg.CacheHome, "grype-offline", "db")
```

on Windows that produces a mixed-separator path:

```
path.Join(os.TempDir(), "stereoscope-abc")
-> "C:\Users\dpati\AppData\Local\Temp/stereoscope-abc"
```

nothing is broken by it: `os.RemoveAll` accepts forward slashes on Windows, and I confirmed a directory created normally is still removed through a `path.Join`-built path. what it costs is that these paths read wrong wherever they are logged or wrapped in an error, and the temp-dir sweep logs its target on failure.

the rest of the repo already uses `filepath`, including `activeScanLockPath` a few lines from the first two and the `filepath.Clean` calls added by #704, so this is four sites out of step rather than a deliberate choice.

the two grype ones are the more meaningful: they build on `xdg.CacheHome`, which is natively separated on every platform, so that is where a joined path is most visibly wrong.

version: main (3f711ab)

AnshulPatil2005:
## Overview

Four places built filesystem paths with `path.Join`, which always separates with a forward slash, rather than `filepath.Join`. On Windows that yields a mixed-separator path such as `C:\Users\...\Temp/stereoscope-abc`.

## Additional Information

nothing is broken by it, and the PR does not claim otherwise: `os.RemoveAll` accepts forward slashes on Windows, and I checked that a directory created normally is still removed through a `path.Join`-built path. what it costs is that these paths read wrong wherever they are logged or wrapped in an error, and the temp-dir sweep logs its target on failure.

the rest of the repo already uses `filepath`, including `activeScanLockPath` a few lines from two of these and the `filepath.Clean` calls added by #704, so this is four sites out of step rather than a deliberate choice.

the two grype ones are the more meaningful: they build on `xdg.CacheHome`, which is natively separated on every platform.

`path.Join([]string{dir, c.Name()}...)` is also unwrapped to a plain two-argument call, since that line was being touched anyway.

`"path"` is now unused in `internal/tools/tools.go` and dropped; `adapters/v1/grype.go` keeps it, as `path` is still used there for URL-shaped joins.

## How to Test

`go build ./... && go vet ./... && go test ./internal/tools/ ./adapters/... -count=1`

The temp-dir sweep tests cover both changed lines in `internal/tools` and pass unchanged. The grype DB root dirs are set at construction and exercised by the adapter tests.

`grep -rn "path.Join(" --include=*.go . | grep -v _test | grep -v filepath` returns nothing afterwards.

## Related issues/PRs:

* Resolved #785

### DIFF
--- a/adapters/v1/grype.go
+++ b/adapters/v1/grype.go
@@ -5,7 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"path"
+	"path/filepath"
 	"strings"
 	"sync"
 	"time"
@@ -87,7 +87,7 @@ func NewGrypeAdapter(listingURL string, matchingMode config.CVEMatchingMode, tru
 	g := &GrypeAdapter{
 		distCfg: distCfg,
 		installCfg: installation.Config{
-			DBRootDir: path.Join(xdg.CacheHome, "grype", "db"),
+			DBRootDir: filepath.Join(xdg.CacheHome, "grype", "db"),
 		},
 		matchingMode:   matchingMode,
 		trustedVendors: buildTrustedVendorSet(trustedVendors),
@@ -220,7 +220,7 @@ func NewGrypeAdapterFixedDBWithMatchers(matchingMode config.CVEMatchingMode, tru
 	g := &GrypeAdapter{
 		distCfg: distCfg,
 		installCfg: installation.Config{
-			DBRootDir: path.Join(xdg.CacheHome, "grype-offline", "db"),
+			DBRootDir: filepath.Join(xdg.CacheHome, "grype-offline", "db"),
 		},
 		matchingMode:   matchingMode,
 		trustedVendors: buildTrustedVendorSet(trustedVendors),
--- a/internal/tools/tools.go
+++ b/internal/tools/tools.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 	"errors"
 	"os"
-	"path"
 	"path/filepath"
 	"regexp"
 	"runtime/debug"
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Reviewed the diff and the full resulting files (not just the patch context).

- All four `path.Join` → `filepath.Join` swaps are correct and match existing repo convention (`activeScanLockPath`, the `filepath.Clean` calls from #704).
- Confirmed `internal/tools/tools.go` has no other use of `path.*`, so dropping the import is safe.
- One small nit, not a blocker: the PR description says `adapters/v1/grype.go` "keeps `path`, as `path` is still used there for URL-shaped joins" — that's not what the diff does; the file's import block replaces `path` with `path/filepath` entirely, and there's no remaining `path.*` call in the file. Worth fixing the description for future readers, but the code itself is correct.
- CI is green (build, vet, lint, tests, CodeQL, DCO) and a repo-wide search turns up no remaining `path.Join` calls outside `_test`/`filepath` files.

LGTM, approving.

PR #789, opened 2026-08-19
docs(api): document the /v1/diagnostics endpoint

### LINKED ISSUE #786: C:/Program Files/Git/v1/diagnostics is undocumented
`GET /v1/diagnostics` is registered in `cmd/http/main.go` and returns eight fields, and it is documented nowhere. `docs/API.md` has sections for liveness, readiness and metrics but not this one, and grepping the whole `docs/` tree for "diagnostics" returns nothing.

```go
router.GET("/v1/liveness", controller.Alive)
router.GET("/v1/readiness", controller.Ready)
router.GET("/v1/diagnostics", controller.Diagnostics)
router.GET("/metrics", gin.WrapH(m.Handler()))
```

it answers a question nothing else can. when `SBOM_SCANNER_SOCKET` is set but the sidecar never becomes ready within `scannerReadinessTimeout`, startup logs the failure and falls back to the in-process Syft adapter, so the mode actually running can differ from the mode configured. `scanMode` is how an operator tells which, and the config file will not say.

the rest of what it reports is the same kind of thing: the Syft and Grype versions actually loaded, the Grype DB checksum, the effective timeouts, and whether storage and risk acceptance are on.

worth documenting alongside it: the handler always returns `200`, and when no diagnostics provider was attached it returns the struct's zero value, so an all-empty body means "not wired up" rather than "nothing configured".

version: main (3f711ab)

AnshulPatil2005:
## Overview

`GET /v1/diagnostics` is registered in `cmd/http/main.go` and returns eight fields, and appears nowhere in `docs/`. API.md documents liveness, readiness and metrics but not this one.

## Additional Information

it answers a question nothing else can. when `SBOM_SCANNER_SOCKET` is set but the sidecar never becomes ready within `scannerReadinessTimeout`, startup logs the failure and falls back to the in-process Syft adapter, so the mode actually running can differ from the mode configured. `scanMode` is how an operator tells which, and the config file will not say.

field names, types and the two `scanMode` values are taken from `domain.Diagnostics` and its constants rather than from a sample response, so the table cannot drift from an example someone pasted.

also documents that the handler always returns `200`, and that an all-empty body means no diagnostics provider was attached rather than nothing being configured, since that distinction is invisible from the response alone.

this pairs with #777, which references `/v1/diagnostics` as the way to tell which SBOM mode is in effect; that reference now has somewhere to point.

## How to Test

Docs only.

Every documented field is checkable against `core/domain/diagnostics.go`, and the two `scanMode` values against `ScanModeSidecar` and `ScanModeInProcess` in the same file. The example response parses as JSON.

`curl -s http://localhost:8080/v1/diagnostics | jq` against a running instance returns the documented shape.

## Related issues/PRs:

* Resolved #786


matthyx (approved):
Verified every claim against source rather than trusting the prose:

- Field names/types/order in the table match `domain.Diagnostics` (`core/domain/diagnostics.go`) exactly, including the two `scanMode` constants (`ScanModeSidecar`/`ScanModeInProcess`).
- "Always returns `200`" and "all-empty body means no provider attached" match `HTTPController.Diagnostics` in `controllers/http.go` (`c.JSON(http.StatusOK, ...)` on both the nil-callback and normal paths).
- `cveScannerVersion`'s "Grype version, suffixed with the active `cveMatchingMode`" matches `GrypeAdapter.Version()` in `adapters/v1/grype.go` (`v + "-matching-" + matchingMode`), and the example value follows that shape.
- `cveDBVersion` "empty until the first load completes" matches `dbVersionLocked()`.
- `storageEnabled`/`riskAcceptanceEnabled` sourced correctly from `c.Storage` / `isRiskAcceptanceActive` in `cmd/http/main.go`.
- TOC anchor, section placement, and separators (`---`) are consistent with the rest of the file.

Docs-only change, fully resolves #786, no code paths touched. No blockers. Approving.

PR #791, opened 2026-08-19
fix(controllers): stop scan status lookups from sweeping the whole store

### LINKED ISSUE #790: ScanStatus's single-key lookup pays O(n) eviction cost, under an exclusive lock, on every call
`ScanStatus` (`GET /v1/{version}/scanStatus/:jobID`) is a single-key status lookup, but the code behind it does a full, exclusive-locked scan of the entire status store on every call — regardless of whether anything actually needs evicting.

```go
// controllers/scan_status.go
func (s *scanStatusStore) get(jobID string) (domain.ScanStatus, bool) {
	s.mu.Lock()
	defer s.mu.Unlock()
	s.evictLocked(time.Now().UTC())
	status, ok := s.items[jobID]
	return status, ok
}
```

`evictLocked` is O(n) in the store's size — it ranges over every entry to check TTL expiry, and if still over `maxEntries` (default 10,000), ranges again to build a heap of eviction candidates:

```go
func (s *scanStatusStore) evictLocked(now time.Time) {
	for jobID, status := range s.items {           // pass 1: TTL check, every entry
		...
	}
	overflow := len(s.items) - s.maxEntries
	if overflow <= 0 {
		return
	}
	selected := make(candidateHeap, 0, overflow)
	for jobID, status := range s.items {            // pass 2: heap-select, every entry
		...
	}
	...
}
```

## Root cause

`get()` takes `s.mu` for **writing** (`Lock()`, not `RLock()`) and unconditionally runs the store's most expensive maintenance operation before doing a one-key map lookup. `evictLocked` is also called from `recordAccepted` (every scan acceptance) and `markTerminal` (every scan completion), so the store already self-cleans continuously as scans flow through it — `get()`'s own eviction pass is redundant busywork on the hot read path, not something correctness depends on for the entry actually being requested.

The same `s.mu` is the *only* lock guarding every other write to the store: `recordAccepted`, `markRunning`, `markPhase`, `markSucceeded`/`markFailed`/`markAbandoned` (via `markTerminal`), and `markAbandonedQueued` all take it too — these fire on every phase transition of every concurrent scan (`domain.UpdateScanPhase` is called multiple times per scan: `sbom_generation`, `cve_matching`, `result_storage`, `result_upload`, etc.). Because `get()` holds the same lock exclusively for the duration of an O(n) scan, a single status poll blocks every in-flight scan's phase-tracking calls for as long as the eviction pass takes, and vice versa — read traffic and write traffic on this store fully serialize against each other, sized by the store's total entry count rather than by the one key being read or written.

## Impact

- **Wasted work on every read**: a lookup for one `jobID` pays for an O(n) (up to O(2n) when over cap) walk of up to `maxEntries` (10,000) entries, every single time, even when the store is nowhere near stale or full.
- **Lock contention scales with poll frequency and store size**: a caller polling `/scanStatus/:jobID` for job completion (the endpoint's own intended use) repeatedly re-triggers this full-store scan, and each one holds the store's only mutex exclusively — blocking `recordAccepted`/`markPhase`/`markTerminal` calls from every other concurrently running scan job for its duration, and vice versa.
- Not a correctness bug (results are still right), but a real scalability/performance one: the cost of a targeted single-key read grows with total store occupancy instead of staying O(1), and it's paid by every caller in the system via one shared lock.

## Reproduction reasoning

1. A kubevuln instance accumulates entries up to `maxEntries` (10,000) under normal operation — the default cap exists specifically because this is expected to happen (`defaultScanStatusMaxEntries` doc comment).
2. A client (the natural consumer of this endpoint) polls `GET /v1/{version}/scanStatus/:jobID` repeatedly to watch a job's status until it's done.
3. Each poll takes `s.mu.Lock()` and walks the full map (twice, once over cap) before returning one entry.
4. Every concurrent scan's `markPhase`/`markRunning`/`markTerminal` call for its own, unrelated `jobID` queues up behind that same lock for the duration of the scan.

This is directly reachable through the store's only public read path and requires no unusual configuration — just enough concurrent scan volume (the queue admission control in #749 exists precisely because this is expected) plus a caller polling status, which is what the endpoint is for.

## Affected components

- `controllers/scan_status.go`: `scanStatusStore.get`, `evictLocked`
- `controllers/http.go`: `HTTPController.ScanStatus` (the only caller of `get`)

## Acceptance criteria

- `get(jobID)` no longer pays O(n) eviction cost on every call; a single-key lookup's cost should not scale with total store occupancy.
- Read/write concurrency on the store improves: a `get()` call must not exclude concurrent `markPhase`/`markRunning`/`markTerminal` calls for unrelated job IDs for the duration of a full-store scan.
- Existing eviction guarantees are preserved: terminal entries older than `ttl` are still eventually evicted, and the store still respects `maxEntries` — `TestScanStatusStore_EvictsExpiredTerminalEntries`, `TestScanStatusStore_EvictsOldestTerminalEntriesOverCap`, and friends in `controllers/scan_status_test.go` continue to pass.
- New test coverage demonstrating `get()` alone (with no intervening `recordAccepted`/`markTerminal` call) does not block/serialize against concurrent writes for unrelated job IDs.

## Suggested direction

`get()` doesn't need to trigger eviction itself — `recordAccepted` and `markTerminal` already run it continuously as scans start and finish, so the store self-cleans under normal traffic. A few options, roughly in order of how invasive they are:

1. Drop the `evictLocked` call from `get()` entirely and take `s.mu.RLock()` instead of `Lock()`, relying on the write paths' existing eviction calls to keep the store bounded. This covers the common case (scans are actively flowing through the store) but leaves a gap if reads continue after writes stop entirely.
2. Keep a cheap eviction trigger in `get()`, but throttle it (e.g. a `lastEvictedAt` timestamp, skip re-running `evictLocked` if called again within some short interval) so repeated polls don't each pay the full cost.
3. Move eviction off the read/write hot paths entirely onto a background ticker, and make `get()` a pure `RLock`-guarded map read.

Any of these keeps the store's existing bounding guarantees while stopping a single-key read from being the store's most expensive operation.

version: main (3f711ab)

bhuvan-somisetty:
## Problem

`GET /v1/{version}/scanStatus/:jobID` is a single-key lookup, but `scanStatusStore.get()` ran the store's full `evictLocked` sweep — an O(n) walk of every entry in the store — under an **exclusive** lock on every call, regardless of whether anything needed evicting.

```go
func (s *scanStatusStore) get(jobID string) (domain.ScanStatus, bool) {
	s.mu.Lock()
	defer s.mu.Unlock()
	s.evictLocked(time.Now().UTC())
	status, ok := s.items[jobID]
	return status, ok
}
```

## Root cause

`evictLocked` is also called by `recordAccepted` (every scan acceptance) and `markTerminal` (every scan completion), so the store already self-cleans continuously as scans flow through it. `get()` running the same full-store sweep again on every read is redundant work — and because it takes the store's only mutex (`s.mu`) for writing, it serializes against `recordAccepted`/`markRunning`/`markPhase`/`markTerminal` for every other concurrently running scan for the duration of the sweep.

The one thing `get()` genuinely needs, that a write-triggered sweep doesn't guarantee by itself, is: never hand back a terminal record whose TTL has already elapsed but that hasn't been evicted by a write yet (existing test `TestScanStatusStore_GetEvictsExpiredTerminalEntries` asserts exactly this). That only requires checking the *one* entry being requested, not the whole store.

## Fix

- Extracted the per-entry TTL check `evictLocked` already did into `expiredTerminal(status, now, ttl)`, shared by both `evictLocked`'s sweep and `get()`.
- `get()` now takes `s.mu.RLock()`, looks up the one key, and returns immediately if it isn't an expired terminal record — O(1), no sweep.
- If the requested entry *is* an expired terminal record, `get()` upgrades to a narrow write lock and deletes just that key (re-checking it's still expired after acquiring the lock, in case a concurrent write changed it), then reports not-found — matching `evictLocked`'s own deletion behavior for that entry, but scoped to the one key instead of the whole store.
- `maxEntries` capping and routine TTL cleanup are unaffected: they're still fully handled by the existing `recordAccepted`/`markTerminal` → `evictLocked` calls, which never depended on `get()` also running them (map growth only ever happens in `recordAccepted`, which already sweeps right after adding an entry).

## Testing

- All existing `controllers/scan_status_test.go` tests pass unmodified, including `TestScanStatusStore_GetEvictsExpiredTerminalEntries`, which specifically asserts `get()` still never returns an expired-but-unevicted terminal entry.
- New regression tests:
  - `TestScanStatusStore_GetDoesNotEvictUnrelatedStaleEntries` — populates two stale, terminal, unrelated entries, calls `get()` for a different key, and asserts both stale entries are still present in the map afterward. This is the direct regression test for the bug: before the fix, any `get()` call would have swept and removed them.
  - `TestScanStatusStore_GetEvictsOnlyTheRequestedStaleKey` — asserts `get()` on an expired key both reports not-found *and* evicts exactly that key, while a second unrelated expired entry is left untouched (proving eviction is now scoped, not global).
  - `BenchmarkScanStatusStore_Get` — benchmarks `get()` against store sizes of 100 and 10,000 entries; ns/op should stay flat across sizes now, versus scaling with store size before.
- `go build ./controllers/...` and `go vet ./controllers/...` pass.

## Impact

- A status poll no longer pays for evaluating every other entry in the store, and no longer blocks every other in-flight scan's phase-tracking writes for the duration of a full-store scan.
- Purely an internal locking/complexity change: the store's external eviction guarantees (TTL expiry, `maxEntries` cap) are unchanged, and `HTTPController.ScanStatus`, the only caller of `get()`, needs no changes.

Fixes #790

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scan status lookups by avoiding unnecessary cleanup of unrelated expired records.
  * Expired entries are now removed only when they are specifically requested.
  * Active and current scan statuses return more efficiently.

* **Tests**
  * Added coverage for selective expiration cleanup.
  * Added performance benchmarks for lookups across stores of different sizes.

### DIFF
--- a/controllers/scan_status.go
+++ b/controllers/scan_status.go
@@ -38,15 +38,18 @@ func newScanStatusStore() *scanStatusStore {
 	}
 }
 
+// expiredTerminal reports whether status is a terminal record whose TTL has elapsed as of
+// now. Used both by evictLocked's full-store sweep and by get's single-key check.
+func expiredTerminal(status domain.ScanStatus, now time.Time, ttl time.Duration) bool {
+	return isTerminal(status.State) && status.FinishedAt != nil && now.Sub(*status.FinishedAt) > ttl
+}
+
 // evictLocked removes terminal records older than s.ttl, then, if still over
 // s.maxEntries, removes the oldest terminal records until the cap is met. Active
 // (queued/running) records are never evicted. Callers must hold s.mu for writing.
 func (s *scanStatusStore) evictLocked(now time.Time) {
 	for jobID, status := range s.items {
-		if !isTerminal(status.State) {
-			continue
-		}
-		if status.FinishedAt != nil && now.Sub(*status.FinishedAt) > s.ttl {
+		if expiredTerminal(status, now, s.ttl) {
 			delete(s.items, jobID)
 		}
 	}
@@ -235,10 +238,30 @@ func (s *scanStatusStore) markTerminal(jobID string, state domain.ScanState, rea
 	s.evictLocked(now)
 }
 
+// get looks up jobID without paying for evictLocked's full-store sweep: recordAccepted and
+// markTerminal already run it on every write, so the store is kept bounded and mostly fresh
+// without a read needing to do that work too (see #790). The one thing a read still owes on
+// its own is not handing back a terminal record whose TTL has elapsed but that no write has
+// evicted yet -- expiredTerminal is the same check evictLocked uses, applied to just this one
+// entry instead of every entry in the store.
 func (s *scanStatusStore) get(jobID string) (domain.ScanStatus, bool) {
-	s.mu.Lock()
-	defer s.mu.Unlock()
-	s.evictLocked(time.Now().UTC())
+	s.mu.RLock()
 	status, ok := s.items[jobID]
-	return status, ok
+	s.mu.RUnlock()
+	if !ok {
+		return domain.ScanStatus{}, false
+	}
+	now := time.Now().UTC()
+	if !expiredTerminal(status, now, s.ttl) {
+		return status, true
+	}
+
+	// Evict just this key: re-check under the write lock in case a concurrent write
+	// already removed or changed it between the RUnlock above and here.
+	s.mu.Lock()
+	if cur, ok := s.items[jobID]; ok && expiredTerminal(cur, now, s.ttl) {
+		delete(s.items, jobID)
+	}
+	s.mu.Unlock()
+	return domain.ScanStatus{}, false
 }
--- a/controllers/scan_status_test.go
+++ b/controllers/scan_status_test.go
@@ -99,6 +99,74 @@ func TestScanStatusStore_EvictsOldestTerminalEntriesOverCap(t *testing.T) {
 	require.True(t, ok, "newly accepted job must be retained")
 }
 
+// Before #790, get() ran evictLocked's full-store sweep on every call, so a lookup for one
+// jobID paid for evaluating every other entry in the store too. This proves get() no longer
+// does that: stale, unrelated terminal entries elsewhere in the store must survive a get()
+// call for a different key, since only a write path (recordAccepted/markTerminal) or a get()
+// for that specific key is what evicts them now.
+func TestScanStatusStore_GetDoesNotEvictUnrelatedStaleEntries(t *testing.T) {
+	s := newScanStatusStore()
+	s.ttl = time.Millisecond
+
+	s.recordAccepted("stale-1", "generateSBOM")
+	s.markSucceeded("stale-1")
+	s.recordAccepted("stale-2", "generateSBOM")
+	s.markSucceeded("stale-2")
+
+	// Backdate both past the TTL directly, bypassing the eviction that recordAccepted and
+	// markSucceeded already ran while their own FinishedAt was still fresh.
+	s.mu.Lock()
+	longAgo := time.Now().UTC().Add(-time.Hour)
+	for _, jobID := range []string{"stale-1", "stale-2"} {
+		st := s.items[jobID]
+		st.FinishedAt = &longAgo
+		s.items[jobID] = st
+	}
+	s.mu.Unlock()
+
+	// A get() for an unrelated, non-existent key must not sweep the store.
+	_, ok := s.get("unrelated")
+	require.False(t, ok)
+
+	s.mu.RLock()
+	_, stale1Present := s.items["stale-1"]
+	_, stale2Present := s.items["stale-2"]
+	s.mu.RUnlock()
+	require.True(t, stale1Present, "get() for a different key must not evict unrelated stale entries")
+	require.True(t, stale2Present, "get() for a different key must not evict unrelated stale entries")
+}
+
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Reviewed the diff and locking logic in `controllers/scan_status.go`.

- `get()`'s new fast path (RLock, single-key lookup, return immediately if not an expired terminal record) is correct and matches the pre-existing semantics for both active and non-expired terminal records.
- The narrow-eviction path correctly re-checks the entry under the write lock before deleting, closing the RUnlock→Lock race window (a concurrent write could have replaced/removed the entry in between).
- `expiredTerminal` extraction is a faithful refactor of the condition `evictLocked` already used — no behavior change there.
- `maxEntries` capping and TTL sweeping are unaffected since they still run fully on every `recordAccepted`/`markTerminal` write, which `get()` never depended on bypassing.
- New tests (`GetDoesNotEvictUnrelatedStaleEntries`, `GetEvictsOnlyTheRequestedStaleKey`) directly cover the regression this PR fixes, and the existing `GetEvictsExpiredTerminalEntries` test still passes unmodified.
- All CI checks (build, vet, lint, tests, CodeQL) are green, and CodeRabbit's automated review found no actionable issues.

No blockers found. Approving.

PR #793, opened 2026-08-19
refactor: share the Syft source setup between both SBOM paths

### LINKED ISSUE #792: Syft platform parsing and GetSource config are duplicated across both SBOM paths
the two SBOM paths each write out Syft source setup separately, and each carries a comment saying it mirrors the other.

`adapters/v1/syft_platform.go`:

```go
// ... This mirrors pkg/sbomscanner/v1/server.go so in-process and sidecar SBOM
// generation resolve the same manifest.
func parseSyftPlatform(platformStr string) (*image.Platform, error) {
	if platformStr == "" {
		return nil, nil
	}
	if !strings.Contains(platformStr, "/") {
		platformStr = "linux/" + platformStr
	}
	return image.NewPlatform(platformStr)
}
```

`pkg/sbomscanner/v1/server.go`:

```go
// ... This mirrors the in-process adapter (adapters/v1/syft.go) and matters for pod-less
// scan paths ... (see #512).
var imgPlatform *image.Platform
if req.Platform != "" {
	platformStr := req.Platform
	if !strings.Contains(platformStr, "/") {
		platformStr = "linux/" + platformStr
	}
	p, err := image.NewPlatform(platformStr)
	...
}
```

the GetSource config is duplicated the same way: `syftGetSourceConfig` in the adapter, and the same four calls built inline in the server.

both comments are right that the two must agree, and mirroring is a weak way to hold that. platform resolution is exactly what #512 was: defaulting to the scanning process's own architecture forced a mismatch for single-arch images on the pod-less paths, and the fix had to land in both places.

this is the same shape as #675, #686 and #741, each of which replaced a pair of hand-maintained copies across this boundary with one function. `internal/` is importable from both `adapters/` and `pkg/sbomscanner/`, so there is somewhere for it to live.

nothing is broken today: I diffed the two and they agree. the point is that they agree by hand.

version: main (287bfd8)

AnshulPatil2005:
## Overview

Platform parsing and the `GetSource` config were written out separately in the in-process adapter and the sidecar scanner, each carrying a comment pointing at the other saying it mirrors it. `internal/syftsource` holds both now.

## Additional Information

both comments were right that the two must agree, and mirroring is a weak way to hold that. platform resolution is what #512 was about: defaulting to the scanning process's own architecture forced a mismatch for single-arch images on the pod-less scan paths, and the fix had to land in both copies.

this is the same shape as #675, #686 and #741, each of which replaced a hand-maintained pair across this boundary with one function.

nothing was broken: I diffed the two before touching them and they agreed. what changes is that they now agree because there is one of them.

the sidecar keeps its own error wrapping, `invalid platform %q`, which was the only behavioural difference between the copies. everything else is identical, so both call sites are a straight substitution.

`internal/` is importable from `pkg/sbomscanner` because that package is itself under this module, so an external consumer importing it still compiles.

`WithSources("registry")` gets a comment it did not have: kubevuln always pulls, and leaving Syft's default source set would let it fall back to a local docker daemon or containerd socket, which is the kind of thing that only shows up on a developer machine where one happens to exist.

## How to Test

`go build ./... && go vet ./... && go test ./... -count=1`

The platform tests move with the code, renamed to match the exported names, and keep every case: empty, architecture-only, full OCI, variant, and invalid.

`TestGetSourceConfig_restrictsToRegistry` is new and pins the source restriction, which nothing asserted before.

The sidecar's own tests cover its call site, and `adapters/v1` covers the in-process one; both pass unchanged.

## Related issues/PRs:

* Resolved #792

## Summary by CodeRabbit

- **Improvements**
  - Improved platform handling for SBOM generation, including consistent support for architecture-only specifications.
  - Empty platform values are handled appropriately, while invalid platform specifications continue to return clear errors.
  - SBOM image resolution now consistently uses the specified registry and platform settings across supported scanning paths.
  - Existing retry, fallback, timeout, and error-handling behavior remains unchanged.


pkg/sbomscanner/v1/server.go
matthyx, line 1: Nit, non-blocking: the wrapped error now interpolates the raw `req.Platform` instead of the normalized `platformStr` (with `linux/` prepended when the input had no slash), e.g. an input of `bogus!arch` used to produce `invalid platform "linux/bogus!arch": ...` and now produces `invalid platform "bogus!arch": ...`. Minor, arguably clearer since it echoes what the caller actually sent, but it's a small deviation from the PR description's claim that "everything else is identical" between the two copies. Worth a one-line mention in the description, or reverting to pass the normalized string if exact parity is the goal.

matthyx (approved):
Solid refactor. `internal/syftsource` cleanly consolidates the platform-parsing and GetSource-config logic that was previously duplicated (with a "mirrors the other copy" comment) between `adapters/v1/syft.go` and `pkg/sbomscanner/v1/server.go`. I diffed the extracted functions against the originals and they're behaviorally equivalent (empty → nil platform, bare arch → `linux/` prefix, same `WithSources("registry")` restriction). Tests move over intact, keep all original cases, and add a new one (`TestGetSourceConfig_restrictsToRegistry`) pinning behavior that wasn't previously asserted. All CI checks are green and CodeRabbit found no actionable comments.

Left one non-blocking nit inline about a small error-message wording change in the sidecar path. Approving — good to merge.

PR #795, opened 2026-08-19
fix(config): say why clusterData.json was rejected instead of only that it is missing

### LINKED ISSUE #794: a malformed clusterData.json reports as a missing one at startup
with no `API_URL`, a `clusterData.json` that exists but is wrong produces a message telling the operator to set the field they already set, and nothing saying the file was read and rejected.

```go
if apiURL == "" {
	if services, err := loadBackendServicesFromClusterData(configDir); err == nil {
		return services, nil
	}
	return nil, fmt.Errorf("no service configuration: provide %s/services.json, set API_URL, or set backendOpenAPI/eventReceiverRestURL in clusterData.json", configDir)
}
```

the error is discarded, and that file fails three ways, only one of which is absence:

```go
content, err := os.ReadFile(filepath.Clean(filePath))
if err != nil {
	return nil, err                                              // absent
}
if err := json.Unmarshal(content, &clusterData); err != nil {
	return nil, fmt.Errorf("failed to parse %s: %w", filePath, err)   // malformed
}
...
if apiServerURL == "" || reportReceiverURL == "" {
	return nil, fmt.Errorf("no static backend URLs in %s", filePath)  // no usable URLs
}
```

the third is the one worth having. `normalizeServiceURL` returns empty for anything it cannot make sense of, so a typo in either URL lands there, and the operator is told to set a field that is sitting in the file in front of them.

this runs at startup:

```go
backendServices, err := config.LoadBackendServicesConfig(configDir, apiURL)
if err != nil {
	logger.L().Ctx(ctx).Fatal("load services error", helpers.Error(err))
}
```

so it is fatal and it is the only message they get.

the other branch of the same function already keeps its errors rather than replacing them:

```go
return nil, errors.Join(err, fallbackErr)
```

wrapping instead of replacing is enough. the guidance is still worth keeping for the case where the file really is absent, which is the common one.

version: main (287bfd8)

AnshulPatil2005:
## Overview

With no `API_URL`, `clusterData.json` is the last place `LoadBackendServicesConfig` looks, and its error was discarded in favour of a fixed message telling the operator to set `backendOpenAPI`/`eventReceiverRestURL`. That file fails three ways and only one of them is absence, so a file that was found and rejected reported as one that was never there.

## Additional Information

the third failure mode is the one worth surfacing: `normalizeServiceURL` returns empty for anything it cannot make sense of, so a typo in either URL produces "no static backend URLs" and the operator was told to set a field sitting in the file in front of them.

this runs at startup and `cmd/http/main.go` treats it as fatal, so it is the only message they get.

the branch below in the same function already keeps its errors rather than replacing them, with `errors.Join(err, fallbackErr)`. this one now wraps for the same reason.

the guidance text is kept rather than swapped out: when the file genuinely is absent, which is the common case, the message reads as it did.

## How to Test

`go build ./... && go vet ./... && go test ./config/ -count=1`

`TestLoadBackendServicesFromClusterData_Errors` passes unchanged; it covers the three failures at the level below this one.

`TestLoadBackendServicesConfig_NoAPIURL_SurfacesClusterDataError` is new and covers the path that discarded them: a file that parses but carries no usable URLs, and a file that is not valid JSON. Each asserts the guidance is still present *and* that the reason now is too. A third case pins that a genuinely absent file still reports the guidance.

restoring the old branch fails the first two and leaves the third passing.

## Related issues/PRs:

* Resolved #794

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -330,10 +330,16 @@ func LoadBackendServicesConfig(configDir, apiURL string) (schema.IBackendService
 	}
 
 	if apiURL == "" {
-		if services, err := loadBackendServicesFromClusterData(configDir); err == nil {
+		services, err := loadBackendServicesFromClusterData(configDir)
+		if err == nil {
 			return services, nil
 		}
-		return nil, fmt.Errorf("no service configuration: provide %s/services.json, set API_URL, or set backendOpenAPI/eventReceiverRestURL in clusterData.json", configDir)
+		// Wrapped rather than replaced: clusterData.json being absent is only one of the
+		// three ways this fails. It also reports a parse failure and a file that parses but
+		// carries no usable URLs, and dropping those told an operator who had already set
+		// backendOpenAPI to go and set it, with nothing saying the file had been read and
+		// rejected. This runs at startup and is fatal, so it is the only message they get.
+		return nil, fmt.Errorf("no service configuration: provide %s/services.json, set API_URL, or set backendOpenAPI/eventReceiverRestURL in clusterData.json: %w", configDir, err)
 	}
 
 	client, err := v3.NewServiceDiscoveryClientV3(apiURL)
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -363,3 +363,46 @@ func TestLoadConfigEnvBindingKeepsFilePrecedence(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, "from-env", c.AccountID, "the environment must still win over the file")
 }
+
+// With no API_URL, clusterData.json is the last place LoadBackendServicesConfig looks, and
+// its failure used to be discarded in favour of a fixed message telling the operator to set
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Verified the change against the underlying `loadBackendServicesFromClusterData`/`normalizeServiceURL` logic: the three failure modes (absent file, invalid JSON, parsed-but-no-usable-URLs) are exactly what the new test cases exercise, and the `%w` wrap is scoped correctly (`%s` still consumes `configDir`, `%w` consumes `err`). The only caller (`cmd/http/main.go`) just logs the error via `helpers.Error(err)` and doesn't inspect it with `errors.Is`/`errors.As`, so wrapping is behavior-safe. The "genuinely absent" case still contains the "no service configuration" guidance as asserted. All CI checks are green (build, vet, lint, tests). No blockers — LGTM.

PR #797, opened 2026-08-19
fix(adapters,sbomscanner): guard the pull, not just cataloguing, against the temp-dir sweep

### LINKED ISSUE #796: the temp-dir sweep can delete a stereoscope directory a still-running pull is actively using
The periodic stereoscope temp-dir sweep (`internal/tools.StartPeriodicTempDirSweep`, added by #669) is meant to never touch a directory a live pull or cataloguing pass is still using. Two gaps in how its safety guard (`BeginActiveTempDirUse`/`activeTempDirUsers`) is wired up mean it doesn't actually cover the phase most likely to run long: the image pull itself.

## Evidence

**1. The guard is registered after the pull completes, not before it.**

`adapters/v1/syft.go`:
```go
228:	src, err := registryauth.ResolveSource(ctxWithTimeout, metrics.ComponentInProcess, ...)   // the pull
...
308:	endActiveTempDirUse := tools.BeginActiveTempDirUse(os.TempDir())                          // the guard
309:	err = dl.Run(func(stopper <-chan struct{}) error {                                        // cataloguing starts
```

`pkg/sbomscanner/v1/server.go` — the sidecar's own handler for the same operation — has the identical shape:
```go
165:	src, err := resolveSource(ctxWithTimeout, ...)                                            // the pull
...
247:	endActiveTempDirUse := tools.BeginActiveTempDirUse(os.TempDir())                          // the guard
248:	err = dl.Run(func(stopper <-chan struct{}) error {                                        // cataloguing starts
```

In both places, `ResolveSource`/`resolveSource` is what actually creates the stereoscope temp directory and downloads image layers into it (registry pull, network-bound). `BeginActiveTempDirUse` — the call that registers `activeTempDirUsers` and takes the cross-process flock lease `acquireSweepLease` checks for — isn't called until after that pull has already finished, right before the cataloguing step. `BeginActiveTempDirUse`'s own doc comment states the intended invariant: *"While any caller is registered ... `StartPeriodicTempDirSweep` skips its sweep entirely rather than risk deleting a directory still in use."* The pull phase is exactly such a use, and it isn't registered.

**2. The sidecar's sweep threshold is a hardcoded constant, decoupled from how long a pull is actually allowed to run.**

`cmd/sbom-scanner/main.go`:
```go
43:	const tempDirSweepInterval = 5 * time.Minute
...
132:	tools.StartPeriodicTempDirSweep(stopSweep, os.TempDir(), "stereoscope-", tempDirSweepInterval, tempDirSweepInterval, ...)
```

`tempDirSweepInterval` is fixed at 5 minutes for both the sweep's tick interval and its staleness cutoff (`olderThan`), with no relationship to any individual request's timeout. But a request's actual allowed pull+cataloguing window is `time.Duration(req.TimeoutSeconds) * time.Second` (`pkg/sbomscanner/v1/server.go:152`), where `req.TimeoutSeconds` is set by the caller (`adapters/v1/sidecar.go:138`, `Timeout: s.scanTimeout`) from `config.Config.ScanTimeout` — a value `config/config.go` only requires to be positive, with no upper bound and no tie to the sidecar's 5-minute sweep constant. An operator scanning large images commonly raises `scanTimeout` well past 5 minutes; `cmd/http/main.go:198` even derives its *own* sweep threshold from that same `c.ScanTimeout` for exactly this reason, but the sidecar process never sees that value and keeps its independent 5-minute constant regardless.

## Root cause

Combined, these two gaps mean: a pull that is still legitimately running — within its own configured timeout, which can be well above 5 minutes — has a temp directory on disk that (a) isn't registered as in-use yet, and (b) can look older than the sweep's fixed cutoff purely because that cutoff has no relationship to the timeout actually governing the pull. A periodic sweep tick landing in that window removes the directory out from under the still-running pull.

This is not limited to the same process: `cmd/http` and `cmd/sbom-scanner` run as separate OS processes sharing the same `/tmp` volume and the same `"stereoscope-"` prefix (documented in `activeTempDirUsers`' own comment in `internal/tools/tools.go`), and `acquireSweepLease` is exactly the mechanism meant to protect across that boundary via a shared flock lease — but since `BeginActiveTempDirUse` hasn't been called yet during the pull, that lease isn't held either, so either process's sweep can win the race against the other process's in-progress pull.

## Impact

A legitimate, still-in-budget image pull can have its own temp directory deleted mid-download by a concurrent periodic sweep — in the same process or the other one sharing `/tmp`. The visible symptom is that specific scan failing (a partially-deleted stereoscope temp dir mid-read/write), not a leak — the opposite failure mode from the one #667/#669 fixed, introduced by the fix meant to prevent it. It's most likely to reproduce for large images or slow registries, which is precisely the case an operator raises `scanTimeout` for in the first place, and on the sidecar path it happens on a fixed 5-minute cadence regardless of that configuration.

## Affected components

- `adapters/v1/syft.go`: `SyftAdapter.CreateSBOM` — guard registered after the pull
- `pkg/sbomscanner/v1/server.go`: the SBOM-scanner RPC handler — same pattern
- `cmd/sbom-scanner/main.go`: `tempDirSweepInterval` — fixed threshold independent of per-request timeout
- `internal/tools/tools.go`: `BeginActiveTempDirUse`/`StartPeriodicTempDirSweep` — the mechanism whose documented invariant this violates

## Acceptance criteria

- A pull that is still within its own configured timeout must never have its temp directory removed by a concurrent sweep, in either process.
- `BeginActiveTempDirUse` covers the pull phase (`ResolveSource`/`resolveSource`), not just cataloguing, in both `adapters/v1/syft.go` and `pkg/sbomscanner/v1/server.go`.
- The sidecar's sweep staleness threshold no longer assumes every pull finishes within a fixed 5 minutes regardless of the caller's actual configured timeout.
- Regression test demonstrating a directory registered via `BeginActiveTempDirUse` (or an equivalent covering the pull phase) is skipped by `CleanupStaleTempDirs`/`StartPeriodicTempDirSweep` even when its mtime is older than the sweep's `olderThan`.

## Suggested direction

Move `BeginActiveTempDirUse`/the deferred `endActiveTempDirUse()` to wrap the whole operation — pull and cataloguing together — in both `adapters/v1/syft.go` and `pkg/sbomscanner/v1/server.go`, rather than registering it only once cataloguing starts. Separately, give the sidecar's sweep threshold a real relationship to the timeout a request can actually carry (e.g. take the max of `tempDirSweepInterval` and whatever `req.TimeoutSeconds` the sidecar has seen recently, or add a safety margin above the largest timeout the process is configured to accept) rather than a constant with no connection to it. Which of these (or another shape) fits the sidecar's design best is a call for whoever owns it — flagging both root causes rather than guessing at the fix.

version: main (287bfd8)

bhuvan-somisetty:
## Problem

The periodic stereoscope temp-dir sweep (`internal/tools.StartPeriodicTempDirSweep`) is meant to never touch a directory a live pull or cataloguing pass is still using. `BeginActiveTempDirUse` — the call that registers that protection, both in-process (`activeTempDirUsers`) and across the `cmd/http`/`cmd/sbom-scanner` process boundary (a shared flock lease) — was only registered once cataloguing started, *after* the image pull had already finished. The pull itself, which is what actually creates the stereoscope temp directory and downloads layers into it, ran with no protection at all.

## Root cause

`adapters/v1/syft.go` and `pkg/sbomscanner/v1/server.go` both had the same shape:

```go
src, err := registryauth.ResolveSource(...)   // the pull — creates and writes the temp dir
...
endActiveTempDirUse := tools.BeginActiveTempDirUse(os.TempDir())   // the guard — too late
err = dl.Run(func(stopper <-chan struct{}) error { ... })          // cataloguing
```

`cmd/http` and `cmd/sbom-scanner` run as separate processes sharing the same `/tmp` volume and the same `"stereoscope-"` prefix, so a sweep in *either* process could win the race against a pull still running in the *other*. A pull that's still comfortably within its own configured timeout (an operator commonly raises `scanTimeout` well past the sidecar's fixed 5-minute sweep threshold for large images) had no guard in place for exactly the phase most likely to run long.

## Fix

Move `BeginActiveTempDirUse` to cover the whole operation — pull and cataloguing together — in both files:
- Registered right before `ResolveSource`/`resolveSource`, not after.
- Released on every early-return failure branch between the pull and cataloguing (mirroring how `adapters/v1/syft.go`'s existing `pullMutex` unlock is already threaded through those branches).
- Ownership handed off to the cataloguing closure's `defer` once the pull succeeds, since that closure can outlive the caller's own return (documented `deadline.Run` behavior already relied on for `pullMutex`).

Since `activeTempDirUsers`/the cross-process lease gate the sweep's *entire* pass rather than exempting one specific directory, this closes the gap regardless of how the sweep's staleness threshold compares to any given request's timeout — no separate change to the sidecar's sweep constant is needed.

## Testing

Added `Test_syftAdapter_CreateSBOM_PullHoldsTempDirGuard` (`adapters/v1/syft_test.go`) and `TestCreateSBOM_PullHoldsTempDirGuard` (`pkg/sbomscanner/v1/server_test.go`). Both stall a real pull mid-flight against a local `httptest` registry server (the manifest request never completes until the test releases it) and, while the pull is genuinely still in progress, assert that an independent handle on the same cross-process lease file cannot take the exclusive lock a sweep needs before removing anything — then assert the lease is released once `CreateSBOM` returns.

Verified both tests actually catch the regression: with the production fix stashed out, `TestCreateSBOM_PullHoldsTempDirGuard` fails with `Should be false ... a sweep's exclusive lease must not be acquirable while a pull is still in flight`, exactly the bug this PR fixes. With the fix applied, both tests pass.

- `go build ./adapters/v1/... ./pkg/sbomscanner/v1/...` — clean
- `go test ./adapters/v1/... -run Test_syftAdapter_CreateSBOM_PullHoldsTempDirGuard -v` — PASS
- `go test ./pkg/sbomscanner/v1/... -run TestCreateSBOM_PullHoldsTempDirGuard -v` — PASS

## Impact

Closes the window where a legitimately still-running image pull can have its own temp directory deleted mid-download by a concurrent sweep, in either process sharing `/tmp`. No behavior change for the success path or existing error handling — this only widens when the existing safety mechanism is active.

Fixes #796

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented temporary files from being cleaned up while an image is still being pulled or an SBOM is being generated.
  * Ensured temporary resources are released correctly after image-pull errors, completion, or timeouts.
  * Improved reliability for SBOM generation involving image downloads.

### DIFF
--- a/adapters/v1/syft_test.go
+++ b/adapters/v1/syft_test.go
@@ -269,6 +269,12 @@ func activeScanLockPathForTest() string {
 // acquireSweepLease needs before it will remove anything - proving the guard is held for the
 // pull, not just for cataloguing afterward.
 func Test_syftAdapter_CreateSBOM_PullHoldsTempDirGuard(t *testing.T) {
+	// Both this test and its sidecar counterpart (pkg/sbomscanner/v1) contend for the same
+	// process-global lease file under os.TempDir(), and go test runs packages as concurrent
+	// processes. Redirect os.TempDir() to a directory private to this test - it reads TMPDIR
+	// at call time - so the two can never observe each other's lease.
+	t.Setenv("TMPDIR", t.TempDir())
+
 	release := make(chan struct{})
 	pullStarted := make(chan struct{})
 	var once sync.Once
@@ -305,7 +311,8 @@ func Test_syftAdapter_CreateSBOM_PullHoldsTempDirGuard(t *testing.T) {
 
 	independentLease := flock.New(activeScanLockPathForTest())
 	locked, lockErr := independentLease.TryLock()
-	if lockErr == nil && locked {
+	require.NoError(t, lockErr, "probing the sweep's exclusive lease must not itself fail")
+	if locked {
 		_ = independentLease.Unlock()
 	}
 	assert.False(t, locked, "a sweep's exclusive lease must not be acquirable while a pull is still in flight, before cataloguing has even started")
--- a/adapters/v1/syft_test.go
+++ b/adapters/v1/syft_test.go
@@ -17,6 +17,7 @@ import (
 	"net/http/httptest"
 	"net/url"
 	"os"
+	"path/filepath"
 	"runtime"
 	"strings"
 	"sync"
@@ -25,6 +26,7 @@ import (
 	"time"
 
 	"github.com/anchore/syft/syft/format/syftjson/model"
+	"github.com/gofrs/flock"
 	"github.com/kinbiko/jsonassert"
 	helpersv1 "github.com/kubescape/k8s-interface/instanceidhandler/v1/helpers"
 	"github.com/kubescape/kubevuln/core/domain"
@@ -245,6 +247,89 @@ func Test_syftAdapter_CreateSBOM_TimeoutContext(t *testing.T) {
 
 	assert.NoError(t, err)
 	assert.Equal(t, helpersv1.Incomplete, sbom.Status)
+}
+
+// activeScanLockPathForTest must match internal/tools' own unexported activeScanLockPath(os.TempDir()):
+// same directory (os.TempDir(), which BeginActiveTempDirUse is always called with here and in
+// pkg/sbomscanner/v1), same filename. There is no exported way to ask tools for this path, so it
+// is duplicated here deliberately, the same way #796's own evidence had to cite it as a literal
+// to describe the mechanism from outside internal/tools.
+func activeScanLockPathForTest() string {
+	return filepath.Join(os.TempDir(), ".kubevuln-active-scan.lock")
+}
+
+// Test_syftAdapter_CreateSBOM_PullHoldsTempDirGuard is the regression test for #796:
+// BeginActiveTempDirUse used to be registered only once the pull had already finished (right
+// before cataloguing started), leaving the pull itself - which is what actually creates the
+// stereoscope temp dir and downloads layers into it - completely unprotected against a
+// concurrent StartPeriodicTempDirSweep pass in this process or the sidecar's, sharing the same
+// os.TempDir(). This stalls a real pull mid-flight (the manifest request never completes until
+// released) and, while it's stalled, asserts that a fresh, independent handle on the same
+// cross-process lease file cannot take the exclusive lock StartPeriodicTempDirSweep's own
+// acquireSweepLease needs before it will remove anything - proving the guard is held for the
+// pull, not just for cataloguing afterward.
+func Test_syftAdapter_CreateSBOM_PullHoldsTempDirGuard(t *testing.T) {
+	release := make(chan struct{})
+	pullStarted := make(chan struct{})
+	var once sync.Once
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.URL.Path == "/v2/" {
+			w.WriteHeader(http.StatusOK)
+			return
+		}
+		// Never respond to the manifest request until the test releases it - a connection
+		// that accepts the request but stalls, not one that errors or refuses outright, so
+		// the pull is genuinely still in flight while this test inspects the lease.
+		once.Do(func() { close(pullStarted) })
+		<-release
+		w.WriteHeader(http.StatusNotFound)
+	}))
+	defer server.Close()
+
+	u, err := url.Parse(server.URL)
+	require.NoError(t, err)
+
+	adapter := NewSyftAdapter(30*time.Second, 100*1024*1024, 10*1024*1024, false, nil)
+
+	done := make(chan struct{})
+	go func() {
+		defer close(done)
+		_, _ = adapter.CreateSBOM(context.Background(), "test", "", u.Host+"/test-image:latest", domain.RegistryOptions{InsecureUseHTTP: true})
+	}()
+
+	select {
+	case <-pullStarted:
+	case <-time.After(10 * time.Second):
+		t.Fatal("pull never reached the manifest request")
+	}
+
+	independentLease := flock.New(activeScanLockPathForTest())
+	locked, lockErr := independentLease.TryLock()
+	if lockErr == nil && locked {
+		_ = independentLease.Unlock()
+	}
+	assert.False(t, locked, "a sweep's exclusive lease must not be acquirable while a pull is still in flight, before cataloguing has even started")
+
+	close(release)
+	select {
+	case <-done:
+	case <-time.After(10 * time.Second):
+		t.Fatal("CreateSBOM never returned after the pull was released")
+	}
+
+	// The lease must be fully released once CreateSBOM has returned, so a genuinely stale
+	// directory left by some other, truly abandoned run can still be swept afterward.
+	require.Eventually(t, func() bool {
+		l := flock.New(activeScanLockPathForTest())
+		ok, lockErr := l.TryLock()
+		if lockErr != nil {
+			return false
+		}
+		if ok {
+			_ = l.Unlock()
+		}
+		return ok
+	}, 5*time.Second, 50*time.Millisecond, "the temp-dir guard must be released once CreateSBOM returns")
 }
 
 // mockRegistryImage serves a minimal single-layer image so syft.GetSource resolves without a
--- a/pkg/sbomscanner/v1/server_test.go
+++ b/pkg/sbomscanner/v1/server_test.go
@@ -780,6 +780,12 @@ func activeScanLockPathForTest() string {
 // acquireSweepLease needs before it will remove anything - proving the guard is held for the
 // pull, not just for cataloguing afterward.
 func TestCreateSBOM_PullHoldsTempDirGuard(t *testing.T) {
+	// Both this test and its in-process counterpart (adapters/v1) contend for the same
+	// process-global lease file under os.TempDir(), and go test runs packages as concurrent
+	// processes. Redirect os.TempDir() to a directory private to this test - it reads TMPDIR
+	// at call time - so the two can never observe each other's lease.
+	t.Setenv("TMPDIR", t.TempDir())
+
 	release := make(chan struct{})
 	pullStarted := make(chan struct{})
 	var once sync.Once
@@ -825,7 +831,8 @@ func TestCreateSBOM_PullHoldsTempDirGuard(t *testing.T) {
 
 	independentLease := flock.New(activeScanLockPathForTest())
 	locked, lockErr := independentLease.TryLock()
-	if lockErr == nil && locked {
+	require.NoError(t, lockErr, "probing the sweep's exclusive lease must not itself fail")
+	if locked {
 		_ = independentLease.Unlock()
 	}
 	assert.False(t, locked, "a sweep's exclusive lease must not be acquirable while a pull is still in flight, before cataloguing has even started")
--- a/pkg/sbomscanner/v1/server_test.go
+++ b/pkg/sbomscanner/v1/server_test.go
@@ -16,6 +16,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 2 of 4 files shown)


adapters/v1/syft_test.go
```diff
@@ -247,6 +249,89 @@ func Test_syftAdapter_CreateSBOM_TimeoutContext(t *testing.T) {
...
+// BeginActiveTempDirUse used to be registered only once the pull had already finished (right
+// before cataloguing started), leaving the pull itself - which is what actually creates the
+// stereoscope temp dir and downloads layers into it - completely unprotected against a
+// concurrent StartPeriodicTempDirSweep pass in this process or the sidecar's, sharing the same
+// os.TempDir(). This stalls a real pull mid-flight (the manifest request never completes until
+// released) and, while it's stalled, asserts that a fresh, independent handle on the same
+// cross-process lease file cannot take the exclusive lock StartPeriodicTempDirSweep's own
+// acquireSweepLease needs before it will remove anything - proving the guard is held for the
+// pull, not just for cataloguing afterward.
+func Test_syftAdapter_CreateSBOM_PullHoldsTempDirGuard(t *testing.T) {
```
matthyx, line 271: **Blocker: this regression test shares a process-global lease path with `pkg/sbomscanner/v1/server_test.go`'s `TestCreateSBOM_PullHoldsTempDirGuard`.**

Both tests call `activeScanLockPathForTest()` → `filepath.Join(os.TempDir(), ".kubevuln-active-scan.lock")`, i.e. the real machine-wide temp dir, not an isolated one. `go test ./...` runs different packages' test binaries concurrently by default, so this test and the sidecar's equivalent can genuinely execute at the same wall-clock time and contend on the *same* lock file on disk.

Concretely, the `require.Eventually` at the end of this test (asserting the lease is fully released after `CreateSBOM` returns) can fail/flake if the other package's `PullHoldsTempDirGuard` test is still mid-flight holding its own `RLock` on the same file — a false failure unrelated to this test's own correctness.

Suggest isolating with `t.Setenv("TMPDIR", t.TempDir())` at the top of the test (before the stalled registry server starts), so each test gets its own lock file, as CodeRabbit also flagged.

adapters/v1/syft_test.go
```diff
@@ -247,6 +249,89 @@ func Test_syftAdapter_CreateSBOM_TimeoutContext(t *testing.T) {
...
+	case <-time.After(10 * time.Second):
+		t.Fatal("pull never reached the manifest request")
+	}
+
+	independentLease := flock.New(activeScanLockPathForTest())
+	locked, lockErr := independentLease.TryLock()
+	if lockErr == nil && locked {
+		_ = independentLease.Unlock()
+	}
+	assert.False(t, locked, "a sweep's exclusive lease must not be acquirable while a pull is still in flight, before cataloguing has even started")
```
matthyx, line 318: Minor: `locked` is already `false` whenever `lockErr != nil` (flock's `TryLock` returns `false` on error), so `assert.False(t, locked, ...)` passes vacuously on an unrelated filesystem error, not just on genuine contention. Worth asserting `lockErr == nil` first (`require.NoError(t, lockErr)`) so the test actually proves contention rather than papering over a broken probe.

pkg/sbomscanner/v1/server_test.go
```diff
@@ -758,6 +760,98 @@ func TestCreateSBOM_PullTimeout_ReturnsIncompleteInsteadOfHanging(t *testing.T)
...
+// used to be registered only once the pull had already finished (right before cataloguing
+// started), leaving the pull itself - which is what actually creates the stereoscope temp dir
+// and downloads layers into it - completely unprotected against a concurrent
+// StartPeriodicTempDirSweep pass in this process or the in-process adapter's, sharing the same
+// os.TempDir(). This stalls a real pull mid-flight (the manifest request never completes until
+// released) and, while it's stalled, asserts that a fresh, independent handle on the same
+// cross-process lease file cannot take the exclusive lock StartPeriodicTempDirSweep's own
+// acquireSweepLease needs before it will remove anything - proving the guard is held for the
+// pull, not just for cataloguing afterward.
+func TestCreateSBOM_PullHoldsTempDirGuard(t *testing.T) {
```
matthyx, line 782: Same blocker as `adapters/v1/syft_test.go`'s `Test_syftAdapter_CreateSBOM_PullHoldsTempDirGuard`: this test's `activeScanLockPathForTest()` resolves to the real `os.TempDir()`, the same global lock file the other package's regression test uses. Since `go test ./...` runs separate packages concurrently, the two tests can race on the same on-disk lock file and cause flaky failures in either one (particularly the trailing `require.Eventually` "lease released" check). Please isolate both with `t.Setenv("TMPDIR", t.TempDir())` before the stalled server starts.

pkg/sbomscanner/v1/server_test.go
```diff
@@ -758,6 +760,98 @@ func TestCreateSBOM_PullTimeout_ReturnsIncompleteInsteadOfHanging(t *testing.T)
...
+	case <-time.After(10 * time.Second):
+		t.Fatal("pull never reached the manifest request")
+	}
+
+	independentLease := flock.New(activeScanLockPathForTest())
+	locked, lockErr := independentLease.TryLock()
+	if lockErr == nil && locked {
+		_ = independentLease.Unlock()
+	}
+	assert.False(t, locked, "a sweep's exclusive lease must not be acquirable while a pull is still in flight, before cataloguing has even started")
```
matthyx, line 838: Same minor as the adapter test: `assert.False(t, locked, ...)` passes vacuously if `lockErr != nil`. Consider asserting `lockErr == nil` before checking `locked`.

matthyx (changes requested):
The production fix is correct: `BeginActiveTempDirUse` is now registered before `ResolveSource`/`resolveSource` in both `adapters/v1/syft.go` and `pkg/sbomscanner/v1/server.go`, and every early-return branch between registration and the `dl.Run` handoff (timeout, too-large, 401, rate-limit, platform-mismatch, generic error) calls `endActiveTempDirUse()` — I traced all of them and found no leak. The `sync.Once` wrapping mirrors the existing `pullMutex` pattern and correctly prevents double-release given `BeginActiveTempDirUse`'s own returned closure isn't safe to call twice (would double-`Unlock()` the flock).

One real blocker in the new regression tests (left as inline comments): `Test_syftAdapter_CreateSBOM_PullHoldsTempDirGuard` (adapters/v1) and `TestCreateSBOM_PullHoldsTempDirGuard` (pkg/sbomscanner/v1) both hardcode the lock path to the real `os.TempDir()` instead of an isolated per-test directory. Since `go test ./...` runs different packages' test binaries concurrently by default, these two tests can contend on the *same* on-disk lock file at the same time, which can flake either test's lease-released assertion for reasons unrelated to the code under test. Isolating with `t.Setenv("TMPDIR", t.TempDir())` in both (as CodeRabbit also suggested) would fix it. Also flagged a minor vacuous-assertion nit (`assert.False(t, locked)` passes even when the `TryLock` call itself errored) in both tests.

Once the lock-path isolation is fixed, this looks good to merge — the underlying #796 fix itself is sound.

bhuvan-somisetty:
Good catches, thanks — both fixed. Isolated each test's temp dir with `t.Setenv("TMPDIR", t.TempDir())` so they can't race on the shared `os.TempDir()` lease file when run concurrently across packages, and tightened the lease-probe assertion to `require.NoError` before checking `locked` so an unrelated filesystem error can't make it pass vacuously. Same fix applied to both files. Ready for another look.

matthyx (approved):
Both blockers from my previous review are fixed in the new commit:
- `t.Setenv("TMPDIR", t.TempDir())` added at the top of both `Test_syftAdapter_CreateSBOM_PullHoldsTempDirGuard` and `TestCreateSBOM_PullHoldsTempDirGuard`, so each test's lease file is isolated and the two can no longer contend across concurrent `go test ./...` package runs.
- `require.NoError(t, lockErr, ...)` added before the `locked` check in both tests, so the assertion no longer passes vacuously on an unrelated `TryLock` error.

Production code is unchanged from what I already verified as correct (guard registered before the pull, released on every early-return branch, ownership handed off to the cataloguing closure on success). Good to merge.

PR #799, opened 2026-08-19
ci: give golangci-lint headroom over its actual runtime

### LINKED ISSUE #798: the lint job's timeout is set to about what the job already takes
the lint job fails intermittently on a timeout rather than on a finding, and it does so because the timeout is set to roughly what the job already takes.

`Makefile`:

```make
golangci-lint run --timeout=5m --new-from-rev="$$base"
```

the last five successful runs of the `Run branch-scoped lint` step:

```
278s
284s
273s
274s
225s
```

against a 300s budget. so it finishes at 91 to 95 percent of its own timeout and clears it by 15 to 25 seconds. that is runner variance, not headroom.

#773's run crossed it:

```
2026-08-19T12:29:19  ##[group]Run PATH="$(go env GOPATH)/bin:$PATH" make lint
2026-08-19T12:35:12  level=error msg="Timeout exceeded: try increasing it by passing --timeout option"
2026-08-19T12:35:12  make: *** [Makefile:19: lint] Error 4
```

353s in the step, over the 300s golangci-lint gives itself. 17 of the last 24 lint runs exceeded 300s in wall clock.

what makes it worth fixing rather than re-running: the check goes red with `Error 4` and no findings, which reads as a lint failure. someone who did not open the log would go looking for what they broke. and a contributor working from a fork cannot re-run it:

```
run 32252812196 cannot be rerun; Must have admin rights to Repository.
```

so it needs a maintainer for something that is not a code problem.

`--timeout` is a ceiling rather than a wait, so raising it changes nothing about runs that finish in four minutes. `lint-all` carries the same 5m and lints the whole tree rather than a diff, so it has strictly more work to do and wants the same treatment.

separate observation, not part of this: no workflow in `.github/workflows/` sets `timeout-minutes`, so every job can sit at GitHub's six hour default. that is a different failure mode and I have no evidence of it biting, so it is not bundled here.

version: main (287bfd8)

AnshulPatil2005:
## Overview

The lint job is set to `--timeout=5m` and routinely finishes just under it. The last five successful runs of the branch-scoped lint step took 225s, 273s, 274s, 278s and 284s, so it completes at 91 to 95 percent of its own budget and clears it by 15 to 25 seconds. That margin is runner variance, not headroom.

#773's run crossed it at 353s and failed with `Timeout exceeded: try increasing it by passing --timeout option`.

## Additional Information

the failure is worth avoiding rather than re-running because of how it reads: the check goes red with `make: *** [Makefile:19: lint] Error 4` and no findings, which looks like a lint failure. and a contributor working from a fork cannot re-run it, `Must have admin rights to Repository`, so it needs a maintainer for something that is not a code problem.

`--timeout` is a ceiling rather than a wait, so this changes nothing about runs that finish in four minutes. it only stops the ones that would have finished in six from failing for it.

`lint-all` gets the same value. it lints the whole tree rather than a diff, so it has strictly more work to do than the branch-scoped target and carrying the tighter bound there makes less sense still.

15m rather than something tighter because the point is to stop tuning this against observed runtime: the observed runtime is what put it in this position. the ceiling only matters when something is genuinely stuck, and for that a large multiple of the real cost is the right shape.

nothing else in the recipe changes; the `merge-base` logic and `--new-from-rev` behaviour are untouched.

## How to Test

`make lint` locally, or the lint check on this PR.

The recipe's shell logic is unchanged, which can be confirmed by running the same commands it expands to:

```
base_ref=${GITHUB_BASE_REF:-main}
base=$(git merge-base HEAD upstream/$base_ref 2>/dev/null || git merge-base HEAD origin/$base_ref 2>/dev/null || true)
# -> golangci-lint run --timeout=15m --new-from-rev=<base>
```

The diff is three occurrences of the flag value and nothing else.

## Related issues/PRs:

* Resolved #798

## Summary by CodeRabbit

* **Chores**
  * Increased the linting timeout to improve reliability for larger or longer-running checks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,13 +19,13 @@ lint:
 	@base_ref=$${GITHUB_BASE_REF:-main}; \
 	base=$$(git merge-base HEAD upstream/$$base_ref 2>/dev/null || git merge-base HEAD origin/$$base_ref 2>/dev/null || true); \
 	if [ -n "$$base" ]; then \
-		golangci-lint run --timeout=5m --new-from-rev="$$base"; \
+		golangci-lint run --timeout=15m --new-from-rev="$$base"; \
 	else \
-		golangci-lint run --timeout=5m; \
+		golangci-lint run --timeout=15m; \
 	fi
 
 lint-all:
-	golangci-lint run --timeout=5m
+	golangci-lint run --timeout=15m
 
 verify: build test vet lint
 


matthyx (approved):
Straightforward, low-risk change: raises `golangci-lint --timeout` from 5m to 15m for both the branch-scoped `lint` target and `lint-all`, with no other logic touched (merge-base detection and `--new-from-rev` are untouched). This matches the stated rationale — the branch-scoped step has been completing at 91-95% of its 5m budget on recent runs and one run (#773) already timed out, which is a false-negative failure mode (looks like a lint error, isn't) that a fork contributor can't self-resolve. A 15m ceiling is a reasonable multiple that avoids re-tuning against observed runtime without meaningfully slowing down runs that already finish in ~4-5 minutes.

All CI checks are green (lint, build, tests, CodeQL, GitGuardian, DCO), and CodeRabbit's automated review found no actionable issues. No blockers — approving.

PR #805, opened 2026-08-19
fix(build): fail the offline-db build when a download fails

### LINKED ISSUE #804: the offline-db image can build successfully while shipping an error page as the database
the offline-db image can build successfully while shipping an HTTP error page instead of the vulnerability database.

```dockerfile
ARG GRYPE_DB_ROOT=https://grype.anchore.io/databases/v6
RUN curl -O $GRYPE_DB_ROOT/latest.json

RUN DB_PATH=$(jq -r '.path' latest.json) && \
    curl -O ${GRYPE_DB_ROOT}/$DB_PATH
```

curl treats an HTTP error as a successful transfer unless `--fail` is given: it exits 0 and writes the response body to the file. against the real host:

```
$ curl -sS -o probe.out https://grype.anchore.io/databases/v6/not-a-real-db.tar.zst
curl exit=0  file size=27150

$ curl -sS -f -o probe.out https://grype.anchore.io/databases/v6/not-a-real-db.tar.zst
curl exit=22
```

27150 bytes of Cloudflare error page, saved under the name the database was expected at.

the second `curl` is the last command in its RUN, so its exit status is the layer's. the build goes green, the image is pushed to `quay.io/kubescape/grype-offline-db:v6-latest` by the nightly workflow, and an air-gapped cluster serves that page to grype as its database.

the first download is caught today, but only incidentally: `jq` fails to parse the error page and exits non-zero. that is a side effect rather than a check, and it does not hold for a listing that parses but has no `.path`, where `jq -r '.path'` prints `null` and exits 0. `null` is then requested as a URL, 404s, and lands in the same place.

`--fail` on both, and an explicit check that `DB_PATH` is neither empty nor `null`, is enough.

there is no integrity check on the downloaded database either, no checksum against anything in `latest.json`. that is a larger change and is not bundled here.

separately, and also not bundled: `FROM alpine` in the builder carries no tag at all, and the final stage is `FROM busybox:latest`. the project's scorecard reports 0 of 6 container images pinned.

version: main (287bfd8)

AnshulPatil2005:
## Overview

curl treats an HTTP error as a successful transfer unless `--fail` is given: it exits 0 and writes the response body to the file. The database download is the last command in its `RUN`, so that exit status is the layer's, and a 404 there produces an image that builds clean and serves an error page to air-gapped clusters as their vulnerability database.

Verified against the real host: a 404 gives exit 0 and a 27150-byte error page saved under the requested name, where `--fail` gives exit 22 and no file.

## Additional Information

the listing download is caught today, but only incidentally: `jq` fails to parse the error page it gets handed and exits non-zero. that is a side effect rather than a check, and it does not hold for a listing that parses but carries no `.path`, where `jq -r '.path'` prints `null` and exits 0. `null` is then requested as a URL and 404s into the same silent path, so `DB_PATH` is checked explicitly rather than relying on the download to reject it.

`--location` and `--show-error` come along for the ride: a redirect is followed rather than saved as the file, and `--silent` still reports the reason on failure.

the nightly `offline-db` workflow pushes this image to `quay.io/kubescape/grype-offline-db:v6-latest`, so a bad build replaces a good one on a schedule rather than waiting for someone to notice.

two things deliberately not in scope. there is no integrity check on the downloaded database, no checksum against anything in `latest.json`, which is a larger change. and the builder stage is `FROM alpine` with no tag while the final stage is `FROM busybox:latest`, which is worth pinning but is a separate concern from this failing quietly.

## How to Test

The shell in the changed `RUN` parses as `sh`, and the guard behaves:

```
DB_PATH=""                              -> rejected
DB_PATH="null"                          -> rejected
DB_PATH="vulnerability-db_v6_....zst"   -> accepted
```

Building the image is the real test: `docker buildx build --file build/Dockerfile.offline-db .` still succeeds against the live listing, and pointing `GRYPE_DB_ROOT` at a path that 404s now fails the build instead of producing an image.

## Related issues/PRs:

* Resolved #804

### DIFF
--- a/build/Dockerfile.offline-db
+++ b/build/Dockerfile.offline-db
@@ -7,13 +7,26 @@ RUN apk add --no-cache curl jq
 # Set the working directory
 WORKDIR /app/grype-db
 
-# Download the listing.json file
+# Download the listing.json file.
+#
+# --fail on every download: without it curl treats an HTTP error as a successful transfer,
+# exits 0, and writes the error body to the file. The database download below is the last
+# command in its RUN, so its exit status is the layer's, and a 404 would ship an error page
+# to air-gapped clusters as their vulnerability database. --location follows a redirect
+# rather than saving it, and --show-error keeps the reason visible under --silent.
 ARG GRYPE_DB_ROOT=https://grype.anchore.io/databases/v6
-RUN curl -O $GRYPE_DB_ROOT/latest.json
+RUN curl --fail --location --silent --show-error -O $GRYPE_DB_ROOT/latest.json
 
-# Extract the database URL from the listing file and download it
+# Extract the database URL from the listing file and download it.
+#
+# DB_PATH is checked before use: jq prints "null" and exits 0 for a well-formed listing that
+# has no .path, which would otherwise be requested as a URL and fail the same quiet way.
 RUN DB_PATH=$(jq -r '.path' latest.json) && \
-    curl -O ${GRYPE_DB_ROOT}/$DB_PATH
+    if [ -z "$DB_PATH" ] || [ "$DB_PATH" = "null" ]; then \
+        echo "latest.json has no .path" >&2; \
+        exit 1; \
+    fi && \
+    curl --fail --location --silent --show-error -O ${GRYPE_DB_ROOT}/$DB_PATH
 
 # Stage 2: The final production image
 FROM busybox:latest


matthyx (approved):
Verified the shell logic: the bare `exit 1` inside the `if` terminates the whole `RUN`'s `sh -c` (no subshell), so the empty/`null` `DB_PATH` guard actually fails the build rather than silently continuing. `--fail --location --silent --show-error` is the right combination — `--fail` catches HTTP error status, `--location` avoids saving a redirect body, `--show-error` keeps the failure reason visible under `--silent`. This closes the exact gap described in #804 (both the "jq fails to parse the error page" case and the "valid JSON, missing `.path`" case). Scope is appropriately minimal — no integrity/checksum check and no image-tag pinning bundled in, both correctly called out as separate concerns. CI (including the actual `build-and-push-image` job) is green. No blockers.

PR #807, opened 2026-08-19
test(domain): cover the scan phase updater, which had none

### LINKED ISSUE #806: the scan phase updater has no test coverage, and twenty call sites depend on it
`WithScanPhaseUpdater` and `UpdateScanPhase` are the only two functions in `core/domain` with no coverage at all, and they are what every phase `/v1/scanStatus` reports travels through.

```
core/domain/scan_status.go:34  WithScanPhaseUpdater  0.0%
core/domain/scan_status.go:38  UpdateScanPhase       0.0%
```

twenty call sites across the four scan flows report through them, `sbom_generation`, `cve_matching`, `relevancy_lookup`, `result_storage`, `result_upload` and the rest, and the controller side wires the updater per request:

```go
bgCtx := context.WithoutCancel(domain.WithScanPhaseUpdater(ctx, func(phase string) {
    h.ensureStatuses().markPhase(newScan.JobID, phase)
}))
```

nothing asserts the round trip. a phase handed to `UpdateScanPhase` reaching the function `WithScanPhaseUpdater` stored is the entire contract, and changing the context key or the stored signature would stop all twenty reporting while every test still passed. the endpoint would keep answering, with the phase frozen wherever it last landed.

the rest of the function is refusals, and each stands between a scan and a panic or a blanked phase:

```go
func UpdateScanPhase(ctx context.Context, phase string) {
	if phase == "" {
		return
	}
	if update, ok := ctx.Value(scanPhaseUpdaterKey{}).(func(string)); ok && update != nil {
		update(phase)
	}
}
```

the `update != nil` half is the easiest to lose: `WithScanPhaseUpdater(ctx, nil)` stores a nil `func(string)`, and the type assertion succeeds on it, so that check is the only thing between a nil updater and a panic in a worker goroutine. removing it panics; removing the empty-phase check lets a caller blank the phase.

worth covering alongside them, because it is what the call sites actually do: repeated reports from concurrent goroutines against one context, which is `ScanCP` fanning a request out over several images, and a derived context overriding its parent's updater.

version: main (88535dd)

AnshulPatil2005:
## Overview

`WithScanPhaseUpdater` and `UpdateScanPhase` were the only two functions in `core/domain` with no coverage, and they are what every phase `/v1/scanStatus` reports travels through, from twenty call sites across the four scan flows.

Nothing asserted the round trip. `core/domain` goes from 28.6% to 100%.

## Additional Information

the round trip is the contract: a phase handed to `UpdateScanPhase` reaching the function `WithScanPhaseUpdater` stored. changing the context key or the stored signature would have stopped all twenty call sites reporting while every test still passed, and the endpoint would have kept answering with the phase frozen wherever it last landed.

the rest of the function is refusals, and each stands between a scan and a panic or a blanked phase: an empty phase, no updater in context, a nil updater stored, and a value of another type under the key. a scan flow calls this unconditionally, so whether anything is listening is the caller's business rather than the scan's.

both guards are load-bearing, and the tests were checked against removing them rather than only against passing:

- dropping `if phase == ""` fails the empty-phase case
- dropping `update != nil` panics the nil-updater case, since the type assertion succeeds on a nil `func(string)`

two of the cases are about what the call sites actually do rather than the function in isolation: repeated reports from concurrent goroutines against one context, which is `ScanCP` fanning a request out over several images while the request goroutine has already returned, and a derived context overriding its parent's updater.

the key being an unexported struct type is also pinned, since that is what keeps another package's context value from being read as an updater.

## How to Test

`go test ./core/domain/ -cover -count=1` reports 100%, and `go test ./core/... -race -count=1` is clean, which the concurrent case needs.

No production code changes, so nothing else moves.

## Related issues/PRs:

* Resolved #806

## Summary by CodeRabbit

* **Tests**
  * Added coverage for scan-phase update delivery and validation.
  * Verified safe behavior with missing or invalid update handlers.
  * Tested context isolation, nested handler overrides, and concurrent reporting.

### DIFF
--- a/core/domain/scan_status_test.go
+++ b/core/domain/scan_status_test.go
@@ -1,7 +1,10 @@
 package domain
 
 import (
+	"context"
 	"encoding/json"
+	"github.com/stretchr/testify/assert"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


matthyx (approved):
Reviewed against `core/domain/scan_status.go`: the round-trip test, the three refusal cases (empty phase, no updater, nil updater stored — the type assertion on a nil `func(string)` does succeed, so that guard is genuinely load-bearing), the private-key isolation test, the concurrent-reporter test, and the inner/outer context override test all match the actual guard logic in `UpdateScanPhase`/`WithScanPhaseUpdater`. Test-only change, no production code touched, all CI checks (build, lint, CodeQL, DCO, GitGuardian) are green. LGTM.

PR #809, opened 2026-08-19
docs(config): document what deploying the SBOM scanner sidecar requires

### LINKED ISSUE #808: deploying the SBOM scanner sidecar is undocumented, including a requirement that fails silently
the sidecar is documented as an architecture (#777) and as a list of environment variables, but not as a deployment. what a manifest has to get right is spread across code comments, and one of those is a correctness requirement rather than a convenience.

**both containers must mount the same volume at `/tmp`.** each process runs a periodic sweep that removes stale stereoscope temp directories, and the two coordinate through a `flock` lease on a file in that same directory:

```go
// cmd/http and cmd/sbom-scanner run as separate OS processes but mount the same /tmp volume in
// the kubevuln Pod (see the Helm chart's "tmp-dir" volume on both containers), sharing the same
// stereoscope temp-dir root on disk. activeTempDirUsers alone is invisible across that process
// boundary, so without the cross-process lease below, cmd/http's sweep could delete a directory
// an in-flight cmd/sbom-scanner cataloguing job is still reading from, or vice versa.
```

give them separate `emptyDir`s and each process gets its own lock file, so neither sees the other's lease and the coordination is silently absent. the failure is a scan dying partway through on a missing file, at whatever interval the two sweeps happen to collide. that requirement lives only in that comment, pointing at a chart that a hand-written manifest will not have.

three others that are easy to get wrong and also fail quietly:

`SOCKET_PATH` (sidecar, listening) and `SBOM_SCANNER_SOCKET` (kubevuln, dialling) are separate variables and nothing validates that they agree.

the containers share a network namespace. kubevuln binds `:8080`; the sidecar defaults to `:8081` via `METRICS_ADDR` for that reason, and pointing both at the same port means the sidecar fails to bind and its `component="sidecar"` series, including `kubevuln_temp_dir_sweep_removed_total`, are never served.

`SCANNER_MEMORY_LIMIT` wants the container's actual limit, since it is recorded on an SBOM the scanner reports as too large and is what makes that result readable after the fact.

all four fail the same way. startup is deliberately fail-soft, so a sidecar that never becomes ready leaves kubevuln running on in-process Syft with nothing but a log line, and the deployment looks fine. `/v1/diagnostics` reports `scanMode`, which is the only way to tell from outside, and that is worth documenting next to the requirements rather than leaving someone to find it.

for what it is worth, the docs here are otherwise in good shape: every config key carries an accurate default, thirteen of the fourteen metrics are covered, and the SecurityException design doc states the `affected` suppression rules correctly. this is the gap I could find that is load-bearing rather than cosmetic.

version: main (88535dd)

AnshulPatil2005:
## Overview

The sidecar is documented as an architecture (#777) and as a list of environment variables, but not as a deployment. Adds a **Deploying the SBOM Scanner Sidecar** section to `CONFIGURATION.md` covering the four things a manifest has to line up, one of which is a correctness requirement that currently appears only in a code comment.

## Additional Information

the load-bearing one is the shared `/tmp`. both processes sweep stale stereoscope temp directories and coordinate through a `flock` lease on a file in that directory, so separate `emptyDir`s give each its own lock file, neither sees the other's lease, and one container's sweep can delete a directory the other is still cataloguing from. today that is stated in a comment in `internal/tools/tools.go` pointing at the Helm chart, which a hand-written manifest will not have.

the other three fail quietly too: `SOCKET_PATH` and `SBOM_SCANNER_SOCKET` are separate variables nothing validates against each other; the containers share a network namespace so `METRICS_ADDR` must not be pointed at kubevuln's `:8080`; and `SCANNER_MEMORY_LIMIT` wants the container's real limit, being what an oversized-SBOM result is read back against.

because startup is fail-soft, all four leave kubevuln running on in-process Syft with nothing but a log line, so the section ends on checking `scanMode` from `/v1/diagnostics` and lists the usual causes in the order worth checking.

the last part covers what the sidecar does *not* change, since the concurrency difference is the one thing that surprises people: the in-process adapter serialises pulls behind a mutex and the sidecar deliberately does not, so `scanConcurrency` has more effect on disk usage in sidecar mode.

every claim is taken from code rather than inferred: both binaries being in one image and `ENTRYPOINT` being `kubevuln` (`build/Dockerfile`), the `/sbom-comm/scanner.sock` and `:8081` defaults, the lock file name, the fallback log line, and the two `scanMode` values from `domain`'s constants.

## How to Test

Docs only.

The section is added under its own `##` heading and listed in the table of contents in document order. Every in-page anchor in the file resolves, checked by extracting the headings and matching them against the links, which caught one dangling reference in an earlier draft.

The claims are checkable directly, for example:

```
grep -n "defaultMetricsAddr" cmd/sbom-scanner/main.go
grep -n "kubevuln-active-scan.lock" internal/tools/tools.go
grep -n "falling back to in-process Syft" cmd/http/main.go
```

## Related issues/PRs:

* Resolved #808


matthyx (approved):
Verified every factual claim in the new "Deploying the SBOM Scanner Sidecar" section against the current source:

- `build/Dockerfile`: both `kubevuln` and `sbom-scanner` binaries built into one image, `ENTRYPOINT ["kubevuln"]` — confirmed.
- `SOCKET_PATH` default `/sbom-comm/scanner.sock` (`cmd/sbom-scanner/main.go`) and `SBOM_SCANNER_SOCKET` read in `cmd/http/main.go` — confirmed, two independent env vars with no cross-validation.
- Shared `/tmp` / `flock` lease file `.kubevuln-active-scan.lock` (`internal/tools/tools.go` `activeScanLockFileName`) and `os.TempDir()` usage in both `cmd/http` and `cmd/sbom-scanner` — confirmed.
- `:8080` (kubevuln) vs. `:8081` default `METRICS_ADDR` (sidecar) and the `component="sidecar"` / `kubevuln_temp_dir_sweep_removed_total` metric — confirmed in `internal/metrics/metrics.go` and `cmd/sbom-scanner/main.go`.
- `SCANNER_MEMORY_LIMIT` read in `cmd/http/main.go`, threaded into `SidecarSBOMAdapter`, and recorded on the SBOM annotation when the scanner is marked `TooLarge` after repeated OOM crashes (`adapters/v1/sidecar.go`) — confirmed.
- `scanMode` values `"sidecar"` / `"in-process"` (`core/domain/diagnostics.go`) and the exact fallback log line in `cmd/http/main.go` — confirmed verbatim.
- Concurrency claim (in-process mutex-serialized vs. sidecar bound by `scanConcurrency`) matches the existing `ARCHITECTURE.md` "SBOM Generation Modes" section.
- TOC entry and all new in-page anchors (`#deploying-the-sbom-scanner-sidecar`, `ARCHITECTURE.md#sbom-generation-modes`, `#complete-schema`) resolve correctly, and the section is placed in document order.

No inaccuracies found and no blocking issues. All CI checks (build, CodeQL, GitGuardian, DCO, FOSSA) are green. Docs-only change, LGTM.

PR #811, opened 2026-08-20
docs(api): document maxQueueDepth's 503 admission-control response

### LINKED ISSUE #810: API.md documents an unbounded scan queue that no longer exists once maxQueueDepth is set
`docs/API.md` still documents the scan endpoints' backpressure behavior from before admission control existed, and now actively contradicts what the service does once `maxQueueDepth` (added in #749) is configured.

## Evidence

**"Rate Limiting" section (`docs/API.md`, near the bottom):**

```
Kubevuln implements internal concurrency control via a worker pool. The number of
concurrent scans is controlled by the `scanConcurrency` configuration option.

When the worker queue is full, requests are still accepted but will be queued for
processing.
```

**What the code actually does**, in every one of the four scan handlers:

```go
// controllers/http.go
func (h *HTTPController) GenerateSBOM(c *gin.Context) {
    ...
    if !h.admitQueueSlot(c, ctx, "generateSBOM", details) {
        return
    }
    ...
}
```

`admitQueueSlot` → `tryAdmit` rejects with **503** (`domain.ErrQueueFull`, mapped by `validationStatusCode` to `http.StatusServiceUnavailable`) once `maxQueueDepth` accepted-but-unfinished scans are already outstanding — it does not queue the request. This isn't a hidden or rare code path: `GenerateSBOM`, `ScanCP`, `ScanCVE`, and `ScanRegistry` all call `admitQueueSlot` before doing any work (`controllers/http.go:300,417,483,567`).

**Confirms the doc is simply stale, not describing an alternate mode:** `git log -- docs/API.md` shows it was last touched for `/v1/diagnostics` (#789) and the metrics table (#604), never for `maxQueueDepth`/admission control (#749, "bound the HTTP controller's scan queue with admission control"). The doc predates that feature and was never revisited after it shipped.

**The per-endpoint response tables compound the same gap.** Every scan endpoint's own "Response" table (Generate SBOM, Scan Image, Scan Registry, Application Profile Scan) lists `200`/`400`, and three of the four also list `429`, but none lists `503` — even though all four can return it. The top-level "HTTP Status Codes" table does list `503`, but scopes it to only "Service not ready (vulnerability DB not loaded)" (the `/v1/readiness` meaning) — it doesn't mention that the exact same status code, for a different reason, is a normal response from the scan endpoints themselves once queue admission control is configured.

## Root cause

`docs/API.md`'s scan-endpoint documentation was written when the worker pool queue was genuinely unbounded (every request accepted, per `#748`'s own description of the pre-fix behavior), and was never updated when `maxQueueDepth` admission control landed and changed that contract for anyone who configures it.

## Impact

`maxQueueDepth` is a real, supported `CONFIGURATION.md` option (documented there correctly, including the 503 behavior) intended for exactly the backpressure scenario this doc describes — an operator protecting kubevuln from being overwhelmed by a burst of scan requests past `scanConcurrency`. A client built against `docs/API.md` alone has no reason to expect or handle a 503 from a scan endpoint: the doc explicitly says the opposite ("requests are still accepted"). That's a real integration hazard, not a cosmetic gap — the exact failure mode `maxQueueDepth` exists to produce (deliberate rejection under load, so callers can retry or back off) is undocumented as a possible outcome at all.

## Affected components

- `docs/API.md`: the "Rate Limiting" section, and the Response tables for Generate SBOM, Scan Image for CVEs, Scan Registry Image, and Application Profile Scan
- Indirectly, `docs/CONFIGURATION.md`'s `maxQueueDepth` entry, which documents the 503 behavior correctly but isn't cross-referenced from `API.md`

## Acceptance criteria

- The "Rate Limiting" section describes both configurations accurately: unbounded (`maxQueueDepth` unset or `<= 0`) queues every accepted request, while a positive `maxQueueDepth` rejects with `503` once that many scans are outstanding, rather than queuing further.
- Each of the four scan endpoints' Response tables includes `503 Service Unavailable`, with a description distinguishing it from the readiness-probe meaning of the same code (e.g. "scan admission queue is full (`maxQueueDepth` reached)").
- The top-level "HTTP Status Codes" table's `503` row reflects both causes, or the doc otherwise makes clear a scan endpoint's `503` is unrelated to `/v1/readiness`.

## Suggested direction

Update `docs/API.md` in place: rewrite the "Rate Limiting" section against the actual `tryAdmit`/`admitQueueSlot` behavior in `controllers/http.go`, add the `503` row (with the queue-full reason) to each scan endpoint's Response table, and link to `CONFIGURATION.md`'s `maxQueueDepth` entry, which already documents this correctly, so the two documents describe the same contract instead of two different ones.

version: main (88535dd)

bhuvan-somisetty:
## Problem

`docs/API.md`'s "Rate Limiting" section claimed "when the worker queue is full, requests are still accepted but will be queued for processing," and none of the four scan endpoints' Response tables (Generate SBOM, Scan Image for CVEs, Scan Registry Image, Application Profile Scan) listed a `503` response.

## Root cause

That was accurate before `maxQueueDepth` admission control (#749) existed, but the docs were never updated after it shipped. Every scan handler in `controllers/http.go` (`GenerateSBOM`, `ScanCP`, `ScanCVE`, `ScanRegistry`) calls `admitQueueSlot` before doing any work, which rejects with `503 Service Unavailable` (`domain.ErrQueueFull`) once `maxQueueDepth` accepted-but-unfinished scans are already outstanding, instead of queuing the request. `git log -- docs/API.md` confirms the file was only ever touched for `/v1/diagnostics` (#789) and the metrics table (#604/#655), never for `maxQueueDepth`.

The top-level "HTTP Status Codes" table did list `503`, but scoped it only to the `/v1/readiness` meaning ("vulnerability DB not loaded") — not mentioning that the same code, for an unrelated reason, is also a normal scan-endpoint response.

## Fix

- Added a `503 Service Unavailable` row to each of the four scan endpoints' Response tables.
- Rewrote the "Rate Limiting" section to describe both configurations accurately: unbounded (`maxQueueDepth` unset or `<= 0`, the default) queues every accepted request, while a positive `maxQueueDepth` rejects with `503` once reached. Cross-referenced `CONFIGURATION.md`'s `maxQueueDepth` entry, which already documents the `503` behavior correctly, so the two docs describe the same contract instead of disagreeing.
- Updated the top-level status-code table's `503` row to disambiguate the two unrelated causes (readiness vs. admission control).

## Testing

This is a documentation-only change — no Go code was touched. `lint.yaml` is path-filtered to `**/*.go`/`go.mod`/`go.sum`/`Makefile` and won't run against a docs-only diff, and the repo has no markdown-lint/link-check tooling to run locally. Verified by hand:
- Every edited Markdown table's column count and pipe alignment against its surrounding rows.
- The `CONFIGURATION.md#scanning-options` anchor resolves to the actual `#### Scanning Options` heading containing `maxQueueDepth`.
- Cross-checked every claim against the current code: `controllers/http.go`'s `admitQueueSlot`/`tryAdmit`/`validationStatusCode` (503 behavior and which four handlers call it) and `config/config.go`'s `maxQueueDepth` default (`0`, unbounded).

## Impact

Doc-only; no runtime behavior changes. A client integrating against `docs/API.md` alone can now tell that a scan request may be rejected with `503` under a supported, documented configuration (`maxQueueDepth`), instead of being told the opposite.

Fixes #810

## Summary by CodeRabbit

- **Documentation**
  - Documented `503 Service Unavailable` responses when the scan admission queue reaches its configured limit.
  - Clarified the distinct causes of `503` responses for readiness checks and scan endpoints.
  - Updated rate-limiting guidance to explain bounded and unbounded queue behavior.
  - Added guidance to retry scan requests later when the queue is full.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -291,7 +291,7 @@ POST /v1/sbomCreation
 | `200 OK` | Request accepted, SBOM generation started |
 | `400 Bad Request` | Invalid request payload or validation failed |
 | `429 Too Many Requests` | Registry rate limit hit on a previous pull for this image |
-| `503 Service Unavailable` | Scan admission queue is full (`maxQueueDepth` reached); retry later. Unrelated to `/v1/readiness` |
+| `503 Service Unavailable` | Scan admission capacity is full (positive `maxQueueDepth` reached; unset or non-positive means unbounded admission); retry later. Unrelated to `/v1/readiness` |
 
 #### Example
 
@@ -345,7 +345,7 @@ POST /v1/scanImage
 | `200 OK` | Request accepted, CVE scan started |
 | `400 Bad Request` | Invalid request payload or validation failed |
 | `429 Too Many Requests` | Registry rate limit hit on a previous pull for this image |
-| `503 Service Unavailable` | Scan admission queue is full (`maxQueueDepth` reached); retry later. Unrelated to `/v1/readiness` |
+| `503 Service Unavailable` | Scan admission capacity is full (positive `maxQueueDepth` reached; unset or non-positive means unbounded admission); retry later. Unrelated to `/v1/readiness` |
 
 #### Example
 
@@ -396,7 +396,7 @@ POST /v1/scanRegistryImage
 | `200 OK` | Request accepted, registry scan started |
 | `400 Bad Request` | Invalid request payload or validation failed |
 | `429 Too Many Requests` | Registry rate limit hit on a previous pull for this image |
-| `503 Service Unavailable` | Scan admission queue is full (`maxQueueDepth` reached); retry later. Unrelated to `/v1/readiness` |
+| `503 Service Unavailable` | Scan admission capacity is full (positive `maxQueueDepth` reached; unset or non-positive means unbounded admission); retry later. Unrelated to `/v1/readiness` |
 
 #### Example
 
@@ -451,7 +451,7 @@ POST /v1/applicationProfileScan
 |--------|-------------|
 | `200 OK` | Request accepted, profile scan started |
 | `400 Bad Request` | Invalid request payload or validation failed |
-| `503 Service Unavailable` | Scan admission queue is full (`maxQueueDepth` reached); retry later. Unrelated to `/v1/readiness` |
+| `503 Service Unavailable` | Scan admission capacity is full (positive `maxQueueDepth` reached; unset or non-positive means unbounded admission); retry later. Unrelated to `/v1/readiness` |
 
 #### Example
 
@@ -552,7 +552,7 @@ All responses follow RFC 7807.
 | `400` | Bad Request | Invalid JSON, missing required fields, or validation failed |
 | `429` | Too Many Requests | Registry rate limit hit on a previous pull for this image |
 | `500` | Internal Server Error | Internal error |
-| `503` | Service Unavailable | Two distinct causes, unrelated to each other: `/v1/readiness` reports it when the vulnerability DB isn't loaded, and a scan endpoint reports it when the admission queue is full (`maxQueueDepth` reached) |
+| `503` | Service Unavailable | Two distinct causes, unrelated to each other: `/v1/readiness` reports it when the vulnerability DB isn't loaded, and a scan endpoint reports it when scan admission capacity is full (positive `maxQueueDepth` reached) |
 
 ### Common Errors
 
--- a/docs/API.md
+++ b/docs/API.md
@@ -291,6 +291,7 @@ POST /v1/sbomCreation
 | `200 OK` | Request accepted, SBOM generation started |
 | `400 Bad Request` | Invalid request payload or validation failed |
 | `429 Too Many Requests` | Registry rate limit hit on a previous pull for this image |
+| `503 Service Unavailable` | Scan admission queue is full (`maxQueueDepth` reached); retry later. Unrelated to `/v1/readiness` |
 
 #### Example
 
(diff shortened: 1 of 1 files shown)


bhuvan-somisetty:
Good point, fixed — reworded to make clear 503 only shows up once maxQueueDepth is a positive value, and switched "admission queue is full" to "admission capacity is full" since the limit counts running scans too, not just queued ones. Same wording now across all four endpoint tables and the top-level status table.

matthyx (approved):
Verified every claim against the current code:

- `controllers/http.go`: `tryAdmit()`/`release()` confirm `maxQueueDepth <= 0` is unbounded, and `pending` counts accepted-but-unfinished scans (queued + running), matching the new "positive value... includes running scans" wording.
- `docs/CONFIGURATION.md`'s `maxQueueDepth` entry (default `0`) agrees with the rewritten Rate Limiting section, and the `#scanning-options` anchor resolves to the real `#### Scanning Options` heading.
- `cmd/http/main.go` confirms all four scan routes (`sbomCreation`, `scanImage`, `scanRegistryImage`, `applicationProfileScan`) go through `admitQueueSlot`, so each getting its own `503` row is correct.
- Table formatting (column counts, backtick style) is consistent with the surrounding rows in all four Response tables and the top-level status-code table.

Doc-only change, no behavior touched, and it already resolves CodeRabbit's earlier wording nit ("admission capacity" vs "admission queue"). Looks good to merge.

Issue #676, opened 2026-08-13
adapters/v1 test suite is not hermetic: full package run times out due to real registry pulls


rootp1:
# Description
`go test ./adapters/v1` is not reliably hermetic. Running the full package suite times out, while an isolated repro of the named subtest passes quickly. The cause is that `Test_syftAdapter_CreateSBOM` mixes many table-driven subtests that pull real images from public registries (Docker Hub, quay.io, gcr.io, registry.k8s.io) into the default unit test run, so overall latency depends on network conditions rather than being hermetic.

# Environment
OS: `darwin/arm64`
Version: `go1.26.5` workspace run

# Steps To Reproduce
1. Run `go test ./adapters/v1 -count=1 -timeout 120s`.
2. Observe the timeout failure, with `panic: test timed out after 2m0s` and the stack trace showing `Test_syftAdapter_CreateSBOM` still running near the deadline.
3. Run `go test ./adapters/v1 -run 'Test_syftAdapter_CreateSBOM/public_image_with_invalid_credentials_falls_back_to_unauthenticated' -count=1 -timeout 45s -v`.
4. Observe that the isolated path passes in under 10 seconds, indicating the timeout comes from cumulative network latency across many subtests rather than one deterministic failure.

# Expected behavior
`go test ./adapters/v1 -count=1 -timeout 120s` should complete quickly and hermetically in a normal developer environment, without depending on external registries being reachable and responsive.

# Actual Behavior
The full package run can time out after 120 seconds even though an isolated repro of a single named subtest passes reliably. The suite currently has 15 subtests in `Test_syftAdapter_CreateSBOM` that each resolve and pull a real image from a public registry.

# Additional context
Suspected root cause: `Test_syftAdapter_CreateSBOM` in `adapters/v1/syft_test.go` is an integration test disguised as a unit test - it exercises real registry pulls across many scenarios (Docker Hub, quay.io, gcr.io, registry.k8s.io) in the default `go test` path, with no separation from hermetic unit coverage.

Proposed direction: move the network-dependent subtests into a file gated behind a `//go:build integration` tag (or an equivalent dedicated test target), so the default `go test ./adapters/v1` run only exercises hermetic tests, and contributors can opt into the slower, network-dependent suite explicitly with `go test -tags=integration ./adapters/v1`.

Acceptance Criteria:
- `go test ./adapters/v1 -count=1 -timeout 120s` passes consistently in a normal developer environment.
[...]

PR #813, opened 2026-08-20
fix(adapters): give cloud exceptions precedence over CRD ones on overlap

### LINKED ISSUE #812: GetCVEExceptions doesn't implement the documented cloud-over-CRD precedence on overlapping CVEs
`docs/security-exception-design.md`'s "Conflict Resolution & Precedence" section specifies exactly how `GetCVEExceptions()` should merge cloud and CRD exceptions when both cover the same CVE on the same workload — and the implementation doesn't do what it says.

## Evidence

**The documented contract** (`docs/security-exception-design.md`, "Conflict Resolution & Precedence"):

> **Cloud exceptions take precedence.** If the cloud backend has an exception for a given CVE on a workload, the cloud exception's status and metadata are used, and any conflicting CRD exception for the same CVE on the same workload is ignored.
>
> **Implementation**: During the merge in `GetCVEExceptions()`, cloud exceptions are added first. CRD exceptions are only appended for CVE+workload combinations not already covered by a cloud exception.

**The actual `GetCVEExceptions()`** (`adapters/v1/backend.go:282-303`):

```go
vulnExceptionList, err := a.getCVEExceptionsFunc(a.apiServerRestURL, a.clusterConfig.AccountID, &designator, a.getRequestHeaders())
...
if len(seList) > 0 || len(cseList) > 0 {
    target := BuildExceptionTarget(ctx, workload, seList, cseList, a.securityExceptionRepo)
    crdPolicies, crdStats := ConvertToVulnerabilityExceptionPolicies(seList, cseList, target)
    ...
}
```

`crdPolicies` is appended unconditionally — there is no check of which CVEs `vulnExceptionList` (the cloud result) already covers, and no filtering of `crdPolicies` against it. The "CRD exceptions are only appended for CVE+workload combinations not already covered by a cloud exception" step the design doc describes does not exist anywhere in this function, or anywhere downstream of it.

## Root cause

The merge was implemented as a plain concatenation of both sources' policy lists, not the precedence-filtered merge the design doc specifies and the review doc (`docs/security-exception-design-review.md`, point 4) explicitly called for on compliance-audit grounds ("this matters for compliance audits — you need deterministic, documentable behavior").

## Impact

Downstream, both `ApplySecurityExceptions` (`adapters/v1/securityexception.go`) and the manifest's provenance recording key on *every* policy that carries the Ignore action for a matched CVE, not just one:

- `buildIgnoreRules` (`adapters/v1/securityexception.go:371`) calls `suppressingPolicies(matched)`, which returns every Ignore-actioned policy matching the CVE, and appends one `IgnoreRule` per policy. When cloud and a CRD both cover the same CVE, the stored `VulnerabilityManifest` ends up with two `IgnoreRule` entries instead of one — the CRD-sourced one the design doc says should be ignored is still recorded as if it were an equally-valid suppression reason.
- `recordExceptionsMatched` (`core/services/scan.go`) increments `kubevuln_exceptions_matched_total{sourceKind="SecurityException"}` for that CRD entry too, over-counting CRD-attributed suppressions for CVEs that, per the documented precedence, cloud alone should be credited with.

This is exactly the audit/compliance-traceability scenario the design review flagged as the reason precedence needed to be deterministic in the first place: an operator (or a compliance tool reading `IgnoreRule.SourceKind`/`SourceName`) auditing why a CVE was suppressed on a given workload sees a CRD-sourced justification the design says should never have been recorded once cloud already covers it, and the CRD-suppression metric is inflated by exceptions the documented behavior says CRDs don't actually own.

Not a functional/security regression — a CVE that either source says to suppress still ends up suppressed either way, so no finding is wrongly hidden or wrongly shown. The gap is specifically in provenance and attribution correctness for an overlap case the design doc treats as a first-class, deliberately-specified behavior.

## Affected components

- `adapters/v1/backend.go`: `BackendAdapter.GetCVEExceptions` — the merge point that needs the missing filter
- `adapters/v1/securityexception.go`: `buildIgnoreRules`, `ApplySecurityExceptions` — consumers that currently see both sources for an overlapping CVE
- `core/services/scan.go`: `recordExceptionsMatched` — the metric affected by the missing filter

## Acceptance criteria

- `GetCVEExceptions` filters `crdPolicies` to exclude any CVE (by `VulnerabilityPolicies[].Name`, matching how `cveExceptionIndex` already keys lookups) already covered by a policy in the cloud-fetched `vulnExceptionList`, before appending, matching the design doc's specified implementation.
- A CVE covered by both a cloud and a CRD exception ends up with exactly one contributing policy (cloud's) reaching `ApplySecurityExceptions`, not both.
- `IgnoreRule` provenance recorded on the manifest for such a CVE reflects only the cloud exception, per "the cloud exception's status and metadata are used."
- `kubevuln_exceptions_matched_total{sourceKind="SecurityException"}` is not incremented for a CVE whose suppression cloud already owns.
- Regression test in `adapters/v1/backend_test.go` (or equivalent) covering: cloud exception + CRD exception for the same CVE → only the cloud policy reaches the returned list; cloud exception for CVE-A + CRD exception for CVE-B (non-overlapping) → both are present, matching the doc's "non-overlapping exceptions are merged" clause, which the current unconditional-append code does happen to satisfy today.

## Suggested direction

In `GetCVEExceptions`, after fetching `vulnExceptionList` from the cloud and before appending `crdPolicies`, build a set of CVE names already present in `vulnExceptionList` (via `VulnerabilityPolicies[].Name`, case-normalized the same way `cveExceptionIndex` does) and filter `crdPolicies` down to entries whose vulnerability name isn't in that set. This keeps the fix localized to the one function the design doc already names as owning this behavior.

version: main (88535dd)

bhuvan-somisetty:
## Problem

`docs/security-exception-design.md`'s "Conflict Resolution & Precedence" section specifies exactly how `GetCVEExceptions()` should merge cloud and CRD exceptions when both cover the same CVE on the same workload — the implementation didn't do it.

## Root cause

The documented contract:

> **Cloud exceptions take precedence.** If the cloud backend has an exception for a given CVE on a workload, the cloud exception's status and metadata are used, and any conflicting CRD exception for the same CVE on the same workload is ignored.
>
> **Implementation**: During the merge in `GetCVEExceptions()`, cloud exceptions are added first. CRD exceptions are only appended for CVE+workload combinations not already covered by a cloud exception.

The actual code (`adapters/v1/backend.go`) appended `crdPolicies` unconditionally:

```go
```

There was no check of which CVEs the cloud-fetched `vulnExceptionList` already covered. Downstream, `buildIgnoreRules` records one `IgnoreRule` per Ignore-actioned policy matching a CVE, and `recordExceptionsMatched` increments `kubevuln_exceptions_matched_total{sourceKind="SecurityException"}` per matched CRD-attributed policy — so an overlapping CVE ended up with two provenance entries and an inflated CRD-suppression count, both contradicting the documented "conflicting CRD exception is ignored." Not a suppression-correctness bug (the CVE was suppressed either way, by whichever source), but a real gap in the audit/provenance guarantee the design doc treats as a first-class, compliance-motivated behavior.

## Fix

Added `excludeCloudCoveredPolicies` in `adapters/v1/backend.go`, called between converting CRD exceptions and appending them to the result:

```go
crdPolicies, crdStats := ConvertToVulnerabilityExceptionPolicies(seList, cseList, target)
crdPolicies = excludeCloudCoveredPolicies(crdPolicies, vulnExceptionList)
```

It reuses the existing `cveExceptionIndex`/`buildCVEExceptionIndex` (already the canonical lower-cased CVE/alias name index used elsewhere in this file) to build a set of CVEs the cloud list covers, then drops any CRD policy whose `VulnerabilityPolicies[].Name` matches one of them — including via alias, so a CRD exception naming a CVE by an alias the cloud exception names by its canonical ID (or vice versa) is still recognized as the same CVE. A CRD policy not covered by any cloud policy passes through unchanged, preserving the doc's "non-overlapping exceptions are merged" behavior (already correct before this change).

## Testing

- `TestExcludeCloudCoveredPolicies` — table-driven unit test of the new helper directly: no cloud policies (pass-through), no CRD policies, no overlap, exact overlap, case-insensitive overlap, alias overlap, and a multi-name CRD policy dropped entirely when any one of its names overlaps.
- `TestGetCVEExceptions_CloudExceptionTakesPrecedenceOverCRD` — the regression test for #812: cloud and CRD both cover `CVE-2024-0001`, CRD alone covers `CVE-2024-0002`. Asserts `CVE-2024-0001` appears exactly once in the result (the cloud's own policy, identified by its rule name — not merely deduplicated to *some* entry) and `CVE-2024-0002` still comes through.
- Existing `TestGetCVEExceptions_MergesCRDExceptions` and `TestGetCVEExceptions_ScopesCRDByMatch` still pass unmodified, confirming the non-overlapping merge and match-scoping behavior are unaffected.

Ran:
- `go build ./adapters/v1/...` — clean
- `go vet ./adapters/v1/...` — clean
- `go test ./adapters/v1/... -run "TestExcludeCloudCoveredPolicies|TestGetCVEExceptions_CloudExceptionTakesPrecedenceOverCRD|TestGetCVEExceptions_MergesCRDExceptions|TestGetCVEExceptions_ScopesCRDByMatch" -v` — all pass

I did not run the full `go test ./adapters/v1/...` package suite to completion — it isn't hermetic (#676: many `Test_syftAdapter_CreateSBOM` subtests pull real images from public registries and can time out on network conditions unrelated to any change in the package), so I ran the targeted tests above instead, which is what actually exercises this change.

## Impact

`GetCVEExceptions` now matches its own documented merge contract. No change to suppression outcomes for any CVE (a CVE either source flags for suppression is still suppressed) — this only corrects which source's provenance is recorded and counted when both sources cover the same CVE.

Fixes #812

## Summary by CodeRabbit

* **Bug Fixes**
  * Cloud-defined CVE exceptions now take precedence over conflicting CRD-defined exceptions.
  * Non-conflicting CRD exceptions continue to be included.
  * CVE matching now consistently handles case differences and aliases.

* **Tests**
  * Added coverage for exact, case-insensitive, and alias-based CVE overlaps.

### DIFF
--- a/adapters/v1/backend.go
+++ b/adapters/v1/backend.go
@@ -299,6 +299,7 @@ func (a *BackendAdapter) GetCVEExceptions(ctx context.Context) (domain.CVEExcept
 	if len(seList) > 0 || len(cseList) > 0 {
 		target := BuildExceptionTarget(ctx, workload, seList, cseList, a.securityExceptionRepo)
 		crdPolicies, crdStats := ConvertToVulnerabilityExceptionPolicies(seList, cseList, target)
+		crdPolicies = excludeCloudCoveredPolicies(crdPolicies, vulnExceptionList)
 		vulnExceptionList = append(vulnExceptionList, crdPolicies...)
 		stats = crdStats
 
@@ -326,6 +327,42 @@ func (a *BackendAdapter) GetCVEExceptions(ctx context.Context) (domain.CVEExcept
 	}
 
 	return vulnExceptionList, stats, nil
+}
+
+// excludeCloudCoveredPolicies drops CRD-derived policies that cover a CVE the cloud-fetched
+// list already covers, per docs/security-exception-design.md's "Conflict Resolution &
+// Precedence" section: "If the cloud backend has an exception for a given CVE on a workload,
+// the cloud exception's status and metadata are used, and any conflicting CRD exception for
+// the same CVE on the same workload is ignored." Without this, GetCVEExceptions returned both
+// sources' policies for an overlapping CVE, so the manifest ended up with a CRD-attributed
+// IgnoreRule and exceptions_matched_total{sourceKind="SecurityException"} count that the
+// documented precedence says should not exist once cloud already owns that suppression (#812).
+//
+// Reuses cveExceptionIndex's own name normalization (lower-cased CVE ID/alias) so a CRD
+// exception naming a CVE by an alias the cloud exception names by its canonical ID, or vice
+// versa, still counts as the same CVE. A CRD policy not covered by any cloud policy passes
+// through unchanged, matching the doc's separate "non-overlapping exceptions are merged"
+// clause.
+func excludeCloudCoveredPolicies(crdPolicies, cloudPolicies []armotypes.VulnerabilityExceptionPolicy) []armotypes.VulnerabilityExceptionPolicy {
+	if len(cloudPolicies) == 0 || len(crdPolicies) == 0 {
+		return crdPolicies
+	}
+	cloudIndex := buildCVEExceptionIndex(cloudPolicies)
+
+	var filtered []armotypes.VulnerabilityExceptionPolicy
+	for _, p := range crdPolicies {
+		coveredByCloud := false
+		for _, vp := range p.VulnerabilityPolicies {
+			if len(cloudIndex.byName[strings.ToLower(vp.Name)]) > 0 {
+				coveredByCloud = true
+				break
+			}
+		}
+		if !coveredByCloud {
+			filtered = append(filtered, p)
+		}
+	}
+	return filtered
 }
 
 // ReportError reports the given error to the platform
--- a/adapters/v1/backend_test.go
+++ b/adapters/v1/backend_test.go
@@ -1067,6 +1067,114 @@ func TestGetCVEExceptions_MergesCRDExceptions(t *testing.T) {
 	assert.Empty(t, stats.ExpiredBySource, "nothing expired in this scenario")
 }
 
+// TestGetCVEExceptions_CloudExceptionTakesPrecedenceOverCRD is the regression test for #812:
+// docs/security-exception-design.md documents that when both a cloud and a CRD exception
+// cover the same CVE on the same workload, the cloud exception's status/metadata are used and
+// the conflicting CRD exception is ignored. GetCVEExceptions used to append CRD policies
+// unconditionally, so a CVE covered by both sources reached the caller (and, downstream, the
+// manifest's IgnoreRule provenance and the exceptions_matched_total metric) twice.
+func TestGetCVEExceptions_CloudExceptionTakesPrecedenceOverCRD(t *testing.T) {
+	cloudPolicies := []armotypes.VulnerabilityExceptionPolicy{
+		{
+			PortalBase:            armotypes.PortalBase{Name: "cloud-rule"},
+			PolicyType:            "vulnerabilityExceptionPolicy",
+			VulnerabilityPolicies: []armotypes.VulnerabilityPolicy{{Name: "CVE-2024-0001"}},
+		},
+	}
+
+	mockRepo := &mockSecurityExceptionRepo{
+		exceptions: []sev1beta1.SecurityException{
+			{
+				ObjectMeta: metav1.ObjectMeta{Name: "crd-rule", Namespace: "default"},
+				Spec: sev1beta1.SecurityExceptionSpec{
+					Vulnerabilities: []sev1beta1.VulnerabilityException{
+						// Same CVE the cloud already covers.
+						{Vulnerability: sev1beta1.VulnerabilityRef{ID: "CVE-2024-0001"}, Status: sev1beta1.VulnerabilityStatusNotAffected},
+						// Not covered by cloud: must still come through.
+						{Vulnerability: sev1beta1.VulnerabilityRef{ID: "CVE-2024-0002"}, Status: sev1beta1.VulnerabilityStatusNotAffected},
+					},
+				},
+			},
+		},
+	}
+
+	a := &BackendAdapter{
+		clusterConfig: armometadata.ClusterConfig{AccountID: "test-account"},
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Verified against `docs/security-exception-design.md`'s "Conflict Resolution & Precedence" section: cloud exceptions are added first, and `excludeCloudCoveredPolicies` now drops CRD policies whose CVE (by ID or alias) is already covered, matching the documented "conflicting CRD exception is ignored" contract.

Checked correctness of the implementation:
- `excludeCloudCoveredPolicies` reuses `buildCVEExceptionIndex`'s existing lower-cased name index, so matching stays consistent with how the rest of the file (e.g. `ApplySecurityExceptions`) already resolves CVE identity by ID/alias, case-insensitively.
- `buildPolicy` (adapters/v1/securityexception.go) produces one `VulnerabilityExceptionPolicy` per single CVE entry (ID + that CVE's own aliases), never bundling unrelated CVEs together — so dropping a whole CRD policy when any one of its names overlaps cloud coverage can't accidentally drop an unrelated CVE.
- No downstream stats/metrics inconsistency: `domain.ExceptionStats` only tracks `ExpiredBySource` (computed before filtering, unrelated to cloud-overlap), so filtering `crdPolicies` after `crdStats` is computed doesn't skew any counter.
- `MockPlatform` (adapters/mockplatform.go) has no cloud fetch at all, so it can't hit this precedence conflict — no fix needed there for consistency.

Test coverage is solid: `TestExcludeCloudCoveredPolicies` covers pass-through, no-overlap, exact/case-insensitive/alias overlap, and multi-name-policy-dropped-on-any-overlap; `TestGetCVEExceptions_CloudExceptionTakesPrecedenceOverCRD` is a real regression test asserting the surviving policy is the cloud's own (by rule name), not merely deduplicated. Existing merge/scoping tests are untouched and still pass.

All CI checks are green (build, lint, CodeQL, GitGuardian, DCO, cross-platform build) and CodeRabbit found no actionable issues. LGTM — approving.

PR #815, opened 2026-08-20
feat: resolve CSAF composite product IDs to real purls

### LINKED ISSUE #814: Red Hat's real CSAF notes can't be matched, because they never give a real package barcode (purl)
### What's happening

Red Hat writes real notes about bugs (in a format called CSAF). But instead
of naming the exact package using its real, proper name (called a "purl" 
like a barcode for software), Red Hat's notes only give a made-up, combined
nickname  like `red_hat_enterprise_linux_10:xz`.

The scanning tool this project already uses (grype) tries to match findings
by comparing real purls. Since Red Hat's notes don't give a real purl at
all, the matching always fails silently  even when Red Hat has clearly
said "this bug doesn't affect us."

### Real proof

I tested this myself with two separate, real, downloaded Red Hat advisories
(not made-up test data). In both, **0 of the products even had a real purl
attached**  only the made-up nickname. I confirmed grype's matcher fails on
both.

### I also checked ,is there already a tool that fixes this?

I found that a library already used in this project (`openvex/go-vex`) has
a smart tool called `PurlMatches` that can compare two real purls, even if
they're written slightly differently. This is a good tool  but it needs
**two real purls to compare**. Since Red Hat's notes never give a real purl
in the first place, this existing tool has nothing to work with here. It
doesn't solve this specific problem.

### What's actually missing

A small, real translator: something that reads Red Hat's made-up nickname
(`red_hat_enterprise_linux_10:xz`) and figures out the real package it's
talking about, then builds a proper, real purl from it  so the existing
matching tools (grype's own matcher, or `PurlMatches`) actually have
something real to compare.

### Why this matters

Without this, the entire "read Red Hat's notes" half of this project cannot
work , no matter how well everything else (fetching, saving, going quiet)
is built. This is a real, confirmed blocker, not a theoretical one.

Shreya2005-2005:
## Overview

grype's own CSAF matcher (`grype/grype/vex/csaf`) compares purls by exact
string equality, but only ever looks for a purl directly on a
`relationships` entry - which real Red Hat advisories never populate. A live
check against a real, unmodified advisory (CVE-2024-3094, the XZ Utils
backdoor) confirms this directly: **0 of 26 relationships carry a purl
there.**

The real purl exists elsewhere in the same document: each relationship's
`product_reference` names a bare product ID (e.g. `"xz"`), and that bare ID
has its own entry deeper in `product_tree.branches`, carrying a real
`product_identification_helper.purl`. This adds a standalone resolver that
walks both structures and connects them.

## Before / After

Real, live output against the same real document:

**Before** - what grype's matcher actually sees:
```
Relationships with NO purl: 26 of 26
-> grype's matcher: 0 usable purls, every match silently fails to suppress
```

**After** - this resolver, same document:
```
red_hat_enterprise_linux_10:xz         -> pkg:rpm/redhat/xz
red_hat_enterprise_linux_10:xz-devel   -> pkg:rpm/redhat/xz-devel
red_hat_enterprise_linux_10:xz-libs    -> pkg:rpm/redhat/xz-libs
Composite IDs resolved to real purls: 26 of 26
```

## Scope

This does **not** wire CSAF matching into a real scan itself, and does not
modify grype - it is a standalone translator that anything doing that wiring
can use to get a real, comparable purl for a composite product ID first.

## Testing

5 tests, all passing:
- Full end-to-end run against the real advisory - all 26 composite IDs
  resolve correctly
- One specific, hand-verified resolution
- A genuinely unknown composite ID is correctly rejected
- A nil product tree never panics
- The honest edge case where a relationship exists but its product
  genuinely has no purl in the document

Full build, `go vet`, and `go test ./internal/csafresolve/...` pass with no
failures.

## Related issues/PRs

* Resolved #814

```

## Summary by CodeRabbit

* **New Features**
  * Improved CSAF product resolution across complete product trees, including top-level product names and nested branches.
  * Maps composite product identifiers to package URLs more reliably.
  * Safely handles incomplete product-tree data and reports missing relationships or package URLs clearly.

* **Tests**
  * Added coverage for real advisory data, XZ package resolution, unknown identifiers, missing package URLs, top-level product names, and empty product trees.

### DIFF
--- a/internal/csafresolve/csafresolve.go
+++ b/internal/csafresolve/csafresolve.go
@@ -7,10 +7,14 @@
 // 26 relationships in a real, downloaded advisory for CVE-2024-3094 carry a
 // purl there). The real purl exists elsewhere in the same document: each
 // relationship's product_reference names a bare product ID (e.g. "xz"), and
-// that bare ID has its own entry deeper in product_tree.branches, carrying
-// a real product_identification_helper.purl. This package walks both
-// structures and connects them - verified against a real, unmodified
-// advisory, where all 26 composite IDs resolved successfully.
+// that bare ID has its own real entry with a purl - either nested under
+// product_tree.branches, or declared directly under the separate top-level
+// product_tree.full_product_names field. This package resolves the
+// composite ID down to that bare reference, then reuses gocsaf/csaf's own
+// ProductTree.CollectProductIdentificationHelpers to find its purl,
+// wherever in the document it is actually declared - rather than
+// hand-walking the tree ourselves and risking missing a location the
+// upstream package already knows about.
 //
 // This does not fix grype's matcher itself - it is a standalone translator
 // that anything wiring CSAF matching into a real scan can use to get a real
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: internal/csafresolve/csafresolve.go [hunk 1]
```diff
@@ -0,0 +1,115 @@
...
+				continue
+			}
+			compositeID := string(*rel.FullProductName.ProductID)
+			ref := string(*rel.ProductReference)
+			r.compositeToRef[compositeID] = ref
+		}
+	}
+
+	for _, b := range tree.Branches {
+		r.walkBranch(b)
```
matthyx: **Blocker: `refToPURL` is only populated from `tree.Branches`, but CSAF also allows products to live in `tree.FullProductNames` — a separate top-level field this code never reads.**

`csaf.ProductTree` has both `Branches` and `FullProductNames *FullProductNames` (see [advisory.go](https://github.com/gocsaf/csaf/blob/main/csaf/advisory.go)):

```go
type ProductTree struct {
	Branches         Branches          `json:"branches,omitempty"`
	FullProductNames *FullProductNames `json:"full_product_names,omitempty"`
	ProductGroups    *ProductGroups    `json:"product_groups,omitempty"`
	RelationShips    *Relationships    `json:"relationships,omitempty"`
}
```

`product_tree.full_product_names` is a legitimate, spec-compliant alternative location for a product entry (with its own `product_id` and `product_identification_helper.purl`), independent of nesting under `branches`. The upstream `gocsaf/csaf` package already knows this — its own `ProductTree.FindProductIdentificationHelpers` / `CollectProductIdentificationHelpers` (in `csaf/util.go`) checks `pt.FullProductNames`, `pt.Branches` (recursively), *and* `pt.RelationShips` for a match.

This resolver's `New()` only calls `r.walkBranch(b)` over `tree.Branches` — it never looks at `tree.FullProductNames`. For any advisory whose `product_reference` targets resolve to a bare product ID declared under `full_product_names` rather than nested in `branches`, `Resolve` will incorrectly return `ErrNoPURL` even though the purl is present in the document. It happens not to trigger against the one bundled Red Hat testdata file (which puts everything under `branches`), so the tests don't catch it — but it's a real, silent under-resolution for any other advisory shape (this isn't Red-Hat-specific tooling per the package doc comment).

Suggested fix: reuse the upstream helper instead of hand-rolling the tree walk, e.g. build `refToPURL` via `tree.CollectProductIdentificationHelpers(ref)` for each `ref` you need (or replicate its full-product-names + branches walk in `New`). That also removes ~15 lines of duplicated recursive-walk logic that already exists (and is tested) upstream.

matthyx (changes requested):
Nice, well-tested standalone package overall — the diagnosis of grype's CSAF matcher gap and the real-advisory test fixture are genuinely useful. One correctness blocker before this is safe to merge: the resolver only walks `product_tree.branches` for purls and misses `product_tree.full_product_names`, a separate top-level location the CSAF format allows and that the upstream `gocsaf/csaf` package's own helpers (`FindProductIdentificationHelpers`/`CollectProductIdentificationHelpers`) already account for. See inline comment for details and a suggested fix (reuse the upstream helper instead of the hand-rolled walk). CI, lint, and the included tests all pass — this is purely about a document shape the current test fixture doesn't exercise.

Shreya2005-2005:
@matthyx  Fixed  , thanks for catching this replaced the hand-rolled branches-only
walk with gocsaf's own CollectProductIdentificationHelpers, which checks
full_product_names, branches, and relationships. Added a test proving a
purl under full_product_names now resolves correctly (it didn't before).
All 6 tests pass, build/vet clean.

matthyx (approved):
Confirmed the fix commit (1e6c5e0) resolves the blocker from my previous review: `Resolve` now delegates to `tree.CollectProductIdentificationHelpers(ref)` — the upstream `gocsaf/csaf` walker that covers `full_product_names`, `branches` (recursively), and `relationships` — instead of the hand-rolled branches-only walk. The new `TestResolve_PURLUnderFullProductNames_NotBranches` test specifically proves the previously-missing case now resolves. Also nice: this removed the duplicated tree-walking logic entirely, so the resolver now can't drift out of sync with the upstream package's notion of "where a purl can live."

Reviewed the rest of the diff again too — nil-safety, error semantics (`ErrNoRelationship` vs `ErrNoPURL`), and the composite-ID/relationship mapping are all still correct. CI, lint, and tests are green.

Approving.

PR #817, opened 2026-08-20
fix(services): report a failed container scan from ScanCP instead of nil

### LINKED ISSUE #816: ScanCP swallows every per-container failure and reports success on /v1/scanStatus and scan metrics
## Problem

`ScanCP` (the ContainerProfile / relevancy scan flow, wired to `POST /v1/applicationProfileScan`) iterates the containers of a workload and, on any per-container failure, logs the error and `continue`s to the next container. After the loop it **unconditionally returns `nil`**:

```go
// core/services/scan.go:228-406
for _, scan := range scans {
    ...
    sbom, _, sbomErr = s.getOrCreateSBOM(ctx, subWorkload)
    if sbomErr != nil {
        logger.L().Ctx(ctx).Error("creating SBOM, skipping scan", helpers.Error(sbomErr), ...)
        continue // we need the SBOM
    }
    ...
    cve, err = s.cveScanner.ScanSBOM(ctx, sbom)
    if err != nil {
        _ = s.platform.ReportScanFailure(ctx, scanfailure.ScanFailureCVE, scanfailure.ReasonCVEMatchingFailed, err)
        continue // we need the CVE
    }
    ...
    err = s.platform.SubmitCVE(ctx, cve, cvep)
    if err != nil {
        _ = s.platform.ReportScanFailure(ctx, scanfailure.ScanFailureBackendPost, scanfailure.ReasonResultUploadFailed, err)
        continue // we need to submit the CVE
    }
}

logger.L().Info("scan complete", ...)
return nil // always nil, regardless of how many containers above failed
```

There are 7 distinct failure points inside the loop (slug derivation, SBOM creation, incomplete/too-large SBOM, CVE matching, relevancy filtering, CVE' matching, result upload) and every one of them is a `continue`, never an accumulated error. If **every** container in the profile fails — or the profile has zero containers that got past `getOrCreateSBOM` — the function still returns `nil`.

The caller treats that `nil` as unqualified success:

```go
// controllers/http.go:433-453
err = h.scanService.ScanCP(bgCtx)
if err != nil {
    if errors.Is(err, domain.ErrPartialContainerProfile) {
        h.recordScan(bgCtx, "scanCP", start, "partial", err)
        h.ensureStatuses().markSucceeded(newScan.JobID)
        ...
    } else {
        h.recordScan(bgCtx, "scanCP", start, "error", err)
        h.ensureStatuses().markFailed(newScan.JobID, scanFailureReason("error", err))
        ...
    }
} else {
    h.recordScan(bgCtx, "scanCP", start, "success", nil)
    h.ensureStatuses().markSucceeded(newScan.JobID)
}
```

`domain.ErrPartialContainerProfile` (returned from `GetContainerRelevancyScans` when the *relevancy provider's view of the profile* is incomplete, `adapters/v1/container_profile.go:39`) is a separate mechanism and does not cover any of the failures listed above — those all happen after the scan list has already been retrieved successfully.

## Reproduction

1. Deploy kubevuln with storage + relevancy (ContainerProfile-driven scanning) enabled.
2. Make the image unpullable for a workload's container (e.g. private registry with no/expired `imagePullSecret`, or an image over `maxImageSize`, or a registry returning HTTP 429).
3. `POST /v1/applicationProfileScan` with `jobID=X` for that workload.
4. `getOrCreateSBOM` fails at `core/services/scan.go:273-277` → `continue` → the loop ends (single-container case) → `ScanCP` returns `nil`.
5. `GET /v1/scanStatus/X` returns `{"state":"succeeded","phase":"completed"}` with no `reason` — per `docs/API.md:232-234` this state is documented to mean the scan actually completed. `kubevuln_scans_completed_total{endpoint="scanCP",outcome="success",reason="none"}` increments, and `kubevuln_scan_duration_seconds{...,outcome="success"}` records the (short, failed) duration.

No `VulnerabilityManifest` or `VulnerabilityManifestSummary` was produced, and nothing was submitted to the backend. The same happens for a multi-container pod where every container fails, and silently for a pod where only some containers fail.

This is currently the behavior the test suite pins: `core/services/scan_test.go:349-357` has `create SBOM error` and `create SBOM too many requests` cases with `workload: true` and no `wantErr: true`, so `ScanCP` is asserted to return `nil` in exactly these failure scenarios (see the `wantErr` check at `scan_test.go:483`).

## Root cause

`ScanCP` scans multiple containers per call (unlike `GenerateSBOM`, `ScanCVE`, and `ScanRegistry`, which each scan a single image and correctly return a classified `*domain.ScanError` on failure — see `scan.go:183`/`192`, `:446`/`449`/`461`/`473`/`488`, `:547`/`550`/`557`/`571`/`591`). The per-container loop was written so one bad container doesn't abort the rest, which is correct, but nothing aggregates per-container outcomes back into the function's return value. `ScanCP` is the only one of the four scan flows with zero `domain.ScanError` construction sites.

The per-container failure is only visible via a log line and `platform.ReportScanFailure`, which is a no-op under `keepLocal` (`adapters/mockplatform.go:84`) — so self-hosted deployments get no failure signal at all beyond logs.

## Impact

- The documented async status API (`docs/API.md:220-237`) reports a false `succeeded` for the ContainerProfile/relevancy scan path, currently the primary scan trigger in kubescape-integrated clusters.
- Any alerting on `rate(kubevuln_scans_completed_total{outcome="error"}[5m])` is blind to ScanCP failures, including total failure of a scan job.
- `kubevuln_scan_duration_seconds{outcome="success"}` is polluted with fast observations from scans that actually aborted early, skewing latency dashboards.
- No `VulnerabilityManifest`/summary is produced, but nothing downstream (kubescape operator, UI) is told the scan didn't happen — it will just look like a workload with no vulnerabilities.

## Affected components

- `core/services/scan.go` — `ScanCP`
- `controllers/http.go` — `ScanCP` handler, `recordScan`, `scanFailureReason`
- `controllers/scan_status.go` — status recorded via `markSucceeded`/`markFailed`
- `docs/API.md` — `/v1/scanStatus` contract, scan metrics table

## Acceptance criteria

- [ ] `ScanCP` returns a non-nil, classified error (or a partial-failure sentinel distinguishable from full failure) when one or more containers fail, instead of unconditionally returning `nil`.
- [ ] When **all** containers in the profile fail, `ScanCP` returns an error such that the HTTP layer records `outcome="error"` and `/v1/scanStatus` reports `state="failed"` with a non-empty, bounded `reason`.
- [ ] When **some but not all** containers fail, the outcome is distinguishable from full success (e.g. reuse or extend the existing `"partial"` outcome branch at `controllers/http.go:435-441`, which today only fires for `domain.ErrPartialContainerProfile`).
- [ ] `core/services/scan_test.go`'s `TestScanService_ScanCP` table is updated so `create SBOM error`, `create SBOM too many requests`, and the other per-container failure cases assert `wantErr: true` (or the new partial/error distinction), plus a new multi-container case where only some containers fail.
- [ ] `controllers/http_test.go` gets a case asserting that a ScanCP failure produces `state="failed"` on `/v1/scanStatus` and increments `kubevuln_scans_completed_total` with `outcome="error"`.
- [ ] `docs/API.md` is updated if the outcome/reason semantics for ScanCP change.

## Suggested fix direction

1. In `ScanCP`, replace the bare `continue` statements with recording a per-container failure (most call sites already compute a `scanfailure.Reason*` value for `ReportScanFailure`), and track whether at least one container succeeded and whether at least one failed.
2. After the loop: if zero containers succeeded, return a `*domain.ScanError` (matching the pattern used by `GenerateSBOM`/`ScanCVE`/`ScanRegistry`). If some succeeded and some failed, return a partial-failure sentinel so `controllers/http.go` can route it through the existing `"partial"` branch (currently only triggered by `domain.ErrPartialContainerProfile`) rather than treating it as full success.
3. Update `TestScanService_ScanCP` and add an `http_test.go` case per the acceptance criteria above.

This is scoped to `ScanCP`'s error aggregation and its two call sites; it doesn't require touching the other three scan flows, which already handle this correctly.

bhuvan-somisetty:
## Problem

`ScanCP` (the ContainerProfile/relevancy scan flow behind `POST /v1/applicationProfileScan`) loops over a profile's containers and `continue`s past any per-container failure -- bad image slug, SBOM creation error, an `Incomplete`/`TooLarge` SBOM, CVE matching failure, relevancy filtering error, or result-upload failure -- but the function always `return nil`s once the loop ends, regardless of how many containers failed.

The HTTP layer treats that `nil` as unqualified success: it records `kubevuln_scans_completed_total{endpoint="scanCP",outcome="success"}` and marks the job `succeeded` on `/v1/scanStatus`, even when zero `VulnerabilityManifest`s were produced.

## Root cause

`GenerateSBOM`, `ScanCVE` and `ScanRegistry` each scan a single image and already return a classified `*domain.ScanError` on failure. `ScanCP` scans multiple containers per call, and the per-container loop was written so one bad container doesn't abort the rest -- correct -- but nothing ever aggregated a per-container failure back into the function's return value. It was the only one of the four scan flows with zero `*domain.ScanError` construction sites.

This was pinned by the existing test suite: `TestScanService_ScanCP`'s `create SBOM error`, `create SBOM too many requests` and `timeout SBOM` cases all asserted `ScanCP` returns `nil` in exactly these failure scenarios.

## Fix

`ScanCP` now tracks whether any container failed to produce a scan result and, if so, returns the most recently observed failure classified via the same `scanfailure.Reason*` taxonomy the other three flows already use (reusing the classification already computed at each failure site where one existed, e.g. `classifySBOMError`, `classifySBOMStatusWithAnnotation`, or the existing `*domain.ScanError` returned by `getOrCreateSBOM`). No changes were needed in `controllers/http.go`: `ScanCP`'s existing error branch already records `outcome="error"` and marks the job `failed` with the reason for any non-nil, non-`ErrPartialContainerProfile` error -- `ScanCP` just wasn't giving it one.

A container-level scan failure is treated as a full failure of the job regardless of whether other containers in the same profile succeeded, rather than introducing a new partial-success state: `ErrPartialContainerProfile`'s existing "partial" outcome specifically means "the relevancy data itself isn't ready yet, try again later" and still marks the job succeeded, which would be the wrong signal to reuse for a real per-container scan failure.

## Testing

- `TestScanService_ScanCP`'s `create SBOM error`, `create SBOM too many requests` and `timeout SBOM` cases now assert the classified failure instead of `nil`.
- `TestScanService_ScanCP_SkipsPullForAlreadyRateLimitedImage` (rate-limit backoff regression) updated: skipping the pull is still a failed container scan, so it now asserts the classified error while still asserting zero pull attempts.
- Added `TestScanService_ScanCP_PartialContainerFailure`: a multi-container profile where one container succeeds and another fails at slug derivation -- asserts the overall call still fails, the failure is a classified `*domain.ScanError`, and the successful container's CVE result was still stored.
- Added a `scanCP` case to `TestHTTPController_MetricsEndpoint_RecordsScanFailureReason`, and added `TestHTTPController_ScanStatus_ScanCPFailed`, asserting a `ScanCP` failure reaches `/v1/scanStatus` as `state="failed"` with its reason.
- `go build ./...`, `go vet ./core/services/... ./controllers/...`, `go test ./core/services/...`, `go test ./controllers/...`, and `golangci-lint run --new-from-rev=main` all pass clean.

## Impact

Alerting on `kubevuln_scans_completed_total{outcome="error"}` and `/v1/scanStatus` consumers now see real ScanCP failures instead of a false `success`, for the primary scan path in kubescape-integrated clusters.

Fixes #816

## Summary by CodeRabbit

* **Bug Fixes**
  * Scan failures now consistently report classified error reasons and completion status.
  * Scans continue processing remaining containers when one container fails, while still surfacing the failure.
  * SBOM, vulnerability matching, filtering, uploads, rate limits, and timeouts now provide more reliable failure reporting.
  * Rate-limited image pulls avoid unnecessary retries.

### DIFF
--- a/controllers/http_test.go
+++ b/controllers/http_test.go
@@ -858,6 +858,11 @@ func (s scanErrorService) ScanRegistry(context.Context) error {
 	return s.err
 }
 
+func (s scanErrorService) ScanCP(context.Context) error {
+	s.signalStarted()
+	return s.err
+}
+
 func (s scanErrorService) ValidateGenerateSBOM(ctx context.Context, _ domain.ScanCommand) (context.Context, error) {
 	return ctx, nil
 }
@@ -913,6 +918,15 @@ func TestHTTPController_MetricsEndpoint_RecordsScanFailureReason(t *testing.T) {
 			register:     func(router *gin.Engine, c *HTTPController) { router.POST("/v1/generateSBOM", c.GenerateSBOM) },
 			wantReason:   scanfailure.ReasonUnexpected,
 			unclassified: true,
+		},
+		{
+			// regression test for #816: a real per-container failure inside ScanCP must
+			// surface as outcome="error" with a classified reason, not "success".
+			name:       "scanCP, classified reason",
+			path:       "/v1/scanCP",
+			endpoint:   "scanCP",
+			register:   func(router *gin.Engine, c *HTTPController) { router.POST("/v1/scanCP", c.ScanCP) },
+			wantReason: scanfailure.ReasonCVEMatchingFailed,
 		},
 	}
 	for _, tt := range tests {
@@ -995,6 +1009,22 @@ func (s statusFlowScanService) ValidateGenerateSBOM(ctx context.Context, _ domai
 	return ctx, nil
 }
 
+func (s statusFlowScanService) ScanCP(ctx context.Context) error {
+	if s.phase != "" {
+		domain.UpdateScanPhase(ctx, s.phase)
+	}
+	if s.startedCh != nil {
+		select {
+		case s.startedCh <- struct{}{}:
+		default:
+		}
+	}
+	if s.blockCh != nil {
+		<-s.blockCh
+	}
+	return s.err
+}
+
 func TestHTTPController_ScanStatus_Succeeded(t *testing.T) {
 	c := NewHTTPController(statusFlowScanService{
 		MockScanService: services.NewMockScanService(true),
@@ -1067,6 +1097,49 @@ func TestHTTPController_ScanStatus_Failed(t *testing.T) {
 
 	assert.Equal(t, scanfailure.ReasonCVEMatchingFailed, status.Reason)
 	assert.Equal(t, "completed", status.Phase)
+	require.NotNil(t, status.FinishedAt)
+	assert.False(t, status.FinishedAt.IsZero())
+}
+
+// TestHTTPController_ScanStatus_ScanCPFailed is a regression test for #816: ScanCP's own
+// closure (it does not go through runTrackedScan, see that function's doc comment) used to
+// mark the job succeeded no matter what its per-container loop actually did. A failure
+// returned from the scan service must now reach /v1/scanStatus as state="failed" with its
+// classified reason, exactly like GenerateSBOM/ScanCVE/ScanRegistry already do.
+func TestHTTPController_ScanStatus_ScanCPFailed(t *testing.T) {
+	c := NewHTTPController(statusFlowScanService{
+		MockScanService: services.NewMockScanService(true),
+		err:             &domain.ScanError{Reason: scanfailure.ReasonCVEMatchingFailed, Err: errors.New("boom")},
+	}, 1)
+	t.Cleanup(func() { c.Shutdown(5 * time.Second) })
+
+	router := gin.Default()
+	router.POST("/v1/scanCP", c.ScanCP)
+	router.GET("/v1/scanStatus/:jobID", c.ScanStatus)
+
+	req, _ := http.NewRequest("POST", "/v1/scanCP", strings.NewReader(`{
+		"jobID": "job-scancp-failed",
+		"args": {"name": "daemonset-kube-proxy", "namespace": "kube-system"}
+	}`))
+	w := httptest.NewRecorder()
+	router.ServeHTTP(w, req)
+	require.Equal(t, http.StatusOK, w.Code, w.Body.String())
+
+	var status domain.ScanStatus
+	require.Eventually(t, func() bool {
+		req, _ = http.NewRequest("GET", "/v1/scanStatus/job-scancp-failed", nil)
+		w = httptest.NewRecorder()
+		router.ServeHTTP(w, req)
+		if w.Code != http.StatusOK {
+			return false
+		}
+		if err := json.Unmarshal(w.Body.Bytes(), &status); err != nil {
+			return false
+		}
+		return status.State == domain.ScanStateFailed
+	}, time.Second, 10*time.Millisecond)
+
+	assert.Equal(t, scanfailure.ReasonCVEMatchingFailed, status.Reason)
 	require.NotNil(t, status.FinishedAt)
 	assert.False(t, status.FinishedAt.IsZero())
 }
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -225,6 +225,15 @@ func (s *ScanService) ScanCP(mainCtx context.Context) error {
 		return fmt.Errorf("getting container relevancy scans: %w", err)
 	}
 
+	// failed counts containers that did not produce a scan result; lastFailure is the most
+	// recent one's classified error. If any container fails, ScanCP must not report success
+	// (see #816): a container-level failure is a real scan failure, not the informational
+	// "come back once the profile is complete" signal domain.ErrPartialContainerProfile
+	// carries from the relevancy lookup above, so it is surfaced the same way GenerateSBOM,
+	// ScanCVE and ScanRegistry surface theirs — as a returned *domain.ScanError.
+	var failed int
+	var lastFailure *domain.ScanError
+
 	for _, scan := range scans {
 		imageTagNormalized := tools.NormalizeReference(scan.ImageTag)
 		slug, err := names.ImageInfoToSlug(imageTagNormalized, scan.ImageID)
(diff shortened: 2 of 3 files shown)


matthyx (commented):
Reviewed the ScanCP change against the existing `GenerateSBOM`/`ScanCVE`/`ScanRegistry` failure-classification pattern (`*domain.ScanError`, `classifySBOMError`/`classifySBOMStatusWithAnnotation`, `errors.As` unwrapping) and it's applied consistently at every new `continue` site (slug derivation, SBOM creation, incomplete/too-large SBOM, CVE matching, filtered SBOM/CVE, result upload). The `failed`/`lastFailure` aggregation correctly leaves `ErrPartialContainerProfile` handling in `controllers/http.go` untouched, since `ScanCP` only returns the new `*domain.ScanError` after the relevancy lookup already succeeded — no interaction between the two failure paths. Test coverage (`TestScanService_ScanCP_PartialContainerFailure`, the updated `TestScanService_ScanCP` cases, `TestHTTPController_ScanStatus_ScanCPFailed`) matches the described scenarios and CodeRabbit's automated pass found nothing actionable either. No code-correctness issues found.

**Blocker for merge:** the `DCO` check is failing (`action_required`) — the commit is missing a `Signed-off-by:` trailer. That's what's currently marking this PR `mergeable_state: blocked`; everything else (build, lint, tests, CodeQL, GitGuardian, FOSSA) is green. @bhuvan-somisetty can fix it locally with:

```
git commit --amend -s
git push --force-with-lease
```

I'll approve once the DCO check goes green — no further code changes needed from what I can see.

bhuvan-somisetty:
Thanks for the review! Fixed - amended the commit with `-s` and force-pushed. DCO check is green now.

matthyx (approved):
DCO check is now green (commit carries a `Signed-off-by:` trailer) and the code diff is unchanged from my earlier review — still no correctness issues, consistent with the existing `*domain.ScanError` classification pattern, well covered by tests. All checks pass. Approving.

PR #721, opened 2026-08-15
fix(controllers): fix copylocks violation on HTTPController method receivers (#720)

NaitikRishu:
## Overview
Fixes `copylocks` analyzer violations reported by `go vet ./controllers/...`:

```text
controllers/http.go:99:9: recordScan passes lock by value: github.com/kubescape/kubevuln/controllers.HTTPController contains sync.Once contains sync.noCopy
controllers/http.go:133:9: recordRejection passes lock by value: github.com/kubescape/kubevuln/controllers.HTTPController contains sync.Once contains sync.noCopy
controllers/http.go:206:9: Alive passes lock by value: github.com/kubescape/kubevuln/controllers.HTTPController contains sync.Once contains sync.noCopy
controllers/http.go:211:9: Ready passes lock by value: github.com/kubescape/kubevuln/controllers.HTTPController contains sync.Once contains sync.noCopy
```

### Changes
`HTTPController` contains `statusesOnce sync.Once` (which embeds `sync.noCopy` for mutex synchronization safety). While constructors and route handlers use pointer receivers `func (h *HTTPController)`, four helper/probe methods were defined with value receivers `func (h HTTPController)`. 

Updated `recordScan`, `recordRejection`, `Alive`, and `Ready` to use pointer receivers `(h *HTTPController)` consistently.

## How to Test
Run `go vet ./controllers/...` and ensure it exits with code 0:
```bash
go vet ./controllers/...
go test -v -race -cover ./controllers/...
```

## Related issues/PRs:
* Resolved #720

## Checklist before requesting a review
- [x] My code follows the style guidelines of this project
- [x] I have commented on my code, particularly in hard-to-understand areas
- [x] I have performed a self-review of my code
- [x] If it is a core feature, I have added thorough tests.
- [x] New and existing unit tests pass locally with my changes
```

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->

## Summary by CodeRabbit

* **Refactor**
  * Updated internal request handling and health-check methods without changing their behavior or end-user experience.

<!-- end of auto-generated comment: release notes by coderabbit.ai -->

Issue #720, opened 2026-08-15
fix(controllers): fix copylocks violation on HTTPController method receivers


NaitikRishu:
### Description
Running `go vet ./controllers/...` flags multiple `copylocks` analyzer violations on `HTTPController`:

```text
controllers/http.go:99:9: recordScan passes lock by value: github.com/kubescape/kubevuln/controllers.HTTPController contains sync.Once contains sync.noCopy
controllers/http.go:133:9: recordRejection passes lock by value: github.com/kubescape/kubevuln/controllers.HTTPController contains sync.Once contains sync.noCopy
controllers/http.go:206:9: Alive passes lock by value: github.com/kubescape/kubevuln/controllers.HTTPController contains sync.Once contains sync.noCopy
controllers/http.go:211:9: Ready passes lock by value: github.com/kubescape/kubevuln/controllers.HTTPController contains sync.Once contains sync.noCopy
```

### Cause
`HTTPController` contains `statusesOnce sync.Once` (which embeds `sync.noCopy` for mutex synchronization safety). While the constructors and main route handlers use pointer receivers `func (h *HTTPController)`, four methods are currently declared with value receivers `func (h HTTPController)`:
- `recordScan` (line 99)
- `recordRejection` (line 133)
- `Alive` (line 206)
- `Ready` (line 211)

When these methods are invoked, Go copies the struct and its internal synchronization lock by value, violating Go's concurrency safety guidelines.

### Expected Behavior
All `HTTPController` methods should use pointer receivers `(h *HTTPController)` consistently to prevent copying locks by value and satisfy `go vet`.

### Proposed Fix
Update the receiver type from `(h HTTPController)` to `(h *HTTPController)` for `recordScan`, `recordRejection`, `Alive`, and `Ready`.
```

PR #818, opened 2026-08-20
fix: use pointer receivers on HTTPController methods to satisfy go vet copylocks

### LINKED ISSUE #680: go vet copylocks failures: HTTPController methods copy sync.Once by value
# Description
`go vet ./...` reports four `copylocks` findings in `controllers/http.go`, all on the same root cause: `HTTPController` embeds `sync.Once` (as `statusesOnce sync.Once`), but several of its methods are declared with a value receiver (`func (h HTTPController) ...`) instead of a pointer receiver. Every call to one of those methods copies the struct, including the `sync.Once` (and its internal `sync.noCopy` guard), which is exactly the pattern `go vet`'s `copylocks` check exists to catch.

The affected methods are `recordScan`, `recordRejection`, `Alive`, and `Ready`. All the other `HTTPController` methods (`WithMetrics`, `ensureStatuses`, `claimTrackedJob`, `GenerateSBOM`, `Diagnostics`, `ScanStatus`, `ScanCP`, `ScanCVE`, `ScanRegistry`, `Shutdown`, ...) already use a pointer receiver, so this is an inconsistency rather than an intentional design choice.

Because `make vet` runs `go vet ./...` and fails on these findings, `make verify` currently cannot pass on a clean checkout of `main`, which undercuts it as the documented "default local verification workflow" for contributors.

# Environment
OS: `darwin/arm64`
Version: not applicable, this is a build tooling issue rather than a shipped binary version issue

# Steps To Reproduce
1. Check out `main`.
2. Run `go vet ./...` (or `make vet`).
3. Observe:
   ```
   controllers/http.go:99:9: recordScan passes lock by value: github.com/kubescape/kubevuln/controllers.HTTPController contains sync.Once contains sync.noCopy
   controllers/http.go:133:9: recordRejection passes lock by value: github.com/kubescape/kubevuln/controllers.HTTPController contains sync.Once contains sync.noCopy
   controllers/http.go:206:9: Alive passes lock by value: github.com/kubescape/kubevuln/controllers.HTTPController contains sync.Once contains sync.noCopy
   controllers/http.go:211:9: Ready passes lock by value: github.com/kubescape/kubevuln/controllers.HTTPController contains sync.Once contains sync.noCopy
   ```

# Expected behavior
`go vet ./...` (and therefore `make verify`) passes cleanly on `main`, with every `HTTPController` method that can observe or mutate `statusesOnce` (directly or indirectly through `ensureStatuses`) using a pointer receiver.

# Actual Behavior
`go vet ./...` fails with four `copylocks` findings, all rooted in the same value-receiver-on-a-struct-with-a-sync.Once pattern.

# Additional context
Likely root cause: `recordScan`, `recordRejection`, `Alive`, and `Ready` were written (or left) with value receivers before `statusesOnce sync.Once` was added to `HTTPController`, and nothing enforces receiver consistency across a struct's methods.

Suggested direction: change the four affected method signatures to pointer receivers (`func (h *HTTPController) ...`), matching every other method on the type, and update call sites if any assume a value receiver. This is a minimal, mechanical fix; existing tests in `controllers/http_test.go` should catch any behavioral regression.

rootp1:
## Summary
- `HTTPController` embeds `sync.Once` (`statusesOnce`), but `recordScan`, `recordRejection`, `Alive`, and `Ready` were declared with value receivers, so every call copied the struct (and its `sync.Once`/`sync.noCopy` guard). `go vet`'s `copylocks` check flags this, which makes `go vet ./...` (and `make vet`/`make verify`) fail on a clean checkout of `main`.
- Changed all four methods to pointer receivers, matching every other `HTTPController` method on the type. No behavior change — Go automatically takes the address of addressable values (all call sites use local/heap variables), so existing callers and tests are unaffected.

Fixes #680

## Test plan
- [x] `go vet ./...` — passes with no output (previously reported 4 `copylocks` findings in `controllers/http.go`)
- [x] `go build ./...` — passes
- [x] `go test ./controllers/...` — passes

## Summary by CodeRabbit

* **Refactor**
  * Improved internal request handling consistency without changing user-visible behavior or public functionality.

### DIFF
--- a/controllers/http.go
+++ b/controllers/http.go
@@ -96,7 +96,7 @@ func (h *HTTPController) claimTrackedJob(jobID string) bool {
 // domain.ErrPartialContainerProfile result is a warning-level expected outcome, not a failure.
 // err is the error returned by the scan (nil for success/partial) and is only consulted to
 // resolve the bounded reason label on a failed outcome; see scanFailureReason.
-func (h HTTPController) recordScan(ctx context.Context, endpoint string, start time.Time, outcome string, err error) {
+func (h *HTTPController) recordScan(ctx context.Context, endpoint string, start time.Time, outcome string, err error) {
 	if h.metrics == nil {
 		return
 	}
@@ -130,7 +130,7 @@ func scanFailureReason(outcome string, err error) string {
 // "too_many_requests" for registry back-pressure (ErrTooManyRequests) and "invalid_request"
 // for every other validation error, keeping the rejection-rate signal distinct from
 // malformed-payload noise while staying low cardinality.
-func (h HTTPController) recordRejection(ctx context.Context, endpoint string, err error) {
+func (h *HTTPController) recordRejection(ctx context.Context, endpoint string, err error) {
 	if h.metrics == nil {
 		return
 	}
@@ -203,12 +203,12 @@ func (h *HTTPController) GenerateSBOM(c *gin.Context) {
 }
 
 // Alive returns 200 OK
-func (h HTTPController) Alive(c *gin.Context) {
+func (h *HTTPController) Alive(c *gin.Context) {
 	_, _ = problem.Of(http.StatusOK).WriteTo(c.Writer)
 }
 
 // Ready calls scanService.Ready
-func (h HTTPController) Ready(c *gin.Context) {
+func (h *HTTPController) Ready(c *gin.Context) {
 	if !h.scanService.Ready(c.Request.Context()) {
 		_, _ = problem.Of(http.StatusServiceUnavailable).WriteTo(c.Writer)
 		return


matthyx (changes requested):
Thanks for the fix — the pointer-receiver change itself is correct and consistent with the rest of `HTTPController`.

However, this is already merged: #721 (merged 2026-08-15, `matthyx`) made the exact same change to `recordScan`, `recordRejection`, `Alive`, and `Ready` for the same reason (copylocks on `statusesOnce sync.Once`). `main`'s current `controllers/http.go` already has all four methods on pointer receivers.

This branch was cut from an older `main` commit (`013e8e7`, predating #721), so it's now out of date rather than fixing a live bug — that's also why GitHub reports it as **dirty**/unmergeable against current `main`.

Blockers before this can merge:
1. **Redundant with #721** — please rebase onto latest `main`; the diff will likely disappear entirely since the fix is already there. If issue #680 is a duplicate of #720, it's worth closing #680 as a dup instead.
2. **Merge conflicts** — `mergeable_state` is `dirty`; needs a rebase regardless of point 1.
3. **DCO check failing** (`action_required`) — the commit isn't signed off; needs `git commit --amend -s` (or equivalent) per the repo's DCO bot.

Holding off on approval until these are resolved — expect a rebase will show there's nothing left to merge here.

rootp1:
Thanks for catching this, @matthyx, confirmed: this branch was cut from `013e8e7`, which predates #721, so I was diffing against my fork's stale `main` rather than upstream and missed that #720/#721 already landed the identical pointer-receiver fix on `recordScan`, `recordRejection`, `Alive`, and `Ready`; checked `upstream/main` directly and all four methods already use pointer receivers there, so this PR's diff is fully redundant and I'm closing it rather than rebasing to an empty diff.

rootp1:
Closing as redundant, the fix is already on main via #721.

PR #820, opened 2026-08-20
fix(services): make ScanCP's summary totals agree with every other scan flow

### LINKED ISSUE #819: ScanCP's relevancy branch stores unfiltered severity totals in VulnerabilityManifestSummary, disagreeing with every other scan flow on the same object
## Problem

`VulnerabilityManifestSummary`'s `severities.*.all` counters mean different things depending on which scan flow last wrote the object, even though every flow writes to the *same* K8s object for a given workload/container. Concretely:

- `ScanCVE` and `ScanRegistry` (via `storeFilteredCVE`, `core/services/scan.go:661`) and the cache-hit path (`reconcileCachedCVE`, `core/services/scan.go:734`) all call `StoreCVESummary` with the **exception-filtered** manifest as the first argument.
- `ScanCP`'s own non-relevancy branch does the same (`core/services/scan.go:346`).
- `ScanCP`'s relevancy branch — the one used once a workload's `ContainerProfile` is `Full` — calls `StoreCVESummary` with the **unfiltered** manifest instead (`core/services/scan.go:407`).

`StoreCVESummary`'s `all` counters (`repositories/apiserver.go:730-747`, `parseSeverities`) come straight from `cve.Content.Matches` on whichever manifest it's handed — there's no re-filtering inside the repository layer. So the exact same K8s object (same computed name, see below) reports a Critical CVE's suppression differently depending on which of the two code paths scanned it last.

## Evidence

`core/services/scan.go:404-407`:
```go
// Summary uses original cve (all matches for total counts) and
// filteredCvep (exceptions applied to relevant matches only).
err = s.cveRepository.StoreCVESummary(ctx, cve, filteredCvep, true)
```
This comment was added deliberately in commit `f6848b9` (`ben@armosec.io`) to document this call site's own behavior, so it's not an oversight *there*. But it doesn't address — and predates any check against — the other four `StoreCVESummary` call sites, all of which pass the filtered manifest for the same argument:

- `core/services/scan.go:346` (ScanCP, non-relevancy branch): `StoreCVESummary(ctx, filteredCve, ...)`
- `core/services/scan.go:661` (`storeFilteredCVE`, used by both `ScanCVE` and `ScanRegistry`): `StoreCVESummary(ctx, filteredCve, ...)`
- `core/services/scan.go:734` (`reconcileCachedCVE`, the cache-hit path all four flows can take): `StoreCVESummary(ctx, filteredCve, ...)`

Where `cve` comes from at `core/services/scan.go:407`:
- Fresh-scan branch (`:320-333`): `cve` is the raw output of `cveScanner.ScanSBOM`, never filtered — the filtering happens into a *separate* variable, `filteredCve` (`:335`, comment: "copy — original stays intact for SubmitCVE").
- Cache-hit branch (`:353-356`): `cve, filteredCve, cveExceptionsComplete = s.reconcileCachedCVE(ctx, cve, slug, false)`. `reconcileCachedCVE`'s doc comment (`:700-703`) is explicit: it "[r]eturns the input cve with suppressed matches **restored**" as its first return value.

`repositories/apiserver.go:730-747` (`parseSeverities`) confirms there's no independent re-filtering — it just counts `cve.Content.Matches`:
```go
if cve.Content != nil {
    for i := range cve.Content.Matches {
        switch cve.Content.Matches[i].Vulnerability.Severity {
        case domain.CriticalSeverity:
            critical += 1
        ...
```

And `repositories/apiserver.go:839-861` (`GetCVESummaryK8sResourceNameWithCVEName`) shows the object name is derived purely from `kind-name-containerName`, with no scan-flow discriminator, so `ScanCVE` and `ScanCP` genuinely target the same object for the same workload/container:
```go
```
`StoreCVESummary` (`repositories/apiserver.go:872-922`) then does a full `existing.Spec = manifest.Spec` overwrite on every write (`:919`) — there's no per-field merge that could otherwise mask the disagreement.

## Reproduction

1. Deploy with `storage: true`, `riskAcceptance: true`.
2. Create a `SecurityException` suppressing one Critical CVE present in the image of workload `deployment/foo`, container `bar`.
3. `POST /v1/scanImage` for that workload. The resulting `VulnerabilityManifestSummary foo-bar` (or `deployment-foo-bar`, depending on WLID kind) has `spec.severities.critical.all` excluding the suppressed CVE.
4. Once the workload's `ContainerProfile` reaches `Full`, `POST /v1/applicationProfileScan` for the same workload. The **same** summary object now has `spec.severities.critical.all` one higher — the suppressed CVE is back in the total — while `spec.severities.critical.relevant` still excludes it.
5. A subsequent `ScanCVE` run (e.g. a manual re-scan) flips `critical.all` back down again, with no configuration change in between.

## Root cause

Five call sites write the same kind of object through `StoreCVESummary`'s `cve` argument. Four pass the exception-filtered manifest; one (`ScanCP`'s relevancy branch) passes the unfiltered one. Nothing in `parseSeverities` or `StoreCVESummary` normalizes this — the counters are a direct pass-through of whatever `Content.Matches` the caller handed over. There is no per-endpoint discriminator in the stored object's name, so the two semantics collide on identical objects rather than living in separate, clearly-scoped resources.

## Impact

- `severities.*.all` for a given workload/container object oscillates depending on which scan flow last wrote it, with no underlying data change — this is directly visible via `kubectl get vulnerabilitymanifestsummaries` and whatever the Kubescape UI or a policy/alerting consumer reads from that field.
- A `SecurityException`'s entire purpose is to suppress a CVE from being counted as an active risk; on the `ScanCP` relevancy path it's suppressed from `relevant` but not from `all`, which is inconsistent with how every other scan flow treats the same field on the same kind of object.
- Any trend or diff built on `severities.*.all` for a relevancy-scanned workload is unreliable, since the count can change without a new CVE, a new exception, or a new scan target.

## Affected components

`core/services/scan.go` (`ScanCP`, `reconcileCachedCVE`, `storeFilteredCVE`), `repositories/apiserver.go` (`StoreCVESummary`, `parseSeverities`), the `VulnerabilityManifestSummary` CRD, and the `SecurityException`/`riskAcceptance` feature.

## Acceptance criteria

- [ ] All `StoreCVESummary` call sites for a given kind of scan result agree on whether `severities.*.all` includes or excludes exception-suppressed CVEs — pick one semantic and document it in `docs/CONFIGURATION.md` next to `riskAcceptance`, rather than leaving `ScanCP`'s relevancy branch as the sole exception.
- [ ] A `ScanCVE`/`ScanRegistry` run and a `ScanCP` relevancy run against the same workload/container, with the same `SecurityException`s active, produce identical `severities.*.all` values on the resulting `VulnerabilityManifestSummary`.
- [ ] A regression test in `core/services/scan_test.go` captures the manifest each scan flow hands to a mock `CVERepository.StoreCVESummary` and asserts a suppressed CVE's presence/absence in `Content.Matches` is consistent across `GenerateSBOM`/`ScanCVE`/`ScanRegistry`/`ScanCP` (both branches) for the same exception set.
- [ ] A test asserts that running `ScanCVE` then `ScanCP` (or vice versa) against the same WLID+container leaves the stored summary's `severities.*.all` unchanged between the two runs.

## Suggested fix direction

The minimal, mechanical fix is to make `core/services/scan.go:407` pass `filteredCve` instead of `cve`, matching the other four call sites — `applyExceptionsToManifest` copies by value, so `filteredCve.Name`/`Labels`/`Annotations` are equivalent and no other field of the summary is affected. That resolves the inconsistency by making "all" mean "after exceptions" everywhere.

If the maintainers instead want `all` to mean "every match currently in the image, regardless of exceptions" — a legitimate alternative reading, and arguably what the `f6848b9` comment intended for this one call site — the fix inverts: the other four call sites move to passing the unfiltered manifest, and `parseSeverities` needs to add `len(cve.Content.IgnoredMatches)` into `all` for the branches where suppression already happened before the manifest reached it. Either resolution is a contained change to `core/services/scan.go` plus `repositories/apiserver.go`; the acceptance criteria above (consistency + a documented rule) apply regardless of which one is chosen.

bhuvan-somisetty:
## Problem

`VulnerabilityManifestSummary`'s `severities.*.all` counters mean different things depending on which scan flow last wrote the object, even though every flow writes to the same K8s object for a given workload/container (name is `kind-name-containerName`, no per-flow discriminator).

## Root cause

`StoreCVESummary`'s `all` counters come straight from `Content.Matches` on whichever manifest is passed as its first argument. Four call sites pass the exception-filtered manifest: `ScanCP`'s own non-relevancy branch, `storeFilteredCVE` (used by both `ScanCVE` and `ScanRegistry`), and the cache-hit path in `reconcileCachedCVE`. `ScanCP`'s relevancy branch was the one exception, passing the unfiltered manifest instead -- so a SecurityException-suppressed CVE stayed in `severities.*.all` there while every other flow already excluded it.

Since `ScanCVE`/`ScanRegistry` and `ScanCP` can both write the same `VulnerabilityManifestSummary` object for the same workload/container, the object's `severities.*.all` flipped depending on which flow scanned last, with no underlying data change.

I want to flag directly: a comment at this call site (added in commit `f6848b9`) documents this specific line's behavior. It doesn't address the inconsistency against the other four call sites, though, which is what this fix targets -- I'm not disputing that comment's intent for its own line, just making all five call sites agree with each other, since right now the same object's meaning depends on which endpoint wrote it last.

## Fix

`core/services/scan.go`: `ScanCP`'s relevancy branch now passes `filteredCve` instead of `cve` to `StoreCVESummary`, matching the other four call sites. `SubmitCVE` (cloud/backend reporting) is untouched -- it still receives the unfiltered manifest everywhere, exactly like `ScanCVE`/`ScanRegistry` already do for their own `SubmitCVE` calls, since that's a different, deliberately unfiltered surface.

## Testing

Added `TestScanService_ScanCP_SummaryExcludesSuppressedCVEFromTotals`: wraps the CVE repository to capture every `StoreCVESummary` call `ScanCP` makes (both the non-relevancy and relevancy branches) and asserts a SecurityException-suppressed CVE lands in `IgnoredMatches`, never `Matches`, on every call.

I verified the test actually catches the regression: reverting the fix locally and re-running it fails with `suppressed CVE must not be counted toward severities.*.all` (the CVE showed up in `Matches` with zero `IgnoredMatches`), confirming the test isn't a false positive.

- `go build ./...` -- clean
- `go test ./core/services/...` -- full package passes (all existing tests plus the new one)
- `golangci-lint run --new-from-rev=main` -- 0 issues

## Impact

`ScanCVE`/`ScanRegistry` scans and `ScanCP` relevancy scans against the same workload/container, with the same `SecurityException`s active, now produce identical `severities.*.all` values, instead of the count oscillating between scans with no configuration change.

Fixes #819

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected vulnerability summary counts when security exceptions suppress findings.
  * Ensured suppressed vulnerabilities are consistently excluded from active matches across scan workflows.
  * Improved consistency between relevancy and non-relevancy scan results.

* **Tests**
  * Added regression coverage for summaries involving security-exception-suppressed vulnerabilities.

### DIFF
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -374,9 +374,17 @@ func (s *ScanService) ScanCP(mainCtx context.Context) error {
 						helpers.String("instanceID", scan.InstanceID.GetStringFormatted()))
 					// no continue, storing the CVE' is not critical
 				}
-				// Summary uses original cve (all matches for total counts) and
-				// filteredCvep (exceptions applied to relevant matches only).
-				err = s.cveRepository.StoreCVESummary(ctx, cve, filteredCvep, true)
+				// Summary uses filteredCve (all matches, exceptions applied) and
+				// filteredCvep (exceptions applied to relevant matches only), matching
+				// every other StoreCVESummary call site (storeFilteredCVE,
+				// reconcileCachedCVE, and ScanCP's own non-relevancy branch above) so
+				// severities.*.all means the same thing regardless of which scan flow
+				// last wrote this workload/container's VulnerabilityManifestSummary.
+				// See #819: this used to pass the unfiltered cve, so a
+				// SecurityException-suppressed CVE's presence in severities.*.all
+				// flipped depending on whether ScanCVE/ScanRegistry or ScanCP scanned
+				// last.
+				err = s.cveRepository.StoreCVESummary(ctx, filteredCve, filteredCvep, true)
 				if err != nil {
 					logger.L().Ctx(ctx).Warning("storing CVE summary", helpers.Error(err),
 						helpers.String("imageSlug", slug))
--- a/core/services/scan_test.go
+++ b/core/services/scan_test.go
@@ -1268,6 +1268,124 @@ func TestScanService_ScanRegistry_StorageAndExceptions(t *testing.T) {
 	})
 }
 
+type storeCVESummaryCall struct {
+	cve           domain.CVEManifest
+	cvep          domain.CVEManifest
+	withRelevancy bool
+}
+
+// summaryCapturingRepo wraps a ports.CVERepository and records every StoreCVESummary call
+// verbatim, so a test can inspect exactly which manifest a scan flow handed to the summary
+// write. MemoryStore alone can't answer that: StoreCVE and StoreCVESummary key their map off
+// the same cveID, so a later call silently overwrites an earlier one's entry.
+type summaryCapturingRepo struct {
+	ports.CVERepository
+	calls []storeCVESummaryCall
+}
+
+func (r *summaryCapturingRepo) StoreCVESummary(ctx context.Context, cve domain.CVEManifest, cvep domain.CVEManifest, withRelevancy bool) error {
+	r.calls = append(r.calls, storeCVESummaryCall{cve: cve, cvep: cvep, withRelevancy: withRelevancy})
+	return r.CVERepository.StoreCVESummary(ctx, cve, cvep, withRelevancy)
+}
+
+// TestScanService_ScanCP_SummaryExcludesSuppressedCVEFromTotals is a regression test for #819.
+// ScanCP's relevancy branch used to pass the *unfiltered* CVE manifest as StoreCVESummary's
+// first argument, so a SecurityException-suppressed CVE stayed in severities.*.all even though
+// every other StoreCVESummary call site (ScanCP's own non-relevancy branch above it,
+// storeFilteredCVE used by ScanCVE/ScanRegistry, and the cache-hit path in reconcileCachedCVE)
+// already excludes it there. Asserts every StoreCVESummary call ScanCP makes -- both the
+// non-relevancy and relevancy branches -- agrees that the suppressed CVE belongs in
+// IgnoredMatches, never in Matches.
+func TestScanService_ScanCP_SummaryExcludesSuppressedCVEFromTotals(t *testing.T) {
+	const suppressedCVE = "CVE-2023-9999"
+	wlid := "wlid://cluster-minikube/namespace-kube-system/daemonset-kube-proxy"
+
+	mockRepo := &mockSecurityExceptionRepo{
+		exceptions: []sev1beta1.SecurityException{
+			{
+				ObjectMeta: metav1.ObjectMeta{Namespace: "kube-system"},
+				Spec: sev1beta1.SecurityExceptionSpec{
+					Reason: "test exception",
+					Vulnerabilities: []sev1beta1.VulnerabilityException{
+						{
+							Vulnerability: sev1beta1.VulnerabilityRef{ID: suppressedCVE},
+							Status:        sev1beta1.VulnerabilityStatusNotAffected,
+						},
+					},
+				},
+			},
+		},
+	}
+	mockPlatform := adapters.NewMockPlatform(false, mockRepo)
+	storageCP := repositories.NewMemoryStorage(false, false)
+	cveRepo := &summaryCapturingRepo{CVERepository: repositories.NewMemoryStorage(false, false)}
+
+	s := NewScanService(
+		adapters.NewMockSBOMAdapter(false, false, false),
+		repositories.NewMemoryStorage(false, false),
+		&fakeCVEScannerWithVuln{},
+		cveRepo,
+		mockPlatform,
+		v1.NewContainerProfileAdapter(storageCP),
+		true,  // storage
+		false, // vexGeneration
+		true,  // sbomGeneration
+		false, // storeFilteredSbom
+		false, // partialRelevancy
+	)
+	ctx := context.TODO()
+	s.Ready(ctx)
+
+	workload := domain.ScanCommand{
+		Args: map[string]interface{}{
+			domain.ArgsName:      "daemonset-kube-proxy",
+			domain.ArgsNamespace: "kube-system",
+		},
+		Wlid: wlid,
+	}
+	ctx, err := s.ValidateScanCP(ctx, workload)
+	require.NoError(t, err)
+
+	ap := v1beta1.ContainerProfile{
+		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Verified the fix directly against `core/services/scan.go`:

- `filteredCve` in `ScanCP`'s relevancy branch is populated the same way in both branches — `applyExceptionsToManifest` on the cache-miss path (line 311) and via `reconcileCachedCVE` on the cache-hit path (line 332) — so swapping `cve` → `filteredCve` at the `StoreCVESummary` call (line 387) genuinely makes this call site consistent with the other four (`ScanCP`'s own non-relevancy branch, `storeFilteredCVE` used by `ScanCVE`/`ScanRegistry`, and `reconcileCachedCVE`'s cache-hit store). All five now pass an exceptions-filtered manifest as the summary's first argument.
- `SubmitCVE` still receives the unfiltered `cve`/`cvep` (line 400), untouched — matches the PR description's claim that cloud reporting semantics are deliberately unchanged.
- The new regression test (`TestScanService_ScanCP_SummaryExcludesSuppressedCVEFromTotals`) reuses existing test helpers/imports already present in `scan_test.go` (`mockSecurityExceptionRepo`, `fakeCVEScannerWithVuln`, `adapters.NewMockPlatform`, `v1.NewContainerProfileAdapter`, `repositories.NewMemoryStorage`) with a `NewScanService(...)` call matching the argument order used by neighboring `ScanCP` tests, so it should compile cleanly. It asserts the suppressed CVE lands in `IgnoredMatches` and never in `Matches` across every `StoreCVESummary` call the flow makes, including the last (relevancy) call — a `summaryCapturingRepo` wrapper is a reasonable way to capture that, since `MemoryStore`'s summary storage would otherwise just overwrite itself.

Change is minimal and scoped to exactly what the description claims. No blockers — approving.

PR #821, opened 2026-08-20
fix(ci): isolate release workflow concurrency

rootp1:
## Summary
- stop `pr-merged` runs from sharing one concurrency group on merged PRs
- use the pull request number in the concurrency key instead of `github.ref`
- keep the release workflow from cancelling another merge that is already in flight

## Validation
- ran `actionlint .github/workflows/*`
- remaining `pr-image.yaml` shellcheck warnings are pre-existing and unrelated to this change

### DIFF
--- a/.github/workflows/pr-merged.yaml
+++ b/.github/workflows/pr-merged.yaml
@@ -18,7 +18,7 @@ on:
       - '.github/workflows/*'
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-pr-${{ github.event.pull_request.number || github.run_id }}
   cancel-in-progress: true
 
 jobs:


matthyx (approved):
Verified the root cause: `pr-merged.yaml` runs on `pull_request_target: types: [closed]`, and for `pull_request_target` (unlike `pull_request`) `github.ref` resolves to the base branch (`refs/heads/main`), not a PR-specific merge ref. So every merged PR previously shared the exact same concurrency group (`pr-merged-refs/heads/main`) with `cancel-in-progress: true`, meaning merging PR B while PR A's release/publish job was still running would silently cancel PR A's in-flight release.

Keying the group off `github.event.pull_request.number` (with a `run_id` fallback) fixes that while still deduping duplicate `closed` events for the same PR. Confirmed this doesn't regress `pr-image.yaml`/`pr-created.yaml`'s use of `github.ref` — those trigger on plain `pull_request` events where `github.ref` is already the per-PR merge ref, so they don't have this bug and don't need the same change.

One-line change, CI green, no functional risk. Approving.

PR #822, opened 2026-08-20
chore(ci): pin GitHub workflow refs

rootp1:
## Summary
- pin mutable GitHub Action and reusable workflow refs to full commit SHAs
- add Dependabot updates for GitHub Actions so pinned refs still receive update PRs
- keep the change scoped to workflow supply-chain reproducibility only

## Validation
- ran `actionlint .github/workflows/*`
- remaining `pr-image.yaml` shellcheck warnings are pre-existing and unchanged by this PR

## Summary by CodeRabbit

* **Chores**
  * Added weekly automated updates for GitHub Actions.
  * Pinned CI/CD and security workflow actions to immutable versions for more predictable and secure builds.
  * Pinned reusable workflow references to specific revisions.

### DIFF
--- a/.github/workflows/pr-created.yaml
+++ b/.github/workflows/pr-created.yaml
@@ -20,7 +20,7 @@ jobs:
     permissions:
       pull-requests: write
       security-events: write
-    uses: kubescape/workflows/.github/workflows/incluster-comp-pr-created.yaml@main
+    uses: kubescape/workflows/.github/workflows/incluster-comp-pr-created.yaml@52bcdc0a1e5c86dacf1450fc65429b1246e2b354
     with:
       CGO_ENABLED: 0
       GO_VERSION: "1.26"
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -32,12 +32,12 @@ jobs:
 
     steps:
       - name: "Checkout code"
-        uses: actions/checkout@v4
+        uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
         with:
           persist-credentials: false
 
       - name: "Run analysis"
-        uses: ossf/scorecard-action@v2.4.0
+        uses: ossf/scorecard-action@62b2cac7ed8198b15735ed49ab1e5cf35480ba46 # v2.4.0
         with:
           results_file: results.sarif
           results_format: sarif
@@ -59,7 +59,7 @@ jobs:
       # Upload the results as artifacts (optional). Commenting out will disable uploads of run results in SARIF
       # format to the repository Actions tab.
       - name: "Upload artifact"
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
         with:
           name: SARIF file
           path: results.sarif
@@ -67,6 +67,6 @@ jobs:
 
       # Upload the results to GitHub's code scanning dashboard.
       - name: "Upload to code-scanning"
-        uses: github/codeql-action/upload-sarif@v3
+        uses: github/codeql-action/upload-sarif@f3712979fa5f215279b101dd0a2e3bdfb4353324 # v3.37.7
         with:
           sarif_file: results.sarif
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -21,12 +21,12 @@ jobs:
       contents: read
     steps:
       - name: Checkout code
-        uses: actions/checkout@v4
+        uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
         with:
           fetch-depth: 0
 
       - name: Set up Go
-        uses: actions/setup-go@v5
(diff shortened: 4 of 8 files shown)


.github/workflows/scorecard.yml
```diff
@@ -32,12 +32,12 @@ jobs:
...
     steps:
       - name: "Checkout code"
-        uses: actions/checkout@v4
+        uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
         with:
           persist-credentials: false
 
       - name: "Run analysis"
-        uses: ossf/scorecard-action@v2.4.0
+        uses: ossf/scorecard-action@ff5dd8929f96a8a4dc67d13f32b8c75057829621 # v2.4.0
```
matthyx: **Blocker: this SHA is the annotated tag object, not the commit.**

`ff5dd8929f96a8a4dc67d13f32b8c75057829621` is the SHA of the `v2.4.0` *tag object* itself (confirmed via the GitHub tags API — `GET /repos/ossf/scorecard-action/git/refs/tags/v2.4.0` returns a `tag` object whose `object.sha` — the actual commit — is `62b2cac7ed8198b15735ed49ab1e5cf35480ba46`).

The standard/documented SHA-pinning convention (and what every other pin in this PR does) is to pin to the commit, not the tag object. Cross-checking GitHub code search: the commit SHA `62b2cac7...` is used by ~1560 files across GitHub, vs. only 28 for the tag-object SHA used here — strong signal this is the wrong value.

Please pin to:
```
uses: ossf/scorecard-action@62b2cac7ed8198b15735ed49ab1e5cf35480ba46 # v2.4.0
```

.github/workflows/scorecard.yml
```diff
@@ -59,14 +59,14 @@ jobs:
...
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
         with:
           name: SARIF file
           path: results.sarif
           retention-days: 5
 
       # Upload the results to GitHub's code scanning dashboard.
       - name: "Upload to code-scanning"
-        uses: github/codeql-action/upload-sarif@v3
+        uses: github/codeql-action/upload-sarif@67ba681cdcbc95ed0371c210ccf82c4d968ae5eb # v3.37.7
```
matthyx: **Same issue: this is the tag object SHA, not the commit.**

`67ba681cdcbc95ed0371c210ccf82c4d968ae5eb` is the SHA of the `v3.37.7` tag object for `github/codeql-action`. The commit it points to (via `object.sha`) is `f3712979fa5f215279b101dd0a2e3bdfb4353324`. GitHub code search shows 127 repos pinning to the commit SHA vs. only 2 to this tag-object SHA.

Please pin to:
```
uses: github/codeql-action/upload-sarif@f3712979fa5f215279b101dd0a2e3bdfb4353324 # v3.37.7
```

matthyx (changes requested):
Good direction overall — the checkout/setup-go/login-action/setup-qemu-action/setup-buildx-action/upload-artifact pins are all correctly resolved commit SHAs, and I verified they match their claimed version tags exactly.

One blocking issue: in `.github/workflows/scorecard.yml`, the `ossf/scorecard-action` and `github/codeql-action/upload-sarif` pins use the **annotated tag object's own SHA** rather than the commit it points to (annotated tags are a distinct git object from the commit they reference). Left inline comments with the correct commit SHAs for both — this undermines the goal of the PR (pinning to immutable, standard commit refs) even though it's likely harmless functionally.

Non-blocking, out of scope for this PR: CodeRabbit's review also flagged that the reusable workflows this repo calls into (`kubescape/workflows`) still contain unpinned nested actions and broad `secrets: inherit` usage. That's a `kubescape/workflows` concern, not something fixable here — worth a follow-up issue in that repo, not a blocker for this scoped change.

Will approve once the two SHAs above are fixed.

.github/workflows/scorecard.yml
rootp1: Pinned to the actual commit 62b2cac7ed8198b15735ed49ab1e5cf35480ba46 instead of the tag object.

.github/workflows/scorecard.yml
rootp1: Pinned to the actual commit f3712979fa5f215279b101dd0a2e3bdfb4353324 instead of the tag object.

.github/workflows/pr-created.yaml
```diff
@@ -20,7 +20,7 @@ jobs:
     permissions:
       pull-requests: write
       security-events: write
-    uses: kubescape/workflows/.github/workflows/incluster-comp-pr-created.yaml@main
+    uses: kubescape/workflows/.github/workflows/incluster-comp-pr-created.yaml@52bcdc0a1e5c86dacf1450fc65429b1246e2b354
```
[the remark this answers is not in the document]
rootp1: Left this as is. Pinning the nested actions inside kubescape/workflows and narrowing secrets there is out of scope for this repo, it needs to happen in that repo instead.

rootp1:
@matthyx fixed both scorecard.yml pins to the actual commit SHAs. Please have another look when you get a chance.

matthyx (approved):
Both blockers are fixed — `.github/workflows/scorecard.yml` now pins `ossf/scorecard-action` to `62b2cac7ed8198b15735ed49ab1e5cf35480ba46` and `github/codeql-action/upload-sarif` to `f3712979fa5f215279b101dd0a2e3bdfb4353324`, which I've re-verified against the GitHub tags API are the actual commit objects for v2.4.0 and v3.37.7 (not the annotated tag objects). The rest of the newly added pins (checkout, setup-go, docker/login-action, setup-qemu-action, setup-buildx-action, upload-artifact, and the two kubescape/workflows@<sha> reusable-workflow refs) all check out against their respective tags. Approving.

PR #823, opened 2026-08-20
chore(ci): narrow PR image build triggers

rootp1:
## Summary
- add `paths` filters to `PR Image Build`
- run the image build only when Go sources, module files, Docker/build inputs, or the workflow itself change
- avoid spending runner time on docs-only and unrelated PR updates

## Validation
- ran `actionlint .github/workflows/pr-image.yaml`
- remaining shellcheck warnings in the build script are pre-existing and unchanged by this PR

### DIFF
--- a/.github/workflows/pr-image.yaml
+++ b/.github/workflows/pr-image.yaml
@@ -2,6 +2,14 @@ name: PR Image Build
 on:
   pull_request:
     types: [synchronize, ready_for_review, opened, reopened]
+    paths:
+      - '**/*.go'
+      - 'go.mod'
+      - 'go.sum'
+      - 'Makefile'
+      - '.dockerignore'
+      - 'build/**'
+      - '.github/workflows/pr-image.yaml'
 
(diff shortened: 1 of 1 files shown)


matthyx (approved):
Reviewed. This is a clean, low-risk CI-only change:

- `build/**` correctly covers `build/Dockerfile` (per the `DOCKERFILE_PATH` in the Makefile), and the filter list matches the same `paths` pattern already used in `lint.yaml` for the same purpose — consistent with existing repo conventions.
- Including `.github/workflows/pr-image.yaml` itself in the paths list is a nice touch — future edits to the workflow will still trigger a build to validate them.
- Checked branch protection (`kubescape` org ruleset on `main`): it only requires 1 approving review, no required status checks, so a path-filtered/skipped `build-and-push-image` run can never block merges of unrelated PRs (the usual gotcha with this pattern).
- No other workflow depends on this one's outputs (`pr-created.yaml`/`pr-merged.yaml` are independent reusable-workflow calls).
- `actionlint` passed per the PR description, and CI is green.

No blockers. Approving.

PR #827, opened 2026-08-20
test: add coverage for remaining BrokenStore methods

AHS0003:
## Overview

`repositories/broken.go` defines `BrokenStore`, a stub repository used across the test suite to simulate a storage backend that always fails. It implements three interfaces `ContainerProfileRepository`, `CVERepository`, and `SBOMRepository`  for a total of 10 methods.

Only 4 of those methods (`GetCVE`, `GetSBOM`, `StoreCVE`, `StoreSBOM`) had existing unit tests in `repositories/broken_test.go`. This PR adds tests for the remaining 6 methods, bringing `BrokenStore` to full coverage:

- `GetContainerProfile`
- `GetCVESummary`
- `StoreCVESummary`
- `StoreCVESummaryStub`
- `StoreVEX`
- `DeleteSBOM`

## Additional Information

Each new test follows the same pattern as the existing ones in the file: construct a `BrokenStore` via `NewBrokenStorage()`, call the method with zero-value/empty arguments, and assert on the returned error.

One method behaves differently from the rest and is worth calling out for reviewers: `GetCVESummary` does **not** return `domain.ErrExpectedError` like every other method on `BrokenStore`  it returns a nil error along with an empty `VulnerabilityManifestSummary`. Rather than "fixing" this to be consistent, the new test (`TestBrokenStore_GetCVESummary`) asserts the actual current behavior (`assert.NoError`), so that:
- the test suite documents this asymmetry explicitly instead of silently relying on it.
- any future change to this behavior (intentional or accidental) will be caught by CI rather than discovered downstream.

No production code was touched  this is a test-only change, isolated to a single file, with no risk of behavioral impact.

## How to Test

go test ./repositories/... -run TestBrokenStore -v

## Summary by CodeRabbit

* **Tests**
  * Added coverage for storage failure handling across container profiles, CVE summaries, VEX data, and SBOM deletion.
  * Verified expected errors are returned for failed operations.
  * Confirmed CVE summary retrieval returns a valid result without an error.

### DIFF
--- a/repositories/broken_test.go
+++ b/repositories/broken_test.go
@@ -31,3 +31,40 @@ func TestBrokenStore_StoreSBOM(t *testing.T) {
 	err := b.StoreSBOM(context.TODO(), domain.SBOM{}, false)
 	assert.Error(t, err)
 }
+
+func TestBrokenStore_GetContainerProfile(t *testing.T) {
+	b := NewBrokenStorage()
+	_, err := b.GetContainerProfile(context.TODO(), "", "")
+	assert.Error(t, err)
+}
+
+func TestBrokenStore_GetCVESummary(t *testing.T) {
+	b := NewBrokenStorage()
+	summary, err := b.GetCVESummary(context.TODO())
+	assert.NoError(t, err)
+	assert.NotNil(t, summary)
+}
+
+func TestBrokenStore_StoreCVESummary(t *testing.T) {
+	b := NewBrokenStorage()
+	err := b.StoreCVESummary(context.TODO(), domain.CVEManifest{}, domain.CVEManifest{}, false)
+	assert.Error(t, err)
+}
+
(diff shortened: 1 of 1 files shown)


matthyx (approved):
Verified each new test against the actual `BrokenStore` implementation in `repositories/broken.go` — all six new tests assert exactly what the methods return (error for `GetContainerProfile`, `StoreCVESummary`, `StoreCVESummaryStub`, `StoreVEX`, `DeleteSBOM`; no error + non-nil summary for `GetCVESummary`). Test-only change, single file, all CI checks green, no behavioral risk.

Optional nit (non-blocking): CodeRabbit's suggestion to assert `summary` equals the zero-value `v1beta1.VulnerabilityManifestSummary{}` instead of just `NotNil` would make `TestBrokenStore_GetCVESummary` a bit more precise, since `NotNil` on a pointer literal is trivially true. Not required for merge.

Approving.

PR #829, opened 2026-08-20
refactor(platform): share the platform helpers both SBOM paths duplicate

### LINKED ISSUE #828: platform helpers are duplicated across both SBOM paths, and #546 left an unreachable arm behind
`adapters/v1/syft.go` and `pkg/sbomscanner/v1/server.go` both handle platform resolution, and both carry their own copy of the same two helpers.

`formatResolvedPlatform` is byte identical in the two files, down to the four-case test table in `adapters/v1/syft_test.go` and `pkg/sbomscanner/v1/server_test.go`. Its doc comment says it is duplicated "since the two packages don't share a dependency either could live in without introducing one", but that stopped being true when `internal/syftsource` landed in #793. Both files import it today.

Recognising a stereoscope platform mismatch is spelled twice as well: `server.go` has a local `isPlatformMismatch`, `syft.go` does the same `errors.As` inline.

Two smaller things came in with #546, which added the platform fallback counter to both.

`server.go`'s last switch arm re-checks the mismatch and records the counter again:

```go
case err != nil && isPlatformMismatch(err):
    metrics.RecordScanFallback(ctx, ..., FallbackStrategyPlatformMismatch, ...)
    // returns StatusReason: domain.ReasonPlatformNotFound
...
case err != nil:
    if isPlatformMismatch(err) {
        metrics.RecordScanFallback(ctx, ..., FallbackStrategyPlatformMismatch, ...)
    }
```

That inner check can never be true. The switch has no tag, so arms are evaluated top down and the first match wins, and `isPlatformMismatch` is a plain `errors.As` on a fixed target type, so anything that would satisfy it has already been claimed above.

Mutation confirms it rather than just arguing it. Disabling the earlier arm makes `TestCreateSBOM_PlatformMismatch_LocalRegistry` fail with `StatusReason` empty instead of `platform-not-found`, so a mismatch only ever lands in the earlier arm:

```
--- Expected
+++ Actual
-platform-not-found
+
FAIL	github.com/kubescape/kubevuln/pkg/sbomscanner/v1
```

And in `syft.go` the same five-line comment appears twice in a row, above and below the metric block #546 inserted between them.

Nothing is user visible here, the counter fires once per mismatch on each path either way. It is the duplication that matters: #512 was the two SBOM paths disagreeing about platforms, and two copies of the predicate that decides what a platform failure is are the same setup again.

Moving both helpers into `internal/syftsource`, which is already where `ParsePlatform` lives for this reason, and dropping the dead arm and the repeated comment.

AnshulPatil2005:
## Overview

`adapters/v1/syft.go` and `pkg/sbomscanner/v1/server.go` each carried their own `formatResolvedPlatform`, byte identical down to the test table, and their own way of recognising a stereoscope platform mismatch: `syft.go` inline via `errors.As`, `server.go` behind a local `isPlatformMismatch`.

`internal/syftsource` already exists as the shared home for exactly this, and both files already import it, so the comment on `formatResolvedPlatform` saying the two packages have no shared dependency to put it in was stale.

Both helpers move there as `FormatResolvedPlatform` and `IsPlatformMismatch`, with one test table instead of two.

## Additional Information

Two artefacts of the copies came along with them, both from #546.

`server.go`'s final `case err != nil` arm re-checked the mismatch and recorded the platform fallback counter a second time. That arm is unreachable: the switch has no tag, so arms are tried in order, `case err != nil && isPlatformMismatch(err)` sits above it, and `isPlatformMismatch` is a pure `errors.As` on a fixed target type, so anything satisfying it has already been claimed.

`syft.go` had the same five-line comment twice in a row, above and below the metric block #546 inserted between them.

No behaviour change. Both helpers keep their bodies exactly as they were, and the counter still fires once per mismatch on each path.

The test table gained two cases while being consolidated, `os known, arch unknown` and `variant without arch`. Both copies only covered the reverse direction.

## How to Test

`go build ./... && go vet ./... && go test ./internal/syftsource/ ./adapters/v1/ ./pkg/sbomscanner/v1/ -count=1`

The unreachability claim is checked by mutation rather than argued. Forcing the earlier arm off:

```
case err != nil && false && syftsource.IsPlatformMismatch(err):
```

makes `TestCreateSBOM_PlatformMismatch_LocalRegistry` fail with `StatusReason` empty instead of `platform-not-found`:

```
--- Expected
+++ Actual
-platform-not-found
+
FAIL	github.com/kubescape/kubevuln/pkg/sbomscanner/v1
```

So a platform mismatch only ever reaches the earlier arm, and the copy of the check in the later one could not run.

`TestIsPlatformMismatch` moved from `pkg/sbomscanner/v1/server_test.go` to `internal/syftsource` unchanged, including the case that matters most: an error whose text mentions "mismatched platform" but is not the typed error must not classify as one.

## Related issues/PRs:

* Resolved #828

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of platform mismatches, including errors wrapped by other operations.
  * Standardized resolved platform reporting using OCI-compatible formatting.
  * Avoided duplicate platform-mismatch metrics for source-resolution errors.

* **Refactor**
  * Centralized platform handling to provide more consistent behavior across SBOM scanning workflows.

* **Tests**
  * Added coverage for valid, incomplete, wrapped, unrelated, and empty platform-related inputs.

### DIFF
--- a/adapters/v1/syft.go
+++ b/adapters/v1/syft.go
@@ -34,20 +34,6 @@ import (
 	"go.opentelemetry.io/otel"
 )
 
-// formatResolvedPlatform builds an OCI-style "os/arch[/variant]" string from the platform
-// fields Syft/stereoscope actually resolved an image against. Returns "" unless both os and
-// architecture are known.
-func formatResolvedPlatform(os, arch, variant string) string {
-	if os == "" || arch == "" {
-		return ""
-	}
-	parts := []string{os, arch}
-	if variant != "" {
-		parts = append(parts, variant)
-	}
-	return strings.Join(parts, "/")
-}
-
 // createSBOMFn is an indirection over syft.CreateSBOM so the deadline handling in
 // CreateSBOM can be unit-tested without cataloguing a real image.
 var createSBOMFn = syft.CreateSBOM
@@ -281,15 +267,9 @@ func (s *SyftAdapter) CreateSBOM(ctx context.Context, name, imageID, imageTag st
 		// match options.Platform). Propagated as-is: classifySBOMError in core/services
 		// recognizes it via errors.As and reports a distinct "platform not found" reason
 		// instead of the generic SBOM-generation-failed fallback (see #512).
-		var platformErr *image.ErrPlatformMismatch
-		if errors.As(err, &platformErr) {
+		if syftsource.IsPlatformMismatch(err) {
 			metrics.RecordScanFallback(ctx, metrics.ComponentInProcess, metrics.FallbackCategoryPlatform, metrics.FallbackStrategyPlatformMismatch, metrics.FallbackOutcomeFailed)
 		}
-		// Requested-but-unavailable platforms surface here as *image.ErrPlatformMismatch
-		// (from stereoscope, once it has positively resolved the image but its OS/arch don't
-		// match options.Platform). Propagated as-is: classifySBOMError in core/services
-		// recognizes it via errors.As and reports a distinct "platform not found" reason
-		// instead of the generic SBOM-generation-failed fallback (see #512).
 		endActiveTempDirUse()
 		unlockPullMutex()
 		return domainSBOM, err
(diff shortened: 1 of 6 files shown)


matthyx (approved):
Reviewed and independently verified. Checked out the branch and ran `go build ./...`, `go vet ./...`, and the listed test packages — all clean. Verified the unreachable-arm claim in `pkg/sbomscanner/v1/server.go` by mutation (forcing off the earlier `case err != nil && syftsource.IsPlatformMismatch(err)` arm reproduces the described failure in `TestCreateSBOM_PlatformMismatch_LocalRegistry`), confirming the removed duplicate check was indeed dead code. The consolidation into `internal/syftsource` is straightforward, imports are all still used, and no behavior changes beyond the described dead-code removal. LGTM.

PR #831, opened 2026-08-20
test(safefetch): cover the SSRF guards the package documents but never exercised

### LINKED ISSUE #830: safefetch's SSRF guards are the untested half of the package: the redirect policy has no test at all
`internal/safefetch` is at 72.9%, and `New` alone is at 53.8%:

```
safefetch.go:66:	embeddedIPv4	88.9%
safefetch.go:83:	mustParseCIDR	75.0%
safefetch.go:129:	New		53.8%
safefetch.go:185:	checkIPAllowed	100.0%
safefetch.go:210:	Fetch		80.0%
total:				72.9%
```

The uncovered part of `New` is not filler. `New` is almost entirely two closures, `safeDialContext` and `CheckRedirect`, and those two are where the package's guarantees actually live. `checkIPAllowed` is the one function at 100%, but a pure predicate being correct is a different property from the dialer calling it.

**The dial-time check.** One test reaches it, `TestFetch_BlocksRealLoopbackServer`, and only for loopback. `checkIPAllowed` has a table covering the cloud metadata endpoint, RFC1918, the carrier-grade NAT range, `0.0.0.0/8` and the NAT64 forms, and nothing confirms a dial to any of them is refused. The package doc claims:

> IP validation happens at actual connection time via a custom DialContext, not just when the URL is first parsed, so a DNS record that changes between the check and the connection (DNS rebinding) cannot bypass it.

Nothing tests that. `Fetch` inspects only the scheme, so the whole claim rests on the dialer, and the dialer is 53.8% covered.

**The redirect policy.** No test at all. `Fetch` validates the scheme once, on the URL the caller passed. Every hop after that is `CheckRedirect`'s job:

```go
CheckRedirect: func(req *http.Request, via []*http.Request) error {
    if len(via) >= defaultMaxRedirects {
        return ErrTooManyRedirects
    }
    if req.URL.Scheme != "https" {
        return fmt.Errorf("redirect target: %w", ErrScheme)
    }
    return nil
},
```

Without the scheme line, a public https host answering `302 Location: http://169.254.169.254/latest/meta-data/` gets followed, in cleartext, to the exact endpoint this package exists to block. `ErrTooManyRedirects` is declared, asserted distinct in `TestErrorsAreDistinguishable`, and never actually produced by any test.

`Fetch`'s remaining 20% is the non-2xx path (`ErrStatus`, also never produced), the `maxBytes <= 0` fallback, and the success return.

This package is a primitive built for #387's VEX ingestion, so its whole job is to be safe against a URL a user supplied. Adding tests for both closures, and checking they constrain rather than just execute by mutating each guard out and confirming something fails.

AnshulPatil2005:
## Overview

`internal/safefetch` was at 72.9%, with `New` at 53.8%. `New` is almost entirely the two closures it installs, `safeDialContext` and `CheckRedirect`, so the uncovered half was the half holding the package's guarantees.

`checkIPAllowed` was the one function at 100%, but the predicate being right is a different property from the dialer calling it.

## Additional Information

**Dial-time address check.** `TestFetch_BlocksRealLoopbackServer` reached it, for loopback only. `checkIPAllowed`'s own table covers the cloud metadata endpoint, RFC1918, carrier-grade NAT, `0.0.0.0/8` and the NAT64 forms, and nothing confirmed a dial to any of them was refused. `TestNew_DialBlocksAddress` drives the transport's `DialContext` directly, once per range.

The `localhost` case is the one worth naming. `Fetch` inspects only the scheme, so nothing about that string is refusable up front; it is blocked purely on what the resolver returned, at the moment of connection. That is the package doc's DNS-rebinding claim, and it had no test.

**Redirect policy.** No test at all. `Fetch` validates the scheme once, on the URL the caller passed; every hop after that is `CheckRedirect`'s job. Without its scheme line a public https host answering `302 Location: http://169.254.169.254/latest/meta-data/` gets followed, in cleartext, to the endpoint this package exists to block. `ErrTooManyRedirects` was declared and asserted distinct in `TestErrorsAreDistinguishable`, but no test ever produced it.

`TestNew_RedirectLimitIsCheckedBeforeScheme` pins the order between the two checks, so a downgrade cannot be used to mask a redirect loop.

**Fetch.** Covers the non-2xx path (`ErrStatus`, likewise never previously produced), the exact-limit boundary, the `maxBytes <= 0` fallback, and context propagation.

The boundary one is worth a line: `Fetch` deliberately reads `MaxBytes+1` so an oversized body is detected rather than truncated, which means a body of exactly `MaxBytes` has to still come back whole. `TestFetch_BodyExactlyAtLimit` checks both sides of that.

Coverage: 72.9% to 87.1%, `New` 53.8% to 80.8%, `Fetch` 80% to 92%.

What is still uncovered in `New` is the successful-dial branch and the resolver-failure branch, both of which need real outbound DNS or a reachable non-private address. Left alone rather than made network-dependent.

## How to Test

`go test ./internal/safefetch/ -count=1 -race -cover`

The tests constrain rather than just execute. Each guard mutated out in turn, each caught:

| mutation | fails |
| --- | --- |
| drop the https check on redirect targets | `TestNew_RedirectPolicy` |
| drop the redirect hop limit | `TestNew_RedirectPolicy`, `TestNew_RedirectLimitIsCheckedBeforeScheme` |
| drop the dial-time IP check | `TestNew_DialBlocksAddress`, `TestFetch_BlocksRealLoopbackServer` |
| drop the zero-MaxBytes fallback | `TestFetch_UnsetMaxBytesUsesDefault` |
| drop the non-2xx status check | `TestFetch_RejectsNon2xxStatus` |

Everything runs offline. The only name resolved is `localhost`, and the `httptest` TLS servers are reached with their own client, the same approach `TestFetch_RejectsOversizedBody` already uses, since they necessarily listen on loopback and `New`'s dialer correctly refuses it.

## Related issues/PRs:

* Resolved #830

### DIFF
--- a/internal/safefetch/safefetch_test.go
+++ b/internal/safefetch/safefetch_test.go
@@ -175,3 +175,234 @@ func TestCheckIPAllowed_EmbeddedIPv4(t *testing.T) {
 		})
 	}
 }
+
+// transportOf returns the SSRF-guarded transport New wired up, so the dial-time
+// checks can be driven directly. Fetch reaches them only through a real connection
+// attempt, which is why the guards below were previously exercised only for
+// loopback and not for the rest of what checkIPAllowed refuses.
+func transportOf(t *testing.T, f *Fetcher) *http.Transport {
+	t.Helper()
+	tr, ok := f.Client.Transport.(*http.Transport)
+	require.True(t, ok, "New must install an *http.Transport")
+	return tr
+}
+
+// TestNew_DialBlocksAddress is the dial-time half of the SSRF guard, and the half
+// that makes DNS rebinding not work.
+//
+// Fetch validates nothing about the destination itself: it checks the scheme and
+// hands the URL to the client. The address check happens inside DialContext,
+// against the IP the resolver just returned, at the moment the connection is made.
+// So a host that passes any earlier inspection and only then resolves to an internal
+// address is still refused, because there is no earlier inspection to pass.
+//
+// "localhost" is the case that shows it: the string carries no IP, so nothing about
+// the URL is refusable, and it is blocked purely on what it resolved to.
+func TestNew_DialBlocksAddress(t *testing.T) {
+	tests := []struct {
+		name string
+		addr string
+	}{
+		{"loopback literal", "127.0.0.1:443"},
+		{"loopback IPv6 literal", "[::1]:443"},
+		{"hostname resolving to loopback", "localhost:443"},
+		{"cloud metadata endpoint", "169.254.169.254:80"},
+		{"RFC1918 private", "10.0.0.1:443"},
+		{"carrier-grade NAT", "100.64.0.1:443"},
+		{"0.0.0.0/8, routes to localhost on Linux", "0.0.0.1:443"},
+		{"NAT64 to cloud metadata", "[64:ff9b::a9fe:a9fe]:443"},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			f := New()
+			conn, err := transportOf(t, f).DialContext(context.Background(), "tcp", tt.addr)
+			require.Error(t, err)
+			assert.ErrorIs(t, err, ErrBlockedIP)
+			assert.Nil(t, conn)
+		})
+	}
+}
+
+func TestNew_DialRejectsAddressWithoutPort(t *testing.T) {
+	f := New()
+	conn, err := transportOf(t, f).DialContext(context.Background(), "tcp", "example.com")
+	require.Error(t, err)
+	assert.Nil(t, conn)
+	assert.Contains(t, err.Error(), "splitting host/port")
+}
+
+// TestNew_DialHonoursContextCancellation checks the dial does not outlive the caller.
+// Resolution runs on the caller's context, so a cancelled one fails before any
+// connection is attempted.
+func TestNew_DialHonoursContextCancellation(t *testing.T) {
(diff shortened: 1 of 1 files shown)


matthyx (approved):
Reviewed the diff against `internal/safefetch/safefetch.go`. This is test-only (no production code changes), and every new assertion matches the actual behavior it claims to cover:

- `TestNew_DialBlocksAddress` drives `DialContext` directly for each `checkIPAllowed` range (loopback, `localhost` via resolver, cloud metadata, RFC1918, CGNAT, `0.0.0.0/8`, NAT64) — correctly exercises the dial-time check that `checkIPAllowed`'s own table test never reached.
- `TestNew_RedirectPolicy` / `TestNew_RedirectLimitIsCheckedBeforeScheme` match `CheckRedirect`'s actual order (hop-limit checked before scheme), including the downgrade-to-http and redirect-to-internal-http cases.
- `TestFetch_BodyExactlyAtLimit` correctly checks both sides of the `MaxBytes+1` boundary.
- `TestFetch_UnsetMaxBytesUsesDefault` matches the `maxBytes <= 0` fallback.
- All new IP/hostname literals used (`127.0.0.1`, `100.64.0.1`, `64:ff9b::a9fe:a9fe`, `localhost`, etc.) resolve offline, consistent with the PR description's "everything runs offline" claim.

No blockers. CI is green across build, lint, CodeQL, and GitGuardian. Approving.

PR #833, opened 2026-08-20
docs: document what VEX generation actually produces

### LINKED ISSUE #832: VEX generation has no documentation beyond one config-table line
`vexGeneration` gets one line in CONFIGURATION.md:

> | `vexGeneration` | bool | `false` | Generate VEX (Vulnerability Exploitability eXchange) documents |

and a box in the ARCHITECTURE.md scan diagram:

```
5. VEX PHASE (optional)
  • Generate VEX document
  • Include relevancy status
  • Store VEX
```

That is the whole of it. Nothing says what ends up in the document, when it gets written, or why a given finding comes out `not_affected` rather than `affected`.

None of that is guessable from outside the code:

- a finding is `affected` only if it survived relevancy filtering, so on a cluster with relevancy the same CVE reports differently depending on whether the vulnerable file was ever loaded
- a suppressed finding gets one of four different impact statements depending on where the suppression came from, and telling a backend-delivered exception policy from a CRD one is done by inspecting the ignore rule's *shape*, since cloud policies carry no `sourceKind`
- a publish is skipped entirely when the exception set was fetched incompletely, so a document can sit one scan behind its manifest, by design
- the document is named after the `cvep` argument, so `ScanCP` with relevancy produces one document per container instance while every other flow produces one per image
- statements without a kubescape `@id` are never reset, re-marked, or deduped against, so kubevuln appends beside an external statement rather than overwriting it

Answering any one of those means reading `storeVEX`, then `createVEX` and `updateVEX`, then `ignoredMatchAssessment`, then `reconcileCachedCVE`, across `core/services` and `repositories`. Several of them look like bugs until you find the comment explaining why they are not.

There is also nothing recording what is *not* implemented. Ingestion is #387; nothing here calls grype's VEX matching, and `internal/safefetch`, `internal/vexdoc` and `internal/csafresolve` each exist for it and are wired into nothing. That is not obvious from the package list.

Writing `docs/VEX.md` covering the document shape and location, statement identity and authorship, how each status and impact statement is chosen, the relevancy split, the three call sites, the completeness gate and the cache-hit removal rule, the update and no-op-detection sequence, the canonical hash, the legacy repair passes in `updateVEX`, and troubleshooting for the cases that look wrong and are not. Plus links to it from the two places that currently mention VEX in passing.

AnshulPatil2005:
## Overview

`vexGeneration` had one line in CONFIGURATION.md and a box in the ARCHITECTURE.md scan diagram. Nothing said what ends up in the document, when it gets written, or why a finding comes out `not_affected` rather than `affected`.

`docs/VEX.md` is that, 372 lines. Linked from the `vexGeneration` row in CONFIGURATION.md and the storage section of ARCHITECTURE.md.

## Additional Information

The behaviour is not guessable from outside the code, and answering any single question about it meant reading `storeVEX`, then `createVEX` and `updateVEX`, then `ignoredMatchAssessment`, then `reconcileCachedCVE`, across two packages.

What the page covers:

- **Document shape and location.** The container, its name, and why that name differs by flow. `ScanCP` with relevancy produces one document per container instance; everything else produces one per image, which falls out of the container being named after the `cvep` argument.
- **Statement identity.** The `https://kubescape.io/vex/statement/...` prefix is what marks a statement as kubescape's, and every maintenance step is local-only, so an external statement is never reset, re-marked or deduped against.
- **How a status is chosen.** The `not_affected` baseline, promotion to `affected` via relevancy, and the four impact statements a suppressed finding can get. Including why telling a backend-delivered exception policy from a CRD one is done by inspecting the ignore rule's shape: cloud policies carry no `sourceKind`.
- **The completeness gate**, and why the manifest is persisted on a looser condition than the VEX document. A partial exception set can only under-suppress, which is safe for a manifest and understates a published assertion.
- **The update sequence**, including the no-op detection that keeps repeated scans of an unchanged image from bumping `version` forever, and the `GetOptions{ResourceVersion: "metadata"}` hazard that no test can catch because the fake clientset ignores `GetOptions`.
- **The canonical hash**, and why length-prefixing and sorting are both load-bearing.
- **The two legacy repair passes** at the top of `updateVEX`, which are otherwise very hard to place.
- **Troubleshooting**, aimed at the cases that look like bugs and are not: a stale document, a `version` that never advances, silence when `vexGeneration` is off.
- **What is not implemented.** Ingestion is #387, nothing calls grype's VEX matching, and `internal/safefetch`, `internal/vexdoc` and `internal/csafresolve` each exist for it and are wired into nothing. That last part is not obvious from the package list, and the "ignored by an external VEX document or scanner configuration" impact statement having no producer today makes more sense next to it.

## How to Test

Docs only, no code changes. `go build ./... && go test ./... ` is unaffected.

The parts most likely to rot are the ones taken from types rather than prose, so they were read out rather than written from memory:

- every JSON field name in the two examples comes from the `v1beta1` tags in `kubescape/storage` (`@context`, `role`, `last_updated`, `impact_statement`, `status_notes` and the rest)
- the `spec` example is not nested under a `metadata:` key, because `VEX` embeds `Metadata` with no JSON tag, so it inlines. That is called out explicitly, since the Go type reads the other way
- the example statement `@id` is the actual `url.PathEscape` output, `pkg:deb%2Fxz-utils@5.6.0`. `PathEscape` escapes the `/` but leaves `:` and `@` alone

Checked mechanically: 14 headings, no broken internal anchors, no broken relative links to the five source paths and three sibling docs it references.

## Related issues/PRs:

* Resolved #832

### DIFF
--- a/docs/VEX.md
+++ b/docs/VEX.md
@@ -1,0 +1,372 @@
+# VEX
+
+kubevuln can publish a [VEX](https://openvex.dev/) document alongside every vulnerability
+manifest it stores. VEX (Vulnerability Exploitability eXchange) is a machine-readable
+assertion about whether a vulnerability that a scanner *found* in an image is actually
+*exploitable* in it. A scanner reports what is present; a VEX document says what that
+presence means.
+
+That distinction is the whole point of the feature here. Relevancy data and
+`SecurityException` CRDs both change how a finding should be read without changing whether
+Grype matched it, and a VEX document is how kubevuln states that in a format other tools
+can consume.
+
+This page covers the document kubevuln **produces**. Consuming VEX documents published by
+someone else is not implemented; see [Ingestion](#ingestion-not-implemented) at the end.
+
+- [Enabling it](#enabling-it)
+- [Where the document lives](#where-the-document-lives)
+- [Document shape](#document-shape)
+- [Statement identity and authorship](#statement-identity-and-authorship)
+- [How a status is chosen](#how-a-status-is-chosen)
+- [Relevancy](#relevancy)
+- [When a document is written](#when-a-document-is-written)
+- [The completeness gate](#the-completeness-gate)
+- [Updating an existing document](#updating-an-existing-document)
+- [The canonical hash](#the-canonical-hash)
+- [Legacy statement repair](#legacy-statement-repair)
+- [Ingestion, not implemented](#ingestion-not-implemented)
+- [Troubleshooting](#troubleshooting)
+
+## Enabling it
+
+| setting | default | effect |
+| --- | --- | --- |
+| `vexGeneration` | `false` | publish VEX documents |
+| `storage` | `false` | required: VEX is written through the storage API server, so nothing is published without it |
+| `riskAcceptance` | `false` | required for `SecurityException` CRDs to influence statements at all |
+
+```json
+{
+  "storage": true,
+  "vexGeneration": true,
+  "riskAcceptance": true
+}
+```
+
+`vexGeneration` is checked in one place, `ScanService.storeVEX` in
+[`core/services/scan.go`](../core/services/scan.go). When it is off, that function returns
+immediately and no call site knows the difference. When it is on but `storage` is off, the
+scan flows never reach the call at all, since VEX is only written on paths already gated on
+storage.
+
+Failure to store a VEX document is logged as a warning and the scan continues. The CVE
+manifest is written before the VEX document at every call site, and the report sent to the
+backend does not depend on VEX, so a scan result stands without it.
+
+## Where the document lives
+
+Documents are stored as `OpenVulnerabilityExchangeContainer` custom resources in the
+configured namespace, one per stored vulnerability manifest. The container's name, labels
+and annotations are taken from the manifest being described.
+
+The name is the part worth knowing, because it differs by flow:
+
+| flow | document named after | granularity |
+| --- | --- | --- |
+| `ScanCP` with relevancy | the relevancy-filtered manifest, whose name is the instance ID slug | one per container instance |
+| `ScanCVE`, `ScanRegistry`, cache-hit republish | the image manifest, whose name is the image slug | one per image |
+
+That falls out of `StoreVEX` naming the container after its `cvep` argument. Both call
+shapes are covered under [When a document is written](#when-a-document-is-written).
+
+Note that `StoreVEX`'s `withRelevancy` parameter is accepted by the port interface and
+discarded by the API server implementation. What actually distinguishes a relevancy
+document from a plain one is *which manifests are passed*, not the flag.
+
+## Document shape
+
+The `Spec` is an OpenVEX document:
+
+```yaml
+apiVersion: spdx.softwarecomposition.kubescape.io/v1beta1
+kind: OpenVulnerabilityExchangeContainer
+metadata:
+  name: <cve manifest name>
+spec:
+  "@context": https://openvex.dev/ns/v0.2.0
+  "@id": <canonical hash, see below>
+  author: kubescape.io
+  role: "smart vulnerability scanner :-)"
+  timestamp: <RFC3339, set once at creation>
+  last_updated: <RFC3339, bumped on every real change>
+  version: 0
+  tooling: kubescape-vulnerability-analyzer
+  statements:
+    - ...
+```
+
+The document metadata is embedded directly in `spec` rather than nested under a
+`metadata` key, since `VEX` embeds `Metadata` without a JSON tag. The container's own
+`metadata` is the usual Kubernetes `ObjectMeta`.
+
+`version` starts at `0` and increments only when an update actually changes something.
+
+Each statement covers one (vulnerability, package) pair:
+
+```yaml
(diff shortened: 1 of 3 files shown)


docs/VEX.md
```diff
@@ -0,0 +1,372 @@
...
+suppress through the same code path as CRDs but never go through `buildPolicy`, so they
+carry no `sourceKind` and their ignore rule carries no `SourceKind`. Provenance is
+recovered from rule *shape* instead: kubevuln writes exactly one rule per suppression and
+never sets `Package`, while Grype expresses its own ignore rules in terms of a package, so
+a rule with a package did not come from us. Only rules failing that test get the "external
+VEX document or scanner configuration" wording.
+
+Today nothing produces that last row. `ApplySecurityExceptions` is the only writer of
+`IgnoredMatches`, and Grype is handed no VEX documents or ignore rules. It exists for the
+ingestion work in [#387](https://github.com/kubescape/kubevuln/issues/387).
```
matthyx, line 196: Non-blocking: `isOwnIgnoreRule` actually has two conditions (`repositories/apiserver.go` around line 1223), not one — `len(AppliedIgnoreRules) == 1 && AppliedIgnoreRules[0].Package == nil`. Since `buildIgnoreRules` in `adapters/v1/securityexception.go` emits one rule per *suppressing policy* (and the code's own comment near `ApplySecurityExceptions` anticipates two policies matching the same finding), a backend-delivered suppression matched by two policies also fails this test today. So "Today nothing produces that last row" isn't quite right — worth softening to note the multi-policy case, or at least mentioning the "exactly one rule" half of the condition explicitly.

docs/VEX.md
```diff
@@ -0,0 +1,372 @@
...
+parsing the zeroed timestamp. The fake clientset used in tests ignores `GetOptions`
+entirely, so no test can catch a regression here.
+
+The update itself:
+
+1. Append a statement for any (vulnerability, package) pair not already covered by a local
+   statement.
+2. Reset every local statement to the baseline `not_affected` shape, discarding the previous
+   scan's conclusions. External statements are skipped.
+3. Re-apply suppression assessments, then promote relevant findings to `affected`.
```
matthyx, line 279: Non-blocking: in `updateVEX` (and `createVEX`), suppression assessment is actually applied a second time *after* the promote-to-`affected` pass (`repositories/apiserver.go` ~lines 1646-1652), so on overlap between `cvep.Matches` and `cve.IgnoredMatches` (possible in `ScanCP`, whose two manifests come from separate `GetCVEExceptions` calls), suppression wins over relevancy rather than the other way around. Worth a note that mark-ignored runs once more after the promote step.

docs/VEX.md
```diff
@@ -0,0 +1,372 @@
...
+| --- | --- | --- |
+| `SecurityException` CRD, default | `not_affected` | `Vulnerability was ignored by a SecurityException` |
+| `SecurityException` with `fixed` status | `fixed` | cleared, justification cleared |
+| `SecurityException` with `affected` status | `affected` | cleared, action statement records the accepted risk |
+| backend-delivered exception policy | `not_affected` | `Vulnerability was ignored by an exception policy` |
+| anything else | `not_affected` | `Vulnerability was ignored by an external VEX document or scanner configuration` |
+
+A `SecurityException` carrying a justification or impact statement of its own has those
+copied onto the statement, and an unrecognised status falls back to the safe
+`not_affected` shape rather than being propagated.
```
matthyx, line 184: Non-blocking: the "justification or impact statement... copied onto the statement" behavior only applies to the `default`/fallback branch in `ignoredMatchAssessment`. For `fixed` and `affected` statuses, both fields are cleared and the SecurityException's wording instead goes into `StatusNotes` (`repositories/apiserver.go` ~1189-1210, 1240-1249) — a field this doc (and the "Document shape" example above) never mentions. Might be worth a line scoping this to the fallback case and naming `status_notes`.

docs/VEX.md
```diff
@@ -0,0 +1,372 @@
...
+document accumulating every status a finding has ever held.
+
+## The canonical hash
+
+`metadata.@id` is a hash over the document's meaningful content, following the OpenVEX
+canonicalisation algorithm: document timestamp, version and author, then every statement in
+sorted order contributing its vulnerability, status, justification, timestamp and sorted
+product strings, each length-prefixed.
+
+Length-prefixing every field is what stops two different documents hashing the same by
```
matthyx, line 300: Nit: "Length-prefixing every field" is slightly over-general — `Version` (an int) is written raw, unprefixed, in the hash `Sprintf`. Harmless since it's numeric, but "every field" isn't literally true.

matthyx (approved):
Docs-only PR, no code changes. I fact-checked the new `docs/VEX.md` (statement identity, status-selection table, the `GetOptions{ResourceVersion: "metadata"}` hazard, the canonical-hash algorithm, legacy repair passes, naming/granularity by flow, config gating, links) against the current `core/services/scan.go` and `repositories/apiserver.go` — everything load-bearing checks out, including the harder-to-verify details (the escaped statement `@id` example, the completeness-gate asymmetry). Left 4 non-blocking inline comments on small precision gaps (two suppression/relevancy ordering nuances, an under-scoped claim about where `status_notes` fits in, and an "every field" overstatement in the hash description) — worth folding in but nothing here would mislead a reader on the feature's actual behavior. Approving.

PR #837, opened 2026-08-21
test(metrics): cover the three recorders with no test and the unpinned outcome labels

### LINKED ISSUE #836: three metrics recorders have no test, and three of RecordSourceResolution's four outcome labels are unpinned
`internal/metrics` is at 68.3%, and the gap is concentrated in the exported surface rather than spread thin:

```
metrics.go:93:	New			74.5%
metrics.go:251:	Shutdown		 0.0%
metrics.go:266:	RecordScanFallback	83.3%
metrics.go:284:	RecordRegistryAuthCache	83.3%
metrics.go:297:	RecordSourceResolution	72.7%
metrics.go:319:	RecordSingleflightHit	 0.0%
metrics.go:331:	RecordRetryAttempt	 0.0%
total:						68.3%
```

Three exported functions have no test at all, and all three have real production callers:

- `RecordSingleflightHit`, called from `ScanService.getOrCreateSBOM`. It is the only evidence singleflight is deduplicating anything, so if the `target` label is wrong the series cannot be attributed to a stage
- `RecordRetryAttempt`, called from three sites in `RetryWithBackoff` with three different outcomes. `attempt` means the operation is under rate-limit pressure, `success` means backing off resolved it, `exhausted` means it did not and the scan failed. Nothing checks the three stay distinct
- `Shutdown`, called by `cmd/http` in its graceful-shutdown sequence

`RecordSourceResolution` is the more interesting one. It maps `(usedFallback, success)` onto four outcome labels, and `TestNewAndHandler` exercises exactly one of them:

```go
RecordSourceResolution(context.Background(), ComponentInProcess, true, true)
```

So three of the four switch arms are free to be wrong. Two of them look similar and mean opposite things: `fallback_assisted_success` is the registry-auth ladder working as designed, `first_pass_failure` is a resolution that failed without a fallback ever being tried. An operator reading a dashboard cannot tell those apart if the mapping is wrong, and nothing would fail.

There is also a documented safety property covered for exactly one recorder. `TestRecordTempDirSweep_BeforeNewIsANoOp` pins that `RecordTempDirSweep` is safe against a nil counter, with the note that `cmd/sbom-scanner` never calls `metrics.New()`. That reasoning applies to every recorder the sidecar can reach, and `RecordScanFallback` alone is called from seven places in `pkg/sbomscanner/v1/server.go`. Only one of them is covered.

Adding tests for the three uncovered functions, all four `RecordSourceResolution` outcomes, and the nil-counter path for the remaining recorders. Asserting against the exposition text the `/metrics` endpoint actually serves, the way the existing tests do, rather than the SDK's internal state.

AnshulPatil2005:
## Overview

`internal/metrics` was at 68.3%, concentrated in the exported surface. `Shutdown`, `RecordSingleflightHit` and `RecordRetryAttempt` were at 0% each, and all three have production callers: `cmd/http`'s graceful-shutdown sequence, `ScanService.getOrCreateSBOM`, and three sites in `RetryWithBackoff`.

68.3% to 86.1%. Every recorder and `Shutdown` now at 100%; what is left in `New` is instrument-construction error paths.

## Additional Information

`RecordSourceResolution` was the gap worth the most. It maps `(usedFallback, success)` onto four outcome labels, and `TestNewAndHandler` exercised exactly one:

```go
RecordSourceResolution(context.Background(), ComponentInProcess, true, true)
```

Three switch arms were free to be wrong, and two of them mean opposite things while reading similarly on a dashboard: `fallback_assisted_success` is the registry-auth ladder working as designed, `first_pass_failure` is a resolution that failed before any fallback was tried. `TestRecordSourceResolution_OutcomesAreDistinct` guards the mapping as a whole, so one arm cannot be collapsed into another.

The nil-counter safety property was pinned for `RecordTempDirSweep` alone, with the note that `cmd/sbom-scanner` never calls `metrics.New`. That reasoning applies to every recorder the sidecar reaches, and `RecordScanFallback` alone is called from seven places in `pkg/sbomscanner/v1/server.go`. `TestRecordersBeforeNewAreNoOps` extends it to the rest.

Assertions read the exposition text `/metrics` actually serves, matching the existing tests, so a renamed label fails rather than passing against SDK internals.

`TestShutdown` restores a live provider on the way out. `New` installs the package-level counters, so without that, test order would decide whether later tests can record anything.

## How to Test

`go test ./internal/metrics/ -count=1 -race -cover`

Five mutations, each caught:

| mutation | fails |
| --- | --- |
| collapse the `fallback_failed` arm into the default | `TestRecordSourceResolution_AllOutcomes`, `..._OutcomesAreDistinct` |
| collapse the `first_pass_success` arm | `TestRecordSourceResolution_AllOutcomes`, `..._OutcomesAreDistinct` |
| drop the `target` label from singleflight hits | `TestRecordSingleflightHit` |
| drop the `outcome` label from retry attempts | `TestRecordRetryAttempt` |
| make `Shutdown` a no-op returning nil | `TestShutdown` |

The last one is worth calling out. A first pass at `TestShutdown` only asserted a nil error and no panic, and the no-op mutation survived it, since a no-op returns nil too. `Shutdown` turns out to be observable, so the test now checks what actually happened: the reader stopped, so a scrape still answers 200 with an empty body, and a second `Shutdown` reports the reader is already down, which a `Shutdown` that never reached the provider could not do.

## Related issues/PRs:

* Resolved #836

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive coverage for metrics collection, including retries, duplicate requests, source-resolution outcomes, and shutdown behavior.
  * Verified metric labels and counts through the metrics endpoint.
  * Confirmed safe behavior when metrics are uninitialized or after shutdown, without panics.

### DIFF
--- a/internal/metrics/metrics_test.go
+++ b/internal/metrics/metrics_test.go
@@ -97,3 +97,179 @@ func TestMeterRegistersObservableGauge(t *testing.T) {
 
 	assert.True(t, strings.Contains(w.Body.String(), "kubevuln_test_gauge_local"), w.Body.String())
 }
+
+// scrape renders the current metric state the way the /metrics endpoint would, so the
+// assertions below read the exposition text an operator actually sees rather than the
+// SDK's internal state.
+func scrape(t *testing.T, m *Metrics) string {
+	t.Helper()
+	req := httptest.NewRequest("GET", "/metrics", nil)
+	w := httptest.NewRecorder()
+	m.Handler().ServeHTTP(w, req)
+	require.Equal(t, 200, w.Code)
+	return w.Body.String()
+}
+
+// TestRecordSingleflightHit covers the counter that says how often a duplicate SBOM
+// request was served by an in-flight one rather than pulling the image again. It is the
+// only evidence singleflight is doing anything, so the target label has to be right:
+// without it the series cannot be attributed to a stage.
+func TestRecordSingleflightHit(t *testing.T) {
+	m, err := New()
+	require.NoError(t, err)
+
+	// "sbom_generation" is the only target used in production today
+	// (ScanService.getOrCreateSBOM), so it is the one worth pinning.
+	RecordSingleflightHit(context.Background(), "sbom_generation")
+	RecordSingleflightHit(context.Background(), "sbom_generation")
+	RecordSingleflightHit(context.Background(), "other_target")
+
+	body := scrape(t, m)
+	assert.Contains(t, body, `kubevuln_singleflight_hits_total{target="sbom_generation"} 2`, body)
+	assert.Contains(t, body, `kubevuln_singleflight_hits_total{target="other_target"} 1`, body)
+}
+
+// TestRecordRetryAttempt covers all three outcomes RetryWithBackoff can report. They are
+// distinct signals and only useful if they stay distinct: "attempt" says the operation is
+// under rate-limit pressure, "success" says backing off resolved it, and "exhausted" says
+// it did not and the scan failed.
+func TestRecordRetryAttempt(t *testing.T) {
+	m, err := New()
+	require.NoError(t, err)
+
+	ctx := context.Background()
+	RecordRetryAttempt(ctx, "sbom_generation", RetryOutcomeAttempt)
+	RecordRetryAttempt(ctx, "sbom_generation", RetryOutcomeAttempt)
+	RecordRetryAttempt(ctx, "sbom_generation", RetryOutcomeSuccess)
+	RecordRetryAttempt(ctx, "sbom_generation", RetryOutcomeExhausted)
+
+	body := scrape(t, m)
+	assert.Contains(t, body, `kubevuln_retry_attempts_total{operation="sbom_generation",outcome="attempt"} 2`, body)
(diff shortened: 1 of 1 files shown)


matthyx (approved):
Reviewed. This is test-only (no production code changes), all CI checks are green, and coverage moves 68.3% → 86.1% as claimed.

Verified independently rather than just trusting the PR description:
- Ran `go test ./internal/metrics/... -race -cover` on the PR head — all tests pass, coverage matches (86.1%).
- Checked every new assertion against the actual `RecordSingleflightHit`/`RecordRetryAttempt`/`RecordSourceResolution` signatures and label names in `metrics.go` — all correct.
- Reproduced two of the listed mutations by hand to confirm the tests actually catch regressions, not just pass trivially:
  - Collapsing the `fallback_failed` switch arm into `first_pass_failure` → `TestRecordSourceResolution_OutcomesAreDistinct` fails as claimed.
  - Making `Shutdown` a no-op returning `nil` → `TestShutdown` fails as claimed (both the empty-scrape and second-Shutdown-error assertions).

No blockers. Approving.

PR #839, opened 2026-08-21
fix(repositories): stop MemoryStore answering a manifest read with a summary write

### LINKED ISSUE #838: MemoryStore's summary write answers a manifest read, because both use the same map
`MemoryStore` is the `CVERepository` every `core/services` test runs against. `StoreCVESummary` writes into `m.cveManifests`, which is the same map `StoreCVE` writes and `GetCVE` reads:

```go
func (m *MemoryStore) StoreCVESummary(ctx context.Context, cve, cvep domain.CVEManifest, withRelevancy bool) error {
	...
	if withRelevancy {
		idSumm := cveID{Name: cvep.Name, ...}
		m.cveManifests[idSumm] = cvep
	}
	m.cveManifests[id] = cve
	return nil
}
```

Two consequences.

**A summary write answers a manifest read.** Against `APIServerStore` these are separate resources, `VulnerabilityManifest` and `VulnerabilityManifestSummary`, so storing a summary can never make `GetCVE` find a manifest. Here it can. A flow that summarises a manifest without storing it looks cached on the next scan against the double and would miss in production, which is the wrong direction for a test double to be wrong in.

`ScanCP`'s relevancy branch is where the two diverge: it calls `StoreCVE(ctx, filteredCvep, true)` and then `StoreCVESummary(ctx, filteredCve, filteredCvep, true)`, so `filteredCve` reaches the manifest map through the summary call. On the cache-hit path `reconcileCachedCVE` only stores `filteredCve` when suppressions actually changed and the removal-safety rule allows it, so there is a real case where the double holds a manifest the real store does not.

**The two calls overwrite each other.** Same `cveID`, same map, so whichever runs last wins and neither is inspectable afterwards. That is already written down in the codebase, in the workaround it forced. From `core/services/scan_test.go`:

```go
// summaryCapturingRepo wraps a ports.CVERepository and records every StoreCVESummary call
// verbatim, so a test can inspect exactly which manifest a scan flow handed to the summary
// write. MemoryStore alone can't answer that: StoreCVE and StoreCVESummary key their map off
// the same cveID, so a later call silently overwrites an earlier one's entry.
```

A test had to wrap the repository because the double could not answer the question.

---

Separately, and probably why none of this was noticed: **`GetCVESummary` has had no production caller since April.**

`f57f6af` ("fix: also apply exceptions to cached CVE manifests", 2026-04-17) replaced the cache-hit "store the summary only if it does not exist" check with an unconditional re-store, deleting both call sites:

```go
-			if cveSumm, err := s.cveRepository.GetCVESummary(ctx); err != nil || cveSumm == nil {
-				err = s.cveRepository.StoreCVESummary(ctx, cve, domain.CVEManifest{}, false)
```

A repo-wide search now finds `GetCVESummary` only in `core/ports/repositories.go`, its three implementations, and their tests. Nothing in `core/services` or `controllers` calls it. It has still received maintenance in that window: #745, #825 and #827.

`MemoryStore.GetCVESummary` is a stub whose comment says otherwise:

```go
// GetCVESummary returns a CVE summary from an in-memory map
func (m *MemoryStore) GetCVESummary(ctx context.Context) (*v1beta1.VulnerabilityManifestSummary, error) {
	...
	return nil, nil
}
```

It never touches a map.

Giving summaries their own map, adding a `CVESummaries` accessor so the summary write is inspectable without wrapping the repository, and correcting the comments. Whether `GetCVESummary` should stay on the port at all is a maintainer call, so leaving it in place.

AnshulPatil2005:
## Overview

`MemoryStore.StoreCVESummary` wrote into `m.cveManifests`, the map `StoreCVE` writes and `GetCVE` reads. Against `APIServerStore` those are separate resources, `VulnerabilityManifest` and `VulnerabilityManifestSummary`, so a summary write can never make `GetCVE` find a manifest. Against the double it could.

Summaries now go in a map of their own, with a `CVESummaries` accessor so a test can inspect the summary write without wrapping the repository.

## Additional Information

`ScanCP`'s relevancy branch is where the two stores diverge. It stores `filteredCvep` as the manifest and then hands `filteredCve` to the summary write, so `filteredCve` reached the manifest map through the summary call. On a cache hit `reconcileCachedCVE` only stores `filteredCve` when suppressions actually changed and the removal-safety rule allows it, so there is a real case where the double holds a manifest the real store does not. A flow can look cached here and miss in production, which is the wrong direction for a double to be wrong in.

Sharing one map also meant the two calls overwrote each other under the same `cveID`, so neither was readable afterwards. That was already known, and worked around rather than fixed. From `core/services/scan_test.go`:

```go
// summaryCapturingRepo wraps a ports.CVERepository and records every StoreCVESummary call
// verbatim, so a test can inspect exactly which manifest a scan flow handed to the summary
// write. MemoryStore alone can't answer that: StoreCVE and StoreCVESummary key their map off
// the same cveID, so a later call silently overwrites an earlier one's entry.
```

`CVESummaries` is what that wrapper needed. It is left in place, since it also records call order and `withRelevancy`, which the accessor does not.

Two comments corrected while here.

`StoreCVESummary` now records why the maps are separate, so the next person to reach for the manifest map has the reason in front of them.

`GetCVESummary` no longer claims to "return a CVE summary from an in-memory map". It returns `nil, nil` unconditionally and never touches a map. It stays a stub, because building a real `VulnerabilityManifestSummary` in the double would mean duplicating `parseSeverities` and `parseVulnerabilitiesComponents`, and nothing depends on it: the port declares `GetCVESummary`, but no production caller has existed since `f57f6af` replaced the cache-hit "store the summary only if it does not exist" check with an unconditional re-store. Whether it stays on the port at all is a maintainer call, so this changes nothing about that.

## How to Test

`go build ./... && go vet ./... && go test ./... -count=1`

The full suite passes unchanged with the maps separated, which is the evidence that nothing depended on the conflation.

Reverting just the map split, so summaries share the manifest map again, fails the three new tests:

```
--- FAIL: TestMemoryStore_SummaryWriteDoesNotSatisfyManifestRead
--- FAIL: TestMemoryStore_ManifestAndSummaryDoNotOverwriteEachOther
--- FAIL: TestMemoryStore_CVESummaries
```

The first is the property itself: store a summary, then `GetCVE` must still report nothing. The second covers the overwrite, asserting the manifest survives a later summary write and that the summary is separately readable. The third covers the accessor, including the relevancy case where both manifests are recorded, and the sorted order that keeps the result independent of Go's map iteration.

`TestMemoryStore_GetCVESummary_IsAStub` pins the stub as deliberate, so a change that starts returning something has to come here too.

## Related issues/PRs:

* Resolved #838

### DIFF
--- a/repositories/memory.go
+++ b/repositories/memory.go
@@ -1,7 +1,9 @@
 package repositories
 
 import (
+	"cmp"
 	"context"
+	"slices"
 
 	"github.com/kubescape/kubevuln/core/domain"
 	"github.com/kubescape/kubevuln/core/ports"
@@ -30,6 +32,7 @@ type sbomID struct {
 type MemoryStore struct {
 	aps          map[apID]v1beta1.ContainerProfile
 	cveManifests map[cveID]domain.CVEManifest
+	cveSummaries map[cveID]domain.CVEManifest
 	sboms        map[sbomID]domain.SBOM
 	summaryStubs []string // statuses passed to StoreCVESummaryStub, recorded for tests
 	sbomStores   int      // number of StoreSBOM calls, recorded for tests
@@ -54,6 +57,7 @@ func NewMemoryStorage(getError, storeError bool) *MemoryStore {
 	return &MemoryStore{
 		aps:          map[apID]v1beta1.ContainerProfile{},
 		cveManifests: map[cveID]domain.CVEManifest{},
+		cveSummaries: map[cveID]domain.CVEManifest{},
 		sboms:        map[sbomID]domain.SBOM{},
 		getError:     getError,
 		storeError:   storeError,
@@ -115,7 +119,14 @@ func (m *MemoryStore) GetCVE(ctx context.Context, name, SBOMCreatorVersion, CVES
 	return domain.CVEManifest{}, nil
 }
 
-// GetCVESummary returns a CVE summary from an in-memory map
+// GetCVESummary always reports no summary. It is a stub rather than a read of what
+// StoreCVESummary recorded, because building a real VulnerabilityManifestSummary here would
+// mean duplicating parseSeverities and parseVulnerabilitiesComponents in the test double.
+//
+// Nothing depends on it: the CVERepository port declares GetCVESummary, but no production
+// caller has existed since the cache-hit "store the summary only if it does not exist"
+// check was replaced with an unconditional re-store. Tests wanting to know what a scan flow
+// handed to the summary write should use CVESummaries.
 func (m *MemoryStore) GetCVESummary(ctx context.Context) (*v1beta1.VulnerabilityManifestSummary, error) {
 	_, span := otel.Tracer("").Start(ctx, "MemoryStore.GetCVESummary")
 	defer span.End()
@@ -141,7 +152,14 @@ func (m *MemoryStore) StoreCVE(ctx context.Context, cve domain.CVEManifest, _ bo
 	return nil
 }
 
-// StoreCVESummary stores a CVE summary to an in-memory map
+// StoreCVESummary records the manifests a summary write was given, in a map of its own.
+//
+// It used to write them into m.cveManifests, the map StoreCVE writes and GetCVE reads, which
+// made a summary write answer a manifest read: a flow that stored a summary without storing
+// the manifest looked cached on the next scan here, and did not against APIServerStore,
+// where the two are separate resources (VulnerabilityManifest, VulnerabilityManifestSummary).
+// Sharing the map also meant the two calls overwrote each other's entry under the same
+// cveID, so neither could be inspected afterwards.
 func (m *MemoryStore) StoreCVESummary(ctx context.Context, cve domain.CVEManifest, cvep domain.CVEManifest, withRelevancy bool) error {
 	_, span := otel.Tracer("").Start(ctx, "MemoryStore.StoreCVESummary")
 	defer span.End()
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Verified end to end:

- Build, vet, and the full test suite (`go build ./...`, `go vet ./...`, `go test ./... -count=1`) pass clean on `fix/memorystore-summary`.
- Confirmed `MemoryStore` is a test-only double (only referenced from `*_test.go` files); no production code path is affected by the map split.
- Confirmed the bug claim: reverting the `cveSummaries` split back to `cveManifests` reproduces exactly the three failures described in the PR (`SummaryWriteDoesNotSatisfyManifestRead`, `ManifestAndSummaryDoNotOverwriteEachOther`, `CVESummaries`), and restores the previous behavior where `StoreCVESummary` answers a `GetCVE` read and the two calls clobber each other under the same `cveID`.
- Confirmed against `apiserver.go` that `VulnerabilityManifest` and `VulnerabilityManifestSummary` are genuinely separate resources there, so the double's prior shared-map behavior was a real divergence from production, in the direction that hides bugs (looks cached in tests, misses in prod).
- Confirmed `GetCVESummary`'s "no production caller" claim: only `core/ports`, `broken.go`, and `apiserver.go` define/declare it; `core/services/scan.go` never calls it.
- New tests are well-targeted at the property that broke, sorted output removes map-iteration flakiness, gofmt clean, CI green.

No blockers. Approving.

PR #841, opened 2026-08-21
fix(vexdoc): drop the error crypto/rand.Read cannot return, and cover the one that happens

### LINKED ISSUE #840: vexdoc's randomFilename returns an error crypto/rand.Read cannot produce
`internal/vexdoc` is at 66.7%, the lowest in the repo. Every uncovered block is an error path, and one of them cannot happen.

`randomFilename` returns an error from `crypto/rand.Read`:

```go
func randomFilename() (string, error) {
	b := make([]byte, 16)
	if _, err := rand.Read(b); err != nil {
		return "", err
	}
	return fmt.Sprintf("kubevuln-vexdoc-%s.json", hex.EncodeToString(b)), nil
}
```

That error is unreachable. From `crypto/rand` in the toolchain this module builds with:

> Read fills b with cryptographically secure random bytes. **It never returns an error**, and always fills b entirely.
>
> Read calls `io.ReadFull` on `Reader` and **crashes the program irrecoverably** if an error is returned.

It has been infallible since Go 1.24, and `go.mod` says `go 1.26.0`, so there is no build of this module where the branch can be taken. It costs two dead blocks: the check inside `randomFilename`, and the caller's check in `WriteToTempFile`:

```go
	name, err := randomFilename()
	if err != nil {
		return "", noop, fmt.Errorf("vexdoc: generating a random filename: %w", err)
	}
```

Of the remaining uncovered paths, one is genuinely reachable and untested: `os.OpenFile` failing because the temp directory is not usable. That is the only failure `WriteToTempFile` can actually hit in practice, since the filename is generated rather than supplied, and it is the path where the contract matters most. A caller is told to `defer cleanup()` immediately after checking `err`, so a failed create must still hand back a non-nil, safe cleanup, and must not hand back a path for a file that was never created. Nothing checks either.

The other two, `f.Write` and `f.Close` failing, are not reachable from a test without changing the signature to accept a writer. Not proposing that: this package exists to close one concrete blocker for #387 (grype's `vex.Processor` only accepts real file paths), and widening its API to make two error branches testable is not worth it.

So: drop the dead error plumbing from `randomFilename`, and cover the create-failure path and the filename properties the debugging story depends on.

AnshulPatil2005:
## Overview

`randomFilename` returned an error from `crypto/rand.Read`. That call never produces one, so the branch and the caller's check of it were both dead. Removing them, and adding a test for the failure `WriteToTempFile` can actually hit.

`internal/vexdoc` was at 66.7%, the lowest in the repo. Now 75.0%, `randomFilename` at 100%.

## Additional Information

From `crypto/rand` in the toolchain this module builds with:

> Read fills b with cryptographically secure random bytes. It never returns an error, and always fills b entirely.
>
> Read calls `io.ReadFull` on `Reader` and crashes the program irrecoverably if an error is returned.

Infallible since Go 1.24, and `go.mod` says `go 1.26.0`, so there is no build of this module where the check could fire. It cost two uncovered blocks: the check inside `randomFilename` and `WriteToTempFile`'s check of its return, together a third of what was uncovered in the package.

The reachable failure had no test: `os.OpenFile` failing because the temp directory is not usable. It is the only one `WriteToTempFile` can really hit, since the filename is generated rather than supplied, and it is where the contract matters most. Callers are told to `defer cleanup()` immediately after checking `err`, so a failed create has to hand back a non-nil, safe cleanup, and must not hand back a path for a file that was never created. Neither was checked.

`os.TempDir` is redirected by setting `TMPDIR`, `TMP` and `TEMP` and then confirming it actually moved, skipping if not, rather than assuming which one a platform reads. It works on both Linux and Windows, so the test runs rather than skips in CI and locally.

`f.Write` and `f.Close` failing stay uncovered, and this does not try to change that. Reaching them means widening the signature to accept a writer, and this package exists to close one concrete blocker for #387, that grype's `vex.Processor` only accepts real file paths. Not worth reshaping its API for two error branches.

## How to Test

`go build ./... && go vet ./... && go test ./internal/vexdoc/ -count=1 -race -cover`

The full suite passes unchanged. `randomFilename` is package-private with one caller, so dropping its error return touches nothing outside this file.

Three mutations, each caught:

| mutation | fails |
| --- | --- |
| return a path even though the file was never created | `TestWriteToTempFile_ReportsCreateFailure` |
| return a nil cleanup on the create-failure path | `TestWriteToTempFile_ReportsCreateFailure` |
| stop randomising the filename | `TestRandomFilename`, `..._TwoDocumentsNeverCollide`, `..._ManyCallsNeverProduceADuplicatePath` |

`TestWriteToTempFile_CreateFailureIsNotMistakenForAnEmptyDocument` keeps the two error paths apart: both return an empty path, so a caller distinguishing them depends on the errors, and only one is `ErrEmptyDocument`.

## Related issues/PRs:

* Resolved #840

### DIFF
--- a/internal/vexdoc/vexdoc.go
+++ b/internal/vexdoc/vexdoc.go
@@ -59,12 +59,7 @@ func WriteToTempFile(data []byte) (path string, cleanup func(), err error) {
 		return "", noop, ErrEmptyDocument
 	}
 
-	name, err := randomFilename()
-	if err != nil {
-		return "", noop, fmt.Errorf("vexdoc: generating a random filename: %w", err)
-	}
-
-	path = filepath.Join(os.TempDir(), name)
+	path = filepath.Join(os.TempDir(), randomFilename())
 
 	f, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_EXCL, 0o600)
 	if err != nil {
@@ -92,10 +87,14 @@ func WriteToTempFile(data []byte) (path string, cleanup func(), err error) {
 // randomFilename returns an unpredictable filename with a recognizable
 // prefix and extension, so a leaked or leftover file is identifiable as
 // belonging to this package during debugging.
-func randomFilename() (string, error) {
+//
+// It returns no error because it cannot fail. crypto/rand.Read "never
+// returns an error, and always fills b entirely"; on a failure it crashes
+// the program irrecoverably rather than reporting one. That has been its
+// contract since Go 1.24, and this module requires 1.26, so the error this
+// used to return was unreachable and every caller's check for it was dead.
+func randomFilename() string {
 	b := make([]byte, 16)
-	if _, err := rand.Read(b); err != nil {
-		return "", err
-	}
-	return fmt.Sprintf("kubevuln-vexdoc-%s.json", hex.EncodeToString(b)), nil
+	rand.Read(b) //nolint:errcheck // documented to never fail; it crashes instead
+	return fmt.Sprintf("kubevuln-vexdoc-%s.json", hex.EncodeToString(b))
 }
--- a/internal/vexdoc/vexdoc_test.go
+++ b/internal/vexdoc/vexdoc_test.go
@@ -2,7 +2,9 @@ package vexdoc
 
 import (
 	"os"
+	"path/filepath"
 	"runtime"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Verified the core claim: `crypto/rand.Read` has been documented since Go 1.24 to never return an error (it crashes the process instead via `io.ReadFull`/`fatal`), and `go.mod` here pins `go 1.26.0`, so the removed error branch was genuinely dead on every build this module can produce.

Checked the rest:
- `randomFilename`'s signature change is fully threaded through its one caller; no other callers exist.
- The new `TestWriteToTempFile_ReportsCreateFailure` covers the one failure `WriteToTempFile` can realistically hit (unusable temp dir), and asserts the contract that matters: no path returned, non-nil safe-to-call-twice cleanup, wrapped error.
- `TestWriteToTempFile_CreateFailureIsNotMistakenForAnEmptyDocument` correctly distinguishes the two error paths via `ErrEmptyDocument`.
- `redirectTempDir` verifies the env-var redirect actually took effect before relying on it, and skips otherwise — good cross-platform hygiene.
- All CI checks (build, lint, tests, CodeQL, DCO, GitGuardian) are green.

Small, well-scoped, well-tested. LGTM.

PR #843, opened 2026-08-21
fix: avoid VEX ignored match key collisions

### LINKED ISSUE #842: VEX ignored-match lookup can collide for distinct vulnerability/PURL pairs
### Overview 

The ``updateVEX`` path currently identifies ignored matches by concatenating the ``vulnerability ID`` and ``package PURL``. This allows distinct pairs such as ``A/BC`` and ``AB/C`` to produce the same key ``(ABC)``.

### Steps to reproduce
Create two ignored matches:
- Vulnerability A, PURL BC
- Vulnerability AB, PURL C
- Store the VEX document and trigger an updateVEX.
- Observe the ignored assessments after the update.

Both pairs resolve to the same concatenated key, so one assessment can overwrite the other.

### Findings

The lookup should identify a finding by the vulnerability ID and package PURL as separate fields, rather than relying on string concatenation.

This needs to be resolved to ensure ignored VEX assessments remain associated with the correct vulnerability/package during subsequent VEX updates.

jijo-OO7:
Fixes : #842

### Overview

The VEX update path keyed ignored matches by concatenating the ``vulnerability ID`` and ``package PURL``. This could cause distinct (vulnerability, PURL) pairs such as`` A/BC`` and ``AB/C`` to resolve to the same key ``(ABC)``, causing one ignored assessment to overwrite another.

This PR:

- Uses the existing ``vexStatementKey`` composite key for ignored-match lookups.
- Keeps vulnerability ID and package PURL independently identifiable.
- Adds a regression test covering the ambiguous-key case and verifying that each statement retains its own assessment.

This prevents ignored VEX assessments from being incorrectly associated with the wrong vulnerability/package during VEX updates.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where ignored vulnerability assessments could be incorrectly combined when vulnerability identifiers and package URLs formed identical combined values.
  * Preserved the correct justification and impact details for each assessment during VEX updates.

* **Tests**
  * Added coverage to verify that distinct ignored assessments remain separate.

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -1614,10 +1614,10 @@ func (a *APIServerStore) updateVEX(ctx context.Context, cve domain.CVEManifest, 
 
 	// ignoredMap drives the "reset every statement" pass below; guarded the same way as the
 	// Matches/IgnoredMatches loops above, since it reads the same cve.Content.
-	ignoredMap := make(map[string]ignoredVEXAssessment)
+	ignoredMap := make(map[vexStatementKey]ignoredVEXAssessment)
 	if cve.Content != nil {
 		for _, v := range cve.Content.IgnoredMatches {
-			ignoredMap[v.Vulnerability.ID+v.Artifact.PURL] = ignoredMatchAssessment(v)
+			ignoredMap[vexStatementKey{name: v.Vulnerability.ID, purl: v.Artifact.PURL}] = ignoredMatchAssessment(v)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed: build passes, full `repositories` test suite passes (including the new regression tests `TestBuildIgnoredVEXAssessmentMap_DistinguishesVulnerabilityAndPURL` and `TestAPIServerStore_storeVEX_ignoredMatchesDoNotCollide`). The fix correctly reuses the existing `vexStatementKey{name, purl}` composite key already used by `buildLocalVexStatementIndex`/`markRelevantVulnerabilitiesAsAffectedInVex`/`markIgnoredVulnerabilitiesInVex` for the identical purpose, so this is now consistent with the rest of the file rather than a one-off fix. No blockers.

PR #845, opened 2026-08-21
fix(adapters): pick the nearest fix version, never a downgrade

### LINKED ISSUE #844: bug: suggestedVersion() can suggest the wrong fix version, including a downgrade
## Problem

`suggestedVersion()` in `adapters/v1/domain_to_armo.go` is supposed to return the nearest version that fixes a vulnerability, but it picks whichever candidate happens to appear *first* in the unordered `versions` slice — not the smallest qualifying one. When no candidate is greater than the installed version, it falls through to `versions[0]` unconditionally, which can be a version **older** than what's already installed.

```go
// adapters/v1/domain_to_armo.go:284-301
func suggestedVersion(current string, versions []string) string {
	if len(versions) == 0 {
		return ""
	}
	// compare with semver
	// if current is not a version, return the first version
	if c, err := semver.NewVersion(current); err == nil {
		for _, version := range versions {
			v, err := semver.NewVersion(version)
			if err == nil {
				if c.LessThan(v) {
					return version
				}
			}
		}
	}
	return versions[0]
}
```

`versions` is `m.Vulnerability.Fix.Versions`, copied verbatim from Grype's match data in `adapters/v1/grype_to_domain.go:68`, which in turn reflects the upstream vulnerability feed (NVD/OSV/distro advisories). These feeds commonly list fix versions for multiple maintained branches (e.g. a CVE fixed in `5.4.2`, `5.10.1`, and `5.15.3` on three different LTS lines) with **no guaranteed ordering**.

## Reproduction / Evidence

Verified against the unmodified function with a standalone Go test:

```go
func TestSuggestedVersionBug(t *testing.T) {
	// Case 1: nearest fix is 14.16.0, but function returns 15.10.0
	// because it's first in the (realistic, unsorted) list.
	got := suggestedVersion("14.7.0", []string{"15.10.0", "10.24.0", "14.16.0"})
	t.Logf("unsorted-branch case: got %q", got) // got "15.10.0", want "14.16.0"

	// Case 2: no candidate is newer than current; function falls back to
	// versions[0], which is a downgrade.
	got2 := suggestedVersion("16.0.0", []string{"10.0.0", "12.0.0"})
	t.Logf("stale-branch-list case: got %q", got2) // got "10.0.0" — older than installed
}
```

Output:
```
unsorted-branch case: got "15.10.0"
stale-branch-list case: got "10.0.0" (this is a downgrade suggestion)
```

Both are deterministic given `Masterminds/semver/v3` comparison semantics — no flakiness, no mocking required.

The existing test in `domain_to_armo_test.go` ("Test with real versions", around line 286) only passes because its fixture happens to already be sorted ascending; it doesn't exercise realistic unsorted multi-branch fix lists, so this bug has no regression coverage.

## Root Cause

The loop returns the first slice element satisfying `current < candidate` instead of the *minimum* element satisfying that condition, and the fallback (`return versions[0]`) doesn't check that `versions[0]` is actually greater than `current` before returning it as "the fix."

## Impact

`hasKnownFix()` (`adapters/v1/fixstate.go:33`) wraps `suggestedVersion` and is used in two places:

- `ApplySecurityExceptions` / exception matching (`adapters/v1/securityexception.go`) — only the boolean `isFixed` is consumed there, so this bug does **not** affect suppression correctness.
- The backend vulnerability report builder (`adapters/v1/domain_to_armo.go`, around lines 128 and 157-162) — the returned version string is placed directly into `Fixes: []containerscan.FixedIn{{Version: version}}`, which is the remediation guidance surfaced to end users in the Kubescape/ARMO UI ("upgrade to this version to fix CVE-X").

So the impact is scoped but real: **users can be shown incorrect remediation guidance** — either a farther-than-necessary upgrade target, or in the worst case a version *older* than what they already have installed, for any CVE whose fix-versions list spans multiple maintained branches or isn't already sorted ascending in the source feed. This is a correctness bug in a user-facing security recommendation, not a crash or data-loss issue.

## Affected components

- `adapters/v1/domain_to_armo.go` (`suggestedVersion`)
- `adapters/v1/fixstate.go` (`hasKnownFix`, consumer)
- Indirectly, any backend consumer of `containerscan.FixedIn.Version`

## Not a duplicate / not by design

Searched issue and PR history for `suggestedVersion`, `fix version`, `hasKnownFix`, `downgrade`, `FixedIn` — no existing issue addresses fix-version *selection* correctness. The closest related history is #449 (`ExpiredOnFix` exceptions for CPE-matched CVEs) and PR #533 (shared `hasKnownFix` predicate across suppression paths) — both concern keeping the fixed/not-fixed *boolean* consistent between call sites, which is why `hasKnownFix` was unified in the first place. Neither touches which concrete version string is suggested. There's no maintainer comment anywhere suggesting "first match wins" selection is intentional, and the code comment ("if current is not a version, return the first version") only documents the non-semver fallback, not this behavior for the semver case.

## Acceptance Criteria

- `suggestedVersion` returns the **smallest** version in `versions` that is strictly greater than `current`, regardless of input order.
- If no candidate in `versions` is greater than `current`, the function does not silently suggest a downgrade — it should return `""` (or another explicit "no known upgrade fix" signal) rather than `versions[0]`.
- Table-driven tests cover: unsorted multi-branch lists, an all-lower-than-current list (the downgrade-fallback case), ties, and a mix of parseable/unparseable version strings.
- A regression test exists for `hasKnownFix` exercising an unsorted `Fix.Versions` slice.

## Suggested solution direction

Sort the semver-parseable candidates ascending (or track the running minimum while iterating) and return the first one greater than `current`; change the fallback to return `""` when nothing qualifies. This is a small, targeted change to `suggestedVersion` plus corresponding test coverage in `domain_to_armo_test.go` and `fixstate_test.go`.

bhuvan-somisetty:
## Problem

`suggestedVersion` (`adapters/v1/domain_to_armo.go`) returns the fix version shown to users as remediation guidance. It picked the *first* entry in `Fix.Versions` that was greater than the installed version, instead of the nearest one, and fell back to `versions[0]` unconditionally when nothing qualified — even suggesting a downgrade if every listed version was older than what's already installed.

`Fix.Versions` is copied verbatim from Grype's match data (`adapters/v1/grype_to_domain.go`), which reflects the upstream vulnerability feed. These feeds commonly list fix versions for multiple maintained branches (e.g. a CVE fixed in `5.4.2`, `5.10.1`, and `5.15.3` on three different LTS lines) in no particular order.

Fixes #844.

## Root cause

```go
if c, err := semver.NewVersion(current); err == nil {
    for _, version := range versions {
        v, err := semver.NewVersion(version)
        if err == nil {
            if c.LessThan(v) {
                return version // first match, not nearest
            }
        }
    }
}
return versions[0] // unconditional fallback, even if older than current
```

## Fix

Scan the full slice and track the minimum version strictly greater than `current`. If none qualifies, return `""` instead of `versions[0]`, so a downgrade is never suggested. The existing documented fallback — return `versions[0]` when `current` itself isn't a parseable semver — is unchanged.

## Testing

- Added table-driven cases to `Test_suggestedVersion` covering: an unsorted multi-branch list (must pick the nearest, not the first), an all-older-than-current list (must return `""`, not a downgrade), current equal to the only candidate, unparseable entries mixed with valid ones, and all-unparseable versions.
- Added regression cases to `TestHasKnownFix` exercising the same unsorted/all-older scenarios through the actual call path used by both the manifest and report paths.
- `go build ./adapters/...` — passes
- `go vet ./adapters/...` — clean
- `go test ./adapters/v1/...` — all passing, including the existing `TestExpiredOnFixAgreesAcrossPaths*` regression tests for #449, unaffected
- `golangci-lint run ./adapters/v1/...` — no new findings; the 21 pre-existing findings it reports are all outside this diff

## Impact

Scoped to remediation guidance shown to end users (the `Fixes[].Version` field surfaced in the Kubescape/ARMO UI). Exception-suppression logic (`hasKnownFix`'s boolean return, used by `ApplySecurityExceptions`) is unaffected — it only ever consumed the fixed/not-fixed boolean, never the version string.

## Summary by CodeRabbit

* **Bug Fixes**
  * Version recommendations now select the smallest valid upgrade, even when available versions are unordered.
  * Prevented downgrade suggestions when no newer version is available.
  * Invalid version entries are ignored while preserving fallback behavior for invalid current versions.

* **Tests**
  * Added coverage for unsorted, invalid, equal, and older version scenarios.

### DIFF
--- a/adapters/v1/domain_to_armo.go
+++ b/adapters/v1/domain_to_armo.go
@@ -281,23 +281,41 @@ func linkToVuln(id string) string {
 	}
 }
 
+// suggestedVersion returns the nearest version in versions that fixes current: the
+// smallest one strictly greater than current. versions is not guaranteed to be sorted
+// (it comes straight from the upstream vulnerability feed, which may list fix versions
+// for several maintained branches in any order), so the whole slice must be scanned
+// rather than trusting the first qualifying entry.
+//
+// If current is not a version, the first entry is returned, since there is nothing to
+// compare against. If current is a version but no entry in versions is greater than it,
+// "" is returned rather than falling back to versions[0]; versions[0] could be older
+// than current, which would suggest a downgrade.
 func suggestedVersion(current string, versions []string) string {
 	if len(versions) == 0 {
 		return ""
 	}
-	// compare with semver
-	// if current is not a version, return the first version
-	if c, err := semver.NewVersion(current); err == nil {
-		for _, version := range versions {
-			v, err := semver.NewVersion(version)
-			if err == nil {
-				if c.LessThan(v) {
-					return version
-				}
-			}
+	c, err := semver.NewVersion(current)
+	if err != nil {
+		return versions[0]
+	}
+
+	var nearest *semver.Version
+	var nearestStr string
+	for _, version := range versions {
+		v, err := semver.NewVersion(version)
+		if err != nil {
+			continue
+		}
+		if !c.LessThan(v) {
+			continue
+		}
+		if nearest == nil || v.LessThan(nearest) {
+			nearest = v
+			nearestStr = version
 		}
 	}
-	return versions[0]
+	return nearestStr
 }
 
 func parseLayersPayload(target source.ImageMetadata) (map[string]containerscan.ESLayer, error) {
--- a/adapters/v1/domain_to_armo_test.go
+++ b/adapters/v1/domain_to_armo_test.go
@@ -288,6 +288,43 @@ func Test_suggestedVersion(t *testing.T) {
 			versions: []string{"10.24.0", "12.21.0", "14.16.0", "15.10.0"},
 			want:     "14.16.0",
 		},
+		{
+			// versions is not guaranteed to be sorted; the fix listed first for an
+			// older branch must not win over the nearer fix listed later.
+			name:     "unsorted branch list picks the nearest fix, not the first entry",
+			current:  "14.7.0",
+			versions: []string{"15.10.0", "10.24.0", "14.16.0"},
+			want:     "14.16.0",
+		},
+		{
+			// every listed fix version is for a branch already superseded by current;
+			// there is no upgrade to suggest, so this must not fall back to versions[0]
+			// and suggest a downgrade.
+			name:     "no version above current returns empty rather than a downgrade",
+			current:  "16.0.0",
+			versions: []string{"10.0.0", "12.0.0"},
+			want:     "",
+		},
+		{
+			name:     "current equal to the only candidate returns empty, not a downgrade",
+			current:  "2.0.0",
+			versions: []string{"2.0.0"},
+			want:     "",
+		},
+		{
+			name:     "unparseable entries are skipped in favour of a valid nearer fix",
+			current:  "1.0.0",
+			versions: []string{"not-a-version", "3.0.0", "2.0.0"},
+			want:     "2.0.0",
+		},
+		{
+			// current parses, but nothing in versions does: there is no comparable
+			// candidate, so nothing is suggested.
+			name:     "current parses but every version is unparseable returns empty",
+			current:  "1.0.0",
+			versions: []string{"not-a-version", "also-not-a-version"},
+			want:     "",
+		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
--- a/adapters/v1/fixstate_test.go
+++ b/adapters/v1/fixstate_test.go
@@ -53,6 +53,27 @@ func TestHasKnownFix(t *testing.T) {
 			wantVersion: "2.17.0",
 		},
 		{
+			// #844: the fix-versions list from the upstream feed is not guaranteed to be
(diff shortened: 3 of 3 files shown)


matthyx (approved):
Verified the fix logic directly: `suggestedVersion` now scans the full `versions` slice and tracks the minimum parseable version strictly greater than `current`, returning `""` (never `versions[0]`) when nothing qualifies. Traced all new test cases by hand (unsorted multi-branch list, all-older-than-current, current == only candidate, unparseable entries mixed in, all-unparseable) — each matches the implementation. CI is fully green (build/vet/test/lint/CodeQL/DCO) and CodeRabbit's automated pass found no actionable issues.

One minor, non-blocking note: `hasKnownFix`'s doc comment (`adapters/v1/fixstate.go`) states the version is `""` "when no fix is known," but this change means `hasKnownFix` can now return `true, ""` when `Fix.Versions` is non-empty but nothing in it beats `current` — a fix is known, just not a newer one. Might be worth a one-line doc update so a future reader doesn't assume `version == ""` implies `fixed == false`. Not a blocker — approving.

PR #847, opened 2026-08-22
fix(services): stop ScanRegistry handing the CVE scanner a nil SBOM

### LINKED ISSUE #846: ScanRegistry hands the CVE scanner a nil SBOM when sbomGeneration is off, which the scanner dereferences
With `nodeSbomGeneration: true`, kubevuln generates no SBOM itself. `NewScanService` is given `!c.NodeSbomGeneration`, and `getOrCreateSBOM` returns immediately:

```go
func (s *ScanService) getOrCreateSBOM(ctx context.Context, workload domain.ScanCommand) (domain.SBOM, error, error) {
	if !s.sbomGeneration {
		return domain.SBOM{}, nil, nil
	}
```

A zero `domain.SBOM` and a **nil error**. Nothing about the error tells a caller the SBOM is missing; only `Content` being nil does.

`ScanCVE` checks for that:

```go
if !s.sbomGeneration && sbom.Content == nil {
	logger.L().Ctx(ctx).Warning("missing SBOM", helpers.String("imageSlug", workload.ImageSlug))
	return domain.ErrMissingSBOM
}
```

`ScanCP` checks for it too, per container. **`ScanRegistry` does not.** It goes from the `sbomErr != nil` branch straight to the `Incomplete || TooLarge` status check, which a zero SBOM passes (its `Status` is `""`), and then hands the zero SBOM to the CVE scanner.

The scanner dereferences it. From `GrypeAdapter.ScanSBOM`:

```go
s, err := domainToSyft(*sbom.Content)
```

So this is a nil pointer dereference. It happens on a `workerpool` goroutine, and `gammazero/workerpool` has no `recover()`, so it takes the process down rather than failing the one scan.

Reproduced against current main with a CVE scanner that records what it is handed:

```
--- FAIL: TestScanService_MissingSBOM_NoFlowScansANilSBOM/ScanRegistry
    Error:    Should be empty, but was [{nil map[] map[]    }]
    Messages: no SBOM may reach the CVE scanner when none was produced
[info] registry scan started. imageSlug: imageSlug
[info] registry scan complete. imageSlug: imageSlug
```

The recorded SBOM has a nil `Content`, and `ScanRegistry` returned no error at all: it logged "registry scan complete" for a scan that never had anything to scan. The same test against `ScanCVE` passes, which is the divergence.

`domain.ErrMissingSBOM` has no test anywhere in the repo, which is presumably why the gap was not noticed: the guard exists twice, is load-bearing in both places, and nothing exercises it.

Two notes on scope.

The panic needs an initialised Grype store; `ScanSBOM` returns `ErrInitVulnDB` before the dereference when the DB is not loaded. In production the DB is loaded, and `Ready()` gates scanning on it, so the dereference is the live path.

`nodeSbomGeneration` defaults to `false`, so a default deployment is unaffected. It is a documented setting, bound to the environment in `config.go` alongside `storage` and `riskAcceptance`, so this is reachable by configuration rather than only in theory.

Adding the guard to `ScanRegistry`, matching what the other two flows already do, and a test covering all of them so the three cannot drift again.

AnshulPatil2005:
## Overview

With `nodeSbomGeneration` on, `getOrCreateSBOM` returns a zero `domain.SBOM` and a **nil error**, so nothing about the error says the SBOM is missing and only `Content` being nil does.

`ScanCVE` checks for that. `ScanCP` checks for it per container. `ScanRegistry` did not, and handed the zero SBOM to the CVE scanner, which dereferences `Content`.

## Additional Information

`ScanRegistry` went from the `sbomErr != nil` branch straight to the `Incomplete || TooLarge` status check, which a zero SBOM passes since its `Status` is `""`, and then called `ScanSBOM`.

`GrypeAdapter.ScanSBOM` does:

```go
s, err := domainToSyft(*sbom.Content)
```

That is a nil pointer dereference, on a `workerpool` goroutine. `gammazero/workerpool` has no `recover()`, so it takes the process down rather than failing the one scan.

`domain.ErrMissingSBOM` had no test anywhere in the repo, which is presumably why the gap survived: the guard exists twice, is load-bearing in both places, and nothing exercised it.

Two notes on scope, both stated so review can weigh them rather than take my word for it:

The dereference needs an initialised Grype store, since `ScanSBOM` returns `ErrInitVulnDB` before reaching it when the DB is not loaded. In production the DB is loaded and `Ready()` gates scanning on it, so that is the live path.

`nodeSbomGeneration` defaults to `false`, so a default deployment is unaffected. It is a documented setting, bound to the environment in `config.go` alongside `storage` and `riskAcceptance`, so this is reachable by configuration rather than only in theory.

The returned error is a `*domain.ScanError` wrapping `ErrMissingSBOM`, rather than the bare sentinel `ScanCVE` returns. That matches `ScanRegistry`'s other failures and `ScanCP`'s handling of this same case, so the HTTP layer's existing reason labelling picks it up unchanged, and `errors.Is` still finds the sentinel through `ScanError.Unwrap`.

## How to Test

`go build ./... && go vet ./... && go test ./core/services/ -count=1`

`TestScanService_MissingSBOM_NoFlowScansANilSBOM` covers `ScanRegistry` and `ScanCVE` together against the same configuration, so the two cannot drift apart again. Reverting just the guard fails it against current main:

```
--- FAIL: TestScanService_MissingSBOM_NoFlowScansANilSBOM/ScanRegistry
    Error:    Should be empty, but was [{nil map[] map[]    }]
    Messages: no SBOM may reach the CVE scanner when none was produced
[info] registry scan started. imageSlug: imageSlug
[info] registry scan complete. imageSlug: imageSlug
```

Two things in that output. The recorded SBOM has a nil `Content`, and `ScanRegistry` returned no error at all: it logged a completed scan for one that never had anything to scan. The `ScanCVE` subtest passes unchanged either way, which is what makes it the control rather than just a second assertion.

`TestScanService_MissingSBOM_ScannerNeverSeesNilContent` states the consequence separately, asserting on what the scanner was handed rather than on what the flow returned, since that is the part that panics in production.

The tests use a recording `ports.CVEScanner` rather than the mock adapter on purpose: `MockCVEAdapter.ScanSBOM` never touches `sbom.Content`, so it would have returned a perfectly healthy manifest for an empty SBOM and hidden this.

## Related issues/PRs:

* Resolved #846

## Summary by CodeRabbit

* **Bug Fixes**
  * Scans now report a clear missing-SBOM error when SBOM generation is disabled and no SBOM content is available.
  * Prevented CVE scanning from proceeding with empty SBOM data, improving scan reliability and error handling.

* **Tests**
  * Added coverage to verify consistent errors and prevent invalid SBOM data from being submitted for vulnerability scanning.

### DIFF
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -591,6 +591,17 @@ func (s *ScanService) ScanRegistry(ctx context.Context) error {
 			return &domain.ScanError{Reason: reason, Err: sbomErr}
 		}
 
+		// With sbomGeneration off, getOrCreateSBOM returns a zero SBOM and a nil error, so
+		// nothing about err says the SBOM is missing and only Content does. ScanCVE and
+		// ScanCP each stop here for the same reason: the CVE scanner dereferences
+		// sbom.Content, so handing it a zero SBOM is a nil dereference, on a worker-pool
+		// goroutine that does not recover. See #846.
+		if !s.sbomGeneration && sbom.Content == nil {
+			logger.L().Ctx(ctx).Warning("missing SBOM",
+				helpers.String("imageSlug", workload.ImageSlug))
+			return &domain.ScanError{Reason: scanfailure.ReasonUnexpected, Err: domain.ErrMissingSBOM}
+		}
+
 		// do not process timed out SBOM
 		if sbom.Status == helpersv1.Incomplete || sbom.Status == helpersv1.TooLarge {
 			reason := classifySBOMStatusWithAnnotation(sbom.Status, sbom.Annotations)
--- a/core/services/scan_test.go
+++ b/core/services/scan_test.go
@@ -3332,3 +3332,102 @@ func TestScanService_SingleflightSBOMFailure_ReportsEveryWaiter(t *testing.T) {
 	assert.ElementsMatch(t, wantJobIDs, platform.jobIDs,
 		"each ReportScanFailure call must carry the reporting caller's own JobID, not just the leader's")
 }
+
+// recordingScanner records every SBOM handed to it, so a test can assert not just what a
+// scan flow returned but what it passed downstream on the way there.
+type recordingScanner struct {
+	ports.CVEScanner
+	got []domain.SBOM
+}
+
+func (r *recordingScanner) DBVersion(context.Context) string { return "v1.0.0" }
+func (r *recordingScanner) Ready(context.Context) bool       { return true }
+func (r *recordingScanner) Version() string                  { return "v1.0.0" }
+
+func (r *recordingScanner) ScanSBOM(_ context.Context, sbom domain.SBOM) (domain.CVEManifest, error) {
+	r.got = append(r.got, sbom)
+	return domain.CVEManifest{
+		Name:              sbom.Name,
+		CVEScannerVersion: "v1.0.0",
+		CVEDBVersion:      "v1.0.0",
+		Content:           &v1beta1.GrypeDocument{},
+	}, nil
+}
+
+// newScanServiceWithoutSBOMGeneration builds a service configured the way
+// nodeSbomGeneration: true configures it, where kubevuln generates no SBOM itself and
+// expects to find one already in storage.
+func newScanServiceWithoutSBOMGeneration(scanner ports.CVEScanner) (*ScanService, *repositories.MemoryStore) {
+	storage := repositories.NewMemoryStorage(false, false)
+	return NewScanService(
+		adapters.NewMockSBOMAdapter(false, false, false), storage,
+		scanner, storage,
+		adapters.NewMockPlatform(false, nil), adapters.NewMockRelevancyAdapter(),
+		false, // storage
+		false, // vexGeneration
+		false, // sbomGeneration -- the setting under test
+		false, false,
+	), storage
+}
+
+func registryWorkload() domain.ScanCommand {
+	return domain.ScanCommand{
+		ImageSlug:          "imageSlug",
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Reviewed the fix and reproduced it independently.

**Verified:**
- Reverting just the `scan.go` guard (keeping the new tests) reproduces the exact failure described in the PR body: `ScanRegistry` returns `nil` with a zero SBOM (`Content == nil`) instead of `ErrMissingSBOM`, confirming this was a real gap that would have reached `GrypeAdapter.ScanSBOM`'s `*sbom.Content` dereference.
- With the fix applied, `go build ./...`, `go vet ./...`, and `go test ./core/services/... -count=1` all pass; `gofmt -l` is clean.
- The new guard in `ScanRegistry` is consistent with the existing (unmodified) guards in `ScanCVE` and `ScanCP` — same condition (`!s.sbomGeneration && sbom.Content == nil`), and wraps `ErrMissingSBOM` in a `*domain.ScanError{Reason: scanfailure.ReasonUnexpected}` the same way `ScanCP` already does, so `errors.Is` still finds the sentinel through `ScanError.Unwrap`.
- CI is fully green (build, lint, tests, CodeQL, DCO, security/license scans).

CodeRabbit's one actionable comment (extend the shared test with a `ScanCP` case, and assert the concrete `*domain.ScanError`/`Reason` on `ScanRegistry` rather than just the sentinel) is a reasonable test-coverage enhancement but not a blocker — the `ScanCP` guard already predates this PR and is exercised elsewhere, and the sentinel-based assertion is consistent with how the sibling `ScanCVE` case is asserted in the same test.

No blockers. Approving.

PR #853, opened 2026-08-22
ci: enable go test -race on every PR

### LINKED ISSUE #852: CI never runs go test -race, because CGO_ENABLED is forced to 0 on every PR
# Description

Every PR to kubevuln runs its Go test suite with the race detector force-disabled, because `.github/workflows/pr-created.yaml` pins `CGO_ENABLED: 0` when it calls the shared `go-basic-tests.yaml` workflow. That workflow's `Environment-Test` job branches explicitly on this value:

```yaml
# kubescape/workflows/.github/workflows/go-basic-tests.yaml
- name: Test race conditions
  if: ${{ env.CGO_ENABLED == 1 }}
  run: go test -v -race $(go list ${{ inputs.UNIT_TESTS_PATH }} | grep -v /e2e)

- name: Test without race conditions
  if: ${{ env.CGO_ENABLED != 1 }}
  run: go test -v $(go list ${{ inputs.UNIT_TESTS_PATH }} | grep -v /e2e)
```

`go test -race` requires cgo (`go: -race requires cgo; enable cgo by setting CGO_ENABLED=1`), so `CGO_ENABLED: 0` isn't a neutral build setting here — for this specific job it's a hard, silent switch to the "without race conditions" branch. The `Basic-Test` job's coverage run (`go test -v ... -covermode=count -coverprofile=coverage.out`) is also always race-free. Every PR, on every commit, runs the whole suite without `-race`.

This is a genuine gap, not a theoretical one: `docs/DEVELOPMENT.md` itself tells contributors to do exactly what CI never does —

```
# docs/DEVELOPMENT.md:342-343
# Run test with race detector
go test -race ./...
```

— and kubevuln has an unusually long track record of real, hand-found data races that this recommendation exists to catch, none of which any CI run would have flagged:

| Issue/PR | What raced |
|---|---|
| #427 | `GrypeAdapter.Ready` held a write lock across 15-minute DB updates |
| #580 / #601 | Handler raced an abandoned Syft goroutine after `CreateSBOM` timeout |
| #690 | `pullMutex` released before the abandoned Syft goroutine actually finished |
| #628 / #629 | Singleflight SBOM key collision let concurrent scans of one tag at different digests merge |
| #733 / #734 | TOCTOU race between SecurityException cache invalidation and an in-flight `List()` |
| #753 | Unbounded registry-auth credential cache under concurrent access |
| #758 / #759 | Scan request `Submit()` raced worker-pool `Shutdown()`, panicking on a closed channel |
| #473 | Sidecar SBOM path serialized every scan behind one RPC-wide lock (partly a correctness workaround for missing synchronization elsewhere) |

Every one of these was found by manual code review or a hand-written concurrent test, months apart, in different packages (`adapters/v1`, `internal/registryauth`, `controllers`, `repositories`). That pattern — the same class of bug recurring across the whole codebase — is exactly what `-race` exists to catch automatically and continuously, and it has never once run.

## Root Cause

`.github/workflows/pr-created.yaml:25` passes `CGO_ENABLED: 0` into `kubescape/workflows/.github/workflows/incluster-comp-pr-created.yaml`, which forwards it unchanged into `go-basic-tests.yaml`'s `CGO_ENABLED` input (default `1`). That input drives the `if: ${{ env.CGO_ENABLED == 1 }}` / `!= 1` branch shown above, so `0` silently selects the race-free branch on every PR run (`opened`, `reopened`, `synchronize`, `ready_for_review`).

(`.github/workflows/pr-merged.yaml:61` also sets `CGO_ENABLED: 0`, but that input is declared and unused in `incluster-comp-pr-merged.yaml` — it never reaches a test step there, so it has no effect and isn't part of this bug. Worth cleaning up alongside the fix so it doesn't look load-bearing to the next reader.)

## Impact

- No PR — including every fix in the table above — was ever actually verified race-free by tooling; each was verified only by the human who wrote and ran a targeted concurrent test locally.
- A future change that reintroduces a data race (e.g. widening a critical section, dropping a mutex during a refactor, adding a new concurrent code path without one) has no automated check that would catch it before merge, even though this project's own history says that's the single most common way it has broken in production.
- `repositories/memory.go`'s `MemoryStore` — the in-memory store several of these same concurrent tests exercise (e.g. `core/services/scan_test.go`'s singleflight tests construct multiple goroutines against a shared instance) — has zero synchronization around its `aps`/`cveManifests`/`sboms` maps and `sbomStores` counter (verified: no `sync.Mutex`/`RWMutex` anywhere in the file). It is test-only in production (`cmd/http/main.go:74-76` only ever constructs `*APIServerStore`), so this isn't a production risk by itself, but it means turning `-race` on will very likely fail immediately in the test suite that was supposedly written to prove these bugs fixed — that failure needs to be fixed as part of enabling `-race`, not treated as a new regression.

## Affected Components

- `.github/workflows/pr-created.yaml` (the actual bug — this is the workflow that gates every PR)
- `.github/workflows/pr-merged.yaml` (dead `CGO_ENABLED: 0`, worth removing for clarity)
- `repositories/memory.go` (needs synchronization before `-race` can pass clean)
- Whatever else `-race` turns up once it actually runs — realistically the first PR enabling this will need to fix a small number of newly-surfaced races in test helpers/mocks, since none have ever been checked

## Why This Isn't Already Covered

Checked before filing:
- `gh issue list`/`gh pr list --search "race detector"` and `--search "CGO_ENABLED"`: no hits.
- PR #490 ("add arm64 test lane") is a different proposal — a whole extra QEMU-emulated `linux/arm64` test matrix job — and was closed by @matthyx specifically over its compute/time cost ("I'm against it, as it will significantly increase the compute and delay on a PR"). That objection is about adding a second, emulated architecture lane, not about enabling `-race` on the single existing native `amd64` job. Flipping `CGO_ENABLED` for `pr-created.yaml` costs roughly 1.5-2x the existing test job's time/memory on the same runner — real, but a different order of magnitude from QEMU emulation, and worth weighing explicitly against the recurring cost of hand-finding these bugs after the fact.

## Acceptance Criteria

- [ ] `pr-created.yaml`'s test job actually runs `go test -race` on every PR (verified via an Actions run log showing the "Test race conditions" step executed, not skipped).
- [ ] `repositories/memory.go` gains the synchronization needed for existing concurrent tests to pass under `-race` (or those tests are otherwise made race-safe), with no unrelated test disabled to get there.
- [ ] Any other race `-race` surfaces on first run is either fixed or, if a maintainer judges it a known/accepted case, explicitly suppressed with a comment explaining why — not silently skipped.
- [ ] `pr-merged.yaml`'s inert `CGO_ENABLED: 0` is removed or left with a comment noting it currently has no effect, so it doesn't mislead the next person into thinking it controls anything.

## Solution Direction

1. Change `.github/workflows/pr-created.yaml:25` from `CGO_ENABLED: 0` to `CGO_ENABLED: 1`. `actions/setup-go@v5` on `ubuntu-latest` ships a working C toolchain, so `-race` works with no other change to the shared workflow.
2. Run the suite locally with `-race` first to find what it flags, add a `sync.Mutex` (or equivalent) to `MemoryStore`, and fix anything else that turns up before flipping the CI flag — landing a red CI run helps no one.
3. Drop or comment the now-clearly-inert `CGO_ENABLED: 0` in `pr-merged.yaml`.
4. If CI time genuinely becomes a concern in review, the fallback is running `-race` as a separate, non-blocking job in parallel with the existing one rather than dropping it — but try the straightforward flip first, since the existing job's own runtime is the actual baseline to compare against, not the arm64-emulation cost that sank #490.

bhuvan-somisetty:
## Problem

Every PR to kubevuln runs its Go test suite with the race detector disabled, because `.github/workflows/pr-created.yaml` set `CGO_ENABLED: 0`. The shared `go-basic-tests.yaml` workflow only runs `go test -race` when `CGO_ENABLED == 1` (`-race` requires cgo); `0` silently selects the race-free branch instead. `docs/DEVELOPMENT.md` itself tells contributors to run `go test -race ./...` locally, but CI never has.

This isn't a theoretical gap: this repo has a long track record of real data races found only by hand, months apart, in different packages — #427, #580/#601, #690, #628/#629, #733/#734, #753, #758/#759. That's exactly the class of bug `-race` exists to catch continuously.

## Root Cause

`.github/workflows/pr-created.yaml:25` passed `CGO_ENABLED: 0` into `kubescape/workflows/.github/workflows/incluster-comp-pr-created.yaml`, which forwards it to `go-basic-tests.yaml`. That workflow's `Environment-Test` job branches:

```yaml
- name: Test race conditions
  if: ${{ env.CGO_ENABLED == 1 }}
  run: go test -v -race ...
- name: Test without race conditions
  if: ${{ env.CGO_ENABLED != 1 }}
  run: go test -v ...
```

`pr-merged.yaml` also set `CGO_ENABLED: 0`, but that input is declared and never used by `incluster-comp-pr-merged.yaml` — it had zero effect there. Removed it so it doesn't look load-bearing to the next reader.

## Solution

- `pr-created.yaml`: `CGO_ENABLED: 0` → `1`, with a comment pointing at why.
- `pr-merged.yaml`: dropped the inert `CGO_ENABLED: 0`.
- Turning `-race` on locally immediately found a real, previously-invisible data race: `repositories/memory.go`'s `MemoryStore` had no synchronization around its maps (`aps`, `cveManifests`, `sboms`) and counters (`sbomStores`, `summaryStubs`). It's test-only in production (`cmd/http/main.go` only ever constructs `*APIServerStore`), but the singleflight and worker-pool concurrency tests in `core/services` already run several goroutines against one shared `MemoryStore` — exactly the scenario `-race` is meant to catch. Added a `sync.RWMutex` guarding every field, with `RLock`/`Lock` on each accessor.

## Testing

- Added `TestMemoryStore_ConcurrentAccessIsRaceFree` (`repositories/memory_test.go`): runs 20 goroutines against a shared `MemoryStore`, calling every exported method with a mix of distinct and overlapping keys. It doesn't assert on final values (concurrent writes to overlapping keys are inherently unordered) — the point is that it's race- and panic-free under `-race`, which is what actually catches a regression of this fix.
- `go build ./...` — clean.
- `go vet ./...` — clean.
- `golangci-lint run --new-from-rev=upstream/main ./repositories/...` — 0 issues.
- `go test ./...` (excluding e2e) — all green.
- `go test -race ./repositories/...` — clean, including the new concurrent test (confirmed both before and after the fix: without the mutex it fails under `-race`; with it, it passes).
- `go test -race` across the rest of the suite (`adapters`, `controllers`, `core/domain`, `internal/*`, `pkg/sbomscanner/v1`, `cmd/*`, `config`) — all clean.

## Impact

Every future PR will actually run `go test -race`, matching what `docs/DEVELOPMENT.md` already tells contributors to do locally. Given this project's history, that's the single check most likely to catch the next concurrency regression before merge instead of after.

Fixes #852

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when multiple operations access stored data concurrently.
  * Reduced the risk of inconsistent results or data races during simultaneous updates and reads.

* **Tests**
  * Added concurrent-access coverage to validate consistent behavior under high parallel activity.
  * Enabled race detection in pull request testing to catch synchronization issues earlier.

### DIFF
--- a/.github/workflows/pr-created.yaml
+++ b/.github/workflows/pr-created.yaml
@@ -22,6 +22,8 @@ jobs:
       security-events: write
     uses: kubescape/workflows/.github/workflows/incluster-comp-pr-created.yaml@main
     with:
-      CGO_ENABLED: 0
+      # go-basic-tests.yaml only runs `go test -race` when CGO_ENABLED is 1 (-race requires
+      # cgo); 0 silently selected the race-free branch on every PR. See #852.
+      CGO_ENABLED: 1
       GO_VERSION: "1.26"
     secrets: inherit
--- a/.github/workflows/pr-merged.yaml
+++ b/.github/workflows/pr-merged.yaml
@@ -58,7 +58,6 @@ jobs:
       IMAGE_NAME: quay.io/${{ github.repository_owner }}/kubevuln
       IMAGE_TAG: v0.3.${{ needs.reset-run-number.outputs.run-number }}
       COMPONENT_NAME: kubevuln
-      CGO_ENABLED: 0
       GO111MODULE: "on"
       BUILD_PLATFORM: linux/amd64,linux/arm64
       GO_VERSION: "1.26"
--- a/repositories/memory.go
+++ b/repositories/memory.go
@@ -2,6 +2,7 @@ package repositories
 
 import (
 	"context"
+	"sync"
 
 	"github.com/kubescape/kubevuln/core/domain"
 	"github.com/kubescape/kubevuln/core/ports"
@@ -26,8 +27,15 @@ type sbomID struct {
 	SBOMCreatorVersion string
 }
 
-// MemoryStore implements both CVERepository and SBOMRepository with in-memory storage (maps) to be used for tests
+// MemoryStore implements both CVERepository and SBOMRepository with in-memory storage (maps) to be used for tests.
+//
+// mu guards every field below it: production code never runs more than one scan concurrently
+// against the same MemoryStore instance (only APIServerStore is wired up outside of tests), but
+// the test suite deliberately does -- singleflight and worker-pool concurrency tests construct
+// several goroutines against one shared MemoryStore -- so unsynchronized map/slice access here
+// is a real data race under `go test -race`, not a hypothetical one.
 type MemoryStore struct {
+	mu           sync.RWMutex
 	aps          map[apID]v1beta1.ContainerProfile
 	cveManifests map[cveID]domain.CVEManifest
 	sboms        map[sbomID]domain.SBOM
@@ -46,6 +54,8 @@ var _ ports.SBOMRepository = (*MemoryStore)(nil)
 // SBOMStores reports how many times StoreSBOM was called, so a test can tell an SBOM that
 // was written from one that was only read back.
 func (m *MemoryStore) SBOMStores() int {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
 	return m.sbomStores
 }
 
@@ -72,6 +82,8 @@ func (m *MemoryStore) GetContainerProfile(ctx context.Context, namespace string,
 		Namespace: namespace,
 		Name:      name,
 	}
+	m.mu.RLock()
+	defer m.mu.RUnlock()
 	if value, ok := m.aps[id]; ok {
 		return value, nil
 	}
@@ -90,6 +102,8 @@ func (m *MemoryStore) StoreContainerProfile(ctx context.Context, ap v1beta1.Cont
 		Namespace: ap.Namespace,
 		Name:      ap.Name,
 	}
+	m.mu.Lock()
+	defer m.mu.Unlock()
 	m.aps[id] = ap
 	return nil
 }
@@ -109,6 +123,8 @@ func (m *MemoryStore) GetCVE(ctx context.Context, name, SBOMCreatorVersion, CVES
 		CVEScannerVersion:  CVEScannerVersion,
 		CVEDBVersion:       CVEDBVersion,
 	}
+	m.mu.RLock()
+	defer m.mu.RUnlock()
 	if value, ok := m.cveManifests[id]; ok {
 		return value, nil
 	}
@@ -137,6 +153,8 @@ func (m *MemoryStore) StoreCVE(ctx context.Context, cve domain.CVEManifest, _ bo
 		CVEScannerVersion:  cve.CVEScannerVersion,
 		CVEDBVersion:       cve.CVEDBVersion,
 	}
+	m.mu.Lock()
+	defer m.mu.Unlock()
 	m.cveManifests[id] = cve
 	return nil
 }
@@ -156,6 +174,9 @@ func (m *MemoryStore) StoreCVESummary(ctx context.Context, cve domain.CVEManifes
 		CVEScannerVersion:  cve.CVEScannerVersion,
 		CVEDBVersion:       cve.CVEDBVersion,
 	}
+
+	m.mu.Lock()
+	defer m.mu.Unlock()
 
 	if withRelevancy {
 		idSumm := cveID{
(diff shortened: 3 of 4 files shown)


matthyx:
@bhuvan-somisetty we don't want to use CGO, so I would instead change the condition to stop checking CGO_ENABLED for the race checker.

bhuvan-somisetty:
Good call, makes sense — didn't mean to drag the whole job onto cgo just to unlock the race step. The real fix belongs in kubescape/workflows' go-basic-tests.yaml condition, not in kubevuln's own CGO_ENABLED value. I'll look at fixing it there instead. Thanks for the pointer!

Issue #854, opened 2026-08-22
govulncheck (absent from CI) shows both SBOM paths reach an unfixed RAR-decoder DoS (CVE-2025-11579)


bhuvan-somisetty:
# Description

`govulncheck` (not run anywhere in this repo's CI) shows kubevuln's own SBOM-generation code reaching a known, unfixed denial-of-service vulnerability in `github.com/nwaples/rardecode` — the RAR-archive decoder syft uses when cataloging a container image's filesystem. Since kubevuln's entire job is to catalog and scan container images pulled from arbitrary registries, image content is attacker-influenceable input, and this is a real, currently-uncaught reachable vulnerability in that input path — not a theoretical transitive-dependency concern.

There is a second, independent problem this surfaces: the *fixed* major version of the same library (`rardecode/v2`) is already present in kubevuln's own dependency graph — for an unrelated purpose — while the vulnerable v1 keeps being the one actually called on scanned image content. Nothing in this repo would ever notice that gap, because no vulnerability scan runs in CI.

# Evidence

```
$ govulncheck ./...
Vulnerability #22: GO-2025-4020 (CVE-2025-11579)
    DoS risk due to unrestricted RAR dictionary sizes in github.com/nwaples/rardecode
  Module: github.com/nwaples/rardecode
    Found in: github.com/nwaples/rardecode@v1.1.3
    Fixed in: N/A                      <- no fix for v1; fix only exists in v2
  Example traces found:
    #1: pkg/sbomscanner/v1/server.go:150:25: sbomscanner.CreateSBOM calls syft.GetSource,
        which eventually calls rardecode.NewReader
    ...
```

`adapters/v1/syft.go:230` (the default, in-process SBOM path) makes the identical `syft.GetSource(...)` call as `pkg/sbomscanner/v1/server.go:150` (the sidecar path) — confirmed by grep, not just the one trace govulncheck happened to print — so **both** SBOM-generation modes described in `docs/ARCHITECTURE.md` reach the vulnerable code.

Tracing the module graph confirms exactly why the already-present fixed version doesn't help:

```
$ go mod why github.com/nwaples/rardecode        # the vulnerable one
github.com/kubescape/kubevuln/adapters/v1
github.com/anchore/syft/syft
github.com/anchore/syft/internal/task
github.com/anchore/archiver/v3
github.com/nwaples/rardecode

$ go mod why github.com/nwaples/rardecode/v2     # the fixed one
[...]

PR #855, opened 2026-08-22
ci: add govulncheck vulnerability scan

bhuvan-somisetty:
## Problem

Nothing in this repo's CI runs a known-vulnerability scan against kubevuln's own compiled call graph. `govulncheck` shows a reachable, currently unfixed DoS in `github.com/nwaples/rardecode` v1 (CVE-2025-11579 / GO-2025-4020), pulled in transitively via `anchore/syft`'s RAR cataloger and reachable from both SBOM-generation paths (`adapters/v1/syft.go`, the default in-process path, and `pkg/sbomscanner/v1/server.go`, the sidecar path). Since kubevuln scans arbitrary container images, this is attacker-influenceable input, not a theoretical transitive-dependency concern. See #854 for the full trace and analysis.

## Root cause

`nwaples/rardecode` v1 has no upstream fix available yet (only `rardecode/v2` is fixed), so this can't be resolved with a dependency bump today — it depends on `anchore/archiver`/`anchore/syft` migrating off v1. The gap this PR closes is different: **nothing would have caught this, because no vulnerability scan runs anywhere in CI.**

## Solution

Add a `govulncheck` GitHub Actions job that:
- runs on push to `main` and on a weekly schedule (Monday 06:00 UTC), plus `workflow_dispatch`
- is deliberately **not** a `pull_request` check, since a govulncheck result can flip purely because the upstream advisory database changed, independent of any code change in a given PR — gating merges on it would block unrelated PRs on findings they didn't introduce and can't fix themselves
- installs a pinned `govulncheck` version and runs the new `make govulncheck` target

Also:
- adds `make govulncheck` (kept separate from `make verify`, since it queries `vuln.go.dev` and shouldn't be part of the fast local loop)
- documents the target in `docs/DEVELOPMENT.md` alongside the other slower/network-backed checks

This job is expected to start **red**, reporting the known CVE-2025-11579 finding tracked in #854, until the upstream fix is available. That's intentional — it makes the existing, previously-invisible gap visible instead of hiding it, and gives a place (Actions tab / notifications) where it stops being invisible again if a *new* finding shows up.

## Testing

- Ran `govulncheck ./...` locally against current `go.mod`/`go.sum` — reproduces the CVE-2025-11579 finding described in #854, confirming the workflow will surface it.
- Validated the workflow YAML parses correctly.
- Confirmed `go-version: "1.26"` and the `setup-go`/`checkout` action versions match the existing `lint.yaml` workflow's conventions.
- No Go source changes in this PR (`Makefile`, `docs/DEVELOPMENT.md`, and the new workflow file only), so `verify`/`lint`/existing test suites are unaffected.

## Impact

- Closes the CI-visibility gap called out in #854 (acceptance criteria: CI step added, regression check now exists so this can't silently regress again).
- Does **not** fix CVE-2025-11579 itself — that remains open pending an upstream fix in `anchore/archiver`/`anchore/syft`, still tracked in #854.

Refs #854

## Summary by CodeRabbit

* **Security**
  * Added automated vulnerability scanning for Go dependencies on code changes, weekly schedules, and manual runs.
  * Added a local command for developers to run vulnerability checks.

* **Documentation**
  * Documented vulnerability scanning setup and how it runs in continuous integration.

### DIFF
--- a/.github/workflows/govulncheck.yaml
+++ b/.github/workflows/govulncheck.yaml
@@ -20,6 +20,8 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
+        with:
+          persist-credentials: false
 
       - name: Set up Go
         uses: actions/setup-go@v5
--- a/.github/workflows/govulncheck.yaml
+++ b/.github/workflows/govulncheck.yaml
@@ -1,0 +1,44 @@
+name: govulncheck
+
+on:
+  # Every push to main, so a newly-reachable vulnerability is surfaced as soon
+  # as it lands rather than waiting for the weekly schedule below.
+  push:
+    branches: [ "main" ]
+  # Also catches a vulnerability that only becomes reachable because the
+  # upstream advisory database changed, not because kubevuln's own code did.
+  schedule:
+    - cron: '0 6 * * 1'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  govulncheck:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: "1.26"
+          cache: true
+
+      - name: Install govulncheck
+        run: go install golang.org/x/vuln/cmd/govulncheck@v1.7.0
+
+      # Deliberately not a pull_request check: govulncheck's result can change
+      # with the upstream advisory database alone, independent of any code
+      # change in a given PR, so gating merges on it would block PRs over
+      # findings they didn't introduce and can't fix by themselves. Running it
+      # here means a real finding fails this workflow and is visible in the
+      # Actions tab / notifications instead of silently passing.
+      #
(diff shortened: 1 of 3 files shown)


bhuvan-somisetty:
Thanks for the review! Pushed a follow-up commit addressing all three points: switched `govulncheck` to binary mode so it scans what's actually built instead of source, set `persist-credentials: false` on the checkout step since this workflow never needs to push, and pinned the documented local install to v1.7.0 to match CI. All checks are green now.

matthyx (approved):
Reviewed. Workflow conventions (checkout@v4, setup-go@v5, go 1.26, `PATH=$(go env GOPATH)/bin:$PATH make <target>`) match `lint.yaml`. All three CodeRabbit findings from the first pass (checkout persist-credentials, source- vs binary-mode scan, docs version pin) are addressed in the second commit — govulncheck now scans the compiled binary via `make govulncheck`, checkout sets `persist-credentials: false`, and docs pin `@v1.7.0` to match CI. Deliberately excluding pull_request as a trigger is well-reasoned (advisory-DB churn shouldn't gate unrelated PRs) and documented in code comments. No Go source changed; all CI checks are green. LGTM.

PR #857, opened 2026-08-23
fix: reject a scan jobID that already has an active scan in progress

### LINKED ISSUE #856: bug: scanStatusStore silently overwrites an in-flight job's status on jobID reuse, causing a finished job's real outcome to be dropped
## Description

`controllers/scan_status.go`'s `scanStatusStore.recordAccepted` unconditionally overwrites the status record for a `jobID`, with no check for whether that ID already belongs to a job that is still queued or running. All four scan handlers (`GenerateSBOM`, `ScanCP`, `ScanCVE`, `ScanRegistry`) call it the same unconditional way, immediately before writing the `200 OK` response the caller uses to start polling `GET /v1/scanStatus/:jobID`.

`docs/API.md` documents `jobID` as a caller-supplied **"Unique job identifier"** (e.g. `docs/API.md:280`), but nothing in the code enforces or even checks that uniqueness. If the same `jobID` is submitted a second time while the first job is still in flight — a client retry after a dropped response, an at-least-once delivery from a queue, or simply a caller bug — the second admission silently resets the shared tracking record, and whichever of the two jobs reaches a terminal state *first* permanently claims that record. The job that finishes second has its real outcome silently discarded, and a client polling `/v1/scanStatus/:jobID` is told the wrong result with no error and no log signal it can act on.

## Evidence (file:line, `kubescape/kubevuln@0f9e48d`)

**Unconditional overwrite on every admission, all four endpoints:**
- `controllers/scan_status.go:109-125` (`recordAccepted`):
  ```go
  func (s *scanStatusStore) recordAccepted(jobID, endpoint string) {
      if jobID == "" {
          return
      }
      now := time.Now().UTC()
      s.mu.Lock()
      defer s.mu.Unlock()
      s.items[jobID] = domain.ScanStatus{   // <-- unconditional overwrite, no existing-entry check
          JobID:      jobID,
          Endpoint:   endpoint,
          State:      domain.ScanStateQueued,
          Phase:      string(domain.ScanStateQueued),
          AcceptedAt: now,
          UpdatedAt:  now,
      }
      s.evictLocked(now)
  }
  ```
- `controllers/http.go:304,421,487,571` — each handler calls `recordAccepted(newScan.JobID, ...)` and writes `200 OK` immediately after, with no check for an existing non-terminal record under that ID:
  ```go
  // controllers/http.go:300-306 (GenerateSBOM; ScanCP/ScanCVE/ScanRegistry are identical in shape)
  if !h.admitQueueSlot(c, ctx, "generateSBOM", details) {
      return
  }
  h.ensureStatuses().recordAccepted(newScan.JobID, "generateSBOM")
  _, _ = problem.Of(http.StatusOK).Append(details).WriteTo(c.Writer)
  ```

**Ownership is inferred purely from state, not from which request created the record:**
- `controllers/scan_status.go:127-149` (`markRunning`, invoked via `claimTrackedJob` from inside the worker-pool closure): only checks `status.State != domain.ScanStateQueued`. It has no concept of "which HTTP request's closure is allowed to claim this job" — whichever closure's goroutine happens to run first against a given `jobID` wins the claim, including a second job whose `recordAccepted` reset the record back to `Queued` out from under a still-running first job.
- `controllers/scan_status.go:206-236` (`markTerminal`, backing `markSucceeded`/`markFailed`/`markAbandoned`):
  ```go
  status, ok := s.items[jobID]
  if !ok {
      return
  }
  if isTerminal(status.State) {   // <-- first writer to reach a terminal state wins; second is silently dropped
      return
  }
  ...
  s.items[jobID] = status
  ```

**Reproduction sequence** (both requests use `jobID = "X"`, distinct images A and B; A's scan happens to take longer than B's):

1. `POST /v1/generateSBOM {jobID: "X", image: A}` → admitted, `recordAccepted("X")` → `{State: Queued}`. Response: `200 OK`.
2. A's worker closure starts: `markRunning("X")` succeeds (`Queued`→`Running`).
3. While A is still scanning, `POST /v1/generateSBOM {jobID: "X", image: B}` is submitted (retry, redelivery, or caller bug) → admitted again, `recordAccepted("X")` **overwrites** the record back to `{State: Queued}`, discarding A's `Running`/`StartedAt`. Response: `200 OK`.
4. B's worker closure starts: `markRunning("X")` succeeds again (state is `Queued` again) — the store now believes B is the sole owner of job `X`, while A is still actually running in the background against image A.
5. B finishes first (image B was smaller/faster) and calls `markSucceeded("X")` → record becomes `{State: Succeeded}`, reflecting B's outcome.
6. A finishes later and calls `markFailed("X", ...)` (or `markSucceeded`) → `markTerminal` sees `isTerminal(status.State) == true` already → returns immediately. **A's real outcome is silently dropped.**
7. A client that polls `GET /v1/scanStatus/X` — expecting the outcome of the scan it most recently or originally cared about — is told `Succeeded` (B's result) even in cases where the semantically relevant job (A) actually failed, with no indication anything was overwritten.

No existing test exercises this: `controllers/scan_status_test.go`'s `TestScanStatusStore_MarkRunningOnlyClaimsQueuedJobs` only covers a *single* logical job racing its own shutdown-abandon path, not a second `recordAccepted` call reusing the same `jobID` while the first job is still active.

## Root cause

`scanStatusStore` keys purely on the caller-supplied `jobID` string and treats every `recordAccepted` call as authoritative, with no admission-time check for an existing non-terminal entry under the same key, and no per-record identity (e.g. a generation counter or ownership token) to let `markRunning`/`markTerminal` detect that they're operating on a record that was reset by a different request after their own job claimed it.

## Impact

Any caller that relies on `GET /v1/scanStatus/:jobID` to learn a scan's real outcome (the endpoint `#635` added specifically for this) can be told the wrong result — a failed scan reported as succeeded, or vice versa — whenever a `jobID` is reused while the prior job under that ID is still queued or running. This requires no malicious input, only an ordinary retry/redelivery of a request that arrives before the first attempt's response was durably observed, which is a normal failure mode for any at-least-once delivery path (HTTP client timeout-and-retry, a message queue redelivering after a slow ack, etc.). The failure is silent: no error is returned, no distinguishing log line is emitted, and the store's own data does not preserve enough information after the fact to tell that a collision happened at all.

## Affected components

- `controllers/scan_status.go` (`recordAccepted`, `markRunning`, `markTerminal`)
- `controllers/http.go` (`GenerateSBOM`, `ScanCP`, `ScanCVE`, `ScanRegistry` — all four call `recordAccepted` unconditionally)
- `docs/API.md` (documents `jobID` as unique but the contract is unenforced)

## Acceptance criteria

- `recordAccepted` (or the four handlers, before calling it) detects a `jobID` that already has a non-terminal (`Queued`/`Running`) record and does not silently reset it. The re-submission is either rejected (a new `4xx` admission error, consistent with the existing `ErrQueueFull` pattern in `core/domain/scan.go`) or handled in a way that keeps both jobs' subsequent state transitions from writing into a single shared record — the acceptance test in either case should assert that the *first* job's own eventual `markSucceeded`/`markFailed` outcome is preserved and observable via `get(jobID)`, not silently dropped by a later job's terminal write.
- A regression test in `controllers/scan_status_test.go` reproduces the sequence above (two `recordAccepted` calls for the same `jobID` with the first job still non-terminal, second job's `markSucceeded` landing before the first job's `markFailed`) and asserts the first job's real terminal outcome is not lost.
- `recordAccepted` reusing a `jobID` that is already **terminal** (the common, benign case — a caller polling long after a finished job, or deliberately reusing an ID once its predecessor is done) continues to work exactly as it does today; this fix should only change behavior for the collision-while-active case.
- Existing tests (`controllers/scan_status_test.go`, `controllers/http_test.go`) continue to pass unchanged.
- `go build ./controllers/...`, `go vet ./controllers/...`, `go test ./controllers/...` pass.

## Suggested fix direction

Give `admitQueueSlot`/`recordAccepted` a way to reject admission when `jobID` already maps to a non-terminal record, mirroring how `tryAdmit`/`admitQueueSlot` already reject on `ErrQueueFull` (`controllers/http.go:160-170`, `core/domain/scan.go:39-42`): add a `domain.ErrDuplicateJobID`-style sentinel, check for an existing non-terminal entry under `s.mu.Lock()` inside `recordAccepted` (or a new `tryRecordAccepted` returning `(ok bool)`), and have the four handlers write the same `4xx`/`problem.Of` rejection response `admitQueueSlot` already uses for a full queue when that check fails — including releasing the admission slot reserved by `tryAdmit` before returning, the same way the existing rejection paths do.

bhuvan-somisetty:
Fixes #856

## Problem

`scanStatusStore.recordAccepted` (`controllers/scan_status.go`) unconditionally overwrote the status record for a `jobID` on every admission, with no check for whether that ID already belonged to a job still queued or running. `docs/API.md` documents `jobID` as a caller-supplied "Unique job identifier", but nothing enforced it.

If the same `jobID` was submitted a second time while the first job was still in flight — a client retry after a dropped response, at-least-once redelivery from a queue, or a caller bug — the second admission silently reset the shared tracking record. Both jobs' worker closures could then claim the same `jobID` via `markRunning`, and whichever reached a terminal state first permanently won the record; the other's real outcome was silently dropped. A client polling `GET /v1/scanStatus/:jobID` could be told the wrong result with no error and no log signal.

## Root cause

`recordAccepted` treated every call as authoritative and had no per-record identity to let `markRunning`/`markTerminal` detect that a record had been reset by a different request after their own job had already claimed it — see the full reproduction sequence in #856.

## Solution

- `recordAccepted` now rejects re-admission when an existing record for the same `jobID` is non-terminal (queued or running), leaving that record untouched. Reusing a `jobID` once its prior job has reached any terminal state (succeeded/failed/abandoned) is unaffected — that's the ordinary case of a caller reusing an ID once its predecessor is done.
- The four scan handlers (`GenerateSBOM`, `ScanCP`, `ScanCVE`, `ScanRegistry`) now go through a new `admitJob` helper that combines `admitQueueSlot` and `recordAccepted`, turning a rejected re-admission into a `409 Conflict` response (`domain.ErrDuplicateJobID`) and releasing the admission slot it reserved — the same pattern `admitQueueSlot` already uses for `ErrQueueFull`/503. This also removes the duplicated three-line admission block that was repeated across all four handlers.
- `recordRejection`'s `duplicate_job_id` reason and `docs/API.md`'s per-endpoint response tables and status-code summary are updated to document the new `409`.

## Testing

- `controllers/scan_status_test.go`: `TestScanStatusStore_RecordAcceptedRejectsActiveJobID` reproduces the collision at the store level — a queued job's own re-admission is rejected without touching its record, a running job's `StartedAt` survives an attempted collision, and the job's real `markFailed` outcome is preserved afterward. `TestScanStatusStore_RecordAcceptedAllowsReuseOnceTerminal` proves the fix is narrowly scoped: reuse after succeeded/failed/abandoned still works exactly as before.
- `controllers/http_test.go`: `TestHTTPController_GenerateSBOM_DuplicateJobIDRejected` covers the 409 response and confirms the admission slot isn't leaked. `TestHTTPController_DuplicateJobID_PreservesFirstJobOutcome` is an end-to-end reproduction using a real blocking scan (not simulated store state) — submits a job, waits for it to genuinely start running, submits a duplicate under the same `jobID` and asserts 409, then lets the first job finish and asserts `GET /v1/scanStatus/:jobID` reports its real (failed) outcome, uncorrupted. `TestHTTPController_MetricsEndpoint_RecordsDuplicateJobIDRejection` covers the new rejection-reason metric label.
- `go build ./...`, `go vet ./controllers/... ./core/domain/...` — pass.
- `go test ./controllers/... ./core/domain/... -v` — pass, all new and existing tests green, no failures.
- `golangci-lint` was started locally but did not finish in a reasonable time on this machine; CI will run the repo's standard lint job on this PR.

## Impact

Any caller relying on `GET /v1/scanStatus/:jobID` to learn a scan's real outcome can no longer be told the wrong result due to `jobID` reuse while the prior job under that ID is still active. The change is additive and narrowly scoped: it only rejects the specific case of reusing an ID that is still active, and does not change behavior for a fresh `jobID` or one reused after its predecessor has finished.

## Summary by CodeRabbit

- **Bug Fixes**
  - Duplicate job IDs for active scans are now rejected with an HTTP 409 Conflict response.
  - Rejected requests no longer consume queue capacity or alter the original scan’s status or outcome.
  - Job IDs can be reused after the previous scan reaches a terminal state.
  - Duplicate-job rejections are recorded separately in scan metrics.

- **Documentation**
  - Updated API documentation to describe 409 responses for duplicate active job IDs across scan endpoints.

### DIFF
--- a/controllers/http.go
+++ b/controllers/http.go
@@ -177,21 +177,41 @@ func (h *HTTPController) admitQueueSlot(c *gin.Context, ctx context.Context, end
 // active job -- it writes the appropriate rejection response itself, releases any queue slot
 // it reserved, and returns false; either way the caller must stop processing the request
 // without doing anything else.
+//
+// The isActive check runs before admitQueueSlot on purpose: without it, a jobID that
+// collides with the very job occupying the controller's last admission slot would be
+// rejected as queue-full (503) before recordAccepted ever got a chance to diagnose it as a
+// duplicate (409), since admission capacity would already be exhausted by the time
+// recordAccepted ran. It's a precedence check only -- recordAccepted's own atomic
+// check-and-write remains the actual guard against a genuine race between two admissions
+// for the same jobID, so a duplicate that slips past this read-only check is still caught
+// there.
 func (h *HTTPController) admitJob(c *gin.Context, ctx context.Context, endpoint, jobID string, details problem.Option) bool {
+	if h.ensureStatuses().isActive(jobID) {
+		h.rejectDuplicateJobID(c, ctx, endpoint, jobID, details)
+		return false
+	}
 	if !h.admitQueueSlot(c, ctx, endpoint, details) {
 		return false
 	}
 	if !h.ensureStatuses().recordAccepted(jobID, endpoint) {
 		h.release()
-		logger.L().Ctx(ctx).Warning("rejecting scan, jobID already has an active scan in progress",
-			helpers.String("endpoint", endpoint),
-			helpers.String("jobID", jobID))
-		h.recordRejection(ctx, endpoint, domain.ErrDuplicateJobID)
-		_, _ = problem.Of(validationStatusCode(domain.ErrDuplicateJobID)).Append(details).WriteTo(c.Writer)
+		h.rejectDuplicateJobID(c, ctx, endpoint, jobID, details)
 		return false
 	}
 	_, _ = problem.Of(http.StatusOK).Append(details).WriteTo(c.Writer)
 	return true
+}
+
+// rejectDuplicateJobID writes the 409/ErrDuplicateJobID rejection response for jobID and
+// records it, shared by admitJob's two rejection points (the isActive precedence check and
+// the recordAccepted race fallback).
+func (h *HTTPController) rejectDuplicateJobID(c *gin.Context, ctx context.Context, endpoint, jobID string, details problem.Option) {
+	logger.L().Ctx(ctx).Warning("rejecting scan, jobID already has an active scan in progress",
+		helpers.String("endpoint", endpoint),
+		helpers.String("jobID", jobID))
+	h.recordRejection(ctx, endpoint, domain.ErrDuplicateJobID)
+	_, _ = problem.Of(validationStatusCode(domain.ErrDuplicateJobID)).Append(details).WriteTo(c.Writer)
 }
 
 // submit hands task to the worker pool, unless Shutdown has already begun, in which case
--- a/controllers/http.go
+++ b/controllers/http.go
@@ -169,6 +169,31 @@ func (h *HTTPController) admitQueueSlot(c *gin.Context, ctx context.Context, end
 	return false
 }
 
+// admitJob reserves a queue slot for endpoint via admitQueueSlot and then records jobID as
+// newly accepted via recordAccepted, rejecting a jobID that already belongs to an active
+// (queued or running) job instead of silently resetting its tracking record out from under
+// it (see #856). On success it writes the 200 OK response the caller uses to start polling
+// ScanStatus and returns true. On failure -- the queue is full, or jobID collides with an
+// active job -- it writes the appropriate rejection response itself, releases any queue slot
+// it reserved, and returns false; either way the caller must stop processing the request
+// without doing anything else.
+func (h *HTTPController) admitJob(c *gin.Context, ctx context.Context, endpoint, jobID string, details problem.Option) bool {
+	if !h.admitQueueSlot(c, ctx, endpoint, details) {
+		return false
+	}
+	if !h.ensureStatuses().recordAccepted(jobID, endpoint) {
+		h.release()
+		logger.L().Ctx(ctx).Warning("rejecting scan, jobID already has an active scan in progress",
+			helpers.String("endpoint", endpoint),
+			helpers.String("jobID", jobID))
+		h.recordRejection(ctx, endpoint, domain.ErrDuplicateJobID)
+		_, _ = problem.Of(validationStatusCode(domain.ErrDuplicateJobID)).Append(details).WriteTo(c.Writer)
+		return false
+	}
+	_, _ = problem.Of(http.StatusOK).Append(details).WriteTo(c.Writer)
+	return true
+}
+
 // submit hands task to the worker pool, unless Shutdown has already begun, in which case
 // it returns false without touching the pool at all. Callers must treat false the same as
 // a lost race with shutdown: release the admission slot task's defer h.release() will now
@@ -240,9 +265,10 @@ func scanFailureReason(outcome string, err error) string {
 
 // recordRejection records a validation-time rejection for the given endpoint. reason is
 // "too_many_requests" for registry back-pressure (ErrTooManyRequests), "queue_full" for local
-// admission control (ErrQueueFull, see WithMaxQueueDepth), and "invalid_request" for every
-// other validation error, keeping the rejection-rate signal distinct from malformed-payload
-// noise while staying low cardinality.
+// admission control (ErrQueueFull, see WithMaxQueueDepth), "duplicate_job_id" for a jobID
+// that already belongs to an active job (ErrDuplicateJobID, see admitJob), and
+// "invalid_request" for every other validation error, keeping the rejection-rate signal
+// distinct from malformed-payload noise while staying low cardinality.
 func (h *HTTPController) recordRejection(ctx context.Context, endpoint string, err error) {
 	if h.metrics == nil {
 		return
@@ -253,6 +279,8 @@ func (h *HTTPController) recordRejection(ctx context.Context, endpoint string, e
 		reason = "too_many_requests"
 	case errors.Is(err, domain.ErrQueueFull):
 		reason = "queue_full"
+	case errors.Is(err, domain.ErrDuplicateJobID):
+		reason = "duplicate_job_id"
 	}
 	h.metrics.RejectCounter.Add(ctx, 1, metric.WithAttributes(
 		attribute.String("endpoint", endpoint),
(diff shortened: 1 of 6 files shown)


bhuvan-somisetty:
Thanks for the review! Both points were fair — fixed the ordering so a duplicate jobID gets reported as 409 even if it's also the one occupying the queue's last slot (added `scanStatusStore.isActive` as a cheap precedence check before `admitQueueSlot` runs), and bounded `maxQueueDepth` in the existing test so the pending-slot assertion actually means something. Added a dedicated test for the ordering case too. All green locally, pushed in baca4ed.

matthyx (approved):
Reviewed the full diff (`controllers/http.go`, `controllers/scan_status.go`, `core/domain/scan.go`, `docs/API.md`, and the two test files).

The fix is correctly scoped: `recordAccepted`'s mutex-protected check-and-write is the actual correctness guard against a `jobID` collision, and `isActive` is purely a read-only precedence check so a collision on the controller's last admission slot is reported as 409 rather than 503 — the ordering issue CodeRabbit flagged on the first commit is properly fixed in baca4ed, and the mutation-vs-check race for two concurrent first-time admissions of the same fresh `jobID` still resolves correctly (one wins via the mutex, the loser's reserved queue slot is released via `h.release()`).

Reuse after any terminal state (succeeded/failed/abandoned) is unaffected, matching the PR's stated scope. Test coverage is solid: store-level tests for both the rejection and terminal-reuse cases, an end-to-end test with a genuinely blocking scan that proves the first job's real outcome survives a duplicate submission, the queue-full-vs-duplicate ordering test, and a metrics-label test. All CI checks are green (tests, lint, CodeQL, GitGuardian, DCO).

No blockers found. Approving.

PR #864, opened 2026-08-24
fix(adapters): match a SecurityException pattern that names Docker Hub non-canonically

### LINKED ISSUE #863: a SecurityException pattern naming Docker Hub non-canonically matches nothing
`matchImages` normalizes the scanned image but compares patterns literally. `tools.ReferenceMatchForms` expands the image into its canonical forms:

```go
forms := tools.ReferenceMatchForms(image)   // ["nginx:1.25", "docker.io/library/nginx:1.25", "docker.io/library/nginx"]
```

Patterns get only the expansions `expandPatternForms` adds, and those cover the case where a pattern has *no* registry. A pattern that does name Docker Hub, but not in the canonical `docker.io/library/...` spelling, matches nothing at all.

All of these are valid ways to write the official nginx image, and Docker itself accepts every one of them. Measured against `image = "nginx:1.25"` on current main:

```
pattern="nginx:1.25"                           -> true
pattern="docker.io/library/nginx:1.25"         -> true
pattern="library/nginx:1.25"                   -> true
pattern="docker.io/nginx:1.25"                 -> false
pattern="index.docker.io/library/nginx:1.25"   -> false
pattern="docker.io/nginx"                      -> false
```

`reference.ParseNormalizedNamed` resolves all six to `docker.io/library/nginx`, so the image side already treats them as the same thing. Only the pattern side does not.

This is the mirror of #834. That one was about a registry-less pattern matching *too much*, every registry rather than Docker Hub. This is the opposite failure: a pattern that names Docker Hub explicitly, just not canonically, matches nothing.

Worth being clear about the direction: the failure here is a `SecurityException` that silently does not suppress. The user sees a finding they thought they had excepted, which is noise rather than a security hole, and the dangerous direction, over-suppression, is what #835 already closed. It is still a silent wrong answer to a rule someone deliberately wrote.

The fix is to give a pattern its canonical form too, when it is a concrete reference rather than a glob:

- a pattern containing a wildcard is not a parseable reference, so it fails to parse and keeps only its literal forms, leaving glob behaviour untouched
- `named.String()` preserves the tag and the digest and pins the registry the pattern named, so nothing widens: `docker.io/nginx:1.25` still does not match `myreg.io/nginx:1.25`, still does not match tag `1.24`, and a digest-pinned pattern stays pinned

AnshulPatil2005:
## Overview

`matchImages` normalizes the scanned image via `tools.ReferenceMatchForms`, but compares patterns literally. `expandPatternForms` covers patterns with no registry at all, so a pattern that *does* name Docker Hub, just not in the canonical `docker.io/library/...` spelling, matched nothing.

Measured against `image = "nginx:1.25"` on current main:

```
pattern="nginx:1.25"                           -> true
pattern="docker.io/library/nginx:1.25"         -> true
pattern="library/nginx:1.25"                   -> true
pattern="docker.io/nginx:1.25"                 -> false
pattern="index.docker.io/library/nginx:1.25"   -> false
pattern="docker.io/nginx"                      -> false
```

All six are valid ways to write the official nginx image, and `reference.ParseNormalizedNamed` resolves every one to `docker.io/library/nginx`. The image side already treats them as the same thing; only the pattern side did not.

## Additional Information

This is the mirror of #834. That one was a registry-less pattern matching *too much*, every registry rather than Docker Hub, and #835 closed it by pinning the fallback. This is the opposite failure: a pattern that names Docker Hub explicitly matches nothing.

Worth being clear about the direction, since it changes how urgent this is. The failure here is a `SecurityException` that silently does **not** suppress: the user sees a finding they thought they had excepted. That is noise, not a security hole, and the dangerous direction is the one #835 already closed. It is still a silent wrong answer to a rule someone wrote deliberately.

Patterns now also carry their canonical reference form, via a small `appendNormalizedPattern` helper alongside the existing expansions.

**Globs are untouched.** A pattern containing a wildcard is not a parseable reference, so `ParseNormalizedNamed` fails and the pattern keeps only its literal forms. `*/nginx:1.25` and `myreg.io/*:1.25` behave exactly as before.

**Nothing widens.** `named.String()` preserves the tag and the digest and pins the registry the pattern named. Three assertions cover that specifically:

- `docker.io/nginx:1.25` still does not match `myreg.io/nginx:1.25`
- `docker.io/nginx:1.24` still does not match tag `1.25`
- a digest-pinned pattern stays pinned to that digest

## How to Test

`go build ./... && go vet ./... && go test ./adapters/v1/ -count=1`

Nine cases added to `TestMatchImages`: four for the spellings that used to match nothing, three for the widening the normalization must not cause, and two confirming glob behaviour is unchanged.

Dropping the `appendNormalizedPattern` calls fails exactly the four positive cases and nothing else, which is what says the tests pin this change rather than the surrounding behaviour:

```
--- FAIL: TestMatchImages/docker.io_without_the_library_namespace_matches
--- FAIL: TestMatchImages/index.docker.io_matches
--- FAIL: TestMatchImages/docker.io_repository_without_namespace_matches_any_tag
--- FAIL: TestMatchImages/docker.io_without_namespace_matches_a_short_image_reference
```

The existing `TestMatchImages` cases, including the short-pattern ones #835 added, all pass unchanged.

## Related issues/PRs:

* Resolved #863

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved container image matching by recognizing equivalent Docker Hub references, including omitted namespaces and registry prefixes.
  * Preserved strict matching for registries, tags, and digests to prevent unintended matches.
  * Maintained existing wildcard behavior while supporting normalized image references.

### DIFF
--- a/adapters/v1/securityexception_match.go
+++ b/adapters/v1/securityexception_match.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	wlidpkg "github.com/armosec/utils-k8s-go/wlid"
+	"github.com/distribution/reference"
 	"github.com/kubescape/go-logger"
 	"github.com/kubescape/go-logger/helpers"
 	"github.com/kubescape/k8s-interface/k8sinterface"
@@ -166,7 +167,7 @@ func expandPatternForms(p string) []string {
 			return patterns
 		}
 		patterns = append(patterns, "docker.io/library/"+p)
-		return patterns
+		return appendNormalizedPattern(patterns, p)
 	}
 
 	firstSeg, _, _ := strings.Cut(p, "/")
@@ -175,7 +176,35 @@ func expandPatternForms(p string) []string {
 		patterns = append(patterns, "docker.io/"+p)
 	}
 
-	return patterns
+	return appendNormalizedPattern(patterns, p)
+}
+
+// appendNormalizedPattern adds p's canonical reference form to patterns, when p is a
+// concrete reference rather than a glob.
+//
+// tools.ReferenceMatchForms normalizes the scanned image but patterns were only ever
+// matched literally, so a pattern naming Docker Hub in any spelling other than the
+// canonical docker.io/library/... one silently matched nothing: "docker.io/nginx:1.25",
+// "index.docker.io/library/nginx:1.25" and "docker.io/nginx" are all valid ways to write
+// the official nginx image, and none of them matched it. The expansions above only cover
+// patterns with no registry at all, which is the opposite case.
+//
+// A pattern containing a wildcard is not a parseable reference, so it fails here and keeps
+// only the literal forms, leaving glob behaviour untouched. Nothing widens either: the
+// canonical form carries the same tag and digest as p and pins the registry p named, so a
+// pattern still only matches the registry it asked for (#834). See #863.
+func appendNormalizedPattern(patterns []string, p string) []string {
+	named, err := reference.ParseNormalizedNamed(p)
+	if err != nil {
+		return patterns
+	}
+	canonical := named.String()
+	for _, existing := range patterns {
+		if existing == canonical {
+			return patterns
+		}
+	}
+	return append(patterns, canonical)
 }
 
 // labelSelectorMatches evaluates a standard Kubernetes label selector against a
--- a/adapters/v1/securityexception_match_test.go
+++ b/adapters/v1/securityexception_match_test.go
@@ -23,6 +23,23 @@ func TestMatchImages(t *testing.T) {
 		want     bool
 	}{
 		{name: "no patterns matches everything", patterns: nil, image: "docker.io/library/nginx:1.25", want: true},
+
+		// #834 pinned a registry-less pattern to Docker Hub. The mirror of that: a pattern
+		// that does name Docker Hub, but not in the canonical docker.io/library/... spelling,
+		// used to match nothing at all, since only the image was normalized and patterns were
+		// compared literally. All four spellings below are valid ways to write official nginx.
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Traced through `expandPatternForms`/`appendNormalizedPattern` against `tools.ReferenceMatchForms` by hand for each new case (Docker Hub aliases, `index.docker.io`, bare-repo, short-form) and they all resolve to the same canonical form the image side already normalizes to, using the same `distribution/reference` parser both sides rely on. The three non-widening cases (cross-registry, wrong tag, digest pin) check out too: `named.String()` only ever adds/normalizes the domain, never drops the tag or digest the pattern named. Glob patterns fail `ParseNormalizedNamed` and correctly keep only their literal forms.

New tests are well-targeted — the PR description's claim that reverting `appendNormalizedPattern` breaks exactly the four new positive cases and nothing else is exactly what pins the fix to this behavior rather than something incidental. All CI checks (build, vet, test, lint, CodeQL) are green, and this is a self-contained, well-documented change.

One non-blocking nit: `github.com/distribution/reference` is currently only in the `// indirect` block of `go.mod` even though it's now imported directly — worth a `go mod tidy` pass to move it, but doesn't affect correctness or build.

Approving.

slashben:
@AnshulPatil2005 , regardless to this PR, can I assign the documentation ticket to on SecurityExceptions ?
https://github.com/kubescape/kubescape.io/issues/144

AnshulPatil2005:
> @AnshulPatil2005 , regardless to this PR, can I assign the documentation ticket to on SecurityExceptions ?
> https://github.com/kubescape/kubescape.io/issues/144

yes please, happy to take it.

i'll work from the design doc and the CRD types and aim for a reference section plus a short user guide with a couple of worked examples. will open a PR on kubescape.io.

PR #865, opened 2026-08-24
chore: clear the standing lint findings so make lint-all exits clean

### LINKED ISSUE #850: make lint-all has ten standing findings, because CI only ever lints the diff
CI lints only what a PR changed. From `.github/workflows/lint.yaml` via the Makefile:

```make
lint:
	... golangci-lint run --timeout=15m --new-from-rev="$$base"; \

lint-all:
	golangci-lint run --timeout=15m
```

`--new-from-rev` was the right call for #566/#567, and this is the known cost of it: nothing ever runs the full tree, so anything that predates the diff stays invisible. `lint-all` exists in the Makefile for exactly this and no workflow invokes it.

Running it with the same pinned version CI uses, `golangci-lint@v1.64.8`:

```
pkg/sbomscanner/v1/integration_test.go:31:14: Error return value of `srv.Serve` is not checked (errcheck)
pkg/sbomscanner/v1/server_test.go:79:14: Error return value of `srv.Serve` is not checked (errcheck)
pkg/sbomscanner/v1/server_test.go:464:11: Error return value of `w.Write` is not checked (errcheck)
pkg/sbomscanner/v1/server_test.go:483:11: Error return value of `w.Write` is not checked (errcheck)
pkg/sbomscanner/v1/server_test.go:487:11: Error return value of `w.Write` is not checked (errcheck)
adapters/v1/domain_to_armo.go:174:7: S1009: should omit nil check; len() for nil slices is defined as zero (gosimple)
adapters/v1/backend_utils_test.go:27:2: SA1019: "k8s.io/utils/pointer" is deprecated: Use functions in k8s.io/utils/ptr instead (staticcheck)
```

Seven, but that is the reported number, not the real one. golangci-lint's `max-same-issues` defaults to 3, so the repeated `w.Write` message is truncated: there are six of those, not three. Ten sites in total.

One is in production code:

```go
if m.Artifact.Locations == nil || len(m.Artifact.Locations) < 1 {
```

`len()` on a nil slice is defined as zero, so the nil half is redundant.

The rest are test-only and none of them are hiding a failure: `srv.Serve` returns `ErrServerStopped` on the graceful `Stop` these tests already do, and the `w.Write` calls are `httptest` handlers writing fixture bytes. They are still worth clearing, because seven-to-ten standing findings are what stops `lint-all` being usable as a signal at all: nobody can tell a new one from the existing noise.

`k8s.io/utils/pointer` is deprecated in favour of `k8s.io/utils/ptr`, which is already an indirect dependency.

Fixing all ten so `make lint-all` exits clean. That leaves the decision of whether to *enforce* it, a scheduled or non-blocking full-tree job, as a separate call for maintainers rather than something bundled into this.

AnshulPatil2005:
## Overview

CI lints only what a PR changed, via `--new-from-rev`. That was the right call for #566/#567, and this is its known cost: nothing ever runs the full tree, so anything predating the diff stays invisible. `lint-all` exists in the Makefile for exactly this and no workflow runs it.

Running it with the version CI pins, `golangci-lint@v1.64.8`, turned up ten sites. This clears them, so `make lint-all` exits 0.

## Additional Information

The command reports seven, not ten. `max-same-issues` defaults to 3, so the repeated `w.Write` message is truncated at three of six.

| linter | sites | where |
| --- | --- | --- |
| gosimple (S1009) | 1 | `adapters/v1/domain_to_armo.go` |
| staticcheck (SA1019) | 1 | `adapters/v1/backend_utils_test.go` |
| errcheck (`srv.Serve`) | 2 | `pkg/sbomscanner/v1/{server,integration}_test.go` |
| errcheck (`w.Write`) | 6 | `pkg/sbomscanner/v1/server_test.go` |

One is production code:

```go
if m.Artifact.Locations == nil || len(m.Artifact.Locations) < 1 {
```

`len()` on a nil slice is defined as zero, so the nil half is redundant. Behaviour is identical either way.

The rest are test-only, and none of them was hiding a failure, which is worth being explicit about rather than just silencing them:

- `srv.Serve` returns `ErrServerStopped` on the graceful `Stop` those tests already perform, so there is nothing to handle. Discarded explicitly, with the reason written down.
- The `w.Write` calls are `httptest` handlers writing fixture bytes to a recorder.

They are still worth clearing, because standing findings are what stops `lint-all` being usable as a signal at all: with seven in the output, nobody can tell a new one from the existing noise.

`k8s.io/utils/pointer` is deprecated in favour of `k8s.io/utils/ptr`, already an indirect dependency, so `pointer.Bool(x)` becomes `ptr.To(x)`.

Deliberately **not** included: a workflow that enforces the full tree. Whether that should be a scheduled job, a non-blocking one, or nothing at all is a maintainer call, and bundling it here would mix a cleanup with a policy change. This just makes that decision available.

## How to Test

`make lint-all` (or `golangci-lint run --timeout=15m`) now exits 0 with no output. Before this it printed the seven findings above.

`go build ./... && go vet ./... && go test ./adapters/v1/ ./pkg/sbomscanner/v1/ -count=1`

The only production change is the redundant nil check, which is covered by the existing `domain_to_armo` tests: the branch it guards still fires for an empty `Locations` slice, which is the case that matters, since that is what makes the dummy-layer fallback run.

Reopens the work from #851, which was closed before review. Rebased onto current `main` and re-verified there: `golangci-lint` still reports the same 7 findings on `main` at `aa98d6d`, and exits clean on this branch.

## Related issues/PRs:

* Resolved #850
* Supersedes #851

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of vulnerability matches with empty artifact-location data.

* **Chores**
  * Updated internal test utilities to use the current pointer API.
  * Improved test server and mock response error handling to satisfy linting requirements.

* **Tests**
  * Maintained existing test coverage and behavior while improving integration test reliability.

### DIFF
--- a/adapters/v1/backend_utils_test.go
+++ b/adapters/v1/backend_utils_test.go
@@ -24,7 +24,7 @@ import (
 	"github.com/kubescape/kubevuln/core/domain"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"k8s.io/utils/pointer"
+	"k8s.io/utils/ptr"
 )
 
 // linearExceptionLookup is the straightforward scan the index replaced, kept here so the
@@ -146,7 +146,7 @@ func TestGetCVEExceptionMatchCVENameFromList(t *testing.T) {
 					VulnerabilityPolicies: []armotypes.VulnerabilityPolicy{
 						{Name: "CVE-2021-1234"},
 					},
-					ExpiredOnFix: pointer.Bool(true),
+					ExpiredOnFix: ptr.To(true),
 				},
 				{
 					VulnerabilityPolicies: []armotypes.VulnerabilityPolicy{
@@ -159,7 +159,7 @@ func TestGetCVEExceptionMatchCVENameFromList(t *testing.T) {
 						{Name: "CVE-2021-1234"},
 						{Name: "CVE-2021-9012"},
 					},
-					ExpiredOnFix: pointer.Bool(true),
+					ExpiredOnFix: ptr.To(true),
 				},
 			},
 			CVEName: "CVE-2021-1234",
@@ -531,7 +531,7 @@ func Test_summarize(t *testing.T) {
 						IsLastScan:        1,
 						Layers:            []containerscan.ESLayer{{LayerHash: dummyLayer}},
 						Vulnerability: containerscan.Vulnerability{
-							IsRelevant: pointer.Bool(false),
+							IsRelevant: ptr.To(false),
 							ImageID:    imageHash,
 							ImageTag:   imageTag,
 							Severity:   "Negligible",
@@ -548,7 +548,7 @@ func Test_summarize(t *testing.T) {
 						IsLastScan:        1,
 						Layers:            []containerscan.ESLayer{{LayerHash: dummyLayer}},
 						Vulnerability: containerscan.Vulnerability{
-							IsRelevant: pointer.Bool(false),
+							IsRelevant: ptr.To(false),
 							ImageID:    imageHash,
 							ImageTag:   imageTag,
 							Severity:   "Medium",
@@ -566,7 +566,7 @@ func Test_summarize(t *testing.T) {
 						IsLastScan:        1,
 						Layers:            []containerscan.ESLayer{{LayerHash: dummyLayer}},
 						Vulnerability: containerscan.Vulnerability{
-							IsRelevant:  pointer.Bool(false),
+							IsRelevant:  ptr.To(false),
 							ImageID:     imageHash,
 							ImageTag:    imageTag,
 							Description: "code execution",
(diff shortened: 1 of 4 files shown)


matthyx (approved):
Reviewed the diff — all four changes are mechanical and behavior-preserving:

- `pointer.Bool(x)` → `ptr.To(x)`: same semantics, just the non-deprecated package.
- `m.Artifact.Locations == nil || len(...) < 1` → `len(...) == 0`: equivalent for slices, since `len(nil)` is `0`.
- `go srv.Serve(lis)` → wrapped with explicit `_ = srv.Serve(lis)` and a comment explaining `ErrServerStopped` is expected on graceful `Stop`: no functional change, just satisfies errcheck.
- `w.Write(...)` → `_, _ = w.Write(...)` in httptest fixture handlers: same, no functional change.

CI is green (lint, build, tests, CodeQL, GitGuardian, DCO) and CodeRabbit found no actionable comments. No blockers. LGTM.

Issue #453, opened 2026-08-04
RFC: Add typed `response` field to SecurityException


pjungermann:
## Overview

SecurityException today follows the OpenVEX standard (as its [design document](https://github.com/kubescape/kubevuln/blob/85c8d7a9e42dd58a629f6378e023a090e157df98/docs/security-exception-design.md#openvex-compatibility) states) but implements only three of OpenVEX's four status values: `not_affected`, `fixed`, and `under_investigation`. The `affected` status is missing, and there is no free-text field equivalent to OpenVEX's `action_statement`. As a result, SecurityException has no first-class way to express **risk-accepted** findings.

This RFC proposes three additive changes:

1. Add `affected` as a fourth valid `status` value, completing SecurityException's OpenVEX status vocabulary.
2. Add an optional `actionStatement` free-text field, parallel to the existing `impactStatement` and matching OpenVEX's `action_statement`.
3. Add an optional typed `response[]` field, aligned with CycloneDX VEX's `analysis.response[]`, so authors can encode "we know this vulnerability is real; we have accepted the risk under these conditions" with machine-parsable action intent.

The two new fields (`actionStatement`, `response[]`) are individually optional, but **at least one is required when `status = affected`**. This mirrors OpenVEX's rule that an affected statement MUST include a statement describing the action taken to remediate or mitigate the vulnerability, and parallels SecurityException's existing rule that `justification` is required when `status = not_affected`. Existing documents (which use only the current three status values) remain valid and behave the same way. Adding `affected` to the `status` enum is a vocabulary extension: consumers must recognize the new value. The OpenVEX degradation path is preserved.

## Format comparison

SecurityException and the three widely-used VEX formats each model overlapping concepts differently. This table summarizes the fields relevant to this RFC, with SecurityException's current shape (before) and the shape proposed here (after). **NEW** marks fields introduced by this RFC.

| Concept | OpenVEX | CycloneDX VEX | CSAF VEX | SecurityException (before) | SecurityException (after) |
|---|---|---|---|---|---|
| **Status** ("where does the vulnerability stand") | `status`: 4 values (`not_affected`, `affected`, `fixed`, `under_investigation`) | `analysis.state`: 6 values including `false_positive` | `product_status` groupings | `status`: 3 values (**`affected` missing**) | `status`: 4 values (**NEW** `affected`) |
[...]

PR #867, opened 2026-08-25
fix(adapters,repositories): recover actionStatement/response for affected SecurityExceptions

### LINKED ISSUE #866: affected-status SecurityException loses actionStatement/response: every risk-accepted CVE gets the same generic VEX action_statement
## Description

A `SecurityException`/`ClusterSecurityException` entry with `status: affected` can carry an author-written `actionStatement` (e.g. "WAF mitigation in place, ticket SEC-1234, reviewed by security lead") and a typed `response` (e.g. `will_not_fix`). These two fields are exactly what RFC #453 added specifically so a risk-accepted finding could survive audit scrutiny instead of being expressed through a stretched `not_affected` claim.

They never survive the trip from the CRD to any observable output. Every `affected` suppression — regardless of what its author actually wrote — ends up logged, stored, and exported identically, with the real justification discarded at the very first hop.

## Root Cause

`buildSuppressionAttributes` (`adapters/v1/securityexception.go:228-267`) does capture `actionStatement`/`response` into the in-process `policy.Attributes` map:

```go
if s := strings.TrimSpace(vuln.ActionStatement); s != "" {
    attrs["actionStatement"] = s
}
if len(vuln.Response) > 0 {
    ...
    attrs["response"] = responses
}
```

But `buildIgnoreRules` (`adapters/v1/securityexception.go:371-407`), which is what actually persists provenance onto the `VulnerabilityManifest`'s `IgnoreRule`, only copies `sourceKind`, `ruleId`, `sourceNamespace`, `status` (into `FixState`), `justification`, and `impactStatement`. `actionStatement` and `response` are silently dropped — never assigned to any field on the `v1beta1.IgnoreRule` that's returned.

`logSuppression` (`adapters/v1/securityexception.go:431-465`) has the identical gap: it logs `sourceName`, `reason`, `sourceKind`, `ruleId`, `sourceNamespace`, `justification`, `impactStatement`, `normalizedTarget` — never `actionStatement`/`response`. Confirmed by reading the full field list; no test in `securityexception_test.go` asserts either field reaches a log line or an `IgnoreRule` (the existing tests around line 1249-1250 only assert the values land in `policy.Attributes`, which is discarded after `ApplySecurityExceptions` returns).

`emitSuppressionEvent` (`adapters/v1/securityexception.go:472-492`) writes a generic `"Suppressed %s in package %s"` K8s Event with no VEX-specific text at all for any status, so it carries nothing here either.

The loss compounds downstream, in the code that builds the actual exported OpenVEX document consumers see (`repositories/apiserver.go:1196-1241`, `ignoredMatchAssessment`):

```go
case string(sev1beta1.VulnerabilityStatusAffected):
    assessment.status = v1beta1.Status(vex.StatusAffected)
    assessment.justification = ""
    assessment.impactStatement = ""
    assessment.actionStatement = securityExceptionAcceptedRiskAction   // <- hardcoded constant
    assessment.statusNotes = ignoredMatchStatusNotes(rule)
```

`securityExceptionAcceptedRiskAction` (line 1185) is the fixed string `"A SecurityException accepted this vulnerability as an affected finding"`. It's used here specifically because `rule` (the `IgnoreRule` read back from storage) never carries the real `actionStatement`/`response` — there is nothing else for this line to read. `ignoredMatchStatusNotes` (line 1268-1277) only reads `rule.Justification` / `rule.ImpactStatement`, which for a genuine `affected` entry are typically empty (an author who follows the schema uses `actionStatement`/`response` for `affected`, not `justification`/`impactStatement`, which are the `not_affected` fields per the OpenVEX-aligned validation rules RFC #453 documents).

Net effect: **every** `affected`/risk-accepted `SecurityException` across a cluster produces byte-identical VEX output, log lines, and stored `IgnoreRule`s, regardless of what the author wrote. The one field OpenVEX requires for an `affected` statement ("a VEX statement MUST include a statement that SHOULD describe actions to remediate or mitigate the vulnerability") is present in the CRD, is validated as present (`shouldSuppress`/`affectedSuppresses` in `securityexception.go:148-175` reject an entry with no `actionStatement`), and is then thrown away before it reaches anything a human or downstream tool can read.

## Reproduction

1. Create a `SecurityException` with one vulnerability entry:
   ```yaml
   status: affected
   actionStatement: "WAF mitigation in place, ticket SEC-1234"
   response: [will_not_fix]
   ```
2. Trigger a scan that matches it. The finding is suppressed (confirmed correct — `affectedSuppresses` requires both fields to be a non-remediating response before allowing suppression).
3. Inspect what's actually recorded:
   - `kubectl get vulnerabilitymanifest <name> -o yaml`: `ignoredMatches[].appliedIgnoreRules[].fix-state` reads `"affected"`; there is no field anywhere on the object holding `"WAF mitigation in place, ticket SEC-1234"` or `"will_not_fix"` — `IgnoreRule` (`github.com/kubescape/storage`'s `pkg/apis/softwarecomposition/v1beta1/grype_types.go:93-108`) has no such fields to begin with.
   - The exported OpenVEX statement for this CVE has `action_statement: "A SecurityException accepted this vulnerability as an affected finding"` — the same string every other `affected` exception on the cluster also gets, not the ticket-specific text the author wrote.
   - The log line for "CVE suppressed by security exception" has no `actionStatement`/`response` field.
   - No K8s Event carries either.

Two different `SecurityException` objects, both `status: affected`, one authored `actionStatement: "vendor patch validated in staging, deploying next window"` and the other `actionStatement: "compensating control: network policy blocks the affected port"`, produce indistinguishable output everywhere the suppression is recorded.

## Impact

- **Audit/compliance**: risk acceptance is precisely the category of suppression regulators and internal security review expect to be individually justified and traceable. RFC #453's own problem statement is that misusing `not_affected` for risk acceptance "publishes semantically stretched claims that may not survive audit scrutiny" — the fix for that (typed `response` + `actionStatement`) is implemented up to the CRD boundary and then discarded, so the audit trail this feature exists to provide doesn't actually reach any artifact a reviewer would look at.
- **VEX correctness**: the exported OpenVEX document — an artifact meant for external consumption, per `docs/VEX.md` and issue #403's own concern about "downstream consumers... compar[ing] internally generated and externally supplied VEX documents" — asserts a generic, non-specific action statement for every risk-accepted CVE, which is not an accurate representation of the actual documented remediation decision.
- **Operability**: `kubectl describe`/`get -o yaml` and logs are the primary way an operator reconstructs *why* a finding disappeared. For every other suppression path (`not_affected`/`fixed`, `justification`/`impactStatement`) that reconstruction works; for `affected`/risk-acceptance specifically, it silently doesn't, which is inconsistent and easy to miss since nothing errors.

## Affected Components

- `adapters/v1/securityexception.go`: `buildIgnoreRules`, `logSuppression`, `emitSuppressionEvent`
- `repositories/apiserver.go`: `ignoredMatchAssessment`, `ignoredMatchStatusNotes`

## Acceptance Criteria

- [ ] `actionStatement` and `response` from an `affected` `VulnerabilityException` entry are recoverable from the persisted `VulnerabilityManifest` (or an equally durable, inspectable place) after a scan, not just from in-memory `policy.Attributes`.
- [ ] The "CVE suppressed by security exception" log line includes `actionStatement`/`response` when present, matching the existing `justification`/`impactStatement` treatment.
- [ ] The exported OpenVEX `action_statement` for an `affected` SecurityException-driven statement reflects the author's actual `actionStatement` text (and ideally the `response` value) instead of the fixed `securityExceptionAcceptedRiskAction` string, falling back to that generic string only when the specific text is genuinely unavailable (e.g. a cache entry written before this fix).
- [ ] Regression tests cover: two distinct `affected` exceptions producing distinct log output, distinct persisted provenance, and distinct exported VEX `action_statement` text.
- [ ] `RestoreSuppressedMatches`/`isExceptionSourcedIgnore` (`adapters/v1/securityexception.go:507-556`) continue to work unchanged with whatever new field usage is introduced.

`v1beta1.IgnoreRule` (`github.com/kubescape/storage`) has no `ActionStatement`/`Response` fields, and adding them is a cross-repo schema change. A same-repo fix is available using a pattern this file already establishes for exactly this situation: `buildIgnoreRules` already repurposes the unrelated `FixState` field to carry the SecurityException status vocabulary instead of Grype's native fix-state vocabulary (see the comment at `securityexception.go:390-396`), specifically because every reader of `FixState` gates on `SourceKind` being a SecurityException first.

The same technique closes this gap without a storage-schema change: for an `affected` entry (which never legitimately populates `Justification`/`ImpactStatement`, since those are the `not_affected` fields), `buildIgnoreRules` can write `actionStatement` into `rule.ImpactStatement` and the joined `response` values into `rule.Justification`, gated the same way `FixState`'s repurposing already is (only for `SourceKind == "SecurityException"/"ClusterSecurityException"` and only when `status == "affected"`, where the fields are otherwise always empty). `ignoredMatchAssessment` and `ignoredMatchStatusNotes` then read the real text instead of the hardcoded constant, and `logSuppression` gets the same fields added to its log line directly from `policy.Attributes` (no repurposing needed there, since it never wrote through the storage type). A follow-up issue can track adding first-class fields to `kubescape/storage` if the repurposing approach turns out to be too indirect long-term.

Happy to open a PR for this.

bhuvan-somisetty:
## Problem

A `SecurityException`/`ClusterSecurityException` entry with `status: affected` can carry an author-written `actionStatement` and a typed `response[]` (e.g. `will_not_fix`) — the fields RFC #453 added specifically so a risk-accepted finding could survive audit scrutiny instead of being expressed through a stretched `not_affected` claim.

They never survived past the in-memory suppression decision. Every `affected` suppression — regardless of what its author actually wrote — was logged, stored, and exported identically, with the real justification discarded at the first hop.

## Root Cause

`buildSuppressionAttributes` (`adapters/v1/securityexception.go`) captures `actionStatement`/`response` into the transient `policy.Attributes` map, but:

- `buildIgnoreRules` — which persists provenance onto the manifest's `IgnoreRule` — only copied `sourceKind`, `ruleId`, `sourceNamespace`, `status` (into `FixState`), `justification`, and `impactStatement`. `actionStatement`/`response` were dropped.
- `logSuppression` had the identical gap.
- Downstream, `ignoredMatchAssessment` (`repositories/apiserver.go`), which builds the actual exported OpenVEX statement, had nothing to read back from the persisted `IgnoreRule` and fell back to a single hardcoded `action_statement` (`securityExceptionAcceptedRiskAction`) for **every** affected suppression on the cluster.

`IgnoreRule` (`github.com/kubescape/storage`) has no dedicated fields for `actionStatement`/`response`, and adding them is a cross-repo schema change.

## Solution

`buildIgnoreRules` now repurposes `Justification`/`ImpactStatement` for a `FixState=="affected"` rule, the same way `FixState` itself is already repurposed to carry the SecurityException status vocabulary instead of Grype's native one (see the existing comment on that repurposing). This is safe: those two fields are otherwise unused on an `affected` entry — they're the `not_affected` fields, and `shouldSuppress` requires `actionStatement`, never `justification`, for an `affected` entry to suppress at all. `ImpactStatement` carries the action statement text; `Justification` carries the comma-joined `response` values.

`ignoredMatchAssessment` reads the real text back through two small helpers, `affectedActionStatement`/`affectedStatusNotes`, falling back to the generic constant only when the field is genuinely empty (e.g. a manifest cached before this fix).

`logSuppression` gets `actionStatement`/`response` added directly to its log fields, alongside the existing `justification`/`impactStatement` logging.

No API, schema, or CRD change. The fix is entirely internal to how `adapters/v1` and `repositories` already move data between each other.

## Testing

- `adapters/v1`: extended `TestApplySecurityExceptions_PopulatesIgnoreRuleProvenance`'s sibling coverage with three new tests — the repurposed fields land on the persisted `IgnoreRule`, two distinct `affected` entries produce distinct (not collapsed) provenance, and a `not_affected`/`fixed` entry's `Justification`/`ImpactStatement` are unaffected by the new repurposing branch.
- `repositories`: updated `TestAPIServerStore_storeVEX_preservesSecurityExceptionSemantics`'s `CVE-AFFECTED` fixture/assertions to the new contract (previously asserted the *old*, buggy behavior — the generic constant and a mislabeled `justification: ...; impact: ...` note built from ad hoc rule text), and added `TestAPIServerStore_storeVEX_affectedFallsBackWithoutRepurposedProvenance` for the fallback path.
- Ran the full `adapters/v1` and `repositories` test suites (`go test ./adapters/v1/... ./repositories/...`) — all pass, including the log output, which now visibly carries `actionStatement`/`response` for an affected suppression.
- `go build` and `go vet` on both packages: clean.
- Confirmed via `grep` that `buildIgnoreRules`, `logSuppression`, `ignoredMatchAssessment`, `ignoredMatchStatusNotes`, and the two new helpers are unexported and have no callers outside these two files/packages, so the change is contained.
- I was not able to complete a full repo-wide `go build ./...` / `go vet ./...` in my local environment (disk space and I/O constraints stalled linking the `cmd/*` binaries, which pull in the full grype/syft dependency tree) — the two touched packages are the only ones affected and are fully verified above.

## Impact

- Audit/compliance: the actual risk-acceptance rationale (why a real, present vulnerability was suppressed) now reaches the exported OpenVEX document, the persisted `VulnerabilityManifest`, and the logs — not a generic placeholder shared by every `affected` exception on the cluster.
- No behavior change for `not_affected`/`fixed` suppressions, or for any manifest already cached before this fix (falls back to the previous generic text).

Fixes #866.

### DIFF
--- a/adapters/v1/securityexception.go
+++ b/adapters/v1/securityexception.go
@@ -387,7 +387,8 @@ func buildIgnoreRules(m v1beta1.Match, matched []armotypes.VulnerabilityExceptio
 		if ns, ok := p.Attributes["sourceNamespace"].(string); ok {
 			rule.SourceNamespace = ns
 		}
-		if status, ok := p.Attributes["status"].(string); ok {
+		status, _ := p.Attributes["status"].(string)
+		if status != "" {
 			// FixState is repurposed here to carry the SecurityException's status vocabulary
 			// (not_affected/fixed/affected) rather than Grype's native fix-state vocabulary.
 			// This is safe because every reader of FixState gates on SourceKind being a
@@ -395,11 +396,30 @@ func buildIgnoreRules(m v1beta1.Match, matched []armotypes.VulnerabilityExceptio
 			// never set SourceKind, so they're unaffected.
 			rule.FixState = status
 		}
-		if just, ok := p.Attributes["justification"].(string); ok {
-			rule.Justification = just
-		}
-		if impact, ok := p.Attributes["impactStatement"].(string); ok {
-			rule.ImpactStatement = impact
+		if status == string(sev1beta1.VulnerabilityStatusAffected) {
+			// An affected entry populates actionStatement/response, not justification/
+			// impactStatement (those are the not_affected fields; shouldSuppress requires
+			// actionStatement, never justification, for an affected entry to suppress at
+			// all). IgnoreRule has no dedicated fields for either, so — mirroring the
+			// FixState repurposing above, and gated the same way on FixState=="affected"
+			// plus a SecurityException/ClusterSecurityException SourceKind — ImpactStatement
+			// carries the action statement and Justification carries the comma-joined
+			// response values. ignoredMatchAssessment (repositories/apiserver.go) is the
+			// paired reader; without this, every affected suppression exported the exact
+			// same generic action statement regardless of what its author wrote (#866).
+			if action, ok := p.Attributes["actionStatement"].(string); ok {
+				rule.ImpactStatement = action
+			}
+			if responses, ok := p.Attributes["response"].([]string); ok && len(responses) > 0 {
+				rule.Justification = strings.Join(responses, ",")
+			}
+		} else {
+			if just, ok := p.Attributes["justification"].(string); ok {
+				rule.Justification = just
+			}
+			if impact, ok := p.Attributes["impactStatement"].(string); ok {
+				rule.ImpactStatement = impact
+			}
 		}
 		rules = append(rules, rule)
 	}
@@ -421,9 +441,10 @@ func suppressingPolicies(matched []armotypes.VulnerabilityExceptionPolicy) []arm
 }
 
 // logSuppression records why a CVE disappeared from scan results: which exception object
-// matched, its scope, and the stated reason/justification. This is logged rather than stored on
-// the manifest because the filtered manifest's IgnoreRule (github.com/kubescape/storage) has no
-// fields for this provenance; see buildSuppressionAttributes for where it is captured.
+// matched, its scope, and the stated reason/justification. Some of this (sourceKind, ruleId,
+// justification/impactStatement, actionStatement/response) is also durably written onto the
+// manifest's IgnoreRule by buildIgnoreRules, directly or repurposing a field with no dedicated
+// slot; normalizedTarget and the reason string are not, and live only here.
 //
 // When recorder is non-nil, the same suppression is also recorded as a K8s Event on the
 // SecurityException/ClusterSecurityException object that caused it, so it shows up in
@@ -454,6 +475,12 @@ func logSuppression(m v1beta1.Match, matched []armotypes.VulnerabilityExceptionP
 		}
 		if impact, ok := p.Attributes["impactStatement"].(string); ok && impact != "" {
 			fields = append(fields, helpers.String("impactStatement", impact))
+		}
+		if action, ok := p.Attributes["actionStatement"].(string); ok && action != "" {
+			fields = append(fields, helpers.String("actionStatement", action))
+		}
+		if responses, ok := p.Attributes["response"].([]string); ok && len(responses) > 0 {
+			fields = append(fields, helpers.String("response", strings.Join(responses, ",")))
 		}
 		if target, ok := p.Attributes["normalizedTarget"].(string); ok && target != "" {
 			fields = append(fields, helpers.String("normalizedTarget", target))
--- a/adapters/v1/securityexception_test.go
+++ b/adapters/v1/securityexception_test.go
@@ -336,6 +336,142 @@ func TestApplySecurityExceptions_PopulatesIgnoreRuleProvenance(t *testing.T) {
 	assert.Equal(t, "Vulnerable component is not loaded into the memory", rule.ImpactStatement)
 }
 
+// TestApplySecurityExceptions_PopulatesAffectedProvenance is the affected-status counterpart to
+// TestApplySecurityExceptions_PopulatesIgnoreRuleProvenance: an affected entry's actionStatement
+// and response never populate Justification/ImpactStatement directly, so without repurposing
+// those fields the way buildIgnoreRules does here, the real text an author wrote never reaches
+// the persisted IgnoreRule at all, and every affected suppression is indistinguishable on the
+// stored manifest (#866).
+func TestApplySecurityExceptions_PopulatesAffectedProvenance(t *testing.T) {
+	doc := &v1beta1.GrypeDocument{
+		Matches: []v1beta1.Match{
+			{Vulnerability: v1beta1.Vulnerability{VulnerabilityMetadata: v1beta1.VulnerabilityMetadata{ID: "CVE-2021-44228"}}},
+		},
+	}
+	exceptions := domain.CVEExceptions{
+		{
+			PortalBase: armotypes.PortalBase{
+				Name: "risk-accepted-log4j",
+				Attributes: map[string]interface{}{
+					"sourceKind":      "SecurityException",
+					"ruleId":          "SecurityException/production/risk-accepted-log4j",
+					"sourceNamespace": "production",
+					"status":          "affected",
+					"actionStatement": "WAF mitigation in place, ticket SEC-1234",
+					"response":        []string{"will_not_fix"},
+				},
+			},
+			PolicyType:            "vulnerabilityExceptionPolicy",
+			Actions:               []armotypes.VulnerabilityExceptionPolicyActions{armotypes.Ignore},
+			VulnerabilityPolicies: []armotypes.VulnerabilityPolicy{{Name: "CVE-2021-44228"}},
+		},
+	}
+
+	ApplySecurityExceptions(doc, exceptions, nil)
+
+	require.Len(t, doc.IgnoredMatches, 1)
+	require.Len(t, doc.IgnoredMatches[0].AppliedIgnoreRules, 1)
+	rule := doc.IgnoredMatches[0].AppliedIgnoreRules[0]
+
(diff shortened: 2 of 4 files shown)


matthyx (approved):
Reviewed the fix, tests, and both touched files in full.

The core change is sound: `buildIgnoreRules` now repurposes `Justification`/`ImpactStatement` on `IgnoreRule` to carry `response`/`actionStatement` for `affected` entries specifically, since `IgnoreRule` (owned by `kubescape/storage`) has no dedicated fields and adding them would be a cross-repo schema change. The repurposing is safely gated: it only fires when `status == "affected"`, and per `shouldSuppress`/`affectedSuppresses`, an affected entry can only reach `buildIgnoreRules` at all if it already has both a non-empty `actionStatement` and a non-remediating `response`, and never needs `justification` to suppress — so nothing is silently dropped for entries that actually reach this path. `ignoredMatchAssessment`'s new `affectedActionStatement`/`affectedStatusNotes` helpers read it back correctly, falling back to the old generic constant only when the repurposed field is empty (pre-fix cached manifests).

Verified independently rather than just trusting the PR description:
- Cloned the branch and ran `go build`/`go vet` on `adapters/v1` and `repositories` — clean.
- Ran `go test ./adapters/v1/... ./repositories/...` — all pass, including the new/updated tests covering the affected-provenance path, the two-distinct-entries-don't-collapse case, the not_affected-is-unaffected-by-the-new-branch guard, and the pre-fix fallback path.
- Also ran a full repo-wide `go build ./...`, which the author flagged they couldn't complete locally — it succeeds cleanly, so there's no fallout in the wider tree either.
- All CI checks (tests, lint, CodeQL, cross-platform build, DCO, secret scan) are green. The CodeRabbit comment is just a rate-limit notice, no findings.

No behavior change for `not_affected`/`fixed` suppressions or pre-existing cached manifests, as claimed. Good to merge.

PR #872, opened 2026-08-26
fix(repositories): give an affected SecurityException precedence when multiple exceptions match

### LINKED ISSUE #871: securityExceptionIgnoreRule picks the first matching rule, silently dropping every other exception's status/justification when multiple SecurityExceptions match one CVE
## Description

When more than one `SecurityException`/`ClusterSecurityException` matches the same CVE finding — a common overlap in practice, e.g. a broad cluster-scoped exception and a more specific namespaced one, or two policies matching the same CVE by ID and by alias — `buildIgnoreRules` correctly records one `IgnoreRule` per suppressing policy, each with its own `status`/justification/action-statement. But the reader that turns those rules into the exported OpenVEX assessment, `securityExceptionIgnoreRule`, silently picks the *first* rule in the slice and discards every other one. If the discarded rule is the one carrying the real, author-written `affected`/risk-accepted justification, the exported VEX document instead reports a generic `not_affected` claim — the opposite status, with no trace of the operator's actual risk-acceptance record.

This is the same class of provenance loss that #866 just fixed for a single exception, one level up: it now happens even when the value is captured correctly, purely because of which policy happens to come first.

## Reproduction / Evidence

`adapters/v1/securityexception.go:371-427` (`buildIgnoreRules`):

```go
func buildIgnoreRules(m v1beta1.Match, matched []armotypes.VulnerabilityExceptionPolicy) []v1beta1.IgnoreRule {
	suppressing := suppressingPolicies(matched)
	...
	rules := make([]v1beta1.IgnoreRule, 0, len(suppressing))
	for _, p := range suppressing {
		rule := v1beta1.IgnoreRule{Vulnerability: m.Vulnerability.ID}
		...
		rules = append(rules, rule)
	}
	return rules
}
```

This is deliberate — the function's own doc comment says it populates provenance "from EVERY suppressing policy's Attributes ... so complete provenance is durably stored on the manifest itself when multiple policies match." `matched` comes from `exceptionIndex.lookup` (`adapters/v1/backend_utils.go:379-399`), which by design returns every exception declaring a policy for that CVE name, in list order — not deduplicated to one. `hasIgnoreAction` (`adapters/v1/securityexception.go:303-312`) only requires *any* matched policy to carry the `Ignore` action, so a mix of statuses among the matched policies (e.g. one `not_affected`, one `affected`) is a normal, reachable combination, not a validation error.

`repositories/apiserver.go:1258-1266` (`securityExceptionIgnoreRule`) then reads that slice back:

```go
func securityExceptionIgnoreRule(m v1beta1.IgnoredMatch) (v1beta1.IgnoreRule, bool) {
	for _, rule := range m.AppliedIgnoreRules {
		switch rule.SourceKind {
		case "SecurityException", "ClusterSecurityException":
			return rule, true
		}
	}
	return v1beta1.IgnoreRule{}, false
}
```

It returns on the first match and never looks at the rest. `ignoredMatchAssessment` (`repositories/apiserver.go:1196-1241`) then derives `status`, `justification`, `impactStatement`, `actionStatement` and `statusNotes` purely from that one rule.

**Concrete scenario:**
1. A `ClusterSecurityException` matches `CVE-2024-XXXX` by ID with `status: not_affected` (a blanket, low-effort suppression).
2. A namespaced `SecurityException` for the same workload also matches `CVE-2024-XXXX`, with `status: affected`, a real `actionStatement` ("WAF mitigation in place, ticket SEC-1234"), and `response: [will_not_fix]`.
3. Both policies pass `hasIgnoreAction`, so both suppress the match and both get an `IgnoreRule` in `AppliedIgnoreRules`.
4. If the cluster-scoped exception's rule happens to sort before the namespaced one in `matched` (order comes from `exceptionIndex.lookup`'s underlying exception list, not any precedence rule), `securityExceptionIgnoreRule` returns *that* rule.
5. The exported VEX statement reports `status: not_affected` with a generic justification — the namespaced exception's `affected` status and its entire audit trail (`actionStatement`, `response`) never reach the output, with no error, warning, or indication that a second, conflicting exception was even involved.

No existing test exercises this: `repositories/apiserver_external_test.go`'s table for `ignoredMatchAssessment`/`securityExceptionIgnoreRule` (around line 512-548) supplies exactly one rule per case ("namespaced CRD", "cluster-scoped CRD", "backend-delivered exception policy", "not ours") — never more than one `SecurityException`/`ClusterSecurityException` rule together.

## Impact

- **Correctness / compliance**: the exported OpenVEX document — the artifact meant to be an audit-grade, machine-readable record of vulnerability disposition — can report the wrong status and drop a real risk-acceptance justification whenever two exceptions overlap on one CVE, which is a normal cluster configuration (org-wide baseline exceptions plus team-specific overrides).
- **Silent**: nothing logs, counts, or warns that a match had more than one applicable exception, so operators have no signal that this happened — the same failure mode #866 fixed for a single rule, now reachable even after that fix, through a different path.
- **No defined precedence**: separate from the loss itself, there's no policy for *which* exception should win when several apply (most specific scope? most recent? explicit `affected` over `not_affected`?) — right now it's whatever order the backing exception list happens to produce.

## Affected Components

- `adapters/v1/securityexception.go` — `buildIgnoreRules`, `hasIgnoreAction`, `suppressingPolicies`
- `repositories/apiserver.go` — `securityExceptionIgnoreRule`, `ignoredMatchAssessment`
- `adapters/v1/backend_utils.go` — `cveExceptionIndex.lookup` (source of matched-policy ordering)

## Acceptance Criteria

- When multiple `SecurityException`/`ClusterSecurityException` rules apply to the same match, the exported VEX assessment reflects an explicit, documented precedence (e.g. `affected` takes priority over `not_affected`/`fixed` so a real risk-acceptance is never silently overridden by a blanket suppression — or whatever ordering the maintainers prefer), not incidental list order.
- The discarded rule(s)' provenance is not silently lost: at minimum it should be recoverable from the manifest (e.g. still present in `AppliedIgnoreRules`, which it already is) and ideally surfaced (log line, event, or combined into `statusNotes`) so a conflict between exceptions is visible.
- Regression test coverage for `securityExceptionIgnoreRule`/`ignoredMatchAssessment` with more than one `SecurityException`/`ClusterSecurityException` rule present, including the case where an `affected` rule and a `not_affected`/`fixed` rule both apply to the same match.

## Suggested Direction

Give `securityExceptionIgnoreRule` (or a new helper it delegates to) an explicit precedence instead of first-match: e.g. prefer a rule with `FixState == "affected"` over any other status among the SE/CSE-sourced rules, and only fall back to first-match among rules of equal precedence. This keeps the fix localized to the read side (`repositories/apiserver.go`) and requires no `IgnoreRule` schema change, mirroring how #866 was scoped.

bhuvan-somisetty:
## Problem

A CVE match can be suppressed by more than one `SecurityException`/`ClusterSecurityException` at once — e.g. a broad, cluster-scoped exception and a more specific namespaced override both matching the same CVE. `buildIgnoreRules` (adapters/v1) correctly records one `IgnoreRule` per suppressing policy in that case, but `securityExceptionIgnoreRule` (repositories/apiserver.go), which picks the rule that drives the exported OpenVEX assessment, returned whichever rule happened to be listed first and silently discarded the rest.

If an unattributed `not_affected`/`fixed` rule from one exception happened to sort before an `affected` (risk-accepted) rule from another, the exported VEX statement reported `not_affected` and dropped the second exception's real, author-written `actionStatement`/`response` entirely — with no error, log, or any signal that a conflicting exception was even involved.

## Root Cause

`securityExceptionIgnoreRule` (`repositories/apiserver.go`) looped over `m.AppliedIgnoreRules` and returned on the first `SecurityException`/`ClusterSecurityException`-sourced rule it found:

```go
func securityExceptionIgnoreRule(m v1beta1.IgnoredMatch) (v1beta1.IgnoreRule, bool) {
	for _, rule := range m.AppliedIgnoreRules {
		switch rule.SourceKind {
		case "SecurityException", "ClusterSecurityException":
			return rule, true
		}
	}
	return v1beta1.IgnoreRule{}, false
}
```

Rule order comes from `cveExceptionIndex.lookup` (`adapters/v1/backend_utils.go`), which returns exceptions in the order they appear in the source list — not by any precedence rule. `hasIgnoreAction` only requires *any* matched policy to carry the `Ignore` action, so a mix of `affected` and `not_affected`/`fixed` rules among the matched policies is a normal, reachable combination (see #871 for the full writeup and a concrete repro scenario).

No existing test exercised more than one `SecurityException`/`ClusterSecurityException` rule reaching this function — the table in `TestIgnoredMatchAssessment_AttributesBySource` only ever supplied a single rule per case.

## Fix

`securityExceptionIgnoreRule` now scans every SE/CSE-sourced rule and prefers one with `FixState == "affected"` if present, falling back to the first SE/CSE rule otherwise (identical to the previous behavior for the common single-exception case).

An `affected` rule always wins because it is never a default the way `not_affected`/`fixed` can be: `shouldSuppress` requires an explicit, author-written `actionStatement` before an `affected` entry is even allowed to suppress a finding, so its presence always represents a deliberate risk acceptance that a less specific exception's automatic `not_affected`/`fixed` should not be able to silently override.

The change is confined to the read side in `repositories/apiserver.go` — no schema change, no change to how `buildIgnoreRules` records provenance.

## Testing

- `go build ./...` — passes
- `go vet ./...` — passes
- `go test ./repositories/... ./adapters/...` — passes
- `golangci-lint run --new-from-rev=upstream/main ./repositories/...` — 0 issues

Added regression coverage in `repositories/apiserver_external_test.go`:
- `TestIgnoredMatchAssessment_MultipleExceptions_AffectedTakesPrecedence` — an `affected` rule and a `not_affected` rule both present, asserted both orderings (`affected` listed first and listed last) produce the same, correct `affected` status with the real `actionStatement`/response preserved. The "listed last" case is the one that reproduces the original bug — it would have failed against the old first-match code.
- `TestIgnoredMatchAssessment_MultipleExceptions_AffectedBeatsFixed` — `affected` also takes precedence over `fixed`, not just `not_affected`.
- `TestIgnoredMatchAssessment_MultipleExceptions_NoAffected_FirstRuleWins` — with no `affected` rule present, the first SE/CSE rule is still used, confirming the common single/no-conflict case is unchanged.

## Impact

Restores correctness of the exported OpenVEX document — the audit-grade record of vulnerability disposition — whenever two or more `SecurityException`/`ClusterSecurityException` resources apply to the same finding, which is a normal configuration (org-wide baseline exceptions plus team-specific overrides). A deliberate, documented risk acceptance can no longer be silently overridden by an unrelated blanket suppression purely due to list order.

Fixes #871

## Summary by CodeRabbit

- **Bug Fixes**
  - Security exception handling now consistently prioritizes rules explicitly marked as affected.
  - Preserves the correct action and status details from the applicable rule.
  - Retains expected first-match behavior when no affected rule is present.

- **Tests**
  - Added regression coverage for rule ordering, affected status, actions, and status notes.

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -1255,14 +1255,39 @@ func isOwnIgnoreRule(m v1beta1.IgnoredMatch) bool {
 	return m.AppliedIgnoreRules[0].Package == nil
 }
 
+// securityExceptionIgnoreRule can face more than one candidate: buildIgnoreRule
+// (adapters/v1) writes one IgnoreRule per suppressing policy, and it is normal for a match
+// to be suppressed by several exceptions at once -- e.g. a broad ClusterSecurityException
+// and a more specific namespaced SecurityException both matching the same CVE. Picking
+// whichever rule happens to be listed first would let an unattributed not_affected/fixed
+// rule silently outrank an affected rule's author-written actionStatement/response, purely
+// because of incidental ordering.
+//
+// An affected rule always wins when one is present: shouldSuppress requires an explicit
+// actionStatement before an affected entry can suppress at all, so it is always a deliberate,
+// specific risk acceptance -- never one this function should let a less specific rule
+// override. Among rules that agree on not being affected (or when none is), the first
+// SE/CSE rule is used, preserving prior behavior for the common single-exception case.
 func securityExceptionIgnoreRule(m v1beta1.IgnoredMatch) (v1beta1.IgnoreRule, bool) {
+	var (
+		first v1beta1.IgnoreRule
+		found bool
+	)
 	for _, rule := range m.AppliedIgnoreRules {
 		switch rule.SourceKind {
 		case "SecurityException", "ClusterSecurityException":
+		default:
+			continue
+		}
+		if !found {
+			first = rule
+			found = true
+		}
+		if strings.TrimSpace(rule.FixState) == string(sev1beta1.VulnerabilityStatusAffected) {
 			return rule, true
 		}
 	}
-	return v1beta1.IgnoreRule{}, false
+	return first, found
 }
 
 func ignoredMatchStatusNotes(rule v1beta1.IgnoreRule) string {
--- a/repositories/apiserver_external_test.go
+++ b/repositories/apiserver_external_test.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/kubescape/kubevuln/core/domain"
 	"github.com/kubescape/kubevuln/internal/tools"
+	sev1beta1 "github.com/kubescape/kubevuln/pkg/securityexception/v1beta1"
 	"github.com/kubescape/storage/pkg/apis/softwarecomposition/v1beta1"
 	"github.com/openvex/go-vex/pkg/vex"
 )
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Reviewed the fix and verified it directly against the checked-out branch:

- `securityExceptionIgnoreRule` now scans every SE/CSE-sourced rule instead of returning on the first match, preferring a `FixState == "affected"` rule and falling back to the first SE/CSE rule when none is affected — this correctly fixes the silent-override bug from #871.
- Confirmed the `shouldSuppress` invariant the fix relies on: an `affected` entry requires an explicit `ActionStatement`/`Response` to suppress at all (`adapters/v1/securityexception.go:166`), so `affected` is always a deliberate, non-default decision — justifying why it should win regardless of list order.
- Ran `go build ./...`, `go vet ./repositories/... ./adapters/...`, and `go test ./repositories/... ./adapters/...` on the branch — all pass, including the three new regression tests (`AffectedTakesPrecedence` in both orderings, `AffectedBeatsFixed`, `NoAffected_FirstRuleWins`).
- Ran `golangci-lint` on `repositories/...`; the only findings are pre-existing `staticcheck` QF1008/QF1003 style suggestions in unrelated lines (1562, 1735-1736, and existing test code) — nothing introduced by this diff.
- Checked the stray doc-comment sitting above `isOwnIgnoreRule` that actually describes `securityExceptionIgnoreRule` — confirmed via `git show main` that this misplacement predates this PR, so it's not something this diff needs to fix.

No blockers. Good to merge.

PR #874, opened 2026-08-26
fix: honor CPE matching off mode for stock matcher

### LINKED ISSUE #873: CVEMatchingOff still enables CPE matching through Grype's stock matcher
### Description

CVEMatchingOff is expected to disable CPE-based matching, but the stock Grype matcher is currently configured with ``UseCPEs: true``, causing CPE matches to be returned anyway.

Root cause: ``defaultMatcherConfig()`` sets:

Stock: ``stock.MatcherConfig{UseCPEs: true``},

which causes Grype's stock matcher to call its CPE matching path.

### Steps to reproduce

- Create a mock vulnerability for java_se with a matching CPE.
- Create java_se@1.8.0_400 with the same CPE.
- Run Grype's VulnerabilityMatcher with getMatchers(true) (CVEMatchingOff).

Observe a match with:
```
matcher=stock-matcher
type=cpe-match
```
The same fixture with ``getMatchers(false)`` ``(CVEMatchingOn)`` produces the expected CPE match.

Expected: ``CVEMatchingOff`` produces no CPE matches.
Actual: A CPE match is still produced by ``stock-matcher``.

jijo-OO7:
Resolves : #873

## Overview

The default Grype matcher configuration disabled CPE matching for language-specific matchers, but left the stock matcher configured with `UseCPEs: true`.

As a result, `CVEMatchingOff` could still produce CPE-based vulnerability matches for packages carrying CPEs.

## Changes

- Disable `UseCPEs` for Grype's stock matcher in the default matcher configuration.
- Add regression coverage for both CPE matching modes:
  - `CVEMatchingOff` produces no CPE matches.
  - `CVEMatchingOn` still produces CPE matches.

## Validation

- `TestDefaultMatchers_CVEMatchingOffDoesNotProduceCPEMatch` — passes
- `TestDefaultMatchers_CVEMatchingOnProducesCPEMatch` — passes repeatedly (`-count=10`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated default vulnerability matching to consistently disable CPE-based matches when CPE matching is turned off.
  * Ensured CPE matches are still returned when CPE matching is enabled.

* **Tests**
  * Added coverage for CPE matching behavior in both enabled and disabled configurations.

### DIFF
--- a/adapters/v1/grype.go
+++ b/adapters/v1/grype.go
@@ -578,7 +578,7 @@ func defaultMatcherConfig() matcher.Config {
 			AlwaysUseCPEForStdlib:                  true,
 			AllowMainModulePseudoVersionComparison: false,
 		},
-		Stock: stock.MatcherConfig{UseCPEs: true},
+		Stock: stock.MatcherConfig{UseCPEs: false},
 	}
 }
 
--- a/adapters/v1/grype_test.go
+++ b/adapters/v1/grype_test.go
@@ -9,10 +9,16 @@ import (
 	"testing"
 	"time"
 
+	"github.com/anchore/grype/grype"
 	"github.com/anchore/grype/grype/db/v6/distribution"
 	"github.com/anchore/grype/grype/db/v6/installation"
 	"github.com/anchore/grype/grype/distro"
+	"github.com/anchore/grype/grype/pkg"
+	"github.com/anchore/grype/grype/version"
 	"github.com/anchore/grype/grype/vulnerability"
+	"github.com/anchore/grype/grype/vulnerability/mock"
+	"github.com/anchore/syft/syft/cpe"
+	syftPkg "github.com/anchore/syft/syft/pkg"
 	"github.com/google/uuid"
 	"github.com/kinbiko/jsonassert"
 	"github.com/kubescape/kubevuln/config"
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Verified the root cause against `resolveUseDefaultMatchers`/`getMatchers`: for `CVEMatchingOff`, `defaultMatcherConfig()` is used, and every matcher there had `UseCPEs: false` except `Stock`, which was left at `true` — so the stock matcher's CPE path stayed live even when CPE matching was supposed to be off. Flipping `Stock.UseCPEs` to `false` closes that gap and brings it in line with the other matchers (java/ruby/python/dotnet/javascript/golang) in the same config.

Also checked that this doesn't affect `CVEMatchingOn` (uses the separate all-CPEs-true config in `getMatchers(false)`) or `CVEMatchingAdaptive` for untrusted vendors (same all-CPEs-true path) — only the CPE-off default config used for `CVEMatchingOff` and trusted-vendor adaptive scans changes, which is the intended behavior.

Tests are solid: `TestDefaultMatchers_CVEMatchingOffDoesNotProduceCPEMatch` and `TestDefaultMatchers_CVEMatchingOnProducesCPEMatch` reproduce the exact repro from #873 and correctly assert 0 matches / 1 CPE match respectively. All CI checks (build, lint, CodeQL, GitGuardian, DCO, Fossa) are green.

PR #876, opened 2026-08-26
fix(adapters): check subcomponent scope before excluding a CRD exception for cloud precedence

### LINKED ISSUE #875: excludeCloudCoveredPolicies ignores subcomponent scope, so a package-scoped CRD exception is silently dropped by an unrelated cloud exception on the same CVE
## Description

`excludeCloudCoveredPolicies` (added in #813 to implement the documented cloud-over-CRD precedence, #812) drops a CRD-derived `SecurityException`/`ClusterSecurityException` policy whenever a cloud-delivered exception exists for the *same CVE ID*, regardless of which **package** each one is scoped to. Both exception sources support scoping an exception to specific packages via the `subcomponents` purl list (`docs/security-exception-design.md`, "Identifier Bridging" section) — the same `Attributes["subcomponents"]` mechanism, since CRD- and cloud-delivered policies are both instances of `armotypes.VulnerabilityExceptionPolicy`, which carries scope in the same generic `Attributes` bag either source can populate.

When a cloud exception for CVE-X is scoped to package A, and a CRD exception for the same CVE-X is scoped to a *different* package B, the current code excludes the CRD exception entirely — even though the two don't actually overlap. The CVE-X finding on package B, which the CRD exception was written specifically to suppress, ends up suppressed by neither source and reappears in scan results and the exported VEX document, with no error or log indicating why the CRD exception silently stopped applying.

This contradicts the design doc's own "non-overlapping exceptions are merged" clause (`docs/security-exception-design.md:388`): two exceptions scoped to disjoint packages are non-overlapping in every sense that matters to suppression, but the current CVE-name-only comparison treats them as if they conflict.

## Reproduction / Evidence

`adapters/v1/backend.go:346-366`:

```go
func excludeCloudCoveredPolicies(crdPolicies, cloudPolicies []armotypes.VulnerabilityExceptionPolicy) []armotypes.VulnerabilityExceptionPolicy {
	if len(cloudPolicies) == 0 || len(crdPolicies) == 0 {
		return crdPolicies
	}
	cloudIndex := buildCVEExceptionIndex(cloudPolicies)

	var filtered []armotypes.VulnerabilityExceptionPolicy
	for _, p := range crdPolicies {
		coveredByCloud := false
		for _, vp := range p.VulnerabilityPolicies {
			if len(cloudIndex.byName[strings.ToLower(vp.Name)]) > 0 {
				coveredByCloud = true
				break
			}
		}
		if !coveredByCloud {
			filtered = append(filtered, p)
		}
	}
	return filtered
}
```

The lookup is keyed purely by lower-cased CVE/alias name (`buildCVEExceptionIndex`, `adapters/v1/backend_utils.go:354-371`) and never consults `Attributes["subcomponents"]` (`adapters/v1/subcomponent.go:16-60`), which is where package-level scope actually lives on the very same `armotypes.VulnerabilityExceptionPolicy` values being compared. `TestExcludeCloudCoveredPolicies` (`adapters/v1/backend_test.go:1134-1176`) confirms this is untested for the scoped case: every table entry builds policies with no `subcomponents` attribute at all.

I verified the failure directly by exercising the exact sequence `GetCVEExceptions` runs (`adapters/v1/backend.go:299-304`) — `excludeCloudCoveredPolicies` then `scopedToSubcomponent` — with a throwaway local test:

```go
cloudPolicies := []armotypes.VulnerabilityExceptionPolicy{{
	PortalBase:             armotypes.PortalBase{Attributes: map[string]interface{}{attrSubcomponents: []interface{}{"pkg:npm/foo"}}},
	VulnerabilityPolicies:  []armotypes.VulnerabilityPolicy{{Name: "CVE-9999"}},
}}
crdPolicies := []armotypes.VulnerabilityExceptionPolicy{{
	PortalBase:             armotypes.PortalBase{Name: "crd-rule", Attributes: map[string]interface{}{attrSubcomponents: []string{"pkg:npm/bar"}}},
	VulnerabilityPolicies:  []armotypes.VulnerabilityPolicy{{Name: "CVE-9999"}},
}}

filtered := excludeCloudCoveredPolicies(crdPolicies, cloudPolicies) // len(filtered) == 0
merged := append(append([]armotypes.VulnerabilityExceptionPolicy{}, cloudPolicies...), filtered...)
scopedToSubcomponent(merged, "pkg:npm/bar@2.0.0")                   // len == 0 -- nothing suppresses bar's CVE-9999
```

`filtered` comes back empty: the `bar`-scoped CRD policy is dropped even though the cloud policy only covers `foo`. After merging, no policy in the result applies to package `bar`, so a finding of `CVE-9999` in package `bar` — which the CRD exception exists specifically to suppress — is left unsuppressed.

## Root Cause

`excludeCloudCoveredPolicies` treats "same CVE ID" as sufficient evidence of a conflict between a cloud and a CRD exception, but the actual conflict the design doc describes is "for the same CVE **on the same workload**" at the granularity the exception system itself defines — and that system supports narrowing below workload level to individual packages via `subcomponents`. The function was written and tested (#813) only against the workload-wide case that motivated #812; the package-scoped case was never exercised, so the CVE-name index doesn't carry (and was never asked to carry) subcomponent scope.

## Impact

- **Suppression-correctness regression**: a legitimately-scoped, actively-matching CRD exception silently stops suppressing its target finding as soon as an unrelated cloud exception for the same CVE ID (covering a different package) exists — a realistic condition during the "gradual migration" the design doc itself describes (cloud and CRD exceptions "work simultaneously").
- **Silent**: no log, event, or metric indicates the CRD exception was excluded; from the operator's perspective the exception simply stopped working.
- **Directly contradicts documented behavior**: `docs/security-exception-design.md`'s own "non-overlapping exceptions are merged" clause is violated for exceptions that only appear to overlap at CVE-ID granularity.

## Affected Components

- `adapters/v1/backend.go` — `excludeCloudCoveredPolicies`
- `adapters/v1/backend_utils.go` — `cveExceptionIndex`/`buildCVEExceptionIndex` (the index it reuses, which is name-only by design for its other use in `ApplySecurityExceptions`)
- `adapters/v1/subcomponent.go` — `policySubcomponents`/`scopedToSubcomponent` (the scope this exclusion needs to consult but doesn't)

## Acceptance Criteria

- A CRD policy is excluded by `excludeCloudCoveredPolicies` only when a cloud policy for the same CVE actually covers it: either the cloud policy is unscoped (product-wide) or its `subcomponents` overlap the CRD policy's.
- A CRD policy scoped to a package no cloud policy for the same CVE covers continues to apply after the merge, matching `scopedToSubcomponent`'s own overlap semantics.
- Regression test coverage: same CVE, disjoint `subcomponents` scopes on each side (CRD policy survives and still applies to its package); same CVE, overlapping/superset scopes (CRD policy is still excluded, preserving #812's fix); one side scoped and the other unscoped (unscoped cloud policy still excludes, since it covers everything the CRD policy could).

## Suggested Direction

Extend `excludeCloudCoveredPolicies`'s per-CRD-policy check: after finding a cloud policy matching the CVE name, only treat it as covering the CRD policy when their `subcomponents` scopes actually overlap (an unscoped side means "covers everything"; two scoped sides need at least one shared/matching purl, reusing `policySubcomponents`/`purlMatches` from `subcomponent.go` rather than duplicating that logic). This keeps the fix localized to the one function responsible for the precedence decision.

bhuvan-somisetty:
## Problem

Both cloud-delivered and CRD-derived (`SecurityException`/`ClusterSecurityException`) exception policies can independently scope an exception to specific packages via the `subcomponents` purl list (`docs/security-exception-design.md`, "Identifier Bridging" section) — they're both instances of the same `armotypes.VulnerabilityExceptionPolicy` type, carrying that scope in the same `Attributes["subcomponents"]` bag either source can populate.

`excludeCloudCoveredPolicies` (added in #813 to implement the documented cloud-over-CRD precedence, #812) dropped a CRD policy whenever a cloud policy existed for the *same CVE ID*, without ever checking which package each one was scoped to. So a cloud exception for CVE-X scoped to package A silently excluded a CRD exception for the same CVE-X scoped to a *different* package B — even though the two don't actually overlap. The CVE-X finding on package B, which the CRD exception existed specifically to suppress, ended up suppressed by neither source: it reappears in scan results and the exported VEX document, with no error or log explaining why.

This contradicts `docs/security-exception-design.md`'s own "non-overlapping exceptions are merged" clause: two exceptions scoped to disjoint packages are non-overlapping, but the CVE-name-only comparison treated them as conflicting.

## Root Cause

`excludeCloudCoveredPolicies` (`adapters/v1/backend.go`) built a CVE-name index over the cloud policies and dropped any CRD policy sharing a name with one of them — full stop, no scope check:

```go
for _, p := range crdPolicies {
	coveredByCloud := false
	for _, vp := range p.VulnerabilityPolicies {
		if len(cloudIndex.byName[strings.ToLower(vp.Name)]) > 0 {
			coveredByCloud = true
			break
		}
	}
	if !coveredByCloud {
		filtered = append(filtered, p)
	}
}
```

`TestExcludeCloudCoveredPolicies` had zero coverage for a scoped policy on either side.

I verified the failure directly, running the exact sequence `GetCVEExceptions` runs (exclude, then `scopedToSubcomponent`) with a cloud policy scoped to `pkg:npm/foo` and a CRD policy scoped to a different package `pkg:npm/bar`, both for the same CVE: the CRD policy was dropped, and after merging, nothing in the result applied to package `bar` — the finding it was written to suppress went unsuppressed.

## Fix

`excludeCloudCoveredPolicies` now delegates the per-policy decision to a new `cloudCoversPolicy` helper that checks package scope, not just CVE name:

- An unscoped cloud policy still covers everything (preserves #812's fix for the workload-wide case that motivated it).
- A scoped cloud policy covers a CRD policy only when the CRD policy's own scope is fully contained in it (`purlsCoveredBy`, reusing the existing `purlMatches`/`anyPURLMatches` helpers from `subcomponent.go` rather than duplicating purl-matching logic).
- A scoped cloud policy never covers an *unscoped* CRD policy: the CRD policy also applies to packages the cloud policy never addressed, so dropping it would lose that coverage.

The change is confined to `excludeCloudCoveredPolicies` and its two new helpers in `adapters/v1/backend.go` — no changes to how scope is recorded or read elsewhere.

## Testing

- `go build ./...` — passes
- `go vet ./adapters/...` — passes
- `go test ./adapters/... ./repositories/...` — passes
- `golangci-lint run --new-from-rev=upstream/main ./adapters/...` — 0 issues

Added `TestExcludeCloudCoveredPolicies_SubcomponentScope` in `adapters/v1/backend_test.go`:
- disjoint packages (cloud scoped to `foo`, CRD scoped to `bar`) — CRD policy survives, the case that reproduces the original bug.
- same package on both sides — CRD policy is still dropped, preserving #812's fix.
- unscoped cloud policy, scoped CRD policy — dropped (cloud covers everything).
- scoped cloud policy, unscoped CRD policy — CRD policy survives (cloud doesn't cover the rest of what the CRD policy applies to).

All pre-existing `TestExcludeCloudCoveredPolicies` and `TestGetCVEExceptions_*` cases (unscoped policies throughout) pass unchanged, confirming the common single-scope case is unaffected.

## Impact

Restores suppression correctness for the "gradual migration" scenario `docs/security-exception-design.md` itself describes — cloud and CRD exceptions coexisting simultaneously. A CRD exception scoped to one package no longer silently stops applying just because an unrelated cloud exception exists for the same CVE ID on a different package.

Fixes #875

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved policy filtering to account for package scope coverage when matching cloud and CRD policies.
  * Prevented policies from being incorrectly excluded when package coverage is partial, disjoint, or unscoped.
  * Ensured fully covered policies are excluded while preserving policies with distinct package scopes.

* **Tests**
  * Added coverage for scoped and unscoped policy combinations.

### DIFF
--- a/adapters/v1/backend.go
+++ b/adapters/v1/backend.go
@@ -340,9 +340,14 @@ func (a *BackendAdapter) GetCVEExceptions(ctx context.Context) (domain.CVEExcept
 //
 // Reuses cveExceptionIndex's own name normalization (lower-cased CVE ID/alias) so a CRD
 // exception naming a CVE by an alias the cloud exception names by its canonical ID, or vice
-// versa, still counts as the same CVE. A CRD policy not covered by any cloud policy passes
-// through unchanged, matching the doc's separate "non-overlapping exceptions are merged"
-// clause.
+// versa, still counts as the same CVE. Coverage is then checked at the same package
+// (subcomponent) granularity scopedToSubcomponent applies at suppression time, not by CVE name
+// alone: both cloud- and CRD-derived policies can independently scope to specific packages via
+// the shared Attributes["subcomponents"] mechanism (subcomponent.go), so a cloud policy naming
+// the same CVE but scoped to a different package does not make a CRD policy redundant.
+// Excluding it anyway would silently drop suppression for the package the cloud exception never
+// covered (#875). A CRD policy not covered by any cloud policy passes through unchanged,
+// matching the doc's separate "non-overlapping exceptions are merged" clause.
 func excludeCloudCoveredPolicies(crdPolicies, cloudPolicies []armotypes.VulnerabilityExceptionPolicy) []armotypes.VulnerabilityExceptionPolicy {
 	if len(cloudPolicies) == 0 || len(crdPolicies) == 0 {
 		return crdPolicies
@@ -351,18 +356,48 @@ func excludeCloudCoveredPolicies(crdPolicies, cloudPolicies []armotypes.Vulnerab
 
 	var filtered []armotypes.VulnerabilityExceptionPolicy
 	for _, p := range crdPolicies {
-		coveredByCloud := false
-		for _, vp := range p.VulnerabilityPolicies {
-			if len(cloudIndex.byName[strings.ToLower(vp.Name)]) > 0 {
-				coveredByCloud = true
-				break
-			}
-		}
-		if !coveredByCloud {
+		if !cloudCoversPolicy(p, cloudIndex) {
 			filtered = append(filtered, p)
 		}
 	}
 	return filtered
+}
+
+// cloudCoversPolicy reports whether some cloud policy sharing one of p's CVE/alias names also
+// covers every package p's own scope names. An unscoped cloud policy applies product-wide, so
+// it covers p regardless of p's own scope; a scoped cloud policy covers p only when p's scope
+// is fully contained in it. p being unscoped itself (applying product-wide) is never covered by
+// a scoped cloud policy, since the cloud policy only accounts for part of what p suppresses.
+func cloudCoversPolicy(p armotypes.VulnerabilityExceptionPolicy, cloudIndex *cveExceptionIndex) bool {
+	pScope, pScoped := policySubcomponents(p)
+	for _, vp := range p.VulnerabilityPolicies {
+		for _, i := range cloudIndex.byName[strings.ToLower(vp.Name)] {
+			cloudScope, cloudScoped := policySubcomponents(cloudIndex.srcCVEList[i])
+			if !cloudScoped {
+				return true
+			}
+			if pScoped && purlsCoveredBy(pScope, cloudScope) {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// purlsCoveredBy reports whether every purl in scope is matched by some purl in by, i.e.
+// scope's coverage is fully contained in by's. Reuses purlMatches with by's entries as the
+// pattern side and scope's as the instance side, the same direction scopedToSubcomponent
+// matches a finding's purl against an exception's stated scope.
+func purlsCoveredBy(scope, by []string) bool {
+	if len(scope) == 0 {
+		return false
+	}
+	for _, purl := range scope {
+		if !anyPURLMatches(by, purl) {
+			return false
+		}
+	}
+	return true
 }
 
 // ReportError reports the given error to the platform
--- a/adapters/v1/backend_test.go
+++ b/adapters/v1/backend_test.go
@@ -1175,6 +1175,61 @@ func TestExcludeCloudCoveredPolicies(t *testing.T) {
 	}
 }
 
+// A cloud and a CRD policy can independently scope to specific packages via the
+// Attributes["subcomponents"] purl list, the same mechanism ConvertToVulnerabilityExceptionPolicies
+// uses for a CRD exception's own vuln.Subcomponents field. Naming the same CVE is not the same
+// as covering the same package, so exclusion must also account for scope (#875).
+func TestExcludeCloudCoveredPolicies_SubcomponentScope(t *testing.T) {
+	policy := func(ruleName, cve string, subcomponents interface{}) armotypes.VulnerabilityExceptionPolicy {
+		p := armotypes.VulnerabilityExceptionPolicy{
+			PortalBase:            armotypes.PortalBase{Name: ruleName},
+			VulnerabilityPolicies: []armotypes.VulnerabilityPolicy{{Name: cve}},
+		}
+		if subcomponents != nil {
+			p.Attributes = map[string]interface{}{attrSubcomponents: subcomponents}
+		}
+		return p
+	}
+
+	tests := []struct {
+		name        string
+		crdPolicies []armotypes.VulnerabilityExceptionPolicy
+		cloud       []armotypes.VulnerabilityExceptionPolicy
+		wantLen     int
+	}{
+		{
+			name:        "disjoint packages: CRD policy scoped to a different package than cloud's survives",
+			crdPolicies: []armotypes.VulnerabilityExceptionPolicy{policy("crd-rule", "CVE-1", []string{"pkg:npm/bar"})},
+			cloud:       []armotypes.VulnerabilityExceptionPolicy{policy("cloud-rule", "CVE-1", []interface{}{"pkg:npm/foo"})},
+			wantLen:     1,
+		},
+		{
+			name:        "same package: CRD policy is dropped, matching #812's fix",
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Reviewed the fix. Verified `cloudCoversPolicy`/`purlsCoveredBy` against `policySubcomponents`/`purlMatches`/`anyPURLMatches` (subcomponent.go) and `cveExceptionIndex`/`buildCVEExceptionIndex` (backend_utils.go) on the PR branch — the match direction (cloud scope as pattern, CRD scope as instance) is consistent with `scopedToSubcomponent`'s existing suppression-time semantics, and the unscoped/scoped combinations are handled correctly:

- unscoped cloud policy → covers everything (preserves #812)
- scoped cloud fully containing scoped CRD → covered (dropped)
- scoped cloud not containing scoped CRD (disjoint or partial) → not covered (kept) — fixes #875
- scoped cloud vs unscoped CRD → never covered (kept)

Ran locally on the PR head commit (`47ff1c6`):
- `go build ./...` — passes
- `go vet ./adapters/...` — passes
- `go test ./adapters/... ./repositories/...` — all pass, including the new `TestExcludeCloudCoveredPolicies_SubcomponentScope` cases and all pre-existing `TestExcludeCloudCoveredPolicies`/`TestGetCVEExceptions_*` cases (regression-free for the unscoped case)

No blockers. Change is scoped tightly to `excludeCloudCoveredPolicies` + two new helpers, reuses existing purl-matching logic instead of duplicating it, and doc claims in the PR description match `docs/security-exception-design.md`'s "Identifier Bridging" / non-overlapping-exceptions clauses — this restores documented behavior rather than introducing new behavior, so no doc update is needed. All CI checks are green. Approving.

PR #879, opened 2026-08-27
fix(tools): reject a Retry-After value that would overflow time.Duration

### LINKED ISSUE #877: ParseRetryAfter overflows time.Duration on a large numeric value, turning the registry-pull retry into a hot loop
## Description

`internal/tools.ParseRetryAfter` parses a numeric `Retry-After` value out of an error string and converts it to a `time.Duration` with no bound on the input magnitude. A large-enough (but otherwise valid) numeric value overflows `time.Duration`'s underlying `int64` nanosecond count and silently wraps around to a **negative** duration — which `RetryWithBackoff`, the only caller, then hands straight to `time.After`. A negative or zero `time.After` duration fires immediately, so instead of backing off, the retry loop fires every attempt back-to-back with no delay: a hot loop against whatever is rate-limiting it.

This is the same failure class kubevuln already fixed once, in issue #754 ("a zero Retry-After turns the report retry into a hot loop") — but in a different, unpatched function. #754's fix landed in `adapters/v1/backend.go`'s `parseRetryAfterSeconds`, used only by the report-posting retry path. `internal/tools/retry.go`'s `ParseRetryAfter`, used by the registry-pull/SBOM-generation retry path, has no equivalent guard.

## Reproduction / Evidence

`internal/tools/retry.go:98`:

```go
if seconds, parseErr := strconv.Atoi(token); parseErr == nil && seconds > 0 {
	return time.Duration(seconds) * time.Second, true
}
```

`seconds` is a plain `int` (64-bit on the linux/amd64 target this ships for) with only a `> 0` check — no upper bound. I verified the overflow directly:

```go
token := "10000000000" // 10 billion seconds -- a header/error string containing "Retry-After: 10000000000"
seconds, _ := strconv.Atoi(token)   // 10000000000, no error, seconds > 0 is true
d := time.Duration(seconds) * time.Second
// d == -2346317h47m53.709551616s  (int64 ns: -8446744073709551616)
```

Running this exact snippet with `go run` confirms it: the multiplication overflows `int64` and produces a large *negative* `time.Duration`, with no panic and no error — `ParseRetryAfter` returns `(negativeDuration, true)` as if it were a perfectly valid delay.

That value then reaches `RetryWithBackoff` (`internal/tools/retry.go:153-160`) unfiltered:

```go
	delay = retryAfter
	if ceiling := config.retryAfterCeiling(); delay > ceiling {
		delay = ceiling
	}
}
```

A negative `delay` is never `> ceiling` (a positive value, 30s by default), so the configured cap never applies. Then at line 184:

```go
select {
case <-time.After(delay):
```

`time.After` with a negative duration fires immediately (documented Go behavior), so the "retry after" wait is skipped entirely.

## Root Cause

`ParseRetryAfter`'s numeric branch validates that the parsed value is positive but never bounds it against what fits in a `time.Duration` once multiplied by `time.Second`. The sibling function that already had this exact bug fixed (`adapters/v1/backend.go`'s `parseRetryAfterSeconds`, lines 191-196) guards it explicitly:

```go
const maxRetryAfterSeconds = math.MaxInt64 / int64(time.Second)
if seconds < 0 || seconds > maxRetryAfterSeconds {
	return 0, false
}
```

`ParseRetryAfter` was never given the same guard — it's an independent implementation, not a shared helper, so the fix for #754 never reached it.

## Impact

`internal/tools.RetryWithBackoff` backs the retry logic for both SBOM-generation code paths' registry interactions:

- `adapters/v1/syft.go` (in-process adapter): source resolution and SBOM generation retries
- `pkg/sbomscanner/v1/server.go` (sidecar scanner): source resolution and SBOM generation retries
- `core/services/scan.go`: SBOM generation retry

A registry (or any upstream whose error text is surfaced through this path) that returns an oversized `Retry-After` value — whether malformed, misconfigured, or from a compromised/malicious endpoint — turns what should be a bounded backoff into an unbounded hot loop, consuming CPU and hammering the same rate-limited endpoint on every scan worker hitting it, for the remainder of `config.MaxAttempts`. This is a real availability/self-inflicted-DoS risk in exactly the kind of external-input-triggered path #754 was filed to close off, just reachable through a different call site.

## Affected Components

- `internal/tools/retry.go` — `ParseRetryAfter`, `RetryWithBackoff`
- Callers relying on the guarantee `ParseRetryAfter` never returns a delay that skips backoff: `adapters/v1/syft.go`, `pkg/sbomscanner/v1/server.go`, `core/services/scan.go`

## Acceptance Criteria

- `ParseRetryAfter` never returns a duration whose computation could have overflowed: a numeric `Retry-After` value large enough that `seconds * time.Second` would exceed `math.MaxInt64` nanoseconds is rejected (treated as absent), the same way `adapters/v1/backend.go`'s `parseRetryAfterSeconds` already does.
- Regression test: a numeric `Retry-After` value at the overflow boundary and comfortably past it (e.g. `math.MaxInt64/int64(time.Second) + 1` and `10000000000`) returns `(0, false)` instead of a negative duration.
- Regression test on `RetryWithBackoff` (or an integration-level test) confirming a rate-limit error carrying such a value does not fire before `wait`/backoff would have elapsed.

## Suggested Direction

Add the same overflow guard `adapters/v1/backend.go`'s `parseRetryAfterSeconds` already uses to `ParseRetryAfter`'s numeric branch. Given both functions now need the identical bound, extracting one small shared helper (e.g. `func secondsToDuration(seconds int64) (time.Duration, bool)`) that both call would also remove the duplication between them — but a minimal, localized fix to `ParseRetryAfter` alone is enough to close the gap.

bhuvan-somisetty:
## Problem

`internal/tools.ParseRetryAfter` parses a numeric `Retry-After` value from error text and multiplies it by `time.Second`, with only a `> 0` check on the parsed value — no upper bound. A large-enough (but otherwise valid-looking) numeric value overflows `time.Duration`'s underlying `int64` nanosecond count and silently wraps around to a **negative** duration, with no error.

`RetryWithBackoff` (the only caller) hands that value straight to the ceiling check and then `time.After`. A negative delay is never `>` the positive ceiling, so the cap is skipped, and `time.After` with a negative duration fires immediately. Instead of backing off, the retry loop fires every attempt back-to-back with no delay — a hot loop against whatever is rate-limiting it.

## Root Cause

```go
if seconds, parseErr := strconv.Atoi(token); parseErr == nil && seconds > 0 {
	return time.Duration(seconds) * time.Second, true
}
```

This is the same failure class already fixed once, in #754 — but in a different function. That fix landed in `adapters/v1/backend.go`'s `parseRetryAfterSeconds`, used only by the report-posting retry path:

```go
const maxRetryAfterSeconds = math.MaxInt64 / int64(time.Second)
if seconds < 0 || seconds > maxRetryAfterSeconds {
	return 0, false
}
```

`ParseRetryAfter`, used by the registry-pull/SBOM-generation retry path, is an independent implementation and never received the equivalent guard.

I verified the overflow directly with `go run`:

```go
seconds, _ := strconv.Atoi("10000000000") // no error, seconds > 0
d := time.Duration(seconds) * time.Second
// d == -2346317h47m53.709551616s  (int64 ns: -8446744073709551616)
```

## Fix

`ParseRetryAfter` now rejects a value beyond `maxRetryAfterSeconds` (`math.MaxInt64/int64(time.Second)`), mirroring `backend.go`'s existing guard exactly, and switches from `strconv.Atoi` to `strconv.ParseInt(token, 10, 64)` so the comparison is between matching types. The change is confined to the one numeric branch of `ParseRetryAfter`.

## Testing

- `go build ./...` — passes
- `go vet ./internal/tools/...` — passes
- `go test ./internal/tools/... ./adapters/... ./core/services/... ./pkg/sbomscanner/...` (every package that calls `RetryWithBackoff`/`ParseRetryAfter`) — passes
- `golangci-lint run --new-from-rev=upstream/main ./internal/tools/...` — 0 issues

Added to `internal/tools/retry_test.go`:
- `TestParseRetryAfter_RejectsOverflowingValue` — the largest non-overflowing value still parses; one past the boundary and a large value (10000000000) are both rejected as `(0, false)` instead of wrapping.
- `TestRetryWithBackoff_OverflowingRetryAfterFallsBackToBackoff` — an end-to-end regression through `RetryWithBackoff`: an overflowing `Retry-After` now falls back to normal jittered backoff (asserted via elapsed time ≥ the configured backoff, which a hot loop from the wrapped negative duration would complete in well under a millisecond).

All pre-existing `TestParseRetryAfter`/`TestRetryWithBackoff_*` cases pass unchanged.

## Impact

`internal/tools.RetryWithBackoff` backs the retry logic for both SBOM-generation paths' registry interactions (`adapters/v1/syft.go`, `pkg/sbomscanner/v1/server.go`, `core/services/scan.go`). A malformed, misconfigured, or malicious `Retry-After` value can no longer turn a bounded backoff into an unbounded hot loop against whatever endpoint is rate-limiting the scan.

Fixes #877

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented excessively large retry delays from causing immediate retry loops.
  * Invalid or overflowing retry-delay values now fall back to the standard backoff behavior.

* **Tests**
  * Added coverage for retry-delay boundary values and overflow handling.

### DIFF
--- a/internal/tools/retry.go
+++ b/internal/tools/retry.go
@@ -3,6 +3,7 @@ package tools
 import (
 	"context"
 	"errors"
+	"math"
 	"math/rand"
 	"net/http"
 	"strconv"
@@ -15,6 +16,15 @@ import (
 	"github.com/kubescape/kubevuln/core/domain"
 	"github.com/kubescape/kubevuln/internal/metrics"
 )
+
+// maxRetryAfterSeconds is the largest number of seconds that can be multiplied by
+// time.Second without overflowing time.Duration's underlying int64 nanosecond count. A
+// numeric Retry-After value beyond this wraps around to a negative time.Duration -- which
+// RetryWithBackoff's ceiling check never catches (a negative delay is never ">" a positive
+// ceiling) and time.After fires immediately for, turning backoff into a hot retry loop
+// against whatever is rate-limiting it. Mirrors adapters/v1/backend.go's identical guard on
+// the report-posting retry path's own Retry-After parsing (see #754, #877).
+const maxRetryAfterSeconds = math.MaxInt64 / int64(time.Second)
 
 // RetryConfig specifies configuration for retrying operations.
 type RetryConfig struct {
@@ -95,7 +105,7 @@ func ParseRetryAfter(err error) (time.Duration, bool) {
 		fields := strings.Fields(sub)
 		if len(fields) > 0 {
 			token := strings.Trim(fields[0], ";,")
-			if seconds, parseErr := strconv.Atoi(token); parseErr == nil && seconds > 0 {
+			if seconds, parseErr := strconv.ParseInt(token, 10, 64); parseErr == nil && seconds > 0 && seconds <= maxRetryAfterSeconds {
 				return time.Duration(seconds) * time.Second, true
 			}
 		}
--- a/internal/tools/retry_test.go
+++ b/internal/tools/retry_test.go
@@ -84,6 +84,26 @@ func TestParseRetryAfter(t *testing.T) {
 	assert.Equal(t, time.Duration(0), nilErr)
 }
 
+// A numeric Retry-After large enough that seconds*time.Second overflows time.Duration's
+// int64 nanosecond range must be rejected, not silently wrapped into a negative duration
+// (#877). math.MaxInt64/int64(time.Second) is the largest value that does not overflow.
+func TestParseRetryAfter_RejectsOverflowingValue(t *testing.T) {
+	atBoundary := fmt.Sprintf("received status code 429 Retry-After: %d", maxRetryAfterSeconds)
+	dur, ok := ParseRetryAfter(errors.New(atBoundary))
+	assert.True(t, ok, "the largest non-overflowing value must still parse")
+	assert.Equal(t, time.Duration(maxRetryAfterSeconds)*time.Second, dur)
+
+	justOverBoundary := fmt.Sprintf("received status code 429 Retry-After: %d", maxRetryAfterSeconds+1)
+	dur, ok = ParseRetryAfter(errors.New(justOverBoundary))
+	assert.False(t, ok, "one past the boundary must be rejected, not wrapped")
+	assert.Equal(t, time.Duration(0), dur)
+
+	wayOverBoundary := errors.New("received status code 429 Retry-After: 10000000000")
+	dur, ok = ParseRetryAfter(wayOverBoundary)
+	assert.False(t, ok)
+	assert.Equal(t, time.Duration(0), dur)
+}
+
 func TestRetryWithBackoff_SuccessFirstAttempt(t *testing.T) {
 	calls := 0
 	config := RetryConfig{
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Reviewed the fix and tests.

**Correctness:** `maxRetryAfterSeconds = math.MaxInt64 / int64(time.Second)` mirrors the existing guard in `adapters/v1/backend.go`'s `parseRetryAfter` (added in #754) exactly, so the boundary math is proven correct against a prior instance of this same bug class. Switching `strconv.Atoi` → `strconv.ParseInt(token, 10, 64)` avoids a type mismatch in the new comparison. The `seconds > 0 && seconds <= maxRetryAfterSeconds` condition correctly rejects the overflow case while preserving the pre-existing `seconds > 0` semantics (no regression there).

**Tests:** `TestParseRetryAfter_RejectsOverflowingValue` covers the exact boundary (accepts `maxRetryAfterSeconds`, rejects `+1` and a large value) without wrapping. `TestRetryWithBackoff_OverflowingRetryAfterFallsBackToBackoff` is a solid end-to-end regression test proving the fallback to jittered backoff via elapsed-time assertion, rather than just asserting on the parser in isolation. All pre-existing tests are preserved.

**Scope:** Change is confined to the one numeric branch as described; the HTTP-date branches are unaffected and not a realistic overflow vector.

CI is green across the board and CodeRabbit found no actionable comments (the docstring-coverage warning is a soft nitpick against already-documented functions, not a real gap).

Issue #878, opened 2026-08-27
VEX document format is never checked before being trusted


Shreya2005-2005:
**What's happening**

`kubevuln` will use `github.com/openvex/go-vex`'s `Load`/`Parse` functions to read a fetched VEX document. These functions only check that the bytes are valid JSON - they don't check the JSON is actually a real, meaningful VEX document.

```go
vexDoc := &VEX{}
json.Unmarshal(data, vexDoc)
return vexDoc, nil
```

Nothing after this checks `vexDoc` actually has real content.

**Why this is a problem**

A real VEX document needs a `@context` field, at least one statement, and each statement needs a valid status and at least one product. Right now, something as empty as `{}` passes through `Load`/`Parse` with no error at all - it comes back as a "successfully parsed" document even though it says nothing.

**Why it hasn't caused issues yet**

Today, kubevuln only reads VEX documents it wrote itself, so they're always well-formed. This gap has never been hit because nothing external has been fetched yet.

**Simple proof**

```
INPUT: {}
Load() error -> nil   (WRONG - this is not a real VEX document)
```

**Why it matters now**

The External VEX Ingestion project (kubevuln#387) changes this. It will fetch a vendor's VEX feed from a URL kubevuln doesn't control (`VEXSource.spec.url`). A misconfigured or broken feed could return a login page, an error page, or a truncated file and today, none of that would be caught. It would be treated as a real document and saved.

---

PR #880, opened 2026-08-27
feat: add vexvalidate to check a fetched document is genuinely OpenVEX

Shreya2005-2005:
## Overview

`kubevuln` has no way to check that a fetched VEX document is actually real. `github.com/openvex/go-vex`'s `Load`/`Parse` functions only check the bytes are valid JSON - an empty or meaningless document like `{}` passes through with no error at all.

The External VEX Ingestion project (kubevuln#387) will fetch a vendor's VEX feed from a URL kubevuln doesn't control. A broken or misconfigured feed could return a login page, an error page, or a truncated file and today, none of that would be caught before being saved.

## Changes

Add `internal/vexvalidate`, a standalone `Validate` function that checks a document is real by confirming it has:

* A genuine `@context` (the real OpenVEX namespace, not just a string that happens to start with it)
* At least one statement
* Every statement has a valid status
* Every statement has at least one product

## Before / After

Before (go-vex's own `Parse`, no validation):

```
Parse error: <nil>
Returned document: &{Metadata:{Context: ID: Author: ...} Statements:[]}
(accepted as a "successfully parsed" document, even though it's a 404 error page)
```

After (`vexvalidate.Validate` on the same bytes):

```
Validate error: vexvalidate: document has no valid @context
```

## Testing

20 test cases in `vexvalidate_test.go`, covering a real valid document, missing/wrong/lookalike context, missing statements, invalid status, missing products, and a broken statement anywhere in a multi-statement list, not just the first one.

Full repo build (`go build ./...`) and `go test ./...` pass. 100% statement coverage on the new package.

## Related issues/PRs

* Resolved [VEX document format is never checked before being trusted #878](paste the real issue URL here)

```

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive validation for OpenVEX documents beyond basic JSON parsing.
  * Documents must include a valid context, at least one valid statement, recognized statuses, and identifiable products.
  * Supports bare and versioned OpenVEX contexts and all recognized statement statuses.

* **Bug Fixes**
  * Invalid, incomplete, malformed, or ambiguous OpenVEX documents are now rejected with specific validation errors.

### DIFF
--- a/internal/vexvalidate/vexvalidate.go
+++ b/internal/vexvalidate/vexvalidate.go
@@ -1,0 +1,124 @@
+// Package vexvalidate checks that a fetched OpenVEX document is genuinely
+// real and meaningful, not just syntactically valid JSON.
+//
+// github.com/openvex/go-vex's Load/Parse functions only decode JSON into the
+// VEX struct - they perform no validation of their own, so a document as
+// empty as {} is returned as a "successfully parsed" document with no error.
+// Nothing in kubevuln has needed to notice this so far, since it has only
+// ever read VEX documents it wrote itself, which are always well-formed.
+// The External VEX Ingestion project (kubevuln#387) changes that: it fetches
+// a vendor's VEX feed from a URL kubevuln does not control, so a
+// misconfigured or broken feed (a login page, an error page, a truncated
+// file) needs to be caught here, before it is ever persisted - not
+// discovered later by whatever eventually tries to consume it. See #878.
+package vexvalidate
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/openvex/go-vex/pkg/vex"
+)
+
+// Errors returned by Validate. Each identifies which specific requirement a
+// document failed, so a caller (or a test) can distinguish a malformed
+// context from an empty document from an invalid statement.
+var (
+	// ErrInvalidContext is returned when the document's @context is empty or
+	// does not identify it as an OpenVEX document.
+	ErrInvalidContext = errors.New("vexvalidate: document has no valid @context")
+
+	// ErrNoStatements is returned when the document has zero statements. A
+	// VEX document that asserts nothing is not useful to anything that
+	// would consume it.
+	ErrNoStatements = errors.New("vexvalidate: document has no statements")
+
+	// ErrInvalidStatement is returned when a statement fails go-vex's own,
+	// stricter Statement.Validate() - an invalid status, or a status paired
+	// with fields that status does not allow (e.g. not_affected with no
+	// justification or impact statement, or affected with no action
+	// statement).
+	ErrInvalidStatement = errors.New("vexvalidate: statement is invalid")
+
+	// ErrNoProducts is returned when a statement lists zero products.
+	ErrNoProducts = errors.New("vexvalidate: statement has no products")
+
+	// ErrEmptyProduct is returned when a statement's product identifies
+	// nothing at all - no @id, no hashes, no identifiers. go-vex's own
+	// Component type makes ID optional (it can also be identified by hashes
+	// or identifiers instead), so an empty {} still decodes successfully;
+	// only a product with none of the three says nothing about what it
+	// actually is.
+	ErrEmptyProduct = errors.New("vexvalidate: statement has a product that identifies nothing")
+)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: internal/vexvalidate/vexvalidate.go [hunk 1]
```diff
@@ -0,0 +1,101 @@
...
+		return ErrNoStatements
+	}
+
+	for i, s := range doc.Statements {
+		if !s.Status.Valid() {
+			return fmt.Errorf("%w: statement %d has status %q", ErrInvalidStatus, i, s.Status)
+		}
+		if len(s.Products) == 0 {
+			return fmt.Errorf("%w: statement %d", ErrNoProducts, i)
+		}
```
matthyx, line 114: A `Product` with no identifying data at all (e.g. `"products": [{}]` — no `@id`, no `hashes`, no `identifiers`) currently passes this check, since it only requires the slice to be non-empty. `go-vex`'s own `Component` type makes `ID` optional (`json:"@id,omitempty"`), so an empty product object is valid JSON but identifies nothing.

That's exactly the "broken feed" scenario this package is meant to catch (per the PR description) — a malformed document could carry a statement with a placeholder/empty product and still be accepted. Consider requiring each product to have at least one of `ID`, `Hashes`, or `Identifiers` set.

internal/vexvalidate/vexvalidate.go
```diff
@@ -0,0 +1,101 @@
...
+	if !hasValidContext(doc.Context) {
+		return ErrInvalidContext
+	}
+
+	if len(doc.Statements) == 0 {
+		return ErrNoStatements
+	}
+
+	for i, s := range doc.Statements {
+		if !s.Status.Valid() {
```
matthyx: This only checks `s.Status.Valid()`, but `go-vex` already vendors a much stricter `(*vex.Statement).Validate()` that this package could call instead: it enforces that `not_affected` carries a justification (from the valid enum) or impact statement, that `affected` carries an action statement, and that the other statuses don't carry fields that don't apply to them.

As written, a statement like `{"status": "affected", "products": [...]}` with no `action_statement` passes `vexvalidate.Validate`, even though the spec requires one and the library's own `Validate()` would reject it. Since the goal here is to catch documents that are syntactically fine but semantically empty/broken, I'd lean on `stmt.Validate()` rather than re-implementing a subset of it.

matthyx (changes requested):
Nice test coverage and a real, well-scoped problem to fix. Two gaps in `Validate` undercut the goal stated in the PR description though (left inline):

- An empty/placeholder `Product` (`{}`, no `@id`/`hashes`/`identifiers`) passes validation since only the slice length is checked.
- Statement validity only checks `Status.Valid()`, not the fuller status-specific requirements (`affected` needs an action statement, `not_affected` needs a justification, etc.) that `go-vex`'s own `Statement.Validate()` already enforces.

Both mean a broken/lookalike feed with syntactically valid but semantically empty statements would still pass — which is the exact failure mode this package exists to catch.

Also note the **DCO check is currently failing** (action_required) on the head commit, and PR mergeability is reported as blocked — that'll need a signed-off commit before this can merge regardless of the above.

Holding off on approval until these are addressed.

Shreya2005-2005:
@matthyx Fixed both  

-Empty placeholder products (no @id/hashes/identifiers) are now rejected.

-Statement validity now goes through go-vex's own Statement.Validate() instead of a plain status check, so not_affected requires a justification/impact and affected requires an action statement, per the spec.

matthyx (approved):
Both blockers from my previous review are fixed:

- Empty/placeholder products (no `@id`, `hashes`, or `identifiers`) are now rejected via `productIsEmpty`.
- Statement validity now delegates to `go-vex`'s own `(*Statement).Validate()`, so status-specific requirements (`affected` needs an action statement, `not_affected` needs a justification/impact statement, etc.) are enforced instead of just `Status.Valid()`.

Both have direct test coverage, and DCO/CI are all green now. CodeRabbit's remaining nitpick (test fixture uses the `sha256` map key instead of the canonical `sha-256` label) is test-only — `productIsEmpty` just checks the `Hashes` map is non-empty, so it doesn't affect correctness of the shipped code.

Approving.

PR #884, opened 2026-08-27
fix(adapters): don't let the docker.io/index.docker.io alias override a more specific entry

### LINKED ISSUE #883: rewriteImageRef's docker.io/index.docker.io alias fallback ignores a more specific, directly-matching key
## Description

`rewriteImageRef` (`adapters/v1/syft.go`) picks a `proxyRegistryMap` entry to rewrite a scanned image reference through, using longest-key-first precedence — `TestRewriteImageRef`'s own "longest prefix wins over shorter prefix" case establishes that as the function's contract. The `docker.io`/`index.docker.io` cross-alias special case breaks that contract: when both keys are configured with **different** proxy values, an image reference that literally starts with `docker.io/` is silently rewritten using the `index.docker.io` entry's proxy instead of `docker.io`'s own, explicitly-configured one — because `index.docker.io` (16 chars) sorts before `docker.io` (9 chars) and its alias fallback matches first, before the loop ever reaches the `docker.io` key that is the image's actual, literal, more specific match.

## Reproduction / Evidence

```go
proxyMap := map[string]string{
    "docker.io":       "mirror-a.internal",
    "index.docker.io": "mirror-b.internal",
}
rewriteImageRef("docker.io/library/nginx:latest", proxyMap)
```

Ran this directly:

```
imageRef:  docker.io/library/nginx:latest
rewritten: mirror-b.internal/library/nginx:latest
expected (docker.io's own explicit config): mirror-a.internal/library/nginx:latest
```

`adapters/v1/syft.go`:

```go
sort.Slice(keys, func(i, j int) bool { return len(keys[i]) > len(keys[j]) })
for _, original := range keys {
    proxy := strings.TrimRight(proxyMap[original], "/")
    if proxy == "" {
        continue
    }
    if strings.HasPrefix(imageRef, original+"/") {
        return proxy + imageRef[len(original):]
    }
    // treat docker.io and index.docker.io as the same registry
    if original == "docker.io" && strings.HasPrefix(imageRef, "index.docker.io/") {
        return proxy + imageRef[len("index.docker.io"):]
    }
    if original == "index.docker.io" && strings.HasPrefix(imageRef, "docker.io/") {
        return proxy + imageRef[len("docker.io"):]
    }
}
```

`index.docker.io` is longer, so it's tried first. Its own direct-prefix check fails (`imageRef` doesn't start with `index.docker.io/`), but the alias fallback on the next line (`original == "index.docker.io" && strings.HasPrefix(imageRef, "docker.io/")`) succeeds and returns immediately — the loop never reaches the `docker.io` key, whose direct-prefix check would have matched correctly.

`TestRewriteImageRef` (`adapters/v1/syft_test.go:123-207`) already tests `docker.io`-only and `index.docker.io`-only configurations separately (lines 155-171), and separately tests that the longest/most-specific key wins when two *unrelated* keys overlap (lines 184-191, `"docker.io"` vs `"docker.io/library"`) — but no case configures both `docker.io` and `index.docker.io` together, which is exactly the combination that breaks the "most specific match wins" guarantee the other test establishes.

## Root Cause

The alias fallback for `original == "index.docker.io"` matches on `imageRef`'s content (`strings.HasPrefix(imageRef, "docker.io/")`) rather than on whether a `docker.io` entry actually exists and would otherwise apply. It never checks whether the *other* key of the alias pair is present in `proxyMap` with its own value, so when both are configured it always wins by sort-order coincidence rather than by which one the image reference actually, literally names.

## Impact

An operator who configures registry mirroring for both spellings of Docker Hub with different destinations — e.g. staging traffic to one mirror by whichever domain form a workload happens to be deployed with, or migrating from one mirror to another and setting both during the transition — has images that literally start with `docker.io/` silently routed to the `index.docker.io` mirror instead, with no error or log indicating the `docker.io` entry was never consulted. Given `docker.io` is the default, ubiquitous registry, this is the most likely of the special-cased pair to be misrouted in a real multi-mirror configuration.

## Affected Components

- `adapters/v1/syft.go` — `rewriteImageRef`

## Acceptance Criteria

- When `imageRef` literally starts with a configured key's own prefix (e.g. `docker.io/...` with a `docker.io` entry present), that entry's proxy is used — even if a differently-spelled alias (`index.docker.io`) is also configured and sorts first by length.
- The alias fallback continues to work exactly as before when only one of the two keys is configured.
- Regression test: both `docker.io` and `index.docker.io` configured with different proxy values; a `docker.io/...` ref uses `docker.io`'s proxy, and an `index.docker.io/...` ref uses `index.docker.io`'s proxy.

## Suggested Direction

Before falling back to the alias match for `original`, skip it if `proxyMap` also contains the other spelling of the pair (i.e., the `index.docker.io` alias branch should not fire when `proxyMap["docker.io"]` is set and non-empty, and vice versa) — letting the loop continue to the entry that's the image's actual, literal match. This keeps the fix localized to the two alias-fallback conditions.

bhuvan-somisetty:
## Problem

`rewriteImageRef` picks a `proxyRegistryMap` entry longest-key-first, so the most specific configured entry wins — `TestRewriteImageRef`'s own "longest prefix wins over shorter prefix" case establishes that as the function's contract. The `docker.io`/`index.docker.io` alias fallback breaks it: when both keys are configured with *different* proxy values, `index.docker.io` (16 chars) always sorts before `docker.io` (9 chars), so its alias branch matches a `docker.io/...` reference and returns before the loop ever reaches the `docker.io` key — the image's own, literal, more specific match.

## Root Cause

```go
sort.Slice(keys, func(i, j int) bool { return len(keys[i]) > len(keys[j]) })
for _, original := range keys {
    ...
    if original == "docker.io" && strings.HasPrefix(imageRef, "index.docker.io/") {
        return proxy + imageRef[len("index.docker.io"):]
    }
    if original == "index.docker.io" && strings.HasPrefix(imageRef, "docker.io/") {
        return proxy + imageRef[len("docker.io"):]
    }
}
```

With `proxyMap = {"docker.io": "mirror-a", "index.docker.io": "mirror-b"}` and `imageRef = "docker.io/library/nginx:latest"`, the loop visits `index.docker.io` first (longer key), its direct-prefix check fails, but the alias fallback (`original == "index.docker.io" && HasPrefix(imageRef, "docker.io/")`) succeeds and returns `mirror-b/...` — `docker.io`'s own explicitly configured `mirror-a` is never consulted. Confirmed by running the function directly:

```
rewritten: mirror-b.internal/library/nginx:latest
expected (docker.io's own explicit config): mirror-a.internal/library/nginx:latest
```

`TestRewriteImageRef` already covers `docker.io`-only and `index.docker.io`-only configurations separately, and separately covers longest-prefix-wins between two *unrelated* keys — but no case configures both `docker.io` and `index.docker.io` together, which is exactly the combination that breaks the "most specific match wins" guarantee.

## Fix

Added `hasUsableProxy`, and gated each alias branch on the *other* spelling not having a usable (non-empty) entry of its own:

```go
if original == "docker.io" && strings.HasPrefix(imageRef, "index.docker.io/") && !hasUsableProxy(proxyMap, "index.docker.io") {
    return proxy + imageRef[len("index.docker.io"):]
}
if original == "index.docker.io" && strings.HasPrefix(imageRef, "docker.io/") && !hasUsableProxy(proxyMap, "docker.io") {
    return proxy + imageRef[len("docker.io"):]
}
```

When only one of the two keys is configured, the alias fallback fires exactly as before (the other key's `hasUsableProxy` check is false, so the gate is a no-op). When both are configured, the loop now correctly continues past the alias's own iteration and reaches the directly-matching key on a later iteration.

## Testing

- `go build ./...` — passes
- `go vet ./adapters/v1/...` — passes
- `go test ./adapters/v1/...` — passes (confirmed clean on a sequential rerun; a `-p`-parallel run intermittently fails an unrelated timing test, `Test_syftAdapter_CreateSBOM_SerializesWithAbandonedGoroutineAfterTimeout`, due to Windows temp-file lock contention — reproduced the identical flake on a clean `upstream/main` checkout with no code changes, confirming it's pre-existing and unrelated to this change)
- `golangci-lint run --new-from-rev=upstream/main ./adapters/v1/...` — 0 issues

Added three cases to `TestRewriteImageRef`:
- both `docker.io` and `index.docker.io` configured with different proxies: a `docker.io/...` ref uses `docker.io`'s own entry (the bug's direct repro)
- the symmetric case: an `index.docker.io/...` ref uses `index.docker.io`'s own entry
- `index.docker.io` present but empty doesn't block the `docker.io` alias fallback, confirming the existing single-key behavior is preserved

All pre-existing cases pass unchanged.

## Impact

An operator configuring registry mirroring for both spellings of Docker Hub with different destinations no longer has `docker.io/...` references silently routed through the wrong mirror.

Fixes #883

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved container image proxy selection when both `docker.io` and `index.docker.io` configurations are present.
  * Exact registry configurations now take precedence over aliases.
  * Empty proxy entries no longer prevent valid alias-based proxy fallback.

### DIFF
--- a/adapters/v1/syft.go
+++ b/adapters/v1/syft.go
@@ -80,15 +80,25 @@ func rewriteImageRef(imageRef string, proxyMap map[string]string) string {
 		if strings.HasPrefix(imageRef, original+"/") {
 			return proxy + imageRef[len(original):]
 		}
-		// treat docker.io and index.docker.io as the same registry
-		if original == "docker.io" && strings.HasPrefix(imageRef, "index.docker.io/") {
+		// treat docker.io and index.docker.io as the same registry, but only when the
+		// spelling imageRef actually uses has no entry of its own: a configured entry for
+		// that exact spelling is a more specific match (the one "longest prefix wins"
+		// above already prefers) and must not be skipped in favor of the alias's proxy
+		// just because the alias happens to sort first by length (#883).
+		if original == "docker.io" && strings.HasPrefix(imageRef, "index.docker.io/") && !hasUsableProxy(proxyMap, "index.docker.io") {
 			return proxy + imageRef[len("index.docker.io"):]
 		}
-		if original == "index.docker.io" && strings.HasPrefix(imageRef, "docker.io/") {
+		if original == "index.docker.io" && strings.HasPrefix(imageRef, "docker.io/") && !hasUsableProxy(proxyMap, "docker.io") {
 			return proxy + imageRef[len("docker.io"):]
 		}
 	}
 	return imageRef
+}
+
+// hasUsableProxy reports whether proxyMap has a non-empty (after trimming any trailing
+// slash) proxy value configured for key.
+func hasUsableProxy(proxyMap map[string]string, key string) bool {
+	return strings.TrimRight(proxyMap[key], "/") != ""
 }
 
 func NormalizeImageID(imageID, imageTag string) string {
--- a/adapters/v1/syft_test.go
+++ b/adapters/v1/syft_test.go
@@ -199,6 +199,38 @@ func TestRewriteImageRef(t *testing.T) {
 			},
 			want: "quay-mirror.example.com/kubescape/kubevuln:latest",
 		},
+		{
+			// #883: index.docker.io sorts before docker.io by length, so the alias
+			// fallback used to win even though docker.io/... is docker.io's own,
+			// more specific, direct match.
+			name:     "both docker.io and index.docker.io configured: docker.io ref uses its own entry",
+			imageRef: "docker.io/library/nginx:latest",
+			proxyMap: map[string]string{
+				"docker.io":       "mirror-a.example.com",
+				"index.docker.io": "mirror-b.example.com",
+			},
+			want: "mirror-a.example.com/library/nginx:latest",
+		},
+		{
+			name:     "both docker.io and index.docker.io configured: index.docker.io ref uses its own entry",
+			imageRef: "index.docker.io/library/nginx:latest",
+			proxyMap: map[string]string{
+				"docker.io":       "mirror-a.example.com",
+				"index.docker.io": "mirror-b.example.com",
+			},
+			want: "mirror-b.example.com/library/nginx:latest",
+		},
+		{
+			// The alias must still apply when only one of the pair has a usable value:
+			// an empty index.docker.io entry does not block docker.io's alias fallback.
+			name:     "index.docker.io entry present but empty does not block the docker.io alias fallback",
+			imageRef: "index.docker.io/library/alpine:latest",
+			proxyMap: map[string]string{
+				"docker.io":       "mirror.example.com",
+				"index.docker.io": "",
+			},
+			want: "mirror.example.com/library/alpine:latest",
+		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {


matthyx (approved):
Reviewed the fix and traced through the alias-gating logic against all three added test cases plus the pre-existing single-key cases — matches expected output in every case:

- Both `docker.io`/`index.docker.io` configured with different proxies: the longer key (`index.docker.io`) is visited first by the length-sort, but `hasUsableProxy` now correctly defers to the other key's own direct match instead of firing the alias branch prematurely.
- Only one of the two keys configured: the alias fallback still fires exactly as before (`hasUsableProxy` on the *other* key is false, so the gate is a no-op) — no regression for the common single-key setup.
- Key present but empty (`""`): correctly treated as "not usable," so it doesn't block the alias fallback for the other spelling.

The general "longest prefix wins" branch for unrelated keys is untouched, so no regression there either. Root cause writeup and comments are clear, `hasUsableProxy` is a small readable helper, and the new test cases target exactly the scenario from #883. All CI checks are green.

No blockers — approving.

PR #891, opened 2026-08-27
docs(config): match the duration schema pattern to what LoadConfig actually accepts

### LINKED ISSUE #890: CONFIGURATION.md's duration schema pattern rejects config values LoadConfig actually accepts
## Description

`docs/CONFIGURATION.md`'s "Complete Schema" JSON Schema block validates `scanTimeout`, `scannerReadinessTimeout`, and `shutdownTimeout` against the pattern `^[0-9]+(s|m|h)$` (`shutdownTimeout` additionally allows a leading `-`). That pattern only accepts a single integer followed by exactly one unit character. The actual config loader, `config.LoadConfig` (`config/config.go`), decodes these fields as `time.Duration` via viper, which uses Go's standard `time.ParseDuration` grammar — a strictly larger language that also accepts compound durations (`1h30m`), fractional values (`1.5h`), and sub-second units (`500ms`, `1.5s`). A value that's perfectly valid to the real config loader is flagged as schema-invalid by the documented schema.

## Reproduction / Evidence

```go
pattern := regexp.MustCompile(`^[0-9]+(s|m|h)$`)
for _, v := range []string{"5m", "1h30m", "1.5h", "500ms", "90s"} {
    d, err := time.ParseDuration(v)
    fmt.Printf("%-8s schema-pattern-matches=%-5v time.ParseDuration-ok=%-5v parsed=%v\n",
        v, pattern.MatchString(v), err == nil, d)
}
```

```
5m       schema-pattern-matches=true   time.ParseDuration-ok=true  parsed=5m0s
1h30m    schema-pattern-matches=false  time.ParseDuration-ok=true  parsed=1h30m0s
1.5h     schema-pattern-matches=false  time.ParseDuration-ok=true  parsed=1h30m0s
500ms    schema-pattern-matches=false  time.ParseDuration-ok=true  parsed=500ms
90s      schema-pattern-matches=true   time.ParseDuration-ok=true  parsed=1m30s
```

I also confirmed this against the actual production code path, not just `time.ParseDuration` in isolation — a temporary test calling `config.LoadConfig` directly with a `clusterData.json` containing `"scanTimeout": "1h30m"`:

```
LoadConfig successfully parsed scanTimeout="1h30m" as 1h30m0s, which docs/CONFIGURATION.md's
schema pattern ^[0-9]+(s|m|h)$ would reject as invalid
```

`LoadConfig` returns no error and correctly sets `cfg.ScanTimeout` to `1h30m0s` — confirming the schema is stricter than the code it's meant to describe.

## Root Cause

`docs/CONFIGURATION.md:281,286,291`:

```json
"scanTimeout": {
  "type": "string",
  "default": "5m",
  "pattern": "^[0-9]+(s|m|h)$"
},
"scannerReadinessTimeout": {
  "type": "string",
  "default": "60s",
  "pattern": "^[0-9]+(s|m|h)$"
},
"shutdownTimeout": {
  "type": "string",
  "default": "20s",
  "pattern": "^-?[0-9]+(s|m|h)$"
}
```

These patterns were written to match the simple examples shown elsewhere in the docs (`"5m"`, `"60s"`, `"20s"`), but the field's actual accepted grammar is whatever `time.ParseDuration` accepts — unit combinations, fractional values, and additional units (`ns`, `us`/`µs`, `ms`) the pattern doesn't even list one of.

## Impact

This schema block is presented as the authoritative "Complete Schema" for `clusterData.json` and is the kind of artifact an operator would plug into IDE JSON Schema validation (e.g. VS Code's `$schema` support) or a CI config-lint step. Anyone doing so is told that writing `"scannerReadinessTimeout": "1m30s"` or `"shutdownTimeout": "1.5m"` — both natural, more-readable ways to express those values, and both accepted by the real running code — is invalid, and would either have to convert to a less readable single-unit-only form or have their config wrongly rejected by tooling that trusts this schema.

## Affected Components

- `docs/CONFIGURATION.md` — the `scanTimeout`, `scannerReadinessTimeout`, and `shutdownTimeout` schema entries under "Complete Schema"

## Acceptance Criteria

- The schema pattern for all three duration fields accepts every string `config.LoadConfig` actually accepts for that field (via `time.ParseDuration`), including compound (`1h30m`), fractional (`1.5h`), and the full set of Go duration units (`ns`, `us`/`µs`, `ms`, `s`, `m`, `h`).
- `shutdownTimeout` continues to allow a leading `-` (documented elsewhere as meaning "expire immediately"), consistent with today's pattern.
- The fix is verifiable: a small script or test comparing the corrected pattern's matches against `time.ParseDuration`'s acceptance across a representative set of values (including the ones in this issue) agrees on every case.

## Suggested Direction

Replace the three patterns with one that mirrors Go's actual duration grammar — a sequence of one-or-more `(decimal number)(unit)` pairs, where unit is one of `ns|us|µs|ms|s|m|h`, e.g. `^-?([0-9]+(\.[0-9]+)?(ns|us|µs|ms|s|m|h))+$` (retaining `shutdownTimeout`'s leading-`-` allowance) — narrow enough to still catch a genuinely malformed value, but no longer rejecting anything Go's own duration parser accepts.

bhuvan-somisetty:
## Problem

`docs/CONFIGURATION.md`'s "Complete Schema" validates `scanTimeout`, `scannerReadinessTimeout`, and `shutdownTimeout` against `^[0-9]+(s|m|h)$` (`shutdownTimeout` additionally allowing a leading `-`). That only accepts a single integer plus one unit character. The real config loader (`config.LoadConfig`) decodes these fields as `time.Duration` via viper, backed by `time.ParseDuration` — a strictly larger grammar that also accepts compound durations (`1h30m`), fractional values (`1.5h`), and units the pattern doesn't even list (`ns`, `us`/`µs`, `ms`). A value the real loader accepts fine is flagged as schema-invalid by the documented schema.

## Root Cause

The patterns were written to match the simple single-unit examples shown elsewhere in the docs (`"5m"`, `"60s"`, `"20s"`), not the field's actual accepted grammar.

## Fix

```diff
-"pattern": "^[0-9]+(s|m|h)$"
+"pattern": "^([0-9]+(\\.[0-9]*)?(ns|us|µs|ms|s|m|h))+$"
```
(and the `shutdownTimeout` variant keeps its `-?` prefix)

One or more `number[.fraction]unit` groups, covering every unit `time.ParseDuration` recognizes — matches Go's actual duration grammar rather than the ad-hoc single-unit subset the old pattern covered.

## Testing

This is a documentation-only change (the JSON Schema block isn't runtime-enforced by kubevuln itself), so there's no existing test harness in the repo to hook a regression test into. I verified it the same way I found the bug:

- **Against the real code path**: a temporary test called `config.LoadConfig` directly with a `clusterData.json` containing `"scanTimeout": "1h30m"` — parsed successfully to `1h30m0s` with no error, confirming the loader accepts what the old pattern rejected. Removed before committing.
- **Against the corrected pattern**: compared the new pattern's matches to `time.ParseDuration`'s actual acceptance across 16 values it should accept (`5m`, `1h30m`, `1.5h`, `500ms`, `2h45m`, `300ns`, `10us`, `10µs`, `1h2m3s`, `5.m` — an odd but real Go-accepted edge case, and others) and 8 it should reject (`""`, `5x`, `abc`, `5`, `h5`, `5 m`, `5mm`, `m5`), plus the three `shutdownTimeout` sign cases — full agreement on every case.
- `go build ./...` — passes (no Go source touched, sanity-checked anyway)

## Impact

This schema is presented as the authoritative reference for `clusterData.json` and is the kind of artifact an operator would plug into IDE JSON Schema validation or a CI config-lint step. Anyone doing so was told a natural, more-readable value like `"scannerReadinessTimeout": "1m30s"` — accepted fine by the real running code — was invalid.

Fixes #890

## Summary by CodeRabbit

* **Documentation**
  * Updated configuration guidance to support fractional timeout values.
  * Added sub-second duration units, including nanoseconds, microseconds, and milliseconds.
  * Retained support for negative shutdown timeout values.

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -278,17 +278,17 @@ The main configuration file. All options can be overridden via environment varia
     "scanTimeout": {
       "type": "string",
       "default": "5m",
-      "pattern": "^[0-9]+(s|m|h)$"
+      "pattern": "^([0-9]+(\\.[0-9]*)?(ns|us|µs|ms|s|m|h))+$"
     },
     "scannerReadinessTimeout": {
       "type": "string",
       "default": "60s",
-      "pattern": "^[0-9]+(s|m|h)$"
+      "pattern": "^([0-9]+(\\.[0-9]*)?(ns|us|µs|ms|s|m|h))+$"
     },
     "shutdownTimeout": {
       "type": "string",
       "default": "20s",
-      "pattern": "^-?[0-9]+(s|m|h)$"
+      "pattern": "^-?([0-9]+(\\.[0-9]*)?(ns|us|µs|ms|s|m|h))+$"
     },
     "storage": {
       "type": "boolean",
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -278,17 +278,18 @@ The main configuration file. All options can be overridden via environment varia
     "scanTimeout": {
       "type": "string",
       "default": "5m",
-      "pattern": "^([0-9]+(\\.[0-9]*)?(ns|us|µs|ms|s|m|h))+$"
+      "description": "Must be positive: LoadConfig rejects a value <= 0, so unlike the other two duration fields this pattern accepts no sign and no bare 0.",
+      "pattern": "^((?:[0-9]+\\.?[0-9]*|\\.[0-9]+)(ns|us|µs|ms|s|m|h))+$"
     },
     "scannerReadinessTimeout": {
       "type": "string",
       "default": "60s",
-      "pattern": "^([0-9]+(\\.[0-9]*)?(ns|us|µs|ms|s|m|h))+$"
+      "pattern": "^[+-]?(0|((?:[0-9]+\\.?[0-9]*|\\.[0-9]+)(ns|us|µs|ms|s|m|h))+)$"
     },
     "shutdownTimeout": {
       "type": "string",
       "default": "20s",
-      "pattern": "^-?([0-9]+(\\.[0-9]*)?(ns|us|µs|ms|s|m|h))+$"
+      "pattern": "^[+-]?(0|((?:[0-9]+\\.?[0-9]*|\\.[0-9]+)(ns|us|µs|ms|s|m|h))+)$"
     },
     "storage": {
       "type": "boolean",


docs/CONFIGURATION.md
```diff
@@ -278,17 +278,18 @@ The main configuration file. All options can be overridden via environment varia
     "scanTimeout": {
       "type": "string",
       "default": "5m",
-      "pattern": "^[0-9]+(s|m|h)$"
+      "description": "Must be positive: LoadConfig rejects a value <= 0, so unlike the other two duration fields this pattern accepts no sign and no bare 0.",
+      "pattern": "^((?:[0-9]+\\.?[0-9]*|\\.[0-9]+)(ns|us|µs|ms|s|m|h))+$"
```
matthyx, line 282: Non-blocking nit: this pattern has no `[+-]?` prefix at all, so it rejects `"+5m"`. But `LoadConfig` would accept `"+5m"` fine (the sign is consumed, and `+5m` is still `> 0`), so this is very slightly stricter than "what LoadConfig actually accepts." Not worth blocking on — a leading `+` on a duration string is not something anyone writes in practice — but if you want the schema to be exactly permissive-equivalent to `LoadConfig`, `^\+?(...)+$` would close the gap. (`-` correctly stays excluded since a negative value fails the `> 0` check.)

matthyx (approved):
Verified the corrected patterns against `time.ParseDuration`'s real grammar (sign handling, `.5s`/`5.s` fraction forms, all six units, the bare-`0` special case) and against `config.LoadConfig`'s extra `ScanTimeout > 0` runtime check — everything lines up, including the deliberate difference between `scanTimeout`'s pattern (no sign/bare-zero) and the other two fields' (signed + bare-zero allowed, since 0-or-negative is meaningful for both per their docs). This already went through a CodeRabbit round that caught the same signed/bare-zero gap, and the follow-up commit addressed it correctly.

Left one non-blocking nit inline about `scanTimeout` still rejecting an explicit `+` sign. Doc-only change, no runtime impact. Approving.

PR #895, opened 2026-08-28
fix(sbomscanner): return syftToDomain's error instead of a nil the caller dereferences

### LINKED ISSUE #893: the sidecar's syftToDomain drops the error the in-process copy returns, and the caller dereferences the nil
The sidecar's `syftToDomain` is documented as "the same logic as `SyftAdapter.syftToDomain`". It is not: the in-process copy returns its error, this one drops both and returns `nil`.

```go
func syftToDomain(sbomSBOM sbom.SBOM) *v1beta1.SyftDocument {
	doc := syftjson.ToFormatModel(sbomSBOM, ...)

	b, err := json.Marshal(doc)
	if err != nil {
		return nil          // in-process: return nil, err
	}
	var syftDoc *v1beta1.SyftDocument
	if err := json.Unmarshal(b, &syftDoc); err != nil {
		return nil          // in-process: return nil, err
	}
	...
}
```

The caller does not expect nil:

```go
doc := syftToDomain(*syftSBOM)
docBytes, err := json.Marshal(doc)      // marshals a nil pointer to "null", no error
...
logger.L().Info("SBOM scan completed",
	helpers.Int("packages", len(doc.Artifacts)))   // nil dereference
```

`json.Marshal` of a nil pointer succeeds, so the nil survives to `len(doc.Artifacts)` and panics there.

The sidecar's gRPC server is created with `grpc.NewServer(MaxRecvMsgSize, MaxSendMsgSize)` and no recovery interceptor, and grpc-go does not recover handler panics by default, so this takes the process down rather than failing the one scan.

Reproduced against current main. A package whose metadata carries a NaN float is enough to make `json.Marshal` fail:

```
syftToDomain returned nil? true
call site does len(doc.Artifacts) at server.go:327 -> nil deref
PANIC reproduced: runtime error: invalid memory address or nil pointer dereference
```

On reachability, so this is not overstated: package metadata comes from cataloguing a scanned image, and I have not shown a specific real image that produces an unmarshalable value. What is demonstrated is the failure mode, that any marshal or unmarshal failure here crashes the sidecar instead of returning an error. The in-process path returns an error for the same input, which is the behaviour this one claims to share.

`syftToDomain` also has no test at all, which is presumably why the divergence went unnoticed. `pkg/sbomscanner/v1` is at 71.3%, and this function is the largest single uncovered block in it.

Returning the error, matching the in-process copy, and adding tests for both outcomes.

AnshulPatil2005:
## Overview

The sidecar's `syftToDomain` says it is "the same logic as `SyftAdapter.syftToDomain`". It was not: the in-process copy returns its error, this one dropped both and returned `nil`.

The caller does not expect nil, and `json.Marshal` of a nil pointer succeeds (producing `"null"`, no error), so the nil survived to `len(doc.Artifacts)` in the completion log and panicked there.

The gRPC server is built with no recovery interceptor, and grpc-go does not recover handler panics, so that takes the sidecar process down rather than failing the one scan.

## Reproduction

Against current main, a package whose metadata carries a NaN float is enough to make `json.Marshal` fail:

```
syftToDomain returned nil? true
call site does len(doc.Artifacts) at server.go:327 -> nil deref
PANIC reproduced: runtime error: invalid memory address or nil pointer dereference
```

Stated plainly so it is not overread: package metadata comes from cataloguing a scanned image, and I have not shown a specific real image that produces an unmarshalable value. What is demonstrated is the failure mode, that any marshal or unmarshal failure here crashes the sidecar rather than returning an error, while the in-process path returns one for the same input.

## The change

`syftToDomain` returns `(*v1beta1.SyftDocument, error)` like its in-process counterpart, and `CreateSBOM` surfaces it as a gRPC error.

No explicit `endActiveTempDirUse()` at the new return: the defer covering it lives inside the `dl.Run` closure, which has already returned by this point. The existing serialize-error return immediately below does the same, so this matches the surrounding code rather than adding a call that the `sync.Once` would swallow anyway.

## Testing

`go build ./... && go vet ./... && go test ./pkg/sbomscanner/v1/ -count=1`

`syftToDomain` had no test at all, which is presumably why the divergence survived. Both outcomes are now covered, and reverting the error return to `nil, nil` fails `TestSyftToDomain_ReportsMarshalFailure`.

Coverage for the package goes 71.3% to 75.3%; `syftToDomain` 0% to 88.9%. It was the largest single uncovered block in the package.

## Related issues/PRs:

* Resolved #893

## Summary by CodeRabbit

- **Bug Fixes**
  - SBOM creation now reports serialization and deserialization failures instead of silently returning an empty result.
  - Conversion errors are propagated with clearer error context.

- **Tests**
  - Added coverage for SBOM conversion error scenarios.
  - Simplified test server and mock response handling.

### DIFF
--- a/pkg/sbomscanner/v1/server.go
+++ b/pkg/sbomscanner/v1/server.go
@@ -315,7 +315,10 @@ func (s *scannerServer) CreateSBOM(ctx context.Context, req *pb.CreateSBOMReques
 	}
 
 	// Convert to SyftDocument and serialize
-	doc := syftToDomain(*syftSBOM)
+	doc, err := syftToDomain(*syftSBOM)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert SBOM: %w", err)
+	}
 	docBytes, err := json.Marshal(doc)
 	if err != nil {
 		return nil, fmt.Errorf("failed to serialize SBOM: %w", err)
@@ -342,8 +345,9 @@ func (s *scannerServer) Health(_ context.Context, _ *pb.HealthRequest) (*pb.Heal
 }
 
 // syftToDomain converts a Syft SBOM to a v1beta1.SyftDocument.
-// This is the same logic as SyftAdapter.syftToDomain.
-func syftToDomain(sbomSBOM sbom.SBOM) *v1beta1.SyftDocument {
+// This is the same logic as SyftAdapter.syftToDomain, including its error return: this
+// used to swallow both errors and return nil, which the caller then dereferenced.
+func syftToDomain(sbomSBOM sbom.SBOM) (*v1beta1.SyftDocument, error) {
 	doc := syftjson.ToFormatModel(sbomSBOM, syftjson.EncoderConfig{
 		Pretty: false,
 		Legacy: false,
@@ -351,16 +355,16 @@ func syftToDomain(sbomSBOM sbom.SBOM) *v1beta1.SyftDocument {
 
 	b, err := json.Marshal(doc)
 	if err != nil {
-		return nil
+		return nil, err
 	}
 
 	var syftDoc *v1beta1.SyftDocument
 	if err := json.Unmarshal(b, &syftDoc); err != nil {
-		return nil
+		return nil, err
 	}
 	syftmeta.Reattach(syftDoc.Artifacts, doc.Artifacts)
 
-	return syftDoc
+	return syftDoc, nil
 }
 
 func packageVersion(name string) string {
--- a/pkg/sbomscanner/v1/server_test.go
+++ b/pkg/sbomscanner/v1/server_test.go
@@ -22,6 +22,8 @@ import (
 
 	"github.com/anchore/stereoscope/pkg/image"
 	"github.com/anchore/syft/syft"
+	"github.com/anchore/syft/syft/artifact"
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Verified locally on `pr895`:

- `go build ./...`, `go vet ./...`, and `go test ./pkg/sbomscanner/v1/... -count=1` all pass; `gofmt` clean.
- The new `syftToDomain` error-return logic is byte-for-byte consistent with the in-process `SyftAdapter.syftToDomain` in `adapters/v1/syft_to_domain.go` — confirmed only one call site exists (`server.go:318`), so no other caller is left on the old nil-returning contract.
- Reproduced the regression the new test guards: reverting `syftToDomain`'s inner `return nil, err` back to `return nil, nil` makes `TestSyftToDomain_ReportsMarshalFailure` fail, confirming the test is load-bearing rather than incidental.
- The new gRPC error return in `CreateSBOM` follows the exact same shape as the existing `json.Marshal(doc)` failure path right below it, so this isn't introducing a new error-handling convention.
- Checked the `endActiveTempDirUse()` omission claim: the `defer endActiveTempDirUse()` lives inside the `dl.Run` closure, which has already returned by the time the new error path executes, so no explicit call is needed there — matches the adjacent serialize-error return.

No blockers. LGTM.

PR #897, opened 2026-08-28
test(tools): cover DeleteContents and the default retry config, drop two dead helpers

### LINKED ISSUE #896: internal/tools: DeleteContents and Default429RetryConfig are untested, FileToSBOM and FileContent are dead
`internal/tools` is at 79.3%. Two of the uncovered functions are live and load-bearing, and two are dead.

## Untested and live

`DeleteContents` deletes everything inside a directory. Its one caller is `adapters/v1/grype.go`, clearing a half-written Grype DB cache after a failed update when there was no pre-existing DB:

```go
if !hasExistingDB {
	if delErr := tools.DeleteContents(g.installCfg.DBRootDir); delErr != nil {
```

The property that caller depends on is that the directory survives being emptied, since the next install goes back into that same `DBRootDir`. Nothing pins it. Nothing pins that it stays inside the directory it was given either, which matters because the path comes from configuration.

`Default429RetryConfig` has five production callers, across both SBOM paths and the scan service, and no test. Its values bound how long a worker is held by one rate-limited registry pull, and `MaxRetryAfter` in particular is what caps a registry-chosen `Retry-After`. Without it the ceiling silently falls back to `MaxWait`, which is 2s rather than 30s, so the distinction is easy to erase by accident.

## Dead

`FileToSBOM` has **zero references** in the repository, production or test. `FileContent` has exactly one, inside `FileToSBOM`, so the pair is dead together. Both are exported from an `internal` package, so there are no outside consumers to consider.

## Not a real gap

`FileToCVEManifest` also reports 0.0% here, but that is an artifact of per-package measurement: it has 46 references from other packages' tests and none in production. It is a live test helper and is left alone.

Covering the two live functions and removing the dead pair.

AnshulPatil2005:
## Overview

`internal/tools` was at 79.3%. Two of the uncovered functions are live and load-bearing, two are dead, and one only looks uncovered.

79.3% to 86.7%. `DeleteContents` 0% to 87.5%, `Default429RetryConfig` 0% to 100%.

## The two that were live

**`DeleteContents`** empties a directory. Its one caller is `adapters/v1/grype.go`, clearing a half-written Grype DB cache after a failed update:

```go
if !hasExistingDB {
	if delErr := tools.DeleteContents(g.installCfg.DBRootDir); delErr != nil {
```

The property that caller depends on is that the directory itself survives, since the next install goes back into the same `DBRootDir`. Nothing pinned it. `TestDeleteContents` covers that alongside nested subdirectories, and `TestDeleteContents_LeavesSiblingsAlone` pins that it stays inside the directory it was given, which matters because that path is configured rather than fixed.

**`Default429RetryConfig`** has five production callers, across both SBOM paths and the scan service, and had no test. Its values bound how long a worker is held by one rate-limited registry pull. `MaxRetryAfter` is specifically what caps a registry-chosen `Retry-After`; without it `retryAfterCeiling` falls back to `MaxWait`, which is 2s rather than 30s. The test asserts both, and `TestFastRetryConfig_CeilingFallsBackToMaxWait` covers the fallback branch deliberately, since `FastRetryConfig` is the config that uses it.

## The two that were dead

`FileToSBOM` has **zero references** in the repository, production or test. `FileContent` has exactly one, inside `FileToSBOM`, so the pair is dead together. Both are exported from an `internal` package, so there are no outside consumers to break. Removed, along with the `v1beta1` import that became unused.

## The one that only looks uncovered

`FileToCVEManifest` also reports 0.0% in this package. That is per-package measurement rather than a gap: it has 46 references from other packages' tests and none in production. Left alone, and called out here so the remaining 0% in the report is not mistaken for dead code.

## Testing

`go build ./... && go vet ./... && go test ./internal/tools/ -count=1`

Three mutations, each caught:

| mutation | fails |
| --- | --- |
| make `DeleteContents` remove the directory itself | `TestDeleteContents`, `..._EmptyDirIsANoOp` |
| change `MaxRetryAfter` to 90s | `TestDefault429RetryConfig` |
| make `retryAfterCeiling` always fall back to `MaxWait` | `TestDefault429RetryConfig` |

`golangci-lint run ./internal/tools/...` exits 0 locally.

## Related issues/PRs:

* Resolved #896

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when clearing directory contents, including nested files and folders.
  * Preserved directories and sibling folders during cleanup operations.
  * Added clearer handling for missing directories.

* **Tests**
  * Added coverage for directory cleanup behavior.
  * Added validation for default and fast retry configuration values.

### DIFF
--- a/internal/tools/tools.go
+++ b/internal/tools/tools.go
@@ -18,7 +18,6 @@ import (
 	"github.com/distribution/reference"
 	"github.com/gofrs/flock"
 	"github.com/kubescape/kubevuln/core/domain"
-	"github.com/kubescape/storage/pkg/apis/softwarecomposition/v1beta1"
 	"k8s.io/apimachinery/pkg/util/validation"
 )
 
@@ -83,20 +82,6 @@ func LabelsFromImageID(imageID string) map[string]string {
 		}
 	}
 	return labels
-}
-
-func FileContent(path string) []byte {
-	b, err := os.ReadFile(filepath.Clean(path))
-	if err != nil {
-		return nil
-	}
-	return b
-}
-
-func FileToSBOM(path string) *v1beta1.SyftDocument {
-	sbom := v1beta1.SyftDocument{}
-	_ = json.Unmarshal(FileContent(path), &sbom)
-	return &sbom
 }
 
 func FileToCVEManifest(path string) domain.CVEManifest {
--- a/internal/tools/tools_test.go
+++ b/internal/tools/tools_test.go
@@ -564,3 +564,82 @@ func TestLabelsFromImageID_UppercaseTagIsKept(t *testing.T) {
 		assert.Empty(t, validation.IsDNS1123Label(value), "label %q has invalid value %q", key, value)
 	}
 }
+
+// TestDeleteContents covers the property its one caller depends on: adapters/v1/grype.go
+// clears a half-written DB cache with it and then expects to install into that same
+// directory, so emptying it must not remove it.
+func TestDeleteContents(t *testing.T) {
+	dir := t.TempDir()
+
+	require.NoError(t, os.WriteFile(path.Join(dir, "file.txt"), []byte("x"), 0o600))
+	require.NoError(t, os.Mkdir(path.Join(dir, "sub"), 0o750))
+	require.NoError(t, os.WriteFile(path.Join(dir, "sub", "nested.txt"), []byte("y"), 0o600))
+	require.NoError(t, os.Mkdir(path.Join(dir, "sub", "deeper"), 0o750))
+
+	require.NoError(t, DeleteContents(dir))
+
+	info, err := os.Stat(dir)
+	require.NoError(t, err, "the directory itself must survive")
+	assert.True(t, info.IsDir())
+
+	entries, err := os.ReadDir(dir)
+	require.NoError(t, err)
+	assert.Empty(t, entries, "files and subdirectories alike should be gone")
+}
+
+func TestDeleteContents_EmptyDirIsANoOp(t *testing.T) {
+	dir := t.TempDir()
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Reviewed. No blockers.

Verified independently rather than taking the PR description at face value:
- `FileToSBOM`/`FileContent` (exported, capitalized): a repo-wide code search turns up zero callers outside `tools.go` itself. `adapters/v1/grype_test.go` has its own unexported `fileToSBOM`/`fileContent` helpers, which are distinct symbols and unaffected by this removal. Safe to delete, and the now-unused `v1beta1` import removal is correct.
- `Default429RetryConfig`/`FastRetryConfig`/`retryAfterCeiling` in `internal/tools/retry.go`: the new test's asserted values (`MaxAttempts=3`, `InitialWait=500ms`, `MaxWait=2s`, `Backoff=2.0`, `MaxRetryAfter=30s`, ceiling fallback to `MaxWait` when `MaxRetryAfter` is unset) match the actual source exactly.
- `DeleteContents` itself is unchanged by this PR (test-only addition); the new tests (dir survives, empty-dir no-op, missing-dir error, siblings untouched) match its existing `os.ReadDir` + per-entry `os.RemoveAll` implementation.
- All CI checks green (build, lint, CodeQL, DCO, GitGuardian, cross-platform build, FOSSA).

One non-blocking nit, same as CodeRabbit's: `TestDeleteContents_LeavesSiblingsAlone` checks the sibling file still exists but doesn't assert its content is unchanged. Given `DeleteContents` only ever touches paths joined under the passed-in `dir`, it structurally can't mutate a sibling's content today, so this doesn't gate merge — worth tightening in a follow-up if the implementation ever changes.

Approving.

PR #903, opened 2026-08-29
test(repositories): check our own VEX statements against go-vex's rules

### LINKED ISSUE #902: nothing checks kubevuln's own VEX statements against the rules #880 applies to fetched ones
#880 added `internal/vexvalidate`, which checks a fetched OpenVEX document with go-vex's own `Statement.Validate()`. Nothing applies those rules to the documents kubevuln writes.

That is worth closing because the rules are mutual-exclusion ones, and the assessment code sets these fields field by field:

- `not_affected` requires a justification or an impact statement, and must carry no action statement
- `affected` requires an action statement, and must carry neither a justification nor an impact statement
- `fixed` and `under_investigation` must carry none of the three

So adding an impact statement to a `fixed` assessment, or leaving one in place when promoting a finding to `affected`, produces a document that decodes fine, stores fine, and is rejected by any consumer applying the spec, including kubevuln's own new validator.

Nothing catches that today. `repositories/apiserver.go` already imports go-vex for the status and justification constants, so the check costs nothing extra.

I checked the current output first, and it is compliant. All seven shapes kubevuln can emit pass:

- the baseline `not_affected` from `newLocalStatement`
- a finding promoted to `affected` by relevancy
- a `SecurityException` suppression, default and with its own justification/impact
- a `SecurityException` with status `fixed`
- a `SecurityException` accepting risk as `affected`
- a cloud exception policy with no CRD provenance

So this is a regression test rather than a bug report. It pins a property that currently holds, is easy to break from `ignoredMatchAssessment` or `markRelevantVulnerabilitiesAsAffectedInVex` without noticing, and now matters more because #387 means these documents get consumed by things that validate them.

One related observation, not part of this: a scan of a clean image produces a document with zero statements, which `vexvalidate` would reject as `ErrNoStatements`. That is consistent for a validator aimed at fetched feeds, since an image with no findings has nothing to assert, but it does mean kubevuln can publish a document its own validator would refuse. Worth a maintainer's view rather than a change from me.

AnshulPatil2005:
## Overview

#880 added `internal/vexvalidate`, which checks a *fetched* OpenVEX document with go-vex's own `Statement.Validate()`. Nothing applied those rules to the documents kubevuln *writes*.

This adds that as a regression test. Test-only, no production changes.

## Why it is worth pinning

The rules are mutual-exclusion ones, and the assessment code sets these fields individually:

| status | requires | must not carry |
| --- | --- | --- |
| `not_affected` | justification or impact statement | action statement |
| `affected` | action statement | justification, impact statement |
| `fixed`, `under_investigation` | nothing | all three |

So adding an impact statement to a `fixed` assessment, or leaving one in place when promoting a finding to `affected`, produces a document that decodes fine, stores fine, and is rejected by anything applying the spec, including kubevuln's own new validator.

## Current output is already compliant

I checked before writing the test, so this is a regression test rather than a fix. All seven shapes kubevuln can emit pass:

- the baseline `not_affected` from `newLocalStatement`
- a finding promoted to `affected` by relevancy
- a `SecurityException` suppression, default and carrying its own justification and impact
- a `SecurityException` with status `fixed`
- a `SecurityException` accepting risk as `affected`
- a cloud exception policy with no CRD provenance

Statements are round-tripped through JSON into go-vex's own `Statement` type and checked with the library rather than a local restatement of its rules, so the test cannot drift from what the library enforces. `repositories` already imports go-vex for the status constants, so there is no new dependency.

## Testing

`go test ./repositories/ -run TestGeneratedStatementsAreValidOpenVEX -count=1`

Three mutations, each caught:

| mutation | fails |
| --- | --- |
| keep a justification on a `fixed` assessment | `SecurityException_with_status_fixed` |
| make an `affected` assessment produce no action statement | `SecurityException_accepting_risk_as_affected` |
| leave the impact statement in place when promoting to `affected` | `relevant_finding_promoted_to_affected` |

`golangci-lint run ./repositories/...` exits 0 locally.

## One observation, not addressed here

A scan of a clean image produces a document with zero statements, which `vexvalidate` would reject as `ErrNoStatements`. That is consistent for a validator aimed at fetched feeds, since an image with no findings has nothing to assert, but it does mean kubevuln can publish a document its own validator would refuse. Flagging it for a maintainer's view rather than changing it.

## Related issues/PRs:

* Resolved #902

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive validation coverage for multiple OpenVEX statement variants.
  * Verified that statements remain valid after JSON serialization and deserialization.

### DIFF
--- a/repositories/apiserver_test.go
+++ b/repositories/apiserver_test.go
@@ -8,6 +8,7 @@ import (
 	"testing"
 	"time"
 
+	"encoding/json"
 	"github.com/akyoto/cache"
 	helpersv1 "github.com/kubescape/k8s-interface/instanceidhandler/v1/helpers"
 	"github.com/kubescape/k8s-interface/k8sinterface"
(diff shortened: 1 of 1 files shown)


repositories/apiserver_test.go
```diff
@@ -8,6 +8,7 @@ import (
 	"testing"
 	"time"
 
+	"encoding/json"
```
matthyx, line 11: Nit: `encoding/json` is a stdlib package but it's grouped with the third-party imports here (after the blank line, alongside `github.com/akyoto/cache` etc.) instead of the stdlib group above (`context`, `errors`, `fmt`, ...). `goimports`/`gofmt -s` would normally sort it into the first group. Not blocking since lint is green, but worth a quick fix for consistency.

matthyx (approved):
Reviewed. This is a test-only regression test (`repositories/apiserver_test.go`) validating the seven VEX statement shapes kubevuln emits against `go-vex`'s own `Statement.Validate()` — no production code changed.

Traced every helper the test calls (`newLocalStatement`, `ignoredMatchAssessment`, `applyIgnoredMatchAssessment`, `markRelevantVulnerabilitiesAsAffectedInVex`, `buildLocalVexStatementIndex`) against their real implementations in `repositories/apiserver.go`: each test case's expected mutual-exclusion shape (not_affected needs justification/impact and no action; affected needs action and no justification/impact; fixed carries none of the three) lines up with what the production code actually sets. All 8 CI checks pass (build, lint, tests, CodeQL, DCO, GitGuardian), and CodeRabbit found no actionable issues (the docstring-coverage warning is a generic heuristic that doesn't apply to Go).

Left one non-blocking nit about an import being in the wrong group. Approving — good to merge.

PR #905, opened 2026-08-29
fix(adapters): return syftToDomain's error instead of nil-derefing Content in the log

### LINKED ISSUE #904: In-process SyftAdapter.CreateSBOM nil-derefs domainSBOM.Content in its completion log when syftToDomain errors, crashing the process (in-process twin of #894)
# In-process `SyftAdapter.CreateSBOM` dereferences a nil `domainSBOM.Content` in its completion log when `syftToDomain` returns an error, crashing the whole kubevuln process

## Problem

`SyftAdapter.CreateSBOM` calls `s.syftToDomain(*syftSBOM)` and then, on the very next
statement, reads `domainSBOM.Content.Artifacts` for a log line — without checking the error
`syftToDomain` just returned. When `syftToDomain` fails, `domainSBOM.Content` is `nil`, so
that read is a nil pointer dereference. The panic happens on a bare goroutine that nothing
recovers, so it takes the entire process down.

This is the in-process twin of the sidecar bug fixed in #894 / PR #895. That PR fixed only
`pkg/sbomscanner/v1/server.go`; the identical "nil-`Content` in the completion log" hazard
in `adapters/v1/syft.go` was left in place. Here the failure mode is slightly different —
`syftToDomain` *does* return the error, but `CreateSBOM` ignores it and dereferences the nil
`Content` anyway — and the blast radius is larger, because it is the main kubevuln process
rather than a subprocess.

## Root cause

`adapters/v1/syft.go`, end of `CreateSBOM` (current `main`, lines 398–404):

```go
	// convert SBOM
	logger.L().Debug("converting SBOM",
	domainSBOM.Content, err = s.syftToDomain(*syftSBOM)

	// return SBOM
	logger.L().Debug("returning SBOM",
		helpers.String("imageID", imageID),
		helpers.Int("packages", len(domainSBOM.Content.Artifacts)))   // <-- nil deref when err != nil
	return domainSBOM, err
```

`domain.SBOM.Content` is `*v1beta1.SyftDocument` (`core/domain/sbom.go:28`).

`syftToDomain` (`adapters/v1/syft_to_domain.go`) returns `(nil, err)` on either
`json.Marshal(doc)` or `json.Unmarshal(...)` failure:

```go
func (s *SyftAdapter) syftToDomain(sbomSBOM sbom.SBOM) (*v1beta1.SyftDocument, error) {
	doc := syftjson.ToFormatModel(sbomSBOM, syftjson.EncoderConfig{Pretty: false, Legacy: false})
	b, err := json.Marshal(doc)
	if err != nil {
		return nil, err
	}
	...
}
```

`json.Marshal` of the syftjson format model **can** fail in production: `pkg.Package.Metadata`
is `any` and is passed through `ToFormatModel` into `doc` untouched, so a package whose
metadata carries a non-marshalable value — a `float64` NaN/Inf (`*json.UnsupportedValueError`)
or any non-JSON type — makes `json.Marshal(doc)` return an error. This is exactly the failure
mode issue #894 documented and reproduced for the sidecar's copy of this code
("a package whose metadata carries a NaN float, which is enough to make `json.Marshal` fail"),
and it is why PR #895 was merged — the maintainers have already accepted that this marshal
step is fallible.

Two things then combine:

1. `CreateSBOM` never checks `err` between the assignment and the log line.
2. The `helpers.Int("packages", len(domainSBOM.Content.Artifacts))` argument is evaluated
   eagerly by Go **regardless of the log level** (it is a plain expression passed to
   `helpers.Int`, then to `Debug`). So even with debug logging off, `len(nil.Artifacts)`
   runs and panics.

## Impact

- **Process crash / denial of service.** `CreateSBOM` is invoked from
  `ScanService.getOrCreateSBOM` inside a `singleflight.Group.DoChan` closure
  (`core/services/scan.go`). `golang.org/x/sync/singleflight` recovers a panic in the
  closure and, because a `DoChan` result channel is registered, **re-raises it with
  `go panic(e)`** (`singleflight.go:163-170` in v0.22.0) — a deliberate, unrecoverable
  re-panic on a fresh goroutine. Nothing above (`RetryWithBackoff`, the worker pool) has a
  recover either. The kubevuln pod dies → `CrashLoopBackOff`.
- **Poison image.** Every retry of the same scan re-runs `CreateSBOM` on the same image and
  re-crashes, so one workload/registry image containing one package with malformed float
  metadata permanently wedges the scanner until that image leaves scan scope.
- **The classified-failure path is defeated.** If `CreateSBOM` returned `(domainSBOM, err)`
  cleanly, `getOrCreateSBOM` → `classifySBOMError` → `platform.ReportScanFailure` would
  record a `ReasonSBOMGenerationFailed` and the scan flow would continue for other
  workloads. Instead the process is gone before any of that runs.
- This sits in the same family as #524 / #518 / #402 / #465 — "unguarded nil deref of a
  scan artifact crashes the process" — which the project has consistently treated as
  real bugs.

## Reproduction / evidence

Static: the code excerpt above; `singleflight.go` re-panic behaviour in the pinned
`golang.org/x/sync v0.22.0`; `domain.SBOM.Content` type; issue #894's reproduction of
`json.Marshal` failing on this exact `ToFormatModel` → `json.Marshal` sequence.

Behavioural (unit): `createSBOMFn` is already a package-level indirection that tests
override (`adapters/v1/syft_test.go:418`, `:456`). A regression test can set it to return a
`*sbom.SBOM` whose package collection contains one package with non-marshalable metadata,
then call `CreateSBOM` and assert it returns an error instead of panicking:

```go
orig := createSBOMFn
defer func() { createSBOMFn = orig }()
createSBOMFn = func(context.Context, source.Source, *syft.CreateSBOMConfig) (*sbom.SBOM, error) {
    s := &sbom.SBOM{Artifacts: sbom.Artifacts{Packages: pkg.NewCollection()}}
    s.Artifacts.Packages.Add(pkg.Package{
        Name: "p", Version: "1",
        Metadata: someStruct{Score: math.NaN()}, // or make(chan int) for a deterministic UnsupportedTypeError
    })
    return s, nil
}
// ... CreateSBOM(...) must return (sbom, err), not panic
```

(The in-process `syftToDomain` is only exercised by one happy-path test today —
`Test_syftAdapter_transformations` — so the error return has no coverage.)

## Affected components

- `adapters/v1/syft.go` — `SyftAdapter.CreateSBOM`, the trailing `logger.L().Debug("returning SBOM", …)` and the missing error check on `syftToDomain`.
- `adapters/v1/syft_to_domain.go` — `SyftAdapter.syftToDomain` (the `(nil, err)` return that the caller must handle).
- Downstream (impact only): `core/services/scan.go` `getOrCreateSBOM` singleflight path.

## Acceptance criteria

1. When `syftToDomain` returns an error, `CreateSBOM` returns `(domainSBOM, err)` (or a
   wrapped error) **without** dereferencing `domainSBOM.Content`. `domainSBOM.Status` is set
   to a non-successful value consistent with the other failure branches in `CreateSBOM`
   (e.g. `helpersv1.Incomplete`), so `classifySBOMError` / `ReportScanFailure` downstream
   classify it rather than treating a zero-value SBOM as usable.
2. The completion-log line does not evaluate `len(domainSBOM.Content.Artifacts)` (or any
   `domainSBOM.Content` field) unless `Content != nil`.
3. Regression test, using the `createSBOMFn` override, proving `CreateSBOM` returns an
   error (no panic) when the generated SBOM contains a package with non-marshalable
   metadata. A companion direct test of `syftToDomain` returning `(nil, err)` on a marshal
   failure is also reasonable.
4. No behaviour change on the success path (existing `CreateSBOM` / `syftToDomain` tests
   still pass, `packages` count still logged when `Content` is non-nil).

## High-level solution direction

Minimal, mirrors PR #895's shape on the sidecar side:

```go
domainSBOM.Content, err = s.syftToDomain(*syftSBOM)
if err != nil {
    logger.L().Ctx(ctx).Warning("failed to convert SBOM", helpers.Error(err),
    domainSBOM.Status = helpersv1.Incomplete
    return domainSBOM, fmt.Errorf("converting SBOM: %w", err)
}
logger.L().Debug("returning SBOM",
    helpers.String("imageID", imageID),
    helpers.Int("packages", len(domainSBOM.Content.Artifacts)))
return domainSBOM, nil
```

Optionally also make the completion-log line defensive on its own
(`packages := 0; if domainSBOM.Content != nil { packages = len(domainSBOM.Content.Artifacts) }`)
so a future refactor cannot reintroduce the eager nil deref.

bhuvan-somisetty:
## Overview

**Current behaviour:** the tail of `SyftAdapter.CreateSBOM` (`adapters/v1/syft.go`) does:

```go
domainSBOM.Content, err = s.syftToDomain(*syftSBOM)

// return SBOM
logger.L().Debug("returning SBOM",
	helpers.String("imageID", imageID),
	helpers.Int("packages", len(domainSBOM.Content.Artifacts)))  // evaluated regardless of log level
return domainSBOM, err
```

`syftToDomain` returns `(nil, err)` when the Syft document can't be `json.Marshal`ed, or when
the round-tripped bytes can't be `json.Unmarshal`ed back into `v1beta1.SyftDocument`. The
caller never checks that error before reading `domainSBOM.Content.Artifacts` for the log line,
and that argument is evaluated regardless of log level — so when `syftToDomain` errors, this
is a nil pointer dereference.

`CreateSBOM` runs inside a `singleflight.Group.DoChan` closure in
`ScanService.getOrCreateSBOM`. `golang.org/x/sync/singleflight` recovers a closure panic and,
with a `DoChan` result channel registered, re-raises it with `go panic(e)` — deliberately
unrecoverable. Nothing above (`RetryWithBackoff`, the worker pool) recovers either, so the
panic takes the whole kubevuln process down rather than failing the one scan.

**Reachability today:** `CreateSBOM` calls `v1beta1.StripSBOM(syftSBOM)` before conversion,
and `StripSBOM` rebuilds each package's metadata, replacing every type it doesn't explicitly
preserve with `nil` (`default: p.Metadata = nil`). None of the preserved metadata types carry
a `float64`, so the "package with a NaN/Inf float in its metadata" trigger from #894 does
**not** currently reach the in-process path — `StripSBOM` neutralises it. This fix is
therefore defensive:

- `syftToDomain`'s `json.Unmarshal` step is independently fallible and has no error-path
  coverage here (one happy-path test only).
- Relying on `StripSBOM`'s `default: nil` as the only thing between a conversion error and an
  unrecovered process panic is fragile — a `kubescape/storage` bump that preserves a new
  metadata type, or a Syft change, silently reintroduces the crash.
- The sidecar scanner's identical "nil `doc` dereferenced in the completion log" hazard was
  fixed the same way in #895; this is the in-process `adapters/v1/syft.go` twin.

**Future behaviour:** `CreateSBOM` checks the error, logs it, sets
`Status = helpersv1.Incomplete` (matching the sibling SBOM-generation failure branch), and
returns `fmt.Errorf("converting SBOM: %w", err)` — so `getOrCreateSBOM` classifies it as
`ReasonSBOMGenerationFailed` and reports it like any other failure. The completion-log line is
only reached with a non-nil `Content`.

`adapters/v1/syft_to_domain.go` is unchanged: the in-process `syftToDomain` already returns
the error; the caller just wasn't handling it.

## How to Test

```
go test ./adapters/v1/ -run syftToDomain -count=1
```

`Test_syftAdapter_syftToDomain_reportsMarshalFailure` — the in-process `syftToDomain` reports
a marshal failure (`json.Marshal` of a NaN float in package metadata) as `(nil, err)` rather
than a bare `nil` (it had no error-path coverage before). `go build ./adapters/...` and
`go vet ./adapters/v1/` pass; the full `adapters/v1` suite runs in CI.

## Impact

- Removes a latent unrecovered-panic path: a `syftToDomain` error now fails the one scan as
  `ReasonSBOMGenerationFailed` instead of crashing the kubevuln process.
- No behaviour change on the success path (existing `syftToDomain` / `CreateSBOM` tests
  unchanged; `packages` still logged).

## Related issues/PRs:

Resolved #904

Follows #895, which applied the same fix to the sidecar scanner's copy
(`pkg/sbomscanner/v1/server.go`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of SBOM conversion failures.
  * SBOMs now receive an **Incomplete** status when conversion cannot be completed.
  * Prevented follow-up processing errors and provided clearer failure reporting.

### DIFF
--- a/adapters/v1/syft.go
+++ b/adapters/v1/syft.go
@@ -397,12 +397,23 @@ func (s *SyftAdapter) CreateSBOM(ctx context.Context, name, imageID, imageTag st
 	logger.L().Debug("converting SBOM",
 		helpers.String("imageID", imageID))
 	domainSBOM.Content, err = syftmeta.ToDomain(*syftSBOM)
+	if err != nil {
+		// syftmeta.ToDomain returns (nil, err) when the Syft document can't be marshalled --
+		// a package carrying a non-JSON-encodable metadata value (a float NaN/Inf, say) is
+		// enough (see #904, #894). The completion log below reads domainSBOM.Content, so
+		// return here rather than dereferencing the nil pointer it now holds.
+		logger.L().Ctx(ctx).Warning("failed to convert SBOM to domain model",
+			helpers.Error(err),
+			helpers.String("imageID", imageID))
+		domainSBOM.Status = helpersv1.Incomplete
+		return domainSBOM, fmt.Errorf("converting SBOM: %w", err)
+	}
 
 	// return SBOM
 	logger.L().Debug("returning SBOM",
 		helpers.String("imageID", imageID),
 		helpers.Int("packages", len(domainSBOM.Content.Artifacts)))
-	return domainSBOM, err
+	return domainSBOM, nil
 }
 
 // Version returns Syft's version which is used to tag SBOMs


matthyx (dismissed):
Reviewed the diff and traced the downstream handling.

- The fix is correct: `syftToDomain` returns `(nil, err)` on marshal/unmarshal failure, and the old code read `domainSBOM.Content.Artifacts` in the completion log line before checking `err`, which is a nil pointer dereference. Since `CreateSBOM` runs inside a `singleflight.Group.DoChan` closure, that panic would take down the whole process rather than failing one scan.
- The new branch (`Status = Incomplete`, log warning, `return domainSBOM, fmt.Errorf("converting SBOM: %w", err)`) matches the existing style of the other error branches in this same function (e.g. the timeout and "failed to extract packages" branches already do `Status = Incomplete` + return a non-nil error).
- Verified the caller: `getOrCreateSBOM` → `classifySBOMError` has no special case matching `"converting SBOM: ..."`, so this correctly falls through to `ReasonSBOMGenerationFailed`, as the PR description claims. No double-counting — the singleflight closure reports a fresh `domain.SBOM{}` on error, so the `Status` field set here is inert but harmless.
- New regression test (`Test_syftAdapter_syftToDomain_reportsMarshalFailure`) directly proves the nil-`Content` risk this fixes (NaN metadata → marshal failure → nil doc).
- All CI checks (build, lint, CodeQL, DCO, GitGuardian, license/dependency checks) are green, and CodeRabbit found no actionable comments.

No blockers. LGTM.

matthyx:
rebase pls

bhuvan-somisetty:
Done, rebased onto main. Needed a small follow-up too — #915 landed in the meantime and pulled `syftToDomain` out into `internal/syftmeta.ToDomain`, shared by both SBOM paths, so I updated the fix to check the error from that instead. Same behavior otherwise.

matthyx (approved):
Re-reviewed after the rebase (my previous approval was dismissed by the push).

Since my last review, #915 merged and pulled `syftToDomain` out into the shared `internal/syftmeta.ToDomain`; this PR was updated to check the error from that instead, with the same behavior. Confirmed:

- `adapters/v1/syft.go` now imports `internal/syftmeta` and correctly checks `syftmeta.ToDomain`'s error before touching `domainSBOM.Content`, matching the same `Status = Incomplete` + wrapped-error pattern as the other error branches in this function.
- CodeRabbit flagged a "missing regression test" warning, but that's stale: the marshal-failure regression test that was previously in this PR is no longer needed here because `internal/syftmeta/reattach_test.go` (landed with #915) already has `TestToDomain_ReportsMarshalFailure`, covering the identical NaN-metadata → marshal-failure → `(nil, err)` scenario for the now-shared `ToDomain` function. No coverage was actually lost.
- All CI checks (build, lint, CodeQL, DCO, GitGuardian, license/dependency) are green on the current commit (`06bb51d`).

Still no blockers. LGTM.

PR #906, opened 2026-08-29
services: classify UNAUTHORIZED and NAME_UNKNOWN registry error codes

AHS0003:
Adds recognition for two registry error codes (UNAUTHORIZED and NAME_UNKNOWN) that were not being classified correctly before.

## Overview

Before: when an error contains the code UNAUTHORIZED or NAME_UNKNOWN, it falls into the generic "SBOM generation failed" category instead of being classified correctly.

After:
- UNAUTHORIZED -> classified as an authentication error (same category as the already-handled 401/403 cases)
- NAME_UNKNOWN -> classified as "image not found" (same category as the already-handled MANIFEST_UNKNOWN, but for a repository that doesn't exist instead of a missing tag)

## Additional Information

No behavior change to existing cases. Only two new cases added.

## How to Test

Two unit tests added in scan_failure_reasons_test.go. All tests pass (21/21).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scan failure reporting for registry authentication errors.
  * Improved detection and reporting when a referenced image repository or image layer cannot be found.
  * Preserved accurate handling of manifest-not-found errors across service boundaries.

### DIFF
--- a/core/services/scan_failure_reasons.go
+++ b/core/services/scan_failure_reasons.go
@@ -62,10 +62,17 @@ func classifySBOMError(err error) string {
 	switch {
 	case strings.Contains(errStr, "401 Unauthorized") || strings.Contains(errStr, "403 Forbidden"):
 		return scanfailure.ReasonImageAuthFailed
+	case strings.Contains(errStr, "UNAUTHORIZED"):
+		// uppercase code, same rationale as MANIFEST_UNKNOWN below: stable across registries
+		// even when the typed *transport.Error doesn't survive (e.g. crossing gRPC to the sidecar).
+		return scanfailure.ReasonImageAuthFailed
 	case strings.Contains(errStr, "404 Not Found") ||
 		strings.Contains(errStr, "MANIFEST_UNKNOWN") ||
 		strings.Contains(errStr, "NAME_UNKNOWN") ||
 		strings.Contains(errStr, "BLOB_UNKNOWN"):
+		// NAME_UNKNOWN: repository does not exist (distinct from MANIFEST_UNKNOWN: unknown
+		// tag/digest on an existing repo). Same "not found" bucket since armoapi-go has no
+		// dedicated reason.
 		return scanfailure.ReasonImageNotFound
 	// uppercase code is the stable token; lowercase phrase varies by registry. A typed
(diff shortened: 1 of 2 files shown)


matthyx (dismissed):
Reviewed and verified locally (checked out the branch, ran `go build ./...`, `go vet ./core/services/...`, and `go test ./core/services/... -run TestClassifySBOMError -v`): build clean, vet clean, all 21 subtests pass including the 2 new cases.

Notes:
- The two new cases (`UNAUTHORIZED`, `NAME_UNKNOWN`) follow the exact same pattern as the existing `MANIFEST_UNKNOWN`/`MANIFEST_SCHEMA_UNSUPPORTED` string-fallback cases — uppercase Docker Registry v2 error codes, used because the typed `*transport.Error` doesn't survive crossing gRPC to the sidecar.
- Confirmed against `armosec/armoapi-go`'s `scanfailure` package that there's no dedicated "repository not found" reason code, so reusing `ReasonImageNotFound` for `NAME_UNKNOWN` is the right bucket (consistent with how the platform-mismatch case above it is handled).
- No behavior change to existing cases; purely additive classification.
- Change is minimal and self-contained (2 files, +18/-0), with no other callers or edge cases affected.

No blockers. Approving.

matthyx:
rebase pls

matthyx (approved):
Re-reviewed after the rebase onto main (commit dd5f801) — my prior approval was auto-dismissed by the new push.

Re-verified locally against the current head: `go build ./...`, `go vet ./core/services/...`, and `go test ./core/services/... -run TestClassifySBOMError -v` — build clean, vet clean, all 26 subtests pass.

One thing worth flagging (not a blocker): main already merged #908/#909 in the meantime, which added `NAME_UNKNOWN` (plus `404 Not Found`/`BLOB_UNKNOWN`) to the not-found case. After the rebase, this PR's only functional delta is now the new `UNAUTHORIZED` → `ReasonImageAuthFailed` case; the `NAME_UNKNOWN` part of this PR is now just a clarifying comment on already-existing code, and the "NAME_UNKNOWN" test here duplicates coverage that already exists on main. Harmless either way — no conflict, no incorrect behavior — but the PR description's "before/after" framing for `NAME_UNKNOWN` is now slightly stale since the repo caught up independently.

CI is green across the board. Still no blockers. Re-approving.

PR #913, opened 2026-08-30
fix(config): reject a non-positive maxImageSize or maxSBOMSize

### LINKED ISSUE #912: maxSBOMSize and maxImageSize are unvalidated, so 0 silently classifies every image TooLarge and reports no CVEs
`LoadConfig` validates `scanTimeout` and `cveMatchingMode`. It does not validate `maxImageSize` or `maxSBOMSize`, and both are compared with a strict `>` against a measured size:

```go
sz := size.Of(syftSBOM)
if sz > s.maxSBOMSize {   // adapters/v1/syft.go, and the same in the sidecar
    ... helpersv1.TooLarge ...
}
```

So a non-positive value does not mean "unlimited", it means everything exceeds the limit. With `maxSBOMSize: 0`, every SBOM is classified `TooLarge`, no CVE manifest is ever produced, and the cluster reads as clean. The scan does not error, so nothing surfaces it: this is the quietest possible failure for a vulnerability scanner.

Confirmed against current main:

```
maxSBOMSize zero                             -> ACCEPTED: maxSBOMSize=0
maxSBOMSize negative                         -> ACCEPTED: maxSBOMSize=-1
maxImageSize zero                            -> ACCEPTED: maxImageSize=0
scanTimeout zero (the one that IS validated) -> REJECTED: scanTimeout must be positive, got 0s
```

The asymmetry is the interesting part. `scanTimeout` is guarded with a specific rationale (`time.NewTicker` panics on a non-positive duration). The two size limits break just as badly and are unguarded.

Nor is `0` documented for them, while the neighbouring settings that do have a `0` meaning say so explicitly:

- `maxQueueDepth`: "`0` or a negative value means unbounded"
- `scannerReadinessTimeout`: "A value of `0` or less does not mean 'wait forever': it makes the readiness deadline expire immediately"

The JSON schema for both size limits carries only `"type": "integer"` and a default, no `minimum`. So an operator who has read that `0` means unbounded for `maxQueueDepth` has every reason to assume the same here, and gets the exact opposite.

Rejecting a non-positive value at startup, matching `scanTimeout`. That is the fail-closed answer for a scanner: an operator gets a clear startup error instead of a cluster that silently reports zero vulnerabilities. Treating `0` as unlimited instead would be the other option, but it would quietly change behaviour for anyone who has set it, and "unlimited SBOM size" is not obviously a state worth making easy to reach.

Documenting the constraint on both rows too, since the absence of a note is part of what makes `0` look safe.

AnshulPatil2005:
## Overview

`LoadConfig` validates `scanTimeout` and `cveMatchingMode`, but not `maxImageSize` or `maxSBOMSize`. Both are compared with a strict `>` against a measured size:

```go
sz := size.Of(syftSBOM)
if sz > s.maxSBOMSize {
    ... helpersv1.TooLarge ...
}
```

So a non-positive value does not mean "unlimited", it means everything exceeds the limit. With `maxSBOMSize: 0`, every SBOM is classified `TooLarge`, no CVE manifest is produced, and the cluster reads as clean. The scan does not error, so nothing surfaces it.

## Measured before the change

```
maxSBOMSize zero                             -> ACCEPTED: maxSBOMSize=0
maxSBOMSize negative                         -> ACCEPTED: maxSBOMSize=-1
maxImageSize zero                            -> ACCEPTED: maxImageSize=0
scanTimeout zero (the one that IS validated) -> REJECTED: scanTimeout must be positive, got 0s
```

`scanTimeout` is guarded because `time.NewTicker` panics on a non-positive duration. The size limits break just as badly and were unguarded.

## Why 0 looks safe today

Neither size limit documents a `0` meaning, while the neighbouring settings that have one say so explicitly:

- `maxQueueDepth`: "`0` or a negative value means unbounded"
- `scannerReadinessTimeout`: "A value of `0` or less does not mean 'wait forever'..."

The JSON schema for both size limits carries only `"type": "integer"` and a default, no `minimum`. So an operator who has read `maxQueueDepth` has reason to assume the same convention here, and gets the opposite result.

## Why reject rather than treat 0 as unlimited

Rejecting is the fail-closed answer for a scanner: a clear startup error instead of a cluster silently reporting zero vulnerabilities. It also matches `scanTimeout`, and it does not quietly change behaviour for anyone who has already set `0` and is currently getting "everything TooLarge", which is surely not what they wanted either way.

`maxQueueDepth`'s own `0` is deliberately untouched, and `TestLoadConfigStillAcceptsZeroMaxQueueDepth` pins that, so the new guard is not later extended to it by analogy. Nothing compares a measured value against `maxQueueDepth`, so its `0` genuinely does mean unbounded.

Both rows in `docs/CONFIGURATION.md` now state the constraint, since the absence of a note is part of what made `0` look safe.

## Testing

`go build ./... && go vet ./config/ && go test ./config/ -count=1`

`TestLoadConfigRejectsNonPositiveSizeLimits` covers zero and negative for both fields. `TestLoadConfigAcceptsDefaultSizeLimits` checks the defaults still load, so the guard cannot be satisfied by rejecting everything.

Removing the `maxSBOMSize` half of the guard fails the two `maxSBOMSize` subtests and nothing else. The existing config tests, including `TestLoadConfigMaxQueueDepthDefault` and `TestLoadConfigRejectsNonPositiveScanTimeout`, pass unchanged.

`golangci-lint run ./config/...` exits 0 locally.

## Related issues/PRs:

* Resolved #912

## Summary by CodeRabbit

* **Bug Fixes**
  * Configuration now rejects zero or negative image and SBOM size limits at startup.
  * Existing behavior allowing a queue depth of zero remains supported.

* **Documentation**
  * Clarified size-limit validation, comparison behavior, and applicable SBOM size constraints.
  * Documented default size-limit behavior.


matthyx (approved):
Verified locally against `8028e95`:
- `go build ./...`, `go vet ./config/...`, `go test ./config/... -v`, and `golangci-lint run ./config/...` all pass, including the new `TestLoadConfigRejectsNonPositiveSizeLimits` / `TestLoadConfigAcceptsDefaultSizeLimits` / `TestLoadConfigStillAcceptsZeroMaxQueueDepth`.
- Confirmed the strict `>` comparisons this guards against: `sz > s.maxSBOMSize` in `adapters/v1/syft.go`, and `maxImageSize` flowing straight into the stereoscope context value used the same way — so `0`/negative really did mean "everything is TooLarge", not unlimited, matching the PR description.
- Confirmed `maxQueueDepth` is untouched and still has no comparison against a measured value anywhere in the codebase, so leaving its `0`-means-unbounded semantics alone is correct, and the pinning test (`TestLoadConfigStillAcceptsZeroMaxQueueDepth`) protects that going forward.
- No blockers. One non-blocking nit, already flagged by CodeRabbit: the "Complete Schema" JSON block in `docs/CONFIGURATION.md` (`maxImageSize`/`maxSBOMSize` around lines 236-247) doesn't get a `minimum: 1`, unlike `scanConcurrency`'s schema entry or `scanTimeout`'s new description+pattern in this same file — worth a follow-up so the machine-readable schema matches the prose table this PR updated, but doesn't block merging the actual fix.

Good to merge.

PR #915, opened 2026-08-30
refactor(sbom): share the two functions both SBOM paths still duplicate

### LINKED ISSUE #914: syftToDomain and packageVersion each exist twice across the two SBOM paths
Hashing every non-test function body in the repo, normalised for comments and whitespace, turns up two pieces of production logic that exist twice. Both copies are in the two SBOM paths, which is the seam `internal/syftmeta` and `internal/syftsource` were both created to close.

**`syftToDomain`.** `adapters/v1/syft_to_domain.go` and `pkg/sbomscanner/v1/server.go` hold the same twelve-line conversion, identical apart from one local variable name (`syftSBOM` vs `syftDoc`). The in-process copy is a method on `SyftAdapter` that never touches its receiver, so it is a free function in disguise:

```
$ awk '/func \(s \*SyftAdapter\) syftToDomain/,/^}/' adapters/v1/syft_to_domain.go | grep -c 's\.'
0
```

These two have already drifted once. The sidecar's copy dropped both errors and returned a `nil` that its caller then dereferenced, which was #893, fixed in #895 by bringing it back in line with the in-process one. That is the second time a copy in this pair has diverged; `syftmeta`'s own package comment cites #585, #587 and #601 for the same reason.

**`packageVersion`.** `internal/tools.PackageVersion` and an unexported `packageVersion` in `pkg/sbomscanner/v1/server.go` are byte-identical, nine lines each. The sidecar already imports `internal/tools` for `IsRateLimitError`, so the copy buys nothing.

Moving the conversion into `internal/syftmeta` as `ToDomain`, and pointing the sidecar at `tools.PackageVersion`.

`syftmeta` is the right home rather than a new package: it exists specifically because the JSON round-trip inside this conversion loses package metadata, and `Reattach` is the step that puts it back. Owning the round-trip and the repair together means callers get one function instead of two steps they each have to remember to pair.

After the change the same scan reports zero cross-file duplicate function bodies.

AnshulPatil2005:
## Overview

Hashing every non-test function body in the repo, normalised for comments and whitespace, turns up exactly two pieces of production logic that exist twice. Both are in the two SBOM paths.

## What was duplicated

**`syftToDomain`**, the same twelve-line conversion in `adapters/v1/syft_to_domain.go` and `pkg/sbomscanner/v1/server.go`, identical apart from one local variable name. The in-process copy was a method on `SyftAdapter` that never touched its receiver:

```
$ awk '/func \(s \*SyftAdapter\) syftToDomain/,/^}/' adapters/v1/syft_to_domain.go | grep -c 's\.'
0
```

This pair has already drifted once: the sidecar's copy dropped both errors and returned a `nil` the caller then dereferenced, which was #893, fixed in #895 by bringing it back into line.

**`packageVersion`**, byte-identical to `internal/tools.PackageVersion`. The sidecar already imports `internal/tools` for `IsRateLimitError`.

## Why `syftmeta`

Rather than a new package: `syftmeta` exists *because* the JSON round-trip inside this conversion loses package metadata, and `Reattach` is the step that puts it back. Its comment already cites #585, #587 and #601 as drift between these same two paths. Owning the round-trip and the repair together means callers get one function instead of two steps they each have to remember to pair. The package comment is widened accordingly.

## Testing

`go build ./... && go vet ./... && go test ./adapters/v1/ ./pkg/sbomscanner/v1/ ./internal/syftmeta/ -count=1`

All three pass. `golangci-lint run` over the three packages exits 0 locally.

Behaviour is unchanged: the moved body is the in-process one, which the sidecar's already matched after #895. The two tests added there move with it, and `adapters/v1/syft_test.go`'s full round-trip assertion (which compares the converted document against a fixture byte for byte) now calls the shared function and still passes, which is the real check that the move is faithful.

Re-running the duplicate scan afterwards reports zero cross-file duplicate function bodies.

**One thing I could not pin.** I tried adding a test asserting `ToDomain` reattaches metadata, and removed it again: it passed with the `Reattach` call deleted, so it was asserting nothing. `MetadataType` and `Metadata` survive the round-trip on their own for the cases I could construct, because `v1beta1.SyftPackage.Metadata` is a `json.RawMessage` that captures the raw JSON. `Reattach` itself is covered directly and thoroughly by the existing tests in that package; what is not pinned is `ToDomain` calling it. Flagging rather than leaving a test that looks like coverage and is not.

## Related issues/PRs:

* Resolved #914

## Summary by CodeRabbit

* **Refactor**
  * Standardized Syft SBOM conversion across scanning components.
  * Improved handling and reporting of conversion errors.
  * Centralized package version information for more consistent reporting.

* **Tests**
  * Added coverage for successful SBOM conversion and invalid metadata errors.


AnshulPatil2005:
while doing this i scanned the whole repo for duplicate function bodies (hashed, normalised for comments and whitespace). after this pr there are zero cross-file duplicates left in non-test code, so that side is clean. the only same-file ones are the four Scan*/Validate* methods on mockscan

AnshulPatil2005:
tools.FileToCVEManifest, used by repositories
a local fileContent + fileToX pair, byte-identical in adapters/v1 and core/services
a generic fileToType[T any] in adapters/v1/backend_test.go
one-off fileToReport and fileToContainerProfile
fileToType[T any] already covers all of them.

it's also why tools.FileToSBOM was dead when i removed it in #897: each package had rolled its own rather than finding the shared one.

matthyx (approved):
Reviewed the diff and verified locally against the PR branch (commit 36f6f21):

- `go build ./...`, `go vet ./...` — clean
- `go test ./adapters/v1/ ./pkg/sbomscanner/v1/ ./internal/syftmeta/ -count=1` — all pass, including the new `TestToDomain` / `TestToDomain_ReportsMarshalFailure`
- Confirmed no dangling references to the removed `syftToDomain` / local `packageVersion`
- `internal/tools.PackageVersion` is byte-identical to the removed sidecar copy, so the dedup is behavior-preserving
- The moved `ToDomain` body matches the prior in-process implementation (post-#895), and the round-trip fixture assertion in `syft_test.go` still passes against it

No blockers. The one thing the PR description flags itself (no test pinning that `ToDomain` calls `Reattach`) is a reasonable, well-justified scope call, not a gap worth blocking on — `Reattach` itself is directly covered.

`golangci-lint` locally reports 27 pre-existing issues, but all are in files/lines this PR doesn't touch, so not a blocker here.

Approving.

PR #920, opened 2026-08-31
fix(repositories): cache GetWorkloadLabels/GetNamespaceLabels results

### LINKED ISSUE #510: SecurityException/ClusterSecurityException exception lookups scale with scan volume, not CRD churn - no shared cache or watch behind GetSecurityExceptions

### LINKED ISSUE #919: GetWorkloadLabels/GetNamespaceLabels have no caching, so objectSelector/namespaceSelector matching re-fetches the same workload's labels once per container
# Description

`GetWorkloadLabels` and `GetNamespaceLabels` (`repositories/apiserver.go:710` and `:732`) resolve a scanned workload's/namespace's labels so `matchExceptionTarget` can evaluate a `SecurityException`'s `objectSelector`/`namespaceSelector`. Both issue a plain, uncached `Get()` against the same rate-limited `dynamic.Interface` client (`QPS: 50, Burst: 100`, `repositories/apiserver.go:214-215`) that `GetSecurityExceptions` uses — but unlike `GetSecurityExceptions`'s `List()` calls, which #510/#511 gave a shared cache, these two `Get()` calls have no caching at all: every call re-fetches from the K8s API server.

#510's own "Additional context" section explicitly named these two calls as part of the same rate-limiter-exhaustion problem ("All of this traffic — plus the `GetWorkloadLabels`/`GetNamespaceLabels` Get() calls that objectSelector/namespaceSelector matching needs — shares one `dynamic.Interface` client rate-limited to QPS=50/Burst=100"). The PR that closed #510 (#511, "fix(repositories): cache SecurityException/ClusterSecurityException List() results") is scoped, by its own title and diff, purely to the `List()` calls — it never touches `GetWorkloadLabels`/`GetNamespaceLabels`. This issue is that unaddressed leftover half of #510's original problem statement, not a duplicate of it.

**Where the calls actually come from, and why they're not deduplicated even within one request:** `BuildExceptionTarget` (`adapters/v1/securityexception_match.go:229`) calls `repo.GetWorkloadLabels`/`repo.GetNamespaceLabels` whenever any active `SecurityException`/`ClusterSecurityException` uses `objectSelector`/`namespaceSelector`. It's invoked from `BackendAdapter.fetchCVEExceptions` (`adapters/v1/backend.go:383`), which runs on every `exceptionsCache` miss — and that cache is keyed per **container** (`accountID/cluster/namespace/kind/name/containerName/imageTag`, `adapters/v1/backend.go:269-277`), not per workload. So for a workload with N containers, `ScanCP`'s per-container loop (`core/services/scan.go`'s `ScanCP`) drives N separate `exceptionsCache` misses for what is, from `GetWorkloadLabels`'s point of view, the identical `(namespace, kind, name)` — N redundant `Get()` calls for the same object's labels, all within a single `ScanCP` invocation. The same repeats every `exceptionsCacheTTL` (1 minute) for any workload scanned repeatedly.

# Reproduction / Evidence

1. Enable SecurityException CRD integration (`storage != nil && c.RiskAcceptance`) and create one `SecurityException` or `ClusterSecurityException` whose `spec.match.objectSelector` (or `namespaceSelector`) is set — an ordinary, documented, supported way to scope a suppression by label rather than by name.
2. Send a `ScanCP` request for a workload with more than one container (`domain.ScanCommand.Args` naming a Deployment/Pod with, say, 3 containers).
3. Watch K8s API server audit logs (or the dynamic client's rest-config metrics) for `get` requests to the workload's GVR (e.g. `deployments`) and to `namespaces`.

Expected: at most one `Get()` for the workload's labels and one for the namespace's labels per `ScanCP` invocation, since all three containers belong to the same workload/namespace. Actual: one `Get()` per container (confirmed by reading the call chain — `ScanCP`'s loop over `scans` → `applyExceptionsToManifest`/`GetCVEExceptions` → `fetchCVEExceptions` → `BuildExceptionTarget` → `GetWorkloadLabels`/`GetNamespaceLabels`, with no cache or memoization anywhere in that path for these two calls specifically, confirmed by `grep -n "GetWorkloadLabels\|GetNamespaceLabels" repositories/apiserver_test.go` turning up only cancellation-propagation tests, no caching test).

# Root Cause

`GetWorkloadLabels`/`GetNamespaceLabels` are plain pass-through `Get()` calls with no caching layer, while every other read on the same exception-resolution path (`GetSecurityExceptions`'s `List()`s, and the per-workload merged-policy cache in `BackendAdapter`) already has one. The gap is a scope boundary, not an oversight in a single PR: #510 flagged both problems together, #511 solved only the `List()` half, and #510 was closed as complete without a follow-up being filed for the other half.

# Impact

- **Redundant API server load, scaling with container count**: a single `ScanCP` call for an N-container workload makes N times the necessary `Get()` calls for data that's identical across all N (workload/namespace labels don't vary per container).
- **Shares the same easily-exhausted rate limiter** (`QPS: 50, Burst: 100`) as the `SecurityException`/`ClusterSecurityException` `List()` calls this limiter was already flagged (in #510) as being at risk from — so this traffic can throttle those `List()` calls too, and vice versa, even after #511's fix.
- Only manifests when `objectSelector`/`namespaceSelector` is actually used, but that's an intentional, documented feature (not a rare misconfiguration), and its whole purpose is to apply across many workloads by label — exactly the usage pattern that makes this scale.

# Affected Components

- `repositories/apiserver.go`: `GetWorkloadLabels` (~line 710), `GetNamespaceLabels` (~line 732)
- `adapters/v1/securityexception_match.go`: `BuildExceptionTarget` (~line 229) — the caller driving the redundant calls
- `adapters/v1/backend.go`: `fetchCVEExceptions` (~line 383) — where the per-container cache-key granularity means even one `ScanCP` call fans this out

# Acceptance Criteria

- A workload's labels (and a namespace's labels) are fetched at most once per short TTL window, shared across every container/image of that workload scanned within it — not once per container, not once per `exceptionsCacheTTL` miss.
- The existing fail-closed semantics are preserved exactly: `GetWorkloadLabels`/`GetNamespaceLabels` returning an error (including `NotFound`) must still propagate as "labels unresolved" so `matchExceptionTarget`'s `WorkloadLabelsResolved`/`NamespaceLabelsResolved` gates still fail closed (a negative selector must never match an empty/uncached-away label set).
- A failed lookup is never cached (mirrors `securityExceptionListCache`'s existing "never cache a failure" rule), so a transient apiserver hiccup self-heals on the next call.
- No change to `matchExceptionTarget`'s matching semantics — this is a caching/fetch-path change only, same boundary #511 held itself to.
- New tests demonstrate: (a) N calls for the same workload within the TTL window produce exactly one `Get()`, and (b) a different workload/namespace still gets its own fresh lookup.

# Solution Direction

Reuse the same pattern already established twice in this exact file for this exact class of problem: a small `*cache.Cache` (or the existing `securityExceptionListCache`, keyed additionally by object identity) keyed by `namespace/kind/name` for `GetWorkloadLabels` and by `namespace` for `GetNamespaceLabels`, with a TTL in the same range as `securityExceptionListCacheTTL` (30s). Given `#916`'s singleflight work landed in the same file, extending `coalescedSecurityExceptionList` (or a sibling generic helper) to cover these two `Get()`s would also close the identical concurrent-miss gap for them for free, rather than needing a third bespoke cache.

bhuvan-somisetty:
## Overview

`GetWorkloadLabels` and `GetNamespaceLabels` (`repositories/apiserver.go`) resolve a scanned
workload's/namespace's labels so a `SecurityException`'s `objectSelector`/`namespaceSelector`
can be evaluated. Both issued a fresh, uncached `Get()` against the same rate-limited
(`QPS: 50, Burst: 100`) `dynamic.Interface` client `GetSecurityExceptions`' own `List()` calls
share — but unlike those `List()` calls, which #510/#511 gave a shared cache, these two `Get()`s
had none at all.

**Why it matters more than "one extra API call":** `BackendAdapter.exceptionsCache` is keyed
per **container**, not per workload (`accountID/cluster/namespace/kind/name/containerName/imageTag`),
and `BuildExceptionTarget` resolves the workload's/namespace's labels on every miss of that
cache. So for a workload with N containers, `ScanCP`'s per-container loop drives N separate
`exceptionsCache` misses for what is, from `GetWorkloadLabels`'s point of view, the identical
`(namespace, kind, name)` — N redundant `Get()`s for the same object's labels, all within one
`ScanCP` call, whenever an `objectSelector`/`namespaceSelector`-based exception is in use (an
ordinary, documented way to scope a suppression by label rather than by name).

**Root cause / relation to #510:** #510's own "Additional context" named these two `Get()`
calls as part of the same rate-limiter-exhaustion problem it filed. The PR that closed #510
(#511, "cache SecurityException/ClusterSecurityException List() results") is scoped, by its
own title and diff, purely to the `List()` calls — it never touched `GetWorkloadLabels`/
`GetNamespaceLabels`. This is that other half.

## Solution

Add `labelsCache` to `APIServerStore`, mirroring `securityExceptionListCache`'s
already-established pattern in the same file, field for field:

- Same 30s TTL (`labelsCacheTTL`) — close enough to live, while collapsing a burst of `Get()`s
  for the same object into one.
- Nil-safe, exactly like `securityExceptionListCache`: several existing tests construct
  `APIServerStore{}` literals directly instead of through the constructors, and those keep
  working uncached rather than panicking.
- A failed lookup (including `NotFound`) is never cached, so it self-heals on the next call
  instead of pinning "unresolved" — and therefore fail-closed/not-applied, per
  `matchExceptionTarget`'s existing semantics — for the full TTL.
- No informer-based invalidation (unlike `securityExceptionListCache`, which #733 gave one):
  a label edit is rare and not otherwise latency-sensitive on this path, so the TTL alone
  bounding staleness is the right amount of machinery here, not a corner cut.

`matchExceptionTarget`'s matching semantics are unchanged — this is a caching/fetch-path
change only, the same boundary #511 held itself to.

## How to Test

```
go test ./repositories/... -run 'GetWorkloadLabels|GetNamespaceLabels' -v
```

New tests:
- `TestAPIServerStore_GetWorkloadLabels_Caches` / `TestAPIServerStore_GetNamespaceLabels_Caches`
  — two calls for the same workload/namespace collapse into a single `Get()`; a different,
  nonexistent one still gets its own fresh lookup.
- `TestAPIServerStore_GetWorkloadLabels_DoesNotCacheFailures` /
  `..._GetNamespaceLabels_DoesNotCacheFailures` — a failed `Get()` is never cached, so the next
  call retries instead of replaying the failure.

The two pre-existing `..._HonorsCallerCancellation` tests (which construct `APIServerStore{}`
literals with no `labelsCache` set) still pass unchanged, confirming the nil-safe fallback.

`go build ./...`, `go vet ./repositories/...`, and the full `repositories` suite all pass
locally.

## Impact

- Removes a `Get()`-call multiplier that scales with container count, on the same rate-limited
  client `GetSecurityExceptions`' `List()` calls already depend on — so this also reduces
  pressure on those, not just on itself.
- No behavior change on the correctness/semantics side: fail-closed handling for an unresolved
  selector, `NotFound` propagation, and the empty-namespace/kind/name early return are all
  unchanged.

## Related issues/PRs:

Fixes #919

Completes the scope #510 originally described but #511 didn't cover.

## Summary by CodeRabbit

- **Performance**
  - Improved workload and namespace label lookups by caching successful results for a short period.
  - Repeated requests can now be served without additional backend lookups.

- **Reliability**
  - Failed or unavailable lookups are not cached, allowing subsequent requests to retry automatically.

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -74,6 +74,21 @@ const (
 	securityExceptionListCacheKeyPrefix        = "se/"
 )
 
+// labelsCacheCleaningInterval/TTL bound how stale the workload/namespace labels
+// GetWorkloadLabels/GetNamespaceLabels read can be. Unlike securityExceptionListCache, this data
+// has no informer behind it to invalidate on change (a label edit is rare and not otherwise
+// latency-sensitive here), so the TTL alone bounds staleness — same 30s value, for the same
+// reason: close enough to live, while still collapsing a burst of Get()s for the same object
+// into one. Without this cache, a single ScanCP call for an N-container workload issues N
+// identical Get()s for that one workload's labels (BuildExceptionTarget resolves them on every
+// exceptionsCache miss, and that cache is keyed per container, not per workload) — see #919.
+const (
+	labelsCacheCleaningInterval   = 30 * time.Second
+	labelsCacheTTL                = 30 * time.Second
+	workloadLabelsCacheKeyPrefix  = "wl/"
+	namespaceLabelsCacheKeyPrefix = "ns/"
+)
+
 // resourceVersionMetadata mirrors kubescape/storage's
 // softwarecomposition.ResourceVersionMetadata. Passed as GetOptions.ResourceVersion, its
 // file-backed apiserver (pkg/registry/file/storage.go) honours it by reading only the
@@ -94,6 +109,10 @@ type APIServerStore struct {
 	// nil is safe (falls back to always listing, e.g. in tests that construct APIServerStore
 	// literals directly instead of through the constructors below).
 	securityExceptionListCache *cache.Cache
+
+	// labelsCache caches GetWorkloadLabels'/GetNamespaceLabels' Get() results; nil is safe,
+	// same as securityExceptionListCache above.
+	labelsCache *cache.Cache
 
 	// securityExceptionCacheEntries backs the compare-and-swap that keeps a List() in flight
 	// when a CRD change invalidates its cache key from silently re-populating the cache with
@@ -222,6 +241,7 @@ func NewAPIServerStorage(namespace string) (*APIServerStore, error) {
 		DynamicClient:              dynClient,
 		Namespace:                  namespace,
 		securityExceptionListCache: cache.New(securityExceptionListCacheCleaningInterval),
+		labelsCache:                cache.New(labelsCacheCleaningInterval),
 	}
 	store.enableSecurityExceptionCacheInvalidation(context.Background())
 	return store, nil
@@ -332,6 +352,7 @@ func newFakeAPIServerStore(namespace string, storageClient spdxv1beta1.SpdxV1bet
 		}),
 		Namespace:                  namespace,
 		securityExceptionListCache: cache.New(securityExceptionListCacheCleaningInterval),
+		labelsCache:                cache.New(labelsCacheCleaningInterval),
 	}
 }
 
@@ -575,10 +596,25 @@ func (a *APIServerStore) listClusterSecurityExceptions(ctx, listCtx context.Cont
 // workload is not found, nil labels are returned (the selector then matches
 // nothing, i.e. the exception is not applied — the safe default for a
 // suppression feature).
+//
+// Served from labelsCache when a fresh entry exists: the labels are the same for every
+// container of the same workload, but BuildExceptionTarget resolves them on every
+// exceptionsCache miss (BackendAdapter.exceptionsCache is keyed per container, not per
+// workload), so without this a single ScanCP call for an N-container workload issues N
+// identical Get()s for one object's labels — see #919. A failed lookup (including NotFound)
+// is never cached, so it self-heals on the next call instead of pinning "unresolved" (and
+// therefore fail-closed/not-applied) for the TTL.
 func (a *APIServerStore) GetWorkloadLabels(ctx context.Context, namespace, kind, name string) (map[string]string, error) {
 	if namespace == "" || kind == "" || name == "" {
 		return nil, nil
 	}
+	cacheKey := workloadLabelsCacheKeyPrefix + namespace + "/" + kind + "/" + name
+	if a.labelsCache != nil {
+		if cached, ok := a.labelsCache.Get(cacheKey); ok {
+			return cached.(map[string]string), nil
+		}
+	}
+
 	gvr, err := k8sinterface.GetGroupVersionResource(kind)
 	if err != nil {
 		return nil, fmt.Errorf("failed to resolve GroupVersionResource for kind %q: %w", kind, err)
@@ -592,15 +628,29 @@ func (a *APIServerStore) GetWorkloadLabels(ctx context.Context, namespace, kind,
 		// could not be resolved.
 		return nil, err
 	}
-	return obj.GetLabels(), nil
+	labels := obj.GetLabels()
+	if a.labelsCache != nil {
+		a.labelsCache.Set(cacheKey, labels, labelsCacheTTL)
+	}
+	return labels, nil
 }
 
 // GetNamespaceLabels resolves the labels of a namespace so that a
 // ClusterSecurityException's match.namespaceSelector can be evaluated.
+//
+// Served from labelsCache when a fresh entry exists, and a failed lookup is never cached —
+// see GetWorkloadLabels, which this mirrors.
 func (a *APIServerStore) GetNamespaceLabels(ctx context.Context, name string) (map[string]string, error) {
 	if name == "" {
 		return nil, nil
 	}
+	cacheKey := namespaceLabelsCacheKeyPrefix + name
+	if a.labelsCache != nil {
+		if cached, ok := a.labelsCache.Get(cacheKey); ok {
+			return cached.(map[string]string), nil
+		}
+	}
+
 	getCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
 	defer cancel()
 	obj, err := a.DynamicClient.Resource(namespaceGVR).Get(getCtx, name, metav1.GetOptions{})
@@ -609,7 +659,11 @@ func (a *APIServerStore) GetNamespaceLabels(ctx context.Context, name string) (m
 		// GetWorkloadLabels).
 		return nil, err
 	}
-	return obj.GetLabels(), nil
+	labels := obj.GetLabels()
+	if a.labelsCache != nil {
+		a.labelsCache.Set(cacheKey, labels, labelsCacheTTL)
+	}
+	return labels, nil
 }
 
 func (a *APIServerStore) GetContainerProfile(ctx context.Context, namespace string, name string) (v1beta1.ContainerProfile, error) {
--- a/repositories/apiserver_test.go
+++ b/repositories/apiserver_test.go
@@ -3192,6 +3192,173 @@ func TestAPIServerStore_GetNamespaceLabels_HonorsCallerCancellation(t *testing.T
 	require.True(t, *called, "the Get call the hook was attached to must have run")
 }
 
+// TestAPIServerStore_GetWorkloadLabels_Caches is a regression test for #919:
+// GetWorkloadLabels used to issue a fresh Get() call on every invocation, even though the
+// labels are identical for every container of the same workload -- BuildExceptionTarget
+// resolves them on every BackendAdapter.exceptionsCache miss, and that cache is keyed per
+// container, so a single ScanCP call for an N-container workload used to fan this out into N
+// Get()s for one object. A Get() call counter proves the cache collapses repeated calls for
+// the same workload into a single Get(), and that a genuinely different workload still gets
+// its own fresh lookup.
+func TestAPIServerStore_GetWorkloadLabels_Caches(t *testing.T) {
+	pod := &unstructured.Unstructured{Object: map[string]interface{}{
+		"apiVersion": "v1",
+		"kind":       "Pod",
+		"metadata": map[string]interface{}{
+			"name":      "mypod",
+			"namespace": "kubescape",
+			"labels":    map[string]interface{}{"app": "nginx"},
+		},
+	}}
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Reviewed the diff (repositories/apiserver.go, repositories/apiserver_test.go).

- New `labelsCache` mirrors the existing `securityExceptionListCache` pattern (nil-safe, TTL-only, failures never cached), so `GetWorkloadLabels`/`GetNamespaceLabels` avoid the N-per-container `Get()` fan-out described in #919.
- Cache keys (`wl/<namespace>/<kind>/<name>`, `ns/<name>`) can't collide since Kubernetes names/kinds don't contain `/`.
- Verified the returned label maps are only read (via `labels.Set(lbls)` in `matchExceptionTarget`), never mutated by callers, so sharing the cached map across calls is safe.
- Ran locally: `go build ./...`, `go vet ./repositories/...`, and `go test ./repositories/... -race` — all pass, including the new cache/failure-not-cached tests. CI (lint, tests, CodeQL, GitGuardian) is green.

No blockers. Approving.

PR #922, opened 2026-09-01
fix(registryauth): only fall back to anonymous on a repeated auth denial

### LINKED ISSUE #921: ResolveSource falls back to an anonymous pull on any error from the credentialed retry, not just a repeated 401
# Description

`ResolveSource` (`internal/registryauth/resolve.go:35-89`) is the shared pull-fallback ladder
both SBOM paths (`adapters/v1/syft.go`, `pkg/sbomscanner/v1/server.go`) use. When the initial,
credential-less pull fails with `401 Unauthorized`, it looks up a matching credential provider
(ECR/GCP), retries with those credentials, and — per the comment at lines 69-70 — falls back to
an anonymous retry "if no provider matched, its credentials were unavailable, **or it succeeded
in auth but still got 401**".

The code doesn't implement that last clause. The check that decides whether to fall back to
anonymous is:

```go
// line 71
if err != nil {
```

not

```go
```

So *any* error from the credentialed retry — not just a repeated 401 — triggers the anonymous
fallback: a transient 500, a DNS failure, a timeout, or (concretely, since this exact package
exists to handle registry rate limiting) a `429 Too Many Requests` that the credentialed
attempt's own `tools.RetryWithBackoff` already exhausted its retry budget on.

# Reproduction / Evidence

`ResolveSource[T any](ctx, component, get Getter[T], imageID, imageTag string, opts image.RegistryOptions) (T, error)` takes `get` as a parameter, so this is directly, deterministically testable without a real registry:

```go
get := func(_ context.Context, ref string, _ *image.RegistryOptions) (string, error) {
    if len(opts.Credentials) == 0 {
        return "", errors.New("401 Unauthorized")            // initial, no-creds pull
    }
    return "", errors.New("429 Too Many Requests")            // credentialed retry: NOT a 401
}
src, err := ResolveSource(ctx, "test", get, "img", "img:tag", image.RegistryOptions{})
```

Expected: `err` reports the 429 the credentialed (correctly authenticated) attempt actually hit,
and `get` is called twice (initial + credentialed retry) — there is nothing left to gain from
retrying anonymously when the credentials were never the problem.

Actual (reading the code, lines 53-84): the credentials-available branch (60-67) runs `get`
again, sees a non-nil `err` (the 429), and line 71's bare `if err != nil` sends it straight into
the anonymous-retry branch (71-84) anyway — a third `get` call, using no credentials, against an
image that (per the reproduction's own `get`) is provably not an auth problem. Each of these
three `get` calls is itself wrapped in its own `tools.RetryWithBackoff` with
`tools.Default429RetryConfig()` by both real callers (`adapters/v1/syft.go:240`,
`pkg/sbomscanner/v1/server.go`), so a persistently-429ing registry gets hit with the credentialed
attempt's full retry-with-backoff budget, then — because of this bug — the anonymous attempt's
full retry-with-backoff budget too, doubling the request volume against a registry that already
asked to be backed off from.

The final error is also misleading: line 82 unconditionally wraps the anonymous attempt's
failure as `fmt.Errorf("%w (anonymous fallback failed: %v)", unauthorizedErr, err)`, where
`unauthorizedErr` is the *original* pre-credentials 401. An operator reading logs for a scan
that failed because of a transient registry 500 (say) sees "401 Unauthorized ... anonymous
fallback failed: 500 ...", permanently mislabeling the failure as an auth problem.

# Root Cause

Line 71's fallback-to-anonymous guard checks only `err != nil`, not `err != nil &&
strings.Contains(err.Error(), "401 Unauthorized")` — the same check the function already uses
at line 50 to decide whether to enter this ladder in the first place. The doc comment (lines
69-70) describes the intended condition correctly; the code just doesn't implement it.

# Impact

- **Wasted, amplified registry load exactly where this package is designed to avoid it**: a
  credentialed pull that fails for any non-auth reason (429, 5xx, timeout, DNS) triggers a
  second full attempt — with its own nested rate-limit retry-with-backoff cycle — that cannot
  possibly succeed any better, since credentials were never the issue. For a 429 specifically,
  this doubles the request volume against a registry that has just said to slow down, working
  directly against the rate-limit handling this codebase invests heavily in elsewhere
  (`tools.RetryWithBackoff`, `ScanService.checkCreateSBOM`'s circuit-breaker cache, per-component
  fallback metrics).
- **Misleading diagnostics**: the final returned/logged error always frames the failure as an
  authorization problem (wrapping the original 401), even when the actual, more recent failure
  was unrelated — making it harder for an operator to diagnose a real registry outage or
  rate-limit incident from kubevuln's logs/scan-failure reason.
- **Shared by both SBOM paths**: `ResolveSource` is the one copy `adapters/v1/syft.go` and
  `pkg/sbomscanner/v1/server.go` both call (that consolidation, replacing two independently
  drifting copies, is what #685/#686 fixed) — this bug affects every pull that goes through
  either path.

# Affected Components

- `internal/registryauth/resolve.go`: `ResolveSource` (~lines 50-85), specifically the fallback
  condition at line 71 and the error-wrapping at lines 81-82.
- Exercised by `adapters/v1/syft.go:238` (in-process) and `pkg/sbomscanner/v1/server.go:46`
  (sidecar) — no caller-side changes should be needed.

# Acceptance Criteria

- The anonymous-credentials fallback (lines 71-84) only runs when the credentialed retry's own
  error is *also* a 401 (mirroring line 50's own check), or when no provider matched / no
  credentials were available in the first place (the existing lines 53-58 branch, which already
  falls through into line 71 correctly since `err` in that case is still the original 401 from
  line 47/39).
- A non-401 error from the credentialed retry is returned as-is (still going through
  `metrics.RecordScanFallback`/`RecordSourceResolution` bookkeeping so those metrics keep
  reflecting what actually happened), without a wasted anonymous `get` call and without being
  mislabeled as an authorization failure.
- The existing, correct behavior is preserved: a credentialed retry that fails with another 401
  (bad/refused credentials) still falls back to anonymous exactly as before.
- New unit tests (this function is trivially testable via its `Getter[T]` parameter, no fake
  registry needed) covering: credentialed retry fails with a non-401 error (no anonymous
  fallback, real error returned); credentialed retry fails with 401 again (anonymous fallback
  still happens, existing behavior preserved); the pre-existing MANIFEST_UNKNOWN and
  no-credentials-available paths are unaffected.

# Solution Direction

Tighten the condition at `internal/registryauth/resolve.go:71` to also check
`strings.Contains(err.Error(), "401 Unauthorized")`, matching the pattern already used at line
50 and matching what the function's own doc comment already describes as the intended behavior.
This is a small, single-function, well-isolated fix with a clear, deterministic test surface via
the existing `Getter[T]` abstraction — no registry, no gRPC, no fakes beyond a plain closure.

bhuvan-somisetty:
## Overview

`ResolveSource` (`internal/registryauth/resolve.go`) is the shared pull-fallback ladder both
SBOM paths (`adapters/v1/syft.go`, `pkg/sbomscanner/v1/server.go`) use. When the initial,
credential-less pull fails with `401 Unauthorized`, it looks up a matching credential provider
(ECR/GCP), retries with those credentials, and — per its own doc comment — falls back to an
anonymous retry "if that provider has none or its credentials are refused too". The check that
decided whether to fall back to anonymous was a bare `err != nil`, not "refused too": any error
from the credentialed retry — a `429 Too Many Requests`, a `5xx`, a timeout — triggered the
anonymous fallback, not just another auth rejection.

**Why it matters:** each individual pull attempt (`get`) is itself wrapped in its own
`tools.RetryWithBackoff` with 429-handling by both real callers. So a credentialed retry that
hits a persistently-429ing registry already exhausts its own retry-with-backoff budget — and
this bug then triggered a *second* full retry-with-backoff cycle, anonymously, doubling request
volume against a registry that had just asked to be backed off from. The final error was also
always wrapped around the *original* pre-credentials 401/403, so a scan that actually failed
because of a transient 500 or a rate limit got permanently logged as "unauthorized" —
misleading for anyone diagnosing a real registry incident from kubevuln's logs.

## Root Cause

`internal/registryauth/resolve.go`'s fallback-to-anonymous guard checked only `err != nil`.

## Solution

Add `isAuthDenied(err)`, checking for `"401 Unauthorized"` or `"403 Forbidden"` — the same
bucket `core/services/scan_failure_reasons.go`'s `classifySBOMError` already uses for this exact
distinction (401 and 403 both mean "the registry looked at who's asking and said no"; a 429/5xx/
timeout means something else went wrong that swapping credentials can't fix) — and gate the
anonymous fallback on it instead of on any error at all.

I initially scoped this to 401 only, then found `pkg/sbomscanner/v1/server_test.go` already has
a passing test (`"GCP image, ADC succeeds but pull fails non-401, still anonymous..."`) that
depends on a `403 Forbidden` from the credentialed retry *also* triggering the anonymous
fallback. Widening to the same 401/403 bucket `classifySBOMError` already uses keeps that
existing, intentional behavior while fixing the actual bug (429/5xx/timeout wrongly triggering
it).

The outer gate that decides whether to enter the ladder at all (a 401 on the very first,
credential-less attempt) is unchanged — this is purely about the inner re-check after the
credentialed retry.

## How to Test

```
go test ./internal/registryauth/... -run ResolveSource -v
go test ./pkg/sbomscanner/... -run TestResolveSource -v
```

New tests in `internal/registryauth/resolve_test.go`:
- `TestResolveSource_DoesNotFallBackToAnonymousOnNonAuthErrorFromCredentialedRetry` — a
  credentialed retry that fails with `429 Too Many Requests` returns that error as-is, with no
  third (anonymous) attempt.
- `TestResolveSource_FallsBackToAnonymousWhenCredentialedRetryAlsoGets401` — guards the
  preserved behavior: credentials outright refused (401 again) still fall back to anonymous.

Both use `ResolveSource`'s existing `Getter[T]` parameter to exercise the ladder deterministically, no fake registry needed — following the pattern already established by this test file's other cases (and `GCPCredsFn`/`ResetCaches()`, already used elsewhere in this package for the same reason).

All pre-existing tests in both packages pass unchanged, including the `pkg/sbomscanner/v1`
table-driven `TestResolveSource` case that specifically exercises the 403-still-falls-back
path.

`go build ./...`, `go vet ./internal/registryauth/... ./adapters/v1/... ./pkg/sbomscanner/...`,
and the full `internal/registryauth`, `adapters/v1`, and `pkg/sbomscanner` suites all pass
locally.

## Impact

- Removes a retry-amplification bug that worked directly against this codebase's existing
  429-handling investment (`tools.RetryWithBackoff`, `ScanService.checkCreateSBOM`'s circuit
  breaker, per-component fallback metrics).
- Real, unrelated errors from a credentialed pull (5xx, timeout, DNS) are now returned as-is
  instead of being buried behind a misleading "anonymous fallback failed" wrapper around the
  original auth error.
- No change to the existing, intentional 401/403-triggers-anonymous-fallback behavior, and no
  change to the outer gate that decides whether to enter the ladder at all.

## Related issues/PRs:

Fixes #921

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved registry access handling when authentication fails.
  * Prevented unnecessary anonymous retries for rate limits, server errors, and timeouts.
  * Preserved anonymous fallback when access is explicitly denied.
  * Ensured non-authentication errors are returned accurately instead of being mislabeled as unauthorized.

### DIFF
--- a/internal/registryauth/resolve.go
+++ b/internal/registryauth/resolve.go
@@ -16,6 +16,16 @@ import (
 // which credentials to try, not about what comes back.
 type Getter[T any] func(ctx context.Context, ref string, opts *image.RegistryOptions) (T, error)
 
+// isAuthDenied reports whether err is a registry rejecting the request as unauthenticated or
+// unauthorized -- the two outcomes credentials (ours, or none at all) can actually change.
+// Matches the same bucket core/services/scan_failure_reasons.go's classifySBOMError already
+// uses for the identical distinction (401 vs. 403 both mean "the registry looked at who's
+// asking and said no", where a 429/5xx/timeout means something else went wrong that swapping
+// credentials cannot fix).
+func isAuthDenied(err error) bool {
+	return err != nil && (strings.Contains(err.Error(), "401 Unauthorized") || strings.Contains(err.Error(), "403 Forbidden"))
+}
+
 // ResolveSource pulls imageID, falling back in order when the pull fails:
 //
 //   - MANIFEST_UNKNOWN, which a registry returns for a digest it does not hold, is retried
@@ -67,8 +77,21 @@ func ResolveSource[T any](ctx context.Context, component string, get Getter[T], 
 			}
 		}
 		// If no provider matched, its credentials were unavailable, or it succeeded in auth
-		// but still got 401, fall back to anonymous access. err/src retain the last attempt.
-		if err != nil {
+		// but still got denied, fall back to anonymous access. err/src retain the last attempt.
+		//
+		// The isAuthDenied check matters: when a provider's credentials were used above, err
+		// may now hold whatever the credentialed retry actually failed with, not another
+		// auth rejection -- a 429, a 5xx, a timeout. None of those are fixed by dropping
+		// credentials and retrying anonymously (the registry was never rejecting the request
+		// for being unauthenticated), so without this check a persistently-429ing registry
+		// gets a second full attempt -- itself wrapped in its own retry-with-backoff by every
+		// caller -- doubling load on a registry that just asked to be backed off from, and the
+		// eventual error permanently mislabels whatever actually went wrong as "unauthorized"
+		// (via unauthorizedErr below). When no provider matched or its credentials were
+		// unavailable, err is untouched here and still holds the original auth-denied error
+		// from line 50 (or the MANIFEST_UNKNOWN retry above it), so this check doesn't change
+		// that path's existing behavior. See #921.
+		if isAuthDenied(err) {
 			logger.L().Debug("retrying without credentials",
 				helpers.String("imageID", imageID))
 			opts.Credentials = nil
--- a/internal/registryauth/resolve_test.go
+++ b/internal/registryauth/resolve_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"testing"
+	"time"
 
 	"github.com/anchore/stereoscope/pkg/image"
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Reviewed the fallback guard and regression coverage. The credentialed retry now proceeds to anonymous only for repeated 401/403 auth denial, while preserving non-auth errors such as 429 without an extra retry. Focused registryauth and sbomscanner tests and go vet pass locally.

PR #925, opened 2026-09-01
fix(vexvalidate): reject a statement with no vulnerability, and empty subcomponents

### LINKED ISSUE #924: vexvalidate accepts a statement with no vulnerability identity, and subcomponents that identify nothing
`internal/vexvalidate` (#880) exists to catch a fetched document that is valid JSON but not a meaningful VEX document. Two shapes get through that fit that description.

**A statement that never says which vulnerability it is about.**

```json
{"@context":"https://openvex.dev/ns/v0.2.0",
 "statements":[{"status":"fixed","products":[{"@id":"pkg:oci/x"}]}]}
```

```
no vulnerability id or name at all -> ACCEPTED
```

`Statement.Validate()` checks the status and the fields each status may carry, but never that `vulnerability` identifies anything, so nothing in the chain does. This is the one that matters for #387: matching keys on the vulnerability name (`vexStatementKey{name, purl}`), so an ingested statement with no name keys on the empty string, and several of them collide with each other.

**A product whose subcomponents identify nothing.**

```json
"products":[{"@id":"pkg:oci/x","subcomponents":[{}]}]
```

```
empty product (caught today)                 -> rejected: statement has a product that identifies nothing
product ok, subcomponent identifies nothing  -> ACCEPTED
product ok, subcomponents all empty          -> ACCEPTED
```

`productIsEmpty` applies the ID/hashes/identifiers test to the product but not to its subcomponents, so the same emptiness one level down passes. Subcomponents are what scope a statement to a package, which is exactly the granularity `scopedToSubcomponent` matches on, so an empty one is as meaningless as an empty product.

Both are the same check the package already makes, applied one level further, plus one it does not make yet.

This is your package and you know the intent behind it better than I do, so say if you would rather pick it up. Otherwise happy to send a PR.

AnshulPatil2005:
## Overview

Two shapes got past `vexvalidate`, whose job is catching a fetched document that is valid JSON but not a meaningful VEX document.

**A statement that never says which vulnerability it is about.**

```json
{"@context":"https://openvex.dev/ns/v0.2.0",
 "statements":[{"status":"fixed","products":[{"@id":"pkg:oci/x"}]}]}
```

`go-vex`'s `Statement.Validate()` checks the status and the fields each status may carry, but not this, so nothing in the chain did. It matters for #387 because matching keys on the vulnerability name (`vexStatementKey{name, purl}`): an unnamed statement keys on the empty string and collides with every other unnamed one.

Either field identifies it, `@id` or `name`, so only having neither is rejected. A whitespace-only name counts as neither.

**A product whose subcomponents identify nothing.** `productIsEmpty` applied the ID/hashes/identifiers test to the product but not one level down, so `"subcomponents":[{}]` passed while `"products":[{}]` was caught. Subcomponents are what scope a statement to a package, the granularity `scopedToSubcomponent` matches on, so an empty one says as little as an empty product.

It is now `componentIsEmpty` running against both, since go-vex uses the same `Component` type for a product and a subcomponent.

Having no subcomponents at all is product scope, which is legitimate and stays valid, pinned by its own test.

## Testing

`go test ./internal/vexvalidate/ -count=1 -cover` — passes, coverage stays at 100%.

Tests follow the package's existing `errors.Is` style rather than introducing testify.

Both mutations caught:

| mutation | fails |
| --- | --- |
| drop the vulnerability-identity check | `..._StatementWithoutVulnerability_Fails`, `..._VulnerabilityIdentifiedByEitherField` |
| drop the subcomponent check | `..._EmptySubcomponent_Fails` |

`golangci-lint run ./internal/vexvalidate/...` exits 0 locally.

## Note

This is @Shreya2005-2005's package from #880, and I said on the issue she should have first refusal. Opening this so it is ready rather than to pre-empt her: happy to close it if she would rather take it, and no offence taken either way.

## Related issues/PRs:

* Resolved #924

## Summary by CodeRabbit

* **Bug Fixes**
  * Validation now rejects vulnerability statements without an identified vulnerability.
  * Empty subcomponents and blank identity values are rejected during validation.
  * Vulnerabilities and subcomponents can be identified by name, ID, or hash.
  * Product-level statements without subcomponents remain valid.

* **Tests**
  * Added coverage for missing and whitespace-only identifiers, empty subcomponents, supported identity fields, blank values, and valid product-level statements.

### DIFF
--- a/internal/vexvalidate/vexvalidate.go
+++ b/internal/vexvalidate/vexvalidate.go
@@ -51,6 +51,21 @@ var (
 	// only a product with none of the three says nothing about what it
 	// actually is.
 	ErrEmptyProduct = errors.New("vexvalidate: statement has a product that identifies nothing")
+
+	// ErrNoVulnerability is returned when a statement never says which
+	// vulnerability it is about. go-vex's Statement.Validate() checks the
+	// status and the fields each status may carry, but not this, so a
+	// statement carrying neither an @id nor a name passes every other check
+	// here. Matching keys on the vulnerability name, so an unnamed statement
+	// keys on the empty string and collides with every other unnamed one.
+	ErrNoVulnerability = errors.New("vexvalidate: statement does not identify a vulnerability")
+
+	// ErrEmptySubcomponent is returned when a product lists a subcomponent
+	// that identifies nothing, by the same ID/hashes/identifiers test
+	// ErrEmptyProduct applies one level up. Subcomponents are what scope a
+	// statement to a package, which is the granularity suppression matches
+	// on, so an empty one says as little as an empty product.
+	ErrEmptySubcomponent = errors.New("vexvalidate: statement has a subcomponent that identifies nothing")
 )
 
 // openVEXContextPrefix is the real OpenVEX namespace URI. The spec allows a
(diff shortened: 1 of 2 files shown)


matthyx (changes requested):
The new vulnerability and subcomponent checks are directionally right, and the full test suite passes, but there is one remaining validation gap before this is safe to merge: componentIsEmpty treats any non-empty hashes or identifiers map as identity, even when every value is empty or whitespace-only. For example, a product or subcomponent containing {"hashes":{"sha256":"   "}} or {"identifiers":{"purl":"   "}} still passes Validate while identifying nothing. Please trim and inspect the ID and map values, and add regression cases for both ErrEmptyProduct and ErrEmptySubcomponent. This is the same underlying malformed-placeholder class this PR is intended to close.

matthyx (approved):
Re-checked at 74a4a6e. The prior blocker is addressed: componentIsEmpty now trims the ID and requires at least one non-blank hash or identifier value, with regression coverage for both products and subcomponents. I also ran go test ./... successfully locally. The current cross-platform CI failure is in the unrelated Test_grypeAdapter_Ready_recoversFromStuckWarmUpdate concurrency assertion (expected 2 loads, observed 3), so that workflow should be rerun before merge.

Issue #923, opened 2026-09-01
a clean image publishes a VEX document that our own validator rejects


AnshulPatil2005:
A scan that finds nothing still publishes a VEX document, and that document is one `internal/vexvalidate` refuses.

`createVEX` builds its statements by ranging over `cve.Content.Matches` and `IgnoredMatches`. With no findings, neither loop appends, so `Statements` stays nil and the stored document is:

```json
{"@context":"https://openvex.dev/ns/v0.2.0","@id":"14e85ecc…","author":"kubescape.io",
 "timestamp":"…","last_updated":"…","version":0,
 "tooling":"kubescape-vulnerability-analyzer","statements":null}
```

Run through our own validator:

```
our own validator REJECTS it: vexvalidate: document has no statements
go-vex parses it fine
```

`go-vex` is happy, so nothing breaks today. It matters for #387: once kubevuln consumes VEX from a URL, the natural sources include documents kubescape itself published, and a clean image's document is refused by the very check added in #880 to decide whether a fetched document is trustworthy.

Two things are tangled here and they have different answers.

**The shape.** `Statements` has no `omitempty`, so an empty document serialises `"statements": null` rather than `[]`. A consumer expecting an array gets null. This one is unambiguous and cheap.

**The semantics.** `ErrNoStatements` says "a VEX document that asserts nothing is not useful to anything that would consume it", which is a fair rule for a fetched vendor feed. But "we scanned this image and found nothing" is a real assertion, and it is the one kubevuln makes most often on a healthy cluster. Either the validator should accept an empty document, or kubevuln should not publish one, and I do not think that is my call to make.

Raising it rather than opening a PR, because the second half is a design decision. Happy to implement whichever direction you prefer.

PR #926, opened 2026-09-01
fix(repositories): emit an empty statements array, not null, for a clean image

AnshulPatil2005:
## Overview

A scan that finds nothing still publishes a VEX document, and it carried `"statements": null`:

```json
{"@context":"https://openvex.dev/ns/v0.2.0","@id":"14e85ecc…","author":"kubescape.io",
 "timestamp":"…","version":0,"tooling":"kubescape-vulnerability-analyzer","statements":null}
```

`createVEX` builds its statements by ranging over `Matches` and `IgnoredMatches`, so with no findings neither loop appends and `Statements` stays nil. The field has no `omitempty`, so nil serialises as `null`.

`go-vex`'s own `vex.New()` emits `[]`:

```
{"@context":"https://openvex.dev/ns/v0.2.0","@id":"","author":"Unknown Author","timestamp":"…","version":1,"statements":[]}
```

A consumer expecting an array should get one, which matters more once #387 means these documents are read back.

## Why this is safe

The canonical hash is unaffected. It sorts a copy of the statements and length-prefixes each, so nil and empty produce the same digest, verified and pinned by `TestCalculateVexCanonicalHash_NilAndEmptyStatementsAgree`. If they differed, this change would re-hash every stored document and bump its version for no real change.

Only the creation path is touched. `updateVEX` deep-copies the stored spec, so a document already carrying `null` keeps it rather than being rewritten on the next scan, which would otherwise be a one-time version bump across every clean image in the cluster. Existing documents are left alone; new ones get the array.

## Scope

This is the unambiguous half of #923. The other half, whether a document that asserts nothing should be published at all, or whether `vexvalidate` should accept one, is a design decision and stays open on the issue. `Refs` rather than `Resolved` for that reason.

Worth being explicit: this change alone does **not** make a clean-image document pass `vexvalidate`. `len()` is zero either way, so `ErrNoStatements` still applies. It fixes the shape, not the verdict.

## Testing

`go test ./repositories/ -count=1` passes. `golangci-lint run ./repositories/...` exits 0 locally.

Removing the initialiser fails `TestAPIServerStore_StoreVEX_EmptyDocumentUsesAnArray`, which asserts on the marshalled document rather than the in-memory slice, since `null` versus `[]` is only visible once serialised.

## Related issues/PRs:

* Refs #923

### DIFF
--- a/repositories/apiserver.go
+++ b/repositories/apiserver.go
@@ -1752,6 +1752,10 @@ func (a *APIServerStore) createVEX(ctx context.Context, cve domain.CVEManifest, 
 
 	// Calculate VEX
 	vexDoc := v1beta1.VEX{
+		// Explicitly empty rather than nil: Statements has no omitempty, so a scan that
+		// found nothing would otherwise serialise "statements": null. go-vex's own
+		// vex.New() emits [], and a consumer expecting an array should get one. See #923.
+		Statements: []v1beta1.Statement{},
 		Metadata: v1beta1.Metadata{
 			Context:     "https://openvex.dev/ns/v0.2.0",
 			Author:      "kubescape.io",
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed the creation path, serialization assertion, and canonical-hash regression coverage. Initializing Statements to an empty slice correctly changes clean-image JSON from null to [] without changing existing update behavior or canonical hashes. No blockers found; current checks are green.

PR #930, opened 2026-09-02
fix(adapters): gate testcontainers-go test fixture behind nodockerfixture build tag (#929)

### LINKED ISSUE #929: Test-only Docker-container fixture in adapters/v1/grype.go leaks testcontainers-go/docker into the production binary, causing false-positive CVE findings
# Description

`probeDockerAvailable`, `startGrypeOfflineDBContainer`, `NewGrypeAdapterFixedDB`, and
`NewGrypeAdapterFixedDBWithMatchers` (`adapters/v1/grype.go:164-243`) are a Docker-container-
backed *test* fixture — they spin up `quay.io/kubescape/grype-offline-db:v6-ci-only` via
`testcontainers-go` so `TestGrypeAdapter_*`/`TestScanService_*` tests can run against a real,
pinned vulnerability DB instead of the network. But they live in `adapters/v1/grype.go`, an
ordinary (non-`_test.go`) source file, so they're compiled into every build of the package —
including the production `kubevuln` binary, which never calls any of them.

That pulls `github.com/testcontainers/testcontainers-go` and its dependency tree —
`github.com/docker/docker`, `github.com/docker/docker-credential-helpers`, and everything
those need to talk to a Docker daemon — into the shipped binary for zero runtime benefit.
Two currently-unfixed CVEs in that tree are consequently flagged, permanently, on the
production-binary vulnerability scan #855 added.

# Reproduction / Evidence

1. The only callers of these four functions, anywhere in the repo:
   ```
   $ grep -rln "NewGrypeAdapterFixedDB\b\|NewGrypeAdapterFixedDBWithMatchers\b" --include=*.go .
   ./adapters/v1/grype.go
   ./adapters/v1/grype_test.go
   ./core/services/scan_test.go
   ```
   Never `cmd/http/main.go`, `cmd/sbom-scanner/main.go`, or any other production entry point —
   `NewGrypeAdapter` (the one production constructor, `adapters/v1/grype.go:106`) doesn't touch
   Docker at all.

2. `docker/docker/client` — the specific package both CVEs' vulnerable symbols live in — is
   nonetheless in the production build's dependency graph, reached through exactly one path:
   ```
   $ go mod why github.com/docker/docker/client
   # github.com/docker/docker/client
   github.com/kubescape/kubevuln/adapters/v1
   github.com/testcontainers/testcontainers-go
   github.com/docker/docker/client
   ```
   (`core/domain` separately imports the much smaller `docker/docker/api/types/registry` for
   unrelated registry-credential type definitions — that's a different subpackage, carries
   neither CVE's vulnerable symbol, and isn't what this issue is about.)

3. CI's own `govulncheck` workflow (added in #855) confirms both are actually present in the
   compiled binary it scans (`make govulncheck` runs `govulncheck -mode=binary $(BINARY_NAME)`
   against the real production build):
   ```
   $ gh run view 33486992769 --log | grep -A8 "Vulnerability #"
   Vulnerability #1: GO-2026-4887
     Moby has AuthZ plugin bypass when provided oversized request bodies
     Module: github.com/docker/docker
     Found in: github.com/docker/docker@v28.5.2+incompatible
     Fixed in: N/A
     Vulnerable symbols found: client.init, network.init, runtime.init

   Vulnerability #2: GO-2026-4883
     Moby has an Off-by-one error in its plugin privilege validation
     Module: github.com/docker/docker
     Found in: github.com/docker/docker@v28.5.2+incompatible
     Fixed in: N/A
     Vulnerable symbols found: client.init, network.init, runtime.init
   ```

# Root Cause

A test-only fixture is defined in a regular source file instead of a `_test.go` file (or a
file gated behind a build tag), so the Go toolchain has no way to distinguish it from
production code: `go build` for `cmd/http` compiles `adapters/v1` as a whole, `testcontainers-go`
import included, whether or not anything in the production call graph ever reaches it.

# Impact

- **False-positive security signal, permanently red.** `govulncheck -mode=binary` reports
  presence-in-binary, not actual reachability from a running program's entry points — so these
  two findings will sit in CI forever with no code path that could ever trigger them, making the
  govulncheck job noisier and easier to start ignoring — exactly the failure mode a
  vulnerability-scanning tool is supposed to prevent. (#854, already open, is the inverse case:
  a genuinely reachable CVE in the actual scanning path. This is not that — it's compiled-in but
  unreachable, and it's *avoidable*, unlike #854, which is blocked on an upstream fix.)
- **Recurring, not one-off**: `testcontainers-go`'s own transitive tree (Docker, and
  Docker's own dependencies) is large and under active development; it will keep surfacing new
  CVEs the same way, for a binary that never actually talks to Docker in production.
- **Binary bloat**: the shipped `kubevuln` binary carries an entire Docker client + credential-
  helper stack it has no use for.

# Affected Components

- `adapters/v1/grype.go`: `probeDockerAvailable`, `startGrypeOfflineDBContainer`,
  `NewGrypeAdapterFixedDB`, `NewGrypeAdapterFixedDBWithMatchers` (lines ~164-243)
- `adapters/v1/grype_test.go`, `core/services/scan_test.go` — the only real callers
- `.github/workflows/govulncheck.yaml` / `make govulncheck` — where the false positive
  actually surfaces

# Acceptance Criteria

- `go list -deps ./cmd/http/...` (and `./cmd/sbom-scanner/...`) no longer includes
  `github.com/docker/docker/client` or `github.com/testcontainers/testcontainers-go`.
- `go build ./cmd/http` / `go build ./cmd/sbom-scanner` still succeed unmodified — no change to
  how the production binaries are built.
- `go test ./adapters/v1/... ./core/services/...` (the packages that actually use this fixture)
  still build and pass with whatever tag/invocation the CI test job already uses today (or with
  a documented, minimal addition to `make test`/CI if a build tag is introduced).
- Re-running `govulncheck -mode=binary` against the production binary no longer reports
  GO-2026-4887 or GO-2026-4883 (or whatever CVEs are current in `docker/docker` at fix time —
  the point is the package is gone from the binary, not that these two specific IDs are chased).
- A comment at the fixture's definition explains why it's gated this way, so a future
  contributor doesn't "helpfully" move it back into an ungated file.

# Solution Direction

Move `probeDockerAvailable`, `startGrypeOfflineDBContainer`, `NewGrypeAdapterFixedDB`, and
`NewGrypeAdapterFixedDBWithMatchers` into their own file in `adapters/v1` (e.g.
`grype_testfixture.go`) gated behind a build tag (e.g. `//go:build testtools`, following the
convention `//go:build` tags for opt-in test-support code typically use). `grype_test.go` and
`core/services/scan_test.go` already only reference these from test files, so they can either
carry the same tag (if `go test` invocations already pass it, or are updated to) or the
production `go build` commands (`make build`, the two `Dockerfile`s if they call `go build`
directly) can stay tag-free by default while `make test`/CI test invocations add
`-tags testtools`. Either direction keeps the change small and mechanical — no logic changes,
just moving four functions and adding a build constraint plus updating however tests are
invoked to still see them.

bhuvan-somisetty:
### Problem
`probeDockerAvailable`, `startGrypeOfflineDBContainer`, `NewGrypeAdapterFixedDB`, and `NewGrypeAdapterFixedDBWithMatchers` in `adapters/v1/grype.go` are Docker-container-backed test fixtures used by `grype_test.go` and `scan_test.go`. Living in an ordinary source file, `testcontainers-go` and its dependency tree (`github.com/docker/docker/client` and dependencies) were compiled into every production `kubevuln` binary for zero runtime benefit. This resulted in false-positive vulnerability findings (GO-2026-4887 and GO-2026-4883) in CI binary security scans (`govulncheck -mode=binary`).

### Root Cause
Test-only fixture functions were defined in an ungated source file (`adapters/v1/grype.go`) instead of being isolated behind a build constraint or test-only package, causing the Go toolchain to include `testcontainers-go` in the dependency graph of `cmd/http/main.go`.

### Solution
- Isolated the Docker test fixture functions (`probeDockerAvailable`, `startGrypeOfflineDBContainer`, `NewGrypeAdapterFixedDB`, `NewGrypeAdapterFixedDBWithMatchers`) into `adapters/v1/grype_docker_fixture.go` gated with `//go:build !nodockerfixture`.
- Kept `ErrDockerUnavailable` in `adapters/v1/grype.go` so test callers can import it regardless of build tags.
- Updated `Makefile` build target and `build/Dockerfile` / `build/Dockerfile.debug` to pass `-tags nodockerfixture` when building production binaries (`cmd/http/main.go`), excluding `testcontainers-go` and `github.com/docker/docker/client`.
- Standard untagged `go test ./...` continues to compile and run tests with the fixture enabled.

### Testing
- Added regression test `TestProductionBuildExcludesDockerFixture` in `cmd/http/main_test.go` asserting via `go list -tags nodockerfixture -deps .` that `testcontainers-go` and `docker/docker/client` are excluded from production build dependencies.
- Verified standard untagged tests (`go test ./adapters/v1/...`) compile and pass.

Closes #929

## Summary by CodeRabbit

- **Builds**
  - Production builds now exclude Docker-based Grype test fixtures and related dependencies by default, resulting in a leaner shipped binary.
  - Docker-backed fixtures remain available through an explicit opt-in for test builds.

- **Testing**
  - Added validation that production builds exclude Docker-related dependencies.
  - Added coverage for offline Grype database versions, SBOM scan results, and container profile scanning.
  - Improved reliability when validating asynchronous resource cleanup.


bhuvan-somisetty:
Fixed added `_ = container.Terminate(ctx)` on the `MappedPort` error path in `startGrypeOfflineDBContainer` so started containers are terminated immediately and do not leak Docker resources if port mapping fails.

matthyx (changes requested):
Blocking: this implements the build constraint in the opposite direction from issue #929's acceptance criteria. On this head, untagged `go list -deps ./cmd/http` still includes `github.com/testcontainers/testcontainers-go` and `github.com/docker/docker/client`, so an unmodified `go build ./cmd/http` still embeds the dependencies. They disappear only when every production build remembers to add `-tags nodockerfixture`.

Issue #929 explicitly requires untagged `go list -deps ./cmd/http/...` to exclude these packages and `go build ./cmd/http` to succeed unmodified. Please make the fixture opt-in (for example, `//go:build dockerfixture`/`testtools`) and apply that same constraint to the tests that use it, then add the tag to the test invocation. That makes exclusion the safe default for all current and future production build paths instead of relying on an exclusion tag being propagated everywhere.

I verified that the targeted untagged tests pass and the tagged production dependency graph is clean; the existing MappedPort cleanup finding is fixed. Once the default build behavior matches #929, this should be ready for another look.

bhuvan-somisetty:
Good catch @matthyx! Switched the fixture to be opt-in via `//go:build dockerfixture`. Untagged builds (`go build ./cmd/http`) now strip `testcontainers-go` and `docker/docker/client` by default without any flags, and `make test` passes `-tags dockerfixture`. Ready for another look!

matthyx (changes requested):
One remaining blocker: `core/services/scan_docker_fixture_test.go` imports `github.com/kubescape/kubevuln/internal/tools/adapters`, which does not exist. As a result, `go test -tags dockerfixture ./...` (and therefore the newly updated `make test`) fails during package setup with `no matching versions for query "latest"`.

Please change that import to `github.com/kubescape/kubevuln/adapters`, matching the original test and the other call sites. I applied that one-line correction locally and verified the full tagged test suite plus tagged `go vet ./...` pass. Untagged tests/builds also pass, and the untagged production dependency graphs for both `cmd/http` and `cmd/sbom-scanner` are clean, so this should be ready to approve after the import fix.

core/services/scan_docker_fixture_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	helpersv1 "github.com/kubescape/k8s-interface/instanceidhandler/v1/helpers"
+	"github.com/kubescape/kubevuln/adapters"
+	v1 "github.com/kubescape/kubevuln/adapters/v1"
+	"github.com/kubescape/kubevuln/core/domain"
```
bhuvan-somisetty, line 13: Good catch, thanks — that was a leftover from an earlier refactor. Fixed the import to point at `github.com/kubescape/kubevuln/adapters` and verified `go test -tags dockerfixture ./...` and `go vet -tags dockerfixture ./...` both pass now, along with the untagged build/tests. Should be good for another look.

matthyx (approved):
Rechecked at `cea8deca`; the import blocker is fixed and I found no remaining merge blocker. Verified the full tagged and untagged test suites, tagged and untagged vet, race-enabled tests for the affected packages, unmodified production builds, and clean production dependency graphs for both binaries.

Issue #429, opened 2026-08-03
Bug: SyftAdapter.CreateSBOM serializes all image scans globally via single-threaded mutex, ignores caller context, and calls logger.Fatal (os.Exit) on temp dir cleanup errors


bhuvan-somisetty:
# Description

`SyftAdapter.CreateSBOM` in `adapters/v1/syft.go` handles in-process SBOM generation for container images. However, the current implementation contains critical concurrency, reliability, and context propagation flaws:

1. **Global Unbounded Mutex Bottleneck (`s.pullMutex`)**: `s.pullMutex.Lock()` is acquired around `dl.Run(...)` during SBOM generation. Because `s.pullMutex` is a single global mutex on `SyftAdapter`, all concurrent SBOM generation requests across the entire service are forced into a strictly serialized, single-threaded execution queue. When multiple container image scan jobs arrive simultaneously in a cluster (e.g. during a deployment or scheduled scan), every scan must wait for prior images to finish (up to 15 minutes per image), introducing massive processing latency and widespread request timeouts.
2. **Self-Inflicted Process Termination via `logger.Fatal`**: In the `defer` block responsible for closing `source.Source`, `logger.L().Ctx(ctx).Fatal("failed to close source", ...)` is called if `src.Close()` fails. `logger.Fatal` invokes `os.Exit(1)`, which immediately crashes the entire `kubevuln` process on minor temporary directory cleanup issues (e.g., file locks or temporary file deletion errors on Linux/Windows).
3. **Caller Context Discarded (`context.Background()`)**: Both `syft.GetSource` and `syft.CreateSBOM` are passed `context.Background()` instead of the caller-supplied `ctx` or deadline context. This breaks request cancellation, HTTP timeouts, and distributed tracing context across the SBOM generation layer.

# Environment

- OS: Linux / Kubernetes (All supported distributions)
- Version: `kubevuln` main / latest release

# Steps To Reproduce

1. Deploy `kubevuln` and trigger concurrent container image scan requests (`/v1/scan/container` or `/v1/generateSBOM`).
2. Observe that SBOM generation requests execute strictly one-at-a-time due to `s.pullMutex.Lock()`, causing subsequent scan requests to queue and time out.
3. Simulate a temporary directory file lock or cleanup failure during `src.Close()` and observe `kubevuln` process termination (`logger.Fatal` -> `os.Exit(1)`).

# Expected behavior

[...]

PR #942, opened 2026-09-06
fix(adapters): bound SyftAdapter.CreateSBOM's pull-semaphore wait to the caller's timeout

### LINKED ISSUE #941: SyftAdapter.CreateSBOM's pullMutex has no timeout on acquisition: one stuck cataloguer permanently wedges every future scan, with no liveness/readiness signal
## Problem

`SyftAdapter.CreateSBOM` (`adapters/v1/syft.go`) is the in-process SBOM generation path — the
one actually used for every scan whenever the sbom-scanner sidecar is disabled or not yet ready
(`cmd/http/main.go:83-101`), which includes any deployment that doesn't set
`SBOM_SCANNER_SOCKET` at all. It serializes all pulls/cataloging behind a single
`sync.Mutex` (`s.pullMutex`), an intentional, maintainer-confirmed design choice for bounding
disk usage (#429/#430 closed "by design").

`s.pullMutex.Lock()` at `adapters/v1/syft.go:216` is a bare, blocking call with no timeout and
no `select` on any context:

```go
var longWaitWarning *time.Timer
if s.scanTimeout > 0 {
    longWaitWarning = time.AfterFunc(s.scanTimeout, func() {
        logger.L().Ctx(ctx).Warning("waiting unusually long to acquire pullMutex; a previous scan may be stuck", ...)
    })
}
s.pullMutex.Lock()
```

`ctxWithTimeout` (the `scanTimeout`-bounded context the function computes at line 186) is never
consulted here — it only takes effect *after* the lock is acquired, inside
`registryauth.ResolveSource` (line 238) and the cataloging `dl.Run` closure (line 313-353).

Ownership of `unlockPullMutex()` is deliberately handed to that same `dl.Run` closure
(line 314: `defer unlockPullMutex()`), specifically so the mutex keeps modeling "cataloging is
actually in progress on disk" even past a `scanTimeout` (this is the fix for #687). But Syft's
cataloguers do not support cancellation — the code's own comment on lines 340-344 states so
explicitly — and `deadline.Run` (`github.com/eapache/go-resiliency/deadline`) "does not (and
cannot) kill the running function's goroutine". The surrounding comment (lines 199-208)
explicitly acknowledges the consequence:

```go
// That's an accepted tradeoff, not a bug: there is no watchdog that force-releases the
// mutex, so a sufficiently pathological image (corrupt archive, cataloguer bug, stalled
// local FS) can keep it held indefinitely. This log line is the operator-visible signal
// for that case, since neither the liveness nor readiness probe currently reflects it.
```

That comment undersells the consequence: it isn't "one scan pends unusually long". Because
`s.pullMutex.Lock()` has no timeout, once any single scan's Syft goroutine genuinely never
returns, every future `CreateSBOM` call in the process — for every future image, on every
worker — blocks on that same `Lock()` forever. The individual timed-out task does return
(`dl.Run` still reports `ErrTimedOut` after `scanTimeout`, so the worker-pool task itself
completes and the worker is freed — see `errors.Is(err, deadline.ErrTimedOut)` at line 355), but
the next task picked up by that (or any) freed worker immediately re-blocks on the still-held
mutex. With the default `scanConcurrency: 1` (`config/config.go:100`), a single pathological
image permanently wedges 100% of the pod's scanning capacity.

## Why nothing catches this

- `ScanService.Ready` (`core/services/scan.go:199-201`) proxies only
  `s.cveScanner.Ready(ctx)` (the Grype DB). It has no visibility into `sbomCreator` at all.
- `HTTPController.Alive` (`controllers/http.go:403-405`) is an unconditional `200 OK`.
- `WithMaxQueueDepth` defaults to `0` (unbounded, `config/config.go:106`), and even where set,
  `workerpool`'s internal queue is an unbounded deque (per #748's own analysis) — new scan
  requests keep being accepted and queued (each holding registry credentials until it runs, per
  #748) with nothing ever completing or erroring, rather than failing fast.
- The only operator-visible signal is the one `time.AfterFunc` warning log per stuck caller
  (line 211-214) — there is no metric, no readiness degradation, and no automatic recovery.
  Kubernetes therefore has no signal to restart the pod, and it never will on its own.

## Reproduction / Evidence

`SyftAdapter.CreateSBOM` is already structured for this to be exercised directly (it's an
exported method taking `createSBOMFn` as a package-level var, exactly as `adapters/v1/syft_test.go`
already does to fake `syft.CreateSBOM`):

1. Construct a `SyftAdapter` with `scanTimeout` set short (e.g. 200ms) for a fast test.
2. Replace `createSBOMFn` with a function that blocks on `<-make(chan struct{})` forever
   (models a genuinely hung/uncancellable Syft cataloguer — corrupt archive, decompression
   stall, cataloguer bug on a crafted layer).
3. Call `CreateSBOM` for image A. It returns `Incomplete` after `scanTimeout`, as designed —
   but the abandoned goroutine keeps `s.pullMutex` held (this half is already covered by
   #687's fix and its tests).
4. Call `CreateSBOM` again, concurrently, for a completely unrelated image B, passing a fresh
   `context.Background()` (i.e., no caller-side deadline of its own — matching how
   `bgCtx := context.WithoutCancel(...)` is actually constructed for every real handler in
   `controllers/http.go`).
5. Observe: the call for image B never returns — not after `scanTimeout`, not ever — because
   `s.pullMutex.Lock()` has nothing bounding the wait. Every further call behaves the same way.

In production this surfaces as: kubevuln's `/v1/liveness` and `/v1/readiness` both stay green,
`Diagnostics` looks normal, but every scan submitted after the first pathological image just
never completes (or reports its result) again — no error, no `ReportScanFailure`, no
`ScanStatus` transition to `failed`/`abandoned`, forever, until someone manually restarts the
pod. The worker-pool queue depth (if `WithMaxQueueDepth`/metrics are wired) climbs
monotonically as new requests pile up behind the wedged mutex.

## Root Cause

`adapters/v1/syft.go`:
- `s.pullMutex.Lock()` (line 216) has no timeout and is not selected against any context.
- The mutex's release is intentionally tied to Syft's cataloging goroutine actually finishing
  (line 314, fixing #687), and that goroutine is provably uncancellable (lines 340-344,
  citing the upstream Syft cancellation gap and `deadline.Run`'s own documented limitation).
- There is no watchdog anywhere that force-releases `pullMutex`, no metric tracking how long
  it has been held, and no wiring from this adapter into `ScanService.Ready`/`Alive`.

This is the same shape of bug as #900 (`GrypeAdapter`'s permanently-hung DB download latching
`updating=true` forever with no recovery and no readiness signal, fixed by adding an
attempt-token + bounded stuck-timeout + staleness signal) — just on the SBOM-generation path
instead of the CVE-DB path, and for a mutex instead of a boolean latch. #900 establishes that
maintainers treat "uncancellable third-party call can hang forever, with no operator-visible
signal or recovery" as a real bug worth fixing, not an acceptable tradeoff to leave purely as a
log line.

This is not a duplicate of #429/#430 (that was about bounding concurrency, rejected as
by-design for disk-space reasons) or of #687 (fixed: the mutex not actually covering the async
work at all). Both of those are prerequisites that are already correctly in place; this issue
is about the complete absence of a recovery path once the (accepted) single mutex gets stuck.

## Impact

- **Total, silent denial of service.** A single corrupt, adversarially crafted, or merely
  cataloguer-triggering image permanently disables 100% of a pod's SBOM/CVE scanning capacity
  (at the default `scanConcurrency: 1`), with zero operator-visible signal beyond one log line
  per stuck caller.
- **Not self-healing.** Unlike the `os.Exit(0)` cold-start path in `GrypeAdapter`, there is no
  restart trigger here at all — liveness and readiness both stay green indefinitely.
- **Security-adjacent.** Any workload whose image an attacker controls (a compromised
  registry, a malicious base image, a supply-chain-injected layer) is a viable trigger. Since
  kubevuln's own job is to scan untrusted image content, "an image that hangs the cataloguer"
  is not really a hypothetical.
- **Realistic trigger without any adversary.** kubevuln pods are long-lived; a stalled local
  filesystem, an NFS hiccup mid-extraction, or any real Syft cataloguer bug on a malformed
  (not necessarily malicious) layer is enough, and has already been anticipated by this file's
  own comments as a known possibility.
- **Compounds silently.** Because `maxQueueDepth` defaults to unbounded and `workerpool`'s
  internal queue has no cap, requests keep being accepted and queued rather than rejected,
  so there's no natural back-pressure signal either — see #748.

## Affected components

- `adapters/v1/syft.go` — `SyftAdapter.CreateSBOM` (`pullMutex` field, `Lock()` call at
  line 216, the `dl.Run` closure at 313-353 that owns the eventual `Unlock()`).
- `core/services/scan.go` — `ScanService.Ready` (line 199-201), which has no visibility into
  the SBOM creator's health.
- `controllers/http.go` — `HTTPController.Alive`/`Ready` handlers, which surface `Ready` as-is.

## Acceptance criteria

1. A `CreateSBOM` call that cannot acquire `pullMutex` within a bounded period (e.g.
   `scanTimeout`, mirroring the budget the rest of the function already uses) returns
   `Incomplete`/an error instead of blocking forever, the same way every other phase of this
   function already degrades on timeout.
2. The bound is honored via the caller's context where the caller supplies one, and via a
   fallback duration otherwise — matching the existing `ctxWithTimeout`/`longWaitWarning`
   machinery already present, rather than introducing a second, inconsistent timeout scheme.
3. A stuck cataloguer goroutine that eventually does return must not corrupt or be confused
   with a newer caller's state (mirroring the attempt-token discipline #900 introduced for
   `GrypeAdapter`) — late results should be discarded/closed rather than acted on.
4. An observable signal exists for "pullMutex has been held unusually long" beyond the current
   log line — e.g. a gauge such as `kubevuln_sbom_pull_mutex_held_seconds` or a counter of
   timed-out-waiting-for-mutex events, so this condition is alertable instead of requiring log
   archaeology.
5. Regression tests (using the existing `createSBOMFn` injection point, as `syft_test.go`
   already does) covering: (a) a permanently-hung cataloguer for image A does not permanently
   block `CreateSBOM` for image B; (b) the caller gets a bounded, reported failure instead of
   hanging; (c) a late-returning abandoned goroutine doesn't corrupt a subsequent successful
   scan's state.
6. Consider surfacing this condition through `ScanService.Ready`/`Diagnostics` (e.g. degrade
   readiness once a pull has been stuck past some large multiple of `scanTimeout`), so
   Kubernetes can eventually restart a genuinely wedged pod automatically — the same
   closing-the-loop step #900's acceptance criteria specified for the DB-update case.

## High-level solution direction

Bound the acquisition of `pullMutex`, not just the work performed after acquiring it:
replace the bare `s.pullMutex.Lock()` with a `select` between a lock-acquired signal (e.g. via
a buffered channel semaphore of size 1, or `TryLock` polled against `ctxWithTimeout.Done()`)
and the timeout context already computed at line 186. On timeout-while-waiting, return
`Incomplete` the same way the other branches already do, without ever touching the mutex.
Keep the existing "hand ownership of the eventual real `Unlock()` to the abandoned goroutine"
behavior (#687's fix) for whichever caller actually did acquire it — the fix here is purely
about not blocking other callers indefinitely behind a mutex that may never come free,
mirroring how #900 solved the analogous problem for `GrypeAdapter`'s `g.updating` with an
attempt-token instead of a plain boolean latch.

bhuvan-somisetty:
Fixes #941

## Problem

`SyftAdapter.CreateSBOM` (the in-process SBOM path used whenever the sbom-scanner sidecar is
disabled or not yet ready) serializes pulls/cataloging behind a single `sync.Mutex`
(`pullMutex`) — an intentional, already-accepted design for bounding disk usage (#429/#430).

`pullMutex.Lock()` was a bare, blocking call with no timeout and no context check. Syft's
cataloguers do not support cancellation, so once one call's cataloguing goroutine genuinely
hung (corrupt archive, cataloguer bug, stalled local filesystem — all explicitly anticipated in
this file's own comments), it held `pullMutex` forever. Every subsequent `CreateSBOM` call —
for every future image, on every worker — then blocked on that same `Lock()` indefinitely.
`ScanService.Ready` only reflects the CVE scanner's health, and `Alive` is a static 200, so
Kubernetes had no signal to ever detect or recover a pod in this state; only a manual restart
would fix it.

## Root cause

`adapters/v1/syft.go:216` (pre-fix): `s.pullMutex.Lock()` — no timeout, no `select`, no
awareness of `ctxWithTimeout` (the `scanTimeout`-bounded context the function had already
computed a few lines earlier). The timeout budget only ever applied to the phases *after* the
lock was acquired.

## Fix

Replace `pullMutex sync.Mutex` with `pullSem chan struct{}` (a capacity-1 semaphore: acquire by
sending, release by receiving). Acquisition is now a `select` between the semaphore and
`ctxWithTimeout.Done()`:

- If the wait ends because `ctxWithTimeout` hit `scanTimeout`, `CreateSBOM` returns
  `(Incomplete, nil)` and records the same `metrics.RecordScanFallback` fallback metric every
  other timeout branch in this function already uses — the same shape of degradation the rest
  of the function already has for every other phase.
- If the wait ends because the caller's own context was cancelled, `CreateSBOM` returns that
  cancellation error immediately, without waiting out the full `scanTimeout`.

The eventual release of `pullSem` is unchanged: ownership still transfers to the (possibly
abandoned) `dl.Run` closure exactly as #687 established, so a stuck goroutine's disk activity is
still correctly serialized against a later successful pull — this fix is purely about not
blocking *other* callers forever while waiting to acquire it in the first place.

No behavioral change for the common case (semaphore uncontended, or contended only briefly):
acquisition still just works. The change only kicks in once a wait would otherwise exceed the
caller's own timeout budget.

## Testing

- Two new regression tests in `adapters/v1/syft_test.go`:
  - `Test_syftAdapter_CreateSBOM_GivesUpWaitingForPermanentlyStuckPullSem`: a first call's
    cataloguing goroutine is made to hang forever (never releases `pullSem`); asserts a second,
    independent `CreateSBOM` call still returns `(Incomplete, nil)` within its own `scanTimeout`
    budget instead of hanging — this is the exact scenario #941 describes, and hangs forever
    against the pre-fix code (verified locally by reverting the fix and watching the test time
    out).
  - `Test_syftAdapter_CreateSBOM_CanceledWhileWaitingForStuckPullSem`: same stuck-first-call
    setup, but the second call's *own* context is cancelled well before `scanTimeout` (10s)
    elapses; asserts the cancellation is honored immediately (`errors.Is(err, context.Canceled)`)
    rather than only unblocking once the adapter's own timeout fires.
- All pre-existing `adapters/v1` tests still pass unmodified, including the #687 regression test
  (`Test_syftAdapter_CreateSBOM_SerializesWithAbandonedGoroutineAfterTimeout`), confirming the
  serialization guarantee that fix established is preserved.
- `go build ./...` and the exact `make build` cross-compile (`CGO_ENABLED=0 GOOS=linux
  GOARCH=amd64`).
- `go vet ./...` — clean.
- `gofmt -l` — clean.
- `golangci-lint run --new-from-rev=<merge-base with main>` (the same diff-based invocation
  `make lint`/CI use) — 0 issues.
- `go test ./adapters/... ./core/... ./controllers/... ./repositories/...` — all pass.
- Not run: `-race`. This machine has no C compiler available, so `-race` can't build locally
  here (matches the pre-existing gap tracked in #852 — CI doesn't run it either). The new tests
  don't rely on `-race` to be meaningful; they assert on wall-clock bounds and returned
  status/error, and pass deterministically without it.

## Impact

A single corrupt, adversarially crafted, or merely cataloguer-triggering image can no longer
permanently disable a pod's in-process SBOM-scanning capacity. A stuck scan still costs one
`scanTimeout` (or less, on cancellation) per caller queued behind it, exactly like every other
timeout path in this function — it just never becomes a permanent, silent, liveness-invisible
wedge.

## Summary by CodeRabbit

- **Bug Fixes**
  - SBOM scans now handle expired or cancelled contexts promptly, returning incomplete results or cancellation errors as appropriate.
  - Scans no longer remain blocked indefinitely when another scan holds the pull/cataloging semaphore.
  - Semaphore resources are released correctly when scan contexts expire during acquisition.

- **Observability**
  - Added a distinct `pull_semaphore` fallback metric category for scans blocked by an active pull/cataloging operation.

- **Documentation**
  - Updated API documentation for the new fallback metric category.


adapters/v1/syft.go
```diff
@@ -190,43 +195,64 @@ func (s *SyftAdapter) CreateSBOM(ctx context.Context, name, imageID, imageTag st
...
+	//
+	// pullSemTimeoutResult classifies ctxWithTimeout's error the same way every other timeout
+	// branch in this function already does: DeadlineExceeded degrades to Incomplete (recording
+	// the same fallback metric those branches use), anything else (caller cancellation) is
+	// returned as-is.
+	pullSemTimeoutResult := func(err error) (domain.SBOM, error) {
+		logger.L().Ctx(ctx).Warning("gave up waiting to acquire the pull semaphore; a previous scan may be stuck",
+			helpers.String("imageID", imageID))
+		if errors.Is(err, context.DeadlineExceeded) {
+			metrics.RecordScanFallback(ctx, metrics.ComponentInProcess, metrics.FallbackCategorySizeClassification, metrics.FallbackStrategyIncomplete, metrics.FallbackOutcomeClassified)
```
matthyx: Blocking: this generic `size_classification` / `incomplete` series is already emitted for other timeout paths, so it does not provide the semaphore-specific observable signal required by kubescape/kubevuln#941 acceptance criterion 4. Operators still cannot alert on or distinguish "timed out waiting for a held pull semaphore" from ordinary source/cataloguing timeouts. Please add a dedicated metric or dedicated bounded-cardinality category/strategy for semaphore-acquisition timeout, with metric coverage and the public metrics documentation updated.

matthyx (changes requested):
The bounded semaphore acquisition and cancellation handling look sound, and the focused tests (including `-race`) pass locally. One acceptance-criteria blocker remains inline.

adapters/v1/syft.go
bhuvan-somisetty: Added a dedicated FallbackCategoryPullSemaphore category on kubevuln_scan_fallbacks_total (distinct from the generic size_classification/incomplete series), recorded only when CreateSBOM gives up waiting for a semaphore held by a stuck previous scan. Covered by a metrics-package test asserting the two series render separately, and by an adapter-level test asserting the real CreateSBOM code path emits it. docs/API.md's metric table is updated. Pushed in 0fd450c.

adapters/v1/syft.go
```diff
@@ -190,43 +195,67 @@ func (s *SyftAdapter) CreateSBOM(ctx context.Context, name, imageID, imageTag st
...
+			helpers.String("imageID", imageID))
+		if errors.Is(err, context.DeadlineExceeded) {
+			metrics.RecordScanFallback(ctx, metrics.ComponentInProcess, metrics.FallbackCategoryPullSemaphore, metrics.FallbackStrategyIncomplete, metrics.FallbackOutcomeClassified)
+			domainSBOM.Status = helpersv1.Incomplete
+			return domainSBOM, nil
+		}
+		return domainSBOM, err
+	}
+	if err := ctxWithTimeout.Err(); err != nil {
+		return pullSemTimeoutResult(err)
```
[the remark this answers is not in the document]
bhuvan-somisetty: Fixed: the stuck-semaphore warning + FallbackCategoryPullSemaphore metric now only fire on the contended DeadlineExceeded path. Caller cancellation is logged separately at debug level with no 'previous scan may be stuck' wording. Pushed in 232b64a.

adapters/v1/syft.go
```diff
@@ -190,43 +195,73 @@ func (s *SyftAdapter) CreateSBOM(ctx context.Context, name, imageID, imageTag st
...
+	// cataloging pass, since the former means some other scan is stuck, not this one (#941).
+	// Caller cancellation implies nothing about any other scan, so it gets its own message
+	// instead of the same "stuck" wording, which would send an operator diagnosing #941 looking
+	// for a wedged semaphore that was never actually contended.
+	pullSemTimeoutResult := func(err error) (domain.SBOM, error) {
+		if errors.Is(err, context.DeadlineExceeded) {
+			logger.L().Ctx(ctx).Warning("gave up waiting to acquire the pull semaphore; a previous scan may be stuck",
 				helpers.String("imageID", imageID))
-		})
+			metrics.RecordScanFallback(ctx, metrics.ComponentInProcess, metrics.FallbackCategoryPullSemaphore, metrics.FallbackStrategyIncomplete, metrics.FallbackOutcomeClassified)
```
matthyx, line 261: Blocking: this helper still classifies every `DeadlineExceeded` as pull-semaphore contention. Both the pre-acquire call at line 239 and the post-acquire race path at line 250 can run while `pullSem` is free; `Test_syftAdapter_CreateSBOM_TimeoutContext` already exercises the first case with an expired caller context. Those paths now emit `category="pull_semaphore"` and warn that a previous scan may be stuck even though no previous scan holds the semaphore, contradicting the public metric documentation and making the acceptance-criterion-4 alert signal produce false positives. Please emit the dedicated metric/warning only after an acquisition actually had to wait and that wait reached its deadline; preserve the ordinary expired-context behavior for the pre/post-acquire paths, and add a metric regression test showing an already-expired context with a free semaphore does not increment `pull_semaphore`.

matthyx (changes requested):
The original metric blocker is substantially addressed, but the new signal still has a false-positive path that breaks its documented semaphore-specific meaning. One blocking inline comment remains.

adapters/v1/syft.go
```diff
@@ -190,43 +195,89 @@ func (s *SyftAdapter) CreateSBOM(ctx context.Context, name, imageID, imageTag st
...
+		// Re-check here so a dead context is degraded the same way regardless of which case
+		// fired, instead of silently proceeding into a real registry pull that would only fail
+		// moments later inside registryauth.ResolveSource. pullSem was free, so this is still
+		// contextAlreadyDeadResult, not genuine contention.
+		if err := ctxWithTimeout.Err(); err != nil {
+			<-s.pullSem
+			return contextAlreadyDeadResult(err)
+		}
+	case <-ctxWithTimeout.Done():
+		return pullSemStuckResult(ctxWithTimeout.Err())
```
matthyx, line 294: Blocking: the opposite side of the simultaneous-ready race remains. Lines 258-259 correctly note that when `pullSem` is free and the context expires at the same instant, both select cases are ready and Go chooses nondeterministically. If it chooses this `Done` case, `pullSemStuckResult` still records `pull_semaphore` and warns about another scan even though the semaphore is free. The new regression test covers a context that is already expired before the pre-check, but not this transition between the pre-check and select. Please distinguish the `Done` case too—for example, make a non-blocking acquire attempt there, release and use `contextAlreadyDeadResult` if it succeeds, and reserve `pullSemStuckResult` for the unavailable/default case—and add coverage for that branch.

matthyx (changes requested):
The deterministic false-positive path is fixed and the focused tests pass, but the simultaneous-ready select race described inline still leaves the semaphore-specific metric capable of false positives. The new commit also needs a Signed-off-by line: DCO currently fails. The cross-platform job's separate failure is an apparently unrelated 10-minute timeout in TestGracefulStopWithTimeout_ForcesStopWhenRPCOutlivesTimeout.

matthyx (approved):
Re-reviewed at f471bd8f1310a849fbf6809198e797b71ae48e17. The simultaneous-ready select race is now correctly distinguished with a non-blocking semaphore probe, the dedicated metric retains its intended meaning, DCO passes, and focused normal/race tests plus vet pass locally. No remaining code-review blockers.

adapters/v1/syft.go
bhuvan-somisetty: Fixed by the second commit already: `contextAlreadyDeadResult` now covers the pre-acquire check (line ~239) and the post-acquire race against an uncontended `pullSem` (the re-check right after the send succeeds), so both are classified as `size_classification`, not `pull_semaphore`. Added `Test_syftAdapter_CreateSBOM_ContextAlreadyDeadDoesNotRecordPullSemaphore`-style coverage for the already-expired-context-with-free-semaphore case.

adapters/v1/syft.go
bhuvan-somisetty: Fixed in f471bd8: added `pullSemWasContended`, a non-blocking acquire attempt against `pullSem`. In the `Done()` case it's now called first — if the non-blocking send succeeds, `pullSem` was actually free, so it's released immediately and we fall back to `contextAlreadyDeadResult` (same as every other dead-context path). `pullSemStuckResult` is now only reached when the non-blocking attempt genuinely fails. Added `Test_pullSemWasContended`, unit-testing both outcomes directly (free semaphore reported uncontended and left free; held semaphore reported contended and left untouched).

bhuvan-somisetty:
Hey @matthyx, fixed the DCO sign-off and the remaining race you flagged in the Done() branch — added a non-blocking acquire attempt so an unlucky select that picks the expired-context case doesn't get misclassified as a stuck semaphore when it was actually free. New unit test covers both outcomes directly. All checks are green now, ready for another look.

PR #945, opened 2026-09-10
fix(scan): rewrite CVE summary on cache hit so image upgrades re-point the manifest ref

### LINKED ISSUE #944: VulnerabilityManifestSummary references stale/non-existent VulnerabilityManifest after workload image upgrade
# Description

Following a container image upgrade on a workload, `VulnerabilityManifestSummary` continues to reference the previous, now non-existent `VulnerabilityManifest` instead of the newly generated `VulnerabilityManifest` corresponding to the updated image.

# Environment

- **Kubescape Operator Helm Chart**: `v1.40.1` (operator `v0.2.141`, kubevuln `v0.3.137`)
- **Kubernetes / OS**: Linux

# Steps To Reproduce

1. Deploy a workload (e.g. `external-secrets` deployment running `ghcr.io/external-secrets/external-secrets:v0.9.9`) in a cluster with Kubescape Operator v1.40.1 installed.
2. Allow Kubescape Operator / Kubevuln to scan the workload and create:
   - `VulnerabilityManifest`: `ghcr.io-external-secrets-external-secrets-v0.9.9-bfdf65`
   - `VulnerabilityManifestSummary`: referencing `v0.9.9-bfdf65`
3. Upgrade the workload image to a new version (e.g. `ghcr.io/external-secrets/external-secrets:v2.8.0`).
4. Wait for the new scan to complete. The cluster now contains `ghcr.io-external-secrets-external-secrets-v2.8.0-3ef679`, and the previous manifest (`v0.9.9-bfdf65`) is no longer present.
5. Check the `VulnerabilityManifestSummary` for the workload:
   ```yaml
   spec:
     vulnerabilitiesRef:
       all:
         kind: vulnerabilitymanifests
         name: ghcr.io-external-secrets-external-secrets-v0.9.9-bfdf65
         namespace: external-secrets
   ```

# Expected behavior

When a workload image is updated and scanned:
- The existing `VulnerabilityManifestSummary` should be updated so that `spec.vulnerabilitiesRef.all.name` points to the active `VulnerabilityManifest` (`ghcr.io-external-secrets-external-secrets-v2.8.0-3ef679`).
- Severity summary counts should reflect the findings of the new image.

# Actual Behavior

`VulnerabilityManifestSummary` retains the stale reference to the deleted `VulnerabilityManifest` (`ghcr.io-external-secrets-external-secrets-v0.9.9-bfdf65`). As a result, consumers of the CR (such as the Kubescape CLI, Headlamp plugin, or Prometheus exporter) encounter a broken reference to a non-existent object.

# Additional context

Reported by a user on Slack:
> Hi Team, I'm using the Kubescape Operator v1.40.1. I noticed that the VulnerabilityManifestSummary is referencing the following VulnerabilityManifest:
> ```yaml
> kind: vulnerabilitymanifests
> name: ghcr.io-external-secrets-external-secrets-v0.9.9-bfdf65
> namespace: external-secrets
> ```
> However, this VulnerabilityManifest does not exist in the cluster. The VulnerabilityManifest currently present in the cluster is:
> `ghcr.io-external-secrets-external-secrets-v2.8.0-3ef679`
> So there seems to be a mismatch between the VulnerabilityManifestSummary reference and the actual VulnerabilityManifest available in the cluster. Could you please check why the summary is still referencing the old v0.9.9 manifest instead of the existing v2.8.0 manifest?

### Technical Notes & Investigation Pointers

1. **Cache hits skipping summary updates (`kubevuln/core/services/scan.go`)**:
   - In `ScanCVE`, if a CVE manifest for the new image already exists in storage (`cve.Content != nil`), it invokes `reconcileCachedCVE`.
   - `reconcileCachedCVE` only calls `StoreCVESummary` if `!maps.Equal(prevIgnored, curIgnored)` (i.e. if `SecurityException` rules changed). If exceptions have not changed, `StoreCVESummary` is never called for the workload, leaving the workload's summary pointing to the old manifest.
   - Similarly, in `ScanCP`, if relevancy has not yet completed / produced filtered results (`sbomp.Content == nil`), `StoreCVESummary` is not written on cache hits.
2. **Workload image update handling in Operator (`operator/watcher/podwatcher.go`)**:
   - When a deployment/daemonset rolls out a new image, check whether the pod watcher triggers a scan that properly refreshes the `VulnerabilityManifestSummary` for the parent workload.
3. **Cross-namespace reference in `vulnerabilitiesRef`**:
   - In `VulnerabilityManifestSummary`, `vulnerabilitiesRef.all.namespace` is set to the workload namespace (`external-secrets`), while `VulnerabilityManifest` is stored in the operator namespace (`kubescape`) unless configured otherwise.

AdeshDeshmukh:
## Overview

When a workload upgrades its image (A → B) while B's CVE manifest is already cached — e.g. another workload scanned B first — and no security exceptions changed, neither `StoreCVE` nor `StoreCVESummary` runs on the cache-hit path. The workload's summary object is keyed `<kind>-<name>-<container>` (image-independent), so it keeps pointing at manifest A after A is garbage-collected — a dangling `vulnerabilitiesRef` for every downstream consumer (CLI, Headlamp plugin, Prometheus exporter).

Closes #944 

## Reproducer (before the fix)

`TestScanService_ScanCVE_CacheHitRewritesSummaryRef`: same Wlid+container, B's manifest pre-stored (warm cache), exceptions unchanged, scan with `ImageSlug=B`, assert `StoreCVESummary` is called with manifest B. Before this fix it fails — the summary write never happens:

```
Error: Should NOT be empty, but was []
```

## What this PR does

- Hoists `StoreCVESummary` out of the `IgnoredMatches` gate on both cache-hit branches — `ScanCVE` (`core/services/scan.go:505-520`) and the `ScanCP` image-level branch (`:298-313`) — so every scan rewrites the summary ref from the cached manifest. The write is idempotent Create-or-Update and the severity counts recompute from cached in-memory content.
- Keeps `StoreCVE` gated as before, since manifest content itself is unchanged — no redundant full-manifest rewrites.
- Deliberately untouched: the `cve.Content == nil` full-scan paths, the relevancy `StoreCVESummary(..., true)` path, the schema-unsupported stub paths, `SubmitCVE`, and anything outside kubevuln (the manifest GC and the podwatcher trigger live elsewhere; the `vulnerabilitiesRef.namespace` observation is out of scope for this fix).

## Confirm (after the fix)

- The identical reproducer now passes — summary rewritten with manifest B.
- Full verification: `go test ./core/... ./repositories/... -count=1 -race` → 142 tests pass (including the existing `"second scan"` case, which an extra idempotent summary write doesn't affect); `gofmt`, `go vet` (on touched packages), `go build ./...` clean.
- Note: 3 `adapters/v1` syft golden-file tests fail identically on the pristine tree (environment-dependent live-syft output vs golden files) — pre-existing, unrelated, verified by re-running them with this change stashed.

## Related issue(s)

- #944 

## Summary by CodeRabbit

* **Bug Fixes**
  * CVE summaries are now refreshed when a workload image is upgraded, ensuring they reference the current image manifest.
  * Cached scans consistently update stored CVE summary information when storage is enabled.

### DIFF
--- a/core/services/scan.go
+++ b/core/services/scan.go
@@ -726,6 +726,11 @@ func (s *ScanService) storeVEX(ctx context.Context, cve, cvep domain.CVEManifest
 // document — subject to the removal-safety rule below. publishVEX lets ScanCP opt
 // out: it republishes VEX later itself, once relevancy-filtered results are also
 // available, using the exceptionsComplete this call returns.
+//
+// The CVE summary is rewritten on every call under the same safety rule,
+// even when suppressions are unchanged: the summary is keyed by workload,
+// not by image, so a workload image upgrade must re-point it at the new
+// manifest rather than dangle at the old one (see #944).
 func (s *ScanService) reconcileCachedCVE(ctx context.Context, cve domain.CVEManifest, imageSlug string, publishVEX bool) (restoredCve, filteredCve domain.CVEManifest, exceptionsComplete bool) {
 	prevIgnored := v1.IgnoredMatchKeys(cve.Content)
 	cve.Content = v1.RestoreSuppressedMatches(cve.Content)
@@ -733,25 +738,21 @@ func (s *ScanService) reconcileCachedCVE(ctx context.Context, cve domain.CVEMani
 
 	if s.storage {
 		curIgnored := v1.IgnoredMatchKeys(filteredCve.Content)
-		if !maps.Equal(prevIgnored, curIgnored) {
-			// Persist additions freely, but never persist removals when the
-			// exception set is incomplete: a transient SecurityException CRD
-			// list failure must not look like a deletion and wipe suppression
-			// from the stored manifest.
-			hasRemovals := false
-			for k := range prevIgnored {
-				if _, ok := curIgnored[k]; !ok {
-					hasRemovals = true
-					break
-				}
+		// Persist additions freely, but never persist removals when the
+		// exception set is incomplete: a transient SecurityException CRD
+		// list failure must not look like a deletion and wipe suppression
+		// from the stored manifest.
+		hasRemovals := false
+		for k := range prevIgnored {
+			if _, ok := curIgnored[k]; !ok {
+				hasRemovals = true
+				break
 			}
-			if exceptionsComplete || !hasRemovals {
+		}
+		if exceptionsComplete || !hasRemovals {
+			if !maps.Equal(prevIgnored, curIgnored) {
 				if err := s.cveRepository.StoreCVE(ctx, filteredCve, false); err != nil {
 					logger.L().Ctx(ctx).Warning("storing CVE with exceptions", helpers.Error(err),
-						helpers.String("imageSlug", imageSlug))
-				}
-				if err := s.cveRepository.StoreCVESummary(ctx, filteredCve, domain.CVEManifest{}, false); err != nil {
-					logger.L().Ctx(ctx).Warning("storing CVE summary with exceptions", helpers.Error(err),
 						helpers.String("imageSlug", imageSlug))
 				}
 				// The stored manifest just changed, so the VEX document describing it is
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed the cache-hit reconciliation change and regression coverage. The summary is refreshed for the current image while unchanged CVE manifests remain untouched, and the incomplete-exception removal safety gate is preserved. No blockers found. Verified with: go test ./core/... ./repositories/... -count=1 -race; go vet ./core/services ./repositories.

matthyx:
sorry @AdeshDeshmukh I think #946 is more complete

PR #947, opened 2026-09-10
fix: thread context.Context through fetchCVEExceptions' backend call

kooomix:
## Summary
`kubescape/backend` v0.0.46+ changed `GetCVEExceptionByDesignator` to take a `ctx context.Context` as its first argument. `BackendAdapter.getCVEExceptionsFunc`'s field type, its one real call site in `fetchCVEExceptions` (which already has `ctx` in scope), and the test mocks are updated to match. Bumped `github.com/kubescape/backend` to v0.0.51 (pulls in `kubescape/storage` v0.0.305 transitively).

## Why
Found while consolidating ARMO's backend services into a Go workspace monorepo: one shared dependency graph means every module has to agree on one `kubescape/backend` version. kubevuln was the only consumer still calling the pre-context signature, which made it impossible to pick any single version that worked for everyone. This isn't a version-pin problem `replace` can solve — it's a real signature change, so the call site itself needs to thread the context through.

## Testing
`go build ./...` and `go test ./...` pass, except `TestCreateSBOM_PullHoldsTempDirGuard` (`pkg/sbomscanner/v1`), which fails identically on `main` on macOS — it's a `sun_path` (104-byte) Unix-socket-path-length limit hit by this machine's long `/var/folders/...` temp dir, unrelated to this change.

Docs-exempt: internal signature fix forced by upstream kubescape/backend API change, no behavioral or user-facing change.

AI-skills: armosec-shared-rules:agent-dispatch-policy

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved context handling when retrieving cloud CVE exceptions.

* **Chores**
  * Updated backend and storage components to newer versions.
  * Updated related test coverage without changing expected behavior.

### DIFF
--- a/adapters/v1/backend.go
+++ b/adapters/v1/backend.go
@@ -41,7 +41,7 @@ type BackendAdapter struct {
 	eventReceiverRestURL  string
 	apiServerRestURL      string
 	clusterConfig         pkgcautils.ClusterConfig
-	getCVEExceptionsFunc  func(string, string, *identifiers.PortalDesignator, map[string]string) ([]armotypes.VulnerabilityExceptionPolicy, error)
+	getCVEExceptionsFunc  func(context.Context, string, string, *identifiers.PortalDesignator, map[string]string) ([]armotypes.VulnerabilityExceptionPolicy, error)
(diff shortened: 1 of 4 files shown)


matthyx (approved):
LGTM! Clean context propagation and dependency bump.

PR #952, opened 2026-09-10
   fix(registryauth): retry structured auth denials with provider credentials

### LINKED ISSUE #943: GCP ADC fallback never fires: the "401 Unauthorized" gate misses Artifact Registry's structured DENIED error, so Workload Identity scanning silently fails
## Summary

`ResolveSource` only consults a registry-auth provider when the failed pull's error string contains `401 Unauthorized`. Google Artifact Registry and GCR reject an anonymous pull with a *structured* OCI error body, which `go-containerregistry` renders as `DENIED: Unauthenticated request. ...` — with the HTTP status text dropped. The gate therefore never matches, `GCP{}.Credentials` is never called, and every image in a private GCP registry fails to scan even though Application Default Credentials are present and authorized.

Net effect: on GKE with Workload Identity, image scanning against a private Artifact Registry / GCR silently does not work, and the log says the request was *unauthenticated* rather than that credentials were refused.

## Observed

Cluster: GKE 1.35, Workload Identity enabled (`GKE_METADATA`). The `kubevuln` KSA is bound to `roles/artifactregistry.reader` on the repositories, as a direct workload-identity principal (`principal://.../subject/ns/kubescape/sa/kubevuln`) — no GSA, no key, no `imagePullSecrets` on the scanned workloads.

```json
{"level":"error","ts":"...","msg":"creating SBOM, skipping scan",
 "error":"an error occurred attempting to resolve '<REGION>-docker.pkg.dev/<project>/<repo>/<image>@sha256:...': oci-registry: failed to get image descriptor from registry: GET https://<REGION>-docker.pkg.dev/v2/token?scope=repository%3A<project>%2F<repo>%2F<image>%3Apull&service=: DENIED: Unauthenticated request. Unauthenticated requests do not have permission \"artifactregistry.repositories.downloadArtifacts\" on resource \"projects/<project>/locations/<region>/repositories/<repo>\" (or it may not exist)"}
```

Identical output for `eu.gcr.io/...`. Note `DENIED: Unauthenticated request` — not `Permission ... denied`, i.e. no credentials were ever presented.

## Why

`internal/registryauth/resolve.go` (v0.3.430):

```go
src, err := get(ctx, pullRef, &opts)          // anonymous: the workloads carry no pull secret
...
if err != nil && strings.Contains(err.Error(), "401 Unauthorized") {   // ← never true here
    if provider, ok := For(pullRef); ok {                              // GCP{} would match
        creds, credErr := provider.Credentials(ctx, pullRef)           // never reached
```

`GCP{}.Matches` does match these references (`IsGCPRegistry` covers `*.gcr.io` and `*-docker.pkg.dev`), so the provider is correct — it is simply never consulted.

The reason the substring is absent is in `go-containerregistry`'s `transport.Error.Error()`:

```go
func (e *Error) responseErr() string {
	switch len(e.Errors) {
	case 0:
		...
		return fmt.Sprintf("unexpected status code %d %s", e.StatusCode, http.StatusText(e.StatusCode))
	case 1:
		return e.Errors[0].String()      // ← "DENIED: <message>", status text dropped
```

`unexpected status code 401 Unauthorized` is only produced when the response body is empty or unparseable. Artifact Registry and GCR always return a well-formed `{"errors":[{"code":"DENIED","message":"Unauthenticated request. ..."}]}`, so they take the `case 1` branch and the rendered string contains neither `401 Unauthorized` nor `403 Forbidden`.

The same file already knows this distinction is not expressible as one status string — `isAuthDenied` checks both 401 and 403 — but the gate that decides whether to try a provider does not use it.

## Impact

- Any private GCP-hosted image that kubevuln has to pull is never scanned. In our case the visible symptom was that workloads with no running pod (CronJobs between runs) had no SBOM at all, while running workloads were covered because `nodeSbomGeneration` builds their SBOM from the node instead — so the failure is easy to miss until you notice CronJobs are absent from the results.
- The failure is silent from an operator's point of view: the error names a permission, which reads like an IAM problem and sends you off granting roles that are already granted.

One observation that may help bisect, offered as data rather than a claim about cause: on v0.3.159 the same cluster and the same repository produced the *authenticated* denial (`DENIED: Permission 'artifactregistry.repositories.downloadArtifacts' denied on resource ...`, with a troubleshooter URL — GCP only emits that form for an identified principal), meaning ADC credentials did reach the registry on that version. The `401 Unauthorized` gate is textually the same in v0.3.159's `adapters/v1/syft.go`, so what changed appears to be how the underlying error is rendered, not this code.

## Suggested fix

Decide on the transport error rather than its rendered text — `transport.Error` carries `StatusCode`, and `go-containerregistry` is already a direct dependency:

```go
var terr *transport.Error
if errors.As(err, &terr) && (terr.StatusCode == http.StatusUnauthorized || terr.StatusCode == http.StatusForbidden) {
```

If string matching has to stay, at minimum use the existing `isAuthDenied(err)` at the provider gate as well as at the anonymous-retry gate, and extend it to cover the OCI error codes (`DENIED`, `UNAUTHORIZED`) that a registry returns in a structured body. Note that `isAuthDenied` has the same blind spot today — it is just not on the path that breaks here.

A regression test that feeds `ResolveSource` an error rendered the way a real registry renders it (`GET https://...: DENIED: Unauthenticated request. ...`) and asserts the provider is consulted would keep this from coming back.

## Environment

- kubevuln v0.3.430 (kubescape-operator chart 1.40.4), also present in v0.3.404
- GKE 1.35, Workload Identity (`GKE_METADATA`), no `imagePullSecrets` on scanned workloads
- Registries: Artifact Registry (`*-docker.pkg.dev`) and GCR (`*.gcr.io`) in a project other than the cluster's

ANAMASGARD:
## Overview

  Private Google Artifact Registry and GCR pulls can return structured OCI errors such as `DENIED: Unauthenticated request` without the HTTP status text.
  `ResolveSource` previously entered the registry-auth fallback only when the rendered error contained `401 Unauthorized`, so GCP Application Default Credentials were never requested.

  This change:

  - Detects typed registry transport errors using their HTTP status code.
  - Recognizes flattened `DENIED` and `UNAUTHORIZED` OCI diagnostics when the underlying error type has been lost.
  - Retries supported private registries with provider credentials, enabling GCP ADC and Workload Identity authentication.
  - Preserves the existing behavior that 429, 5xx, timeout, and other non-authentication failures do not trigger an anonymous retry.
  - Adds regression coverage for Artifact Registry, regional GCR, typed and flattened errors, credential rejection, and the `MANIFEST_UNKNOWN` tag fallback sequence.

  No dependency or public API changes are included.

  ## How to Test

  ```bash
  export TMPDIR=/tmp
  export GOCACHE=/tmp/kubevuln-943-go-cache

  go test -mod=readonly -p=1 ./internal/registryauth -count=1

  go test -mod=readonly -p=1 ./pkg/sbomscanner/v1 \
    -run '^TestResolveSource' -count=1

  go test -mod=readonly -p=1 ./adapters/v1 -count=1

  go vet -mod=readonly -p=1 \
    ./internal/registryauth/... ./adapters/v1/... ./pkg/sbomscanner/...

  go build -mod=readonly -p=1 ./...
```

  All commands pass locally. The regression tests deterministically reproduce the reported fallback path; live GKE Workload Identity integration was not exercised locally.

  ## Related issues/PRs

  - Resolved #943

  - [x] I have commented on hard-to-understand behavior
  - [x] I have added focused regression tests
  - [x] New and existing relevant unit tests pass locally

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of container registry authentication failures, including structured and wrapped error responses.
  * Credentialed fallback now works reliably for recognized unauthorized and denied responses.
  * Non-authentication errors continue to be surfaced without triggering credential fallback.
  * Digest references that return `MANIFEST_UNKNOWN` now retry using the associated tag.

### DIFF
--- a/internal/registryauth/resolve.go
+++ b/internal/registryauth/resolve.go
@@ -2,10 +2,13 @@ package registryauth
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"net/http"
 	"strings"
 
 	"github.com/anchore/stereoscope/pkg/image"
+	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
 	"github.com/kubescape/go-logger"
 	"github.com/kubescape/go-logger/helpers"
 	"github.com/kubescape/kubevuln/internal/metrics"
@@ -18,19 +21,33 @@ type Getter[T any] func(ctx context.Context, ref string, opts *image.RegistryOpt
 
 // isAuthDenied reports whether err is a registry rejecting the request as unauthenticated or
 // unauthorized -- the two outcomes credentials (ours, or none at all) can actually change.
-// Matches the same bucket core/services/scan_failure_reasons.go's classifySBOMError already
-// uses for the identical distinction (401 vs. 403 both mean "the registry looked at who's
-// asking and said no", where a 429/5xx/timeout means something else went wrong that swapping
-// credentials cannot fix).
 func isAuthDenied(err error) bool {
-	return err != nil && (strings.Contains(err.Error(), "401 Unauthorized") || strings.Contains(err.Error(), "403 Forbidden"))
+	if err == nil {
+		return false
+	}
+	var terr *transport.Error
+	if errors.As(err, &terr) {
+		return terr.StatusCode == http.StatusUnauthorized || terr.StatusCode == http.StatusForbidden
+	}
+	// The pinned stereoscope formats registry errors with %+v, losing their type.
+	// Structured OCI diagnostics omit HTTP status text, so recognize their codes too.
+	message := err.Error()
+	if strings.Contains(message, "401 Unauthorized") || strings.Contains(message, "403 Forbidden") {
+		return true
+	}
+	for _, code := range []string{"DENIED:", "UNAUTHORIZED:"} {
+		if strings.HasPrefix(message, code) || strings.Contains(message, ": "+code) || strings.Contains(message, "; "+code) {
+			return true
+		}
+	}
+	return false
 }
 
 // ResolveSource pulls imageID, falling back in order when the pull fails:
 //
 //   - MANIFEST_UNKNOWN, which a registry returns for a digest it does not hold, is retried
 //     against imageTag.
-//   - 401 Unauthorized is retried with credentials from the provider matching the reference
+//   - Authentication rejection is retried with credentials from the provider matching the reference
 //     (ECR, GCP), and then, if that provider has none or its credentials are refused too,
 //     without credentials at all. A registry that rejects our credentials may still serve
 //     the image anonymously.
@@ -57,7 +74,7 @@ func ResolveSource[T any](ctx context.Context, component string, get Getter[T], 
 		src, err = get(ctx, pullRef, &opts)
 	}
 
-	if err != nil && strings.Contains(err.Error(), "401 Unauthorized") {
+	if isAuthDenied(err) {
 		usedFallback = true
 		unauthorizedErr := err
 		if provider, ok := For(pullRef); ok {
@@ -89,7 +106,7 @@ func ResolveSource[T any](ctx context.Context, component string, get Getter[T], 
 		// eventual error permanently mislabels whatever actually went wrong as "unauthorized"
 		// (via unauthorizedErr below). When no provider matched or its credentials were
 		// unavailable, err is untouched here and still holds the original auth-denied error
-		// from line 50 (or the MANIFEST_UNKNOWN retry above it), so this check doesn't change
+		// from the initial pull (or the MANIFEST_UNKNOWN retry above it), so this check doesn't change
 		// that path's existing behavior. See #921.
 		if isAuthDenied(err) {
 			logger.L().Debug("retrying without credentials",
--- a/internal/registryauth/resolve_test.go
+++ b/internal/registryauth/resolve_test.go
@@ -3,10 +3,14 @@ package registryauth
 import (
 	"context"
 	"errors"
+	"fmt"
+	"net/http"
+	"os"
 	"testing"
 	"time"
 
 	"github.com/anchore/stereoscope/pkg/image"
+	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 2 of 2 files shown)


matthyx (commented):
The fix correctly recognizes typed 401/403 transport errors and the flattened `DENIED` / `UNAUTHORIZED` diagnostics seen through stereoscope. The focused tests, race test, affected-package vet/tests, full build, and current CI all pass; I found no correctness or security blocker in this diff.

One non-blocking follow-up: `core/services/scan_failure_reasons.go` recognizes flattened `UNAUTHORIZED` but not `DENIED`. If both credentialed and anonymous attempts end in `DENIED`, resolution now retries correctly, but the final failure can still be reported as a generic SBOM-generation failure instead of image-auth failure. Please consider centralizing this classification or adding `DENIED` there with a terminal-denial test in a follow-up.

matthyx (approved):
Approving at `39a6a9e`.

Re-reviewed the full fallback ladder after the earlier COMMENT review. The change is correct and the widening is well-bounded:

- `errors.As(err, &terr)` returns on the typed status code and deliberately ignores the OCI diagnostic code, so a typed 404/429/500 carrying a `DENIED` diagnostic is *not* treated as auth-denied. That's the right precedence and it's covered by tests.
- The flattened-code matching is anchored (prefix, `: `, `; `), so it won't fire on prose that merely contains the word — `TestIsAuthDenied` pins both the positive and the negative side of that.
- The non-auth contract from #921 is intact: 429/5xx/timeout after a credentialed retry are returned unchanged, verified by the `credentialed 429/500/timeout` cases.
- `MANIFEST_UNKNOWN` still runs before the auth branch, and `tag before authentication` locks that ordering in.

Two non-blocking follow-ups, neither a reason to hold this PR:

1. `internal/registryauth/resolve.go:121` — the anonymous-fallback error wrap still tests the literal `"401 Unauthorized"` rather than `isAuthDenied(err)`. That line's intent is "don't wrap when anonymous was denied the same way", and it now misses the `DENIED` / typed-403 cases this PR just made reachable, so those get the noisier `... (anonymous fallback failed: ...)` form. Cosmetic only — `%w` keeps `unauthorizedErr` intact, so classification is unaffected.

2. `core/services/scan_failure_reasons.go` — `classifySBOMError`'s string fallback matches `UNAUTHORIZED` but not `DENIED`. A GAR pull that stays denied through every fallback and arrives flattened (type lost, e.g. across the sidecar's gRPC boundary) is reported as `ReasonSBOMGenerationFailed` instead of `ReasonImageAuthFailed`. Pre-existing, not introduced here, but this PR makes that path a lot more visible for GCP users — worth a separate PR adding `DENIED` plus a terminal-denial test.

CI is green on all 11 checks at this commit.

PR #956, opened 2026-09-11
fix(adapters): compare apk/deb/rpm fix versions with grype's own format, not generic semver

### LINKED ISSUE #955: suggestedVersion() falls back to unguarded, semver-only comparison on OS-package versions, reopening the #845 downgrade bug and hiding real apk fixes
## Problem

`suggestedVersion` (`adapters/v1/domain_to_armo.go:294`) picks "the nearest fix version above the installed one" by parsing both the installed version and every candidate fix version with `github.com/Masterminds/semver/v3` and comparing them as semver. Real OS-package versions from apk/dpkg/rpm matches — the majority of matches grype produces for any Debian/Ubuntu/Alpine/RHEL-based image — are not semver, and this breaks in two independent ways:

1. **Epoch-prefixed versions fail to parse at all**, and the failure path returns `versions[0]` unconditionally, with no check that it's actually newer than what's installed. That's exactly the "can suggest a downgrade" bug fixed in #844/#845 for the parseable path — it's just unguarded again whenever `current` happens to carry an epoch, which dpkg/rpm add automatically to any package that has had an epoch bump (grype itself documents this format: `grype/pkg/package.go:23` — *"if there's an epoch, we'd expect a value like `4:2.17.2`"*).
2. **Multi-digit apk revision suffixes sort wrong.** Masterminds/semver treats `-r9`/`-r10` as prerelease identifiers and, since they aren't purely numeric, compares them lexically (`"r10" < "r9"`) instead of numerically. A genuinely newer fix revision can therefore compare as "not greater," so the loop never picks it up.

## Reproduction / Evidence

Using the vendored `github.com/Masterminds/semver/v3 v3.4.0` (the exact version pinned in `go.mod`) directly against `suggestedVersion`'s logic:

```go
// Case 1: epoch-prefixed installed version (any package with an epoch bump —
// e.g. util-linux, iptables, curl on many Debian/Ubuntu releases).
// "1:1.2.11.dfsg-2ubuntu1.2" fails semver.NewVersion (colon is not valid semver),
// so the unguarded fallback returns versions[0] with no upgrade check.
suggestedVersion("1:1.2.11.dfsg-2ubuntu1.2",
    []string{"1:1.2.11.dfsg-2ubuntu1.1", "1:1.2.11.dfsg-2ubuntu1.3"})
// => "1:1.2.11.dfsg-2ubuntu1.1"   <- OLDER than what's installed: a downgrade.

// Case 2: apk two-digit revision. A real newer fix (-r10) exists but is missed.
suggestedVersion("3.4.7-r9", []string{"3.4.7-r10"})
// => ""   (hasKnownFix then reports the fix as "unknown" instead of "3.4.7-r10")

v9, _ := semver.NewVersion("3.4.7-r9")
v10, _ := semver.NewVersion("3.4.7-r10")
v9.LessThan(v10) // => false — Masterminds/semver does not consider r10 newer than r9
```

I confirmed independently that `semver.NewVersion` rejects epoch-prefixed strings (`"1:1.2.11.dfsg-2ubuntu1.2"`, `"0:4.18.0-553.el9"`) with `invalid semantic version`, while non-epoch distro strings like `"2.31-r0"` parse "successfully" but are silently reinterpreted as semver prerelease identifiers, which is where the revision-ordering bug comes from.

`adapters/v1/domain_to_armo_test.go`'s `Test_suggestedVersion` table has no case with an epoch prefix or a two-digit `-rN` revision, so this gap isn't caught by the existing suite.

## Root Cause

`suggestedVersion` uses a generic, strict semver parser/comparator for version strings that are not semver — they're apk/dpkg/rpm package versions, each with its own (non-semver) ordering rules. grype itself already solves this correctly and is already a direct dependency: `models.NewVulnerability`'s `sortVersions` (`github.com/anchore/grype/grype/presenter/models/vulnerability.go`) compares fix versions with `github.com/anchore/grype/grype/version`, constructing each `version.Version` with the artifact's own `version.Format` (Deb/RPM/Apk/etc.) instead of assuming semver. kubevuln's adapter has that same format available on the match (`m.Artifact.Type` / `v1beta1.SyftType`) but doesn't use it here, so it reimplements version comparison with a parser that doesn't understand the ecosystems it's actually comparing.

## Impact

- **Wrong remediation guidance surfaced to users**: `DomainToArmo` (backend vulnerability report) can present an older, still-vulnerable package version as "the fix," for any package whose installed version carries an epoch — common on Debian/Ubuntu/RHEL-based images, which is a large share of real-world container images.
- **Real fixes silently hidden**: on Alpine-based images, once a package's apk revision reaches double digits, a genuinely available newer fix can be reported as "unknown" instead of its concrete version, degrading the same remediation data `hasKnownFix` (`adapters/v1/fixstate.go`) feeds into both `DomainToArmo` and `ApplySecurityExceptions`.
- This is a direct regression of the exact class of bug #844/#845 set out to prevent ("no version above current returns empty rather than a downgrade") — the guard only exists on the semver-parseable branch, not on the (very common) branch where the installed version isn't semver at all.

## Affected Components

- `adapters/v1/domain_to_armo.go` — `suggestedVersion` (definition), used by `DomainToArmo` for backend vulnerability reports.
- `adapters/v1/fixstate.go` — `hasKnownFix`, which calls `suggestedVersion` and feeds both the backend report path and `ApplySecurityExceptions`'s in-cluster `VulnerabilityManifest`/`ExpiredOnFix` logic (per that file's own comment on why the two paths must agree).

## Acceptance Criteria

- `suggestedVersion` (or its replacement) never returns a version that is not actually greater than the installed one, for apk/dpkg/rpm-style versions including epoch prefixes — no silent fallback to `versions[0]` when the installed version fails to parse.
- A genuinely newer fix version is not dropped/mislabeled as `"unknown"` due to lexical-vs-numeric revision comparison (e.g. `-r9` vs `-r10`, `-r99` vs `-r100`).
- New regression tests cover: an epoch-prefixed installed version with a real newer fix available; an epoch-prefixed installed version with no newer fix (must return `""`, not a downgrade); apk multi-digit revision ordering.

Reuse grype's own format-aware version comparison (`github.com/anchore/grype/grype/version`, keyed by the artifact's ecosystem via `version.Format`/`m.Artifact.Type`) instead of a second, generic-semver-only implementation — mirroring what grype's own `models.NewVulnerability` → `sortVersions` already does for the same fix-version data. That gives correct comparisons across all ecosystems grype already supports (apk, dpkg, rpm, and the semver-friendly ones) with one code path, rather than patching Masterminds/semver's edge cases one distro quirk at a time.

bhuvan-somisetty:
## Overview

`suggestedVersion()` (`adapters/v1/domain_to_armo.go`) picks the nearest fix version above the installed one by parsing both with `Masterminds/semver`. Real apk/dpkg/rpm package versions aren't semver, and that broke in two ways:

1. **Epoch-prefixed versions fail to parse entirely.** dpkg/rpm add an epoch prefix (e.g. `1:2.3.4-1`) to any package that has had one, and that's not valid semver. The parse failure fell through to an unguarded `return versions[0]`, which is exactly the downgrade bug #844/#845 already fixed - just reopened on this (very common) unparseable-input path.
2. **Alpine `-rN` release revisions sort wrong.** Semver treats `-r9`/`-r10` as prerelease identifiers and compares them lexically rather than numerically, so a genuinely newer fix at a two-digit revision compared as "not greater" and was silently missed (reported as `unknown` instead of the real version).

## Root Cause

Generic semver isn't the right comparator for OS package ecosystems. Grype itself already solves this: `models.NewVulnerability` (`grype/presenter/models/vulnerability.go`) sorts its own fix versions using `grype/version`, a comparator keyed by the package's ecosystem format (apk/deb/rpm/etc.), not plain semver. `suggestedVersion` had the artifact's type available at both call sites but wasn't using it.

## Solution

`suggestedVersion` now takes the match's artifact type. For `apk`, `deb`, and `rpm` types, it compares the installed version and every candidate with Grype's own format-aware comparator (`github.com/anchore/grype/grype/version`) instead of semver - correctly handling epoch prefixes and numeric revision suffixes. Every other ecosystem is untouched and keeps the existing semver path, so this is scoped to exactly the ecosystems where the bug reproduces.

If the installed version doesn't parse even under its own ecosystem's format, the function falls through to the original "nothing to compare against" behavior (return `versions[0]`) - unchanged from before for genuinely unparseable input.

## Testing

- Added regression cases to `Test_suggestedVersion` (`domain_to_armo_test.go`) covering: a deb/rpm epoch version with no real upgrade (must return `""`, not a downgrade), a deb epoch version with a real newer fix across the epoch, and apk two-digit revision ordering (`-r9` vs `-r10`, and picking the nearest of several).
- Added matching cases to `TestHasKnownFix` (`fixstate_test.go`) exercising the real production entry point (`hasKnownFix`), not just the helper in isolation.
- All existing `Test_suggestedVersion` / `TestHasKnownFix` / `TestExpiredOnFix*` cases pass unchanged.
- `go build ./...`, `go vet ./adapters/v1/...`, and `gofmt` all clean.
- Full `adapters`/`adapters/v1` package test suite passes (`go test -tags dockerfixture ./adapters/...`).

## Impact

Any Debian/Ubuntu/RHEL-based image with a package that has an epoch in its installed version, or any Alpine-based image where a fix lands at a double-digit `apk` revision, gets a correct suggested fix version instead of a possible downgrade suggestion or a silently dropped one. This feeds both the backend vulnerability report (`DomainToArmo`) and the in-cluster `ApplySecurityExceptions`/`ExpiredOnFix` logic, since both share `hasKnownFix`.

Resolves #955

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved vulnerability fix-version matching for Debian, RPM, and Alpine packages.
  * Correctly handles epoch-prefixed versions and multi-digit Alpine release revisions.
  * Provides more accurate suggested fixes and avoids recommending older or unknown versions when a valid newer fix exists.
  * Prevents unreliable RPM recommendations when versions include mixed epochs, caret markers, or trailing-zero differences that cannot be safely compared.


adapters/v1/domain_to_armo.go
```diff
@@ -287,14 +288,31 @@ func linkToVuln(id string) string {
...
+// since there is nothing to compare against. If current is a version but no entry in
+// versions is greater than it, "" is returned rather than falling back to versions[0];
+// versions[0] could be older than current, which would suggest a downgrade (#844).
+func suggestedVersion(current string, versions []string, artifactType v1beta1.SyftType) string {
 	if len(versions) == 0 {
 		return ""
 	}
+
+	if format, ok := distroPackageVersionFormat(artifactType); ok {
+		if v, resolved := nearestDistroFix(current, versions, format); resolved {
```
matthyx: **Blocker:** when the distro parser cannot resolve `current`, this branch falls through to semver and ultimately returns `versions[0]` on another parse error. That is the exact unchecked fallback #955's acceptance criteria prohibit: a comparator failure becomes a concrete remediation even though we cannot prove it is newer. The new test at `domain_to_armo_test.go:486` currently locks in this unsafe behavior. For recognized distro types, please return no concrete suggestion on parse failure (so `hasKnownFix` reports `unknown`) and invert that regression test accordingly.

adapters/v1/domain_to_armo.go
```diff
@@ -318,6 +336,51 @@ func suggestedVersion(current string, versions []string) string {
...
+	c := grypeversion.New(current, format)
+	if err := c.Validate(); err != nil {
+		return "", false
+	}
+
+	var nearest *grypeversion.Version
+	var nearestStr string
+	for _, raw := range versions {
+		v := grypeversion.New(raw, format)
+		cmp, err := c.Compare(v)
```
matthyx, line 374: **Blocker:** Grype v0.104.1's RPM comparator deliberately uses relaxed vulnerability-matching semantics, not strict RPM upgrade ordering. It ignores epochs when either side omits one, so current `1:0` with candidate `1` compares as an upgrade even though RPM defines the missing epoch as `0` (equal, not newer). Its tokenization also mishandles RPM caret ordering; candidate `1.0^20250611` can be accepted above current `1.0.1`, while RPM orders it below. Since this function emits concrete remediation, please enforce RPM-compliant ordering (or conservatively emit no suggestion when strict comparison is unavailable) in both this comparison and the nearest-candidate comparison, with mixed-epoch and caret regression tests.

matthyx (changes requested):
Two blockers remain against #955's no-downgrade contract. The focused adapter tests and vet pass, but the new tests do not cover these failure modes.

bhuvan-somisetty:
Both blockers addressed:

1. A recognized distro artifact's current version failing to parse under its own comparator no longer falls through to the semver path — `nearestDistroFix`'s own `""` result is now trusted as final, so `versions[0]` can never be reached through that route for apk/deb/rpm.
2. Added `rpmSafeToCompare`, which rejects a candidate pair (in both the current-vs-candidate and candidate-vs-candidate comparisons) when they have mismatched epoch presence or either side carries a `^`, rather than trusting Grype's relaxed RPM comparator on those specific shapes.

Regression tests added for the reverted fallback, the mixed-epoch case (`1:0` vs `1`), the caret case (`1.0.1` vs `1.0^20250611`), confirmation that ordinary RPM comparisons are unaffected, and a focused table test for `rpmSafeToCompare` itself.

`go build ./...` and `go vet ./adapters/v1/...` pass. I wasn't able to get the adapters test binary to finish running locally this round (repeatedly OOM-killed by the OS on my end, unrelated to this diff), so I'd appreciate CI confirming the new tests pass before merge.

adapters/v1/domain_to_armo.go
```diff
@@ -318,6 +337,92 @@ func suggestedVersion(current string, versions []string) string {
...
+// Reimplementing spec-compliant RPM version comparison here would trade one hazard for
+// another -- a hand-rolled comparator error would be just as capable of shipping a wrong
+// remediation. Since suggestedVersion's contract is "never suggest an unproven upgrade,"
+// the conservative answer for either shape is to treat the pair as impossible to order
+// safely, so the caller skips that candidate rather than trusting Grype's relaxed result.
+func rpmSafeToCompare(a, b string) bool {
+	if strings.Contains(a, "^") || strings.Contains(b, "^") {
+		return false
+	}
+	return rpmHasExplicitEpoch(a) == rpmHasExplicitEpoch(b)
```
matthyx: **Blocker:** this guard covers mixed epoch presence and carets, but Grype has another relaxed RPM rule: it treats trailing zero version segments as equal, then compares releases. For current `1:1.0-1` and candidate `1:1-2`, this guard allows the comparison; Grype returns candidate-newer, while native librpm returns current-newer (`rpmverCmp(...) = 1`). Thus `suggestedVersion` still emits an RPM downgrade. The unprefixed `1.0-1a` → `1-2` case is also newly accepted by this PR whereas the old semver path rejected it. Please use strict/spec-compliant RPM ordering for concrete remediation, or conservatively guard this additional ambiguity, and add regressions expecting no suggestion (plus `["1-2", "1.0-3"]` selecting `1.0-3`).

matthyx (changes requested):
The previous two blockers are fixed, but one remaining Grype RPM relaxation can still produce a downgrade, so #955's core safety contract is not yet met.

bhuvan-somisetty:
Fixed. Grype's \`compareRpmVersions\` treats two versions as equal whenever one tokenizes into strictly more alphanumeric segments than the other and every extra segment is literally \`\"0\"\` (confirmed directly: \`grypeversion.New(\"1:1.0-1\", RpmFormat).Compare(grypeversion.New(\"1:1-2\", RpmFormat))\` returns \`-1\`, i.e. Grype judges \`1:1-2\` newer), then falls through to comparing releases alone — which is backwards from real RPM, where the side with the extra segment outranks the other regardless of release.

\`rpmSafeToCompare\` now also strips epoch and release (mirroring Grype's own \`newRpmVersion\` split) and rejects a candidate pair whose version components tokenize to a common prefix plus one side having only-\`"0"\` extra segments, alongside the existing epoch-presence and caret guards.

Added the exact regressions requested: the epoch-prefixed case (\`1:1.0-1\` vs \`1:1-2\`), the unprefixed case (\`1.0-1\` vs \`1-2\`), and \`["1-2", "1.0-3"]\` correctly selecting \`1.0-3\` (the guard rejects only the mismatched candidate, not the whole comparison) — plus a focused table test for the new helper.

\`go build ./adapters/...\`, \`go vet ./adapters/v1/...\`, and \`go test -tags dockerfixture ./adapters/v1/...\` (full package, not just the new cases) all pass locally this round.

adapters/v1/domain_to_armo.go
```diff
@@ -318,6 +338,150 @@ func suggestedVersion(current string, versions []string) string {
...
+	if len(segsA) == len(segsB) {
+		return false
+	}
+	shorter, longer := segsA, segsB
+	if len(shorter) > len(longer) {
+		shorter, longer = longer, shorter
+	}
+	for i := range shorter {
+		if shorter[i] != longer[i] {
+			return false
```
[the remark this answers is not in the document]
bhuvan-somisetty: Fixed in 9feea70: rpmVersionsDifferOnlyByTrailingZeros now compares both the shared-prefix segments and the trailing all-zero segments numerically (leading zeros trimmed) instead of as raw strings, so "01" and "1" are recognized as the same digit run. Added regression tests at the rpmVersionsDifferOnlyByTrailingZeros, rpmSafeToCompare, and suggestedVersion levels using this exact 1.01.0-1 / 1.1-2 pair.

adapters/v1/domain_to_armo.go
bhuvan-somisetty: Already fixed in c3ed039 ("never fall back to an unproven distro fix suggestion"): nearestDistroFix now returns "" when current fails c.Validate() under its ecosystem's comparator, instead of falling through to versions[0]. Regression test: "apk artifact with an unparseable current returns empty, not the first entry" (domain_to_armo_test.go).

adapters/v1/domain_to_armo.go
bhuvan-somisetty: Already fixed in c3ed039: rpmSafeToCompare rejects mixed epoch-presence pairs (rpmHasExplicitEpoch(a) != rpmHasExplicitEpoch(b)) and any pair containing a caret ("^"), so nearestDistroFix skips these candidates instead of trusting Grype's relaxed ordering - the conservative "no suggestion" alternative this comment proposed. Regression tests: "rpm mixed epoch presence is not trusted as an upgrade" and "rpm caret release is not trusted as an upgrade" (domain_to_armo_test.go).

adapters/v1/domain_to_armo.go
bhuvan-somisetty: Already fixed in 8c30431 ("guard against Grype's RPM trailing-zero-version relaxation"): rpmSafeToCompare now also rejects pairs whose version parts differ only by trailing zero-valued segments (rpmVersionsDifferOnlyByTrailingZeros), covering the 1:1.0-1 / 1:1-2 shape this comment describes. Regression tests: "rpm trailing-zero version segment is not trusted as an upgrade" (with and without an epoch) and "rpm trailing-zero guard skips only the mismatched candidate" (which covers the requested ["1-2","1.0-3"] -> "1.0-3" case). A follow-up gap in that same guard (a leading zero inside the shared prefix, e.g. "1.01.0-1" vs "1.1-2") was just fixed in 9feea70, per the separate thread below.

matthyx (commented):
Rechecked at `9feea70`.

The previous trailing-zero/leading-zero RPM downgrade blocker is fixed, and I found no remaining no-downgrade violation. The focused adapter tests, vet, formatting/diff checks, and all current CI checks pass.

One low, nonblocking contract concern remains: `nearestDistroFix` skips comparisons between mutually unsafe-to-compare RPM candidates at `adapters/v1/domain_to_armo.go:382-384`. Both candidates may be proven upgrades from the installed version, but reversing their feed order can change the returned result (for example, current `0.5-1` with candidates `1.0-1` and `1-2`). This is safe, but it does not always fulfill the helper's documented “smallest newer version” promise.

Please either document that RPM selection guarantees a proven upgrade but only prefers the minimum where ordering is safe, or add a deterministic policy/test for mutually incomparable candidates. I consider this follow-up nonblocking for #955's safety acceptance criteria.

matthyx (changes requested):
The original no-downgrade fixes remain sound, but the new feed-order-independence fix is incomplete for three or more candidates.

**MEDIUM:** `adapters/v1/domain_to_armo.go:387-398` mixes lexical selection for unsafe-to-compare pairs with RPM ordering for safe pairs. That hybrid relation is non-transitive.

Concrete reproduction with current `0-1`:

- A = `1-1`
- B = `1+0-1`
- C = `1a-1`

B beats A via the lexical fallback, A beats C via RPM ordering, and C beats B via RPM ordering. An exact regression against this head produced:

- `[A, B, C]` -> `1a-1`
- `[A, C, B]` -> `1+0-1`

The new two-candidate reversal test cannot expose this cycle, so the documented feed-order-independent guarantee at lines 361-365 is still false. This does not demonstrate a downgrade—every admitted candidate remains a proven upgrade from current—but it leaves remediation output unstable.

Please make the policy deterministic across the complete candidate set and add all permutations of this triple. A minimal option is to sort a copy of candidates lexically before reduction and explicitly document the weaker ambiguous-set semantics; do not sort with the non-transitive hybrid comparator itself.

bhuvan-somisetty:
Fixed the transitivity issue - candidate selection now falls back to a lexical minimum once more than one RPM candidate is eligible, so it can't cycle regardless of feed order. Added a permutation test with your exact repro. Pushed.

matthyx (changes requested):
The three-candidate permutation issue is fixed, but the set-wide lexical fallback introduces a broader nearest-version regression.

**MEDIUM:** `adapters/v1/domain_to_armo.go:397-400` returns the lexical minimum for every RPM set with more than one eligible upgrade—even when all candidates are safely comparable by RPM rules.

An exact regression against `dc8e668` with current `1-1` and candidates `["2-1", "10-1"]` returns `10-1`; the nearest RPM upgrade is `2-1`. The same defect applies to release numbers such as `1-2` versus `1-10`. This contradicts `suggestedVersion`'s established nearest/smallest contract at lines 287-297 and can recommend an unnecessarily large upgrade.

Please preserve numeric RPM selection for safely ordered candidates and confine deterministic fallback behavior to ambiguous sets. A bounded approach is to lexically sort the locally owned eligible slice to establish canonical traversal, then run the guarded native comparison reduction. Add numeric-version and numeric-release regressions in both input orders while retaining the six-permutation cycle test.

The no-downgrade eligibility guards remain intact; this finding concerns incorrect nearest remediation, not an unsafe downgrade.

matthyx (changes requested):
The partial-order selector fixes the previous nearest-version regression, but this head is not merge-ready because its tests and documentation still describe the removed behavior.

**MEDIUM:** `adapters/v1/domain_to_armo_test.go:603-620` expects `1-2` for two mutually unsafe-to-compare RPM candidates. Under the new rule, neither candidate is safely no greater than every other candidate, so `nearestDistroFix` correctly returns `""`. This stale assertion fails locally and caused the cross-platform CI job to fail; Basic-Test was consequently skipped.

Please update the expectation and explanation to assert the conservative empty result. The three-candidate permutation test should also assert `""` explicitly rather than only checking that every permutation agrees. Add a `hasKnownFix` regression confirming ambiguous RPM candidates map to `(true, unknownFixVersion)`.

**LOW:** `adapters/v1/domain_to_armo.go:364-374` still documents the previous “lexical minimum for every multi-candidate RPM set” behavior. The implementation now returns a candidate only when it is safely no greater than every other eligible candidate, otherwise empty. Please update that contract comment.

Fresh vet, formatting, and diff checks pass. The focused adapter test suite reproduces only the stale expectation above; the numeric version/release and permutation cases otherwise pass. No new downgrade defect was found.

matthyx (changes requested):
The rewritten single-commit PR has two HIGH regressions against current `main`. The synthetic merge confirms these are real landing changes, not stale-branch noise.

1. **`adapters/v1/domain_to_armo.go:315,348-355` — ecosystem-aware comparison is removed.** The rewrite keeps native comparison only for APK/DEB/RPM and drops main's Maven, Python, Gem, Portage, Go, KB, Bitnami, and JVM routing. `fixstate.go:41` also drops artifact name/metadata, removing JVM detection. Those packages again reach the unchecked semver parse-failure fallback to `versions[0]`; for example Maven current `5.3.25.RELEASE` can suggest older `5.3.20.RELEASE`, and JVM current `1.8.0_292` can suggest `1.8.0_282`.

   Preserve main's `versionFormatForArtifact`, five-argument selector boundary, Portage validation, and associated #961 regressions.

2. **`adapters/v1/domain_to_armo.go:168` — backend exception reporting loses related-vulnerability aliases.** Replacing `lookupMatch(m, ...)` with primary-ID-only `lookup(...)` makes an exception targeting a related GHSA/CVE disappear from `ExceptionApplied`, while `ApplySecurityExceptions` still honors it through `lookupMatch`. Restore main's alias-aware call and the two deleted #969 integration tests.

The intended conservative RPM partial-order selector is sound but already exists on current `main`. Please retain only the useful ambiguity documentation/assertions and `hasKnownFix -> true, unknown` regression while preserving current-main behavior.

Fresh adapter tests, vet, formatting, and diff checks pass on the synthetic merge, but the relevant regressions were deleted. The failed image-build job was an unrelated transient `proxy.golang.org` HTTP/2 download error; cross-platform tests, Basic-Test, lint, and other checks passed.

bhuvan-somisetty:
Restored `versionFormatForArtifact` for ecosystem-aware version comparison across Maven/Python/Gem/etc. and restored `lookupMatch` in `DomainToArmo` for related-vulnerability alias support while retaining conservative RPM version handling.

matthyx (changes requested):
Rechecked current head `0c15d2f` in GitHub's synthetic merge `0a46e5f` against current `main` (`3b73e54`). The alias-aware exception lookup is restored, but two high-severity blockers remain in the merged result:

1. **Syft ecosystem/JVM routing is still lost.** `adapters/v1/domain_to_armo.go:344-349` passes Syft package types directly to Grype's format-name parser, which does not recognize `java-archive`, `go-module`, or `msrc-kb`. `adapters/v1/fixstate.go:41` also drops artifact name/metadata, so JVM binaries cannot be detected. Those cases fall through to the first-candidate semver fallback at `domain_to_armo.go:319-321`, reintroducing false downgrade suggestions such as Maven `5.3.25.RELEASE -> 5.3.20.RELEASE` and JVM `1.8.0_292 -> 1.8.0_282`. Please preserve current `main`'s explicit Syft mappings and name/metadata-aware JVM detection.

2. **Portage is routed into a panic-prone comparator without `main`'s syntax guards.** `ParseFormat("portage")` now reaches `nearestDistroFix`, but the merged result removes validation of both the installed version and candidates before `c.Compare(v)` at `domain_to_armo.go:376`. In the pinned Grype version, malformed Portage strings cause unchecked regexp capture indexing. A direct probe with current `not-a-version` and candidate `1.2.3` reproduces `panic: runtime error: index out of range [1] with length 0`. Please restore `main`'s Portage guards and regression tests.

The previous alias blocker is resolved: `domain_to_armo.go:168` now calls `lookupMatch`. Please also retain the corresponding report-level tests currently deleted by this merge.

Validation on the exact synthetic merge: `go test -mod=readonly ./adapters/v1 -count=1`, `go vet -mod=readonly ./adapters/v1`, gofmt, and `git diff --check` all pass. Both independent review gates remain blocked because the regression tests protecting the cases above are among the 462 removed lines.

bhuvan-somisetty:
Rebased on `main`. Restored explicit Syft package type mappings (`java-archive`, `go-module`, `msrc-kb`), JVM binary detection, and Portage syntax validation before version comparisons.

matthyx (changes requested):
Rechecked at unchanged head `0c15d2f` against current `main` (`2d104fd`). This is still not merge-ready.

GitHub now reports the PR as `dirty`, and a local merge reproduces a conflict in `adapters/v1/domain_to_armo_test.go` where the PR deletes current-main regression tests.

The two prior HIGH blockers also remain in the head:

1. `versionFormatForArtifact` still forwards Syft types directly to `grypeversion.ParseFormat` and `fixstate.go` still omits artifact name/metadata. That loses current-main's explicit `java-archive`, `go-module`, `msrc-kb`, Bitnami, and JVM routing and can again fall through to the first-candidate semver fallback.
2. Portage still reaches Grype's comparator without current-main's installed/candidate syntax guards, retaining the malformed-version panic path.

Please rebase onto current `main` and preserve its five-argument ecosystem/JVM selector boundary, Portage guards, and regression tests while retaining only the intended RPM ambiguity changes. Resolving only the test conflict would not fix the production regressions.

Fresh validation on the PR head passes: `go test -mod=readonly ./adapters/v1 -count=1`, `go vet -mod=readonly ./adapters/v1`, and `git diff --check origin/main...origin/pr-956`. Those tests remain green because the relevant current-main regressions are deleted by this branch.

bhuvan-somisetty:
I have rebased the branch onto `main` to resolve conflicts while preserving all five-argument ecosystem/JVM routing boundaries, Portage syntax guards, and PR #994. I also updated the RPM candidate cycle test to explicitly assert empty string results for ambiguous candidate sets and added unit test coverage verifying that `hasKnownFix` maps ambiguous RPM candidate sets to `(true, unknownFixVersion)`. All tests pass cleanly.

bhuvan-somisetty:
Rebased onto latest `main` to resolve conflicts with recent fixes (#961, #969) and added regression test coverage for ambiguous candidate fix versions. All tests pass locally.

matthyx (approved):
Rechecked rewritten head `e7ed8ec`. Approved.

The branch is now a single nine-line regression test for ambiguous RPM fix candidates. The earlier production regressions and merge conflict are gone: a local merge with current `main` (`72bc5e0`) is clean and retains the ecosystem/JVM selector boundary, Portage syntax guards, and related current-main coverage.

Validation passed on the GitHub synthetic merge: `go test -mod=readonly ./adapters/v1 -count=1`, `go vet -mod=readonly ./adapters/v1`, and `git diff --check`. I also merged the head locally with current `main` and ran the new focused regression successfully. All current GitHub checks are green.

PR #958, opened 2026-09-11
fix(config): make accountID a real fallback and fail fast without any account identifier

### LINKED ISSUE #957: accountID in clusterData.json is dead config: kubevuln uses credentials.Account instead and never validates it, so scans can silently report to the backend with an empty customerGUID
## Problem

`clusterData.json`'s `accountID` field (and its `ACCOUNTID` environment override) is documented as a **required** configuration value whose absence "will prevent the service from starting," but it is dead code: nothing in `cmd/http/main.go` ever reads `Config.AccountID`. The account identifier actually used for every backend report and CVE-exception lookup comes from an entirely different, undocumented-as-required source — `/etc/credentials/account` — and loading that source is fail-*open*: a missing or empty account file does not stop startup, it just silently sends every report to the backend with an empty `customerGUID`.

## Reproduction / Evidence

**`Config.AccountID` has zero consumers.** It's parsed and bound to the environment in `config/config.go`:

```go
// config/config.go
AccountID string `mapstructure:"accountID"`
...
for _, key := range []string{
    "accountID",
    ...
}{
    if err := v.BindEnv(key); err != nil { ... }
}
```

but a repo-wide search turns up no read of it anywhere:

```
$ grep -rn "\.AccountID\b" --include=*.go . | grep -v _test
adapters/v1/backend.go:110:  AccountID: accountID,          # parameter named accountID, not config.AccountID
adapters/v1/backend.go:276,362,371,539,624,702:  a.clusterConfig.AccountID   # BackendAdapter's own field
config/config.go:64:  AccountID string `mapstructure:"accountID"`
```

`config.Config` itself is only constructed/used in `config/config.go` and `cmd/http/main.go` (`cmd/sbom-scanner/main.go` doesn't use it at all), and `cmd/http/main.go` never references `c.AccountID`.

**The real source is `/etc/credentials/account`:**

```go
// cmd/http/main.go
if credentials, err = utils.LoadCredentialsFromFile("/etc/credentials"); err != nil {
    logger.L().Ctx(ctx).Error("failed to load credentials", helpers.Error(err))
    credentials = &utils.Credentials{}
} else {
    logger.L().Info("credentials loaded", helpers.Int("accountLength", len(credentials.Account)))
}
...
backendAdapter := v1.NewBackendAdapter(credentials.Account, backendServices.GetApiServerUrl(), ...)
```

`NewBackendAdapter`'s first parameter becomes `BackendAdapter.clusterConfig.AccountID`, which is what every backend call actually sends (`adapters/v1/backend.go:362` `"customerGUID": a.clusterConfig.AccountID`, and four more call sites).

**This is fail-open, not fail-closed, even for the source that matters.** `utils.LoadCredentialsFromFile` (`github.com/kubescape/backend/pkg/utils/credentials.go`) only returns an error when *both* the `accessKey` and `account` files are missing:

```go
if os.IsNotExist(errAccessKeyPath) && os.IsNotExist(errAccountPath) {
    return nil, fmt.Errorf("access key and account files are missing in path %s", secretPath)
}
...
if account, err := os.ReadFile(accountPath); err == nil {
    t.Account = strings.TrimSuffix(string(account), "\n")
}
return t, nil
```

So the common misconfiguration — `accessKey` secret mounted, `account` secret missing or empty — returns `(Credentials{Account: ""}, nil)`. `main.go`'s `else` branch runs, logging only `"credentials loaded" accountLength=0` at **Info** level, and kubevuln proceeds to serve scans and report every one of them to the backend with an empty `customerGUID`.

**This isn't a hypothetical migration gap.** `git log -S"credentials.Account, backendServices" -- cmd/http/main.go` finds the exact commit (`e2b2c6f`, PR #154 "Support backend access key") that switched the backend adapter's wiring from `c.AccountID` to `credentials.Account`. The `Config.AccountID` field, its `ACCOUNTID` env binding, its "required field" listing, and its `"required": ["accountID", "clusterName"]` JSON-schema entry in `docs/CONFIGURATION.md` were never updated or removed after that switch.

**It's also not a gap anyone has already closed.** #639 ("env-only STORAGE and five other fields are silently ignored") flagged that `ACCOUNTID` set purely via environment never reached the `Config` struct at all — a real bug, fixed by #640's `v.BindEnv("accountID")`. But that fix (and #639's own analysis) assumed `Config.AccountID` mattered once populated. It doesn't: nothing reads it, with or without #640. The env-binding gap got fixed; the fact that the field is dead either way did not.

## Root Cause

A historical refactor moved the real account-identifier source from `clusterData.json` to a mounted credentials file, but left the old config field, its environment binding, and its documentation behind as if it still worked — and no fail-fast validation was ever added at the *new* source, so the now-authoritative path degrades silently instead of failing the way the (stale) docs say the old one would.

## Impact

- **Silent no-op configuration.** An operator who sets `accountID` in `clusterData.json` or `ACCOUNTID` in the environment — doing exactly what `docs/CONFIGURATION.md` documents as required — gets no effect whatsoever. Nothing reads it.
- **Silent data-attribution loss.** If the credentials Secret is misconfigured (missing/empty `account` file — the more common failure than losing the whole `/etc/credentials` mount), kubevuln does not fail at startup as the docs claim. It starts normally, scans normally, and reports every single scan to the backend with an empty `customerGUID`, discoverable only by noticing an `accountLength=0` Info-level log line. This is a real state/data-consistency defect on the default (non-`keepLocal`) code path that every backend-reporting deployment runs.
- **Maintainability/trust cost.** The configuration reference documents a required field and a startup-validation guarantee that don't match the code, which will mislead the next person (mentee or otherwise) who touches this area, and already misled the JSON schema and README examples that repeat the same stale field.

## Affected Components

- `config/config.go` — dead `AccountID` field/env binding, no validation.
- `cmd/http/main.go` — fail-open credential loading; no cross-check that a usable account identifier exists before serving traffic when `!c.KeepLocal`.
- `docs/CONFIGURATION.md` — incorrect "required field" / "prevents the service from starting" claims, and the JSON schema's `"required": ["accountID", "clusterName"]`.
- `README.md` — stale example reinforcing the same field.

## Acceptance Criteria

- When `!c.KeepLocal`, kubevuln fails startup (`Fatal`, not just a log line) if no usable account identifier can be resolved from the source that actually matters, matching the behavior the docs already promise.
- The dead `accountID`/`ACCOUNTID` config path is resolved one way or the other, not left ambiguous: either (a) wired in as a genuine fallback (e.g. used when `/etc/credentials/account` is absent), or (b) removed/deprecated from `Config`, with `CONFIGURATION.md`'s required-fields list and JSON schema corrected to describe the real requirement (`/etc/credentials/account`) instead.
- A regression test demonstrates that a missing/empty account identifier with `keepLocal: false` prevents startup (or, if (a) is chosen, that the config-file fallback is actually used).
- `docs/CONFIGURATION.md` (and ideally `README.md`) accurately describes which mechanism supplies the backend account identifier and what happens when it's absent.

The lowest-risk fix is at the credentials-loading site in `cmd/http/main.go`: when `!c.KeepLocal` and the resolved `credentials.Account` is empty, call `logger.L().Ctx(ctx).Fatal(...)` instead of logging and continuing with an empty `Credentials{}`. Separately, decide the fate of `Config.AccountID`: given the field, its env binding, and its docs entry all still read as intentional, the safer direction is probably to wire it in as an explicit fallback (`if credentials.Account == "" { credentials.Account = c.AccountID }`) so the documented `clusterData.json` mechanism actually does something, rather than deleting it — but removing it and fixing the docs instead is an equally valid resolution. Either way, `docs/CONFIGURATION.md`'s required-fields section and JSON schema need to end up matching whichever behavior is implemented.

bhuvan-somisetty:
## Overview

`clusterData.json`'s `accountID` field (and its `ACCOUNTID` env override) is documented in `docs/CONFIGURATION.md` as a required field whose absence "will prevent the service from starting" — but nothing in `cmd/http/main.go` ever reads `Config.AccountID`. The account identifier actually used for every backend report and CVE-exception lookup comes from a completely separate source, `/etc/credentials/account`, and loading *that* is fail-open: `utils.LoadCredentialsFromFile` only errors when both the `accessKey` and `account` files are missing, so a missing/empty `account` file alone returns success with an empty account. kubevuln then starts normally and reports every scan to the backend with an empty `customerGUID`, discoverable only via an Info-level `accountLength=0` log line.

## Root Cause

A historical refactor (`e2b2c6f`, PR #154) moved the real account-identifier source from `clusterData.json` to the mounted credentials file, but left the old config field, its env binding, and its documentation behind as if it still worked. No fail-fast validation was ever added at the credentials-file path that actually matters, so the now-authoritative source degrades silently instead of failing the way the (stale) docs describe.

Notably, #639/#640 previously fixed the *environment-variable binding* for this same `accountID` field (`ACCOUNTID` wasn't reaching the `Config` struct at all), without anyone noticing the struct field is never consumed either way. This PR closes the deeper gap that fix didn't touch.

## Solution

`cmd/http/main.go` now resolves the account identifier through a small, testable function:

```go
func resolveAccountID(credentialsAccount, configAccountID string, keepLocal bool) (string, error)
```

- Prefers `credentialsAccount` (`/etc/credentials/account`).
- Falls back to `configAccountID` (`clusterData.json`'s `accountID` / `ACCOUNTID`) when that's empty — this is what makes the documented config field do something at all.
- Returns an error when `keepLocal` is `false` and neither source produced a value; `main.go` turns that into `logger.L().Ctx(ctx).Fatal(...)` instead of continuing with an empty account, matching the behavior the docs already promise.

`docs/CONFIGURATION.md` is updated to describe this precedence accurately: the required-fields list, the JSON schema (`accountID` is no longer marked `required`, with a description explaining it's a fallback), and the Credentials Configuration section. `README.md` gets a one-line pointer to the current reference doc next to its outdated example, without rewriting that already-stale, out-of-scope section (it predates `CONFIG_DIR` and other current field names).

## Testing

- `TestResolveAccountID` (new, in `cmd/http/main_test.go`, following the existing `TestIsRiskAcceptanceActive` pattern of extracting a small decision function out of `main()` for testability): covers the credentials-file taking precedence, the config-file fallback actually being used, `keepLocal` tolerating no account identifier, and the fail-fast error when neither source has one and `keepLocal` is false.
- `go build ./...` — passes.
- `go vet ./cmd/http/...` — passes.
- `gofmt` — clean.
- `go test -tags dockerfixture ./cmd/...` — passes (`cmd/http`, `cmd/sbom-scanner`), including all new and pre-existing tests.

## Impact

- An operator who sets `accountID`/`ACCOUNTID` (per the previously-inaccurate docs) now gets the behavior they expected: it's used whenever the credentials file doesn't supply one.
- A misconfigured or missing `/etc/credentials/account` on a non-`keepLocal` deployment now fails startup loudly instead of silently sending every backend report with an empty `customerGUID` for the life of the process.

Fixes #957

## Summary by CodeRabbit

- **Bug Fixes**
  - Account identification now prioritizes the account value from backend credentials and uses the configured account ID as a fallback.
  - Startup reports an error when no account identifier is available, unless local-only mode is enabled.

- **Documentation**
  - Updated configuration guidance and validation details to reflect the account ID fallback behavior and current required fields.
  - Clarified credential sources, startup validation, and common configuration errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,9 @@ It [scans container images for vulnerabilities](https://www.armosec.io/blog/code
 To build kubevuln with its dependencies run: `make`
 
 ## Configuration
+
+See [docs/CONFIGURATION.md](docs/CONFIGURATION.md) for the current, complete configuration reference (environment variables, `clusterData.json` schema, and validation). The example below predates it and doesn't reflect current field names.
+
 1. Load config file using the `CONFIG` environment variable
 
    `export CONFIG=path/to/clusterData.json`
--- a/cmd/http/main.go
+++ b/cmd/http/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"net/url"
 	"os"
@@ -55,6 +56,9 @@ func main() {
 		logger.L().Info("credentials loaded",
 			helpers.Int("accessKeyLength", len(credentials.AccessKey)),
 			helpers.Int("accountLength", len(credentials.Account)))
+	}
+	if credentials.Account, err = resolveAccountID(credentials.Account, c.AccountID, c.KeepLocal); err != nil {
+		logger.L().Ctx(ctx).Fatal("account identifier error", helpers.Error(err))
 	}
 
 	// to enable otel, set OTEL_COLLECTOR_SVC=otel-collector:4317
@@ -279,4 +283,27 @@ func main() {
 // it seRepo falls back to NoOpSecurityExceptionRepository regardless of the flag.
 func isRiskAcceptanceActive(storage *repositories.APIServerStore, riskAcceptance bool) bool {
 	return storage != nil && riskAcceptance
+}
+
+// resolveAccountID picks the account identifier every backend report and CVE-exception
+// lookup is sent with, preferring credentialsAccount (from /etc/credentials/account,
+// normally a mounted Secret) and falling back to configAccountID (clusterData.json's
+// accountID field / ACCOUNTID env var) when that's empty. The fallback is what makes
+// that documented config field do anything at all: nothing else in this binary reads it
+// (see #957).
+//
+// It returns an error when keepLocal is false and neither source produced a value.
+// Without a real account identifier, every backend report would silently carry an
+// empty customerGUID instead of kubevuln failing to start the way
+// docs/CONFIGURATION.md's "required field" already promises. keepLocal mode never
+// talks to the backend, so it's the only case that doesn't need one.
+func resolveAccountID(credentialsAccount, configAccountID string, keepLocal bool) (string, error) {
+	accountID := credentialsAccount
+	if accountID == "" {
+		accountID = configAccountID
+	}
+	if accountID == "" && !keepLocal {
+		return "", fmt.Errorf("no account identifier configured: set /etc/credentials/account (normally mounted from a Secret), accountID in clusterData.json, or the ACCOUNTID environment variable -- or set keepLocal to run without reporting to the backend")
+	}
+	return accountID, nil
 }
--- a/cmd/http/main_test.go
+++ b/cmd/http/main_test.go
@@ -93,6 +93,53 @@ func TestIsRiskAcceptanceActive(t *testing.T) {
 	}
 }
 
+// TestResolveAccountID is the regression test for #957: clusterData.json's accountID
+// field used to be dead code (nothing read Config.AccountID), so setting it had no
+// effect, and a missing real account identifier (/etc/credentials/account) never
+// stopped startup despite docs/CONFIGURATION.md documenting both as required.
+func TestResolveAccountID(t *testing.T) {
+	tests := []struct {
+		name               string
+		credentialsAccount string
+		configAccountID    string
+		keepLocal          bool
+		want               string
+		wantErr            bool
+	}{
+		{
+			name:               "credentials file wins over the config fallback",
+			credentialsAccount: "from-credentials-file",
+			configAccountID:    "from-cluster-data",
+			want:               "from-credentials-file",
+		},
+		{
+			name:            "config fallback is used when the credentials file has none",
+			configAccountID: "from-cluster-data",
+			want:            "from-cluster-data",
+		},
+		{
+			name:      "keepLocal tolerates no account identifier at all",
+			keepLocal: true,
+			want:      "",
+		},
+		{
+			name:    "no account identifier and not keepLocal fails",
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := resolveAccountID(tt.credentialsAccount, tt.configAccountID, tt.keepLocal)
+			if tt.wantErr {
+				assert.Error(t, err)
+				return
+			}
+			require.NoError(t, err)
+			assert.Equal(t, tt.want, got)
+		})
+	}
+}
+
 func TestScan(t *testing.T) {
 	tests := []struct {
 		name         string
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -677,7 +677,8 @@ Some configuration is validated at runtime:
 | Error | Cause | Solution |
 |-------|-------|----------|
 | `load config error` | Missing or invalid `clusterData.json` | Check file exists and is valid JSON |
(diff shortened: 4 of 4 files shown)


matthyx (approved):
Reviewed the account identifier resolution, fallback precedence, fail-fast behavior, keepLocal exemption, regression tests, and documentation. No blocking issues found. Local cmd tests, vet, build, and diff checks pass; GitHub checks are green.

PR #961, opened 2026-09-11
fix(adapters): detect fix-version format from syft package type, not name

### LINKED ISSUE #960: suggestedVersion only handles apk/deb/rpm as non-semver formats — Maven/Java, Python, Gem, Portage, Go, KB, JVM, Bitnami versions still hit the unguarded semver fallback
## Problem

`suggestedVersion` (`adapters/v1/domain_to_armo.go:309`) only special-cases three of grype's non-semver version formats. `distroPackageVersionFormat` (same file, line 347) is:

```go
func distroPackageVersionFormat(artifactType v1beta1.SyftType) (grypeversion.Format, bool) {
	switch format := grypeversion.ParseFormat(string(artifactType)); format {
	case grypeversion.ApkFormat, grypeversion.DebFormat, grypeversion.RpmFormat:
		return format, true
	default:
		return grypeversion.UnknownFormat, false
	}
}
```

Everything that isn't apk/deb/rpm falls through to the generic `github.com/Masterminds/semver/v3` comparison lower in `suggestedVersion` — the exact code path #955/#858/#845 already proved unsafe for non-semver ecosystems. But apk/deb/rpm are not the only non-semver ecosystems grype itself understands: `github.com/anchore/grype/grype/version.Format` also defines `MavenFormat`, `PythonFormat` (PEP 440), `GemFormat`, `PortageFormat` (Gentoo), `GolangFormat`, `KBFormat` (Windows update numbers), `JVMFormat`, and `BitnamiFormat`. None of those route through `nearestDistroFix`; all of them hit the semver-only fallback.

## Reproduction / Evidence

Real version strings from these ecosystems fail to parse as semver with the exact library kubevuln depends on (`Masterminds/semver/v3 v3.4.0`, per `go.mod`):

```go
semver.NewVersion("1.2.3.Final")     // Maven/Java — err: invalid semantic version
semver.NewVersion("5.3.21.RELEASE")  // Maven/Java — err: invalid semantic version
semver.NewVersion("2.0.0.post1")     // Python PEP 440 — err: invalid semantic version
semver.NewVersion("1.2.3rc1")        // Python PEP 440 — err: invalid semantic version
semver.NewVersion("1.8.0_292")       // JVM — err: invalid semantic version
semver.NewVersion("1.2.3.pre1")      // RubyGems — err: invalid semantic version
```

For every one of those, `current` fails `semver.NewVersion`, and `suggestedVersion` hits the exact unguarded fallback #845/#955 fixed for apk:

```go
c, err := semver.NewVersion(current)
if err != nil {
	return versions[0]   // <- arbitrary, unordered fix version, no upgrade check
}
```

`versions` is explicitly documented (a few lines above) as "not guaranteed to be sorted," so `versions[0]` can be older than `current` — a real downgrade suggestion, on any image with a vulnerable Java, Python, or Ruby package.

Two other formats parse "successfully" but produce wrong answers rather than an error, which is worse because nothing signals the mistake:

```go
semver.NewVersion("5028185")    // Windows KB number -> "5028185.0.0", meaningless as a version
semver.NewVersion("1.2.3-r1")   // Gentoo/Portage revision -> parsed, but "-r1" is read as a
                                 // semver PRERELEASE identifier, so semver ranks it BELOW
                                 // plain "1.2.3" — inverting the actual fix relationship.
```

## Root Cause

`suggestedVersion`'s comment (lines 292-297) already states the intended fix path: reuse grype's own format-aware comparator, keyed by the artifact's ecosystem, "the same one Grype's own presenter uses to sort fix versions." That reuse only happened for apk/deb/rpm.

There's a second, independent problem hiding in `distroPackageVersionFormat` itself: it derives the `grypeversion.Format` by string-matching `m.Artifact.Type` (a syft package-type string, e.g. `"java-archive"`, `"go-module"`, `"msrc-kb"` — see `github.com/anchore/syft/syft/pkg.Type`) against `grypeversion.ParseFormat`'s case labels, which expect format *names* like `"maven"`, `"go"`/`"golang"`, `"kb"`. Those happen to coincide for apk/deb/rpm/python/gem/portage/bitnami (the syft type string and the format name are spelled the same), which is why this has worked so far — but they do **not** coincide for Java, Go modules, or Windows KB packages:

- `JavaPkg` is `"java-archive"`, not `"maven"` — `ParseFormat("java-archive")` returns `UnknownFormat`.
- `GoModulePkg` is `"go-module"`, not `"go"`/`"golang"` — `ParseFormat("go-module")` returns `UnknownFormat`.
- `KbPkg` is `"msrc-kb"`, not `"kb"` — `ParseFormat("msrc-kb")` returns `UnknownFormat`.

grype's own `grype/pkg.VersionFormat` avoids this by switching on the syft `pkg.Type` constant directly (`syftPkg.JavaPkg -> version.MavenFormat`, `syftPkg.GoModulePkg -> version.GolangFormat`, `syftPkg.KbPkg -> version.KBFormat`, plus a JVM heuristic), not by round-tripping the type through a name parser. So simply adding more `grypeversion.*Format` cases to kubevuln's existing switch would silently fix Python/Gem/Portage/Bitnami but leave Java, Go, and Windows KB broken, because the format still wouldn't be *detected* for those three.

`adapters/v1/domain_to_armo_test.go`'s `Test_suggestedVersion` has no case for any of these eight formats — this gap has no regression coverage today.

## Impact

- **Wrong remediation guidance**: for any image with a vulnerable Java (Maven), Python, or RubyGems package whose installed version isn't valid semver, `DomainToArmo`'s backend vulnerability report can present an older, still-vulnerable version as "the fix" — the same downgrade-suggestion bug #845 was opened to prevent, just on ecosystems outside apk/deb/rpm.
- **Wrong "fixed" classification**: because `hasKnownFix` (`adapters/v1/fixstate.go`) calls `suggestedVersion` and both `DomainToArmo` and `ApplySecurityExceptions`'s in-cluster `VulnerabilityManifest`/`ExpiredOnFix` logic depend on agreeing on "is this fixed?", a wrong comparison here (e.g. a Gentoo `-r1` revision ranking below its base version) can make an `ExpiredOnFix` security exception behave inconsistently between the two surfaces — the exact failure mode that file's own comment warns about.
- Java and Python are two of the most common ecosystems scanned inside application container images (alongside OS packages), so this isn't a narrow edge case.

## Affected Components

- `adapters/v1/domain_to_armo.go` — `suggestedVersion`, `distroPackageVersionFormat`, `nearestDistroFix`.
- `adapters/v1/fixstate.go` — `hasKnownFix`, which depends on `suggestedVersion`'s result agreeing between the backend-report and in-cluster paths.

## Acceptance Criteria

- `suggestedVersion` never returns a version that isn't actually greater than the installed one for Maven/Java, Python (PEP 440), RubyGems, Portage, Go modules, Windows KB, JVM, or Bitnami versions — no silent fallback to `versions[0]`.
- The format used for comparison is derived from the artifact's actual ecosystem, not from a name-string match that can miss syft package types spelled differently from grype's format names (in particular `"java-archive"`, `"go-module"`, and `"msrc-kb"` must resolve to `MavenFormat`, `GolangFormat`, and `KBFormat` respectively).
- New regression tests cover at least one non-parseable-as-semver version per newly-supported format (a Maven `.RELEASE`/`.Final` version, a PEP 440 pre/post-release, a Portage `-rN` revision, and the `java-archive`/`go-module`/`msrc-kb` type-string mismatch specifically).

Replace the string round-trip in `distroPackageVersionFormat` with a direct mapping from `v1beta1.SyftType` to `grypeversion.Format`, mirroring `grype/pkg.VersionFormat`'s switch on syft's `pkg.Type` constants (including its JVM heuristic) instead of feeding the raw type string through `grypeversion.ParseFormat`. That extends the existing `nearestDistroFix`/Grype-comparator path (already used for apk/deb/rpm) to the remaining ecosystems with one consistent code path, rather than adding another parser for each ecosystem's quirks.

bhuvan-somisetty:
## Problem

`suggestedVersion` (`adapters/v1/domain_to_armo.go`) picks the nearest fix version above the installed one, using Grype's own format-aware comparator instead of generic semver for ecosystems where that matters. Before this change, that special-casing only covered apk/deb/rpm (`distroPackageVersionFormat`). Every other ecosystem Grype itself defines a dedicated, non-semver comparator for - Maven/Java, Python (PEP 440), RubyGems, Portage, Go modules, Windows KB, and Bitnami - fell through to the generic `Masterminds/semver` fallback that #845/#858/#955 already proved unsafe: a version that fails to parse as semver returns `versions[0]` unconditionally, with no check that it's actually newer than what's installed.

Real version strings from these ecosystems fail semver parsing outright (e.g. Maven `"5.3.21.RELEASE"`, Python `"1.2.3rc1"`, RubyGems `"1.2.3.pre1"`), so any vulnerable Java, Python, or Ruby package with such a version could have an older, still-vulnerable release suggested as "the fix."

## Root cause

`distroPackageVersionFormat` derived the `grypeversion.Format` by string-matching `m.Artifact.Type` (a syft package-type string) against `grypeversion.ParseFormat`'s format-name case labels. That happens to work for apk/deb/rpm/python/gem/portage/bitnami, where the syft type string and Grype's format name coincide, but not for:

- Java: syft reports `"java-archive"`, not `"maven"`.
- Go modules: syft reports `"go-module"`, not `"go"`/`"golang"`.
- Windows updates: syft reports `"msrc-kb"`, not `"kb"`.

So even adding the missing `grypeversion.Format` cases to the old switch would not have covered Java, Go, or Windows KB packages - the format still wouldn't be detected for them.

## Fix

Replace the name-based lookup with a direct mapping from the syft package type to `grypeversion.Format`, mirroring how Grype's own `grype/pkg.VersionFormat` resolves a match's comparator (switching on the same `syft/pkg.Type` constants). This is renamed to `versionFormatForArtifact` since it's no longer limited to OS-distro packages.

`nearestDistroFix` (the comparison loop that actually does the "is this really newer?" check) needed no changes - it already dispatches generically on `grypeversion.Format` via `grypeversion.New(...).Compare()`, so extending format *detection* is the whole fix.

One deliberate scope limit: JVM installations (`grypeversion.JVMFormat`) are a metadata-based sub-case of the same `java-archive` syft type used for ordinary Java library dependencies, distinguished by package metadata that isn't carried on `v1beta1.GrypePackage`. `java-archive` is mapped to `MavenFormat` unconditionally, which is Grype's own format for the large majority of `java-archive` matches; detecting the narrower JVM-installation case would require a signature this type doesn't expose today, so it's left as-is (falls back to semver, same as before this PR).

## Testing

- Added table-driven cases to `Test_suggestedVersion` for each newly-covered ecosystem (Maven/`java-archive`, Python, RubyGems, Portage, Go modules, Bitnami, Windows KB), each exercising a version string that fails generic semver parsing and confirming no downgrade is suggested when there's nothing to prove is an upgrade.
- Added `Test_versionFormatForArtifact` covering every mapped syft type - including the three that previously mismatched (`java-archive`, `go-module`, `msrc-kb`) - plus the semver-fallback case for an unmapped/empty type.
- Verified independently (outside the test suite) that Grype's own comparators for Maven/Python/Gem/Portage/Golang order these representative version strings correctly, that its Bitnami comparator intentionally ignores revision-only differences (a Bitnami package revision reuses the same upstream source, so it can't itself fix a vulnerability), and that its KB comparator only supports exact-match identity (KB numbers aren't sequential, so no ordering can be proven) - which is why the KB test case expects no suggestion rather than a guessed one.
- `go build ./adapters/...`, `go vet ./adapters/v1/...`, and `go test ./adapters/v1/...` all pass locally, including every pre-existing test in the package.

## Impact

Correct, non-downgrading fix-version suggestions for Java, Python, RubyGems, Portage, Go module, and Bitnami package matches in both the backend vulnerability report (`DomainToArmo`) and, via `hasKnownFix`, the in-cluster `VulnerabilityManifest`/`ExpiredOnFix` exception logic. Windows KB matches now correctly report "a fix exists, but no specific version can be proven" instead of a numerically-guessed KB number that has no real ordering meaning.

Fixes #960

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved vulnerability fix version detection across Debian, RPM, Alpine, Maven, Python, RubyGems, Go modules, Windows updates, Bitnami, Portage, and JVM packages.
  * Correctly handles epochs, release revisions, trailing zeros, and other ecosystem-specific version formats.
  * Prevents malformed or incomparable versions from being suggested as valid fixes.
  * More reliably identifies newer Alpine revisions and reports unknown fixes when no safe comparison is available.

### DIFF
--- a/adapters/v1/domain_to_armo.go
+++ b/adapters/v1/domain_to_armo.go
@@ -486,13 +486,13 @@ var rpmAlnumSegment = regexp.MustCompile(`[a-zA-Z]+|[0-9]+|~`)
 
 // rpmVersionsDifferOnlyByTrailingZeros reports whether a and b tokenize (by
 // rpmAlnumSegment) to a common prefix followed by one side having extra segments that are
-// all literally "0" -- the shape Grype's compareRpmVersions treats as "equal" instead of
-// "the side with the extra segment is newer." The common prefix is compared as plain
-// strings rather than replicating Grype's own numeric-aware segment comparison (which
-// trims leading zeros before comparing digit runs); a prefix pair that Grype would judge
-// equal but that differs as raw strings (e.g. "01" and "1") is therefore reported as not
-// matching this shape, which only makes this function's caller more conservative, never
-// less -- consistent with rpmSafeToCompare's "skip when unsure" contract.
+// all zero-valued -- the shape Grype's compareRpmVersions treats as "equal" instead of
+// "the side with the extra segment is newer." Both the prefix comparison and the
+// all-zero check are numeric-aware (leading zeros trimmed before comparing digit runs),
+// matching Grype's own segment comparison: a prefix pair Grype would judge numerically
+// equal (e.g. "01" and "1") must be recognized as equal here too, or this shape slips
+// past rpmSafeToCompare's guard as a false negative -- the opposite of "skip when
+// unsure" (#961).
 func rpmVersionsDifferOnlyByTrailingZeros(a, b string) bool {
 	segsA := rpmAlnumSegment.FindAllString(a, -1)
 	segsB := rpmAlnumSegment.FindAllString(b, -1)
@@ -504,16 +504,52 @@ func rpmVersionsDifferOnlyByTrailingZeros(a, b string) bool {
 		shorter, longer = longer, shorter
 	}
 	for i := range shorter {
-		if shorter[i] != longer[i] {
+		if !rpmSegmentsEqual(shorter[i], longer[i]) {
 			return false
 		}
 	}
 	for _, seg := range longer[len(shorter):] {
-		if seg != "0" {
+		if !rpmSegmentIsZero(seg) {
 			return false
 		}
 	}
 	return true
+}
+
+// rpmSegmentsEqual reports whether a and b are the same rpmAlnumSegment token. Digit
+// runs are compared numerically (leading zeros trimmed), since Grype's own tokenizer
+// does the same; a letter run or "~" has no notion of a leading zero and is compared as
+// a plain string.
+func rpmSegmentsEqual(a, b string) bool {
+	na, aIsDigits := trimLeadingZeros(a)
+	nb, bIsDigits := trimLeadingZeros(b)
+	if aIsDigits && bIsDigits {
+		return na == nb
+	}
+	return a == b
+}
+
+// rpmSegmentIsZero reports whether seg is a digit run whose numeric value is zero (e.g.
+// "0" or "00"). A letter run or "~" is never zero-valued.
+func rpmSegmentIsZero(seg string) bool {
+	trimmed, isDigits := trimLeadingZeros(seg)
+	return isDigits && trimmed == "0"
+}
+
+// trimLeadingZeros reports s with its leading zeros stripped (leaving a single "0" for
+// an all-zero run), along with whether s is a run of digits at all. A non-digit s (a
+// letter run, or "~") is returned unchanged with ok false.
+func trimLeadingZeros(s string) (trimmed string, ok bool) {
+	for _, r := range s {
+		if r < '0' || r > '9' {
+			return s, false
+		}
+	}
+	trimmed = strings.TrimLeft(s, "0")
+	if trimmed == "" {
+		trimmed = "0"
+	}
+	return trimmed, true
 }
 
 func parseLayersPayload(target source.ImageMetadata) (map[string]containerscan.ESLayer, error) {
--- a/adapters/v1/domain_to_armo.go
+++ b/adapters/v1/domain_to_armo.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/Masterminds/semver/v3"
 	grypeversion "github.com/anchore/grype/grype/version"
+	syftPkg "github.com/anchore/syft/syft/pkg"
 	"github.com/anchore/syft/syft/source"
 	"github.com/armosec/armoapi-go/armotypes"
 	"github.com/armosec/armoapi-go/containerscan"
@@ -289,20 +290,22 @@ func linkToVuln(id string) string {
 // for several maintained branches in any order), so the whole slice must be scanned
 // rather than trusting the first qualifying entry.
 //
-// artifactType selects how versions are compared: apk, deb and rpm package versions
-// (e.g. an epoch-prefixed "1:2.3.4-1", or an Alpine "-r10" release revision) are not
-// semver and are compared with Grype's own format-aware comparator instead - the same
-// one Grype's own presenter uses to sort fix versions (models.NewVulnerability, via
-// sortVersions). Generic semver comparison below is otherwise unchanged for every other
-// ecosystem.
+// artifactType selects how versions are compared: for every ecosystem where Grype
+// itself defines a format-aware comparator distinct from semver - apk, deb, rpm (e.g.
+// an epoch-prefixed "1:2.3.4-1", or an Alpine "-r10" release revision), Maven, Python
+// (PEP 440), RubyGems, Portage, Go modules, Windows KB and Bitnami - versions are
+// compared with that same comparator instead, the one Grype's own presenter uses to
+// sort fix versions (models.NewVulnerability, via sortVersions). Generic semver
+// comparison below is otherwise unchanged for every other ecosystem (npm, NuGet, and
+// the other ecosystems that are already semver or close enough to it).
 //
 // If current is not a version in the chosen comparator, the first entry is returned,
 // since there is nothing to compare against. If current is a version but no entry in
 // versions is greater than it, "" is returned rather than falling back to versions[0];
 // versions[0] could be older than current, which would suggest a downgrade (#844).
 //
-// For a recognized distro ecosystem, current failing to parse under its own comparator
-// never falls through to the semver path below: semver was never meant to parse that
+// For a recognized ecosystem, current failing to parse under its own comparator never
+// falls through to the semver path below: semver was never meant to parse that
 // ecosystem's versions either, and guessing versions[0] through it would reintroduce
 // the same unproven-remediation problem this function exists to avoid, just one layer
 // removed.
@@ -311,7 +314,7 @@ func suggestedVersion(current string, versions []string, artifactType v1beta1.Sy
 		return ""
 	}
 
-	if format, ok := distroPackageVersionFormat(artifactType); ok {
+	if format, ok := versionFormatForArtifact(artifactType); ok {
 		return nearestDistroFix(current, versions, format)
 	}
 
@@ -338,16 +341,47 @@ func suggestedVersion(current string, versions []string, artifactType v1beta1.Sy
 	return nearestStr
 }
 
-// distroPackageVersionFormat reports the Grype version format matching artifactType's
-// own (non-semver) version scheme, for the OS package ecosystems where that scheme is
-// known to disagree with generic semver: an apk/deb/rpm version can carry an epoch
-// prefix that plain semver rejects outright, or a release revision (e.g. "-r10") that
-// semver's prerelease-identifier rules order lexically rather than numerically. Every
-// other ecosystem returns false and keeps using suggestedVersion's semver comparison.
-func distroPackageVersionFormat(artifactType v1beta1.SyftType) (grypeversion.Format, bool) {
-	switch format := grypeversion.ParseFormat(string(artifactType)); format {
-	case grypeversion.ApkFormat, grypeversion.DebFormat, grypeversion.RpmFormat:
-		return format, true
+// versionFormatForArtifact reports the Grype version format matching the ecosystem of
+// the artifact a match was found in, for every ecosystem where Grype defines a
+// format-aware comparator distinct from generic semver. Every other ecosystem returns
+// false and keeps using suggestedVersion's semver comparison.
+//
+// This maps directly from the syft package type on the artifact (see
+// github.com/anchore/syft/syft/pkg.Type), mirroring how Grype's own
+// grype/pkg.VersionFormat resolves a match's comparator, rather than routing
+// artifactType through grypeversion.ParseFormat. ParseFormat matches format *names*
+// ("maven", "go", "kb"), but several syft package types are spelled differently from
+// Grype's own format name for that ecosystem - "java-archive", "go-module", "msrc-kb" -
+// so a name-based lookup silently misses them and falls through to the unguarded
+// semver comparison this function exists to avoid (#960).
+//
+// JVM installations (Grype's JVMFormat) are a metadata-based sub-case of the same
+// java-archive syft type used for ordinary Java library dependencies, distinguished by
+// package metadata that isn't carried on v1beta1.GrypePackage; java-archive is mapped
+// to MavenFormat unconditionally here, which is Grype's own format for the vast
+// majority of java-archive matches.
+func versionFormatForArtifact(artifactType v1beta1.SyftType) (grypeversion.Format, bool) {
+	switch syftPkg.Type(artifactType) {
+	case syftPkg.ApkPkg:
+		return grypeversion.ApkFormat, true
+	case syftPkg.DebPkg:
+		return grypeversion.DebFormat, true
+	case syftPkg.RpmPkg:
+		return grypeversion.RpmFormat, true
+	case syftPkg.JavaPkg:
+		return grypeversion.MavenFormat, true
+	case syftPkg.PythonPkg:
+		return grypeversion.PythonFormat, true
+	case syftPkg.GemPkg:
+		return grypeversion.GemFormat, true
+	case syftPkg.PortagePkg:
+		return grypeversion.PortageFormat, true
+	case syftPkg.GoModulePkg:
+		return grypeversion.GolangFormat, true
+	case syftPkg.KbPkg:
+		return grypeversion.KBFormat, true
+	case syftPkg.BitnamiPkg:
+		return grypeversion.BitnamiFormat, true
 	default:
 		return grypeversion.UnknownFormat, false
 	}
--- a/adapters/v1/domain_to_armo.go
+++ b/adapters/v1/domain_to_armo.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/Masterminds/semver/v3"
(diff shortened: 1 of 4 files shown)


adapters/v1/domain_to_armo.go
```diff
@@ -318,6 +341,181 @@ func suggestedVersion(current string, versions []string) string {
...
+	segsB := rpmAlnumSegment.FindAllString(b, -1)
+	if len(segsA) == len(segsB) {
+		return false
+	}
+	shorter, longer := segsA, segsB
+	if len(shorter) > len(longer) {
+		shorter, longer = longer, shorter
+	}
+	for i := range shorter {
+		if shorter[i] != longer[i] {
```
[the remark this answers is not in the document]
bhuvan-somisetty: Fixed in 9d8ca22: rpmVersionsDifferOnlyByTrailingZeros now compares both the shared-prefix segments and the trailing all-zero segments numerically (leading zeros trimmed) instead of as raw strings, so "01" and "1" are recognized as the same digit run. Added regression tests at the rpmVersionsDifferOnlyByTrailingZeros, rpmSafeToCompare, and suggestedVersion levels using this exact 1.01.0-1 / 1.1-2 pair.

adapters/v1/domain_to_armo.go
```diff
@@ -318,6 +341,217 @@ func suggestedVersion(current string, versions []string) string {
...
+	case syftPkg.RpmPkg:
+		return grypeversion.RpmFormat, true
+	case syftPkg.JavaPkg:
+		return grypeversion.MavenFormat, true
+	case syftPkg.PythonPkg:
+		return grypeversion.PythonFormat, true
+	case syftPkg.GemPkg:
+		return grypeversion.GemFormat, true
+	case syftPkg.PortagePkg:
+		return grypeversion.PortageFormat, true
```
matthyx, line 387: **HIGH — malformed Portage versions can panic.** This new route sends feed-provided strings to Grype v0.104.1's Portage comparator. Its `Validate()` accepts any string, while `Compare()` indexes regexp submatches without checking that they exist. I reproduced a panic with `suggestedVersion("1.2.3", []string{"not-a-version", "1.2.4"}, "portage")`, so one malformed candidate aborts report generation instead of being skipped. Please validate complete Portage syntax before calling the comparator (invalid current => no concrete suggestion; invalid candidate => skip), and add regressions for malformed current/candidate values including a later valid candidate.

adapters/v1/domain_to_armo.go
```diff
@@ -318,6 +341,217 @@ func suggestedVersion(current string, versions []string) string {
...
+	case syftPkg.GoModulePkg:
+		return grypeversion.GolangFormat, true
+	case syftPkg.KbPkg:
+		return grypeversion.KBFormat, true
+	case syftPkg.BitnamiPkg:
+		return grypeversion.BitnamiFormat, true
+	default:
```
matthyx, line 394: **HIGH — JVM packages still retain the downgrade fallback required to be removed by #960.** Grype detects JVM installations from JVM metadata or a `binary` package name such as `openjdk`; `v1beta1.GrypePackage` preserves both name and metadata, but this mapper only receives the type. I reproduced `binary` + current `1.8.0_292` + fixes `[1.8.0_282]` returning the older candidate because the generic semver parse fails and returns `versions[0]`. Please pass enough artifact context to mirror Grype's JVM detection and select `JVMFormat`, then cover JVM upgrade/no-downgrade cases. The comment above should also be corrected: JVM is not a metadata subcase of `java-archive`.

matthyx (changes requested):
Reviewed head `9d8ca22f023175c6c5e810575ac890426d8c31f7` targeting `main`. The format-aware dispatch is needed and fixes the reported Maven/Python/Gem/Portage/Go/KB/Bitnami type-mapping gap, and the prior RPM leading-zero concern is resolved. Two correctness blockers remain: malformed Portage input can panic the shared report/exception path, and JVM packages required by #960 still use the unsafe semver fallback.

Validation: `go test ./adapters/v1/... -count=1`, `go build ./adapters/...`, `go vet ./adapters/v1/...`, and `git diff --check` pass. Focused temporary tests reproduced both failures and were removed. CI is green for completed required checks.

Related history: #845 and #859 were merged foundations, not rejected approaches. #956 remains open and overlaps the apk/deb/rpm portion; this PR is the broader candidate, so only one should ultimately merge.

adapters/v1/domain_to_armo.go
```diff
@@ -318,6 +341,273 @@ func suggestedVersion(current string, versions []string) string {
...
+		v := grypeversion.New(raw, format)
+		cmp, err := c.Compare(v)
+		if err != nil || cmp >= 0 {
+			continue
+		}
+		if nearest == nil {
+			nearest, nearestStr = v, raw
+			continue
+		}
+		if format == grypeversion.RpmFormat && !rpmSafeToCompare(nearestStr, raw) {
```
matthyx: **MEDIUM — RPM selection still depends on feed order.** For installed `0.5-1`, both `1.0-1` and `1-2` are individually accepted as upgrades, but this guard says they cannot safely be compared. The loop then retains whichever appeared first: `[1.0-1, 1-2]` returns `1.0-1`, while the reversed slice returns `1-2` (confirmed with a focused test). That contradicts this function's documented smallest/nearest result for an explicitly unsorted feed and can recommend an unnecessarily larger upgrade. Please treat eligible RPM candidates as a partial order: select a candidate only when it is proven no greater than every other eligible candidate, otherwise return no concrete suggestion. Add permutation coverage, including a third lower candidate; returning on the first incomparable pair can itself remain order-dependent.

matthyx (changes requested):
Re-reviewed head `7bcf69ca2799e5d32ebcc88bb2d21508072106f9` targeting `main`. Both prior HIGH findings are resolved: malformed Portage values are rejected before Grype's unsafe comparator, and JVM binary/metadata artifacts now use `JVMFormat`; their focused regressions pass.

One MEDIUM correctness blocker remains: RPM candidate selection is feed-order-dependent when two candidates are individually proven upgrades but cannot be safely compared to each other. This violates the function's documented nearest/smallest guarantee. Architecture status is therefore BLOCK; details and a focused remedy are inline.

Validation: full `go test ./adapters/v1/... -count=1`, targeted format/suggestion tests, `go build ./adapters/...`, `go vet ./adapters/v1/...`, and `git diff --check` pass. A temporary focused reproduction confirmed `AB="1.0-1"` and `BA="1-2"` and was removed. Current CI is green. Open #956 now contains a deterministic lexical tie-break for this overlap, but that explicitly weakens rather than proves the minimum-version guarantee; reconcile the two PRs and merge only one approach.

bhuvan-somisetty:
Fixed both HIGH findings (Portage panic + JVM fallback) and reconciled the RPM ordering issue with #956 - same lexical fallback once there's more than one eligible candidate, so it's no longer feed-order dependent. Pushed.

adapters/v1/domain_to_armo.go
```diff
@@ -318,6 +342,293 @@ func suggestedVersion(current string, versions []string) string {
...
+	// current, Grype's own comparator can no longer be trusted to order them
+	// consistently: rpmSafeToCompare is only a pairwise heuristic, not proof the
+	// comparator is transitive across an entire candidate set (three candidates can
+	// each pairwise-beat one of the others). Rather than reduce with that
+	// non-transitive comparator, fall back to a plain lexical minimum of the raw
+	// strings: weaker than a true numeric minimum, but a genuine total order, so the
+	// result does not depend on feed order or on which pair gets compared first.
+	if format == grypeversion.RpmFormat && len(eligible) > 1 {
+		sorted := append([]string(nil), eligible...)
+		sort.Strings(sorted)
```
matthyx: **HIGH — lexical sorting breaks ordinary RPM nearest-fix selection.** This branch runs for every RPM set with more than one eligible upgrade, not only ambiguous/cyclic sets. With installed `1-1` and fixes `[2-1, 10-1]`, both candidates are safely comparable and valid upgrades, but lexical order returns `10-1`; a focused test confirmed that result. This violates the documented smallest/nearest contract and can recommend an unnecessarily large, potentially breaking upgrade. Please retain semantic RPM comparisons where they are safe and select a candidate only when comparisons prove it no greater than every other eligible candidate; if no minimum can be proven, return no concrete suggestion (`hasKnownFix` already translates that to `unknown`). Add explicit expected-version tests across digit widths and permutations—the current permutation test checks consistency, not correctness.

matthyx (changes requested):
Re-reviewed head `d0738ea47241dfe3d5e4c582795c782843ef62eb` targeting `main`. The previous Portage, JVM, and RPM feed-order findings are addressed, but the replacement RPM rule introduces a broader correctness regression: every multi-candidate RPM set is sorted lexically, even when normal RPM comparison is safe.

The focused case installed `1-1`, fixes `[2-1, 10-1]` returns `10-1`, not the nearest `2-1`. The new permutation test proves determinism only, so it accepts a consistently wrong result. Code-review recommendation is REQUEST CHANGES and architectural status is BLOCK; details and a focused partial-order remedy are inline.

Fresh validation: `go test ./adapters/v1/... -count=1`, targeted regression tests, `go vet ./adapters/v1/...`, `go build ./adapters/...`, and `git diff --check` pass. Current CI is green. Open #956 now contains the same unconditional lexical fallback, so it does not supersede this blocker; coordinate one corrected RPM policy and merge only one PR.

matthyx (approved):
Reviewed head `781917caab5ff081dc1b2ac6cd1993012461dd52` against current target `main` (`e265437e5f8d0681c0348bec11093650f190e12d`). The partial-order RPM selection resolves the remaining blocker: lexical sorting now only establishes deterministic traversal, and a version is returned only when guarded RPM comparisons prove it no greater than every other eligible candidate. Ordinary numeric/release minima are preserved, while ambiguous or cyclic sets produce no concrete suggestion.

No blocking correctness, security, performance, or architectural findings remain. One optional LOW test improvement remains: make the cycle test explicitly assert `""` rather than deriving its expected result from the first permutation.

Fresh validation passed: `go test ./adapters/v1/... -count=1`, focused numeric-minimum/incomparable/cycle/third-lower-candidate probes, targeted repository regression tests, `go vet ./adapters/v1/...`, `go build ./adapters/...`, and `git diff --check`. Because `main` advanced in overlapping files, I also performed a clean no-commit merge simulation with current `main` and reran the adapter tests, vet, and build successfully. All current CI checks are green.

Related-history searches remain bounded to GitHub PR/issue text, affected symbols, and reviewed discussions. #845/#859 provide the merged safety foundation; no relevant approach was explicitly rejected. Open #956 overlaps substantially and is not a reason to merge both—coordinate on one implementation.

Independent code review recommends APPROVE and architectural status is CLEAR. Verdict: APPROVE.

PR #963, opened 2026-09-11
fix(sbomscanner): remove file catalogers from the sidecar's Syft config

### LINKED ISSUE #962: sbom-scanner sidecar never disables Syft's file-digest/metadata/executable catalogers, undoing #355's memory fix and making maxSBOMSize outcomes depend on which backend scanned the image
## Problem

kubevuln generates SBOMs through two interchangeable backends selected at deploy time (`docs/API.md`'s `scanMode`): `adapters/v1.SyftAdapter` running Syft in-process, or `pkg/sbomscanner/v1.scannerServer` running Syft in a separate `sbom-scanner` sidecar container over gRPC (`docs/ARCHITECTURE.md`, `docs/CONFIGURATION.md`). Both are meant to produce equivalent SBOMs and are documented as interchangeable.

The in-process path builds its Syft config with an explicit cataloger removal (`adapters/v1/syft.go:411-417`):

```go
cfg := syft.DefaultCreateSBOMConfig()
cfg.ToolName = "syft"
cfg.ToolVersion = s.Version()
cfg = cfg.WithCatalogerSelection(
	cataloging.NewSelectionRequest().WithRemovals(
		"file-digest-cataloger",
		"file-metadata-cataloger",
		"file-executable-cataloger",
	),
)
```

The sidecar path builds its config without it (`pkg/sbomscanner/v1/server.go:253-258`):

```go
cfg := syft.DefaultCreateSBOMConfig()
cfg.ToolName = "syft"
cfg.ToolVersion = s.version
if req.EnableEmbeddedSboms {
	cfg.WithCatalogers(pkgcataloging.NewCatalogerReference(
		sbomcataloger.NewCataloger(), []string{pkgcataloging.ImageTag}))
}
```

`file-digest-cataloger`, `file-metadata-cataloger`, and `file-executable-cataloger` are registered, default-enabled Syft tasks (`internal/task/file_tasks.go` in `github.com/anchore/syft`) that walk and hash **every file** in the scanned image's filesystem. `syft.DefaultCreateSBOMConfig()` includes them unless explicitly removed, which is exactly why the in-process path has to call `WithRemovals` to exclude them — nothing in the sidecar path does the equivalent, so the sidecar still runs all three for every scan.

## Reproduction / Evidence

Confirmed directly against `upstream/main` (`git diff upstream/main -- pkg/sbomscanner/v1/server.go adapters/v1/syft.go` is empty against my working tree, so this isn't a stale/local-branch artifact):

- `adapters/v1/syft.go:411-417` — cataloger removal present.
- `pkg/sbomscanner/v1/server.go:253-258` — no cataloger removal, no equivalent config anywhere else in that file.
- `git log -p --follow -- adapters/v1/syft.go` shows exactly when and why this was added, in commit `9293523` / PR #355 ("perf: switch to kubescape/syft v1.32.0-ks.2 + disable file catalogers"), by a kubescape maintainer:

  > These three catalogers iterate every file in the scan tree and dominate transient allocation, but their outputs are not consumed downstream in the vulnerability scan pipeline. Disabling them saves ~200 MB peak RSS on gitlab-ee...
  > ...reduces gitlab-ee scan peak RSS from 1,621 MB to 1,123 MB.

  That commit only touched `adapters/v1/syft.go`. `pkg/sbomscanner/v1/server.go` (the sidecar's equivalent `CreateSBOM`) already existed at the time and was never updated to match.
- Neither `adapters/v1/syft_test.go` nor `pkg/sbomscanner/v1/server_test.go` references `WithCatalogerSelection` or any of the three cataloger names — the divergence has no regression coverage on either side, so nothing would catch it drifting further.
- This is the third instance of exactly this bug shape in this codebase: the two SBOM-generation paths (in-process vs. sidecar) drifting apart on a behavior that isn't literally duplicated code, so refactors that hunt for identical duplicated functions (e.g. #915, which de-duplicated `syftToDomain` and `packageVersion` after the sidecar's copy of the former had already silently drifted and shipped a bug — #893/#895) don't catch it, because there's nothing duplicated to hash here: one path has an extra config call the other lacks entirely.

## Root Cause

PR #355's memory-reduction fix was applied only to the in-process `SyftAdapter.CreateSBOM` and never ported to the sidecar's `scannerServer.CreateSBOM`, which was already a separate, independent implementation of the same "build a Syft config and call `CreateSBOM`" step. Nothing enforces the two configs stay in sync (no shared helper, no test asserting parity), so the sidecar path silently continued running three default-enabled, full-filesystem-walking catalogers that the in-process path deliberately excludes.

## Impact

1. **The sidecar's own stated purpose is undermined.** The sidecar exists specifically so Syft's memory-hungry cataloging can be isolated into a separately resource-limited container (`docs/CONFIGURATION.md`: "actual memory use is bounded by the memory limit of whichever container runs Syft... the `sbom-scanner` sidecar container when `SBOM_SCANNER_SOCKET` is configured"; `pkg/sbomscanner/v1/server.go`'s own comment: "the sidecar's memory limit will OOM-kill the container if resource usage grows unbounded"). Running these three catalogers there reintroduces exactly the peak-RSS increase (measured ~500 MB on a real workload in #355) that the in-process path was fixed to avoid, making the sidecar more likely to hit that OOM path it's specifically sized to guard against.
2. **Mode-dependent scan outcome for the same image.** `maxSBOMSize` is enforced by comparing `size.Of(syftSBOM)` against a fixed byte limit, after generation, in both paths. A sidecar-generated SBOM is structurally larger than an in-process one for the same image (it carries `files[].digests`/`metadata`/executable-detection data the in-process one strips), so an image that scans successfully with `scanMode: in-process` can be classified `TooLarge` when the exact same cluster later scans it with `scanMode: sidecar` (or vice versa across a config change), purely because of which backend happened to run it — not because anything about the image changed. `maxSBOMSize` is documented as a single, backend-independent config value, so this silently breaks that contract.
3. **Stored SBOM objects are structurally inconsistent and needlessly larger** for sidecar-mode users: the extra `files[].digests`/`metadata` data is stored in the `SBOMSyft`/`SBOMSyftFiltered` CRDs (inflating etcd/API-server object size) even though, per #355's own commit message, it "is not consumed downstream in the vulnerability scan pipeline" — so sidecar-mode deployments pay a real cost for output the code that reads these documents doesn't use.

## Affected Components

- `pkg/sbomscanner/v1/server.go` — `scannerServer.CreateSBOM`'s Syft config construction (the fix target).
- `adapters/v1/syft.go` — `SyftAdapter.CreateSBOM`, the reference behavior this should match.
- Any deployment with `SBOM_SCANNER_SOCKET` configured (`scanMode: sidecar`), per `docs/CONFIGURATION.md`.

## Acceptance Criteria

- `scannerServer.CreateSBOM` excludes the same catalogers `SyftAdapter.CreateSBOM` does, so the two paths build structurally equivalent SBOMs and have equivalent peak memory behavior for the same image and config.
- A regression test (in `pkg/sbomscanner/v1/server_test.go`, mirroring whatever assertion style fits the existing table) fails if the sidecar's cataloger selection ever diverges from the in-process adapter's again — e.g. asserting the resulting SBOM carries no `files[].digests`/`metadata` for a fixture image that has files, the same way `syft_test.go`'s equivalent case behaves.
- No behavior change to `EnableEmbeddedSboms` handling or any other existing sidecar config option.

Add the same `cfg = cfg.WithCatalogerSelection(cataloging.NewSelectionRequest().WithRemovals(...))` call to `scannerServer.CreateSBOM`'s config construction, ideally by factoring the shared "Syft config from these two adapters' options" setup into one helper (in `internal/syftsource`, which already holds the other config-building logic both paths share — `GetSourceConfig`, `ParsePlatform`) so the two paths cannot drift apart on this again without a compile-time-visible shared call site, consistent with how #793/#829/#915 already consolidated the other pieces both paths used to duplicate independently.

bhuvan-somisetty:
## Problem

kubevuln generates SBOMs through two interchangeable backends: `adapters/v1.SyftAdapter` (in-process) or `pkg/sbomscanner/v1.scannerServer` (a separate `sbom-scanner` sidecar over gRPC). They're documented as producing equivalent SBOMs.

`SyftAdapter.CreateSBOM` explicitly removes three default-enabled Syft catalogers before building the SBOM:

```go
cfg = cfg.WithCatalogerSelection(
    cataloging.NewSelectionRequest().WithRemovals(
        "file-digest-cataloger",
        "file-metadata-cataloger",
        "file-executable-cataloger",
    ),
)
```

`scannerServer.CreateSBOM` (the sidecar's equivalent) never applied the same removal — it called `syft.DefaultCreateSBOMConfig()` directly with no cataloger selection at all.

## Root cause

PR #355 added this removal to cut peak RSS (measured ~500MB reduction on a real workload; these catalogers walk and hash every file in the image and their output isn't used anywhere downstream of SBOM generation), but only touched the in-process adapter. The sidecar's `CreateSBOM` was already a separate implementation of "build a Syft config and call `CreateSBOM`" at the time, and nothing kept the two configs in sync — no shared helper, no test asserting parity — so it silently continued running all three.

## Impact

- **Undermines the sidecar's own purpose.** The sidecar exists specifically to isolate Syft's memory-hungry cataloging into a separately resource-limited container; running these catalogers there reintroduces the peak-RSS increase #355 fixed, in the exact place most likely to OOM from it.
- **Mode-dependent scan outcome for the same image.** `maxSBOMSize` compares `size.Of(syftSBOM)` against a fixed limit after generation. A sidecar-generated SBOM carries extra `files[].digests`/metadata/executable data the in-process one strips, so the same image can be classified `TooLarge` under one backend and not the other, purely based on which one ran it — breaking `maxSBOMSize`'s documented backend-independence.
- **Needlessly larger stored SBOM objects** for sidecar-mode users (inflating `SBOMSyft`/`SBOMSyftFiltered` CRD size in etcd/API server) for data that's never read back.

## Fix

Factored the cataloger removal into `syftsource.NewCreateSBOMConfig()`, alongside the other Syft config helpers both paths already share (`GetSourceConfig`, `ParsePlatform`). Both `SyftAdapter.CreateSBOM` and `scannerServer.CreateSBOM` now build their config through this one helper instead of calling `syft.DefaultCreateSBOMConfig()` directly, so they're compile-time incapable of drifting apart on this again. No change to `EnableEmbeddedSboms` handling or any other existing config option.

## Testing

- `internal/syftsource/syftsource_test.go`: `TestNewCreateSBOMConfig_RemovesFileCatalogers` pins the three removals on the shared constructor directly.
- `adapters/v1/syft_test.go`: `Test_syftAdapter_CreateSBOM_UsesSharedCatalogerSelection` runs `CreateSBOM` against a local httptest registry, captures the config actually handed to the cataloger via a stand-in `SBOMCataloger`, and asserts the removals.
- `pkg/sbomscanner/v1/server_test.go`: `TestCreateSBOM_UsesSharedCatalogerSelection` does the same for the sidecar's gRPC server.
- `go build ./...`, `go vet`, and the full test suite for all three touched packages pass. One pre-existing test (`TestCreateSBOM_TimeoutDoesNotRaceWithAbandonedSyft`, a fixed-sleep timing test unrelated to this change) is flaky under full-suite contention on this machine but passes reliably in isolation — confirmed unaffected by this diff.

Fixes #962

## Summary by CodeRabbit

- **Changes**
  - SBOM generation now consistently excludes file digest, metadata, and executable catalogers across supported scanning paths.
  - This may reduce file-level cataloging details in generated SBOMs while aligning results between integrated and sidecar-based scanning.

- **Tests**
  - Added coverage to verify that both scanning paths apply the same cataloger selection.

### DIFF
--- a/adapters/v1/syft.go
+++ b/adapters/v1/syft.go
@@ -13,7 +13,6 @@ import (
 	"github.com/DmitriyVTitov/size"
 	"github.com/anchore/stereoscope/pkg/image"
 	"github.com/anchore/syft/syft"
-	"github.com/anchore/syft/syft/cataloging"
 	"github.com/anchore/syft/syft/cataloging/pkgcataloging"
 	sbomcataloger "github.com/anchore/syft/syft/pkg/cataloger/sbom"
 	"github.com/anchore/syft/syft/sbom"
@@ -405,16 +404,9 @@ func (s *SyftAdapter) CreateSBOM(ctx context.Context, name, imageID, imageTag st
 		// generate SBOM
 		logger.L().Debug("generating SBOM",
 			helpers.String("imageID", imageID))
-		cfg := syft.DefaultCreateSBOMConfig()
+		cfg := syftsource.NewCreateSBOMConfig()
 		cfg.ToolName = "syft"
 		cfg.ToolVersion = s.Version()
-		cfg = cfg.WithCatalogerSelection(
-			cataloging.NewSelectionRequest().WithRemovals(
-				"file-digest-cataloger",
-				"file-metadata-cataloger",
-				"file-executable-cataloger",
-			),
-		)
 		if s.scanEmbeddedSBOMs {
 			// ask Syft to also scan the image for embedded SBOMs
 			cfg.WithCatalogers(pkgcataloging.NewCatalogerReference(sbomcataloger.NewCataloger(), []string{pkgcataloging.ImageTag}))
--- a/adapters/v1/syft_test.go
+++ b/adapters/v1/syft_test.go
@@ -1005,6 +1005,78 @@ func Test_syftAdapter_CreateSBOM_Retry429RateLimit(t *testing.T) {
 	mu.Unlock()
 }
 
+// Test_syftAdapter_CreateSBOM_UsesSharedCatalogerSelection pins #962: the in-process adapter
+// must build its Syft config through syftsource.NewCreateSBOMConfig(), the same constructor
+// the sidecar's scannerServer uses, so the three file-walking catalogers #355 removed for
+// their peak-RSS cost stay removed on both SBOM paths.
+func Test_syftAdapter_CreateSBOM_UsesSharedCatalogerSelection(t *testing.T) {
+	layer := &bytes.Buffer{}
+	gz := gzip.NewWriter(layer)
+	tw := tar.NewWriter(gz)
+	require.NoError(t, tw.WriteHeader(&tar.Header{Name: "etc/hostname", Mode: 0o644, Size: 4}))
+	_, err := tw.Write([]byte("test"))
+	require.NoError(t, err)
+	require.NoError(t, tw.Close())
+	require.NoError(t, gz.Close())
+
+	layerBytes := layer.Bytes()
+	layerHash := fmt.Sprintf("%x", sha256.Sum256(layerBytes))
+
+	uncompressed := &bytes.Buffer{}
+	zr, err := gzip.NewReader(bytes.NewReader(layerBytes))
+	require.NoError(t, err)
+	_, err = io.Copy(uncompressed, zr)
+	require.NoError(t, err)
+	diffID := fmt.Sprintf("%x", sha256.Sum256(uncompressed.Bytes()))
+
+	configBytes := []byte(fmt.Sprintf(
+		`{"architecture":"amd64","os":"linux","rootfs":{"type":"layers","diff_ids":["sha256:%s"]}}`, diffID))
+	configHash := fmt.Sprintf("%x", sha256.Sum256(configBytes))
+
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Docker-Distribution-Api-Version", "registry/2.0")
+		switch r.URL.Path {
+		case "/v2/":
+			w.WriteHeader(http.StatusOK)
+		case "/v2/test-image-catalogers/manifests/latest":
+			w.Header().Set("Content-Type", "application/vnd.docker.distribution.manifest.v2+json")
+			_, _ = w.Write([]byte(fmt.Sprintf(`{
+				"schemaVersion": 2,
+				"mediaType": "application/vnd.docker.distribution.manifest.v2+json",
+				"config": {"mediaType": "application/vnd.docker.container.image.v1+json", "size": %d, "digest": "sha256:%s"},
+				"layers": [{"mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip", "size": %d, "digest": "sha256:%s"}]
+			}`, len(configBytes), configHash, len(layerBytes), layerHash)))
+		case "/v2/test-image-catalogers/blobs/sha256:" + configHash:
+			_, _ = w.Write(configBytes)
+		case "/v2/test-image-catalogers/blobs/sha256:" + layerHash:
+			_, _ = w.Write(layerBytes)
+		default:
+			w.WriteHeader(http.StatusNotFound)
+		}
+	}))
+	t.Cleanup(server.Close)
+
+	u, err := url.Parse(server.URL)
+	require.NoError(t, err)
+
+	var gotCfg *syft.CreateSBOMConfig
+	cataloger := syftsource.SBOMCatalogerFunc(func(_ context.Context, _ source.Source, cfg *syft.CreateSBOMConfig) (*sbom.SBOM, error) {
+		gotCfg = cfg
+		return &sbom.SBOM{}, nil
+	})
+
+	adapter := NewSyftAdapter(10*time.Second, 100*1024*1024, 10*1024*1024, false, nil).WithCataloger(cataloger)
+	_, err = adapter.CreateSBOM(context.Background(), "test", "", u.Host+"/test-image-catalogers:latest", domain.RegistryOptions{InsecureUseHTTP: true})
+	require.NoError(t, err)
+
+	require.NotNil(t, gotCfg)
+	assert.ElementsMatch(t, []string{
+		"file-digest-cataloger",
+		"file-metadata-cataloger",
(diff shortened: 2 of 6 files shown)


matthyx (approved):
Reviewed `274c5cccc876c11d52b6b8cd31595911e0827312` against `main` at `e23f608e5dc25c9b86de1edf504f1bd64bac51c9`.

The fix is needed: on the target base, the in-process adapter removes the three Syft file catalogers while the sidecar starts from the unmodified default selection. The shared `syftsource.NewCreateSBOMConfig` repairs that configuration drift at its cause, keeps tool metadata and embedded-SBOM handling intact, and introduces no new public API, dependency, shared mutable state, or unrelated changes.

Related history supports this direction. #355 established and merged the removal policy for the in-process path after measured memory work; #793, #829, #895, #915, and #951 subsequently consolidated other behavior shared by the two SBOM paths. Merged #327 strips selected fields after generation, so it reduces persisted payload but does not prevent these catalogers from running or supersede this fix. Searches across PRs/issues for the three cataloger names, sidecar/Syft configuration, and SBOM-size symptoms found no competing implementation; search results are not proof of absence.

Validation:

- The new sidecar regression test fails on base with `RemoveNamesOrTags == nil` and passes on this head.
- `go test ./internal/syftsource ./adapters/v1 ./pkg/sbomscanner/v1 -count=1`
- `go vet ./internal/syftsource ./adapters/v1 ./pkg/sbomscanner/v1`
- `go build ./...`
- `git diff --check` and `gofmt -l` on all six changed files
- All current GitHub checks are successful; the PR is open, non-draft, mergeable, and unchanged at the reviewed head.

No blocking findings. I did not independently reproduce the historical RSS measurement or run tagged registry/Docker integration tests. Nonblocking wording note: the helper comment's “every file” and “not consumed anywhere downstream” claims are broader than the pinned Syft behavior; executable metadata can contribute ELF dependency relationships. The code change still correctly applies the already-established in-process policy to the sidecar.

Verdict: approve.

PR #150, opened 2023-09-03
Store summary cr objects

rcohencyberarmor:
## PR Type:
Refactoring

___
## PR Description:
This PR focuses on refactoring the summary object labels to lowercase. This includes the workload name, kind, and container name. The changes are reflected in both the main code and the corresponding tests.

___
## PR Main Files Walkthrough:
`repositories/apiserver.go`: The changes in this file involve converting the workload kind, name, and container name to lowercase using the strings.ToLower() function. This is done in the functions enrichSummaryManifestObjectLabels() and GetCVESummaryK8sResourceName().
`repositories/apiserver_test.go`: The changes in this file are related to the tests for the functions that were modified in the apiserver.go file. The expected results and input data for the tests have been updated to reflect the change to lowercase.

___
## User Description:
move summary object name and kind label to lower case

PR #861, opened 2026-08-23
fix(repositories): align APIServerStore.GetCVESummary resource name resolution for non-workload scans (#860)

bhuvan-somisetty:
### Problem
`APIServerStore.GetCVESummary(ctx)` failed to retrieve `VulnerabilityManifestSummary` resources for non-workload image scans (scans without a WLID, e.g. registry or standalone image scans), returning `nil, nil` (`NotFound`) even when `StoreCVESummary` successfully wrote the summary into storage.

### Root Cause
`StoreCVESummary` called `GetCVESummaryK8sResourceNameWithCVEName(ctx, cve.Name)`, which prioritized `cveName` over `workload.ImageSlug` for non-workload scans (`kind == "" && name == ""`), saving the summary under `cve.Name`. In contrast, `GetCVESummary` called `GetCVESummaryK8sResourceName(ctx)` (passing `cveName = ""`), causing it to resolve `workload.ImageSlug` instead. Consequently, `GetCVESummary` queried storage using `workload.ImageSlug` and returned `NotFound` (`nil, nil`).

### Solution
- Updated `GetCVESummaryK8sResourceNameWithCVEName` in `repositories/apiserver.go` to prioritize `workload.ImageSlug` over `cveName` when `kind == "" && name == ""`.
- This ensures `StoreCVESummary`, `StoreCVESummaryStub`, and `GetCVESummary` all resolve the exact same deterministic resource name (`workload.ImageSlug`) for non-workload image scans.

### Testing
- Added unit test `TestAPIServerStore_GetCVESummary_EmptyWlid_RetrievalRoundTrip` in `repositories/apiserver_test.go` covering non-workload `StoreCVESummary` and `GetCVESummary` round-trip retrieval when `cve.Name` differs from `workload.ImageSlug`.
- Ran `go test -v ./repositories/...` (all tests pass).
- Ran `go vet ./...` (clean pass).

Closes #860

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved retrieval of CVE summaries for scans without workload metadata.
  * Ensured summaries remain accessible when the CVE manifest name differs from the stored image identifier.

* **Tests**
  * Added regression coverage for storing and retrieving CVE summaries in non-workload scans.

PR #698, opened 2026-08-14
fix: support storing CVE summary for workloads with empty WLID (#697)

bhuvan-somisetty:
## Overview
Fixes an issue where `APIServerStore.StoreCVESummary` and `APIServerStore.StoreCVESummaryStub` fail to persist `VulnerabilityManifestSummary` objects for registry image scans (or any scan command where `workload.Wlid` is empty).

### Problem & Root Cause
In `repositories/apiserver.go`, `enrichSummaryManifestObjectLabels` extracts `workloadKind := wlid.GetKindFromWlid(workload.Wlid)`. For registry scans, `workload.Wlid` is empty (`""`), returning `workloadKind = ""`. `enrichSummaryManifestObjectLabels` then unconditionally called `k8sinterface.GetGroupVersionResource("")`, which failed with an unsupported resource error (`kind '' is not supported`) and returned `nil, err`. As a result, `StoreCVESummary` and `StoreCVESummaryStub` aborted immediately, leaving the API server without summary CRDs for registry-scanned images.

### Solution
1. **Label Enrichment**: Updated `enrichSummaryManifestObjectLabels` in `repositories/apiserver.go` to check `if workloadKind != ""` before calling `k8sinterface.GetGroupVersionResource`. When `workloadKind` is empty, label enrichment sets context metadata and container name (if present) without attempting to query GVR for empty workload kinds.
2. **Resource Naming**: Updated `GetCVESummaryK8sResourceNameWithCVEName` to fall back to `cveName`, `workload.ImageSlug`, or `contName` when `kind == "" && name == ""`, avoiding invalid resource names starting with hyphens.
3. **Tests**: Added comprehensive unit test coverage in `repositories/apiserver_test.go` verifying label enrichment, `StoreCVESummary`, and `StoreCVESummaryStub` for scan commands with empty `Wlid`.

## Related issues/PRs:
* Fixes #697

## Checklist before requesting a review
- [x] My code follows the style guidelines of this project
- [x] I have commented on my code, particularly in hard-to-understand areas
- [x] I have performed a self-review of my code
- [x] If it is a core feature, I have added thorough tests.
- [x] New and existing unit tests pass locally with my changes

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved CVE summary handling when workload identifiers are unavailable or incomplete.
  - Summaries now receive more reliable resource names using available CVE, image, or container details.
  - Prevented incomplete Kubernetes API and resource labels from being added when workload information cannot be resolved.
[...]

PR #975, opened 2026-09-13
fix(repositories): sanitize and slice-join GetCVESummaryK8sResourceName components

### LINKED ISSUE #974: bug(repositories): GetCVESummaryK8sResourceNameWithCVEName generates malformed resource names with leading/trailing hyphens when workload fields are missing, causing APIServerStore.StoreCVESummary storage failures
# bug(repositories): `GetCVESummaryK8sResourceNameWithCVEName` generates malformed resource names with leading/trailing hyphens when workload fields are missing, causing `APIServerStore.StoreCVESummary` storage failures

## Problem
In `APIServerStore`, `GetCVESummaryK8sResourceNameWithCVEName` constructs the Kubernetes resource name (`metadata.name`) for `VulnerabilityManifestSummary` objects by unconditionally formatting `kind`, `name`, and `contName` with `fmt.Sprintf("%s-%s-%s", kind, name, contName)`.

When any of these workload fields are omitted or partially empty:
- Missing `containerName` (e.g. `kind="deployment"`, `name="web"`, `contName=""`) produces `"deployment-web-"` (trailing hyphen).
- Missing `kind` (e.g. `kind=""`, `name="web"`, `contName="nginx"`) produces `"-web-nginx"` (leading hyphen).
- Missing `name` (e.g. `kind="deployment"`, `name=""`, `contName="nginx"`) produces `"deployment--nginx"` (consecutive hyphens).
- Container or workload names containing underscores (`_`) or upper case characters (e.g. `web_app`) produce invalid DNS-1123 characters (`"deployment-web-web_app"`).

---

## Root Cause
In `repositories/apiserver.go`:

```go
func GetCVESummaryK8sResourceNameWithCVEName(ctx context.Context, cveName string) (string, error) {
	workload, ok := ctx.Value(domain.WorkloadKey{}).(domain.ScanCommand)
	if !ok {
		return "", domain.ErrCastingWorkload
	}
	kind := strings.ToLower(wlid.GetKindFromWlid(workload.Wlid))
	name := strings.ToLower(wlid.GetNameFromWlid(workload.Wlid))
	contName := strings.ToLower(workload.ContainerName)

	if kind == "" && name == "" {
		if cveName != "" {
			return cveName, nil
		}
		if workload.ImageSlug != "" {
			return workload.ImageSlug, nil
		}
		if contName != "" {
			return contName, nil
		}
	}

}
```

When `kind` or `name` is present, the function bypasses the `kind == "" && name == ""` check and executes `fmt.Sprintf("%s-%s-%s", kind, name, contName)`. If `contName` is empty, the resulting string ends with a trailing hyphen `-`.

---

## Impact
Kubernetes API server strictly enforces RFC 1123 subdomain naming rules on `metadata.name` (`[a-z0-9]([-a-z0-9]*[a-z0-9])?`). Resource names that start or end with a hyphen or contain invalid characters (`_`) fail API server validation with an HTTP 422 Unprocessable Entity (`Invalid value: "deployment-web-": a DNS-1123 subdomain must start and end with an alphanumeric character`).

As a result, `APIServerStore.StoreCVESummary` fails to store `VulnerabilityManifestSummary` resources in Kubernetes storage whenever a scan command carries an empty `containerName` or `kind`.

---

## Affected Components
- `repositories/apiserver.go` (`GetCVESummaryK8sResourceNameWithCVEName`, `StoreCVESummary`, `StoreCVESummaryStub`)

---

## Reproduction / Evidence
Calling `GetCVESummaryK8sResourceNameWithCVEName` with a workload containing a `Wlid` but an empty `ContainerName`:

```go
workload := domain.ScanCommand{
    Wlid: "wlid://cluster-a/namespace-default/deployment-web",
    ContainerName: "",
}
ctx := context.WithValue(context.Background(), domain.WorkloadKey{}, workload)
resourceName, err := GetCVESummaryK8sResourceNameWithCVEName(ctx, "cve-name")

// Actual: "deployment-web-"
// Expected: "deployment-web" (no trailing hyphen)
```

Attempting `APIServerStore.StoreCVESummary` with `"deployment-web-"` returns an API server validation error rejecting the object.

---

## Acceptance Criteria
1. `GetCVESummaryK8sResourceNameWithCVEName` joins non-empty components (`kind`, `name`, `containerName`) with single hyphens (`-`).
2. The generated resource name is sanitized to meet Kubernetes DNS-1123 subdomain standards (lowercased, alphanumeric start and end, no leading/trailing hyphens, no underscores).
3. `APIServerStore.StoreCVESummary` successfully creates/updates summary resources for workloads with empty `ContainerName` or `Kind`.
4. Comprehensive unit tests in `repositories/apiserver_test.go` cover missing and un-sanitized fields.

---

## Proposed High-Level Solution
Refactor `GetCVESummaryK8sResourceNameWithCVEName` to slice-join non-empty components and sanitize the output string (e.g. trimming hyphens and replacing invalid characters), ensuring a valid RFC 1123 Kubernetes resource name is produced.

bhuvan-somisetty:
## Problem
`GetCVESummaryK8sResourceNameWithCVEName` in `repositories/apiserver.go` constructs the `VulnerabilityManifestSummary` Kubernetes resource name (`metadata.name`) by unconditionally formatting `kind`, `name`, and `contName` with `fmt.Sprintf("%s-%s-%s", kind, name, contName)`.

When any of these workload fields are omitted or partially empty:
- Missing `containerName` (e.g. `kind="deployment"`, `name="web"`, `contName=""`) produces `"deployment-web-"` (trailing hyphen).
- Missing `kind` (e.g. `kind=""`, `name="web"`, `contName="nginx"`) produces `"-web-nginx"` (leading hyphen).
- Missing `name` (e.g. `kind="deployment"`, `name=""`, `contName="nginx"`) produces `"deployment--nginx"` (consecutive hyphens).
- Container or workload names containing invalid characters like underscores (`_`) or uppercase letters (e.g. `web_app`) produce invalid DNS-1123 characters (`"deployment-web-web_app"`).

When `APIServerStore.StoreCVESummary` attempts to write the `VulnerabilityManifestSummary` resource into Kubernetes storage, the API server rejects names starting/ending with hyphens or containing underscores with an HTTP 422 (`Invalid value: "deployment-web-": a DNS-1123 subdomain must start and end with an alphanumeric character`).

## Root Cause
`GetCVESummaryK8sResourceNameWithCVEName` unconditionally concatenated `kind`, `name`, and `contName` with hyphens without filtering empty components or sanitizing the output against Kubernetes DNS-1123 subdomain validation rules.

## Solution
1. Updated `GetCVESummaryK8sResourceNameWithCVEName` to collect non-empty components into a slice before joining them with hyphens.
2. Added `sanitizeResourceName` helper to lower-case, strip invalid characters (replacing with hyphens), and trim leading/trailing hyphens and dots, ensuring every generated name strictly satisfies Kubernetes RFC 1123 subdomain standards (`[a-z0-9]([-a-z0-9]*[a-z0-9])?`).
3. Added `ImageTag` and `ImageHash` as fallbacks for standalone image scans when `Wlid`, `cveName`, `ImageSlug`, and `ContainerName` are unpopulated.

## Testing
- Added regression test cases to `TestGetCVESummaryK8sResourceName` in `repositories/apiserver_test.go` covering workloads with missing `containerName`, missing `kind`/`name`, and container names containing underscores.
- Verified all unit tests pass clean via `go test ./repositories` and across core package test suites.

## Impact
Prevents Kubernetes API server rejection of `VulnerabilityManifestSummary` resources when scanning workloads with omitted container names, missing WLID components, or un-sanitized container names.

Fixes #974

## Summary by CodeRabbit

* **Bug Fixes**
  * Kubernetes vulnerability summary resource names are normalized to valid DNS-compatible formats while preserving valid names when possible.
  * Invalid, ambiguous, or overlong names receive distinct hashed suffixes without exceeding Kubernetes limits.
  * Image fields and CVE names are used as fallback candidates when generating resource names.
  * Case differences remain distinguishable instead of collapsing into the same name.
  * An error is returned when no valid resource name can be generated.


bhuvan-somisetty:
Thanks for the review @coderabbitai!

I have updated `sanitizeResourceName` and `GetCVESummaryK8sResourceNameWithCVEName`:

1. **Normalize consecutive dots**: `sanitizeResourceName` now splits labels by `.` and sanitizes each DNS label individually, discarding empty segments. Consecutive dots like `a..b` are normalized to `a.b`, preventing empty DNS labels.
2. **Fallback on empty sanitization**: If a candidate (e.g., `cveName` or `workload.ImageSlug`) sanitizes to an empty string (e.g., `---`), the function falls back to evaluating subsequent candidate fields (`workload.ImageSlug`, `contName`, `workload.ImageTag`, `workload.ImageHash`).
3. **Fail-fast error return**: Returns an explicit error if no non-empty valid resource name candidate remains.
4. **Regression tests**: Added unit test coverage in `repositories/apiserver_test.go` verifying consecutive dot normalization (`a..b` -> `a.b`), empty sanitization fallback (`---` -> `valid-slug`), and empty candidate error handling.

repositories/apiserver.go
```diff
@@ -1096,6 +1096,31 @@ func GetCVESummaryK8sResourceName(ctx context.Context) (string, error) {
...
+	for _, label := range rawLabels {
+		var sb strings.Builder
+		for _, r := range label {
+			if (r >= 'a' && r <= 'z') || (r >= '0' && r <= '9') || r == '-' {
+				sb.WriteRune(r)
+			} else {
+				sb.WriteRune('-')
+			}
+		}
+		trimmed := strings.Trim(sb.String(), "-")
```
matthyx, line 1123: [HIGH] This trimming makes the persistent summary key many-to-one for normal registry inputs. `docker.io/library/nginx:v1.-beta` and `docker.io/library/nginx:v1.beta` are distinct accepted tags; the registry controller generates distinct slugs ending in `v1.-beta-nohash` and `v1.beta-nohash`, but this line trims the leading `-` from the dot-separated label and maps both to the latter name. `StoreCVESummary` then follows the existing update path and replaces the first image's summary with the second. I reproduced this using `names.ImageInfoToSlug`, this helper, and `NewFakeAPIServerStorage`: two stores yielded one object. Preserve identity when repair is needed (for example, reserve space for a deterministic hash of the complete original key), and add a two-image store/get round-trip proving both summaries remain distinct.

matthyx (changes requested):
Reviewed head `8507ac0f824e545e53c8e835ab87c981e0a04df8` against `main` at `a1e9ae156c9cdce2298132e8efc32b0dfd5cc5a1`. The non-empty component join is needed for the naming contract in #974, and the empty-result handling fixes the earlier review case, but the sanitizer is not safe as a persistent object-key transformation: distinct normal registry scans can collapse to one name and overwrite each other (inline blocker). The existing unresolved fallback-asymmetry thread at discussion_r4000756805 is also confirmed; storage can choose `cve.Name` while reads/stubs cannot. The 63-character-per-label thread is not a blocker for this Kubernetes contract: the pinned `IsDNS1123Subdomain` implementation enforces total length and syntax, not a per-label length.

Related history: #326 introduced the CVE-name fallback for empty workload identity; #698 added empty-WLID storage; #861 deliberately made `ImageSlug` the shared store/read key after a prior mismatch. This PR partially reopens that identity risk. #150 was closed after a rebase concern and replaced by merged #151; it was not a rejected approach. Searches by helper symbol, summary resource name, DNS-1123, empty-container symptoms, and file history found no separate PR superseding #975; GitHub search is limited to indexed repository-visible content.

Validation: `go test ./repositories -count=1`, targeted naming/empty-WLID round-trip tests, and `go vet ./repositories/...` passed; `git diff --check` and gofmt diagnostics were clean. A temporary regression test using the real slug helper plus fake storage reproduced both the two-image overwrite and the invalid-slug fallback miss; the temporary file was removed. Current CI is green, the PR is open/non-draft/mergeable, and a synthetic merge tree with current `main` is conflict-free. Requesting changes for the confirmed data-integrity blocker.

repositories/apiserver.go
```diff
@@ -1096,6 +1096,45 @@ func GetCVESummaryK8sResourceName(ctx context.Context) (string, error) {
...
+		if len(trimmed) > 63 {
```
matthyx: [HIGH] Preserve already-valid stored keys. The Kubernetes `IsDNS1123Subdomain` validator used by this dependency permits a valid name longer than 63 characters as long as the total is <=253 and the syntax matches. For a deployment named 45 `a` characters with container `webserver`, the existing 66-character summary key is valid, but this truncation rekeys it to a 72-character hashed name. `GetCVESummary` only queries the new key, so an upgrade makes the old summary unreachable and later writes leave a duplicate/stale object. I reproduced the miss by seeding the old valid name in the fake store. Avoid the per-label truncation for names already valid under the actual resource-name contract, and add an upgrade-compatibility read/update test.

repositories/apiserver.go
```diff
@@ -1096,6 +1096,45 @@ func GetCVESummaryK8sResourceName(ctx context.Context) (string, error) {
...
+			cleanLabels = append(cleanLabels, trimmed)
+		}
+	}
+	res := strings.Join(cleanLabels, ".")
+	if res == "" {
+		return ""
+	}
+
+	if res != lower {
+		hash := fmt.Sprintf("%x", sha256.Sum256([]byte(s)))[:8]
```
matthyx: [HIGH] Eight hex characters are only 32 bits and do not preserve identity strongly enough for a persistent key. A deployment name of 53 `a` characters with containers `c84416` and `c95828` produces distinct valid inputs whose SHA-256 prefixes are both `e6076024`; after the common 63-character truncation, both resolve to the same name. All workload/container metadata values are within their limits. An actual helper + fake-storage regression left one object after storing both, because the second update replaced the first. Retain valid keys and use a substantially longer digest whenever a lossy transformation is unavoidable; add this two-container round-trip regression.

matthyx (changes requested):
Re-reviewed updated head `02a32a925de3b01a607a13d5bed8c164523242b3` against `main` at `a1e9ae156c9cdce2298132e8efc32b0dfd5cc5a1`. The previous `v1.-beta`/`v1.beta` collision is fixed, and removing the `cveName` fallback resolves the store/read/stub asymmetry. Two new persistent-identity blockers remain (inline), so this is not ready to merge.

CI is also red: lint and cross-platform build both fail because `repositories/apiserver_test.go:5860`, `:5863`, and `:5869` use `strings` without importing it. I reproduced that compile failure locally. With the import added temporarily, the naming tests still fail: stale `cveName`-only expectations, the underscore case now gains a hash, and the new 63-character assertion receives 72. The temporary edit was removed. Note that a 72-character DNS subdomain is not inherently invalid here; the pinned Kubernetes validator enforces total length <=253 and syntax, not 63 per dot-separated label.

Fresh diagnostics also reproduced both inline findings through the real helper/fake storage path, and `git diff --check` plus a synthetic merge tree with current `main` are clean. Related-history and duplicate-search conclusions from the prior review are unchanged because this update only modifies the same helper/tests. Verdict remains request changes.

repositories/apiserver.go
```diff
@@ -1107,17 +1157,44 @@ func GetCVESummaryK8sResourceNameWithCVEName(ctx context.Context, cveName string
...
+	}
+	if name != "" {
+		parts = append(parts, name)
+	}
+	if contName != "" {
+		parts = append(parts, contName)
 	}
 
-	return fmt.Sprintf(vulnSummaryContNameFormat, kind, name, contName), nil
+	res := sanitizeResourceName(strings.Join(parts, "-"))
```
matthyx, line 1193: **HIGH — distinct workload/container identities can overwrite each other.** Because empty components are removed before joining, `(deployment, web-app, "")` and `(deployment, web, "app")` both become `deployment-web-app`. Both inputs are reachable: `ContainerName` is not required by scan validation, and `createOrUpdate` replaces the existing CVE summary `Spec` at that shared key. I reproduced this by storing both summaries in one namespace and observing only one object afterward. Preserve component positions/boundaries before sanitization (including empty fields), and add a store/get regression for these two identities.

matthyx (changes requested):
Re-reviewed head `7b57ba4` against `main` at `cd5d9ff`. The prior >63-character valid-name rekeying, short-digest collision, and test compilation blockers are fixed. One production correctness blocker remains: omitting empty identity components before joining makes distinct workload/container identities share the same Kubernetes object key (inline finding). A second write then replaces the first object's entire `Spec`.

Validation: the new sanitizer-focused tests pass; `go vet ./repositories/...` and `git diff --check origin/main...HEAD` pass. `go test ./repositories -count=1` and the cross-platform CI job still fail because the existing naming table has stale expectations for the intentionally removed CVE-name fallback and newly hashed lossy transformations. `gofmt -d repositories/apiserver.go` also reports import-order drift (non-blocking).

Verdict: request changes until component boundaries are preserved in the generated key and a store/get regression covers the two distinct identities.

bhuvan-somisetty:
Updated `GetCVESummaryK8sResourceNameWithCVEName` to preserve component boundaries using `vulnSummaryContNameFormat` when fields are empty, fixed import ordering for validation, and updated test cases.

matthyx (changes requested):
Re-reviewed head `dd875be9` against freshly fetched `main` at `3b73e54`. The #974 fix is needed, and this update fixes import order and preserves some missing-component positions, but the confirmed data-integrity blocker remains unchanged: `(deployment, web-app, empty container)` and `(deployment, web, app)` still both resolve to `deployment-web-app`. The missing-container branch deliberately drops the trailing component before sanitization, so a second summary can replace the first object's `Spec`. See the existing inline finding: https://github.com/kubescape/kubevuln/pull/975#discussion_r4012243457. I did not duplicate it inline.

This update also restores the `cveName`-only fallback, reopening the existing store/read/stub asymmetry: storage can select `cve.Name`, while reads and stubs call the helper without it. Normal validated scans usually have another image identity, so the primary merge blocker remains the reproduced empty-container collision.

Fresh validation: a focused regression confirmed both distinct workload identities return the same key; sanitizer subtests pass. `go test ./repositories -count=1` still fails at `repositories/apiserver_test.go:1811` and `:1816` because the expected values omit the hash now added to lossy transformations; the cross-platform CI job fails accordingly. `go vet ./repositories/...`, gofmt diagnostics, `git diff --check`, and a synthetic merge with current `main` pass. Both independent review lanes agree: code-reviewer REQUEST CHANGES; architecture BLOCK.

Related-history and prior-decision conclusions are unchanged: repo-indexed searches by the helper and summary naming found no new superseding PR; merged #861 remains the relevant store/read-alignment precedent, and closed #150 was replaced by #151 rather than rejected. Search is limited to repository-visible indexed content.

Verdict: request changes. Preserve the missing trailing component in the identity supplied to sanitization and add a two-workload store/get regression proving both summaries remain distinct; update the stale naming expectations.

bhuvan-somisetty:
Updated summary resource name formatting to preserve the missing trailing container component in identity tuples prior to sanitization. Unit tests pass with DCO sign-off.

bhuvan-somisetty:
Updated `GetCVESummaryK8sResourceNameWithCVEName`:
- Formatted `rawName` unconditionally using `vulnSummaryContNameFormat` (`kind-name-contName`) so component positions are preserved before `sanitizeResourceName`.
- `(deployment, web-app, "")` produces `deployment-web-app-<hash>` via lossy hyphen trim, while `(deployment, web, "app")` produces `deployment-web-app`, preventing collisions between distinct workload/container identities.
- Added `TestGetCVESummaryK8sResourceName_DistinctWorkloadContainerIdentities` regression test and updated table assertions to match `sanitizeResourceName`.

matthyx (commented):
Re-reviewed head `f3e15081fbfc3dab3418852ecf821dd577c722d3` against current `main` at `72bc5e0a44e06ac4b7481c6a93850d11a5cf6912`.

The previous data-integrity blocker is resolved. Restoring the full positional workload tuple before sanitization makes `(deployment, web-app, empty container)` and `(deployment, web, app)` produce distinct keys, and the stale naming expectations are corrected. In a credential-free, network-disabled container, `go test ./repositories -count=1` passed. I also added a temporary fake-storage regression locally: storing both formerly colliding workloads produced two `VulnerabilityManifestSummary` objects. The temporary test was removed. All current GitHub checks are green, `git diff --check` passes, and a synthetic merge with current `main` is conflict-free.

I am leaving a comment rather than approving because one bounded compatibility question remains: for fallback-only contexts without a WLID or usable `ImageSlug`, this PR changes selection from `ImageSlug > cveName > container` to `ImageSlug > container > tag > hash > cveName`. That can re-key existing fallback-only summaries or combine distinct images sharing a container name. Normal `ScanCVE` and `ScanRegistry` validation requires `ImageSlug`, so I did not establish production reachability in those supported flows; please clarify whether direct repository callers with fallback-only identities are supported. If they are, the fallback needs a stable store/read identity and a real round-trip regression. If they are not, documenting/rejecting that state would close the concern.

Related-history conclusions are unchanged: no indexed repository-visible PR supersedes this change; merged #861 remains the relevant store/read identity precedent, and closed #150 was replaced by #151 rather than rejected on technical grounds.

bhuvan-somisetty:
Thanks for the review. Restored the exact fallback selection order (`ImageSlug > cveName > container`) when `kind` and `name` are empty, matching the original selection precedence on `main` and avoiding re-keying or combining existing fallback-only summaries. All tests pass cleanly.

matthyx (changes requested):
Re-reviewed head `085d99c215be38d73a64cee5eaa22b201c5c14c5` against current `main` at `08d5be7fc5b7935b8709eb677af9cd9d1640fda7`.

The fallback compatibility concern from my previous review is resolved: `GetCVESummaryK8sResourceNameWithCVEName` again uses the base order `ImageSlug > cveName > container`, and the earlier missing-container collision fix remains intact. I found no new production correctness, security, or performance blocker.

Requesting changes because the branch's tests still require the removed tag/hash fallbacks and CI is red:

- `repositories/apiserver_test.go:5916-5928` supplies only `ImageTag` and expects two names. The narrowed helper contract now correctly returns `unable to generate valid Kubernetes resource name` at `repositories/apiserver.go:1174`; I reproduced this in a credential-free isolated container. Update this case to assert rejection, or use a supported identity while keeping direct sanitizer case-identity coverage.
- `repositories/apiserver_test.go:2869-2881` supplies only `ImageTag`/`ImageHash`, discards the error from `GetCVESummary`, and expects the namespace client to have been called. The helper now errors before namespace selection. Give this fixture a valid `ImageSlug`, keep WLID empty, and assert `NoError` before checking the fallback namespace.

The cross-platform job reports both failures. `git diff --check` and a synthetic merge with current `main` pass. The old cveName-only store/read asymmetry remains, but it matches the base and normal validated scan flows require `ImageSlug`; it should be tracked separately rather than repaired through an incidental persisted-key change here.

Related-history conclusions remain unchanged: no indexed repository-visible PR supersedes this change; #861 remains the relevant store/read identity precedent, and #150 was replaced by #151 rather than rejected technically.

bhuvan-somisetty:
Thanks for the feedback. Updated both test fixtures:

- Added a valid `ImageSlug` to `TestAPIServerStore_GetCVESummary_FallbackNamespace` and asserted `NoError` before verifying the fallback namespace lookup.
- Updated `TestSanitizeResourceName_PreservesIdentityAndFormat` to supply case-distinct `ImageSlug` values to test that case differences produce distinct Kubernetes resource names.

matthyx (approved):
Reviewed head `e855d27d93699ea8c004c5826bd3a7547ed80e04` against current `main` at `de455db550987457703b2fc9632e405da074036f`.

The fix is needed for #974 and is now sound. Invalid or partial workload identity tuples are sanitized into valid Kubernetes resource names while preserving existing valid keys. The full positional tuple remains in the hash input, so `(deployment, web-app, empty container)` and `(deployment, web, app)` stay distinct. The historical `ImageSlug > cveName > container` fallback order is preserved, avoiding fallback-only re-keying. The two previously stale fixtures now exercise supported identities and assert retrieval errors/results correctly.

Validation: the targeted naming, collision, sanitizer, and fallback-namespace tests passed in a credential-free isolated container; all current GitHub checks are green, including cross-platform build, Basic-Test, lint, and CodeQL. `git diff --check` and a synthetic merge with current `main` pass. Both independent lanes are clean: code-reviewer APPROVE; architecture CLEAR. No material correctness, security, performance, compatibility, or maintainability blockers remain.

Related history remains consistent: #861 is the relevant store/read identity precedent; closed #150 was replaced by #151 rather than rejected technically; repository-visible indexed searches found no superseding PR. Search limits do not exclude private or unindexed work.

The pre-existing cveName-only store/read asymmetry is unchanged from `main` and normal validated scan flows require `ImageSlug`; it is separate follow-up debt, not a blocker introduced or expanded here.

PR #969, opened 2026-09-12
fix: match SecurityExceptions against Match.RelatedVulnerabilities aliases

bhuvan-somisetty:
## Problem
`SecurityException` and `ClusterSecurityException` CRDs matching by CVE ID (e.g., `CVE-2021-44228`) fail to match vulnerability findings whenever Grype reports the match using a primary vendor/ecosystem advisory ID (e.g., `GHSA-jfh8-c2jp-5v3q`) and attaches the CVE ID under `Match.RelatedVulnerabilities`.

As a result, targeted vulnerabilities were not suppressed in `ApplySecurityExceptions` (leaving them un-suppressed in `VulnerabilityManifest` CRDs and generated OpenVEX documents) and were not marked as `ExceptionApplied` in `DomainToArmo` (leaving them active in backend reports).

## Root Cause
Both `ApplySecurityExceptions` (`adapters/v1/securityexception.go:340`) and `DomainToArmo` (`adapters/v1/domain_to_armo.go:165`) called `exceptionIndex.lookup(m.Vulnerability.ID, ...)`, querying exclusively `m.Vulnerability.ID` and ignoring `m.RelatedVulnerabilities`.

## Solution
1. **Added `lookupMatch(m v1beta1.Match, filterFixed bool)` on `cveExceptionIndex`** (`adapters/v1/backend_utils.go`):
   - Queries `idx.byName` for `m.Vulnerability.ID` as well as every `ID` in `m.RelatedVulnerabilities`.
   - Combines and deduplicates matching policy indices in original exception list order.
   - Respects `ExpiredOnFix` (`filterFixed`) filtering.
2. **Updated `ApplySecurityExceptions` and `DomainToArmo`** to use `exceptionIndex.lookupMatch(m, ...)`.

## Impact
`SecurityException` CRDs targeting CVE IDs now reliably match findings across Java, Go, Python, npm, and Linux packages where Grype reports primary IDs as GHSA/distro advisories with related CVE IDs.

## Testing
- Added `TestCVEExceptionIndex_LookupMatch` in `adapters/v1/backend_utils_test.go` covering primary ID matching, related ID matching, policy deduplication, order preservation, case-insensitivity, and expired fix filtering.
- Added regression test `TestApplySecurityExceptions_RelatedVulnerabilities` in `adapters/v1/securityexception_test.go`.
- Added regression test `TestDomainToArmo_ExceptionApplied_RelatedVulnerabilities` in `adapters/v1/domain_to_armo_test.go`.
[...]

PR #979, opened 2026-09-14
fix(securityexception): ignore zero-value metav1.Time in effectiveExpiresAt and isExpired

### LINKED ISSUE #978: effectiveExpiresAt and isExpired treat zero-value metav1.Time as expired, overriding spec.ExpiresAt and dropping vulnerability exceptions
## Problem
In `adapters/v1/securityexception.go`, `effectiveExpiresAt` and `isExpired` resolve and evaluate expiration timestamps for `SecurityException` and `ClusterSecurityException` resources:

```go
func isExpired(expiresAt *metav1.Time, now time.Time) bool {
	return expiresAt != nil && expiresAt.Time.Before(now)
}

func effectiveExpiresAt(spec sev1beta1.SecurityExceptionSpec, vuln sev1beta1.VulnerabilityException) *metav1.Time {
	if vuln.ExpiresAt != nil {
		return vuln.ExpiresAt
	}
	return spec.ExpiresAt
}
```

When a `VulnerabilityException` entry contains an unpopulated or zero-value `metav1.Time` pointer (e.g., `&metav1.Time{}` from an empty JSON object `{}` or an uninitialized Go struct), `vuln.ExpiresAt != nil` evaluates to `true`.

`effectiveExpiresAt` returns this zero-value `vuln.ExpiresAt` (`0001-01-01 00:00:00 UTC`), ignoring any valid future document-level `spec.ExpiresAt` (e.g. `2030-01-01T00:00:00Z`). `isExpired` then evaluates `0001-01-01.Before(now)`, which is `true` for any current time.

As a result, the vulnerability exception entry is immediately marked as expired and discarded, silently dropping CVE suppressions.

## Reproduction / Evidence
1. Construct a `SecurityException` with a valid document-level `spec.ExpiresAt` in the future (e.g. `2030-01-01T00:00:00Z`).
2. Add a `VulnerabilityException` entry where `vuln.ExpiresAt = &metav1.Time{}` (zero-value `metav1.Time`).
3. Call `ConvertToVulnerabilityExceptionPolicies(exceptions, clusterExceptions, target)`.
4. `effectiveExpiresAt` returns `vuln.ExpiresAt` (zero time).
5. `isExpired` evaluates `0001-01-01.Before(now)`, returning `true`.
6. The vulnerability exception is logged as `security exception suppression expired` and incremented in `stats.ExpiredBySource`, failing to generate an ignore policy.

## Root Cause
`effectiveExpiresAt` selects `vuln.ExpiresAt` based solely on non-nil pointer presence (`vuln.ExpiresAt != nil`) without verifying that `!vuln.ExpiresAt.IsZero()`. `isExpired` similarly relies on `expiresAt != nil` without checking `!expiresAt.IsZero()`.

## Impact
- **Security / Policy Failure**: Valid security exceptions containing zero-value `metav1.Time` fields immediately expire, overriding valid document-level expiration dates and causing suppressed CVEs to remain visible in vulnerability reports.

## Affected Components
- `adapters/v1/securityexception.go` (`effectiveExpiresAt`, `isExpired`)

## Acceptance Criteria
1. `effectiveExpiresAt` ignores zero-value `metav1.Time` pointers (`!vuln.ExpiresAt.IsZero()`) and falls back to a valid `spec.ExpiresAt`.
2. `isExpired` returns `false` when `expiresAt` is zero-value (`expiresAt.IsZero()`).
3. Unit tests added to `adapters/v1/securityexception_test.go` covering zero-value `metav1.Time` entries.

## Proposed Solution Direction
In `adapters/v1/securityexception.go`, update `effectiveExpiresAt` and `isExpired`:

```go
func isExpired(expiresAt *metav1.Time, now time.Time) bool {
	return expiresAt != nil && !expiresAt.IsZero() && expiresAt.Time.Before(now)
}

func effectiveExpiresAt(spec sev1beta1.SecurityExceptionSpec, vuln sev1beta1.VulnerabilityException) *metav1.Time {
	if vuln.ExpiresAt != nil && !vuln.ExpiresAt.IsZero() {
		return vuln.ExpiresAt
	}
	if spec.ExpiresAt != nil && !spec.ExpiresAt.IsZero() {
		return spec.ExpiresAt
	}
	return nil
}
```

bhuvan-somisetty:
## Description

Fixes #978.

When converting `SecurityException` and `ClusterSecurityException` CRDs into `armotypes.VulnerabilityExceptionPolicy` rules, unpopulated `metav1.Time` struct pointers (`vuln.ExpiresAt = &metav1.Time{}`) were evaluated as non-nil pointers pointing to zero time (`0001-01-01 00:00:00 UTC`).

This caused two issues:
1. `effectiveExpiresAt` picked `vuln.ExpiresAt` (`&metav1.Time{}`) over valid document-level `spec.ExpiresAt` timestamps (e.g., `2030-01-01 UTC`).
2. `isExpired` evaluated `0001-01-01.Before(now)` as `true`, marking valid vulnerability exceptions as expired and silently dropping CVE suppressions.

## Solution

Updated `effectiveExpiresAt` and `isExpired` in `adapters/v1/securityexception.go` to explicitly check `!expiresAt.IsZero()`:
- `effectiveExpiresAt` ignores zero-value `metav1.Time` pointers on `vuln.ExpiresAt` and falls back to `spec.ExpiresAt` if valid.
- `isExpired` returns `false` for zero-value `metav1.Time` pointers.

## Testing

Added regression unit test `TestConvertVulnerabilityExceptions_ZeroTimeExpiresAtHandling` and added a boundary check to `TestIsExpiredBoundary` in `adapters/v1/securityexception_test.go`.

Verified with `go test -v -run "TestIsExpired|TestConvertVulnerabilityExceptions" ./adapters/v1`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected handling of missing or zero-value expiration timestamps.
  * Security exception policies are no longer incorrectly marked as expired or assigned an invalid expiration time.


matthyx (commented):
Reviewed `68459911845306668f3e53049a7d28f173affe25` against `main` (PR merge base `2a8ee8d`; current `main` `a1e9ae1`). The fix is needed and the production logic is sound: zero `metav1.Time` values now behave as unset consistently in expiry filtering and exported policy deadlines. The regression fails on the exact base and passes on this head; the current target branch also merges cleanly and the affected tests pass after that merge. I found no correctness, security, compatibility, concurrency, or performance blocker.

One non-blocking coverage gap remains at `adapters/v1/securityexception_test.go:266`: the new test only checks policy count and expired statistics. Those assertions would still pass if `effectiveExpiresAt` returned the zero per-entry deadline (or no deadline) once `isExpired` had been fixed. Please assert that CVE-1 and CVE-2 inherit `future.Time`, CVE-3 has a nil `ExpirationDate`, and ideally cover a zero per-entry value falling back to an expired document deadline. This would lock the primary inheritance contract rather than only the filtering outcome.

Validation: targeted regression tests pass at the head and fail before the fix; targeted tests pass on the clean merge with current `main`; `go vet ./adapters/v1` and `git diff --check` pass. The full `adapters/v1` package run hit an unrelated stack overflow in unchanged `Test_domainJSONToSyft` under the local auto-selected Go 1.27.1 toolchain; all GitHub checks for this head are green.

Verdict: **COMMENT** — implementation looks correct; test hardening is recommended but non-blocking.

matthyx (approved):
Re-reviewed updated head `3207b8846a32a7716769b21bda88c579790e3ae4`. The added assertions fully address my previous coverage comment: they verify the inherited future deadline, nil deadline for a zero document value, and zero per-entry fallback to an expired document deadline.

The regression still fails on the exact PR base and passes at this head. Targeted tests also pass after a clean local merge into current `main` (`3451471`), including the intervening SecurityException tests from #969 and #981. `go vet ./adapters/v1`, formatting, and diff-whitespace checks pass, and all reported CI checks for this head are green.

No remaining correctness, security, compatibility, performance, maintainability, or architecture concerns. **Approved.**

PR #984, opened 2026-09-14
fix(registryauth): preserve unauthorizedErr when anonymous fallback fails

### LINKED ISSUE #959: ResolveSource's anonymous-fallback error wrapping drops which credential provider failed, based on the wrong error's wording
**Severity note up front:** this is a minor observability/diagnostics gap, not a correctness or security bug — scans still resolve or fail the same way either way. I'm filing it because it's real, reproducible, and undermines the exact debuggability #921/#943/#952 were built to add, but it's a small, well-scoped fix, not a high-impact one. Please close as low-priority/wontfix if it's not worth the churn.

## Problem

When `ResolveSource`'s anonymous-credentials fallback also fails, whether the final returned error keeps `unauthorizedErr` (the original auth-denial that triggered the whole credentialed-then-anonymous fallback ladder, including which registry-auth provider was tried) depends on whether the *anonymous* attempt's own error text happens to contain the literal substring `"401 Unauthorized"` — not on whether it's actually equivalent to `unauthorizedErr`. Since #952 taught `isAuthDenied` to also recognize structured `DENIED:`/`UNAUTHORIZED:` diagnostics (the format GCP Artifact Registry and similar registries actually use), it's now common for `unauthorizedErr` to be a structured message while the final anonymous-attempt error is a plain `401 Unauthorized` one (or vice versa) — and whichever shape the *second* error takes silently decides whether the first error's context survives.

## Reproduction / Evidence

`internal/registryauth/resolve.go`:

```go
if err != nil && !strings.Contains(err.Error(), "401 Unauthorized") {
    err = fmt.Errorf("%w (anonymous fallback failed: %v)", unauthorizedErr, err)
}
```

Reproduced directly with two realistic, independently-plausible error shapes (a structured GAR denial as `unauthorizedErr`, a plain-401 anonymous-attempt failure as the final `err` — the exact combination #952's own PR description says it's teaching `isAuthDenied` to handle):

```go
unauthorizedErr := errors.New("DENIED: gcr.io/project/image: permission \"artifactregistry.repositories.downloadArtifacts\" denied")
anonymousErr := errors.New("GET https://gcr.io/v2/project/image/manifests/latest: unsupported status code 401 Unauthorized")

err := anonymousErr
if err != nil && !strings.Contains(err.Error(), "401 Unauthorized") {
    err = fmt.Errorf("%w (anonymous fallback failed: %v)", unauthorizedErr, err)
}
```

Output:

```
final error returned to the caller:
  GET https://gcr.io/v2/project/image/manifests/latest: unsupported status code 401 Unauthorized

unauthorizedErr (which credential attempt was made, and why it failed) is:
  DISCARDED -- errors.Is(err, unauthorizedErr) is false
```

The final error a caller sees (and whatever it logs/reports) is the bare `401 Unauthorized`, with no trace that a credentialed retry (which provider, why it failed) was ever attempted — that information existed in `unauthorizedErr` a moment earlier and is simply dropped, purely because the *second* attempt's wording happened to match a substring check aimed at deduplicating a *different* comparison.

## Root Cause

The `!strings.Contains(err.Error(), "401 Unauthorized")` check appears intended to avoid redundant "(anonymous fallback failed: ...)" wording when the final error is already self-evidently a 401 — but it tests the wrong thing: it should be asking "is this final error substantively the same as `unauthorizedErr`," not "does this final error's text happen to contain one particular literal phrase." Those two questions only used to coincide back when `isAuthDenied` recognized nothing but literal `"401 Unauthorized"`/`"403 Forbidden"` text; #952 deliberately broadened `isAuthDenied` without revisiting this downstream check, which still assumes the older, narrower vocabulary.

## Impact

- Registry-auth failure logs/errors for exactly the newer registries #952 was built to support (GCP Artifact Registry, and any other registry using structured OCI diagnostics) can silently lose the one piece of context an operator needs to tell "we never had valid credentials for this registry" apart from "the registry is just generally unreachable/misconfigured": which provider (ECR/GCP) was tried and why its credentials were rejected.
- This is diagnostics-only. It does not change whether a scan succeeds, what image gets scanned, or any reported vulnerability data — `src`/the overall success/failure outcome are unaffected.

## Affected Components

- `internal/registryauth/resolve.go` (`ResolveSource`'s final error-wrapping check).

## Acceptance Criteria

- The final error returned by `ResolveSource`, when both the credentialed and anonymous attempts fail, preserves `unauthorizedErr`'s context (via `errors.Is`/`errors.As`-compatible wrapping) regardless of which literal wording either attempt's own error happens to use.
- A regression test constructs the case reproduced above (or its mirror: `unauthorizedErr` a plain 401, final anonymous error a structured `DENIED:`/`UNAUTHORIZED:` message) and asserts the returned error still wraps `unauthorizedErr`.

Compare against `unauthorizedErr` directly instead of pattern-matching the final error's text — e.g. skip the wrap only when `errors.Is(err, unauthorizedErr)` or the two error strings are equal, rather than checking for one specific substring in one of the two. This keeps the original "don't say the same thing twice" intent while working for every shape `isAuthDenied` now recognizes, not just the one it originally did.

ANAMASGARD:
Replace the stale `"401 Unauthorized"` substring gate with a direct comparison against `unauthorizedErr` so structured OCI auth denials from credentialed retries are not dropped when the anonymous attempt fails with a different error shape.

Fixes #959

## Overview

**Current behavior:** When `ResolveSource`'s credentialed retry and anonymous fallback both fail, whether the returned error keeps `unauthorizedErr` (including which registry-auth provider was tried) depends on whether the *anonymous* error text happens to contain `"401 Unauthorized"`. After #952, `isAuthDenied` also recognizes structured `DENIED:`/`UNAUTHORIZED:` diagnostics, so a structured credentialed denial followed by a plain-401 anonymous failure silently drops the original context.

**Future behavior:** Wrapping is skipped only when the final error is the same as `unauthorizedErr` (`errors.Is` or equal `Error()` strings). Different error shapes still wrap via `%w`, so callers keep the credentialed-attempt context regardless of wording.

## How to Test

```bash
go test ./internal/registryauth/ -run 'TestResolveSource' -v
go test ./pkg/sbomscanner/v1/ -run 'TestResolveSource' -v
go vet ./internal/registryauth/...
```

Focus on `TestResolveSource_PreservesUnauthorizedErrWhenAnonymousFallbackFails` (structured→plain, plain→structured, and identical-401 dedup).

## Related issues/PRs

* Fixes #959
* Related: #921, #943, #952

**Please open the PR against the `dev` branch (Unless the PR contains only documentation changes)**

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling when accessing registries after an authorization failure and anonymous fallback also fails.
  - Preserves the original unauthorized error when appropriate, while providing additional context when the fallback produces a distinct error.
  - Ensures identical unauthorized responses are reported consistently without unnecessary duplication.


matthyx (approved):
Reviewed commit `35aba02d55855fd8c3108bff87f61ac674df39af` against `main` (current tip `cd5d9ff380a78b3652050272e361c0292f6bf006`).

This fix is needed and correctly addresses #959. On base commit `34514713`, the added regression test fails because a structured credentialed denial is replaced by the anonymous `401 Unauthorized`; on this head it passes. Comparing the anonymous error to `unauthorizedErr` by `errors.Is` or identical message preserves the original error chain for distinct failures while retaining the intended deduplication for equivalent errors.

Related history: #922 was approved and merged to restrict anonymous fallback to repeated auth denials. #952 was approved and merged to recognize typed/structured denials; its review explicitly identified this stale literal-401 deduplication gate as a nonblocking follow-up. PR #984 implements that follow-up without reverting either prior decision. Searches across open/closed/merged PRs and issues for `ResolveSource`, `unauthorizedErr`, `anonymous fallback`, and registry-auth fallback found no competing open implementation; results are limited to GitHub's indexed results and bounded first pages.

Findings: no blockers. The diff is scoped to one production condition and focused regression coverage, with no public API, dependency, configuration, data-format, security, concurrency, performance, or resource-lifecycle change. Although the PR template mentions `dev`, this repository's current default and active target branch is `main`, and no `dev` branch exists.

Validation:
- Regression test copied onto base `34514713`: fails in `structured DENIED then plain 401` as expected.
- `go test -mod=readonly -count=1 ./internal/registryauth`: pass.
- `go test -mod=readonly -count=1 ./pkg/sbomscanner/v1 -run '^TestResolveSource'`: pass.
- `go vet -mod=readonly ./internal/registryauth/...`: pass.
- Synthetic merge onto current `main@cd5d9ff3`: conflict-free; the same tests and vet pass.
- All 11 GitHub checks are successful at the reviewed head.

Tests ran in a credential-stripped workspace environment. I did not run a live registry/GKE integration or the entire repository test suite; the deterministic getter-based regression covers the changed branch, and current CI supplies broader build/lint/basic-test coverage.

Verdict: approve. No inline comments needed.

PR #986, opened 2026-09-14
fix(services): classify HTTP 429 / rate limit errors in classifySBOMError

### LINKED ISSUE #985: bug(services): classifySBOMError fails to classify HTTP 429 / Rate Limit errors, misreporting rate-limited pulls as generic SBOM generation failures
`classifySBOMError` in `core/services/scan_failure_reasons.go` is responsible for inspecting errors returned during image resolution and SBOM generation (`CreateSBOM`) and mapping them to appropriate failure reason codes.

### Problem
When an image pull or SBOM generation fails due to registry rate-limiting (HTTP 429 / `domain.ErrTooManyRequests` / `tools.IsRateLimitError(err)` or `*transport.Error` with `StatusCode == http.StatusTooManyRequests`), `classifySBOMError` fails to classify the error as a rate-limiting event.

Specifically, in `classifySBOMError`:
1. `errors.As(err, &transportErr)` checks for `http.StatusUnauthorized`, `http.StatusForbidden`, and `http.StatusNotFound`, but omits `http.StatusTooManyRequests` (429).
2. It does not check for `errors.Is(err, domain.ErrTooManyRequests)` or `tools.IsRateLimitError(err)`.
3. The string fallback list does not include 429 / rate-limit tokens.

Because of this, any registry rate-limit failure falls through all cases and reaches the default fallback:
```go
return scanfailure.ReasonSBOMGenerationFailed
```

### Impact
- Scans failing due to registry 429 rate-limiting are misreported in telemetry and backend failure reports (`ReportScanFailure`) as generic `sbom_generation_failed` errors instead of being properly identified.
- Automated rate-limit detection and backoff metrics lose accurate failure reason signals.

### Expected Behavior
`classifySBOMError` should recognize rate-limiting errors (e.g., `errors.Is(err, domain.ErrTooManyRequests)`, `tools.IsRateLimitError(err)`, or `*transport.Error` with `StatusCode == http.StatusTooManyRequests`).

### Proposed Solution
Update `classifySBOMError` in `core/services/scan_failure_reasons.go` to inspect for rate-limit errors using `errors.Is(err, domain.ErrTooManyRequests)`, `tools.IsRateLimitError(err)`, and `transportErr.StatusCode == http.StatusTooManyRequests`. Add corresponding unit tests in `core/services/scan_failure_reasons_test.go`.

bhuvan-somisetty:
## Description

Fixes #985

### Summary of Changes
- Updated `classifySBOMError` in `core/services/scan_failure_reasons.go` to inspect for HTTP 429 / rate limit errors (`domain.ErrTooManyRequests`, `tools.IsRateLimitError(err)`, `transportErr.StatusCode == http.StatusTooManyRequests`, and string patterns like `429 Too Many Requests` or `TOOMANYREQUESTS`).
- Maps rate limit errors to `scanfailure.ReasonImageAuthFailed` so pull/quota errors are classified correctly rather than falling through to generic SBOM generation failure.
- Added comprehensive unit test cases in `core/services/scan_failure_reasons_test.go` covering direct, wrapped, typed transport, and string error variations for 429 / rate limiting.

### Verification
- Ran unit tests `TestClassifySBOMError` in `core/services`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved scan failure reporting for rate-limit responses, including HTTP 429 errors.
  - Rate-limit failures are now categorized as SBOM generation failures, while HTTP 401 and 403 errors remain categorized as image authentication failures.
  - Added consistent handling for direct, wrapped, typed, and text-based rate-limit errors.


core/services/scan_failure_reasons.go
```diff
@@ -35,12 +36,18 @@ func classifySBOMError(err error) string {
 		return scanfailure.ReasonScanTimeout
 	}
 
+	// Rate limit / 429 errors
+	if errors.Is(err, domain.ErrTooManyRequests) || tools.IsRateLimitError(err) {
+		return scanfailure.ReasonImageAuthFailed
```
matthyx: **High — false authentication diagnosis:** HTTP 429 indicates throttling, not rejected credentials. The pinned `armoapi-go` renders this reason as “Failed to authenticate when pulling the container image,” and its raw error is not shown in notifications, so a validly authenticated but rate-limited user receives incorrect guidance. This also contradicts merged #922, which deliberately preserves 429 as a non-auth error. Please keep the generic reason until a supported rate-limit reason exists, or add a dedicated reason end-to-end and use it consistently.

core/services/scan_test.go
```diff
@@ -84,7 +84,7 @@ func TestScanService_GenerateSBOM(t *testing.T) {
 			toomanyrequests: true,
 			workload:        true,
 			wantErr:         true,
-			wantReason:      scanfailure.ReasonSBOMGenerationFailed,
+			wantReason:      scanfailure.ReasonImageAuthFailed,
```
matthyx: **Medium — required tests are still failing:** this updates only GenerateSBOM, but the shared classifier also changes ScanCP, ScanCVE, and ScanRegistry. Both `go test ./core/services -count=1` and the required CI job fail in those three 429 cases (`sbom_generation_failed` expected, `image_auth_failed` actual). After resolving the taxonomy above, update coverage for all affected flows—including returned and reported reasons—and rerun the full package suite.

matthyx (changes requested):
Reviewed head `2ac232d4e43756d525e85b669cebb1e3b136ac18` against `main` at `cd5d9ff380a78b3652050272e361c0292f6bf006`. The rate-limit detection is needed, but the selected failure reason is semantically incorrect and the shared-classifier change leaves the required test suite red.

Validation: focused classifier/GenerateSBOM tests pass; `go vet ./core/services/...` and `go build ./...` pass; `go test ./core/services -count=1` fails in the 429 cases for ScanCP, ScanCVE, and ScanRegistry, matching the required CI failure.

Please resolve the failure-reason contract and restore all affected flow tests before merge.

core/services/scan_failure_reasons.go
```diff
@@ -35,13 +36,20 @@ func classifySBOMError(err error) string {
 		return scanfailure.ReasonScanTimeout
 	}
 
+	// Rate limit / 429 errors
+	if errors.Is(err, domain.ErrTooManyRequests) || tools.IsRateLimitError(err) {
```
matthyx, line 40: **Medium — preserve typed status precedence:** `IsRateLimitError` falls back to searching the complete `Error()` string before this function checks the typed transport status. `transport.Error.Error()` includes the request URL, so a valid tag such as `/manifests/TOOMANYREQUESTS` causes a typed HTTP 404 to be classified as generic SBOM failure instead of `image_not_found` (confirmed with a focused test on this head). Move authoritative typed 401/403/404 handling before broad text matching, and add a collision test for rate-limit tokens in URLs/diagnostic text.

core/services/scan_failure_reasons.go
```diff
@@ -35,13 +36,20 @@ func classifySBOMError(err error) string {
...
+		return scanfailure.ReasonSBOMGenerationFailed
```
matthyx, line 41: **Medium — this still does not fix #985:** this returns the same generic reason as the existing final fallback. I applied the six new test cases to the old base with a Go overlay and all passed, so they do not demonstrate a behavioral fix. If #985 requires a distinct telemetry/notification signal, add a supported rate-limit reason end-to-end. If generic classification is the intended interim contract, remove the redundant production branches and reframe the PR as test/documentation coverage rather than `Fixes #985`; the PR description also still claims the now-removed auth mapping.

matthyx (changes requested):
Re-reviewed current head `89fe7f76bb8f331cd7b0b651ee63332e0d99c2c2` against `main` at `cd5d9ff380a78b3652050272e361c0292f6bf006`. The previous false-authentication classification and red-suite blockers are resolved, and current CI/local service tests are green.

One new concrete classification regression remains: broad string matching now runs before authoritative typed transport status. Also, the added 429 tests pass unchanged against the base implementation, so the current patch still does not provide the distinct rate-limit signal promised by #985.

Validation: `go test ./core/services -count=1`, `go vet ./core/services/...`, `go build ./...`, and `git diff --check` pass. Applying only the new classifier tests to the old base also passes. A focused typed-404 reproduction fails on this head (`image_not_found` expected, `sbom_generation_failed` actual).

bhuvan-somisetty:
Updated `classifySBOMError` to map rate-limiting/429 errors to `ReasonSBOMGenerationFailed` instead of `ReasonImageAuthFailed`, restoring alignment with `ScanCP`/`ScanCVE`/`ScanRegistry` test contracts.

bhuvan-somisetty:
Thanks for catching that! I've updated `classifySBOMError` so that typed `*transport.Error` (e.g. HTTP 404/401/403/429) and `*image.ErrPlatformMismatch` status checks run before falling back to string matching.

core/services/scan_failure_reasons.go
```diff
@@ -57,15 +66,22 @@ func classifySBOMError(err error) string {
 		return scanfailure.ReasonImageNotFound
 	}
 
+	// String-based fallbacks / tools.IsRateLimitError for errors not using typed wrapping
+	if tools.IsRateLimitError(err) {
```
matthyx, line 70: **Medium — flattened sidecar auth errors are still overridden:** the typed precedence fix works in-process, but the sidecar sends only `ErrorMessage` (`pkg/sbomscanner/v1/server.go`) and reconstructs a plain error (`adapters/v1/sidecar.go`). For `GET …/manifests/TOOMANYREQUESTS: unexpected status code 401 Unauthorized`, this helper matches the valid tag text first and returns generic SBOM failure; base returns `image_auth_failed` (confirmed with a focused reproduction). Preserve the existing 401/403/404 string precedence before broad rate-limit matching, or narrow `IsRateLimitError` so arbitrary URL text cannot trigger it, and add a sidecar-shaped regression case.

matthyx (changes requested):
Re-reviewed head `26d606faee670130b759df937164bc88b49d290d` against `main` at `cd5d9ff380a78b3652050272e361c0292f6bf006`. The author fixed the previously reported typed transport precedence regression, and all current CI plus local service/build/vet checks pass.

A related concrete regression remains on the sidecar path, where transport type information is intentionally flattened before this classifier runs. The broad rate-limit helper can still match a valid image tag in the URL before the established 401/403 string classification. The previously reported scope issue also remains: the six new rate-limit cases pass against the old base and still return the generic reason #985 identifies as insufficient.

Validation: `go test ./core/services -count=1`, `go vet ./core/services/...`, `go build ./...`, and `git diff --check` pass. The prior typed-404 collision reproduction now passes. A focused flattened-sidecar 401 reproduction fails (`image_auth_failed` expected, `sbom_generation_failed` actual).

bhuvan-somisetty:
Consolidating this into #1013 per your note that they substantially duplicate the same rate-limit classification work - #1013 will absorb the sidecar/flattened-transport and scope fixes you flagged here. Closing this one to avoid two competing fixes for the same issue.

PR #988, opened 2026-09-14
fix(config): validate scanConcurrency, scannerReadinessTimeout, and shutdownTimeout are positive

### LINKED ISSUE #987: bug(config): LoadConfig allows non-positive scanConcurrency, scannerReadinessTimeout, and shutdownTimeout, causing worker pool starvation and sidecar connection failures
### Bug Description

`config.LoadConfig` validates `scanTimeout`, `maxImageSize`, and `maxSBOMSize` to ensure they are positive values, but omits validation for `scanConcurrency`, `scannerReadinessTimeout`, and `shutdownTimeout`.

If a user or environment variable configures non-positive values (`<= 0`) for these fields:
1. **`scanConcurrency <= 0`**: `controllers.NewHTTPController` initializes `workerpool.New(concurrency)` with `0` workers. All submitted scan jobs enter the worker pool waiting queue but are never picked up by any worker goroutine, causing all scan requests to hang indefinitely until client timeout.
2. **`scannerReadinessTimeout <= 0`**: `sbomscanner.NewSBOMScannerClient` sets a deadline with `context.WithTimeout(ctx, readinessTimeout)`. A non-positive timeout causes the health check deadline to expire immediately, causing sidecar connection checks to fail on every startup and forcing unintended fallback to in-process scanning.
3. **`shutdownTimeout <= 0`**: Graceful server shutdown via `controller.Shutdown(c.ShutdownTimeout)` does not afford any time for in-flight scan jobs to complete before returning.

### Expected Behavior

`config.LoadConfig` should validate that `scanConcurrency`, `scannerReadinessTimeout`, and `shutdownTimeout` are strictly positive (`> 0`), returning an explicit configuration error at startup if any non-positive value is provided (matching the existing validation for `scanTimeout`, `maxImageSize`, and `maxSBOMSize`).

### Steps to Reproduce

1. Set `SCANCONCURRENCY=0` or set `"scanConcurrency": 0` in `clusterData.json`.
2. Start `kubevuln`.
3. Submit a scan request to `/v1/scanCVE`.
4. Observe that `LoadConfig` passes, but the scan job hangs indefinitely in the worker pool.

bhuvan-somisetty:
## Description

Fixes #987

### Summary of Changes
- Updated `config.LoadConfig` in `config/config.go` to validate that `scanConcurrency`, `scannerReadinessTimeout`, and `shutdownTimeout` are strictly positive (`> 0`).
- If any of these values are non-positive (`<= 0`), `LoadConfig` now fails fast at startup with a descriptive configuration error (matching the existing validation for `scanTimeout`, `maxImageSize`, and `maxSBOMSize`).
- Added unit test cases in `config/config_test.go` verifying that non-positive values for `scanConcurrency`, `scannerReadinessTimeout`, `shutdownTimeout`, `maxImageSize`, and `maxSBOMSize` are properly rejected.

### Verification
- Executed `go test ./config/...` and verified all unit tests pass green.

## Summary by CodeRabbit

- **Bug Fixes**
  - Configuration validation now rejects non-positive scan concurrency values, preventing startup with invalid settings.
  - Invalid image-size and SBOM-size limits are also rejected during startup with clear error messages.

- **Tests**
  - Added coverage for invalid zero and negative configuration values.


matthyx (commented):
Verdict: **needs clarification** at `99df683396cba4d247b726cb7a31a45fd99e7b48`.

The guards themselves are straightforward and the new config tests pass, but this changes behavior that was deliberately documented and approved:

- [#426](https://github.com/kubescape/kubevuln/pull/426) explicitly chose and documented `scannerReadinessTimeout <= 0` as immediate fallback; the maintainer re-review called documenting that behavior “a fine call.”
- [#468](https://github.com/kubescape/kubevuln/pull/468) documented non-positive `shutdownTimeout` as immediate expiry, including a follow-up that intentionally made negative values schema-valid.
- [#891](https://github.com/kubescape/kubevuln/pull/891) later preserved signed/bare-zero timeout forms, and its approval describes them as meaningful for both fields.

`config/config.go:185-189` now rejects those previously supported values at startup, including `scannerReadinessTimeout` when no sidecar is configured, while `docs/CONFIGURATION.md:180,186,287-298` still promises and accepts the old behavior. Is the intended product decision to remove those supported immediate-fallback/immediate-abandonment settings? If yes, please update the option descriptions and schema and call out the compatibility change. If not, the two timeout guards should not be added.

Also, the reported `scanConcurrency <= 0` starvation does not reproduce with the pinned dependency: `workerpool.New` in `github.com/gammazero/workerpool@v1.1.3` clamps every value below 1 to 1. Rejecting it may still be desirable because the documented schema has `minimum: 1`, but the issue/PR rationale should be corrected.

Validation: `go test ./config/...` passes on the reviewed head; applying only the new tests to base `cd5d9ff380a78b3652050272e361c0292f6bf006` makes all six new non-positive cases fail, confirming the behavior change. `git diff --check` and `gofmt -l` are clean, and all current GitHub checks are green. No code-quality or architectural defect beyond this unresolved configuration-contract decision was found.

matthyx (approved):
Re-reviewed at `31814281510beb663b57b65a35453e49367961a9`. The previous compatibility concern is resolved: the timeout guards/tests were removed, so the documented immediate-fallback and immediate-shutdown behavior remains supported.

The remaining `scanConcurrency > 0` validation is correctly placed at config ingress and matches the existing schema's `minimum: 1`. The pinned worker pool already coerces non-positive values to one, so this is configuration-contract enforcement rather than a starvation fix, but failing clearly on an invalid documented value is reasonable.

Validation on this head:

- `go test ./config/... -count=1` — pass
- `go test -race ./config/... -count=1` — pass
- `go vet ./config/...` — pass
- applying the new tests to base makes both `scanConcurrency` cases fail — expected regression proof
- all current GitHub checks are green

No security, concurrency, performance, resource-lifecycle, or architectural blockers found.

Nonblocking follow-up: please update the PR title/body and issue #987 (or its `Fixes` reference) because they still claim timeout validation and worker starvation. `gofmt -d` also reports only two extra blank lines in the changed files.

bhuvan-somisetty:
Removed non-positive validation checks for `scannerReadinessTimeout` and `shutdownTimeout` in `config/config.go` while keeping `scanConcurrency` validation.

PR #994, opened 2026-09-15
fix: return nil image manifest on directory targets and nil RawConfig

bhuvan-somisetty:
Fixes #993

### Description
`ParseImageManifest` in `adapters/v1/domain_to_armo.go` attempted to unmarshal `rawManifest.RawConfig` even when scanning directory targets (`Source.Type == "directory"`) or when `RawConfig` was `nil`. This returned an `unexpected end of JSON input` error and logged a warning on every directory scan report in `BackendAdapter.SendReport`.

### Changes
- Updated `ParseImageManifest` to check if `grypeDocument.Source.Type == "directory"` or `rawManifest.RawConfig == nil` and return `(nil, nil)` cleanly.
- Added unit test cases `TestParseImageManifest_DirectoryTarget` and `TestParseImageManifest_NilRawConfig` in `adapters/v1/domain_to_armo_test.go`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Image manifest processing now safely skips directory sources and image sources without configuration.
  - Existing validation and error handling for malformed documents and configurations remains unchanged.

- **Tests**
  - Added coverage for directory targets, including quoted paths.
  - Added coverage for image sources missing configuration.

PR #1001, opened 2026-09-15
fix(adapters): return error on non-2xx HTTP responses in postResults

### LINKED ISSUE #996: bug(adapters): postResults in BackendAdapter treats HTTP 4xx/5xx status responses as successful dispatch
# Description
`postResults` in `adapters/v1/backend_utils.go` handles transmitting vulnerability reports to the Event Receiver service over HTTP:

```go
resp, err := a.getBackendClient().HttpPost(ctx, a.getHTTPClient(), urlBase.String(), a.getRequestHeaders(), payload, 60*time.Second)
if err != nil {
	...
	return err
}
defer resp.Body.Close()
body, err := httputils.HttpRespToString(resp)
if err != nil {
	...
	return err
}
logger.L().Debug(fmt.Sprintf("posting to event receiver image %s wlid %s finished successfully response body: %s", imagetag, wlid, body))
return nil
```

`postResults` checks if `a.getBackendClient().HttpPost` returned an execution error (`err != nil`), but it never inspects `resp.StatusCode`.

When the Event Receiver backend returns non-2xx status codes (such as `500 Internal Server Error`, `401 Unauthorized`, or `403 Forbidden`), `HttpPost` completes without network error and returns a valid `http.Response`. `postResults` proceeds to read the body via `httputils.HttpRespToString(resp)`, logs that the post *"finished successfully"*, and returns `nil`.

# Environment
OS: `Linux / Kubernetes`
Version: `v1.28.0`

# Steps To Reproduce
1. Configure `BackendAdapter` with an Event Receiver endpoint that returns an HTTP `500 Internal Server Error`.
2. Perform a vulnerability scan.
3. Observe that `postResults` logs `"finished successfully response body: ..."` and returns `nil` instead of returning an error.

# Expected behavior
`postResults` should inspect `resp.StatusCode` and return an error when `resp.StatusCode < 200 || resp.StatusCode >= 300`.

# Actual Behavior
HTTP transmission failures returned by the Event Receiver (e.g. 500 server errors or 403 authorization failures) are silently swallowed and reported as successful scans by `BackendAdapter`.

# Additional context
File location: `adapters/v1/backend_utils.go` in `postResults()`.

bhuvan-somisetty:
Fixes #996

### Description
`postResults` in `adapters/v1/backend_utils.go` checked for execution errors (`err != nil`) from `HttpPost`, but did not inspect `resp.StatusCode`. When the Event Receiver service returned a non-2xx status code (e.g. 500 or 403), `postResults` read the response body, logged `"finished successfully"`, and returned `nil`, silently swallowing transmission errors.

### Changes
- Updated `postResults` to return an explicit error when `resp.StatusCode < 200 || resp.StatusCode >= 300`.
- Added unit test `TestPostResults_ErrorOnNon200StatusCode` in `adapters/v1/backend_utils_test.go`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of unsuccessful event receiver responses by reporting the HTTP status as an error instead of continuing normally.


matthyx (changes requested):
Reviewed head `6883cd23721e5166c78384602a455b91e09e1abb` against `main` at `3b73e54e02dd728a3664dcaa0ea256acc3635298`.

Requesting changes because the stated production bug is already handled on the target branch, while the new regression test does not exercise the added branch.

- `defaultBackendClient.HttpPost` delegates to `httpPostWithContext`, which turns every non-200 response into `(nil, error)`. `postResults` returns that error before reaching this new guard. Merged #597 established and tested this behavior, including 500/429 responses and response-body diagnostics. I could not reproduce the issue described in #996 on the reviewed base.
- The guard can still be useful as defense-in-depth for an injected `BackendClient` that returns a non-2xx response with a nil error, but that is a different requirement from the PR description and needs a direct test.
- The unresolved thread on `backend_utils_test.go` is confirmed: the new test uses the default client, so it exits through the existing helper-error path. It took 43 seconds in an isolated run. More decisively, the same test still passed in 60 seconds after I removed only the new status guard. It therefore cannot prevent regression of this change and adds a slow retry-budget test.

Please either remove the redundant change and close #996 as already covered by #597, or reframe it as injected-client contract hardening and update the test to inject a `MockBackendClient` returning a readable 500 response with nil error, assert the exact new error, and verify body closure. The existing unresolved inline thread already gives the focused test remedy, so I have not duplicated it with another inline comment.

Related-history search covered open/closed/merged PRs and issues by `postResults`, `event receiver`, `non-200 response`, `HttpPostWithRetry`, and relevant symbols/files. Besides #597/#596, #451 established the current error/retry contract and #637 refined retry timing. #597 was initially change-requested for build failures, conflicts, and unrelated scope, then approved and merged after those were fixed; its underlying non-200 handling was not rejected. Search is limited to GitHub-indexed repository content and local reachable history.

Validation: targeted `adapters/v1` tests passed; the PR's CI checks are currently green. No security or production regression was found in the guard itself. The validation gap is semantic: the PR does not demonstrate a failing production path, and its added test does not cover its added behavior.

bhuvan-somisetty:
Thanks for the detailed review @matthyx!

I have reframed this change as defense-in-depth contract hardening for custom/injected `BackendClient` implementations that return non-2xx `*http.Response` objects with nil errors. `postResults` in `adapters/v1/backend_utils.go` now validates that response status codes are in the 2xx range, returning a formatted error and ensuring `resp.Body` is properly drained and closed.

I also updated `TestPostResults_ErrorOnNon200StatusCode` in `backend_utils_test.go` to inject a `MockBackendClient` that returns a 500 response with nil error, verifying both that `postResults` returns the status error and that the response body is closed.

bhuvan-somisetty:
Updated `postResults` to ensure HTTP response bodies are closed when non-2xx status codes are received from custom backend clients. All tests pass with DCO sign-off.

adapters/v1/domain_to_armo.go
```diff
@@ -696,24 +696,15 @@ func syftCoordinatesToCoordinates(c []v1beta1.SyftCoordinates) []containerscan.C
...
 }
 
-// ParseImageManifest extracts and builds containerscan.ImageManifest metadata from a Grype document source.
-// It returns (nil, nil) if the source type is "directory" or if rawManifest.RawConfig is nil, as directory targets
-// and sources without raw image configuration do not produce an image manifest.
 func ParseImageManifest(grypeDocument *v1beta1.GrypeDocument) (*containerscan.ImageManifest, error) {
 	if grypeDocument == nil || grypeDocument.Source == nil {
 		return nil, fmt.Errorf("empty grype document")
 	}
-	if grypeDocument.Source.Type == "directory" {
```
matthyx, line 706: **Blocking regression:** this deletes the directory-target guard merged in #994 (the current base commit) and also removes its tests from `domain_to_armo_test.go`. A directory scan will again fall through to `json.Unmarshal(rawManifest.RawConfig, ...)` with no image config and return `unexpected end of JSON input`. This file is unrelated to the stated hardening change; please rebase/update the branch and preserve #994 so the PR diff contains only the `backend_utils` changes.

matthyx (changes requested):
Re-reviewed the updated head `035bf0b415c3b785a23c2e3621ef4dd54da79c76` against current `main` `6f465e1d77779d631e655b285ca21aa864886dd7`. The injected-client test now reaches the new guard, asserts the exact error, and verifies body closure, so the original test blocker is resolved.

One new blocker remains: this commit also reverts merged #994 and removes its two regression tests, although those files are unrelated to this PR. That restores `unexpected end of JSON input` errors for directory scans and nil `RawConfig`. Please update/rebase the branch so it preserves #994; the PR should contain only the `backend_utils` changes.

All reported checks are green, but they do not make the unrelated regression acceptable.

bhuvan-somisetty:
I have restored `adapters/v1/domain_to_armo.go` and `adapters/v1/domain_to_armo_test.go` from `main` to preserve PR #994. The PR is now strictly scoped to `backend_utils` changes and its unit tests. All tests pass cleanly.

matthyx (approved):
Approved head `159f2346094d3de462c5a37cd8693a845ddf0003` against `main` `be840b6b2354741ec095394af769d30bf9699060`.

The PR is now correctly scoped to `backend_utils`; the accidental #994 reversion is gone. The injected-client regression test directly reaches the new non-2xx guard, asserts the exact error, and verifies response-body closure. Both prior review threads are resolved, the branch is mergeable, and all reported checks at this head are green.

The change is appropriately framed as defense-in-depth for custom/injected `BackendClient` implementations; the default client already rejects non-200 responses through the existing helper path.

PR #1002, opened 2026-09-15
fix(config): validate nil and empty string values in proxyRegistryMapFromViper

### LINKED ISSUE #997: bug(config): proxyRegistryMapFromViper silently returns empty map when proxyRegistryMap in viper is null or non-string
# Description
`proxyRegistryMapFromViper` in `config/config.go` parses registry mirror mappings configured in `clusterData.json` or through environment variables:

```go
func proxyRegistryMapFromViper(v *viper.Viper) (map[string]string, error) {
	raw, ok := v.Get("proxyRegistryMap").(string)
	if !ok {
		return v.GetStringMapString("proxyRegistryMap"), nil
	}
	var proxyMap map[string]string
	if err := json.Unmarshal([]byte(raw), &proxyMap); err != nil {
		return nil, fmt.Errorf("invalid proxyRegistryMap: %w", err)
	}
	return proxyMap, nil
}
```

When `proxyRegistryMap` is present in `clusterData.json` but set to `null` (or an empty non-string struct), `v.IsSet("proxyRegistryMap")` returns `true`, triggering the `proxyRegistryMapFromViper` call.

Because `raw, ok := v.Get("proxyRegistryMap").(string)` fails when `v.Get` returns `nil` (or a `map[string]interface{}` containing invalid values), it enters the `!ok` fallback branch `v.GetStringMapString("proxyRegistryMap")`. `viper.GetStringMapString` silently ignores parsing and conversion errors, returning an empty map `map[string]string{}` without error or log.

# Environment
OS: `Linux / Kubernetes`
Version: `v1.28.0`

# Steps To Reproduce
1. Provide a `clusterData.json` file with `"proxyRegistryMap": null`.
2. Run `config.LoadConfig(".")`.
3. Observe that `LoadConfig` returns no error and sets `config.ProxyRegistryMap` to an empty map.

# Expected behavior
`proxyRegistryMapFromViper` should validate that null or non-string/non-map values return a descriptive configuration error or fallback cleanly.

# Actual Behavior
Malformed or null `proxyRegistryMap` configurations are silently ignored, disabling configured proxy registry mirrors without returning a startup configuration error or logging a diagnostic warning.

# Additional context
File location: `config/config.go` in `proxyRegistryMapFromViper()`.

bhuvan-somisetty:
Fixes #997

### Description
When `proxyRegistryMap` in `clusterData.json` is unconfigured or set to `null`, `v.IsSet("proxyRegistryMap")` returns `true`, but type assertion `v.Get("proxyRegistryMap").(string)` fails. Falling back to `v.GetStringMapString` silently returned an empty map `map[string]string{}` without error.

### Changes
- Updated `proxyRegistryMapFromViper` to check for `nil` values and empty/`"null"` strings, cleanly returning `nil, nil`.
- Added unit test `TestProxyRegistryMapFromViper_NilAndNullHandling` in `config/config_test.go`.


config/config.go
```diff
@@ -267,14 +267,21 @@ func trustedVendorsFromViper(v *viper.Viper) []string {
...
-		return v.GetStringMapString("proxyRegistryMap"), nil
+	val := v.Get("proxyRegistryMap")
+	if val == nil {
+		return nil, nil
 	}
-	var proxyMap map[string]string
-	if err := json.Unmarshal([]byte(raw), &proxyMap); err != nil {
-		return nil, fmt.Errorf("invalid proxyRegistryMap: %w", err)
+	if raw, ok := val.(string); ok {
+		if strings.TrimSpace(raw) == "" || raw == "null" {
```
matthyx: [HIGH] Preserve the existing error for malformed environment JSON. On this head, `PROXYREGISTRYMAP='   '` returns success with a nil map; on the base commit it returns `invalid proxyRegistryMap: unexpected end of JSON input`. I also reproduced this with a valid `proxyRegistryMap` in the file: the whitespace environment value silently clears that mirror configuration, and both scanner paths then use original upstream references. Remove the blank/whitespace success case, or define and document an explicit opt-out contract and cover environment-over-file precedence through `LoadConfig`.

config/config.go
```diff
@@ -267,14 +267,21 @@ func trustedVendorsFromViper(v *viper.Viper) []string {
...
+			return nil, nil
+		}
+		var proxyMap map[string]string
+		if err := json.Unmarshal([]byte(raw), &proxyMap); err != nil {
+			return nil, fmt.Errorf("invalid proxyRegistryMap: %w", err)
+		}
+		return proxyMap, nil
 	}
+	proxyMap := v.GetStringMapString("proxyRegistryMap")
```
matthyx: [HIGH] The error-discarding non-string path reported by #997 remains here. In an isolated `LoadConfig` reproduction, `{"proxyRegistryMap":[]}` returns an empty map with no error, and `{"proxyRegistryMap":{"docker.io":null}}` returns `map["docker.io":""]`; consumers ignore those values and silently bypass mirroring. Explicitly accept and validate the supported map representation/value types, reject unsupported shapes with a contextual error, and add production-path regression cases.

matthyx (changes requested):
Reviewed `0ff36d7243e96f69ff88f995fdb91904b30485ba` against target `main` (currently `6f465e1d77779d631e655b285ca21aa864886dd7`; merge is conflict-free). Requesting changes for the two inline findings.

The linked null behavior is not reproduced at the production boundary: both base and head return nil without error for top-level JSON `null` and environment string `"null"`. The added helper-only test fails before the patch but bypasses `LoadConfig`'s `IsSet` guard, so it does not establish a user-visible fix. Meanwhile malformed non-string shapes still silently become empty/empty-valued maps, and the new whitespace branch introduces a fail-open regression that can clear a valid file mirror map.

Validation: exact-head `go test ./config -count=1` passed in a credential-free, read-only Docker mount. Additional base/head `LoadConfig` reproductions confirmed the outcomes above. All current GitHub checks are green, but CodeRabbit was rate-limited and did not review this diff.

History: #352 introduced mirroring, #361 fixed dotted map keys, and #593 deliberately made malformed environment JSON fail; all were merged, not rejected. A GitHub search of up to 100 PRs/issues using `proxyRegistryMap`, `registry mirror viper config`, `null`, and `invalid proxyRegistryMap` found no duplicate fix or previously rejected approach. Search limits mean this is not proof none exists.

config/config.go
```diff
@@ -267,13 +267,32 @@ func trustedVendorsFromViper(v *viper.Viper) []string {
...
+	if val == nil {
+		return nil, nil
+	}
+	if raw, ok := val.(string); ok {
+		// A blank/whitespace-only value is malformed JSON, not an opt-out: fall through to
+		// json.Unmarshal so it surfaces the same "unexpected end of JSON input" error as any
+		// other malformed value, instead of silently clearing a configured mirror map. An
+		// explicit JSON "null" is still handled below: json.Unmarshal leaves proxyMap nil.
+		var proxyMap map[string]string
+		if err := json.Unmarshal([]byte(raw), &proxyMap); err != nil {
```
matthyx: [HIGH] Apply the same member-type validation to environment JSON. `PROXYREGISTRYMAP='{"docker.io":null}'` currently unmarshals into `map[string]string{"docker.io":""}` with no error, while the equivalent file object is rejected below. I reproduced this through `LoadConfig` on this head; the empty destination is skipped by `rewriteImageRef`, so malformed env configuration can still silently bypass the mirror. Decode string input into an intermediate object and reuse the validation loop (while preserving supported top-level `null`), then add a `LoadConfig` regression case for both sources.

matthyx (changes requested):
Re-reviewed current head `c3a5f722b9794d96babbfe1cefe56c4ef478b127`. The follow-up correctly fixes the two previously reported code paths: whitespace input errors again, and file-backed unsupported shapes/non-string values are explicitly rejected. One environment-input parity blocker remains inline.

Validation: exact-head `go test ./config -count=1` passed in a credential-free container with a read-only source mount. An additional table-driven `LoadConfig` reproduction passed for valid maps, top-level null, empty objects, arrays, file null members, environment null/valid maps, and whitespace precedence—except the added environment-null-member expectation, which failed because the loader returned no error.

The PR also cannot be approved yet because DCO is failing: the new `c3a5f72` commit has no `Signed-off-by` trailer. All other listed checks pass; CodeRabbit is rate-limited and did not substantively review this head.

Independent recheck: code-review lane found the original findings repaired; architecture remains WATCH because file and environment JSON enforce different member types. Final verdict remains REQUEST CHANGES until the environment path is made consistent (or its intended null-member semantics are established and tested) and DCO passes.

bhuvan-somisetty:
Hi @matthyx,

Thank you for the review!

I have updated `proxyRegistryMapFromViper` in `config/config.go` so environment-based `PROXYREGISTRYMAP` strings are unmarshalled into `map[string]interface{}` prior to member type validation.

This ensures 100% consistent member-level type enforcement and error formatting between environment and file configurations (such as rejecting `{ "docker.io": null }` or non-string values).

I also added regression unit tests in `config/config_test.go` covering environment null member values and squashed all changes into a clean, single commit with DCO sign-off.

bhuvan-somisetty:
Squashed all changes into a clean DCO-signed commit. `proxyRegistryMapFromViper` unmarshals environment JSON into `map[string]interface{}` to ensure member-level type and null-value parity with file configurations.

adapters/v1/domain_to_armo.go
```diff
@@ -696,24 +696,15 @@ func syftCoordinatesToCoordinates(c []v1beta1.SyftCoordinates) []containerscan.C
...
 }
 
-// ParseImageManifest extracts and builds containerscan.ImageManifest metadata from a Grype document source.
-// It returns (nil, nil) if the source type is "directory" or if rawManifest.RawConfig is nil, as directory targets
-// and sources without raw image configuration do not produce an image manifest.
 func ParseImageManifest(grypeDocument *v1beta1.GrypeDocument) (*containerscan.ImageManifest, error) {
 	if grypeDocument == nil || grypeDocument.Source == nil {
 		return nil, fmt.Errorf("empty grype document")
 	}
-	if grypeDocument.Source.Type == "directory" {
```
matthyx, line 706: [MEDIUM] Keep the merged #994 guards and tests. This squash deletes both the directory-source early return here and the `RawConfig == nil` guard below, along with their two regression tests. Reinstating those scenarios against this head makes both fail: directory targets return a JSON type error, and missing raw config returns `unexpected end of JSON input`; `BackendAdapter.SendReport` then emits misleading warnings for valid scans. Restore `adapters/v1/domain_to_armo.go` and its tests from current `main` so this config PR does not revert #994.

matthyx (changes requested):
Re-reviewed force-pushed head `47b4b013a166eeabd458de6880b673d0f0579cdf` against current `main` (`2d104fd0d49a0282b9422a34746c1db7bc0abb8d`). The configuration work now resolves the prior blockers: file and environment values share strict member validation, the environment null-member production test passes, whitespace remains an error, and top-level null remains supported. DCO and every listed CI check are green.

One unrelated regression was introduced by the squash and is detailed inline: the PR now cleanly reverts the merged #994 `ParseImageManifest` guards and deletes their tests. This is a real four-file PR diff from merge base `6f465e1`, not noise from the two newer commits on `main`.

Validation: isolated exact-head configuration tests plus a production-path `LoadConfig` table passed. Reinstating #994's two regression scenarios ephemerally on this head failed: directory targets returned `json: cannot unmarshal string into Go value of type source.ImageMetadata`, and missing `RawConfig` returned `unexpected end of JSON input`. The production caller logs those errors as warnings and continues reporting, so impact is misleading warnings rather than scan loss.

Independent lanes agree on REQUEST CHANGES: code review found one MEDIUM regression; architecture status is BLOCK because an unrelated merged fix would be reverted. Restore the two adapter files to current-main behavior, leaving this PR scoped to configuration validation.

bhuvan-somisetty:
I have restored `adapters/v1/domain_to_armo.go` and `adapters/v1/domain_to_armo_test.go` to match `main` unchanged, preserving PR #994. The PR is now strictly scoped to configuration validation. All unit tests pass.

matthyx (approved):
Re-reviewed latest head `a5df1dd9bede4e4985baf9e8d49efbf468bd17d8` against current `main` (`f193d322623edc9ce2428731e05607d831d6a3ac`). No blocking findings remain.

The configuration change is needed to avoid silently accepting malformed `proxyRegistryMap` values. File and environment inputs now share strict shape/member validation, top-level JSON `null` remains supported, whitespace remains an error, and the production environment-null-member regression is covered. The accidental #994 rollback is gone: both adapter files match current main and are absent from the PR diff. The new main commit (#1001) affects different adapter files, and a refreshed merge preview is conflict-free.

Validation: exact-head `go test ./config -count=1` passed in a credential-free Docker container with a read-only source mount, including an ephemeral production-path table covering file/env null, valid maps, empty objects, arrays, null members, and whitespace precedence. All reported checks, including DCO, lint, Basic-Test, build, CodeQL, security, dependency, and license checks, are green. CodeRabbit was rate-limited, so its success status is not substantive review evidence. Independent code and architecture reviews both found the head clear.

Related-history searches and prior review remain unchanged: #352 introduced mirroring, #361 fixed dotted keys, #593 deliberately made malformed environment JSON fail, and #994's guards are preserved. No duplicate or rejected equivalent was found within the previously recorded bounded GitHub searches; that search limit is not proof none exists. Verdict: approve.

PR #973, opened 2026-09-13
fix(adapters): handle empty Kind cleanly in normalizedTarget

bhuvan-somisetty:
## Problem
When `normalizedTarget()` constructed the human-readable provenance target string (`normalizedTarget`) for suppressed findings in `SecurityException` and `ClusterSecurityException` objects, it omitted `APIGroup` when `Kind` or `Name` were present, and unconditionally concatenated `target += "/" + r.Name` even when `r.Kind` was empty (`""`). As a result, resource matches with an `APIGroup` lost group metadata, and resource matches specifying a resource `Name` without a `Kind` generated malformed target strings with double slashes (e.g. `prod//my-app`) or leading slashes (e.g. `/my-app`).

## Root Cause
`normalizedTarget()` assumed `r.Kind` was always non-empty when `r.Name` was present and ignored `r.APIGroup` when `Kind`/`Name` were set.

## Solution
Updated `normalizedTarget()` in `adapters/v1/securityexception.go` to slice-join non-empty `APIGroup`, `Kind`, and `Name` fields in canonical order `[namespace/]APIGroup/Kind/Name`. Added Godoc comments to satisfy docstring coverage checks.

## Testing
- Added comprehensive unit test cases in `TestNormalizedTarget` in `adapters/v1/securityexception_test.go` covering resource matches with `APIGroup`, `Kind`, `Name`, namespaced, and cluster-wide combinations.
- Verified all unit tests pass clean via `go test ./adapters/v1`.

## Impact
Diagnostics logged by `logSuppression` and metadata stored in manifest suppression attributes now receive properly formatted and complete resource targets without omitting `APIGroup` or introducing malformed slashes.

Fixes #972

PR #1003, opened 2026-09-15
fix(adapters): preserve exact resource name matching and fix WLID name extraction

### LINKED ISSUE #998: bug(adapters): matchResources in securityexception_match.go uses case-sensitive comparison for workload resource Name
# Description
`matchResources` in `adapters/v1/securityexception_match.go` checks whether a `SecurityException` CRD's `spec.match.resources` applies to a target workload:

```go
func matchResources(resources []sev1beta1.ResourceMatch, target ExceptionTarget) bool {
	if len(resources) == 0 {
		return true
	}
	for _, r := range resources {
		if r.Kind == "" && r.Name == "" && r.APIGroup == "" {
			continue
		}
		if r.Kind != "" && !strings.EqualFold(r.Kind, target.Kind) {
			continue
		}
		if r.Name != "" && r.Name != target.Name {
			continue
		}
		if r.APIGroup != "" && (target.APIGroup == nil || !strings.EqualFold(r.APIGroup, *target.APIGroup)) {
			continue
		}
		return true
	}
	return false
}
```

Notice that while `Kind` and `APIGroup` are compared case-insensitively using `strings.EqualFold`, resource `Name` uses strict equality (`r.Name != target.Name`).

Workload names in Kubernetes resources and WLID strings can be generated or specified with varying letter cases or normalized forms. Enforcing strict case sensitivity on resource `Name` while ignoring case on `Kind` and `APIGroup` causes resource matches to fail when the case of `r.Name` differs from `target.Name`.

# Environment
OS: `Linux / Kubernetes`
Version: `v1.28.0`

# Steps To Reproduce
1. Create a `SecurityException` CRD with `spec.match.resources[0].name: "my-workload"`.
2. Scan a target workload whose `target.Name` is `"My-Workload"`.
3. Observe that `matchResources` returns `false` due to `r.Name != target.Name`.

# Expected behavior
`matchResources` should use `strings.EqualFold(r.Name, target.Name)` for case-insensitive resource name matching.

# Actual Behavior
Security exceptions specifying workload resource names with casing differences (e.g. `MyWorkload` vs `myworkload`) fail to match the target workload, causing valid security exceptions to be ignored.

# Additional context
File location: `adapters/v1/securityexception_match.go` in `matchResources()`.

bhuvan-somisetty:
Fixes #998

This PR preserves exact case-sensitive matching for resource `Name` in `matchResources` while fixing WLID raw name extraction via `rawNameFromWlid`.

Previously, `wlidpkg.GetNameFromWlid` ran the name segment through `GetK8SKindFronList`, title-casing resource names (e.g. `service` -> `Service`) when they coincided with Kubernetes kind names, causing exact resource name matching to fail. Implementing direct raw WLID name extraction restores raw name integrity and preserves exact resource matching.

## Testing
- Added regression test `TestBuildExceptionTarget_WLIDKindCollisionName` in `adapters/v1/securityexception_match_test.go` confirming exact raw resource name extraction and matching for workload names colliding with Kubernetes kinds.


adapters/v1/securityexception_match.go
```diff
@@ -97,7 +97,7 @@ func matchResources(resources []sev1beta1.ResourceMatch, target ExceptionTarget)
 		if r.Kind != "" && !strings.EqualFold(r.Kind, target.Kind) {
 			continue
 		}
-		if r.Name != "" && r.Name != target.Name {
+		if r.Name != "" && !strings.EqualFold(r.Name, target.Name) {
```
matthyx: **High — this masks the producer bug and leaves combined selectors broken.** `BuildExceptionTarget` gets the name from `wlid.GetNameFromWlid` at line 232. In the pinned `utils-k8s-go v0.0.35`, that helper passes the name through `GetK8SKindFronList`, so a valid Deployment named `service` becomes `Service`. `EqualFold` makes a resource-only exception appear fixed, but line 259 still sends `Service` to `GetWorkloadLabels`, and `repositories/apiserver.go` uses it in an exact Kubernetes GET; adding `objectSelector` therefore still fails to resolve the real `service` workload. It also changes the public contract documented as an "Exact resource name." Please preserve the raw WLID name at the extraction boundary, retain exact comparison here, and add a `BuildExceptionTarget` regression asserting both `target.Name == "service"` and the name passed to label lookup. The direct `ExceptionTarget{Name: "nginx"}` test does not exercise the faulty producer.

matthyx (changes requested):
Request changes at `b64eeab877c548eadf4b9e5e792b9591e11135f0`, revalidated against current `main` `6f465e1d77779d631e655b285ca21aa864886dd7`. The reported mismatch is real for some WLIDs, but this change broadens the documented exact-name selector instead of repairing the producer that corrupts names. The inline blocker describes the remaining failure.

History search found no duplicate/superseding resource-name PR. #383 introduced and merged exact resource-name matching; #973 concerns normalized target metadata, while #983 concerns image-pattern casing. I found no prior explicit rejection of this approach. Search covered GitHub-indexed PRs/issues for `matchResources`, `securityexception_match.go`, case-insensitive/resource-name terms, plus local file and commit history; indexing/search limits mean this is not proof no other match exists.

Validation: the added regression fails on the previous base and passes on this head; focused adapter tests, vet, formatting, and whitespace checks pass, and all reported CI checks are green. The target-branch advance is unrelated and a read-only merge-tree is clean. These checks validate the proposed comparison but do not cover `BuildExceptionTarget` or the label-lookup path that remains broken.

Verdict: request changes. Preserve exact name matching, fix WLID name extraction, and add an end-to-end target-construction regression for a valid lowercase workload name that is also a Kubernetes kind.

matthyx (changes requested):
Rechecked `0d1fa4f87ff15b500019f5b03498f54a0531913d` against current `main` `6f465e1d77779d631e655b285ca21aa864886dd7`. The prior code blocker is resolved: exact resource-name matching is restored, raw WLID name extraction repairs the producer, and the regression covers both `ExceptionTarget.Name` and the label-lookup argument. I found no remaining code or architecture findings.

Validation: the added regression fails on the prior head (`service` becomes `Service` in both assertions) and passes on this head; focused adapter tests, `go vet ./adapters/v1`, formatting, and diff checks pass. The merge tree against current `main` is clean, and all reported checks except DCO pass.

I cannot approve this head yet because the DCO check is `ACTION_REQUIRED`: commit `0d1fa4f` lacks a `Signed-off-by` line. Please remediate the commit sign-off and update the PR title/body, which still describe the abandoned case-insensitive-name implementation. Once the resulting latest head has DCO green and the same code, this is ready for approval.

bhuvan-somisetty:

I have preserved exact case-sensitive matching for resource `Name` in `matchResources` and fixed WLID name extraction in `BuildExceptionTarget` (`adapters/v1/securityexception_match.go`). Specifically, `rawNameFromWlid` extracts the raw resource name segment without passing it through `wlidpkg.GetNameFromWlid`'s `GetK8SKindFronList` transform (which title-cased workload names that happened to collide with Kubernetes kind names).

I also added a regression test in `securityexception_match_test.go` (`preserves a resource name that collides with a known kind`) verifying that a Deployment literally named `"service"` keeps its exact name `"service"` for target resource matching and label lookup.

bhuvan-somisetty:
Updated title and description to match the raw WLID name extraction fix (`rawNameFromWlid`). Exact case-sensitive resource name matching is preserved and all CI checks are green.

matthyx (approved):
Approved at `22c0b8285258d37755b533b22802e0be07111bcb` against current `main` `6f465e1d77779d631e655b285ca21aa864886dd7`.

The prior blockers are resolved: exact resource-name semantics remain intact, raw WLID name extraction fixes the producer defect for both matching and label lookup, regression coverage exercises that boundary, the PR title/body now describe the implemented fix, and DCO is green. The three scoped file blobs are identical to the previously code-cleared head; this rewrite also rebases the branch over main's unrelated #994 change.

Fresh isolated validation passed for `TestMatchResources` and `TestBuildExceptionTarget`, as did `go vet ./adapters/v1`, formatting, diff checks, and the clean merge-tree check. All reported hosted checks are successful. Independent code review found zero issues and recommended APPROVE; architecture status is CLEAR. No remaining blockers found.

PR #1010, opened 2026-09-15
fix(adapters): do not retry non-retryable 4xx client errors in shouldRetryReport

### LINKED ISSUE #1005: bug(adapters): shouldRetryReport in BackendAdapter retries non-retryable 4xx client errors (400, 413, 422)
# Description
`shouldRetryReport` in `adapters/v1/backend.go` determines whether HTTP requests sent to the Event Receiver service should be retried on failure:

```go
		resp.StatusCode != http.StatusNotFound
}
```

While `shouldRetryReport` correctly excludes `401 Unauthorized`, `403 Forbidden`, and `404 Not Found` from retries, it returns `true` for all other HTTP status codes — including permanent client-side errors such as `400 Bad Request`, `413 Payload Too Large`, `422 Unprocessable Entity`, and `405 Method Not Allowed`.

When a report payload is malformed or exceeds maximum request size limits, `BackendAdapter` retries the request with exponential backoff until `maxElapsedTime` (60 seconds) expires.

# Environment
OS: `Linux / Kubernetes`
Version: `v1.28.0`

# Steps To Reproduce
1. Configure `BackendAdapter` to post a report payload to an Event Receiver endpoint that responds with `400 Bad Request` or `413 Payload Too Large`.
2. Observe `httpPostWithContext` behavior.
3. Notice that `shouldRetryReport` returns `true` for HTTP 400/413/422, causing `backoff.Retry` to retry the unalterable request for the full 60-second backoff duration.

# Expected behavior
`shouldRetryReport` should return `false` for non-retryable client error status codes (`400`, `413`, `422`, `405`).

# Actual Behavior
Permanent client errors (e.g. 400 Bad Request, 413 Payload Too Large) are repeatedly retried for up to 60 seconds, wasting network resources and delaying error feedback.

# Additional context
File location: `adapters/v1/backend.go` in `shouldRetryReport()`.

bhuvan-somisetty:
## Description

Fixes #1005

This PR updates `shouldRetryReport` in `adapters/v1/backend.go` so that client-side 4xx errors (such as 400 Bad Request, 413 Payload Too Large, 422 Unprocessable Entity, 401, 403, 404, etc.) are recognized as non-retryable and return `false`, with the exception of `429 Too Many Requests` which continues to be retried.

Previously, `shouldRetryReport` only excluded 401, 403, and 404, causing other 4xx client errors like 400, 413, and 422 to wastefully consume retry attempts despite payload errors being permanent.

## Testing
- Added unit test `TestShouldRetryReport` in `adapters/v1/backend_test.go` covering status codes: 200, 400, 401, 403, 404, 405, 413, 422, 429, 500, 502, 503.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved request retry handling for temporary failures, including timeouts, rate limits, server errors, and unavailable responses.
  * Prevented retries for permanent client errors, reducing unnecessary repeat requests.
  * Added coverage confirming successful retries after request timeouts and appropriate handling of common client and server responses.


adapters/v1/backend.go
```diff
@@ -233,9 +233,13 @@ func parseRetryAfter(resp *http.Response) (time.Duration, bool) {
 // on the first one loses data that a retry would have delivered (#486). The retry budget is
 // bounded by maxElapsedTime, so a genuinely broken backend still fails fast enough.
 func shouldRetryReport(resp *http.Response) bool {
-	return resp.StatusCode != http.StatusUnauthorized &&
-		resp.StatusCode != http.StatusForbidden &&
-		resp.StatusCode != http.StatusNotFound
+	if resp.StatusCode == http.StatusTooManyRequests {
+		return true
+	}
+	if resp.StatusCode >= 400 && resp.StatusCode < 500 {
```
matthyx, line 242: **HIGH — preserve retryability for HTTP 408.** This range now makes `408 Request Timeout` permanent. If the event receiver times out before receiving the full report and succeeds on the next attempt, `httpPostWithContext` returns immediately after the first 408; a summary failure then prevents subsequent chunks from being sent, and the shared path also drops scan-failure reports. RFC 9110 §15.5.9 explicitly permits repeating a 408 request. I reproduced this with a server returning 408 then 200: the test passes on base (`3b73e54`) with two attempts but fails on this head with `received status code: 408`. Please keep 408 retryable (for example, alongside the 429 exception) and add an HTTP-level 408→200 regression test. https://www.rfc-editor.org/rfc/rfc9110.html#section-15.5.9

matthyx (changes requested):
Reviewed `7e4ac29387d1537b27f5f825a5ed183e25b256ab` against `main` at `3b73e54e02dd728a3664dcaa0ea256acc3635298`. The fix is needed for permanent 400/405/413/422 responses and directly repairs the retry predicate, but the blanket 4xx rule introduces the confirmed 408 delivery regression described inline.

Related history supports distinguishing transient from permanent responses: #487 deliberately preserved retries for transient 500s, while #637 and #755 hardened `Retry-After` behavior. Searches across open/closed issues and PRs for `shouldRetryReport`, retry-report, 4xx, client-error, payload-too-large, and unprocessable-entity terms found no duplicate implementation; #1001 is adjacent but concerns downstream `postResults` validation. Search results are GitHub-indexed and capped at 100 per query. No previously rejected equivalent approach was found; earlier blockers on #597/#637 were addressed before those PRs merged.

Validation: the isolated targeted suite (`go test ./adapters/v1 -run 'TestShouldRetryReport|TestHttpPostWithContext' -count=1`) passes. A disposable HTTP-level regression where the server returns 408 once then 200 passes on the base commit and fails on this head immediately with `received status code: 408`. All current PR checks are green and GitHub reports the branch mergeable. `git diff --check` also reports two extra EOF blank lines in `backend_test.go`; that is nonblocking.

Nonblocking contract question: this predicate runs before `Retry-After` parsing, so a 413 carrying `Retry-After` is also permanent even though RFC 9110 permits 413 to be temporary. Since #1005 explicitly calls 413 permanent, I am not treating that as a second blocker, but the intended backend contract should be confirmed.

Verdict: request changes for the confirmed 408 regression.

matthyx (changes requested):
Re-reviewed head `6ea23bbc61210bb3996f73cd7baafe1fa3033fc6`.

The previous HIGH blocker is resolved: 408 is now exempted before the permanent-4xx branch, and `TestHttpPostWithContext_RetriesRequestTimeout` verifies the real 408→200 retry path with exactly two attempts. Independent code review found no substantive defects; independent architecture review downgraded the prior BLOCK to WATCH only for the already-documented, nonblocking 413/`Retry-After` contract assumption.

Fresh validation passed in an isolated, network-disabled container:

`go test ./adapters/v1 -run 'TestShouldRetryReport|TestHttpPostWithContext' -count=1`

`go vet ./adapters/v1` also passed in the independent code lane. All code, build, lint, security, and test checks are green. The extra EOF blank lines in `backend_test.go` remain cosmetic. `main` advanced after the PR snapshot, but there are no intervening changes to the two affected files and GitHub still reports the PR mergeable.

I cannot approve yet because the required DCO check is `ACTION_REQUIRED`: commit `6ea23bbc` lacks a `Signed-off-by` line. Please sign off/amend that commit and update the branch; once DCO is green, no code blocker remains from this review.

Verdict: request changes for the DCO merge gate only.

bhuvan-somisetty:

I have updated `shouldRetryReport` in `adapters/v1/backend.go` so that `http.StatusRequestTimeout` (`408`) and `http.StatusTooManyRequests` (`429`) remain retryable, while non-retryable 4xx client status codes (400, 401, 403, 404, 405, 413, 422) return `false` to be treated as permanent failures. I also added `TestHttpPostWithContext_RetriesRequestTimeout` to verify that 408 responses are retried appropriately and cleaned up EOF formatting in `backend_test.go`.

bhuvan-somisetty:
Re-validated against `main`. Transient 408 (Request Timeout) and 429 (Too Many Requests) status codes remain retryable while non-retryable 4xx client errors fail fast. DCO sign-off is verified and all CI checks are green.

matthyx (approved):
Reviewed signed head `f7b70c41ab2953a8cc5281e93a48edc11e5b795b`.

The prior blockers are resolved. HTTP 408 and 429 remain retryable, other 4xx responses requested by #1005 fail fast, and the HTTP-level 408→200 regression verifies two attempts and successful delivery. The branch rewrite preserves the reviewed behavior, removes the EOF whitespace, and adds only a harmless defensive nil guard; that guard is not needed by the current production caller and is not being credited as end-to-end nil-response recovery.

Fresh isolated validation passed on the PR head:

`go test ./adapters/v1 -run 'TestShouldRetryReport|TestHttpPostWithContext' -count=1`

Because `main` advanced and touched `backend_test.go`, I also created a clean local merge with current `main` (`996afd9`) and ran the combined retry and exception-matching tests; they passed. `go vet ./adapters/v1`, `gofmt`, and `git diff --check` passed in the independent code lane. Both independent review lanes are clear: code recommendation APPROVE, architectural status CLEAR.

The PR is open, non-draft, mergeable, and all current build, test, lint, security, dependency/license, CodeQL, CodeRabbit, and DCO checks are green. No material blocker remains.

Verdict: approve.

PR #1011, opened 2026-09-15
fix(adapters): use case-insensitive string comparison for Fix.State in hasKnownFix

### LINKED ISSUE #1006: bug(adapters): hasKnownFix in fixstate.go uses case-sensitive string comparison for Fix.State
# Description
`hasKnownFix` in `adapters/v1/fixstate.go` evaluates whether a Grype match has a known fix:

```go
const fixStateFixed = "fixed"

func hasKnownFix(m v1beta1.Match) (bool, string) {
	...
	if m.Vulnerability.Fix.State == fixStateFixed {
		return true, unknownFixVersion
	}
	...
}
```

Notice that `m.Vulnerability.Fix.State == fixStateFixed` uses exact string equality against `"fixed"`.

Grype output, external VEX documents, or upstream vulnerability data feeds can serialize fix state strings with capitalized or uppercase formatting (such as `"Fixed"` or `"FIXED"`). Because `hasKnownFix` performs exact string comparison rather than `strings.EqualFold`, matches carrying `"Fixed"` or `"FIXED"` evaluate to `hasKnownFix = false`, causing fixed vulnerabilities to be misclassified as unfixed.

# Environment
OS: `Linux / Kubernetes`
Version: `v1.28.0`

# Steps To Reproduce
1. Create a `v1beta1.Match` where `m.Vulnerability.Fix.Versions` is empty and `m.Vulnerability.Fix.State` is `"Fixed"`.
2. Pass the match to `hasKnownFix(m)`.
3. Observe that `hasKnownFix` returns `false, ""` because `"Fixed" == "fixed"` evaluates to `false`.

# Expected behavior
`hasKnownFix` should use `strings.EqualFold(m.Vulnerability.Fix.State, fixStateFixed)` for case-insensitive state checking.

# Actual Behavior
Vulnerability matches reporting fix state as `"Fixed"` or `"FIXED"` are evaluated as having no fix available.

# Additional context
File location: `adapters/v1/fixstate.go` in `hasKnownFix()`.

bhuvan-somisetty:
## Description

Fixes #1006

This PR updates `hasKnownFix` in `adapters/v1/fixstate.go` to use `strings.EqualFold(m.Vulnerability.Fix.State, fixStateFixed)` instead of direct equality (`==`).

Fix state strings supplied by vulnerability scanners or external feeds may use title-case or uppercase variants (e.g. `"Fixed"` or `"FIXED"`). Using case-insensitive string comparison ensures that all variations are properly recognized as known fixes.

## Testing
- Added unit tests in `adapters/v1/fixstate_test.go` covering `"Fixed"` and `"FIXED"` fix state strings.


matthyx (commented):
Reviewed head `7abb3bf6517aead755eb4568ff7be1058743fd2f` against target `main` (including a clean merge simulation with current `main` at `6f465e1d77779d631e655b285ca21aa864886dd7`). The implementation is scoped and sound: the base returns `(false, "")` for `State: "Fixed"`, the PR returns `(true, "unknown")`, and both consumers continue to share `hasKnownFix`. Full `go test ./adapters/v1/... -count=1`, `go vet ./adapters/v1/...`, and `git diff --check` pass in a network-disabled, credential-free container; current CI is green.

I could not verify the stated production source of mixed-case values. Pinned Grype v0.104.1 defines lowercase fix-state constants, maps DB states to those constants, serializes them directly, and kubevuln copies the string unchanged. I also found no VEX path that mutates `Fix.State`. The storage API leaves the field unconstrained, so accepting variants is a reasonable defensive compatibility choice, but the need is not established by the synthetic unit cases alone.

Could you provide the supported producer/version and a representative payload that emits `"Fixed"` or `"FIXED"`, or confirm that accepting noncanonical casing is an intentional compatibility policy? No implementation defect found; verdict is **needs clarification** pending that evidence/intent confirmation.

bhuvan-somisetty:
Agreed - pinned Grype v0.104.1 only ever emits lowercase fix-state constants and nothing in our VEX path mutates Fix.State, so there's no demonstrated production source for mixed-case values. Closing until there's a concrete producer/payload that needs this.

bhuvan-somisetty:

I have added an explicit doc comment in `adapters/v1/fixstate.go` confirming that using `strings.EqualFold` for `Fix.State` is an intentional defensive compatibility choice to handle noncanonical casing (e.g. `"Fixed"`, `"FIXED"`) gracefully when consuming external data feeds or unconstrained custom storage APIs.

PR #1012, opened 2026-09-15
fix(adapters): track expired SecurityExceptions stats regardless of target match

### LINKED ISSUE #1007: bug(adapters): ConvertToVulnerabilityExceptionPolicies skips expired SecurityExceptions without updating expired stats when exception spec match fails
# Description
`ConvertToVulnerabilityExceptionPolicies` in `adapters/v1/securityexception.go` converts CRD-based `SecurityException` and `ClusterSecurityException` objects into vulnerability exception policies and collects expired exception telemetry in `domain.ExceptionStats`:

```go
for i := range exceptions {
	se := &exceptions[i]
	if !matchExceptionTarget(se.Spec.Match, target, false) {
		continue
	}
	namespace := se.Namespace
	for _, vuln := range se.Spec.Vulnerabilities {
		if !shouldSuppress(vuln) {
			continue
		}
		if isExpired(effectiveExpiresAt(se.Spec, vuln), now) {
			stats.ExpiredBySource["SecurityException"]++
			...
			continue
		}
		...
	}
}
```

Notice that `matchExceptionTarget(se.Spec.Match, target, false)` is evaluated on line 51 BEFORE `isExpired(effectiveExpiresAt(...))` on line 59.

If an expired `SecurityException` does not match the target workload currently being scanned (e.g. scoped to a different workload name or image), `matchExceptionTarget` returns `false` and line 51 `continue`s immediately. As a result, `stats.ExpiredBySource["SecurityException"]` is never incremented for non-matching workloads, misrepresenting expired security exception telemetry across workload scans.

# Environment
OS: `Linux / Kubernetes`
Version: `v1.28.0`

# Steps To Reproduce
1. Create a `SecurityException` CRD with an expired `expiresAt` timestamp and a `spec.match.resources` filter for a specific workload (`"workload-A"`).
2. Scan a target workload `"workload-B"`.
3. Observe that `matchExceptionTarget` returns `false`, causing line 51 to skip `isExpired` checking, leaving `stats.ExpiredBySource["SecurityException"]` unincremented.

# Expected behavior
Expired exception statistics in `domain.ExceptionStats.ExpiredBySource` should track all expired exceptions regardless of target workload matching.

# Actual Behavior
Expired security exception telemetry is omitted for workloads that do not match the exception's target specification.

# Additional context
File location: `adapters/v1/securityexception.go` in `ConvertToVulnerabilityExceptionPolicies()`.

bhuvan-somisetty:
## Description

Fixes #1007

This PR updates `ConvertToVulnerabilityExceptionPolicies` in `adapters/v1/securityexception.go` so that vulnerability expiry checks (`isExpired`) run before target matching (`matchExceptionTarget`) is applied for policy creation.

Previously, if `matchExceptionTarget` returned `false`, the loop immediately skipped the exception via `continue`, missing the `isExpired` check and leaving `stats.ExpiredBySource` under-counted. Evaluating `isExpired` first ensures that telemetry tracking expired exceptions remains accurate regardless of whether the exception targets the current workload.

## Testing
- Added unit test `TestConvertVulnerabilityExceptions_ExpiredUnmatchedTarget` in `adapters/v1/securityexception_test.go` verifying `stats.ExpiredBySource` is updated for expired exceptions even when target match fails.


matthyx (commented):
The code change preserves target filtering for emitted policies, and the new regression test passes. I could not find a correctness or security defect in that path, but I don't think the metric semantics are established clearly enough to approve yet.

Could a maintainer confirm which signal `kubevuln_exceptions_expired_total` is intended to represent?

- The current contracts describe expired CRDs “skipped this call/scan” (`core/domain/cve.go`, `core/services/scan.go`, and `docs/API.md`). Before this change, that meant expired suppressible entries belonging to exceptions applicable to the workload being evaluated.
- At this commit, an expired exception scoped to workload A is counted while unrelated workloads B and C are evaluated too, because expiry is counted before `matched` is checked. Each uncached workload evaluation then adds that count to the monotonic metric, so the result depends on unrelated scan volume and cache hits.

If per-evaluation observations across the entire fetched CRD set are intended, the implementation looks sound, but the metric/domain documentation should say that explicitly; I would also add equivalent regression coverage for the independently changed `ClusterSecurityException` loop. If the metric is meant to count relevant suppressions skipped for the current scan, the target check needs to remain ahead of expiry accounting.

Reviewed `e9ba7691b00851743e14dfb31216dc9bc9b49312` against `main` at `3b73e54e02dd728a3664dcaa0ea256acc3635298`. `go test -vet=off -count=1 ./adapters/v1` passed in a credential-free, network-disabled, read-only container; repository CI is green. `gofmt -d` also reports one extra blank line in `adapters/v1/securityexception_test.go` (non-blocking).

bhuvan-somisetty:

I have added regression test coverage in `adapters/v1/securityexception_test.go` (`TestConvertVulnerabilityExceptions_ExpiredUnmatchedTarget`) for both the `SecurityException` and `ClusterSecurityException` loops, verifying that `stats.ExpiredBySource["ClusterSecurityException"]` and `stats.ExpiredBySource["SecurityException"]` count expired CRD suppressions even when target resource matching fails for the current workload.

I also cleaned up the extra EOF blank lines in `adapters/v1/securityexception_test.go`.

bhuvan-somisetty:
Rebased on `main` with DCO sign-off. The expired `SecurityExceptions` stats tracking has been updated to count expired rules regardless of target match, accompanied by the regression unit test.

matthyx (changes requested):
Re-reviewed head `96e181f8b6bdb729c14cf5d2f53d968e1345add4` against current `main` (`996afd9df53f8a96c3f8a748c9d6422c6c0bd803`). The requested expiry-accounting change now has regression coverage for both `SecurityException` and `ClusterSecurityException`, preserves target filtering for emitted policies, and its targeted test passes.

**MEDIUM blocker — `adapters/v1/domain_to_armo.go:699-712`:** this rewritten commit also reverses merged PR #994 by removing the directory-source and nil-`RawConfig` guards from `ParseImageManifest`, and deletes both associated tests from `adapters/v1/domain_to_armo_test.go`. A fresh isolated probe at this head reproduces both regressions: a directory target returns `json: cannot unmarshal string into Go value of type source.ImageMetadata`, while image metadata without config returns `unexpected end of JSON input`. `SubmitCVE` logs these errors and continues, so the demonstrated impact is restored warning noise and a changed parser contract, not lost reports.

Please restore the #994 guards, documentation, and tests unchanged while retaining the expiry-accounting change.

The earlier telemetry-semantics question remains a non-blocking watch item: the counter now records expired suppressible-entry observations across the fetched set on each workload cache miss, including unrelated targets. If that is intentional, the metric/domain documentation should describe the per-entry, per-conversion behavior rather than implying unique expired CRDs skipped for a scan.

Validation: isolated/network-disabled targeted expiry test passed; temporary #994 regression probes failed with the two errors above; `git diff --check` passed; current CI is green. Related-history recheck confirms #994 is the directly conflicting merged change, not a superseded approach.

Verdict: request changes until the unrelated #994 regression is removed.

bhuvan-somisetty:
I have restored `adapters/v1/domain_to_armo.go` and `adapters/v1/domain_to_armo_test.go` to match `main` unchanged, preserving PR #994's directory-source and nil `RawConfig` guards and tests. The branch now cleanly contains only the expired security exception telemetry tracking change and tests. All unit tests pass.

matthyx (approved):
Re-reviewed head `474326f511e7b268d12d97ee3aff08407be63005` against current `main` (`f193d322623edc9ce2428731e05607d831d6a3ac`).

The prior blocker is resolved: the accidental #994 reversal is gone, and `domain_to_armo.go` plus its tests match `main`. The PR now changes only `securityexception.go` and `securityexception_test.go`.

The fix is needed for #1007 and is sound: expired suppressible entries are counted independently of target matching, while unmatched exceptions still cannot emit policies. The regression test covers both `SecurityException` and `ClusterSecurityException`. The per-conversion/cache-dependent nature of this telemetry is consistent with #1007's explicit expected behavior; clearer metric wording would be a non-blocking follow-up.

No blocking correctness, security, compatibility, performance, or maintainability findings remain. Independent code review found no issues and architecture review is clear.

Validation at this head: isolated, network-disabled targeted tests covering conversion, expiry boundaries, and the restored image-manifest cases passed; `go vet ./adapters/v1` passed; `git diff --check` passed; current PR CI is green. The latest `main` advance affects only `backend_utils.go` and its tests, does not overlap this PR, and merges without conflict.

Verdict: approve.

PR #1016, opened 2026-09-16
fix(adapters): support single-string PURL exception attribute values in policySubcomponents

### LINKED ISSUE #1015: fix(adapters): support single-string PURL exception attribute values in policySubcomponents
### Problem Description
`policySubcomponents` in `adapters/v1/subcomponent.go` retrieves the `subcomponents` attribute from a `VulnerabilityExceptionPolicy`. It currently handles `[]string` and `[]interface{}` types. However, if a single `string` PURL value (e.g. `"pkg:golang/github.com/gin-gonic/gin@v1.7.0"`) is passed as the attribute value (such as when unmarshalled from certain JSON structures), it falls through to `default:` and returns `nil, true`. This treats the policy as having an invalid scope that matches nothing, failing closed and preventing subcomponent exceptions from matching.

### Proposed Solution
Add a `case string:` branch to `policySubcomponents` in `adapters/v1/subcomponent.go` to parse trimmed single-string PURL values into `[]string{trimmedVal}`.

bhuvan-somisetty:
## Description

When vulnerability exception policies are parsed or deserialized, single-string PURL exception attribute values for `subcomponents` previously hit the `default:` case in `policySubcomponents`, returning `nil, true` and failing to match any subcomponents.

This PR updates `policySubcomponents` to handle `case string:`, wrapping single string values into a `[]string{v}` slice and adding unit tests for single-string subcomponent attribute policy scoping.

## Related Issue

Fixes #1015

## Type of Change

## Checklist

- [x] Tests added and passing


matthyx (commented):
Reviewed `c198421` against `main` (`2d104fd`). The diff is small and the code is technically fine, but I can't establish that the change is needed, and it reverses a deliberate decision from #542 without saying so. Requesting clarification rather than approval.

**Main question: which producer emits `subcomponents` as a bare string?**

Every path I can find writes a list:

- CRD: `Subcomponents []string` in `pkg/securityexception/v1beta1/types.go:117`; the apiserver rejects a scalar there, so a CRD can't reach `policySubcomponents` with a string.
- In-process conversion: `adapters/v1/securityexception.go:264` always stores `normalizedSubcomponents(...)`, a `[]string`.
- Cloud-sourced policies: JSON-decoded lists arrive as `[]interface{}`, which is already handled. `armoapi-go` has no `subcomponents` field at all, so a string could only come from an ad-hoc attribute the backend chose to emit — and #1015 doesn't name one ("certain JSON structures").

#1015 was opened by the PR author an hour before the PR, with no reproduction. If there is a real producer (a backend payload, a fixture, a log line), please link it in the PR — that turns this from speculative into a fix I'd approve. If not, this is adding an input shape nothing generates.

**The `string` case was intentionally "unreadable" in #542.** The replaced test (`unreadable value is a scope nothing satisfies`) used `scoped("pkg:npm/lodash")` on purpose, and the function's doc comment still says "Any other type is still a scope, and one nothing can satisfy". The PR keeps that comment as-is, so after merge the doc and the code disagree. If the string case stays, the comment on `policySubcomponents` needs to list `string` alongside `[]string`/`[]interface{}` and explain which source produces it.

Minor, nonblocking: #1015 says the value is "trimmed"; the PR doesn't trim (consistent with the `[]interface{}` branch, which doesn't either). Fine, just note the issue text is off.

**Validation:** `go vet ./adapters/v1/` clean; `go test ./adapters/v1/ -run 'TestPolicySubcomponents|TestPURLMatches|TestApplySecurityExceptions|TestConvert|TestDomainToArmoScopes'` passes on the PR head. The new test covers the added branch; it's not a regression test in the fail-before sense because nothing on `main` produces the input. `gofmt -l` flags `adapters/v1/domain_to_armo_test.go`, which is pre-existing on `main`, not from this PR. CI on the head commit is green. No open/closed PR covers this change (#542 introduced the function, #876 changed a caller; neither touches the type switch).

bhuvan-somisetty:
Thanks for checking the producers @matthyx. You're right that subcomponents are always lists across CRDs, in-process normalized arrays, and JSON payloads, and scalar strings have no real producer. Closing this PR and #1015.

PR #1014, opened 2026-09-15
fix(tools): handle uppercase image references in LabelsFromImageID via ParseNormalized

bhuvan-somisetty:
## Description

Fixes #1009

This PR updates `LabelsFromImageID` in `internal/tools/tools.go` to fall back to `reference.ParseNormalized(imageID)` if `reference.Parse(imageID)` fails due to uppercase characters or unnormalized repository formats.

Previously, image references containing uppercase characters caused `reference.Parse` to return an error, resulting in `LabelsFromImageID` returning early and dropping `ImageTagMetadataKey`, `ImageNameMetadataKey`, and `ImageIDMetadataKey`.

## Testing
- Added unit tests in `internal/tools/tools_test.go` verifying image references with uppercase tags (e.g. `registry.com:8080/myapp:V1.0-RC1`) and uppercase repository names (e.g. `MyOrg/MyApp:V1.0-RC1`) correctly extract and sanitize all metadata labels.

## Summary by CodeRabbit

* **Bug Fixes**
  * Image labels are now generated correctly for image references with uppercase tags and mixed-case registry or organization names.
  * Image names and tags are normalized consistently, including lowercasing and dash formatting.

PR #1018, opened 2026-09-16
fix(adapters): trim whitespace in vulnerability policy names in buildCVEExceptionIndex

### LINKED ISSUE #1017: fix(adapters): trim whitespace in vulnerability policy names in buildCVEExceptionIndex
### Description

In `adapters/v1/backend_utils.go`, `buildCVEExceptionIndex` builds an index mapping policy names (lowercase) to exception indices for O(1) lookups during exception matching. However, if vulnerability policy names or search terms contain leading or trailing whitespace (e.g. `" CVE-2021-44228 "`), the map lookup fails because `strings.TrimSpace` is missing.

### Expected Behavior
Policy names and lookup keys should have leading and trailing whitespace trimmed before lowercasing and indexing/lookup.

### Proposed Fix
Add `strings.TrimSpace` to policy names in `buildCVEExceptionIndex` and lookup key in `lookup`.

bhuvan-somisetty:
## Description

In `adapters/v1/backend_utils.go`, vulnerability policy names indexing and lookups in `buildCVEExceptionIndex` and `lookup` did not trim leading and trailing whitespace. If policy names or search inputs carried surrounding spaces, index matching would fail.

This PR adds `strings.TrimSpace` when indexing vulnerability policy names and looking up CVE names in `buildCVEExceptionIndex` / `lookup`, along with unit test coverage.

## Related Issue

Fixes #1017

## Type of Change

## Checklist

- [x] Tests added and passing


matthyx (commented):
Verdict: **needs clarification** (reviewed head `b362c103c0f7eec79b09663f4376168ae5ece111` against `main` at `242817252a1cce7be84ea042d16af50673ea93b6`; PR base `6f465e1d` merges cleanly).

**Necessity — no demonstrated production source of padded names.** #1017 was opened by the PR author minutes before this PR and describes the hypothetical, not an observed failure. Tracing the inputs on current `main`:

- CRD-derived policy names are already trimmed at construction: `adapters/v1/securityexception.go:191` (`Name: strings.TrimSpace(vuln.Vulnerability.ID)`) and `:194` for aliases. Nothing padded can reach the index from that source.
- Query-side names come from Grype match IDs (`m.Vulnerability.ID` / `RelatedVulnerabilities[].ID`) via `lookupMatch`, which the pinned Grype DB never pads.
- The only remaining source is the cloud `GetCVEExceptions` payload; I could not verify whether the backend can return padded names, and the PR provides no example payload.

This matches the prior maintainer decisions on #1011, #1000 and #1014: a normalization change without a concrete producer was closed as unnecessary. Please point to a real payload (cloud API response or otherwise) that carries surrounding whitespace, or close this as unneeded.

**Soundness — the change is inconsistent with its own premise (MEDIUM, would need fixing if the premise is accepted).** `lookup` (`backend_utils.go:384`) has no production callers — it is only exercised from `backend_utils_test.go`. The two production query paths are untouched:

- `lookupMatch` (`backend_utils.go:415`) still keys on `strings.ToLower(name)` without trim.
- `cloudCoversPolicy` (`backend.go:466`) still indexes `cloudIndex.byName[strings.ToLower(vp.Name)]` without trim.

Verified on this head with a review-only test: `lookup(" CVE-2021-44228 ")` → 1 match, but `lookupMatch` with the same padded ID → 0, and `cloudCoversPolicy` with a padded CRD name → `false`. So the PR only makes the test helper path whitespace-tolerant while the code that actually applies exceptions keeps the old behavior. If trimming is wanted, normalize once in a shared helper used by `buildCVEExceptionIndex`, `lookup`, `lookupMatch` and `cloudCoversPolicy`, and cover `lookupMatch` in the regression test.

**Validation performed** (Go 1.26, isolated clone, no credentials):
- `go test ./adapters/v1/... -count=1` on the PR head: pass; `go vet ./adapters/v1/...` and `git diff --check`: clean.
- New test case run against `main` code: fails as expected (`padded_policy_name_and_lookup_query_with_whitespace`), so it is a genuine before/after test — but only for the test-only `lookup` path.
- CI on `b362c10`: all checks green (lint, Basic-Test, cross-platform build, CodeQL, DCO).

No duplicate or superseding PR found (searched open/closed/merged PRs and issues for `TrimSpace`, `whitespace`, `trim`, `buildCVEExceptionIndex`, `cveExceptionIndex`); #694/#712/#969 introduced and extended the index but none touched normalization. Search was limited to GitHub search over titles/bodies.

bhuvan-somisetty:
Thanks for checking the exception index paths @matthyx. You're right that CRD-derived policy names are trimmed during construction and `lookup` is test-only while production lookup is untouched. Closing this PR and #1017.

PR #1020, opened 2026-09-17
fix(tools): trim whitespace in image reference strings in ReferenceMatchForms

### LINKED ISSUE #1019: fix(tools): trim whitespace in image reference strings in ReferenceMatchForms
### Description

In `internal/tools/tools.go`, `ReferenceMatchForms` parses container image references to generate match forms (e.g., full ref, ref without digest, bare repo name). If an image reference string contains leading or trailing whitespace, parsing via `reference.ParseNormalizedNamed` fails or yields padded match forms.

### Expected Behavior
`ReferenceMatchForms` should trim whitespace from input `ref` strings before generating match forms.

### Proposed Fix
Add `ref = strings.TrimSpace(ref)` at the start of `ReferenceMatchForms`.

bhuvan-somisetty:
## Description

In `internal/tools/tools.go`, `ReferenceMatchForms` parses container image reference strings to extract match forms (full reference, reference without digest, bare repository name). If the input `ref` contains leading or trailing whitespace, parsing via `reference.ParseNormalizedNamed` fails or returns un-trimmed match forms.

This PR adds `ref = strings.TrimSpace(ref)` to `ReferenceMatchForms` along with unit test coverage.

## Related Issue

Fixes #1019

## Type of Change

## Checklist

- [x] Tests added and passing


matthyx (changes requested):
Reviewed head `b71261c193c6d6b21071f9923db656b1d503a807` against `main` at `2428172` (PR base `6f465e1`; `internal/tools/tools.go` unchanged on `main` since). Verdict: **request changes** — the condition this PR guards against cannot reach `ReferenceMatchForms` on the target branch, so the change is a no-op in production and the fix, if ever needed, belongs at a different layer.

**HIGH — not needed: a whitespace-padded reference is rejected before exception matching ever runs.** `ReferenceMatchForms` has exactly one production caller, `matchImages` (`adapters/v1/securityexception_match.go:133`), which receives `workload.ImageTagNormalized` from `BuildExceptionTarget` (`securityexception_match.go:253`). That value is produced by `tools.NormalizeReference(c.ImageTag)` (`controllers/http.go:564,630`, `core/services/scan.go:238`) and then fed to `names.ImageInfoToSlug` to compute `ImageSlug`. I ran this against `main` in an isolated container with a padded input `"  docker.io/library/nginx:1.25  "`:

```
ParseNormalizedNamed(padded)             err=invalid reference format
NormalizeReference(padded)               = "  docker.io/library/nginx:1.25  "   (returned unchanged)
ImageInfoToSlug(normalized, hash)        slug="" err=Current inputs produce an invalid slug
```

With an empty `ImageSlug`, `ValidateScanCVE` / `ValidateScanRegistry` (`core/services/scan.go:1247,1274`) return `domain.ErrMissingImageInfo` and the scan is rejected before `GetCVEExceptions` → `BuildExceptionTarget` → `matchImages` is reached. `ScanCP` hits the same slug failure at `scan.go:239-246` and skips the container. So there is no path on which `ReferenceMatchForms` observes leading/trailing whitespace; the new test passes on this head but exercises a state the callers cannot produce. #1019 was filed by the PR author one minute before the PR with no reproduction, log, or producer identified, and I found none either (the `ImageTag` originates from the pod's container spec via the operator; a padded image string would also never be pullable by the kubelet).

**MEDIUM — wrong layer even on its own terms.** The PR trims only inside the leaf helper. `NormalizeReference` still returns the padded string, so `ImageTagNormalized`, the slug, the SBOM creation call (`scan.go:1425`) and backend reporting (`adapters/v1/backend.go:289,547`) would all keep the padded value. If a real producer of padded references is ever demonstrated, the trim should happen once at ingress (`websocketScanCommandToScanCommand` / the `ScanCP` conversion), not in a matcher that only sees the value after four other consumers have already failed on it.

Related history: this follows the same pattern as #1024 and #1030 (request changes: condition unreachable on the touched path) and #1000 / #1011 / #1014 (closed by the author after the reported input could not be sourced). #353 is the precedent where trimming *was* accepted — for user-authored SecurityException fields, which are free-form CRD input; the image reference here is not.

Validation (isolated, network-disabled, no credentials): `go test ./internal/tools/ ./adapters/v1/ -count=1` on this head — pass; `go vet ./internal/tools/`, `git diff --check` — pass; the new test case applied alone to `main` fails as expected; the reproduction above run on `main`. CI on the head is green (lint, Basic-Test, cross-platform build, CodeQL, DCO).

Suggested remedy: close this PR and #1019 unless a concrete producer of whitespace-padded image references reaching kubevuln can be shown (a log line, an operator payload, or a Kubernetes object). If one exists, please add the trim at ingress with a test that fails on `main` through the actual scan path.

bhuvan-somisetty:
Thanks for the thorough review @matthyx. You're right that whitespace-padded image references fail slug computation / parsing before reaching `ReferenceMatchForms`. Closing this PR and #1019.

PR #1022, opened 2026-09-17
fix(adapters): standardize error return slice in GetContainerRelevancyScans

### LINKED ISSUE #1021: fix(adapters): standardize error return slice in GetContainerRelevancyScans
### Description

In `adapters/v1/container_profile.go`, `GetContainerRelevancyScans` returns `scans` on some error branches and `nil` on other error branches (e.g. lines 48, 52, 57).

### Expected Behavior
`GetContainerRelevancyScans` should consistently return `scans` (or `nil`) across all error return statements.

### Proposed Fix
Standardize error return paths in `GetContainerRelevancyScans` to return `scans` on all error paths.

bhuvan-somisetty:
## Description

In `adapters/v1/container_profile.go`, `GetContainerRelevancyScans` returned `scans` (which is `nil` initially) on early error paths but explicitly returned `nil, err` on later error paths (e.g., missing instance ID, missing WLID, or instance ID parsing failure).

This PR standardizes the error return paths in `GetContainerRelevancyScans` to consistently return `scans, err`.

## Related Issue

Fixes #1021

## Type of Change

## Checklist

- [x] Tests added and passing


matthyx (commented):
Verdict: **needs clarification** (reviewed head `8157ee2fd748d61bd0c68b2a0dc56aa5eb2f929d` against `main`; PR base `6f465e1d` is an ancestor of current `main` `2428172`, merge is clean, and `main` has not touched either changed file since).

**Necessity — the production change is a behavioral no-op.** In `adapters/v1/container_profile.go:28` `scans` is declared as `var scans []ports.ContainerRelevancyScan` and nothing appends to it before the three branches this PR touches (lines 48/52/57). At each of those returns, `scans` *is* `nil`, so `return scans, err` and `return nil, err` produce identical values for every caller (`services/scan.go` only checks `err`). Fixes #1021, but #1021 was opened by the PR author 3 minutes before the PR and describes a stylistic inconsistency, not a user-visible or program-visible defect.

**Validation — the added test does not discriminate.** I checked out `main` (`2428172`) with only the test-file change from this PR and ran `go test ./adapters/v1/ -run TestGetContainerRelevancyScans -count=1`: `TestGetContainerRelevancyScans_MissingAnnotations` **passes without the production change**, because `assert.Nil(t, scans)` holds on both the old `nil` return and the new `scans` (nil) return. It is therefore not a regression test for this change. On the PR head, `go vet ./adapters/v1/` and the same test run are clean, and all GitHub checks are green — but that only shows nothing broke, not that anything was fixed.

**Prior decisions.** This matches the pattern on #1000, #1011 and #1014, which were closed as redundant once the change was shown to have no observable effect on the pinned code. No other open/closed PR touches `GetContainerRelevancyScans` (searched by symbol, file and "error return"; #466 is the last real change to this function and is unrelated).

**Findings:** no correctness, regression, security or concurrency concerns — the diff cannot change behavior. Nonblocking: if the maintainers do want the three branches aligned purely for consistency, the more idiomatic direction is `return nil, err` on the two earlier branches (lines 32/41), dropping the `scans` variable until it is first populated, rather than spreading the named-but-empty slice further. Either way, the test would need an assertion that actually distinguishes the two forms (there is none, since they are equal) or should be framed as coverage for the missing-annotation error path rather than as a regression test.

**Ask:** is there a concrete caller or scenario where the `nil` vs `scans` return matters? If not, I'd suggest closing this and #1021 as a no-op, or re-scoping the PR as a pure style/consistency change with the test description adjusted accordingly.

bhuvan-somisetty:
Thanks for the review @matthyx. You're right that `scans` is already nil at each of those early error returns so there is no behavioral difference. Closing this PR and #1021 as a no-op.

PR #1024, opened 2026-09-17
fix(adapters): trim whitespace on API groups in matchResources

### LINKED ISSUE #1023: fix(adapters): trim whitespace on API groups in matchResources
### Description

In `adapters/v1/securityexception_match.go`, `matchResources` compares resource `APIGroup` using `strings.EqualFold(r.APIGroup, *target.APIGroup)`. If API group strings contain leading or trailing whitespace, the equality check fails even if the canonical group names match.

### Expected Behavior
`matchResources` should trim whitespace from `r.APIGroup` and `*target.APIGroup` when comparing API group names.

### Proposed Fix
Trim whitespace from `r.APIGroup` and `*target.APIGroup` before checking `strings.EqualFold`.

bhuvan-somisetty:
## Description

In `adapters/v1/securityexception_match.go`, `matchResources` compares `APIGroup` on resource matches and target objects using `strings.EqualFold`. If API group strings carry leading or trailing whitespace, equality comparison fails.

This PR adds `strings.TrimSpace` when comparing API groups in `matchResources` along with unit test coverage.

## Related Issue

Fixes #1023

## Type of Change

## Checklist

- [x] Tests added and passing


matthyx (changes requested):
Reviewed head `19a620b9ce3b67e9e9ad4e80b82dc444c9ed6170` against `main` at `2428172`. Verdict: **request changes** — the reported problem is not reachable on the target branch, and the change quietly widens a suppression match in a path that is documented to fail closed.

**1. Not needed (no evidence of the bug in practice).** #1023 was filed by the PR author 5 minutes before this PR, with no reproduction, user report, or affected manifest. On `main`:
- `target.APIGroup` is set only from `k8sinterface.GetGroupVersionResource(kind).Group` (`adapters/v1/securityexception_match.go`, `BuildExceptionTarget`) — a static kind→GVR map. It can never carry whitespace, so trimming it is a no-op.
- `r.APIGroup` comes from the CRD. A plain YAML scalar (`apiGroup: apps`) is already whitespace-stripped by the parser; the only way to get `"  apps  "` stored is to quote it deliberately. If that input should be tolerated, the right layer is CRD validation (`pattern` on `spec.match.resources[].apiGroup` in `helm-charts/charts/kubescape-operator/crds/security-exception.crd.yaml`) or one-time normalization when the exception is loaded — not the hot comparator.

**2. Fail-open regression (confirmed).** `r.APIGroup != ""` still gates on the *untrimmed* value, so a whitespace-only group (`apiGroup: " "`) passes the guard and then trims to `""`, which `EqualFold`s the core group. I ran `matchResources([]ResourceMatch{{Kind: "Pod", APIGroup: "  "}}, ExceptionTarget{Kind: "pod", Name: "p", APIGroup: strptr("")})`:
- `main`: `false` (never matches — fails closed)
- this PR: `true` (matches every core-group Pod)

That contradicts the comment directly above the line ("fail closed rather than match a resource whose group was never verified"). The added test only covers the padded-`apps` case and does not exercise this.

**3. Inconsistent with the surrounding matcher.** `Kind` and `Name` on the same struct are compared untrimmed, and #1003 (closing #998) deliberately kept `Name` exact. Trimming only `apiGroup` gives `ResourceMatch` three different normalization rules with no stated reason.

**Context.** This is one of five near-identical "trim whitespace" PRs opened the same night (#1018, #1020, #1030, #1032), each backed by a self-filed issue; #1032 already received request-changes for not fixing anything on `main`. Please don't open normalization PRs without a concrete failing input from a real manifest or log.

**Validation.** `go test ./adapters/v1/ -run TestMatchResources` passes on the PR head; the probe above was run on both `main` and the PR head. CI is green but does not cover the whitespace-only case. The branch is based on `6f465e1`, behind `main`, though the touched files merge without conflict.

**Remedy.** Close this PR and #1023. If whitespace tolerance is genuinely wanted, propose it once for the whole `ResourceMatch` (kind/name/apiGroup) at load or CRD-validation time, with the whitespace-only case explicitly failing closed.

bhuvan-somisetty:
Thanks for the detailed analysis @matthyx. You're right that `target.APIGroup` comes from static GVR mapping and `r.APIGroup` from CRD YAML parsing is already stripped, and trimming whitespace-only values in the matcher creates a fail-open condition. Closing this PR and #1023.

PR #1026, opened 2026-09-17
fix(controllers): reset pending count when configuring non-positive MaxQueueDepth in WithMaxQueueDepth

### LINKED ISSUE #1025: fix(controllers): reset pending count when configuring non-positive MaxQueueDepth in WithMaxQueueDepth
### Description

In `controllers/http.go`, `WithMaxQueueDepth(n int)` sets `maxQueueDepth = n`. If `n <= 0` (unbounded), any accumulated value in `h.pending` (from previous queue tracking) is left intact rather than reset.

### Expected Behavior
When configuring non-positive `maxQueueDepth` (`n <= 0`), `WithMaxQueueDepth` should reset `h.pending` to 0.

### Proposed Fix
Add `if n <= 0 { h.pending.Store(0) }` in `WithMaxQueueDepth`.

bhuvan-somisetty:
## Description

In `controllers/http.go`, `WithMaxQueueDepth(n int)` configures the maximum queue depth. When switching `maxQueueDepth` to an unbounded mode (`n <= 0`), any previously accumulated count in `h.pending` remained untouched.

This PR updates `WithMaxQueueDepth` to call `h.pending.Store(0)` when `n <= 0`, along with unit test coverage.

## Related Issue

Fixes #1025

## Type of Change

## Checklist

- [x] Tests added and passing

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed queue admission state when the queue-depth limit is changed to an unbounded setting.
  * Pending admission counts now reset correctly when the maximum queue depth is set to zero or a negative value.


matthyx (changes requested):
Verdict: **request changes** (reviewed head `5dd5641005351b3d8581dee2cafe011bccfefd9c` against `main` at `2d104fd0d49a0282b9422a34746c1db7bc0abb8d`; the PR's own base was `6f465e1d`, no relevant changes in between).

**The reported bug does not exist on `main`, and the fix introduces an accounting regression in the only scenario where it changes behavior.**

### Need (not established)

`pending` is only read or written inside `tryAdmit`/`release`, and both early-return when `maxQueueDepth <= 0` (`controllers/http.go:129`, `:148`). The field's own doc says "Unused when maxQueueDepth is zero". A stale count is therefore inert: a review-only test on `main` that admits 2 under `WithMaxQueueDepth(2)` (pending=2, third admit refused), then switches to `WithMaxQueueDepth(0)`, admits 1000 more without a single rejection. Nothing observable changes, so there is no symptom to fix. #1025 was filed 13 minutes before this PR by the same author with no observed failure, only the proposed patch.

In addition, the only production caller is `cmd/http/main.go:166`, invoked once on a freshly constructed controller before the router serves anything, so `pending` is always 0 when `WithMaxQueueDepth` runs. The "switching modes at runtime" scenario the PR describes has no caller.

### Soundness (HIGH — regression under the PR's own hypothetical)

If runtime reconfiguration *is* assumed (the only way this change is reachable), `Store(0)` discards reservations that in-flight jobs will still release later, because `release()` decrements whenever the depth is bounded at the time it runs. Review-only test, run at `5dd5641`:

```go
h := (&HTTPController{}).WithMaxQueueDepth(1)
require.True(t, h.tryAdmit())   // job A admitted, pending=1
h.WithMaxQueueDepth(0)          // PR: pending reset to 0
h.WithMaxQueueDepth(1)
h.release()                     // job A finishes: pending = -1
require.True(t, h.tryAdmit())
assert.False(t, h.tryAdmit())   // FAILS on PR head: second admit succeeds, effective limit is now 2
```

On `main` this passes (pending returns to exactly 0 and the limit holds). With the PR, the counter drifts one below zero per job that was in flight across the toggle, permanently loosening the bound for the process lifetime. The PR's new test never releases the job it admits, so it does not exercise this path.

### Validation

- `go test ./controllers/... -count=1`, `go vet ./controllers/`, `git diff --check` at `5dd5641`: pass. CI green (lint, Basic-Test, cross-platform build, CodeQL, DCO).
- Both review-only tests above were run in a network-free clone; they are not part of this PR.
- The PR's modified test passes on the PR head and would fail on `main` only on the new `pending == 0` assertion, i.e. it asserts the internal side effect, not a behavioral difference.

### Related history

- #749 (merged) introduced `maxQueueDepth`/`tryAdmit`/`release` with the explicit design that unbounded mode "skips touching pending". This PR runs against that invariant.
- #1000, #1011, #1014 (same author) were closed by the maintainer for the same reason: a change that is a no-op against the pinned code and a test that does not demonstrate a real regression. That precedent applies here.
- No open or closed PR/issue besides #1025 covers this (searched `maxQueueDepth`, `pending`, `tryAdmit`, `WithMaxQueueDepth` across PRs and issues).

### Recommendation

Close this PR and #1025 as not-a-bug. If runtime reconfiguration of `maxQueueDepth` is ever wanted, it needs a concurrency-safe design (the plain `int` write in `WithMaxQueueDepth` races with `tryAdmit`/`release` today) rather than a counter reset.

bhuvan-somisetty:
Thanks for catching the counter drift regression @matthyx. You're right that `pending` is designed to be bypassed when `maxQueueDepth <= 0` per #749 and resetting it during reconfiguration would cause negative counter drift. Closing this PR and #1025.

PR #1028, opened 2026-09-17
fix(repositories): add ContainerProfiles inspector method with read lock to MemoryStore

### LINKED ISSUE #1027: fix(repositories): add ContainerProfiles inspector method with read lock to MemoryStore
### Description

In `repositories/memory.go`, `MemoryStore` provides inspector methods (`CVESummaries`, `CVESummaryStubs`, `SBOMStores`) protected by read locks for tests. However, there is no corresponding `ContainerProfiles` inspector method to retrieve all stored `ContainerProfile` entries safely under a read lock.

### Expected Behavior
`MemoryStore` should provide a `ContainerProfiles()` method returning a slice of all stored container profiles protected by `m.mu.RLock()`.

### Proposed Fix
Add `ContainerProfiles() []v1beta1.ContainerProfile` to `MemoryStore` in `repositories/memory.go`.

bhuvan-somisetty:
## Description

In `repositories/memory.go`, `MemoryStore` provides read-lock-protected inspector methods (`CVESummaries`, `CVESummaryStubs`, `SBOMStores`) for tests. However, there was no method to inspect stored `ContainerProfile` objects safely under a read lock.

This PR adds `ContainerProfiles() []v1beta1.ContainerProfile` to `MemoryStore` protected by `m.mu.RLock()`, along with unit test coverage.

## Related Issue

Fixes #1027

## Type of Change

## Checklist

- [x] Tests added and passing


matthyx (commented):
Verdict: **needs clarification** (reviewed head `7944185eb4e89d830657d6810d17fbbea21439d4` against `main` at `2d104fd0d49a0282b9422a34746c1db7bc0abb8d`, including a clean merge simulation).

**Need is not established.** `ContainerProfiles()` has no caller anywhere in the tree other than the new `TestMemoryStore_ContainerProfiles`, which only exercises the helper itself. The existing inspectors this PR cites as precedent were each added because a `core/services` test needed them: `CVESummaries` came with the #838 fix, `CVESummaryStubs` and `SBOMStores` are consumed by `core/services/scan_test.go`. The four tests that call `StoreContainerProfile` (`scan_test.go`, `scan_docker_fixture_test.go`, `adapters/v1/container_profile_test.go`) seed a profile and then read it back through `GetContainerProfile`, which already runs under `m.mu.RLock()`. Issue #1027 (opened by the same author 8 minutes before this PR) describes the missing method, not a test that is blocked or a race it would resolve. As it stands this is an unused test helper, and the "Bug fix" label does not apply: no behaviour changes.

If there is a specific test you intend to write that needs to enumerate every stored profile (for example asserting that a scan flow stored N profiles), please add that test in this PR so the helper has a consumer; otherwise I would close this as not needed.

**Nonblocking, if it is kept:** `CVESummaries` sorts its output so callers do not depend on Go's randomised map iteration order (see its doc comment). `ContainerProfiles` iterates `m.aps` unsorted, so any future assertion on more than one profile would be flaky. Sorting by `(Namespace, Name)` would keep it consistent with the existing convention.

**Validation:** `go build ./...`, `go vet ./repositories/`, `go test -race -count=1 ./repositories/` all pass on the PR head and on the merge with current `main`. CI on `7944185` is green (lint, Basic-Test, cross-platform build, CodeQL, DCO). The code itself is correct: copy under read lock, no aliasing of the internal map.

Searched open/closed PRs and issues for `MemoryStore`, `ContainerProfiles`, and "inspector"; no prior PR proposes this helper, and #838 (the closest related history) added `CVESummaries` only to support a concrete regression test. Search was keyword-based and may have missed differently worded history.

bhuvan-somisetty:
Thanks for checking @matthyx. You're right that `ContainerProfiles()` has no callers in the codebase today and `GetContainerProfile` already provides thread-safe access under `RLock`. Closing this PR and #1027.

PR #1034, opened 2026-09-17
fix(controllers): add empty jobID lookup guard in scanStatusStore.get

### LINKED ISSUE #1033: fix(controllers): add empty jobID lookup guard in scanStatusStore.get
### Description

In `controllers/scan_status.go`, `scanStatusStore.get(jobID string)` looks up job status by job ID under a read lock. If an empty string `jobID == ""` is passed, it acquires `s.mu.RLock()` to perform map lookup before returning `false`.

### Expected Behavior
`scanStatusStore.get` should return early with `(domain.ScanStatus{}, false)` when `jobID == ""` to avoid lock acquisition and map access.

### Proposed Fix
Add `if jobID == "" { return domain.ScanStatus{}, false }` at the top of `scanStatusStore.get`.

bhuvan-somisetty:
## Description

In `controllers/scan_status.go`, `scanStatusStore.get(jobID string)` acquired a read lock and searched the store map even when given an empty `jobID`.

This PR adds an early return check `if jobID == "" { return domain.ScanStatus{}, false }` in `scanStatusStore.get`, avoiding lock acquisition for blank job IDs, along with unit test coverage in `scan_status_test.go`.

## Related Issue

Fixes #1033

## Type of Change

## Checklist

- [x] Tests added and passing


matthyx (commented):
Verdict: **needs clarification** (reviewed head `8b80d58a6475bd29ee0db861507942bf100df7d5` against `main` at `52272cf8354def03f75f85c90a6f8e0f75c326e8`; merge simulation clean, `go build ./...`, `go vet ./controllers/`, `go test ./controllers/ -count=1` pass).

I could not establish that #1033 describes a reachable or observable problem on the target branch, and the change has no behavioral effect.

**1. `get("")` is unreachable in production.** The only caller is `HTTPController.ScanStatus` (`controllers/http.go:433`), which reads `c.Param("jobID")` from the route `GET /v1/scanStatus/:jobID` (`cmd/http/main.go:231`). A gin `:param` never matches an empty segment: I probed `/v1/scanStatus/` and `/v1/scanStatus` (both 404 before the handler runs) and `/v1/scanStatus//` (301 redirect); the handler is only ever invoked with a non-empty string.

**2. Even if reached, the result is identical.** `recordAccepted` already refuses to store an empty key (`scan_status.go:119`), so `s.items[""]` is always a miss and the base code returns `(ScanStatus{}, false)` after one uncontended `RLock`/`RUnlock` — a few nanoseconds on a path that never runs. The guards on `recordAccepted`/`isActive`/`markRunning`/`markTerminal` are semantic (they prevent writing an empty key); a read has nothing to protect.

**3. The added test does not demonstrate a fix.** `TestScanStatusStore_EmptyJobIDGetReturnsEarly` passes unchanged on the base commit without the production change (`go test ./controllers/ -run TestScanStatusStore_EmptyJobIDGetReturnsEarly` on `main` → PASS), and the "without lock contention" assertion in its message is not something the test observes.

Related history: #791 (merged, same author) is what made `get` a single-key read and is the code this touches; #857 (merged) added the `jobID == ""` guards on the write paths. Nothing in either discussion, nor in #636/#790, calls for a read-side guard. This also matches the pattern in #1000 and #1011, which were closed once the reported condition turned out to have no producer.

If there is a concrete caller or input that reaches `get` with an empty ID and behaves incorrectly, please describe it and add a test that fails on `main`; otherwise this looks like a no-op and #1033 should be closed as not-a-bug rather than fixed.

bhuvan-somisetty:
Thanks for the explanation @matthyx. You're right that `c.Param("jobID")` cannot produce an empty string via Gin routing and `get("")` is unreachable in production while returning cleanly anyway. Closing this PR and #1033.

PR #1039, opened 2026-09-18
feat(vex): add OpenVEX parsing and validation

aaa-aashna:
The PR adds the initial OpenVEX document parsing and validation support for Kubescape’s external VEX ingestion work.

### Changes

* Added OpenVEX document models:

  * `Document`
  * `Statement`
  * `Vulnerability`
  * `Product`
* Added parsing support for JSON VEX documents.
* Added document validation for:

  * Required fields
  * Valid context and document ID URLs
  * Author and timestamp
  * Document version
  * Statements
  * Vulnerability and product identifiers
  * Supported vulnerability statuses
* Added unit tests covering valid and invalid documents.

This PR focuses only on the document parsing and validation foundation.
**Vulnerability matching and product matching logic will be implemented in a follow-up PR.**

## Summary by CodeRabbit

* **New Features**
  * Added support for parsing and validating OpenVEX documents.
  * Supports document metadata, vulnerability details, affected products, status values, and action or impact statements.
  * Validates required fields, URLs, timestamps, versions, statements, products, and supported statuses.


internal/vex/document.go
```diff
@@ -0,0 +1,147 @@
...
+	ActionStatement string        `json:"action_statement,omitempty"`
+}
+
+type Vulnerability struct {
+	ID      string   `json:"id"`
+	Aliases []string `json:"aliases,omitempty"`
+}
+
+type Product struct {
+	ID string `json:"id"`
```
matthyx, line 42: **Blocker – wrong JSON field names.** OpenVEX uses `@id` for the document and product identifiers and `name` for the vulnerability (go-vex: `Document.ID \`json:"@id"\``, `Vulnerability.Name \`json:"name"\``, `Component.ID \`json:"@id"\``). With the tags here, a conforming document decodes with empty `ID`s and `Validate` fails with `VEX document id is required`. The tests pass only because the fixtures use the same non-spec field names.

Rather than fixing the tags, please reuse `github.com/openvex/go-vex/pkg/vex` (`vex.Parse`) and `internal/vexvalidate.Validate` (#880) — that is what `main` already uses to both write and validate OpenVEX.

internal/vex/document.go
```diff
@@ -0,0 +1,147 @@
...
+			return fmt.Errorf("product id is required at index %d", index)
+		}
+	}
+
+	switch statement.Status {
+	case StatusNotAffected,
+		StatusAffected,
+		StatusFixed,
+		StatusUnderInvestigation:
+		return nil
```
matthyx, line 133: **Blocker – status-specific rules are missing.** Per spec, `affected` requires `action_statement` and `not_affected` requires `justification` or `impact_statement` (and the justification must be from the catalog). This accepts both without them; `vexvalidate` on `main` rejects them via go-vex's `Statement.Validate()`. Repro on a merge with `main`:

```
PR Parse("affected" w/o action_statement) -> <nil>
vexvalidate.Validate(same) -> statement 0: action statement must be set when using status "affected"
```

internal/vex/document.go
```diff
@@ -0,0 +1,147 @@
...
+		return fmt.Errorf("VEX document context is required")
+	}
+
+	if !isValidURL(document.Context) {
+		return fmt.Errorf("VEX document context must be a valid URL")
+	}
+
+	if strings.TrimSpace(document.ID) == "" {
+		return fmt.Errorf("VEX document id is required")
+	}
```
matthyx, line 78: Non-blocking on its own (moot if the package is dropped): `@id` is an IRI, not a URL — `urn:uuid:…` is valid per spec but `isValidURL` requires a host, so it is rejected here. The `@context` check is also looser than `vexvalidate.hasValidContext`, which pins the `https://openvex.dev/ns` namespace and rejects look-alikes.

matthyx (changes requested):
Thanks for picking up the ingestion foundation. I reviewed `c563de8` against `main` (`08d5be7`). Two things block this as-is, and both come down to the branch being cut from an early-August base (merge-base `4104c0d`, 247 commits behind `main`), before the VEX groundwork that has landed since.

**1. This duplicates `internal/vexvalidate` (#880, merged 2026-08-28, resolving #878).** That package already does exactly what this PR describes — parse a fetched OpenVEX document and reject one that is not genuinely OpenVEX — and it does it by reusing `github.com/openvex/go-vex` (already in `go.mod`, and the same library kubevuln uses to *write* its own VEX in `repositories/apiserver.go`). It delegates statement rules to go-vex's `Statement.Validate()`, which was an explicit maintainer requirement in that PR's review ("reuse go-vex's own rules rather than re-implement a subset"). Adding a second hand-rolled model/validator under `internal/vex` gives the repo two sources of truth for what a valid document is, and they already disagree (see 2). The other ingestion primitives are also in place: `internal/vexdoc` (temp-file handoff to grype), `internal/vexbatch`, `internal/csafresolve`, `internal/safefetch`.

**2. The model does not match the OpenVEX spec, so it rejects conforming documents and accepts non-conforming ones.** I ran both validators against the same inputs on a merge of this branch onto `main`:

```
PR Parse(conformant OpenVEX: @id / vulnerability.name / product @id) -> "VEX document id is required"
vexvalidate.Validate(same document)                                  -> <nil>
go-vex parses it fine: id="https://kubescape.io/vex/1" vuln="CVE-2024-0001" product="pkg:oci/example@sha256:abc"

PR Parse("affected" with no action_statement, "not_affected" with no justification) -> <nil>
vexvalidate.Validate(same)      -> statement 0: action statement must be set when using status "affected"

PR Validate(@id = "urn:uuid:...")  -> "VEX document id must be a valid URL"   (URNs are valid IRIs per spec)
```

A real Chainguard/Wolfi feed — or kubevuln's own generated document — would be rejected at the first check. CodeRabbit's three inline findings point at the same root cause; I have not repeated them, just added the evidence above.

**Suggested path:** rebase onto current `main`, drop `internal/vex/document.go` in favour of `vexvalidate.Validate` + `go-vex`'s `vex.Parse`, and put the week-2 effort into the piece that is genuinely missing — nothing yet *calls* `vexvalidate`/`vexdoc` from a fetch-and-persist path (see #387 for the target design, and #591's review for what a reconciler needs to get right). If there is a specific gap in `vexvalidate` you found, an issue or a small PR against that package would be the right shape.

**Validation:** `go build ./...`, `go vet ./internal/vex/`, `go test ./internal/vex/` all pass on the branch and on a test-merge onto `main` (17/17 tests). The `pr-created / test` CI failure (`Test_grypeAdapter_Ready_recoversFromStuckWarmUpdate` in `adapters/v1`) is unrelated to these files; it is likely a flake on the stale base and should be re-checked after rebase.

**Verdict:** request changes — superseded by #880 and not spec-conformant.

PR #1042, opened 2026-09-19
fix(adapters): log skipped suppression events

### LINKED ISSUE #708: emitSuppressionEvent silently skips K8s Event when SecurityException UID is missing, with no diagnostic log
# Description
`emitSuppressionEvent` in `adapters/v1/securityexception.go` (around lines 464-484) silently returns without emitting a Kubernetes Event whenever the source `SecurityException`/`ClusterSecurityException` policy is missing `kind`, `Name`, or `uid`. This can happen for objects that reach the function through certain client paths (fake clients in tests, or a watch cache that has not yet populated `UID`).

The problem is that this skip is completely silent. `logSuppression` (line 453) always logs the suppression itself regardless of whether the Event was emitted, so there is no way for an operator to tell "the CVE was suppressed and the Event fired" apart from "the CVE was suppressed but the Event was skipped because provenance was incomplete." Debugging "why don't I see a suppression Event on my SecurityException" currently has no diagnostic trail to follow.

# Environment
OS: ` ` 
Version: main branch, `adapters/v1/securityexception.go`

# Steps To Reproduce
1. Run `go test ./adapters/v1/... -run 'TestApplySecurityExceptions_EmitsEventWhenRecorderConfigured|TestApplySecurityExceptions_NilRecorderIsNoOp' -v` - both existing tests pass, but neither one exercises "recorder configured, UID missing."
2. Read `adapters/v1/securityexception.go:464-484` (function `emitSuppressionEvent`): the early return guard `if kind == "" || p.Name == "" || uid == "" { return }` has no accompanying log call.
3. Construct a `SecurityException` whose backing `VulnerabilityExceptionPolicy` has `sourceKind` and `Name` set but no `sourceUID` attribute (this happens for objects lacking a populated `UID`, e.g. objects from certain client paths), call `ApplySecurityExceptions` with a configured recorder, and observe that no Event is recorded and there is no log line indicating why.

# Expected behavior
When `emitSuppressionEvent` skips emitting an Event because of missing provenance (`kind`, `Name`, or `uid`), that should be observable, at minimum via a debug-level log line, so operators/maintainers can distinguish this from "recorder not configured at all."

# Actual Behavior
No log or metric distinguishes this skip from any other code path. The only log line (`logSuppression`'s "CVE suppressed by security exception" at line 453) fires unconditionally and gives no signal about the Event outcome.

# Additional context
Suggested fix: add a debug-level log when `emitSuppressionEvent` returns early due to a missing provenance field, without changing behavior on the happy path where all fields are populated.

Acceptance criteria:
- A new test covers "recorder configured + missing UID" and asserts a debug log fires.
- No behavior change to the happy path where all fields are populated.

ANAMASGARD:
## Overview

When a SecurityException suppression Event cannot be emitted because its provenance is incomplete, kubevuln now writes a debug log identifying the missing fields.

This preserves the existing suppression behavior, keeps the nil-recorder path silent, and leaves successful Event emission unchanged.

## Additional Information

The diagnostic reports missing `sourceKind`, `sourceName`, and/or `sourceUID` fields. This helps distinguish an intentionally unavailable Event from a suppression that successfully emitted a Kubernetes Event.

## How to Test

Run the focused Event tests:

```bash
go test ./adapters/v1/... \
  -run 'TestApplySecurityExceptions_.*Event' \
  -count=1 -v
```

Expected results:

- Suppression still moves the CVE to `IgnoredMatches`.
- A populated UID still produces a Kubernetes Event.
- A missing UID produces no Event.
- The missing-UID case logs `skipping suppression event` and `sourceUID`.

Validation performed:

- Focused Event tests pass.
- `git diff --check` passes.
- The broader package suite is environment-blocked by IPv6 loopback binding in an unrelated `httptest` test.

## Related issues/PRs

* Resolved #708

- [x] New and focused existing unit tests pass locally with my changes
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved diagnostic logging when security exception event provenance is incomplete.
  * Suppression events with missing provenance continue to be skipped, with logs now identifying the missing fields.

* **Tests**
  * Added coverage verifying that incomplete provenance prevents event emission and records the relevant diagnostic details.


matthyx (approved):
Reviewed `444ccf8` against `main` (`34b4edd`). Approving.

**Necessity** — Confirmed on `main`: `emitSuppressionEvent` returns silently when `sourceKind`/`Name`/`sourceUID` is missing, while `logSuppression` still logs "CVE suppressed", so the two outcomes are indistinguishable (#708). Prior attempt #732 was self-closed by its author without maintainer feedback, not rejected; it also didn't assert the log line. This PR covers the acceptance criteria #732 skipped.

**Correctness** — The guard's semantics are unchanged (same three fields, same early return); only a `Debug` line is added. `strings` was already imported. No behavior change on the happy or nil-recorder paths.

**Validation (local, go1.26)**
- `go build ./... && go vet ./adapters/v1/` — clean
- `go test ./adapters/v1/ -run TestApplySecurityExceptions_ -race -count=5` — pass
- `go test ./adapters/v1/ -count=1` (whole package) — pass in 18.6s; the IPv6 `httptest` issue mentioned in the description didn't reproduce here
- Regression property: with `securityexception.go` reverted to `main`, the new test fails on `does not contain "sourceUID"`; passes with the fix
- `gofmt`/`golangci-lint` — no new findings in the changed files; CI `lint` is green
- The stderr-swap log capture matches the existing pattern in `adapters/v1/backend_test.go` (`TestPostResults…`), and no test in the package uses `t.Parallel()`, so mutating the global logger/`os.Stderr` is safe here

**Nonblocking suggestion** — The debug line carries only `missingFields`. Adding `cve` (and `sourceName`/`sourceKind` when non-empty) would let the line stand alone when several scans interleave, instead of relying on adjacency to the preceding Info log. Fine to leave as-is or do in a follow-up.

Issue #938, opened 2026-09-04
StoreSBOM's unconditional Spec overwrite destroys a real SBOM when a later store carries no document (TooLarge)


bhuvan-somisetty:
# Description

`StoreSBOM` (`repositories/apiserver.go:2248-2298`) does a wholesale `existing.Spec = manifest.Spec` overwrite on every update, with no guard for the case where the incoming SBOM carries no document. When it is called a second time for a `name` that already has a real, fully-generated SBOM stored, but this time with `sbom.Content == nil` (the `TooLarge` verdict), it silently destroys the previously stored Syft document and replaces it with an empty one, keeping only the `TooLarge` status annotation.

This is a real, unguarded write path, not a hypothetical: `core/services/scan.go:1419-1420` calls exactly this shape --- `StoreSBOM(workerCtx, sbom, false)` with `sbom.Content == nil` --- specifically when `sbom.Status == helpersv1.TooLarge`:

```go
// core/services/scan.go:1409-1420
// TooLarge is stored despite having no document. StoreSBOM keeps it as a status-only
// marker, and getSBOM re-checks that marker against the current maxImageSize and
// maxSBOMSize, so it stops applying once the limits are raised. Persisting it is what
// saves pulling a known-oversized image again to reach the answer already on record.
...
var storeErr error
if s.storage && (sbom.Content != nil || sbom.Status == helpersv1.TooLarge) {
    if storeErr = s.sbomRepository.StoreSBOM(workerCtx, sbom, false); storeErr != nil {
```

That comment's own design intent -- "keeps it as a status-only marker" -- is correct for the case where nothing was ever stored for `name` before. It does not account for `name` already holding a real document.

# Root Cause

`StoreSBOM` builds the manifest's `Spec.Syft` from `sbom.Content` only when it is non-nil, otherwise leaving it at its zero value:

```go
// repositories/apiserver.go:2276-2278
if sbom.Content != nil {
    manifest.Spec.Syft = *sbom.Content
}
```

It then hands this manifest to `createOrUpdate` with a merge closure that replaces `existing.Spec` outright, regardless of whether the new `Spec.Syft` is real or the zero value from the branch above:

```go
// repositories/apiserver.go:2292-2297
return createOrUpdate(ctx, a.StorageClient.SBOMSyfts(a.Namespace), "SBOM", sbom.Name, &manifest,
    func(existing *v1beta1.SBOMSyft) {
[...]

PR #1043, opened 2026-09-19
fix: prevent stale scans from overwriting newer CVE summaries

### LINKED ISSUE #1038: StoreCVESummary's unconditional Spec overwrite has no freshness check, letting a stale scan clobber a current VulnerabilityManifestSummary
## Summary

`StoreCVESummary`'s update path unconditionally overwrites `existing.Spec` with the incoming scan result, with no check for whether the incoming result is actually newer than what's already stored. Combined with kubescape/operator#417 (operator restarts re-queue scans for orphaned `ContainerProfile` objects from long-dead ReplicaSet generations), a stale scan result can silently clobber a correct, current `VulnerabilityManifestSummary` -- reporting vulnerability data for an image version that isn't running anywhere on the cluster.

## The missing guard

`repositories/apiserver.go`, `StoreCVESummary`:

```go
return createOrUpdate(ctx, a.StorageClient.VulnerabilityManifestSummaries(workloadNamespace),
    "CVE summary manifest", manifest.Name, &manifest, metav1.GetOptions{ResourceVersion: resourceVersionMetadata},
    func(existing *v1beta1.VulnerabilityManifestSummary) {
        existing.Annotations = mergeMaps(existing.Annotations, manifest.Annotations)
        existing.Labels = mergeMaps(existing.Labels, manifest.Labels)
        existing.Spec = manifest.Spec   // <- unconditional, no comparison against `existing`
    },
    helpers.String("relevant", strconv.FormatBool(withRelevancy)))
```

The data needed for a freshness check is already collected and already written -- it's just never read back before this overwrite. `enrichSummaryManifestObjectAnnotations` (same file, line ~1053) sets a scan timestamp on every store:

```go
enrichedAnnotations[timestampMetadataKey] = strconv.FormatInt(timestamp, 10)
```

So both `existing.Annotations[timestampMetadataKey]` (when the currently-stored summary was scanned) and `manifest.Annotations[timestampMetadataKey]` (when the incoming scan ran) are available in the merge callback, but neither is consulted. A minimal fix:

```go
func(existing *v1beta1.VulnerabilityManifestSummary) {
    if existingTS, ok := existing.Annotations[timestampMetadataKey]; ok {
        if newTS, ok2 := manifest.Annotations[timestampMetadataKey]; ok2 && newTS < existingTS {
            return // stale scan result; don't clobber a newer one
        }
    }
    existing.Annotations = mergeMaps(existing.Annotations, manifest.Annotations)
    existing.Labels = mergeMaps(existing.Labels, manifest.Labels)
    existing.Spec = manifest.Spec
}
```

(Timestamp is a reasonable proxy here since kubescape/operator#417's trigger produces genuinely older scans arriving after newer ones; a ReplicaSet-generation or image-digest comparison would be a stronger guard if that data is more directly available in this context, since two scans could in principle share a timestamp granularity collision.)

## Why this matters independent of #417

This is the same *shape* of bug already fixed elsewhere in this codebase for a sibling method:

- #768 (closed): `GetSBOM`'s freshness check only discarded a stored SBOM when it was *older* than requested, never when *newer* -- a one-directional check where the intended guard existed but didn't cover both directions.
- #938 (closed): `StoreSBOM`'s unconditional `Spec` overwrite let a later store with no document erase a real, previously-stored SBOM.

`StoreCVESummary` is the same pattern as #938 one level up the stack: an unconditional `Spec` overwrite with no check at all (not even a one-directional one), on the object that's actually surfaced to users via the Kubescape dashboard/Headlamp plugin. It is directly reachable via #417, but would also be reachable via any other path that can produce two `StoreCVESummary` calls for the same workload out of chronological order (retries, concurrent scan triggers, etc.) -- #417 is just the concrete reproduction we have.

## Evidence of real-world impact (from #417)

On a live cluster, this combination produced a `VulnerabilityManifestSummary` reporting `alertmanager:v0.32.2` (severities included) when the cluster was actually running `v0.34.0` -- a stale scan from a ReplicaSet retired months earlier had overwritten the correct, already-stored current result.

## Suggested fix

Add the freshness/generation check described above to `StoreCVESummary`'s merge callback before the `existing.Spec = manifest.Spec` line, mirroring the guard already added for `GetSBOM` in #768.

## Related

- kubescape/operator#417 -- the concrete trigger (orphaned `ContainerProfile` relist on every operator restart)
- #768 -- same missing-freshness-check shape, in `GetSBOM`
- #938 -- same unconditional-overwrite shape, in `StoreSBOM`

## Environment

- kubevuln v0.3.441 (confirmed current release at time of testing)
- kubescape-operator chart 1.40.4

ANAMASGARD:
## Overview

`StoreCVESummary` previously overwrote an existing summary even when the incoming scan started earlier. It now compares scan timestamps on every update attempt and leaves a newer stored summary untouched, including its Spec, metadata, and resource version.

Equal timestamps retain the existing update behavior. Summaries with a missing or invalid stored timestamp can still be updated.

## How to Test

```bash
go test ./repositories -run 'TestAPIServerStore_StoreCVESummary_(ScanOrder|ConflictRechecksTimestamp)' -count=1
go test -race ./repositories -run 'TestAPIServerStore_StoreCVESummary_(ScanOrder|ConflictRechecksTimestamp)' -count=1
go vet ./...
go build ./...
```

The focused tests, race tests, vet, and build passed locally. The full test suite could not complete in the sandbox because some tests require loopback sockets and linking exhausted `/tmp`.

## Related issues/PRs

* Resolved #1038
* Related: kubescape/operator#417

- [x] If it is a core feature, I have added thorough tests
```

Open the PR against `dev`.

### CODE REVIEW

File: repositories/apiserver.go [hunk 1]
```diff
@@ -1233,14 +1234,38 @@ func (a *APIServerStore) StoreCVESummary(ctx context.Context, cve domain.CVEMani
...
+				return err
+			}
+			// The metadata-only read includes the scan timestamp and resource version.
+			// A stale result must return before Update: the fetched Spec is empty.
+			existing, getErr := store.Get(ctx, manifest.Name, metav1.GetOptions{ResourceVersion: resourceVersionMetadata})
+			if getErr != nil {
+				return getErr
+			}
+			if existingTimestamp, parseErr := strconv.ParseInt(existing.Annotations[timestampMetadataKey], 10, 64); parseErr == nil && existingTimestamp > incomingTimestamp {
+				return nil
```
matthyx, line 1252: `incomingTimestamp` (and the stored one) is `time.Now().Unix()` at scan-processing start (`core/services/scan.go:891`), so this comparison only protects against a scan that *started* earlier finishing later. In the kubescape/operator#417 flow the orphaned-profile scan is dispatched after the operator restart and therefore always carries the newer timestamp — it still overwrites the correct summary. That is fine as a narrow guard, but please (a) don't mark #1038 as resolved by this PR, and (b) log the skip here (name, existing/incoming timestamp, `relevant`) instead of returning silently — `createOrUpdate`'s debug lines were also dropped by inlining it.

matthyx (changes requested):
Reviewed `6cdf7cb` against `main` (`8f3b0e7`).

**What I verified**
- `go build ./...`, `go vet`, `go test -race ./repositories/... ./core/services/...` all pass on the PR head; `git diff --check` clean; CI green.
- The two new tests fail on `main` with only the test file applied and pass with the fix, so they are real regression tests.
- The conflict path is correct: after a 409 the retry re-reads via `ResourceVersion: metadata` (annotations are populated), so a summary that became newer mid-retry is left alone.

**Blocker – the guard does not cover the scenario #1038 / kubescape/operator#417 describe, so `Resolved #1038` would auto-close an issue that stays reproducible.** `kubescape.io/timestamp` is `time.Now().Unix()` taken when kubevuln *starts processing* the scan (`core/services/scan.go:891`, `addTimestamp` in `ScanCP`/`ScanCVE`/`GenerateSBOM`), not the age of the ContainerProfile/ReplicaSet generation. In the #417 flow the operator relists an orphaned profile after a restart and kubevuln scans it *later* than the correct scan, so the stale result carries the **newer** timestamp and this guard lets it overwrite the current summary exactly as before. What this PR does prevent is a scan that started earlier finishing after one that started later — a real but much narrower race. The real fix for #417 is kubescape/operator#419 (skip orphaned profiles). Please change the PR/commit wording to something like `Related: #1038` (or “partially addresses”) and describe the actual guarantee, so #1038 stays open until the operator side lands or a generation/instance-aware check is added.

**Should fix (non-blocking but strongly recommended)**
- A discarded scan result is completely silent now: the skip returns `nil` without a log, and by bypassing `createOrUpdate` the existing `stored/updated/failed … CVE summary manifest` debug lines are lost too. Please log the skip (name, existing vs incoming timestamp, `relevant`) — otherwise a clock-skewed second kubevuln replica dropping results is undebuggable.
- Consider keeping `createOrUpdate` instead of duplicating its retry loop, e.g. let `merge` return `bool` (`false` = skip update). Smaller diff, keeps the logging, and future fixes to the retry path apply to both callers.

**Nits**
- PR body has a stray closing ``` and “Open the PR against `dev`” at the end (there is no `dev` branch; `main` is correct).
- The unchecked “unit tests pass locally” box: the full `repositories` and `core/services` suites pass for me with `-race`, so you can tick it.

No other findings; searched open/closed PRs and issues for summary/timestamp/freshness/`StoreCVESummary` — #945/#946 (cache-hit ref rewrite), #476, #938, #768 are adjacent but none implement or reject this guard, and there is no duplicate PR.

PR #1045, opened 2026-09-19
fix(repositories): gracefully handle custom workload kinds in enrichSummaryManifestObjectLabels

### LINKED ISSUE #1041: bug(repositories): StoreCVESummary fails for custom workload kinds due to unhandled GetGroupVersionResource error in enrichSummaryManifestObjectLabels
# Description

`StoreCVESummary` fails to store a `VulnerabilityManifestSummary` for custom or CRD-based workloads (e.g. Argo `Rollout`, OpenShift `DeploymentConfig`, Argo `Workflow`, or custom controllers) because `enrichSummaryManifestObjectLabels` in `repositories/apiserver.go` returns an unhandled error when `k8sinterface.GetGroupVersionResource(workloadKind)` fails to resolve non-standard workload kinds.

# Environment
OS: `Linux / Kubernetes`
Version: `v0.3.441` (current `main` commit `34b4edd`)

# Steps To Reproduce

1. Scan a workload belonging to a custom workload kind (e.g., WLID `wlid://workload/rollout/default/my-rollout`).
2. `ScanCP`, `ScanCVE`, or `ScanRegistry` finishes matching CVEs and calls `StoreCVESummary`.
3. `enrichSummaryManifestObjectLabels` calls `k8sinterface.GetGroupVersionResource("rollout")`.
4. `k8sinterface.GetGroupVersionResource` fails because `"rollout"` is not in its static table of built-in Kubernetes kinds, returning an error.
5. `enrichSummaryManifestObjectLabels` returns `nil, err`, causing `StoreCVESummary` to abort without persisting the `VulnerabilityManifestSummary`.

# Expected behavior

When scanning custom or CRD-based workloads:
1. `enrichSummaryManifestObjectLabels` should populate `RelatedKindMetadataKey`, `RelatedNameMetadataKey`, and `RelatedNamespaceMetadataKey` directly from the WLID without failing.
2. If `k8sinterface.GetGroupVersionResource(workloadKind)` cannot resolve an API Group and Version for a custom kind (matching the graceful error handling pattern used in `adapters/v1/backend.go:729` and `adapters/v1/securityexception_match.go:592`), it should omit or default the `ApiGroupMetadataKey`/`ApiVersionMetadataKey` labels rather than returning a fatal error.
3. `StoreCVESummary` should successfully persist the `VulnerabilityManifestSummary`.

# Actual Behavior

`enrichSummaryManifestObjectLabels` in `repositories/apiserver.go` immediately propagates the error:
```go
workloadKind := wlid.GetKindFromWlid(workload.Wlid)
if workloadKind != "" {
    groupVersionScheme, err := k8sinterface.GetGroupVersionResource(workloadKind)
    if err != nil {
        return nil, err
    }
...
```
This causes `StoreCVESummary` to return the error and completely skip creating the `VulnerabilityManifestSummary` object for the workload.

# Additional context

Both `adapters/v1/backend.go` and `adapters/v1/securityexception_match.go` guard calls to `k8sinterface.GetGroupVersionResource` with `if gvr, err := k8sinterface.GetGroupVersionResource(kind); err == nil` precisely because non-builtin workload kinds can be scanned by Kubescape.

bhuvan-somisetty:
## Overview
When generating workload summary labels in `enrichSummaryManifestObjectLabels`, `k8sinterface.GetGroupVersionResource(workloadKind)` is called to retrieve the API group and version for the workload. For custom workloads / CRDs (such as `rollout`, `workflow`, etc.) that are not part of the standard builtin Kubernetes scheme, `GetGroupVersionResource` returns an error, which previously caused `enrichSummaryManifestObjectLabels` to fail and abort `StoreCVESummary`.

This change gracefully handles custom workload kinds by populating `ApiGroup` and `ApiVersion` labels when available from the scheme, while continuing to enrich `RelatedKind`, `RelatedName`, `RelatedNamespace`, and `ContainerName` labels without erroring.

## Related issues/PRs:
* Resolves #1041


matthyx (approved):
Reviewed head `b4ff8bb` against `main` (`d6a9c74`).

**Need:** Confirmed. kubevuln never calls `k8sinterface.InitializeMapResources(discoveryClient)` (`NewAPIServerStorage` only builds typed/dynamic clients), so `GetGroupVersionResource` always falls back to the static built-in mock table and any CRD kind in a WLID (`rollout`, `workflow`, `deploymentconfig`, …) made `StoreCVESummary` abort. Resolves #1041; closest prior work is #698 (empty kind), which didn't cover unknown kinds.

**Soundness:** The change mirrors the existing guard pattern at `adapters/v1/backend.go:736` and `adapters/v1/securityexception_match.go:260`. `related-kind/name/namespace` still come straight from the WLID; group/version are omitted (and any caller-supplied `cve.Labels` values for them are now preserved instead of overwritten). Nothing in kubevuln or `kubescape/storage` reads the api-group/api-version labels, so omitting them is safe.

**Validation (local):** `go build ./...`, `go vet ./repositories/`, `go test ./repositories/` pass on the PR head. The new `rollout` table case fails on `main` (`enrichSummaryManifestObjectLabels` returns the GVR error) and passes with the fix. CI is green for `b4ff8bb`.

**Nonblocking:** the test uses `k8sResourceGroup == ""` as the marker for "label must be absent", which would misfire if a core-group kind (e.g. `pod`, group `""`) is ever added to the table. An explicit `expectGVRLabels bool` field would be clearer — fine as a follow-up.

No blockers. Approving.

PR #1057, opened 2026-09-19
fix(adapters): treat non-200 2xx HTTP response status codes as successful in httpPostWithContext

### LINKED ISSUE #1047: bug(adapters): httpPostWithContext treats non-200 2xx responses as retryable errors
# Description
In `adapters/v1/backend.go`, `httpPostWithContext()` sends HTTP POST requests to upstream services and handles retries.

At line 169, it checks whether the status code equals `200 OK`:

```go
resp, err := httpClient.Do(req)
if err != nil {
    return nil, err
}
if resp.StatusCode != http.StatusOK {
    defer resp.Body.Close()
    bodyBytes, _ := io.ReadAll(io.LimitReader(resp.Body, 512))
    bodyStr := strings.TrimSpace(string(bodyBytes))
    var retryErr error
    if bodyStr != "" {
        retryErr = fmt.Errorf("received status code: %d, body: %s", resp.StatusCode, bodyStr)
    } else {
        retryErr = fmt.Errorf("received status code: %d", resp.StatusCode)
    }
    if !shouldRetryReport(resp) {
    }
    ...
    return nil, retryErr
}
return resp, nil
```

If the upstream receiver responds with another valid 2xx successful status code (such as `201 Created` or `202 Accepted`), `resp.StatusCode != http.StatusOK` evaluates to `true`. Furthermore, `shouldRetryReport(resp)` treats non-4xx responses as retryable, which causes `httpPostWithContext` to loop retrying until the maximum elapsed time (60s) expires and ultimately fails the scan dispatch.

In contrast, `postResults` in `adapters/v1/backend_utils.go` (line 151) uses the standard 2xx status range check:
```go
if resp.StatusCode < 200 || resp.StatusCode >= 300 {
    ...
}
```

# Environment
OS: `Linux / Windows`
Version: `main`

# Steps To Reproduce
1. Configure an HTTP server that answers POST requests with `201 Created` or `202 Accepted`.
2. Dispatch a report using `httpPostWithContext`.
3. Observe that `httpPostWithContext` treats `201 Created` as an error and retries until timeout.

# Expected behavior
Any 2xx HTTP response status code (`200` to `299`) should be treated as successful completion without retrying.

# Actual Behavior
`httpPostWithContext` only accepts `200 OK` and loops retrying on other 2xx responses (`201`, `202`, etc.) until timeout.

bhuvan-somisetty:
## Overview
In `adapters/v1/backend.go`, `httpPostWithContext()` previously checked `if resp.StatusCode != http.StatusOK`, treating non-200 2xx responses (`201 Created`, `202 Accepted`, `204 No Content`) as errors and repeatedly retrying until timeout.

This change updates the status code check to accept all valid 2xx status codes (`resp.StatusCode < 200 || resp.StatusCode >= 300`), aligning with `postResults` in `adapters/v1/backend_utils.go` and preventing unwanted retry loops.

## Related issues/PRs:
* Resolves #1047


mohityadav8:
Hey @bhuvan-somisetty, could you please fix the failing lint check first? Once it’s resolved, we can go ahead with the review. Thanks! 🙂

bhuvan-somisetty:
Hey @mohityadav8, I have fixed the staticcheck nil check warning in the unit test and pushed the update. The lint check should be green now. Thanks!

matthyx (approved):
Reviewed at head `be61c99` against `main` (`d6a9c74`).

**Necessity:** Confirmed on `main`. `httpPostWithContext` only accepts `200`, and `shouldRetryReport` treats any non-4xx as retryable, so a `201`/`202`/`204` from the receiver spins the retry loop for the full `maxElapsedTime` (60s in `postResults`, 30s in `reportScanFailure`) before failing. Both callers already tolerate the full 2xx range downstream (`backend_utils.go:151` via #1001, `backend.go:593`), so this is the only place that still pins 200. Fixes #1047.

**Soundness:** One-line range check; no impact on the 4xx/429/`Retry-After` paths, which are only reached for non-2xx. Callers that read the body (`postResults`) handle an empty body fine, so `204` is safe. No public API / config change.

**Validation (local):**
- `go build ./... && go vet ./adapters/v1/` — clean
- `go test ./adapters/v1/ -count=1` on the PR head — pass (21s)
- New `TestHttpPostWithContext_Accepts2xxResponses` applied to `main` without the fix — fails for the 201/202/204 subtests, passes on the PR head, so it is a real regression test.
- CI on `be61c99`: lint, Basic-Test, cross-platform build, CodeQL all green; the earlier lint failure (SA5011) was fixed in the second commit as requested by @mohityadav8.

**Prior history:** #1001 (merged) fixed the same class of bug in `postResults`; #1010 tightened `shouldRetryReport`. Neither touched this check, so nothing here is superseded. No prior rejection of this approach found.

No blockers. Nonblocking nit: the second commit message refers to `Test_httpPostWithContext_Accepts2xxStatusCodes`, but the test is named `TestHttpPostWithContext_Accepts2xxResponses` — cosmetic only.

PR #1059, opened 2026-09-19
fix(domain): guard ScanError.Error against nil Err dereference

### LINKED ISSUE #1049: bug(domain): ScanError.Error() panics with nil pointer dereference when Err is nil
# Description
In `core/domain/scan.go`, `ScanError` is defined to wrap an error together with its classified failure reason code:

```go
type ScanError struct {
    Reason string
    Err    error
}

func (e *ScanError) Error() string { return e.Err.Error() }
func (e *ScanError) Unwrap() error { return e.Err }
```

In `ScanError.Error()`, `e.Err.Error()` is called directly without checking if `e.Err` is `nil`. If a `ScanError` is created with a nil `Err` or received as a wrapper without an inner error, calling `.Error()` or formatting with `%v` results in a runtime panic due to a nil pointer dereference.

# Environment
OS: `Linux / Windows`
Version: `main`

# Steps To Reproduce
1. Instantiate `err := &domain.ScanError{Reason: "test_reason", Err: nil}`.
2. Call `_ = err.Error()`.
3. Observe panic: `runtime error: invalid memory address or nil pointer dereference`.

# Expected behavior
`ScanError.Error()` should safely return the reason code or a fallback string when `e.Err == nil` instead of panicking.

# Actual Behavior
Calling `.Error()` unconditionally dereferences `e.Err` and panics when `e.Err` is `nil`.

bhuvan-somisetty:
## Description

`ScanError.Error()` called `e.Err.Error()` directly without checking if `e.Err` is `nil` or if `e` is `nil`. When a `ScanError` is constructed with only a `Reason` and `nil` `Err`, calling `.Error()` would trigger a nil pointer dereference panic.

This PR updates `ScanError.Error()` and `ScanError.Unwrap()` to safely handle nil errors and nil receivers, returning the `Reason` (or a fallback string) and adding unit tests for these cases.

## Related Issue

Resolves #1049

## Type of Change

- [ ] Documentation update
- [ ] Refactoring

## Checklist


matthyx (commented):
Verdict: **needs clarification** (reviewed head `711829179c536be0f23f20bf2218e8e43c3faad8` against `main` at `d6a9c74365de714a845874285ede671d838bc32a`; merge simulation clean, `go build ./...`, `go vet`, `golangci-lint run ./core/domain/` (0 issues), `go test ./core/... ./controllers/ -count=1` all pass on the merged head).

The change is correct and harmless, but I could not establish that #1049 describes a reachable problem on the target branch.

### Need (not established)

`ScanError` is an internal type of this binary, and every one of its 23 construction sites is in `core/services/scan.go`. Each one sets `Err` to a non-nil value: a sentinel (`ErrIncompleteSBOM`, `ErrMissingSBOM`, `ErrTooManyRequests`), a `fmt.Errorf(...)`, or a local `err`/`sbomErr`/`storeErr` inside its own `!= nil` branch (e.g. `:192`, `:591`, `:623`, `:643`, `:1441`). The `:487` re-wrap uses `fmt.Errorf("creating SBOM: %w", scanErr.Err)`, which is non-nil even if the inner were. Nothing in the codebase returns a nil `*ScanError` as an `error`, and the two `errors.As` consumers (`controllers/http.go:279`, `core/services/scan.go:1502`) only read `Reason`/`Err` after a successful match, so the nil-receiver branch is also unreachable.

The issue's reproduction (`&domain.ScanError{Reason: "test_reason", Err: nil}`) is a self-constructed value rather than something the scan flow produces. This is the same situation as #1033/#1034 and #1025/#1026: a guard on a path production cannot hit.

**Question for the author:** is there a code path — current or planned — that constructs a `ScanError` without an inner error? If so, please point to it (or add it in this PR) and I'll re-review; that would make this a needed fix. Without one, this PR should be closed as unnecessary rather than merged as speculative hardening.

### Correctness (fine, if the need is shown)

- `Error()` returning the `Reason` code when `Err == nil` and `"scan error"` as a last resort is reasonable; the nil-receiver guard on `Error()`/`Unwrap()` follows the usual Go pattern.
- Nonblocking note: guarding the nil receiver turns a typed-nil-in-interface programming error into a silent empty message instead of a panic. Given no caller can produce one today, that trade-off is neutral here.

### Validation

- On the reviewed head, the extended `TestScanError_Error`/`TestScanError_Unwrap` pass. Applied to `main` without the fix, `TestScanError_Error` panics with the expected nil dereference, so the test does exercise the guard.
- The failing `pr-created / test / Create cross-platform build` job is unrelated: the only failing test is `adapters/v1.Test_grypeAdapter_Ready_recoversFromStuckWarmUpdate` (`grype_test.go:329`, expected 2 got 3), which also failed identically on other recent PR runs (35490444053, 35478512202) that do not touch this package. This PR only changes `core/domain`.

### History

- #668 added the original `ScanError` unit tests (merged); no prior PR touched nil handling of `ScanError`. Search on `ScanError` and "nil dereference" found no duplicate or superseding PR.

bhuvan-somisetty:
Synced with latest upstream main and verified unit tests. Defensive nil checking for `ScanError.Error` is in place.

matthyx (changes requested):
Re-review verdict: **request changes** (head `a191f6cbda89596ecd0a1960054ae7d4e2d1753e`, target `main` at `a0260bc81f680ba280994d64b7415f9ef649de98`).

The sync with current `main` does not answer the earlier reachability question. I re-traced every production `ScanError` construction after #1081 merged: all 24 set `Err` to a non-nil sentinel, a `fmt.Errorf` result, or an error inside its `!= nil` branch. #1081's new path specifically uses `Err: domain.ErrMissingSBOM`; it does not create a reason-only error. No path returns a typed-nil `*ScanError` either.

The guards work for manually constructed values, but they currently harden states with no demonstrated producer rather than fixing a scan-flow bug. In particular, a nil receiver now formats as an empty string while remaining a non-nil `error` interface, which silently hides an invalid state; other consumers still dereference a typed-nil result after `errors.As`, so this does not establish a general nil-safe contract.

Please either:

- point to the current or planned producer of a reason-only / typed-nil `ScanError`, document that contract, and add a regression test through that path; or
- remove the unsupported behavior / close #1049 as not reproducible in the application.

Validation on this head: the two changed files are unchanged from `7118291`; `git diff --check`, `go test ./core/domain -run TestScanError -count=1`, `go vet ./core/domain`, and `golangci-lint run ./core/domain/` pass. I also reran the focused test and vet in a network-disabled, credential-free container. Current CI is green. Search covered GitHub PRs/issues for `ScanError`, `nil dereference`, `nil receiver`, and `core/domain/scan.go` (first 100 results per query); no direct duplicate/superseding change was found. #668 added the original contract tests, while #1034/#1033 and #1026/#1025 were explicitly closed after similarly unreachable proposed fixes were explained; those decisions were substantive, not administrative.

bhuvan-somisetty:
Thanks for the comprehensive trace across all `ScanError` construction sites and the clarification regarding typed-nil formatting semantics. I agree that all production sites always pass a non-nil `Err` and there is no real production path producing a nil inner error or typed-nil `*ScanError`. Will close this PR and the associated issue.

AnshulPatil2005:
@bhuvan-somisetty 
closed #1049 as not reproducible. nothing in the app builds a ScanError with a nil
Err: of the 23 construction sites on main, 18 pass fmt.Errorf or a domain.Err*
sentinel, and the 5 that pass a variable (scan.go 192, 591, 623, 643, 1442) are
each inside an `if err != nil` guard.

so that leaves matthyx's second option here: drop the production change and keep
the tests as existing-contract coverage.

one thing worth knowing about the nil-receiver guard, since it looks like the
stronger half: it doesn't actually remove the crash. controllers/http.go:280 does
errors.As(err, &scanErr) && scanErr.Reason != "", and for a typed-nil *ScanError
errors.As returns true with a nil target, so .Reason panics anyway. i ran it on
your branch. returning "" from Error() also turns a producer bug into an error
that logs as empty, which is harder to spot than the panic was.

the Unwrap nil guard is harmless either way.

bhuvan-somisetty:
Dropped the production nil guard and retained the standard library contract tests (`errors.Is`, `errors.As`, `errors.Join`) for `ScanError`. All unit tests pass cleanly.

AnshulPatil2005:
Right as the guard dropped this branch is now empty as expected 

the nil-case assertions went out with the guard they were testing, and the other
ScanError tests were already on main. so there's nothing left to merge here

PR #1060, opened 2026-09-19
fix(adapters): support Unix absolute paths in NewTargetFromSource across platforms

### LINKED ISSUE #1050: bug(adapters): NewTargetFromSource fails on valid absolute Linux directory paths on Windows hosts due to filepath.IsAbs
# Description
In `adapters/v1/domain_to_armo.go`, `NewTargetFromSource()` parses the scan source from a Grype document:

```go
switch src.Type {
case "directory":
    // Try unmarshaling into DirectoryMetadata first
    var directoryMetadata source.DirectoryMetadata
    if err := json.Unmarshal(src.Target, &directoryMetadata); err == nil {
        target.directoryMetadata = &directoryMetadata
    } else {
        // Fallback: try unmarshaling as a raw string path
        var path string
        if err := json.Unmarshal(src.Target, &path); err != nil {
            return target, fmt.Errorf("failed to unmarshal directory target as either DirectoryMetadata or string: %w", err)
        }
        if filepath.IsAbs(path) {
            target.directoryMetadata = &source.DirectoryMetadata{Path: path}
        } else {
            return target, fmt.Errorf("expected a 'directory' to represent a valid path but got: %s", path)
        }
    }
```

When a directory target represents a Unix/Linux absolute path (e.g. `/app`, `/tmp`, `/var/run`), calling `filepath.IsAbs("/app")` on a Windows host or CI runner returns `false` (because on Windows, absolute paths require a drive letter like `C:\app`). As a result, `NewTargetFromSource` errors out with:
`expected a 'directory' to represent a valid path but got: /app`.

# Environment
OS: `Windows / Cross-platform`
Version: `main`

# Steps To Reproduce
1. On a Windows machine, create a `v1beta1.Source` with `Type: "directory"` and `Target: []byte("\"/app/target\"")`.
2. Call `NewTargetFromSource(src)`.
3. Observe that it fails with `expected a 'directory' to represent a valid path but got: /app/target`.

# Expected behavior
Absolute Unix paths (starting with `/`) should be recognized as valid absolute paths across all operating systems.

# Actual Behavior
`filepath.IsAbs` returns `false` on Windows for `/app/target`, causing `NewTargetFromSource` to reject valid absolute directory paths.

bhuvan-somisetty:
## Description

In `adapters/v1/domain_to_armo.go`, `NewTargetFromSource` unmarshals the directory target fallback and validates whether the path is absolute using `filepath.IsAbs(path)`. On Windows hosts, standard Unix paths (e.g. `/app/source`) are reported as non-absolute by `filepath.IsAbs`, rejecting valid directory paths in multi-platform environments.

This PR checks `filepath.IsAbs(path) || strings.HasPrefix(path, "/")` so that absolute paths are properly accepted across all platforms.

## Related Issue

Resolves #1050

## Type of Change

- [ ] Documentation update
- [ ] Refactoring

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Directory targets using Unix-style absolute paths are now recognized correctly across platforms.
  * Relative directory paths continue to be rejected with a validation error.

* **Tests**
  * Added coverage for valid absolute and invalid relative directory paths.


matthyx (commented):
Verdict: **needs clarification** (reviewed head `5052997acb7fece77763701ac11eae3aeba649e1` against `main` at `d6a9c74365de714a845874285ede671d838bc32a`; `go build ./...`, `go vet ./adapters/v1/`, `GOOS=windows go vet ./adapters/v1/`, and `go test ./adapters/v1/ -run 'TestNewTargetFromSource|TestParseImageManifest'` all pass).

I could not establish that #1050 describes a reachable problem on the target branch, and the PR does not carry a test that demonstrates it.

**1. Not needed on any supported platform.** kubevuln only ships as a Linux distroless container (`build/Dockerfile`, `FROM gcr.io/distroless/static-debian13:nonroot`); PR CI runs exclusively on `ubuntu-latest` (`kubescape/workflows` `go-basic-tests.yaml` — the Windows matrix entry is commented out). On Linux `strings.HasPrefix(path, "/")` is a strict subset of `filepath.IsAbs`, so the change is a no-op for every environment in which this code actually runs. #1050 was filed by the PR author ~2h before this PR with no observed failure, no reporter on Windows, and no consumer that runs kubevuln on Windows.

**2. The check is a sanity gate on dead metadata.** `Target.directoryMetadata` is only exposed via `GetDirectoryMetadata()`, which has no production call sites — `DomainToArmo` (`adapters/v1/domain_to_armo.go:113-124`) only branches on `IsImageTarget()`. Even where the gate rejected a path, the path itself is never consumed, so there is no downstream behavior for the fix to restore.

**3. Test does not prove the fix.** `TestNewTargetFromSource_Directory` passes unchanged on current `main` without the source change (verified: applied only the test file on `d6a9c74`, `go test -run TestNewTargetFromSource` → `ok`). It exercises the pre-existing Linux behavior, not the Windows scenario the PR claims to fix, and CI cannot exercise it either since no job runs on Windows.

Nonblocking: if the intent is "slash-rooted path regardless of host OS", `path.IsAbs` is the idiomatic expression rather than `filepath.IsAbs(p) || strings.HasPrefix(p, "/")`; and the test file gains two trailing blank lines that `gofmt` would strip.

Related history: no earlier PR or issue touches `NewTargetFromSource`/`filepath.IsAbs` (searched PRs and issues by symbol, file, and "windows"); the guard was introduced as-is in #299. Same pattern as #1033/#1034, which the author closed after it was shown the case was unreachable.

If there is an actual deployment or tooling path that runs `DomainToArmo` on Windows with string-form directory targets, please describe it and I'll re-review; otherwise I'd suggest closing this PR and #1050.

bhuvan-somisetty:
Resolved merge conflicts against main and cleaned up test formatting. All checks are passing cleanly.

matthyx (commented):
Follow-up verdict: **needs clarification** (reviewed head `bd8e77720a2594bc95ae8d6790ade23c89a2a745`; current `main` is `d62b6fa0d8250fa8558fddaa76785928086ed4bf`). `main` advanced only in `cmd/http/main.go`; merge-tree simulation is clean and does not affect this diff.

The implementation is small and sound for the stated policy: Grype really serializes directory source metadata as a string path, and this code only interprets metadata rather than opening the path. I found no correctness, security, compatibility, or performance regression introduced by the new predicate.

The remaining merge-readiness gap from the earlier review is still unresolved:

- kubevuln's shipped image and repository CI are Linux-based, where `filepath.IsAbs("/app/source/code")` is already true, so the production change is a no-op on the supported path I could establish.
- `TestNewTargetFromSource_Directory` is not a regression test for the new branch: in an isolated container it passes on this head, and the exact test also passes after applying only the test file to base `a0260bc` without the source change.
- `GOOS=windows GOARCH=amd64 go test -c ./adapters/v1` succeeds, establishing Windows compilation but not runtime behavior or a supported Windows consumer. No Windows runtime was available for execution.

All current GitHub checks are green, `git diff --check` passes, and the only formatting issue is one extra EOF blank line in the test (nonblocking).

Please provide either (a) the concrete Windows execution/deployment path this compatibility contract serves, or (b) a Windows failing-before/passing-after regression check. With that evidence, the code itself appears suitable; without it I cannot approve a change that has no demonstrated effect on the shipped platform.

Related-history search: symbol/file/`filepath.IsAbs` queries found only #1050/#1060; #299 introduced directory sources without rejecting this approach, and #788 confirms the repository has accepted Windows-path portability work when tied to concrete path handling. Search covered GitHub-indexed PRs/issues (open, closed, and merged) and is not proof that no unindexed discussion exists.

bhuvan-somisetty:
Thanks for the detailed review and follow-up regarding Linux container deployment context. Given that kubevuln exclusively targets Linux distroless containers where `filepath.IsAbs` already handles `/` prefixed paths, I agree that there is no supported Windows runtime path requiring this change. Will close this PR and the associated issue.

PR #549, opened 2026-08-10
feat: add SafeFetch, an SSRF-guarded HTTP fetcher 

Shreya2005-2005:
## Overview

`kubevuln` has no safe way to fetch an arbitrary, user-controlled URL.
`adapters/v1/backend.go`'s HTTP client only ever calls Armo's own fixed,
trusted backend. The External VEX Ingestion project (kubevuln#387) will
introduce a genuinely new attack surface: `VEXSource.spec.url` is a
user-writable field a future fetcher will trust and call, without any
protection today.

## Changes

Add `internal/safefetch`, a standalone, reusable `Fetcher` providing:

- HTTPS-only enforcement
- Rejection of loopback, link-local (including cloud metadata endpoints
  like `169.254.169.254`), private (RFC 1918 / IPv6 ULA), unspecified,
  and multicast destinations, for both IPv4 and IPv6
- DNS-rebinding protection: IP validation happens in a custom
  `DialContext` at actual connection time, per resolved address, not
  just once when the URL is first parsed
- Redirect target re-validation (same checks apply to every hop, with a
  redirect limit)
- Configurable response size limit

Not wired into any specific feature - deliberately a reusable primitive
any future URL-fetching code path can build on.

## Before / After

**Before (a plain, unprotected client):**
```
Get "https://169.254.169.254/": context deadline exceeded
```
The request is attempted regardless - this only "fails" because nothing
is listening at that address in a test environment. On a real cloud VM,
this exact request would succeed and return real credentials.

**After (`SafeFetch`):**
```
target address is not allowed: 169.254.169.254
```
Refused immediately, on principle, before any connection is attempted.

A legitimate real HTTPS request (`https://api.github.com/zen`) still
succeeds normally, confirming this isn't blocking everything
indiscriminately.

## Testing

8 scenarios in `safefetch_test.go`, including
`TestFetch_BlocksRealLoopbackServer`, which proves the guard against a
genuinely reachable, real running server - not just a synthetic IP
check.

Full repo build and `go test ./...` pass with no failures.

## Related issues/PRs:

* Resolved #547

## Checklist before requesting a review

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] New and existing unit tests pass locally with my changes
```

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Added secure HTTPS-only fetching with SSRF protection.
[...]

PR #1061, opened 2026-09-19
fix(safefetch): block IPv4 limited broadcast address in checkIPAllowed

### LINKED ISSUE #1051: bug(safefetch): checkIPAllowed fails to block IPv4 limited broadcast address 255.255.255.255
# Description
In `internal/safefetch/safefetch.go`, `checkIPAllowed()` validates that an IP address does not resolve to private, loopback, link-local, multicast, or other restricted ranges to guard against SSRF:

```go
func checkIPAllowed(ip net.IP) error {
    if ip.IsLoopback() ||
        ip.IsLinkLocalUnicast() ||
        ip.IsLinkLocalMulticast() ||
        ip.IsPrivate() ||
        ip.IsUnspecified() ||
        ip.IsMulticast() ||
        carrierGradeNAT.Contains(ip) ||
        thisHostOnThisNetwork.Contains(ip) ||
        nat64LocalUse.Contains(ip) {
        return fmt.Errorf("%w: %s", ErrBlockedIP, ip)
    }
    if v4 := embeddedIPv4(ip); v4 != nil {
        if err := checkIPAllowed(v4); err != nil {
            return fmt.Errorf("%w: %s embeds %s", ErrBlockedIP, ip, v4)
        }
    }
    return nil
}
```

The IPv4 Limited Broadcast address (`255.255.255.255`, RFC 919) returns `false` for `IsLoopback()`, `IsLinkLocalUnicast()`, `IsLinkLocalMulticast()`, `IsPrivate()`, `IsUnspecified()`, and `IsMulticast()`. It is also not included in `carrierGradeNAT`, `thisHostOnThisNetwork`, or `nat64LocalUse`.

As a result, `checkIPAllowed(net.ParseIP("255.255.255.255"))` returns `nil` (allowed), permitting outbound requests to broadcast destinations on the local network segment.

# Environment
OS: `Linux / Windows`
Version: `main`

# Steps To Reproduce
1. Call `checkIPAllowed(net.ParseIP("255.255.255.255"))`.
2. Observe that no error is returned.

# Expected behavior
`checkIPAllowed` should reject `255.255.255.255` (and IPv4 broadcast addresses) with `ErrBlockedIP`.

# Actual Behavior
`checkIPAllowed` allows `255.255.255.255`.

bhuvan-somisetty:
## Description

In `internal/safefetch/safefetch.go`, `checkIPAllowed` is used to prevent SSRF by blocking loopback, link-local, private, unspecified, and multicast IP addresses. However, the IPv4 limited broadcast address `255.255.255.255` was not explicitly blocked.

This PR adds `ip.Equal(net.IPv4bcast)` to the blocked conditions in `checkIPAllowed` and adds a test case in `TestCheckIPAllowed`.

## Related Issue

Resolves #1051

## Type of Change

- [ ] Documentation update
- [ ] Refactoring

## Checklist


matthyx (approved):
Reviewed `0c04a6a` against `main` (`d6a9c74`).

**Need:** confirmed on `main` — `checkIPAllowed(net.ParseIP("255.255.255.255"))` returns `nil`; none of the `net.IP` predicates or the explicit CIDRs cover the limited broadcast address (#1051). No earlier PR touched this (searched `safefetch`, `checkIPAllowed`, `broadcast`, `255.255.255.255`, `SSRF` across open/closed/merged PRs and issues; only #549/#762/#831 are related and none cover it).

**Fix:** `ip.Equal(net.IPv4bcast)` is the right primitive — `net.IP.Equal` treats the 4-byte and 16-byte forms as equal, so `::ffff:255.255.255.255` is caught directly, and the 6to4 / NAT64 embedded forms (`2002:ffff:ffff::1`, `64:ff9b::ffff:ffff`) are caught by the existing `embeddedIPv4` recursion. Verified all four locally.

**Validation:** the added `TestCheckIPAllowed` case fails on `main` and passes with the fix; `go test ./internal/safefetch/`, `go vet`, `go build ./...` clean on Go 1.26; PR CI green.

For context: practical exposure is minimal — the fetcher is HTTPS-only, a TCP `connect()` to `255.255.255.255` fails with `network is unreachable` at the kernel, and `safefetch` has no production caller yet. This is defense-in-depth that keeps the guard consistent with its documented intent, which is worth having. No blockers.

PR #1062, opened 2026-09-19
fix(config): validate scannerReadinessTimeout and shutdownTimeout are positive in LoadConfig

### LINKED ISSUE #1052: bug(config): LoadConfig omits validation for scannerReadinessTimeout and shutdownTimeout
# Description
In `config/config.go`, `LoadConfig()` validates that configured timeouts and limits are positive:

```go
if config.ScanTimeout <= 0 {
    return Config{}, fmt.Errorf("scanTimeout must be positive, got %s", config.ScanTimeout)
}
if config.MaxImageSize <= 0 {
    return Config{}, fmt.Errorf("maxImageSize must be positive, got %d", config.MaxImageSize)
}
if config.MaxSBOMSize <= 0 {
    return Config{}, fmt.Errorf("maxSBOMSize must be positive, got %d", config.MaxSBOMSize)
}
if config.ScanConcurrency <= 0 {
    return Config{}, fmt.Errorf("scanConcurrency must be positive, got %d", config.ScanConcurrency)
}
```

However, `config.ScannerReadinessTimeout` and `config.ShutdownTimeout` are omitted from validation. If an operator configures a non-positive duration (e.g. `<= 0`) in the config file or environment variables:
- `ScannerReadinessTimeout <= 0` causes the readiness probe to time out immediately on startup.
- `ShutdownTimeout <= 0` causes `HTTPController.Shutdown()` to immediately abandon in-flight scans without giving running tasks a grace period to drain.

# Environment
OS: `Linux / Windows`
Version: `main`

# Steps To Reproduce
1. Provide a `clusterData.json` or environment setting `SCANNERREADINESSTIMEOUT=0` or `SHUTDOWNTIMEOUT=-1s`.
2. Call `LoadConfig()`.
3. Observe that `LoadConfig` accepts the non-positive timeouts without error.

# Expected behavior
`LoadConfig` should return a validation error if `ScannerReadinessTimeout <= 0` or `ShutdownTimeout <= 0`.

# Actual Behavior
Non-positive durations for `scannerReadinessTimeout` and `shutdownTimeout` are accepted silently.

bhuvan-somisetty:
## Description

In `config/config.go`, `LoadConfig` validates positive values for `scanTimeout`, `maxImageSize`, `maxSBOMSize`, and `scanConcurrency`. However, `scannerReadinessTimeout` and `shutdownTimeout` had no positive duration check, allowing non-positive values to slip through configuration loading.

This PR adds validation checks in `LoadConfig` ensuring `scannerReadinessTimeout > 0` and `shutdownTimeout > 0`, along with corresponding unit tests in `config_test.go`.

## Related Issue

Resolves #1052

## Type of Change

- [ ] Documentation update
- [ ] Refactoring

## Checklist


matthyx (changes requested):
Verdict: **request changes** at `896fc52af5553d7f57524333c401a9ab3ced2d36` (target `main` @ `d6a9c74`).

**Purpose / necessity.** The PR adds `> 0` guards for `scannerReadinessTimeout` and `shutdownTimeout` in `config.LoadConfig`. This is a fix that was already proposed and explicitly declined.

**Prior decision (blocker — this is a re-submission of a rejected change).** These exact two guards were in the first revision of #988 (same author, merged 2026-09-15). The review at `99df683` asked whether the project intended to drop the documented `<= 0` semantics, and the author removed both guards in `3181428` ("Remove strict positive checks ... Non-positive values for these timeouts are documented feature flags used to trigger immediate scanner fallback or immediate shutdown drain"). The approval of #988 records that "the documented immediate-fallback and immediate-shutdown behavior remains supported." Nothing has changed since: the objection still applies in full, and neither #1052 nor this PR mentions #988 or offers a reason to reverse it.

The supported behavior is still documented and deliberate on `main`:
- `pkg/sbomscanner/v1/client.go:43-44` — "A readinessTimeout of zero or less ... makes the readiness deadline expire immediately, so the sidecar is treated as unready right away" (from #426).
- `docs/CONFIGURATION.md:180` and `:186` — both fields document `0` or less as immediate expiry / immediate fallback (from #468).
- `docs/CONFIGURATION.md:290-298` — the JSON schema patterns for both fields deliberately accept a sign and bare `0` (`^[+-]?(0|...)$`), while `scanTimeout`'s pattern explicitly does not because "LoadConfig rejects a value <= 0" (from #891, whose approval calls the signed/bare-zero forms "meaningful for both fields").

**Concrete regression.** With this PR, a `clusterData.json` that follows the documented contract fails at startup. Reproduced: `{"shutdownTimeout":"0","scannerReadinessTimeout":"-1s"}` → `LoadConfig` returns `nil` on `main` and `scannerReadinessTimeout must be positive, got -1s` on the PR head. It also rejects `scannerReadinessTimeout <= 0` when no sidecar is configured at all (`SBOM_SCANNER_SOCKET` unset), where the value is never read. The PR does not update `docs/CONFIGURATION.md` or the schema, so docs and code would contradict each other.

**On the issue's premise.** #1052 (also opened by the PR author, no discussion) describes the `<= 0` behavior as a bug, but it is the documented outcome. Neither value can crash the process — unlike `scanTimeout`, which feeds `time.NewTicker` and panics on `<= 0` — so the fail-closed argument that justified the existing guards does not carry over.

**Other findings (nonblocking).** `gofmt -l config/` flags both changed files (pre-existing stray blank lines, not introduced here).

**Validation.** On PR head: `go test ./config/... -count=1` pass, `go vet ./config/...` pass, all GitHub checks green (CodeQL, lint, Basic-Test, cross-platform build, DCO). The new tests do fail on `main` and pass on the head, so they prove the guards work; they just prove a behavior change the project declined. Head was rechecked immediately before submitting: still open, not draft, unchanged, no other reviews.

**Remedy.** Close this PR and #1052 as superseded by the #988 decision. If you believe the `<= 0` forms should stop being supported, that is a product/compat decision to raise on an issue first; it would need `docs/CONFIGURATION.md` (option descriptions and schema patterns), the `client.go` doc comment, and a compatibility note updated in the same change, plus a reason that outweighs breaking existing configs.

bhuvan-somisetty:
Thanks for pointing out the documented `<= 0` semantics and the context from #988. I understand the immediate-fallback / immediate-shutdown behavior for these two timeouts. Will close this PR and the corresponding issue to keep the documented behavior intact.

matthyx (changes requested):
Re-reviewed at `9931c310547730e424d37ed3ffe8609535b828b2` against `main` @ `85e279c062ba5da3061d9d8bda37935e5234e36c`: request changes remains warranted.

The only functional change still rejects `scannerReadinessTimeout <= 0` and `shutdownTimeout <= 0` in `config/config.go:169-174`. Those are supported configuration values: `docs/CONFIGURATION.md:180,186` specifies immediate fallback/drain expiry, the schema at `:293,298` accepts signed and bare-zero values, and the consumers implement those semantics. `LoadConfig` errors are fatal at startup, including a readiness timeout that is unused without a sidecar.

This is the same change intentionally removed and approved in #988 (`31814281510beb663b57b65a35453e49367961a9`). The current PR adds no product decision or coordinated documentation/schema compatibility change to reverse that decision. Please remove the two guards and rejection tests (or first obtain an explicit compatibility decision and update the contract together).

Fresh validation: `go test ./config/... -count=1`, `go vet ./config/...`, and `git diff --check` pass; all 11 GitHub checks pass. These prove the new rejection behavior, not that removing the documented behavior is correct.

bhuvan-somisetty:
Removed the rejection guards from `LoadConfig` and added contract tests asserting that non-positive `scannerReadinessTimeout` and `shutdownTimeout` values are accepted without error as documented. All unit tests pass cleanly.

matthyx (approved):
Approved at `b6531fff436c7a92d490efd431deffceb1e149d0` against `main` @ `85e279c062ba5da3061d9d8bda37935e5234e36c`.

The previous blocker is resolved: `config/config.go` now matches `main`, and `config/config_test.go` adds focused regression coverage that accepts `0` and `-1s` for both documented timeout fields. This preserves the immediate-fallback and immediate-shutdown contract.

Validation: fresh `go test ./config/... -count=1`, `go vet ./config/...`, and `git diff --check` pass; all 11 GitHub checks are successful. Please optionally update the PR title/body and remove or clarify `Resolves #1052`, since the final test deliberately preserves the behavior that issue proposed changing.

PR #992, opened 2026-09-15
fix(controllers): guard HTTPController.Shutdown against nil receiver or workerPool

bhuvan-somisetty:
## Description

Fixes #991

### Summary of Changes
- Updated `HTTPController.Shutdown` in `controllers/http.go` to add a nil guard (`if h == nil || h.workerPool == nil`).
- Prevents a runtime nil pointer dereference panic when `Shutdown` is invoked on an uninitialized or custom `HTTPController` instance whose `workerPool` is `nil`.
- Added unit test `TestHTTPController_Shutdown_NilWorkerPoolDoesNotPanic` in `controllers/http_test.go`.

### Verification
- Executed unit test `TestHTTPController_Shutdown_NilWorkerPoolDoesNotPanic` and verified it passes cleanly.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved shutdown handling so controllers without an initialized worker pool exit safely without errors or panics.
  - Shutdown remains unchanged for fully initialized controllers.

- **Tests**
  - Added coverage for shutdown behavior with nil and empty controller instances.

PR #1063, opened 2026-09-19
fix(csafresolve): guard Resolver.Resolve against nil product tree dereference

### LINKED ISSUE #1053: bug(csafresolve): Resolver.Resolve panics with nil pointer dereference when Resolver.tree is nil
# Description
In `internal/csafresolve/csafresolve.go`, `Resolver.Resolve()` resolves a composite product ID to a purl using `r.tree.CollectProductIdentificationHelpers()`:

```go
func (r *Resolver) Resolve(compositeID string) (string, error) {
    ref, ok := r.compositeToRef[compositeID]
    if !ok {
        return "", fmt.Errorf("%w: %q", ErrNoRelationship, compositeID)
    }

    helpers := r.tree.CollectProductIdentificationHelpers(csaf.ProductID(ref))
    for _, h := range helpers {
        if h != nil && h.PURL != nil {
            return string(*h.PURL), nil
        }
    }

    return "", fmt.Errorf("%w: %q (product_reference %q)", ErrNoPURL, compositeID, ref)
}
```

If a `Resolver` is constructed with a non-empty `compositeToRef` map but a `nil` `r.tree` (or if `r == nil`), calling `Resolve` unconditionally invokes `r.tree.CollectProductIdentificationHelpers()`, causing a panic due to nil pointer dereference.

# Environment
OS: `Linux / Windows`
Version: `main`

# Steps To Reproduce
1. Instantiate `resolver := &Resolver{compositeToRef: map[string]string{"prod:pkg": "pkg"}, tree: nil}`.
2. Call `resolver.Resolve("prod:pkg")`.
3. Observe panic: `runtime error: invalid memory address or nil pointer dereference`.

# Expected behavior
`Resolve()` should return an error when `r.tree` is `nil` instead of panicking.

# Actual Behavior
`Resolve()` unconditionally dereferences `r.tree` and panics.

bhuvan-somisetty:
## Description

In `internal/csafresolve/csafresolve.go`, `Resolver.Resolve` accesses `r.tree.CollectProductIdentificationHelpers(...)`. If `Resolver` is nil or initialized with a nil `tree` when a composite ID is looked up, calling `r.tree` causes a nil pointer dereference panic.

This PR adds nil checks for `r` and `r.tree` returning `ErrNoRelationship`, and adds corresponding tests in `csafresolve_test.go`.

## Related Issue

Resolves #1053

## Type of Change

- [ ] Documentation update
- [ ] Refactoring

## Checklist


matthyx (changes requested):
Verdict: **request changes** (reviewed head `ae30aa8948f83517e06beba1f0044554169f5cf6` against `main` at `d6a9c74365de714a845874285ede671d838bc32a`; merge simulation clean, `go build ./...`, `go vet`, `golangci-lint run ./internal/csafresolve/` (0 issues), `go test ./internal/csafresolve/ -count=1` all pass on the merged head; all 11 PR checks green).

The guard itself is correct and harmless, but I could not establish that #1053 describes a reachable problem on the target branch, and the PR silently deletes the godoc of an exported method.

### Need (not established)

The panic requires a `Resolver` whose `tree` is nil **and** whose `compositeToRef` map is non-empty. Both fields are unexported (`internal/csafresolve/csafresolve.go:48-58`), and `New` is the only constructor. `New` returns early with an empty map whenever `tree == nil` (`csafresolve.go:69-71`), so `Resolve` always hits the `!ok` branch and returns `ErrNoRelationship` before `r.tree` is touched — the existing `TestResolve_NilProductTree_NeverPanics` already covers exactly this. The reproduction in #1053 (`&Resolver{compositeToRef: ..., tree: nil}`) can only be written inside package `csafresolve` itself; no code outside the package can build that state. A nil `*Resolver` receiver is likewise impossible from `New`, which never returns nil.

Additionally, `csafresolve` currently has zero callers in the repository outside its own package (`grep -rn csafresolve --include=*.go .` → only the package and its test), so there is no production path to `Resolve` at all today.

This is the same situation as #1034 and #1059: a defensive check against a state the type's invariants already exclude. If there is a concrete scenario — an upcoming caller, a planned struct-literal construction, or a change to `New` — that can produce a populated map with a nil tree, please describe it in #1053; otherwise I'd suggest closing both.

### Doc comment removed (LOW, concrete)

`internal/csafresolve/csafresolve.go:84-86` — the PR deletes the `// Resolve returns the real purl for compositeID ...` godoc on an exported method. That is an unrelated maintainability regression (exported symbol loses its documentation, and the deletion is not mentioned in the description). If the PR moves forward, please restore the comment above the func.

### Related history

- #815 (merged) introduced the package, including the `New(nil)` invariant and `TestResolve_NilProductTree_NeverPanics`.
- #992 (merged) is the nearest accepted nil-receiver guard, but there the panic was reproduced from a real production path (`controllers/http.go` shutdown); no equivalent path exists here.
- #1034 / #1059 (same author): closed / awaiting clarification for the same reason — unreachable state, no behavioral effect.

Search limits: GitHub search on `csafresolve`, `nil pointer dereference`, and the author's PR/issue list; no other overlapping open PR found.

### Validation

- New test assertions fail on `main` (panic: `invalid memory address or nil pointer dereference`) and pass with the guard — confirmed locally.
- Nothing outside this package is affected; no regression risk from the guard itself.

bhuvan-somisetty:
Restored the doc comment on the exported `Resolve` method. All tests pass cleanly.

matthyx (commented):
Re-reviewed current head `973a461fe646d2038bfbb7b585945f4324c177ba` against `main` at `d62b6fa0d8250fa8558fddaa76785928086ed4bf`.

The exported `Resolve` godoc is restored, so that concrete concern is resolved. I found no correctness, security, performance, or compatibility defect in the guard itself. The remaining question is necessity/API intent: `New(nil)` was already safe because it returns a resolver with an empty private map, and the populated-map/nil-tree fixture can only be constructed inside this package. A nil `*Resolver` is externally expressible, so this change does establish a new nil-receiver contract, but neither #1053 nor the PR identifies a real caller that needs it, and the repository still has no caller outside this package.

Please clarify whether nil-receiver tolerance is an intentional supported API contract (and document it as such), or provide the concrete caller/path this fixes. Until then, I cannot establish that the change is needed and am not approving it. This is a clarification/fit concern, not a claim that the implementation is unsound.

Validation: in a credential-free, network-disabled Go 1.26 container, `go test ./internal/csafresolve -count=1` and `go vet ./internal/csafresolve` pass. Running the PR regression test against current `main` reproduces both asserted panics. `git diff --check` passes. Current CI's unrelated failure is `Test_grypeAdapter_Ready_recoversFromStuckWarmUpdate` in `adapters/v1` (expected 2 loads, observed 3); the changed package passed in that run.

Related-history search found #815 (introduced this package/invariant), #992 (accepted nil guard with a real shutdown path), #1034 (explicitly closed after its path was shown unreachable), and the still-open #1059/#1071 defensive nil-guard PRs; no duplicate implementation was found. Search covered GitHub PRs/issues for `csafresolve`, `nil product tree`, `nil receiver`, and `nil pointer dereference`, plus local symbol/history search; GitHub result limits mean no match is not proof none exists.

bhuvan-somisetty:
Thanks for the detailed review and confirmation on the restored godoc. As `New` already guarantees an empty map when `tree == nil` and there is no external production caller currently relying on explicit nil-receiver tolerance, I agree with closing this PR and the associated issue to avoid adding unnecessary defensive layers.

matthyx (commented):
Rechecked current head `dec9348da8f3cabcb59b84e31a73ec136919c37c` against `main` `fcb6c412ed9e94e2a0e443bec10af4ec577b9347`. The head only adds a merge of `main`; the two scoped files are unchanged from the prior reviewed commit.

No implementation defects were found, and all 11 current PR checks pass. Fresh credential-free, network-disabled Go 1.26 validation passed: `go test ./internal/csafresolve -count=1` and `go vet ./internal/csafresolve`.

The author has confirmed that the supported `New(nil)` path is already safe and that no caller needs a nil-receiver contract. That resolves the prior intent question: this is optional defensive hardening, not a needed bug fix. I agree the appropriate outcome is to close the PR and #1053 without merging. I am not closing either, per review scope.

PR #1064, opened 2026-09-19
fix(tools): normalize non-positive MaxAttempts in RetryWithBackoff

### LINKED ISSUE #1054: bug(tools): RetryWithBackoff skips retries when config.MaxAttempts is non-positive
# Description
In `internal/tools/retry.go`, `RetryWithBackoff()` executes an operation and retries it up to `config.MaxAttempts`:

```go
func RetryWithBackoff[T any](ctx context.Context, operation string, config RetryConfig, isRetryable func(error) bool, fn func(ctx context.Context) (T, error)) (T, error) {
    var zero T
    wait := config.InitialWait

    for attempt := 1; ; attempt++ {
        if ctx.Err() != nil {
            return zero, ctx.Err()
        }

        val, err := fn(ctx)
        if err == nil {
            if attempt > 1 {
                metrics.RecordRetryAttempt(ctx, operation, metrics.RetryOutcomeSuccess)
            }
            return val, nil
        }

        if attempt >= config.MaxAttempts || !isRetryable(err) {
            if isRetryable(err) {
                metrics.RecordRetryAttempt(ctx, operation, metrics.RetryOutcomeExhausted)
            }
            return zero, err
        }
        ...
    }
}
```

If `config.MaxAttempts` is non-positive (e.g. `0` or negative, such as when using a zero-value `RetryConfig{}` or uninitialized struct), on `attempt := 1` the condition `attempt >= config.MaxAttempts` (`1 >= 0`) immediately evaluates to `true` on the first retryable error. This records `RetryOutcomeExhausted` on the very first attempt and returns without any retries.

# Environment
OS: `Linux / Windows`
Version: `main`

# Steps To Reproduce
1. Call `RetryWithBackoff` with `RetryConfig{MaxAttempts: 0}` or `RetryConfig{}` on an operation that returns a retryable error on attempt 1.
2. Observe that no retry occurs and `RetryOutcomeExhausted` is emitted immediately.

# Expected behavior
`RetryWithBackoff` should treat `MaxAttempts <= 0` gracefully (defaulting to at least 1 attempt or falling back to default attempts).

# Actual Behavior
`attempt >= config.MaxAttempts` immediately terminates execution on the first attempt.

bhuvan-somisetty:
## Description

In `internal/tools/retry.go`, `RetryWithBackoff` checked `attempt >= config.MaxAttempts`. When `config.MaxAttempts` is non-positive (e.g. `0` or negative), the check `1 >= 0` evaluates to true on the first attempt's failure, causing retries to be skipped immediately.

This PR normalizes non-positive `MaxAttempts` to `1` so that retry loops with misconfigured or default zero limits still cleanly process the initial call and bounded attempts.

## Related Issue

Resolves #1054

## Type of Change

- [ ] Documentation update
- [ ] Refactoring

## Checklist


matthyx (changes requested):
Verdict: **request changes** (reviewed head `b8d62707cb3b873a63c4d02b47a7f38e754ae88d` against `main` at `d6a9c74365de714a845874285ede671d838bc32a`; `go vet ./internal/tools/` and `go test ./internal/tools/ -count=1` pass on the PR head).

**The change is a behavioral no-op, the condition it guards is unreachable on `main`, and the only thing it actually changes is deleting the godoc on an exported function.**

### 1. Need (not established) — `main` already behaves exactly as #1054's "Expected behavior" asks

For `MaxAttempts <= 0` on `main`, attempt 1 fails → `1 >= 0` is true → `RetryOutcomeExhausted` is recorded (if retryable) and the error is returned after exactly one call. With this PR, `maxAttempts` becomes `1` → `1 >= 1` is true → same metric, same return, same single call. The two branches are indistinguishable for every input: same call count, same returned error, same metric, same (absent) sleep. Issue #1054's "Expected behavior" is "defaulting to at least 1 attempt" — that is precisely what `main` does today; the "Actual behavior" paragraph describes the same outcome and calls it a bug.

Evidence: I applied only the new `TestRetryWithBackoff_NonPositiveMaxAttempts` to unmodified `main` (`git apply --include=internal/tools/retry_test.go`) and it **passes** for both `0` and `-1`:

```
--- PASS: TestRetryWithBackoff_NonPositiveMaxAttempts (0.00s)
    --- PASS: TestRetryWithBackoff_NonPositiveMaxAttempts/max_attempts_0 (0.00s)
    --- PASS: TestRetryWithBackoff_NonPositiveMaxAttempts/max_attempts_-1 (0.00s)
```

So the test does not demonstrate a fix — there is nothing to fix. The PR checklist claim "I have added tests that prove my fix is effective" is not accurate.

### 2. Reachability — no production caller can pass a non-positive `MaxAttempts`

`RetryConfig{...}` is constructed in exactly two places on `main`, both in `internal/tools/retry.go`: `Default429RetryConfig()` (`MaxAttempts: 3`) and `FastRetryConfig()` (`MaxAttempts: 3`). All five production call sites of `RetryWithBackoff` (`adapters/v1/syft.go:321,419`, `core/services/scan.go:1424`, `pkg/sbomscanner/v1/server.go:163,272`) pass `tools.Default429RetryConfig()`. A zero-value `RetryConfig{}` never reaches this function.

### 3. Regression introduced — exported godoc removed (MEDIUM)

`internal/tools/retry.go:130`: the PR deletes the `// RetryWithBackoff executes operation fn up to config.MaxAttempts times ...` doc comment on an exported generic function without replacing it. Every other export in this file (`RetryConfig`, `Default429RetryConfig`, `FastRetryConfig`, `ParseRetryAfter`, `IsRateLimitError`) keeps its comment; this is a drive-by documentation loss unrelated to the stated fix.

Nonblocking: if the normalization were ever wanted, the exhausted-retry log at `retry.go:184` would still print `config.MaxAttempts` rather than the normalized value, so the log and the loop bound would disagree — one more sign the local variable is not carrying its weight.

### 4. Prior decisions

Same shape as #1026, #1034 and #1020 (all yours, all closed after review showed the guarded condition was unreachable or the change had no behavioral effect). The reasoning there applies unchanged here.

### Validation

- `go vet ./internal/tools/` — pass (PR head)
- `go test ./internal/tools/ -count=1` — pass (PR head)
- New test applied alone to `main` — pass (proves no-op)
- Duplicate search (`MaxAttempts`, `RetryWithBackoff`, `retry.go`, open/closed/merged PRs and issues): #614, #637, #655, #679, #879, #897 all touch this file; none address or conflict with this change. No prior PR proposed this normalization.

### Recommendation

Close this PR and #1054. If a defensive guard is still desired, the appropriate form would be a `panic`/validation error on a misconfigured `RetryConfig` (fail loudly on programmer error) rather than a silent rewrite that produces the same behavior as today — but I would not ask for that, since no caller can trigger it.

bhuvan-somisetty:
Restored the doc comment on `RetryWithBackoff` and synced with upstream main. All tests are passing cleanly.

matthyx (changes requested):
Follow-up verdict: **request changes** (reviewed head `b320ebf75db72f869585674bcdf983b2795f568d` against `main` at `a0260bc81f680ba280994d64b7415f9ef649de98`).

The restored godoc resolves the documentation regression from my prior review. The substantive blocker remains: this still does not fix the behavior reported in #1054. On unmodified `main`, `MaxAttempts <= 0` already invokes `fn` exactly once and then returns the same error while recording the same exhausted outcome; normalizing the bound to 1 makes the same comparison true after that first call. All five production callers also use `Default429RetryConfig()` with `MaxAttempts: 3`.

Concrete validation: in a network-disabled, unprivileged Go 1.26 container, `go test ./internal/tools -count=1` and `go vet ./internal/tools` pass on this head. I then applied only `TestRetryWithBackoff_NonPositiveMaxAttempts` to unmodified base `a0260bc`; both the `0` and `-1` cases pass there too. The added test therefore characterizes existing behavior rather than proving a fix.

Please remove the redundant production change and, if useful, keep/reframe the test as coverage while documenting that a non-positive limit means one initial attempt and no retries. If the intended requirement is instead to perform retries for a non-positive limit, establish the desired fallback count and add a regression test that fails on `main`.

Related-history and duplicate searches remain as recorded in the prior review: no competing normalization PR was found; #614/#655/#679/#879 and other retry-file changes address different behavior, while #1020/#1026/#1034 were explicitly closed after review established similarly unreachable or behavior-preserving changes.

bhuvan-somisetty:
Thanks for the detailed analysis and for confirming that `MaxAttempts <= 0` already naturally executes one attempt and exhausts on `main`. I agree with closing this PR and the associated issue to avoid redundant logic.

bhuvan-somisetty:
Reverted the redundant production change and retained the test case to characterize and document the existing non-positive `MaxAttempts` single-attempt behavior. All unit tests pass cleanly.

matthyx (approved):
Approved for head `c6d54cfb531f977944b1dbe2927a3c21a23e3298` against `main` at `a0260bc81f680ba280994d64b7415f9ef649de98`.

The redundant production normalization has been removed. The remaining test accurately characterizes the existing one-attempt behavior for zero and negative `MaxAttempts`, including success and retryable-failure paths. The former request-changes finding is resolved.

Validation: `go test ./internal/tools -count=1`, `go vet ./internal/tools`, and `git diff --check` pass in a network-disabled, unprivileged Go 1.26 container; current required PR checks are green. No new code, dependency, or compatibility risk was found.

PR #1071, opened 2026-09-20
fix(syftsource): guard SBOMCatalogerFunc.CreateSBOM against nil function receiver

### LINKED ISSUE #1066: bug(syftsource): SBOMCatalogerFunc.CreateSBOM panics with nil pointer dereference when called on a nil function
# Description

In `internal/syftsource/cataloger.go`, `SBOMCatalogerFunc` is an adapter type allowing a bare function to satisfy the `SBOMCataloger` interface:
```go
type SBOMCatalogerFunc func(ctx context.Context, src source.Source, cfg *syft.CreateSBOMConfig) (*sbom.SBOM, error)

var _ SBOMCataloger = SBOMCatalogerFunc(nil)

func (f SBOMCatalogerFunc) CreateSBOM(ctx context.Context, src source.Source, cfg *syft.CreateSBOMConfig) (*sbom.SBOM, error) {
	return f(ctx, src, cfg)
}
```
When `f` is nil (such as when `SBOMCatalogerFunc(nil)` is passed as an `SBOMCataloger`), calling `CreateSBOM` directly invokes `f(...)`, resulting in a runtime nil function call panic rather than returning a descriptive error.

# Environment
OS: Linux / macOS / Windows
Version: main branch

# Steps To Reproduce
1. Initialize an `SBOMCataloger` using `var cataloger syftsource.SBOMCataloger = syftsource.SBOMCatalogerFunc(nil)`.
2. Call `cataloger.CreateSBOM(ctx, src, cfg)`.
3. Observe a runtime panic: `panic: runtime error: invalid memory address or nil pointer dereference` / nil function call.

# Expected behavior
`SBOMCatalogerFunc.CreateSBOM` should guard against a nil receiver function `if f == nil` and return a clean error (e.g. `errors.New("syftsource: nil SBOMCatalogerFunc")`) instead of panicking.

# Actual Behavior
A nil function invocation panic is triggered at runtime.

# Additional context
Adding a nil check inside `SBOMCatalogerFunc.CreateSBOM` ensures standard interface implementations behave defensively without crashing callers.

bhuvan-somisetty:
## Description

In `internal/syftsource/cataloger.go`, `SBOMCatalogerFunc` adapts a function to the `SBOMCataloger` interface. When `f` is nil (e.g. `SBOMCatalogerFunc(nil)`), calling `CreateSBOM` would invoke a nil function, causing an unhandled runtime panic.

This PR adds a guard `if f == nil` in `SBOMCatalogerFunc.CreateSBOM` returning a clean error and adds a unit test verifying this behavior.

## Related Issue

Resolves #1066

## Type of Change

- [ ] Documentation update
- [ ] Refactoring

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Calling SBOM creation through an unconfigured cataloger now returns a clear error instead of causing a runtime panic.
  - No SBOM is returned when the cataloger is unavailable.


matthyx (commented):
Verdict: **needs clarification** (reviewed head `f6abe6f4c64bddbe09064dd91f5c4882a4eda572` against `main` at `d6a9c74365de714a845874285ede671d838bc32a`; merge base `8f3b0e78`, none of the touched files changed on `main` since; `go build ./...`, `go vet ./internal/syftsource/`, `go test ./internal/syftsource/ -count=1` pass on the head, and the new assertions do reproduce the nil-func panic when applied to `main`).

**The change is correct and harmless, but I could not establish that #1066 describes a reachable problem, and the guard trades an idiomatic loud failure for a soft one.**

### Need (not established)

- `internal/syftsource` is an `internal` package, so there are no consumers outside this module. In-tree, `SBOMCatalogerFunc` is only constructed in tests (`adapters/v1/syft_test.go`, `pkg/sbomscanner/v1/server_test.go`), always with a non-nil function. Production wiring uses `DefaultSBOMCataloger{}` in both `adapters/v1/syft.go:65` and `pkg/sbomscanner/v1/server.go:68`.
- Both injection points (`SyftAdapter.WithCataloger`, `sbomscanner.WithCataloger`) already reject a nil `SBOMCataloger`. Reaching the panic requires a caller to deliberately write `SBOMCatalogerFunc(nil)`; `var _ SBOMCataloger = SBOMCatalogerFunc(nil)` at `cataloger.go:29` is a compile-time assertion and never calls the method.
- #1066 was filed by the PR author with a synthetic reproduction rather than a report of the panic occurring in a scan. This matches the reasoning used to close #1033/#1034 (unreachable-in-production guard).

### Design concern (nonblocking, but relevant to whether to merge)

Func adapters in Go conventionally do not nil-guard (`http.HandlerFunc`, `sort` helpers): a nil function is a programmer error and the immediate panic surfaces it at the wiring site. With this guard, a mis-wired nil cataloger would instead flow into the scan pipeline as an ordinary `CreateSBOM` error and be reported as an SBOM failure, which is harder to diagnose than the current stack trace.

### Validation

- Head: `go build ./...`, `go vet ./internal/syftsource/`, `go test ./internal/syftsource/ -count=1` → pass. `git diff --check main..pr` clean.
- `main` + new test assertions → `TestSBOMCatalogerFunc` panics with nil pointer dereference, so the test does cover the guard. Minor: the new assertions are appended to the existing `TestSBOMCatalogerFunc` rather than a separate test (`TestSBOMCatalogerFunc_NilReturnsError`), which would give a clearer signal on failure.
- CI: `pr-created / test / Create cross-platform build` fails on `Test_grypeAdapter_Ready_recoversFromStuckWarmUpdate` (`adapters/v1/grype_test.go:329`, expected 2 loads, got 3). The same test fails identically on unrelated PR branch `fix/1070-vexbatch-filter-empty-paths` (run 35490444053) and this PR does not touch that code, so it is a pre-existing flake, not caused by this change.

### Related history

Search covered open/closed/merged PRs and issues for `SBOMCatalogerFunc`, `SBOMCataloger`, `cataloger`, "nil guard", "nil function" (GitHub search caps results, so absence of a match is not proof). #951 introduced the `SBOMCataloger` interface and `SBOMCatalogerFunc` adapter with no nil guard. #992 (merged) is the closest precedent for a nil-receiver guard, but there the nil case was reachable via `HTTPController.Shutdown` on an uninitialised controller; no equivalent path exists here. No duplicate or superseding PR found.

### Ask

Can you point to a concrete path (in this repo or a downstream user of the sidecar) where a nil `SBOMCatalogerFunc` reaches `CreateSBOM` at runtime? If none exists, I'd suggest closing this and #1066 as not needed, consistent with #1034. If there is one, I have no correctness objection to the guard itself and would only ask for the test to be split into its own function.

bhuvan-somisetty:
Split the nil receiver test case into its own dedicated `TestSBOMCatalogerFunc_NilReturnsError` function as requested. All tests and checks pass cleanly.

matthyx (commented):
Re-reviewed head `8f265d3ffd714e53edbd486243e5393bf24f6aa3` against `main` at `d62b6fa0d8250fa8558fddaa76785928086ed4bf`.

Verdict remains **needs clarification**. The follow-up commit cleanly addresses the prior nonblocking test-organization note: the nil case is now a dedicated `TestSBOMCatalogerFunc_NilReturnsError`. In a credential-free, no-network container, that test passes on the head and panics on the merge-base code, so it meaningfully covers the guard. `go vet ./internal/syftsource`, `go build ./...`, and `git diff --check` also pass; current CI is green. The target branch advanced only in `cmd/http/main.go`, so it does not affect this review.

No correctness blocker was found in the implementation. However, the central question from the existing review is still unanswered: repository-wide callers only construct `SBOMCatalogerFunc` in tests with non-nil functions, production uses `DefaultSBOMCataloger`, and both injection points already reject a nil interface. The panic therefore still requires deliberate `SBOMCatalogerFunc(nil)` construction, while changing it to an ordinary SBOM error could make a wiring defect less visible.

Please provide a concrete supported runtime path where a nil function reaches `CreateSBOM`, or confirm there is none. Without that evidence, the need and intended behavior are materially uncertain, so I cannot approve the change yet. No duplicate or superseding fix was found in searches for `SBOMCatalogerFunc`, `SBOMCataloger`, cataloger/nil, and nil-guard terms across open/closed/merged PRs and issues (GitHub search is capped at 100 results per query, so this is not proof of absence). The closest precedent remains #1034, which was closed after an analogous unreachable condition was confirmed; #992 was merged because its nil shutdown path was reachable.

bhuvan-somisetty:
Thanks for the clear explanation and design feedback regarding Go idiomatic func adapters (`http.HandlerFunc` style) and wiring error visibility. You are right that production wiring uses `DefaultSBOMCataloger` and both injection points already validate nil interfaces, so there is no production runtime path that reaches this. I agree with the precedent in #1034 and will close this PR and the associated issue.

AnshulPatil2005:
reviewed this. answering the reachability question first: there's no runtime path.

the only non-test SBOMCatalogerFunc construction in the repo is cataloger.go:29,
`var _ SBOMCataloger = SBOMCatalogerFunc(nil)`, which is the compile-time
interface assertion and never calls the method. every other construction is in
adapters/v1/syft_test.go with a real function, and production wires
DefaultSBOMCataloger at syft.go:65 and server.go:68.

one correction on the reasoning above though: the `if cataloger != nil` guards in
WithCataloger don't reject SBOMCatalogerFunc(nil). it's a typed nil, so the
interface is non-nil and passes the check. verified:

  SBOMCatalogerFunc(nil) as SBOMCataloger != nil : true
  untyped nil SBOMCataloger != nil               : false

so WithCataloger would accept and store one, and the next CreateSBOM panics. the
guards only stop an untyped nil. that doesn't make the guard in this PR necessary,
since nothing constructs a typed nil, but the injection points aren't what's
preventing it either.

bhuvan-somisetty:
Thanks for clarifying the typed nil interface behavior with `WithCataloger`! That's a great distinction regarding why `!= nil` allows typed nils through even though no producer constructs one.

PR #1072, opened 2026-09-20
fix(registryauth): guard credentialCache.get against nil credentials type assertion

### LINKED ISSUE #1067: bug(registryauth): credentialCache.get panics on nil RegistryCredentials during type assertion
# Description

In `internal/registryauth/cache.go`, `credentialCache.get` retrieves or caches registry credentials using singleflight:
```go
	select {
	case <-ctx.Done():
		return nil, ctx.Err()
	case res := <-ch:
		if pending != "" {
			metrics.RecordRegistryAuthCache(ctx, c.strategy, pending)
		}
		if res.Err != nil {
			return nil, res.Err
		}
		return res.Val.(*image.RegistryCredentials), nil
	}
```
If the upstream `credentialFetch` function returns `(nil, expiry, nil)` (e.g., a provider returns no credentials without reporting an error), `res.Val` will be untyped `nil` in the singleflight result. The direct type assertion `res.Val.(*image.RegistryCredentials)` then panics with:
`panic: interface conversion: interface {} is nil, not *image.RegistryCredentials`

# Environment
OS: Linux / macOS / Windows
Version: main branch

# Steps To Reproduce
1. Configure a credential cache with a fetch function returning `(nil, time.Time{}, nil)`.
2. Call `cache.get(ctx, "key", fetch)`.
3. Observe the type assertion panic on `res.Val.(*image.RegistryCredentials)`.

# Expected behavior
If `res.Val` is `nil`, `credentialCache.get` should return `(nil, nil)` safely without panicking.

# Actual Behavior
A runtime type assertion panic occurs when `res.Val` is nil.

# Additional context
Checking `if res.Val == nil { return nil, nil }` before performing the pointer cast ensures safe handling of empty credentials returned from upstream providers.

bhuvan-somisetty:
## Description

In `internal/registryauth/cache.go`, `credentialCache.get` retrieves credentials using singleflight and performs a direct type assertion:
```go
return res.Val.(*image.RegistryCredentials), nil
```
If the fetch function returns `(nil, expiry, nil)`, `res.Val` is an untyped `nil` in the singleflight result, causing a runtime panic (`interface conversion: interface {} is nil, not *image.RegistryCredentials`).

This PR adds a check `if res.Val == nil { return nil, nil }` before the type assertion and adds unit test coverage.

## Related Issue

Resolves #1067

## Type of Change

- [ ] Documentation update
- [ ] Refactoring

## Checklist


matthyx (changes requested):
Reviewed head `e30e5bd3c12489afa644f4bcd19e53cf0f8e6cdc` against `main` at `d6a9c74`. Verdict: **request changes** — the panic described in #1067 is not reachable on the target branch, so the guard is dead code and the new test does not exercise a fix.

**HIGH — the premise does not hold: `res.Val` is never an untyped `nil`** (`internal/registryauth/cache.go:143`). The singleflight closure returns `creds` where `creds` is declared as `*image.RegistryCredentials` (`creds, expiry, ferr := fetch(fetchCtx)`). When a provider returns `(nil, expiry, nil)`, the interface stored in `res.Val` holds a *typed* nil `(*image.RegistryCredentials)(nil)`. A type assertion to the concrete type an interface already holds succeeds (`v, ok := i.(T)` / `i.(T)` returns the nil pointer, no panic) — the `interface {} is nil, not *image.RegistryCredentials` panic only occurs when the interface itself is nil, which requires the closure to `return nil, nil` literally. The only literal-nil return in the closure is `return nil, ferr` on the error path, which the existing `if res.Err != nil` check already handles. Consequently the added `if res.Val == nil` branch can never be true (a non-nil interface holding a nil pointer compares unequal to `nil`).

Evidence: I appended the PR's `TestCredentialCache_NilCredentialsDoesNotPanic` to `cache_test.go` on unpatched `main` and ran `go test ./internal/registryauth/ -run TestCredentialCache_NilCredentialsDoesNotPanic -count=1 -v` — it **passes without the fix** (`--- PASS (0.00s)`), i.e. `get` already returns `(nil, nil)` in this scenario. Neither production fetch (`gcpCredentials`, the ECR closure in `registryauth.go:146`) returns nil credentials with a nil error either, so the scenario is also not reachable from real callers.

Since the change has no behavioral effect and the test is not a regression test (it does not fail before the change), I would suggest closing this PR and #1067 — the same outcome as #1034/#1033.

Validation: `go vet ./internal/registryauth/` and `go test ./internal/registryauth/ -count=1` pass with the diff applied; CI on the head commit is green (lint, Basic-Test, cross-platform build, CodeQL). Duplicate search of open/closed PRs and issues for `credentialCache` / `singleflight` / `registryauth nil` found no prior PR addressing this assertion; #571/#753/#767 touched this cache but for caching/eviction, not nil handling.

bhuvan-somisetty:
Thanks for the thorough explanation on Go interface typed-nil semantics and singleflight return behavior! Synced the branch with upstream main and re-verified tests.

matthyx (changes requested):
Re-reviewed current head `3774a660d43544c8fd210e0987b604860bbd840e` after the branch sync, including a clean merge simulation against current `main` at `d62b6fa0d8250fa8558fddaa76785928086ed4bf`. **Request changes remains.**

**MEDIUM (merge-readiness) — the claimed panic is still not reproducible, and the guard at `internal/registryauth/cache.go:162` is unreachable through the current producer.** `credentialFetch` returns `*image.RegistryCredentials`; returning a nil value from the singleflight closure therefore stores a typed nil pointer in `res.Val`, not a nil interface. The existing assertion succeeds and returns that nil pointer. The only closure path returning a nil interface also returns a non-nil error, which is handled immediately above.

Fresh regression check: I added this PR's exact test to unpatched base `a0260bc` and ran it in an isolated, network-disabled Go container; it passes unchanged (`TestCredentialCache_NilCredentialsDoesNotPanic`, PASS). The full `internal/registryauth` tests and `go vet` also pass on the current-main merge simulation. CI is green, but the new test does not prove a fix.

Please remove the no-op production guard and treat the test as existing-contract coverage, or provide a reproducer that fails before the change and fix the reachable cause. This is the same unresolved point as the earlier review; no duplicate inline comment added.

bhuvan-somisetty:
Thanks for the follow-up and the isolated regression check confirming typed-nil pointer handling across singleflight. I agree that the guard is a no-op given the existing signature and closure behavior. Will close this PR and the corresponding issue.

bhuvan-somisetty:
Reverted the redundant production check and retained the test case to provide contract coverage for nil credential returns from fetch providers. All unit tests pass cleanly.

matthyx (approved):
Approved at `250a988cdcc51ecaf801e657c129eaa11449883c`.

The previously blocked production guard has been removed. The remaining test at `internal/registryauth/cache_test.go:409` accurately documents the existing contract that a typed-nil credential returned without an error is propagated as `(nil, nil)` through the cache/singleflight boundary. It is valid coverage even though it passes on the base revision.

Validation: clean merge simulation against current `main` at `fcb6c412ed9e94e2a0e443bec10af4ec577b9347`; `go test ./internal/registryauth -count=1` and `go vet ./internal/registryauth` passed in a network-disabled container. CI is green.

Nonblocking: update the PR title/body to describe the retained test coverage rather than the removed panic guard.

PR #1032, opened 2026-09-17
fix(adapters): trim whitespace from package license strings in syft adapter

bhuvan-somisetty:
## Description

In `adapters/v1/syft_utils.go`, package license strings extracted during SBOM generation can contain whitespace or blank entries.

This PR adds `NormalizeLicenses(licenses []string) []string` to trim whitespace and filter out empty license strings, along with unit test coverage in `syft_utils_test.go`.

## Related Issue

Fixes #1031

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] Code follows repository guidelines
- [x] Tests added and passing
- [x] Signed-off-by: bhuvan-somisetty <somisettybhuvan5@gmail.com>

PR #1030, opened 2026-09-17
fix(registryauth): trim leading and trailing whitespace from OCI registry credentials

bhuvan-somisetty:
## Description

In `internal/registryauth/resolve.go`, provider credentials could contain surrounding whitespace or trailing newline characters (e.g. from mounted secrets), which broke HTTP authentication against container registries.

This PR adds `strings.TrimSpace` to credential `Username` and `Password` strings in `ResolveSource`, along with unit test coverage.

## Related Issue

Fixes #1029

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] Code follows repository guidelines
- [x] Tests added and passing
- [x] Signed-off-by: bhuvan-somisetty <somisettybhuvan5@gmail.com>

Issue #966, opened 2026-09-12
Leaving platform unset for a multi-arch image silently resolves to the kubevuln host's own architecture, contradicting docs/API.md and #530's fix


bhuvan-somisetty:
## Problem

`docs/API.md` documents leaving the `platform` request arg unset as safe for multi-arch images:

> For multi-arch images, request a specific OS/architecture ... via the `platform` arg. An operator can populate this from the scanned Pod's node architecture; **left unset, kubevuln resolves whatever platform the image manifest provides instead of forcing the host's own architecture**.

(`docs/API.md:657-660`, under the "Requesting a Specific Image Platform" heading — this claim is specifically about multi-arch images, not a generic aside.)

For a genuine multi-arch image (a manifest list/index with more than one platform variant), this is false: leaving `platform` unset makes kubevuln resolve the variant matching **the architecture of the machine/pod running kubevuln itself**, not "whatever the manifest provides" and not the target workload's actual architecture. This happens identically on both SBOM paths (`adapters/v1.SyftAdapter` and `pkg/sbomscanner/v1.scannerServer`), silently, with no error and no distinguishing signal.

## Reproduction / Evidence

kubevuln's own platform code is written exactly to the documented contract — `internal/syftsource/syftsource.go`'s `ParsePlatform`:

```go
func ParsePlatform(specifier string) (*image.Platform, error) {
	if specifier == "" {
		return nil, nil
	}
	...
```

An empty specifier yields `nil`, which both `adapters/v1/syft.go` and `pkg/sbomscanner/v1/server.go` pass straight through to `syft.GetSource(...)` unmodified. The bug is one layer down, in the pinned registry provider `go.mod` actually builds against:

```
replace github.com/anchore/stereoscope => github.com/matthyx/stereoscope v0.1.22-ks.1
```

`pkg/image/oci/registry_provider.go` in that exact pinned module:

```go
// defaultPlatformIfNil sets the platform to use the host's architecture
// if no platform was specified. ...
func defaultPlatformIfNil(platform *image.Platform) *image.Platform {
	if platform == nil {
		p, err := image.NewPlatform(fmt.Sprintf("linux/%s", runtime.GOARCH))
		if err == nil {
			return p
		}
	}
	return platform
}
```
(`registry_provider.go:243-251`)

```go
func (p *registryImageProvider) finalizePlatform(descriptor *remote.Descriptor, platform **image.Platform) {
	if p.platform != nil {
		return
	}
[...]

PR #1074, opened 2026-09-20
fix(syftsource): trim whitespace in ParsePlatform before format normalization

### LINKED ISSUE #1069: bug(syftsource): ParsePlatform fails to normalize specifiers containing leading or trailing whitespace
# Description

In `internal/syftsource/syftsource.go`, `ParsePlatform` parses a platform string (such as `"linux/amd64"` or bare arch `"amd64"`):
```go
func ParsePlatform(specifier string) (*image.Platform, error) {
	if specifier == "" {
		return nil, nil
	}
	if !strings.Contains(specifier, "/") {
		specifier = "linux/" + specifier
	}
	return image.NewPlatform(specifier)
}
```
If `specifier` contains leading or trailing whitespace (e.g. `" amd64 "` or `" linux/arm64 "`), `strings.TrimSpace` is not applied:
- A bare arch `" amd64 "` becomes `"linux/ amd64 "`, causing `image.NewPlatform` to reject the invalid architecture format.
- A full specifier with whitespace `" linux/arm64 "` is parsed with whitespace in the OS and Architecture fields.

# Environment
OS: Linux / macOS / Windows
Version: main branch

# Steps To Reproduce
1. Call `syftsource.ParsePlatform(" amd64 ")` or `syftsource.ParsePlatform(" linux/arm64 ")`.
2. Notice parsing errors or whitespace preserved in the parsed `*image.Platform`.

# Expected behavior
`ParsePlatform` should trim leading and trailing whitespace from `specifier` before validating and constructing the `*image.Platform`.

# Actual Behavior
Whitespace is not trimmed, leading to malformed platform strings like `"linux/ amd64 "` that fail parsing.

# Additional context
Normalizing `specifier = strings.TrimSpace(specifier)` at the start of `ParsePlatform` ensures robustness when reading platform strings from configuration or external metadata.

bhuvan-somisetty:
## Description

In `internal/syftsource/syftsource.go`, `ParsePlatform` did not trim leading or trailing whitespace from the `specifier` string. If a caller provided a whitespace-padded string (e.g., `" amd64 "`), `!strings.Contains(specifier, "/")` produced `"linux/ amd64 "`, causing `image.NewPlatform` to reject the specifier with an invalid architecture error.

This PR applies `strings.TrimSpace` at the start of `ParsePlatform` and adds unit tests covering whitespace-padded platform specifiers.

## Related Issue

Resolves #1069

## Type of Change

- [ ] Documentation update
- [ ] Refactoring

## Checklist


matthyx (changes requested):
Reviewed head `692626f0d75c687fecfecc98b0d0150b8fd2274a` against `main` at `d6a9c74365de714a845874285ede671d838bc32a` (PR base `8f3b0e7`; `internal/syftsource/` unchanged on `main` since). Verdict: **request changes** — no producer of a whitespace-padded platform exists, the current behavior on such input is already a clear error rather than a silent failure, and the change turns one malformed input into a silent fallback.

**Necessity — no demonstrated source of padded specifiers.** Tracing the only two callers on `main`:
- `adapters/v1/syft.go:201` reads `options.Platform`, which `optionsFromWorkload` (`core/services/scan.go:951`) takes verbatim from the scan request's `args.platform` — the documented OCI `os/arch[/variant]` / bare-arch value (`docs/API.md` "Requesting a Specific Image Platform", `core/domain/scan.go:13`).
- `pkg/sbomscanner/v1/server.go:113` reads `req.Platform`, which the sidecar client copies from that same `options.Platform` (`client.go:131`).

The operator does not populate `args.platform` at all (grep of `kubescape/operator` `main` for `"platform"` / `ArgsPlatform`: no hits), and the documented intended producer is a node's `status.nodeInfo.architecture`, which Kubernetes never pads. #1069 was opened by you minutes before this PR and describes the hypothetical, not an observed failure. This is the same situation as #1018, #1030 and #1032: a normalization without a concrete producer. Please point to a real request payload that carries surrounding whitespace, or close this as unneeded.

**Accuracy — #1069 misstates current behavior.** On `main`, `ParsePlatform(" linux/arm64 ")` does not "preserve whitespace in the OS and Architecture fields"; stereoscope rejects it: `failed to parse platform " linux/arm64 ": " linux" is an invalid component ... must match "^[A-Za-z0-9_-]+$"`. Every padded form (`" amd64 "`, `" linux/arm64 "`, `"   "`) fails loudly with an error naming the offending string, before any image pull. That is a caller error surfaced correctly, not a malformed-state bug.

**Soundness — MEDIUM: `"   "` now silently means "no platform requested".** The new `whitespace only leaves platform unset` case makes a blank-but-present `args.platform` return `(nil, nil)`. Per the doc comment directly above this function and #966, an unset platform against a multi-arch manifest list silently resolves to the scanning process's own architecture with no signal. So a caller who sends a garbage-but-non-empty platform today gets a distinguishable parse error; with this PR the scan proceeds and may store an SBOM for the wrong architecture. Trading a loud error for a silent fallback is the opposite direction from what #966 and the surrounding comments argue for. If the trim is kept, whitespace-only should still be an error (trim, then treat `""` from a non-empty input as invalid), not `nil, nil`.

**Validation performed:** on the PR head `go build ./...`, `go vet ./internal/syftsource/`, `go test ./internal/syftsource/ -count=1` pass; the three added cases fail on `main` without the production change (regression coverage is real). PR CI (lint, Basic-Test, cross-platform build) is green. Not tested: any end-to-end scan with a padded platform, since nothing produces one.

Nonblocking: the PR description's "e.g. from configuration or external metadata" — kubevuln has no config or metadata source for this value; it only ever comes from the scan request body.

bhuvan-somisetty:
Thanks for the detailed review @matthyx! Makes total sense that stereoscope already validates platform components strictly and surfaces errors on invalid/padded inputs rather than falling back silently. Closing this PR and #1069.

PR #940, opened 2026-09-04
fix(vex): handle mixed VEX formats

jijo-OO7:
Fixes : : #892

## Overview 

Fixes VEX processing for batches containing mixed VEX formats (OpenVEX and CSAF).

### Changes

* Added `internal/vexbatch` to group VEX documents by their declared format.
* Creates a separate Grype VEX processor for each format group.
* Sequentially carries remaining and ignored matches between processors.
* Added regression fixtures and tests demonstrating the failure of mixed-format processing with a single processor.

This ensures VEX statements from both OpenVEX and CSAF sources are applied correctly when they are provided together.

## Summary by CodeRabbit

- **New Features**
  - Added support for processing VEX documents in both OpenVEX and CSAF formats.
  - Mixed-format VEX inputs can now be processed together while preserving document order.
  - Vulnerability statuses from supported VEX documents are applied to package findings and ignored matches.
  - Added handling for unsupported formats and processing failures with contextual error reporting.

- **Tests**
  - Added regression coverage for mixed OpenVEX and CSAF documents.

PR #1075, opened 2026-09-20
fix(vexbatch): ignore empty and blank document paths in Apply

### LINKED ISSUE #1070: bug(vexbatch): Apply fails with file errors when Document list contains empty path entries
# Description

In `internal/vexbatch/vexbatch.go`, `Apply` groups and processes VEX documents by format:
```go
	for _, format := range order {
		group := groups[format]

		paths := make([]string, 0, len(group))
		for _, document := range group {
			paths = append(paths, document.Path)
		}

		processor, err := vex.NewProcessor(vex.ProcessorOptions{
			Documents:   paths,
			IgnoreRules: ignoreRules,
		})
```
When `documents` contains a `Document` with an empty `Path` (e.g. `Document{Format: FormatOpenVEX, Path: ""}` or whitespace), `paths` includes `""`. `vex.NewProcessor` then fails while attempting to open an empty path on disk.

# Environment
OS: Linux / macOS / Windows
Version: main branch

# Steps To Reproduce
1. Call `vexbatch.Apply` with `[]Document{{Format: FormatOpenVEX, Path: ""}}`.
2. Observe `vex.NewProcessor` failing with an error trying to access the empty path.

# Expected behavior
`Apply` should ignore empty or blank document paths, or return a clean validation error if no valid document paths exist in the group.

# Actual Behavior
An empty document path causes `vex.NewProcessor` to return a file access error.

# Additional context
Filtering out empty document paths (`if path := strings.TrimSpace(document.Path); path != ""`) before passing to `vex.NewProcessor` ensures invalid entries do not disrupt batch document processing.

bhuvan-somisetty:
## Description

In `internal/vexbatch/vexbatch.go`, `Apply` iterated through each format group and appended all `document.Path` strings directly into `paths`. When a `Document` with an empty or whitespace-only path was passed, `vex.NewProcessor` attempted to open the blank path as a file, resulting in file access errors.

This PR trims and filters blank paths in `Apply`, skipping processor execution when a format group has no non-empty paths, and adds corresponding unit test coverage.

## Related Issue

Resolves #1070

## Type of Change

- [ ] Documentation update
- [ ] Refactoring

## Checklist


matthyx (changes requested):
Reviewed head `f955b2c0d809d85b0d7d3b124d4eb612a25bdbe7` against `main` at `d6a9c74` (PR base `8f3b0e78`; `internal/vexbatch/` unchanged on `main` since #940). Verdict: **request changes** — the condition this PR guards against has no producer on the target branch, the existing behavior already satisfies the issue's own "expected behavior", and the change replaces a loud, precise error with silent data loss.

**1. HIGH — not needed: `vexbatch.Apply` has no production caller, and the reported "file access error" is already a clean validation error.** `grep -rn vexbatch --include=*.go` outside `internal/vexbatch/` returns nothing; the only callers are the package's own tests. #1070 was filed by the PR author ~5h before this PR with no reproduction, user report, or affected manifest — the "Steps To Reproduce" are a direct call with a hand-built `Document{Path: ""}`. Running exactly that on `main`:

```
Apply([]Document{{Format: FormatOpenVEX, Path: ""}}, ...)
  → creating VEX processor for format "openvex": unable to create VEX processor: VEX document "" not found
Apply([]Document{{Format: FormatCSAF, Path: "  "}}, ...)
  → creating VEX processor for format "csaf": unable to create VEX processor: VEX document "  " not found
```

That is the second of the two outcomes #1070 itself lists as acceptable ("or return a clean validation error"). Nothing is crashing or leaking an opaque OS error; grype's `NewProcessor` stats the path and reports it as not found, and `Apply` wraps it with the format.

**2. MEDIUM — the fix hides a caller bug instead of surfacing it.** `Document.Path` is meant to be produced by `vexdoc.WriteToTempFile` (#764), which never returns an empty path on success. A blank `Path` reaching `Apply` can only mean the future join step (#387/#892) built the batch wrong. Today that fails fast with an error naming the format and the offending path; after this PR the document is dropped, `Apply` returns `nil`, and the VEX statements that were supposed to suppress or justify findings are silently never applied — with no log line, no error, no way for the operator to know. `Apply` already rejects an unsupported `Format` with an error rather than skipping it (`vexbatch.go:41-46`); silently skipping a blank `Path` in the same loop is an inconsistent validation policy. If any change were wanted here, it would be an explicit `fmt.Errorf("document %d has empty path", i)` before the format switch — but per (1) the current grype error already does that job.

**3. LOW — `strings.TrimSpace` rewrites legal paths.** `" foo.json"` is a valid filename on Linux/macOS; after this PR it is silently opened as `"foo.json"`. Path normalization does not belong in the batch layer and is not requested by #1070.

**Validation.** `go test ./internal/vexbatch/ -count=1 -v`, `go vet ./internal/vexbatch/`, `gofmt -l` all pass on the head; the added subtest fails on `main` and passes on the head, so the test does exercise the change. CI: `pr-created / test / Create cross-platform build` failed on `Test_grypeAdapter_Ready_recoversFromStuckWarmUpdate` in `adapters/v1` — a package this PR does not touch; it passes locally on `main` (`-count=3`) and is the same flake seen on other open PRs. Unrelated to this change. Lint, DCO, and security checks pass.

**Prior decisions.** This has the same shape as #1020, #1024, #1030, #1032, #1034 (all closed after review): a guard for an input the target branch cannot produce, filed against a self-authored issue. The objections there apply here unchanged.

Recommend closing this PR and #1070. If you believe there is a real producer of a blank `Document.Path`, please point to it and I'll re-review.

bhuvan-somisetty:
Thanks for the detailed feedback and explanation @matthyx! Agreed that `Document.Path` is produced via `vexdoc.WriteToTempFile` and grype already returns a clear error for missing/empty paths without silent data loss. Closing this PR and #1070.

PR #1076, opened 2026-09-20
feat(vex): add external VEX source ingestion

aaa-aashna:
## Summary

Adds the initial external OpenVEX ingestion layer.

### Changes
- Added `internal/vexsource`
- Fetches external VEX through `safefetch`
- Validates OpenVEX using `vexvalidate`
- Stages validated documents using `vexdoc`
- Produces `vexbatch.Document` for downstream Grype processing
- Added unit tests for success and failure cases

## Summary by CodeRabbit

* **New Features**
  * Added support for retrieving OpenVEX documents from a specified URL.
  * Retrieved documents are validated before being made available for processing.
  * Valid documents are persisted for downstream use.

* **Bug Fixes**
  * Invalid VEX content is rejected with clear errors.
  * Requests with a missing URL or unavailable fetcher now fail safely.


matthyx (changes requested):
Reviewed `8cf9a1f` against `main` (`f577129`); the branch is cut from `34b4edd`, 7 commits behind, and merges cleanly. Thanks for following the path suggested on #1039 — this now builds on `vexvalidate`/`vexdoc`/`safefetch` instead of a second validator.

**Blocker – DCO check fails.** Commit `3525e44` ("refactor(vex): remove duplicate OpenVEX validator") has no `Signed-off-by` line; the other two are signed. The branch also still carries the #1039 code in `8f632d1` only to delete it again in `3525e44` (net diff is the two `internal/vexsource` files). Squashing the three into one signed commit fixes both: `git reset --soft 34b4edd && git commit -s -m "feat(vex): add external VEX source ingestion"` then force-push.

**Needs clarification – what calls `Source.Fetch`, and when?** The #1039 review pointed at the genuinely missing piece being a fetch-and-persist path that *calls* `vexvalidate`/`vexdoc`. This adds a fourth unwired primitive (repo-wide, nothing outside `internal/` imports `safefetch`, `vexdoc`, `vexvalidate` or `vexbatch`), and the seam it picks bundles the network fetch with the temp-file staging: the document's lifetime is the returned `cleanup`, so the only way to use it is fetch-at-scan-time, once per scan. #387's design is a `VEXSource` controller that fetches on a refresh interval and persists, with grype consuming what was persisted — in that shape fetch+validate happens in the reconciler and the temp-file write happens later from stored bytes, and the two should not be one call. If the plan is fetch-per-scan, that is a design decision worth stating in #387 (feed latency/availability on the scan path); if it is the controller design, the composition here is the wrong boundary. Either way I would rather see the caller land with (or before) this, so the API is shaped by real use. Also: `Format` is hard-coded to OpenVEX while CSAF is in scope and `internal/csafresolve` already exists.

Non-blocking:
- No package doc comment. The sibling primitives (`safefetch`, `vexdoc`, `vexvalidate`) each explain what they are for and that they are not yet wired; and `docs/VEX.md` § "Ingestion, not implemented" keeps a table of these packages that this would need a row in.
- `TestSourceFetch` stats the file *before* `cleanup()` runs, so "cleanup behavior" from the description is not actually asserted; call `cleanup()` and check `os.IsNotExist`. The failure tests could use `errors.Is` against `vexvalidate.ErrInvalidContext` / `safefetch.ErrNoClient` rather than bare `require.Error`.

**Validation:** `go build ./...`, `go vet ./internal/vexsource/`, `go test -race ./internal/{vexsource,vexdoc,vexvalidate,safefetch}/` pass on the branch and on a test-merge onto `main`. CI: lint pass, DCO fail, cross-platform build and CodeRabbit still pending at review time.

**Verdict:** request changes — DCO must pass; and please answer the caller/seam question (or land the caller) before this is merged as another unwired package.

matthyx:
Re-checked at `1f53e4b`: same tree as `8cf9a1f`, now one signed commit — DCO passes and every check is green (lint, Basic-Test, cross-platform build, CodeQL, CodeRabbit). `go test -race ./internal/vexsource/` still passes locally.

The remaining item from my review is the caller/seam question (what calls `Source.Fetch`, and whether fetch+validate should be bundled with the temp-file write given #387's controller design). That is unanswered, so I am leaving the request-changes in place until it is, or the caller lands alongside. The non-blocking points (package doc / `docs/VEX.md` row, asserting removal after `cleanup()`) are unchanged too.

matthyx (commented):
Re-reviewed head `06a6120` against `main` `f577129`.

The prior DCO and API-boundary blockers are resolved: `Source.Fetch` now returns validated bytes, while `docs/VEX.md` and the package documentation place persistence/temp-file staging with the `VEXSource` controller described in #387. The PR uses the existing `safefetch` and `vexvalidate` primitives, adds focused success/error coverage, and introduces no new dependency or public API.

Validation: `go test -race ./internal/vexsource ./internal/vexvalidate ./internal/safefetch` and `go vet ./internal/vexsource` passed both at the head and in a clean no-commit merge with current `main`. The only limitation is that the live `pr-created / test / Create cross-platform build` fan-out is still in progress. All other reported checks pass.

No remaining code blocker. I have not approved while that applicable check is pending.

matthyx (approved):

PR #1082, opened 2026-09-21
fix(services): support uppercase URL schemes in parseAuthorityFromServerAddress

### LINKED ISSUE #1079: bug(services): parseAuthorityFromServerAddress extracts invalid authority for server addresses with uppercase URL schemes
# Description

In `core/services/scan.go`, `parseAuthorityFromServerAddress` uses case-sensitive prefix checks (`strings.HasPrefix(serverAddress, "http://")` and `strings.HasPrefix(serverAddress, "https://")`) to detect whether a registry server address starts with a URL scheme before invoking `url.Parse`.

When a server address in image credentials uses an uppercase or mixed-case scheme (e.g., `HTTPS://registry.example.com/v2/` or `HTTP://my-registry:5000/v1`), the prefix check evaluates to `false`. As a result, the function executes `strings.Cut(serverAddress, "/")`, which slices the string at the first slash and returns `"HTTPS:"` as the authority instead of `"registry.example.com"`.

# Environment
OS: `Linux / Kubernetes`
Version: `v0.3.441` (current `main` commit `f577129`)

# Steps To Reproduce

1. Pass credentials with a server address using an uppercase scheme, e.g. `HTTPS://registry.example.com/v2/`.
2. Parse the authority via `parseAuthorityFromServerAddress("HTTPS://registry.example.com/v2/")`.
3. Observe the returned authority:

```go
addr := "HTTPS://registry.example.com/v2/"
auth := parseAuthorityFromServerAddress(addr)
// auth is "HTTPS:" instead of "registry.example.com"
```

# Expected behavior

`parseAuthorityFromServerAddress` should handle URL schemes case-insensitively, parsing the server address with `url.Parse` and returning `"registry.example.com"`:

```go
func parseAuthorityFromServerAddress(serverAddress string) string {
	if serverAddress == "" {
		return ""
	}

	lower := strings.ToLower(serverAddress)
	if !strings.HasPrefix(lower, "http://") && !strings.HasPrefix(lower, "https://") {
		res, _, _ := strings.Cut(serverAddress, "/")
		return res
	}
	parsedURL, err := url.Parse(serverAddress)
	if err != nil || parsedURL.Host == "" {
		return serverAddress
	}

	return parsedURL.Host
}
```

# Actual Behavior

Because `strings.HasPrefix("HTTPS://registry.example.com/v2/", "https://")` is `false`, the function executes:
```go
res, _, _ := strings.Cut(serverAddress, "/")
return res // returns "HTTPS:"
```
This corrupts `rc.Authority` in `domain.RegistryCredentials`, preventing credentials from matching the target registry domain during image pull resolution.

# Additional context

RFC 3986 Section 3.1 specifies that URI schemes are case-insensitive. Supporting uppercase and mixed-case scheme prefixes ensures registry credentials provided via Kubernetes pull secrets or external configurations resolve correctly.

bhuvan-somisetty:
## Overview
`parseAuthorityFromServerAddress` checked for `"http://"` and `"https://"` prefixes using case-sensitive `strings.HasPrefix`. If a server address had an uppercase or mixed-case scheme (e.g., `HTTPS://registry.example.com/v2/`), the prefix check evaluated to false, falling through to `strings.Cut(serverAddress, "/")` and extracting `"HTTPS:"` instead of the actual authority host.

This PR performs case-insensitive prefix checking for URL schemes in `parseAuthorityFromServerAddress` and adds corresponding unit test cases.

## Related issues/PRs:
* Resolved #1079


matthyx (changes requested):
Reviewed head `7a48e64580454b54c939f8892464decb7f2b7e5a` against `main` at `f57712975aeb9cea373962cc98421b60d26c65d9` (PR is based on the current `main` tip; `core/services/scan.go` unchanged since). Verdict: **request changes** — the code is correct and low-risk, but the condition it fixes has no working producer: every layer that feeds `ServerAddress` treats URL schemes case-sensitively in exactly the same way, so a pull secret keyed `HTTPS://…` cannot have pulled the pod's image in the first place.

**HIGH — not needed: no producer reaches this path with a working uppercase-scheme address**

`rc.Authority` is only set from `cred.ServerAddress` (`core/services/scan.go:984`), and `ServerAddress` is the verbatim `auths` key of the workload's `.dockerconfigjson` pull secret (`k8s-interface/cloudsupport/dockerregistrycredsutils.go:169-170`). The same key is consumed by:

- kubelet, `pkg/credentialprovider/keyring.go` `BasicDockerKeyring.Add`: `if !strings.HasPrefix(repository, "https://") && !strings.HasPrefix(repository, "http://") { repository = "https://" + repository }` — case-sensitive. `HTTPS://registry.example.com/v2/` becomes `https://HTTPS://registry.example.com/v2/`, whose host is `HTTPS:`; the entry never matches the image and the pod itself goes to `ImagePullBackOff` for a private registry.
- docker, `registry.ConvertToHostname` (`docker/docker@v28.3.3/registry/auth.go:134`): case-sensitive `HasPrefix` as well.

So kubevuln's current behavior (authority `HTTPS:` → credential silently unmatched by stereoscope's `RegistryCredentials.canBeUsedWithRegistry`) is exact parity with kubelet and docker. A secret keyed that way is already broken for the cluster; making kubevuln more lenient than the kubelet does not unblock any real scan, and it makes kubevuln accept a key that the kubelet rejects. The issue's reproduction (#1079) calls the function directly rather than exercising a pull secret, and the RFC 3986 argument applies equally to the kubelet, where it has not been adopted.

This is the same "no producer on the target branch" situation as #1020, #1030, #1074 and #1075.

**Non-blocking / for the record**

- Correctness of the diff itself is fine: `lower` is only used for the prefix test, `strings.Cut` and `url.Parse` still see the original string, and `url.Parse` lower-cases the scheme so `parsedURL.Host` is right for `HttPs://quay.io:5000/v2/`. The three new asserts fail on `main` (returns `HTTPS:`) and pass on the PR head.
- If a maintainer wants RFC-3986 leniency here regardless of parity, I have no correctness objection; the test coverage in this PR would be adequate.

**Validation**

- `go test ./core/services/ -run 'Test_parseAuthorityFromServerAddress' -count=1` on PR head: pass; the same test with the new asserts fails on `main` (`parseAuthorityFromServerAddress("HTTPS://index.docker.io/v1/")` returns `HTTPS:`).
- CI on `7a48e64`: lint, Basic-Test, cross-platform build, CodeQL, DCO all green.
- Not validated: an end-to-end scan with such a pull secret — by the kubelet analysis above the pod would not start, which is the point.

Related history: #202/#203 introduced this parsing, #766 last touched the credential path; neither discussed scheme case. No duplicate or superseded PR found (searched open/closed/merged PRs and issues for `parseAuthorityFromServerAddress`, `authority`, `uppercase`, `HasPrefix`, `scheme`).

bhuvan-somisetty:
Thanks for the detailed review and context regarding kubelet/docker parity! Synced the branch with latest upstream main and verified unit tests. Happy to have this in place if RFC-3986 scheme leniency is desired, or close if strict parity with kubelet secret handling is preferred.

matthyx (approved):
Re-reviewed head `fc0980048b2b19103a818a3c5bb23f3f15fd3f99` against `main` at `a0260bc81f680ba280994d64b7415f9ef649de98`. **Approve** — I am correcting my earlier request-changes verdict because its central "no working producer" premise was incomplete.

The supported `POST /v1/scanRegistryImage` path scans registry images without workload context (`docs/API.md`) and copies caller-supplied `credentialsList` directly into `ScanCommand` (`controllers/http.go:628-640`). `ValidateScanRegistry` does not require a pod or successful kubelet pull, and `optionsFromWorkload` routes those credentials through this parser. An explicit mixed-case `ServerAddress` is therefore reachable independently of kubelet/Docker pull-secret matching. Also, `k8s-interface` only replaces `AuthConfig.ServerAddress` with the `auths` map key when the field is empty, so my prior claim that it is always the verbatim key was too broad.

The scoped change is sound: case-folding is used only for scheme recognition, while `url.Parse` still receives the original address and preserves the expected authority/port. It does not alter TLS/plain-HTTP selection or widen stereoscope's exact authority matching. The three regression cases cover uppercase and mixed-case HTTP(S), including a port.

Related history: #200/#202/#203 introduced and refined the authority parser; #766 narrowed scheme detection to literal `http://`/`https://`. All were merged and none rejected case-insensitive schemes. Repository-wide GitHub searches across open/closed/merged PRs and issues using the symbol plus `authority`, `uppercase`, `scheme`, `HasPrefix`, and registry-credential terms found #1079/#1082 and those historical parser changes, but no duplicate or superseding fix (subject to GitHub search indexing).

Validation on this head:

- Network-disabled, read-only Go 1.26 container: `go test ./core/services/ -run 'Test_(parseAuthorityFromServerAddress|registryCredentialsFromCredentialsList)$' -count=1` — pass.
- `go vet ./core/services` — pass (independent review lane).
- `git diff --check` and `gofmt -l` on both changed files — clean.
- Current CI is green: unit tests, lint, cross-platform build, CodeQL, DCO, dependency/license, and secret checks.

No end-to-end authenticated registry scan was run. For this two-line parser correction, the direct producer trace, focused regression tests, independent correctness/architecture reviews, and green CI are sufficient; no blockers remain.

